PR #55, opened 2015-01-09
Replace custom router with react-router

brian-c:
The new-project an view-project pages were broken up into separate modules by this, which was very much needed.

I still think react-router is pretty horsey to configure (not as horsey as I originally thought), but I really like the way this turned out. Linking by page names is awesome. Big :+1: for something I thought would be :unamused:.

Lots of changes, but not much is actually different. Let me know if anything doesn't makes sense.


aweiksnar:
Looks fantastic, the named-routes and entry file are super readable

PR #66, opened 2015-01-12
Use new translate component

### LINKED ISSUE #64: Switch translations to react-translate-component
In the interest of maintaining less stuff, check out martinandert/react-translate-component

brian-c:
This replaces the kludgey custom `react-translator` component with the very nice `react-translate-component`, which also allows for nice things like date localization and pluralization.

I haven't translated everything (especially things I think will be replaced soon), but this should serve as the model for the future.

Translating things with markup in them is not possible because React doesn't actually do anything with markup. You can, however, interpolate on React components. It looks a little weird but it makes sense:

https://github.com/zooniverse/Panoptes-Front-End/blob/3649ba22e0ec266a53bced414ca321cd5bd1192d/app/partials/register-form.cjsx#L147-L148

Also worth pointing out that the interpolation placeholder is `%(key)s`, with a weirdo "s" at the end.

Closes #64.


brian-c:
@aweiksnar #67 fixes the eval/use-strict issue. Picked up that counterpart dep and snuck in a change to the polyfills file that gets downloaded (IE6's break Chrome, and IE<9 can't render SVGs anyway).

aweiksnar:
Sweet, that looks like a nice translations component

PR #79, opened 2015-01-23
Update the API layer

brian-c:
I made some big changes to the json-api-client lib to make it more predictable and less clumsy to work with.

The JSON-API lib no longer attempts to handle errors on its own, instead promises for failed requests are simply reject with the request. The rejected requests are no processed by the api client instance (/app/api/client) to work around the fact that different back end modules pass errors in different ways. A bad request now now consistently throws an `instanceof Error`, with a simple string `message` property (multi-line if needed) that's the best attempt at how to express the error.

These changes also lead to some refactoring of the auth module and the PromiseToSetState component mixin.

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,7 +41,7 @@
     "postinstall": "npm run get-fallback-polyfills && npm run get-font-awesome",
     "stage": "npm run build && publisssh ./build demo.zooniverse.org/panoptes-front-end",
     "start": "./bin/serve.sh",
-    "test": "browserify ./test/runner.coffee --extension .coffee --transform coffee-reactify --transform envify | testling \"${TESTLING_ARGS:---pass}\" | tap-spec",
+    "test": "browserify ./test/runner.coffee --extension .coffee --transform coffee-reactify --transform envify --ignore-transform coffeeify | testling \"${TESTLING_ARGS:---pass}\" | tap-spec",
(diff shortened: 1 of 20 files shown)


package.json
```diff
@@ -31,9 +30,12 @@
     "uglify-js": "^2.4.16",
     "watchify": "^2.2.1"
   },
+  "peerDependencies": {
+    "coffeeify": "^1.0.0"
```
aweiksnar: I haven't been able to pick up coffeeify as a dependency on a fresh `npm i`, is yours locating it okay?

brian-c: Try `npm link` to link it globally, then `npm link panoptes` in another directory.

You don't need coffeeify unless you're requiring this as a lib in another project.

aweiksnar: I think the browserify transform in package.json is causing it to look for it. Running `npm start` is throwing `Error: Cannot find module 'coffeeify'`

aweiksnar: adding `  --ignore-transform coffeeify \ ` to the watchify command in bin/serve.sh seems to fix it

aweiksnar: Seems to be the case for all of the browserify calls in bin,  `test` & `build` we're the same

brian-c: Ugh, yeah, that makes sense. I just removed coffeeify and I get the same thing.

I'll push a fix in a minute.

aweiksnar:
Cool, everything's running smoothly now. New error handling code looks really nice

PR #85, opened 2015-01-28
Updated classification interface

brian-c:
This won't actually save classifications until zooniverse/Panoptes#449 is fixed, but I'm starting work on the aggregate viewer, which is based on this, and merging all that at once would be pretty painful, so I'd like to get this in now.

Everything should be okay except the final "next" button.

Dev view at #/dev/classifier


aweiksnar:
Wow I could see the polygon tool being really popular, super clever

PR #129, opened 2015-02-27
Phantomjs update

astopy:
Note that this increases the Docker build time to an hour (or more), so we'll want to change this once binaries are available.


brian-c:
@astopy Other than being slow, is this good to merge?

astopy:
Yup, should be.

Issue #120, opened 2015-02-20
Let the user know when a project is complete


brian-c:
Project Home page should show a message when all data has been classified and link to the Results page instead of the Classify page.

Classifier should have a message too, but it should still work (skip Cellect?).

brian-c:
See zooniverse/Panoptes#571, zooniverse/Panoptes#565

PR #135, opened 2015-03-02
Show a warning on .already_seen and .retired subjects for #120

brian-c:
This'll show a little flag on retired subjects and subjects you've already seen.

Depends on zooniverse/Panoptes#571


brian-c:
@edpaget It looks like zooniverse/Panoptes#571 is merged and deployed, but a request for subjects still comes back empty after a user has been through them all. Any ideas?

brian-c:
This can close #312, too. Will rebase after the current round of merges.

brian-c:
Will also close #368.

I think this is just waiting on zooniverse/Panoptes#857, right @edpaget?

brian-c:
Rebased into #382.

PR #137, opened 2015-03-02
Cache busting

brian-c:
Rename JS and CSS files with their hash on build.

index.html is now an ERB template (kinda weird, but I guess Handlebars can't render from the CLI?) that renders with the environment variables from the script config.

This could be cleaned up a lot in the future.


public/index.erb
```diff
@@ -8,10 +8,13 @@
     <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.2.0/css/font-awesome.min.css" />
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Montserrat">
     <link rel="stylesheet" href="//fonts.googleapis.com/css?family=Open+Sans:400,400italic,700,700italic" />
-    <link rel="stylesheet" href="./main.css" />
+    <link rel="stylesheet" href="./<%= ENV['OUT_CSS'] %>" />
   </head>
 
   <body>
+    <!--Last build: <%= Time.now %>-->
```
aweiksnar: Clever!

PR #138, opened 2015-03-02
Expose function to get subject image location

brian-c:
Just exposing the function that finds the locations we know how to display from a subject.


aweiksnar:
Sweet!

PR #149, opened 2015-03-18
Parse metadata files

rogerhutchings:
I've added BabyParse for parsing metadata files and it all seems to be working. There are two things outstanding:

- There's a decision to be made on the manifest schema to determine how to group images as a single subject
- It could stand to have some design work done, as there's going to a be a lot of data on there. It'll need to be paginated (via @brian-c), and probably tabbed (one for subjects + manifest, loose and missing)

Unfortunately I need to get back on to Tate stuff, so I'm submitting a PR for now.


brian-c:
Handled by the new lab subject page

PR #157, opened 2015-03-20
Switch Markdown renderer to MarkdownIt

brian-c:
Thoughts on this @aweiksnar? It's ~45k zipped, which is kinda huge (Marked was ~10, I think), but it's super nice. I think there are plugins for hashtags and @-mentions, and I added custom containers and some extra formatting bits in there for Stargazing.


aweiksnar:
Great find! looks like this engine correctly differentiates hashtags and top-level headers from the space character in-between too, which fixes an issue I was having with marked. I took a quick browse through the plugins too and seems like there's plenty out there.

PR #160, opened 2015-03-20
Fix subject sizing

brian-c:
Just some CSS to improve classify page layout


brian-c:
Revisit when we figure out how we want layout to work.

PR #162, opened 2015-03-21
Refresh bearer tokens

### LINKED ISSUE #154: Keep track of sessions timing out
Use `expires_in` and `refresh_token`.

https://github.com/doorkeeper-gem/doorkeeper/wiki/Enable-Refresh-Token-Credentials

brian-c:
Sessions expire after two hours. This refreshes the bearer token ten seconds before that.

Closes #154.

### DIFF
--- a/app/api/auth.coffee
+++ b/app/api/auth.coffee
@@ -32,7 +32,7 @@ module.exports = new Model
       .then (request) ->
         [_, authTokenMatch1, authTokenMatch2] = request.responseText.match CSRF_TOKEN_PATTERN
         authToken = authTokenMatch1 ? authTokenMatch2
-        console?.info 'Got auth token', authToken
+        console?.info "Got auth token #{authToken[...6]}..."
         authToken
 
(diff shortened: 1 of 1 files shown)


aweiksnar:
Thoughts on emitting a refresh-token type event when this happens? Right now i'm listening to user change signals to update the bearer-token for talk auth (branched), which uses the panoptes bearer-token for authentication

brian-c:
I'd rather token handling not be exposed (I was actually planning on making it all private in the future).

Here (https://github.com/zooniverse/Panoptes-Front-End/blob/talk-int/app/api/talk.coffee#L15-L24) I believe you can just set `talkClient.headers = authClient.headers` and the Talk client will make requests using whatever token is set for the API client. That's a bit of a hack, but we can formalize some way for two API clients to share auth details later.

@parrish Are talk-related resources not going to be accessed through the main API, or is this a temporary thing?

aweiksnar:
Okay cool, that works for me

PR #166, opened 2015-03-27
Add lab pages

brian-c:
From `/lab`, you can now create and delete projects. This will be the main project editor stuff. Basic content editing is implemented (mostly via a general `BoundResourceMixin`) and you can add and edit `ProjectRole`s. Deleting them is currently broken (zooniverse/Panoptes#704).

The subject set editor is coming real soon, followed by a proper workflow editor.

PR #175, opened 2015-04-01
Subject set editor

brian-c:
This fills out the subject set editing pages of the Lab section.

Retirement classification-count should be reflected next time the back end deploys.

#166 should be merged first.


aweiksnar:
Looks solid!

PR #177, opened 2015-04-01
Add lab pages

aweiksnar:
This just puts #175 into master


brian-c:
I think I screwed this up. My later PR was on top of an earlier one, so really the later one should have merged into the earlier one first, and then the whole mess into master. Sorry!

PR #178, opened 2015-04-01
Fix displayName conflict

aweiksnar:
Clicking the new subject set button on the lab page would sometimes render an error about duplicate display names, this just removes the ones that conflict with the new lab pages, and match the corresponding workflow page to those


brian-c:
Weird, didn't know that'd cause problems. I though `displayName` was optional anyway. Thanks.

PR #188, opened 2015-04-09
Add initial inbox functionality

aweiksnar:
**Working**
- You should be able to navigate to another users profile, send over a message, and then they'll get a filled-in envelope in their navigation until they've read the message.
- Adding messages to the conversation once started.

**Todo**
- Still needs validations and more email-style interaction niceties , but I think this is just enough for a basic inbox to start tinkering with
- Improve empty states

### DIFF
--- a/app/api/talk.coffee
+++ b/app/api/talk.coffee
@@ -12,7 +12,18 @@ talkClient = new JSONAPIClient 'https://talk-staging.zooniverse.org',
   'Content-Type': 'application/json'
   'Accept': 'application/json'
 
-talkClient.headers = apiClient.headers
+authClient.listen 'change', ->
+  authClient.checkCurrent()
+    .then (user) ->
+      if user
+        token = authClient._bearerToken
+        talkClient.headers['Authorization'] = "Bearer #{token}"
+      else
+        delete talkClient.headers['Authorization']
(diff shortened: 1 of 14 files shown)


app/api/talk.coffee
```diff
@@ -12,7 +12,18 @@ talkClient = new JSONAPIClient 'https://talk-staging.zooniverse.org',
   'Content-Type': 'application/json'
   'Accept': 'application/json'
 
-talkClient.headers = apiClient.headers
+authClient.listen 'change', ->
```
aweiksnar: I added this back in because I was having headers issues I can't recall right now, but will have a better auth solution for talk in down the line.

brian-c:
- You're not calling `preventDefault` on any of the form submit events, so the page refreshes. Does that not happen for you?

- Why catch and log errors instead of rendering them in the UI?

Works well once the submits are handled! Nice stuff.

aweiksnar:
Ah sorry--wasn't getting the submit refresh in firefox - I'll do a chrome/safari pass for the next one.

Yeah good call on errors, I'll do a pass over the talk section and clean those up. Mostly been logging to console for dev purposes

brian-c:
Looks good, thanks!

PR #194, opened 2015-04-15
Lock all dependencies

chrissnyder:
Fresh install errors out when attempting to serve the project. Locking dependencies solves this, so some "minor version" bump of a module breaks the build.


brian-c:
What error are you getting? It's working okay here.

I'm not crazy locking everything as a solution, but I wouldn't mind switching to the `~` prefix so we'd only get new patch versions.

chrissnyder:
Welp, I reliably produced it earlier and am reliably not able to produce it now. Pretty sure I nuked `node_modules` before each attempt, but maybe not.

I'm definitely for at least only updating on patch version bumps (looking at react/react-router in particular). I'll update the PR ina sec.

brian-c:
Looks good.

I just ran `npm config set save-prefix='~'`, too.

PR #199, opened 2015-04-17
Workflow editor

brian-c:
This is basically all the new stuff for creating and editing workflows, plus some bug fixes for things that came up during demos this week.

The task types that currently exist (single, multiple, and drawing) are similar enough that they can be edited with the same editor. Future task types may need custom editors.


aweiksnar:
Looks great! 
I found that if you save a task without adding a choice it prevents new ones from being created, but that'll get fixed with https://github.com/zooniverse/Panoptes/issues/743

PR #210, opened 2015-04-21
Don't render "pending" promise if an old value exists

brian-c:
This makes it so that when a previously-rendered PromiseRenderer gets a new promise, it shows its old data until the promise resolves instead of blanking out.

The tradeoff is that sometimes old data is displayed for a while. I think it's worth it, but just close this if I'm wrong.


aweiksnar:
Great! this'll fix an issue I was having rendering user-avatars

PR #221, opened 2015-04-27
Update react to 0.13

aweiksnar:
Dependency upgrade to react 0.13.2

### DIFF
--- a/app/lib/draggable.cjsx
+++ b/app/lib/draggable.cjsx
@@ -7,7 +7,6 @@ module.exports = React.createClass
   _previousEventCoords: null
 
   propTypes:
-    children: React.PropTypes.component.isRequired
     onStart: React.PropTypes.oneOfType [
       React.PropTypes.func
       React.PropTypes.bool
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "blue-tape": "~0.1.7",
     "browserify": "~8.1.0",
     "cjsx-loader": "~1.1.0",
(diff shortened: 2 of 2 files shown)


app/lib/draggable.cjsx
```diff
@@ -7,7 +7,6 @@ module.exports = React.createClass
   _previousEventCoords: null
 
   propTypes:
-    children: React.PropTypes.component.isRequired
```
brian-c: Is something here deprecated now?

brian-c:
@aweiksnar Mind bumping coffee-rectify to 3.0.0 while you're in there?

I'll take care of that getDOMNode-in-render warning next time I update the classifier.

app/lib/draggable.cjsx
aweiksnar: Yeah looks like it's no longer in the valid prop-types https://facebook.github.io/react/docs/reusable-components.html

aweiksnar:
Just bumped it--sorry totally missed that warning

brian-c:
Thanks.

The warning will take some work to fix. I think that's where it's trying to find the best position for the drawing tool details tooltip, and I'm not really happy with how that's working anyway.

cc: @saschaishikawa; we were talking about updating React the other day.

app/lib/draggable.cjsx
aweiksnar: Ah I guess it's just React.PropTypes.element now. I'll get that in the next one

PR #224, opened 2015-04-27
Update board section key for api, Add roles inputs

aweiksnar:
The talk-api now expects `section` values to be formatted as project_id-project_name. This also adds permissions assignment when creating a board

### DIFF
--- a/app/talk/home.cjsx
+++ b/app/talk/home.cjsx
@@ -5,9 +5,13 @@ module?.exports = React.createClass
   displayName: 'TalkHome'
 
   getSection: ->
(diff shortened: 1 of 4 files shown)


brian-c:
Cool, looks like it works. I noticed that the discussions count on the init page sidebar doesn't seem to take hidden boards into account. If we can get that number right for the user, I think we should, but I suspect that'll be a lot of work. Maybe we should just take it off?

aweiksnar:
Ah yeah I'll push something up to this branch taking it off and then build stats out a little better later on

PR #237, opened 2015-04-29
Polygon tool always draws a closed shape

### LINKED ISSUE #180: Polygon tool should finish polygon on double-click
When you draw a polygon and get to the last vertex, double clicking causes the polygon to stop being drawn, but doesn't draw the final side. Instead, you have to click on the original vertex to finish it up. (And once you've double-clicked, you no longer have the option of finishing it up by clicking the final vertex.)

I found this unintutive myself and then I just got alpha testing feedback from others in my lab saying that they expected that a double-click would complete the polygon, since that's how most drawing programs work.

brian-c:
The polygon tool previously allowed for open and closed shapes based on how you finished it. The UI was confusing, and that decision shouldn't be made by the user anyway, it should be made by the project creator. So now the polygon tool just makes closed polygons.

@simensta This will probably screw up the multi-point-line tool you were working on, sorry. I think I find it hard enough to extend tools like this that the hassle of setting them all up just right outweighs the benefit of sharing any drawing code. I'd just make it a standalone thing.

Closes #180.

### DIFF
--- a/app/classifier/drawing-tools/polygon.cjsx
+++ b/app/classifier/drawing-tools/polygon.cjsx
@@ -17,7 +17,7 @@ module.exports = React.createClass
 
     defaultValues: ({x, y}) ->
       points: []
-      closed: null
+      closed: false
 
     initStart: ({x, y}, mark) ->
       mark.points.push {x, y}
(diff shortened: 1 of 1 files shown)


aweiksnar:
Cool! that seems like a smart choice

PR #238, opened 2015-04-29
Add some padding around dialogs, closes #234

### LINKED ISSUE #234: Right and left margins for 'help' text

chrissnyder:


brian-c:
Just saw this Chris, sorry.

For now I'd actually like to keep dialog padding at 0 and let the content determine its own layout. In this case I think we can use the `.content-container` class, which puts a consistent space around stuff like this. That'll keep something like the sign-in form form dialog form (a tabbed-content thing) from getting too much.

PR #241, opened 2015-04-29
Project owner can configure allowing user choice of workflow

### LINKED ISSUE #189: Workflow appearance on landing page
@brian-c I'm working on exactly what the two-project solution looks like. We're going to have a totally custom landing page with two buttons that go directly to 'classify' in the two projects. 

Now, once a user is in classify in, say, Project A, and they click the "Project A" word/tab in the upper left, what happens? If we don't do anything special, they'll go to the 'landing page' for Project A. 

Question 1: Is it (easily) possible to override where this link goes so that instead of Project A's landing page, it goes to our custom landing page?

Assuming the answer to Question 1 is 'no', then we'll have to create content for the 'landing pages' of Projects A and B, with links between them and back to the custom landing page. That's fine. But leads to:

Question 2: Can we have a single button on the project landing page that goes to 'classify' without specifying workflow (and so get a random workflow)? 

Right now there's a button for each workflow on the project landing page. But I don't want users to have the option of choosing different workflows.

brian-c:
Allow project owner to specify a single or per-workflow button-to-classify on a project's home page.

Will test when zooniverse/Panoptes#760 is deployed.

Closes #189.

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -28,6 +28,13 @@ module.exports = React.createClass
 
         <p>
           <label>
+            <input type="checkbox" name="configuration.user_choose_workflow" checked={@props.project.configuration?.user_choose_workflow} onChange={@handleChange} />
+            Volunteers can choose which workflow they work on
+          </label>
+        </p>
+
+        <p>
+          <label>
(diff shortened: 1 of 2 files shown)


brian-c:
Oops, typo, hold on...

PR #256, opened 2015-05-01
Add .content-container class to dialogs

### LINKED ISSUE #234: Right and left margins for 'help' text

rogerhutchings:
As suggested in #238 by @brian-c

@aliburchard

Fix #234

### DIFF
--- a/css/dialog.styl
+++ b/css/dialog.styl
@@ -7,7 +7,7 @@
   top: 0
   width: 100%
   z-index: 1
-  
+
 .dialog
   background: white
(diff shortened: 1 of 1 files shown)


brian-c:
I meant literally wrap the content in a div with the class `content-container` where appropriate. The dialog should not impose any layout on its contents.

aliburchard:
I don't understand that specific tech desires on this one, but @brian-c are you implementing something that will add *some* sort of padding within the pop-up? Because right now the text bleeds into the edges and adding some wrapper (however it's done) would make it look a lot cleaner.

PR #261, opened 2015-05-01
Move coffeeify to devDependencies

aweiksnar:
Fixes preventative npm warning:
npm WARN peerDependencies The peer dependency coffeeify included from panoptes will no
npm WARN peerDependencies longer be automatically installed to fulfill the peerDependency 
npm WARN peerDependencies in npm 3+. Your application will need to depend on it explicitly.

Also updates json-api-client to 0.2.6

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "animated-scrollto": "~1.1.0",
     "counterpart": "~0.16.7",
     "debounce": "~1.0.0",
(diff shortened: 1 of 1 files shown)


brian-c:
The reason that was a peer was for custom projects that want to bundle this as a lib themselves.

I don't think anybody's doing that anymore, so this should be fine.

Thanks.

PR #291, opened 2015-05-06
Simplify classifier wrapper, reset state between classifications

### LINKED ISSUE #289: Project freezes repeatedly
I've received 4 comments from collaborators about my WildebeestWatch project freezing after the drawing task. And, in fact, the below screenshot shows the screen after I've tried to draw a line, but now can't seem to end the line, go back, or do anything else.

screen shot 2015-05-06 at 10 42 15

I've just realized this probably has to do with trying to add drawing tasks when some subjects have multiple images in them...I can see why that's a feature not planned for first release (as you've got to tie pixels to unique image IDs instead of just the subject ID), but I think it's functionality to plan for in the future. And in the meanwhile, perhaps some warning when linking a workflow with drawing tasks to subject sets that have multiple files per subject? Perhaps relevant to the discussion re: manifest structure in #282

brian-c:
The classifier can get into a state between ending one classification and starting another where the old classification's workflow is still loaded but the new classification tries to render.

This PR resets the classifier's workflow and subject when a different classification is loaded. Also removes some of my mixins to make things more explicit.

Closes #289.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -4,7 +4,6 @@ ChangeListener = require '../components/change-listener'
 SubjectViewer = require './subject-viewer'
 ClassificationSummary = require './classification-summary'
 tasks = require './tasks'
-PromiseToSetState = require '../lib/promise-to-set-state'
 
 NOOP = Function.prototype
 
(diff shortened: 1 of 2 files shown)


aweiksnar:
Sweet!

PR #297, opened 2015-05-07
Project lab page checks to make sure the user has permission to edit the project

### LINKED ISSUE #270: when you log out while in the workflow editor (right name?) having it redirect you to Zooniverse panoptes home
When you log out from the workflow editor, it keeps you where you are so you can actually add tasks or options to your workflow but you can't save them, so if you forget if you're logged out or come back later to a tab you forgot to close (in my case) and don't look at the top right you don't realize you're not logged in. 

Any changes you make won't be saved and it does bring up a message saying you must be logged in to access this resource, but only after you hit save .  I think it would be helpful to explicitly redirect to Zooniverse panoptes home so that someone new to panoptes doesn't make that mistake. Or maybe it can do the wordpress thing where it redirects you to your project.

brian-c:
Project lab pages weren't checking to make sure that a user is signed in and they have permission to edit the project. Now it does.

Closes #270.

### DIFF
--- a/app/pages/lab/project.cjsx
+++ b/app/pages/lab/project.cjsx
@@ -194,16 +194,40 @@ module.exports = React.createClass
       projectID: '0'
 
   render: ->
-    getProject = auth.checkCurrent().then =>
-      apiClient.type('projects').get @props.params.projectID
-
     <ChangeListener target={auth} handler={=>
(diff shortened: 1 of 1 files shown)


aweiksnar:
Looks good!

PR #309, opened 2015-05-12
Improve workflows

### LINKED ISSUE #304: help text box is too small and difficult to write in when writing more than two sentences
For the P4 project we need to stick everything into the help because it appears there is no tutorial (is that correct?) and because there currently isn't a separate page we can create to be a spotters guide that is easily accessible from the classification interface.  

 I found that the help box position is really small and it's not very easy to enlarge if you need to upload images or write more than two sentences of text.

### LINKED ISSUE #308: Can't re-order first task
You can re-order most tasks by changing the references in the "next task" option. But this breaks when you want to change which task shows up first -- it seems the only way to change which task shows up first is to completely remove the first task (and then the next task automatically becomes the first task)... Is there any way to fix this so we can more effectively reorder tasks?

brian-c:
Allows choosing a first task (close #308)

Hides drawing task details popup unless details exist

Makes the help input bigger (close #304)

### DIFF
--- a/app/classifier/subject-viewer.cjsx
+++ b/app/classifier/subject-viewer.cjsx
@@ -90,7 +90,7 @@ module.exports = React.createClass
 
         {if @state.selectedMark? and @refs.selectedTool?
           toolDescription = @props.workflow.tasks[@props.annotation.task].tools[@state.selectedMark.tool]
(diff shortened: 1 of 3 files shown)


chrissnyder:
:+1:

Issue #305, opened 2015-05-12
avatar icon not showing in firefox but showing in chrome


mschwamb:
I don't know if this happens when there isn't a temp avatar image but I noticed when I switched to test in chrome the blue and green striped avatar temp icon and that it doesn't show up on firefox.  Screenshots below (checked it out on other projects to make sure it wasn't just mine). Shows up in Safari as well. Just not in Firefox. 

Firefox:

![screen shot 2015-05-12 at 3 08 28 pm](https://cloud.githubusercontent.com/assets/753325/7581714/ecc4b880-f8b8-11e4-9010-8cfd1d381109.png)

Chrome:

![screen shot 2015-05-12 at 3 08 32 pm](https://cloud.githubusercontent.com/assets/753325/7581724/ff45f78a-f8b8-11e4-926e-78fb85dd9ca1.png)

mkosmala:
It's worse. Whatever happened this week broke the links to my images. Previously we had a custom avatar that showed up (on Firefox and Chrome), as well as a background image. I'm not sure what the difference is between Firefox and Chrome where a non-existent avatar shows up in one but not the other.

mschwamb:
Fixed with the previous pull request merged. So I'm closing the ticket since I initiated it.

PR #311, opened 2015-05-12
Add user avatars

brian-c:
This adds a drag-and-drop avatar field on the user settings page. This component can be used for project avatar and background image (coming soon).

Neat stuff: it optionally auto-crops images if you give it an aspect ratio (`1` is square) and optionally scales them down until they're under a size limit (in bytes).

Also preps for other user settings (email preferences to come real soon).

Probably needs some CSS work, but so does literally everything (again, soon).

### DIFF
--- a/app/components/image-selector.cjsx
+++ b/app/components/image-selector.cjsx
@@ -1,0 +1,127 @@
+React = require 'react'
+LoadingIndicator = require './loading-indicator'
+
+BASE_64_PNG_HEAD = 'data:image/png;base64,'
+BASE_64_JPEG_HEAD = 'data:image/jpeg;base64,'
+BASE_64_EXPANSION = 3 / 4
+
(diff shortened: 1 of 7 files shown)


aweiksnar:
Super cool, Just updated my own and it worked without error. Looks like this will take care of #305 as well

brian-c:
Nah, #305 is an issue with Firefox not rendering _anything_ for an image if it doesn't have a `src`. Chrome at least fills the space, and the default-stripes-icon thing is just a background on that space. There might be an easy fix for this, but it's not critical at the moment.

I'll have a PR for project images up shortly. Pretty much the same as this one.

aweiksnar:
Ahh gotcha, Yeah maybe we can supply a default zooniverse logo image for users without a custom avatar

PR #314, opened 2015-05-13
Add favoriting, and start working towards collections features

aweiksnar:
- Adds a favorite button 
  - It takes in a single 'subject' prop and handles itself from there
  - The code for this feels overly verbose / fragile. Any objections if I try bringing in flux or some other layer for components like this?
- Lets users (moderators) edit the permissions of their board via a group of radio buttons 
- Standardizes a bunch of errors to just throw the error message
- A very raw collections view (/collections or collections/:id), just a paginated list of subjects right now
  - You should see a favorites collection pop up here with any added subjects
  - the form here was for dev, but I'd like to build it out it for general collections use

### DIFF
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -1,0 +1,89 @@
+React = require 'react'
+apiClient = require '../api/client'
+auth = require '../api/auth'
+getFavoritesName = require './get-favorites-name'
+alert = require '../lib/alert'
+
+module?.exports = React.createClass
+  displayName: 'CollectionFavoritesButton'
+
+  propTypes:
+    subject: React.PropTypes.object # a subject response from panoptes
+
+  getInitialState: ->
+    favorited: false
+
+  componentWillMount: ->
+    # see if the subject is in the project's favorites collection
+    # to see if it's favoriters
+    @props.subject.get('project')
+      .then (project) =>
+        project_id = project.id.toString()
+        apiClient.type('projects').get(project_id)
+          .then (project) =>
+            display_name = getFavoritesName(project)
+
+            apiClient.type('collections').get({project_id, display_name}).index(0)
+              .then (favorites) =>
+                if !!favorites
+                  apiClient.type('subjects').get(collection_id: favorites.id, id: @props.subject.id).index(0)
+                    .then (subject) =>
+                      @setState favorited: !!subject
+
+  addSubjectTo: (collection) ->
+    collection.addLink('subjects', [@props.subject.id.toString()])
+      .then (collection) => @setState favorited: true
+
+  removeSubjectFrom: (collection) ->
+    collection.removeLink('subjects', [@props.subject.id.toString()])
+      .then (collection) => @setState favorited: false
+
+  createFavorites: (user, project) ->
+    project_id = project.id.toString()
+    apiClient.type('projects').get(project_id)
+      .then (project) =>
+        display_name = getFavoritesName(project)
+        project = @props.subject.links.project
+
+        owner = {id: user.id.toString(), type: "users"} #--> must supply type
+        links = {project, owner}
+        collection = {display_name, links}
+
+        apiClient.type('collections').create(collection).save()
+          .then (favorites) =>
+            @addSubjectTo(favorites)
+
+  toggleFavorite: ->
+    project_id = @props.subject.links.project.toString()
+
+    @props.subject.get('project')
+      .then (project) =>
+        project_id = project.id.toString()
+        auth.checkCurrent().then (user) =>
+          if user?
+            apiClient.type('projects').get(project_id)
+              .then (project) =>
+                display_name = getFavoritesName(project)
+                # check for a favorites collection in project first
+                apiClient.type('collections').get({project_id, display_name}).index(0)
+                  .then (favorites) =>
+                    # try to request subject from favorites collection, otherwise create it
+                    if !!favorites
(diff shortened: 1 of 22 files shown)


app/collections/favorites-button.cjsx
```diff
@@ -0,0 +1,89 @@
...
+  displayName: 'CollectionFavoritesButton'
+
+  propTypes:
+    subject: React.PropTypes.object # a subject response from panoptes
+
+  getInitialState: ->
+    favorited: false
+
+  componentWillMount: ->
+    project_id = @props.subject.links.project.toString()
```
brian-c: I wanna say almost never access the `links` field directly, if you can help it. You can request a subject's project with `subject.get('project').then (project) ->`.

app/collections/get-favorites-name.cjsx
```diff
@@ -0,0 +1,5 @@
+# Get the name of the favorites collection for a project
```
brian-c: If favorites collections are per-project, couldn't the name always just be "Favorites"? Project names can change.

aweiksnar: Yeah I initially had just 'favorites'. The api scoping on the uniqueness validation is not defined to the project, so there can only be one 'favorites'. This is a workaround

app/collections/favorites-button.cjsx
```diff
@@ -0,0 +1,89 @@
...
+    apiClient.type('projects').get(project_id)
+      .then (project) =>
+        display_name = getFavoritesName(project)
+
+        apiClient.type('collections').get({project_id, display_name}).index(0)
+          .then (favorites) =>
+            if !!favorites
+              apiClient.type('subjects').get(collection_id: favorites.id, id: @props.subject.id).index(0)
+                .then (subject) => @setState favorited: !!subject
+                .catch (e) -> throw new Error(e)
```
brian-c: `.catch (e) -> throw new Error(e)` should be redundant, no? If these have some effect, there might be a problem in the api client's error handler.

app/collections/favorites-button.cjsx
```diff
@@ -0,0 +1,89 @@
...
+                      else
+                        @addSubjectTo(favorites)
+                    .catch (e) -> throw new Error(e)
+                else
+                  @createFavorites(user)
+      else
+        alert(<div className='content-container'>You must be logged in to favorite</div>)
+
+  render: ->
+    <button
```
brian-c: Good habit to always define `type="button"` with button elements, otherwise they crazily default to type="submit"` and things get confusing if you ever end up using something in a form.

app/collections/home.cjsx
```diff
@@ -0,0 +1,45 @@
...
+
+  setCollections: ->
+    auth.checkCurrent().then (user) =>
+      user.get('collections')
+        .then (collections) =>
+          @setState {collections}
+        .catch (e) -> throw new Error(e)
+
+  collectionLink: (d, i) ->
+    <div key={i}>
```
brian-c: Keying off an index is bad news, since the index doesn't relate to the item itself, just its position in the list.

app/collections/home.cjsx
```diff
@@ -0,0 +1,45 @@
...
+PromiseRenderer = require '../components/promise-renderer'
+
+module?.exports = React.createClass
+  displayName: 'CollectionsHome'
+
+  getInitialState: ->
+    collections: []
+
+  componentWillMount: ->
+    @setCollections()
```
brian-c: Should this be called whenever `auth` changes?

app/collections/favorites-button.cjsx
aweiksnar: I can see that as a good rule. I did that to cut down on nesting. I'll push up a change with an explicit request in a sec

app/collections/show.cjsx
```diff
@@ -0,0 +1,54 @@
...
+          subjects: @subjectsRequest(page)
+          }
+
+  goToPage: (n) ->
+    @transitionTo(@props.path, @props.params, {page: n})
+    @setData(n)
+
+  subject: (d, i) ->
+    # 'http://' may need to prefix some older subjects...
+    <img key={i} src={getSubjectLocation(d).src} />
```
brian-c: Could this be a link to the subject discussion page?

aweiksnar: Yeah it will be--once that page is built I'll point any subject reference in collections/talk to it

app/collections/home.cjsx
aweiksnar: I'll change these to key on resource id. 

As an aside: Why call keying on indexes bad news? React.js does it in their examples, internally when no key is supplied, and recommends it over random numbers

brian-c: React uses the `key` of components in an array to associate them with their data. So if you have a list of three objects `[a, b, c]` and pass them in props to three components keyed by index, and then shift off `a` from the array, the third component (which represented `c`) will be removed (because index `2` no longer exists) and you'll be left with two components. The first will maintain the _state_ it had when it represented `a`, but it'll now have `b` in its _props_. The second will have its state from `b` but with `c` in its props.

Where's React use it in their examples? That seems weird.

aweiksnar: Ah I guess I've never been bitten by that case before. I'll go through and update other places in the code where I've used indexes

app/collections/favorites-button.cjsx
aweiksnar: These catches should be removed now. I think it's just in my muscle memory to add those in

brian-c:
Lookin' good, thanks!

Issue #205, opened 2015-04-20
Home page - near term


chrissnyder:
I wish there was a way to tag everybody for this issue...

Consider this the place to have discussions on what will show up on zooniverse.org when we launch Panoptes. While there has been many ideas bantered about, please keep this scoped to realistic goals that we can achieve in this milestone.

@heathv Please reference any design work you do on the homepage here.

We chatted about this two weeks ago in Chicago, and while I don't want to put words in anyones mouth, here is what I took out of that discussion:

A logged in user simply goes to their profile page (comments on, please see #59). 

A non-logged in user sees a couple things, in order down the page

- Intro to the Zooniverse - large banner image with some brief copy attempting to explain who we are and what we are all about. Call to action should be about creating an account and getting started classifying on a project. 
- Zooniverse Live - In a perfect world, I'd love to simply port ZooLive into this space. Realistically, some way of communicating that stuff is actually happening in the Zooniverse. Feed from Daily Zoo, subject/user/classification counts, whatever.
- Projects - A brief listing, probably a curated one, of various projects built on Panoptes.
- Project Builder - Reference the project builder and message that if you have an idea for a project, you can use our platform to create it.
- Footer - Various links that we have to put somewhere.

Worth mentioning that nothing has been built yet and it's all up in the air.

chrislintott:
Additions from me :

1. A view for beta projects as well as the full on project list
1b. The inclusion of the other live Zooniverse projects (not just Panoptes) 
2. Login. 
3. Whatever content from the CSA site is worth saving.

chrissnyder:
1. The way I currently imagine it, these would live on the project list page. I've opened #206 to discuss concerns there.
2. Already handled via the popup and is available sitewide.
3. I've asked @DarrenMcRoy to do an audit on what content we have available. Will be opening another issue to discuss what other content we'd like to have, outside of what already exists (if any).

chrislintott:
I will take on the task of looking at the CSA content.

Chris

vrooje:
In case it's useful, I've moved the notes I took during the ZooTM unconference session on Zoo Home to [this separate Google Doc](https://docs.google.com/document/d/1OmDSkMiRqrS1vN_u7C3y_4hgDqKgoJD8Zhg6qwNVlRg/edit?usp=sharing).

The doc is mostly looking farther down the line but I've put in bold the things we talked about as potentially more immediate. They were:

Zooniverse Live - including live running classification count (total, from all projects)
[...]

PR #315, opened 2015-05-13
Homepage

srallen:
- Updated Home page styles and content based on @heathv's design in #205
  - Featured projects listing with scroll listener and effects 
- Reworked header
  - Updated styles including moving the main-nav-items to be inline
  - Added on hover context menu for logged in user from discussion with @heathv 
  - Added scroll listener and css transition effects.
- Added basic style guide designs from @heathv for typography and buttons


brian-c:
A couple quick notes:

- Double avatars in the header. Merge error?

- Make sure the sticky header resets itself when switching pages. Does the header need to be sticky anyway? What's that for?

- The header appears to have a fixed height now, which means when it wraps lines on small screens it can't push down the content that follows.

- Let's avoid fixing sizes of things in pixel values (header height, "Get started" button width). I really want us to be in the habit of letting the content and the viewport determine how things lay out, especially since we'll have very little control over most of the content. Might be a specific thing Heath wants; defer to him.

- The featured projects are sticky too, sometimes. Class name collision?

- Lots of font styles across the site are different. Not sure if that was intentional. Might want to namespace typography changes if they only apply to the home page.

srallen:
I've fixed the avatar thing. I didn't catch it because I don't have custom avatar. A default one should be put in place as noted in the pull request for #311. Without any avatar, the context menu isn't aligned correctly, fyi.

Also, I fixed the header when switching pages and removed the max-height. The sticky header on the home page is a design idea from @heathv. I could go either way on it.

The featured projects is supposed to be sticky initially to give site visitors an indication that there's more content to scroll down to. I changed the class names to make this more clear.

The changed typography and buttons are from [the style guide](http://adobe.ly/1E69eLu) @heathv is working on. I could scope the font-sizes, but maybe we should talk about it altogether?

heathv:
There are some weird bits on there right now (i.e., the project cards being half-sticky) due to content not being fully flushed out. We have lots of copy via Darren, but I'm slowly working on getting artwork created to go along with it. Typography conflicts are my fault, I switched to using 10px as a base so font-sizes could be a simple multiple of 10 (i.e., 15px = 1.5em) I have no idea what conflicts this creates in the front-end so it can be switched back if need be.

brian-c:
Cool, thanks @srallen.

Re: font size, if you want to define something in px, use px. I wrote everything using em-sized fonts so they'll grow or shrink appropriately in context but if you want to define font sizes in explicit pixels, no worries. Just keep in mind the consequences of changing the base.

heathv:
Cool, will do.

srallen:
@brian-c, I've changed the css to scope the font-sizes to just the homepage for now. Also added in some more designs and placeholder content that Heath had ready this morning.

brian-c:
Sweet, thanks.

PR #317, opened 2015-05-13
Wrap up tasks' help dialog content

### LINKED ISSUE #234: Right and left margins for 'help' text

brian-c:
Here's all I meant by wrapping up the help dialog content. Sorry for being so picky @chrissnyder and @rogerhutchings, just trying to keep things as consistent when it's easy.

Closes #234, closes #238, closes #256.

### DIFF
--- a/app/classifier/tasks/generic.cjsx
+++ b/app/classifier/tasks/generic.cjsx
@@ -36,4 +36,6 @@ module.exports = React.createClass
     </div>
 
   showHelp: ->
-    alert <Markdown className="classification-task-help">{@props.help}</Markdown>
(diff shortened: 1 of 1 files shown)


chrissnyder:
Merging cause it's better than it is now.

That said, I dislike using the same class for both generic static page content and dialogs (amongst other locations that probably exist). I can see the styling for both of them diverging pretty quickly.

brian-c:
The idea is that the content is what's being styled, not its container. A dialog should not affect the rendering of its contents.

Applying padding to the dialog itself would have put double padding on the sign-in dialog, since it uses a `content-container` in its own layout.

PR #323, opened 2015-05-14
Enable image selection for project avatar and background image

brian-c:
Allows setting project avatars and background images

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -1,5 +1,12 @@
 React = require 'react'
 BoundResourceMixin = require '../../lib/bound-resource-mixin'
+PromiseRenderer = require '../../components/promise-renderer'
+ImageSelector = require '../../components/image-selector'
+apiClient = require '../../api/client'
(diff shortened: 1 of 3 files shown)


chrissnyder:
UI seems to act weird when first adding these to a fresh project

1. Initial state, brand new project - http://i.imgur.com/jnEWk8J.png
2. State after adding avatar - http://i.imgur.com/mvxDqU6.png
3. State after adding background - http://i.imgur.com/HZ80imz.png

Step 3 seems pretty variable, in that sometimes both images will disappear, or both will work. Some race condition maybe?

Refreshing the page solves all the above and makes it work just fine.

brian-c:
Ooo, yeah, something funny going on there. I'll look into it.

brian-c:
Race condition indeed. The JSON-API lib doesn't cache failed requests, so if the project didn't originally have an avatar or background, it would request it again every time the component rendered, so it'd end up trying to get the image at the same time it was trying to create it.

Should be fixed now.

chrissnyder:
Tried again, worked like a charm. Merging...

PR #324, opened 2015-05-14
Project links

brian-c:
Allows adding external links to projects' navigation. Editor is barebones for now, just crammed onto the project details page.

Links without labels get a sweet icon.


chrissnyder:
Today I learned that example.com is actually reserved domain by the IANA.

Works as advertised, merging. Though we will want to add a delete button sooner rather than later.

PR #336, opened 2015-05-15
Move api and auth code to its own module.

chrissnyder:
I'm not 100% sure of the consequences of this, but I wanted to put it out there.

This PR pulls out all the api and auth code into it's own module, the code of which is at https://github.com/zooniverse/panoptes-javascript-client, and available on npm and bower under "panoptes-client". It's a straight es6 rewrite of the existing code (with test coverage that passes!).

I'm not really happy with how the client object itself is organized, but I kept it the same to avoid having to change a bunch of references to api/auth everywhere. Maybe later I'll tackle this.

The only real difference is that PanoptesClient has some sensible defaults set that you overwrite by passing params. Makes initializing a new instance of the client pretty easy.


brian-c:
I believe @rogerhutchings has been working on this exact same thing, you guys should really sync up. I'm totally for this, just not a week before launch, and I really want the new base libs to not have to be compiled.

brian-c:
Closing for now, we can revisit this when we have some free time for maintenance.

Issue #335, opened 2015-05-15
Done button is unreadable (button box is same color as the text)


mschwamb:
![screen shot 2015-05-15 at 9 11 16 am](https://cloud.githubusercontent.com/assets/753325/7645160/6fcf560a-fae3-11e4-9487-c0e841d0d0d3.png)

mschwamb:
Also it's hard to see the outline of the other buttons with the background color.

brian-c:
Closed in #339.

PR #339, opened 2015-05-15
Scope button and color styles to home page

### LINKED ISSUE #334: new default background and html header text same color nearly the same
with the new default background color is a almost the same color as the color of text written when you make it a header is markdown. 

screen shot 2015-05-15 at 5 41 57 am

srallen:
Scopes button and color styles that I added with the home page updates to home page only. Fixed #334 and #335


brian-c:
Looks good, thanks.

PR #345, opened 2015-05-15
Add sign-in prompt to favorite button

### LINKED ISSUE #344: Add a links to pop up login and register window to the 'you must be logged in to register'
More based my own experience, but I've always asked for this on builds I'm involved in when windows pop-up ask for you to login to get those features since I think most people will miss the login/register in the top right corner, and  those projects tend to have higher numbers of registered/logged-in users, but more anecdotal evidence. So I think it would be useful if you could click from the box below and get the register and login windows when you try and favorite non-logged in. 

screen shot 2015-05-16 at 2 14 32 am

aweiksnar:
Closes #344

### DIFF
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -15,7 +15,9 @@ module?.exports = React.createClass
     favorited: false
 
   promptToSignIn: ->
-    alert <SignInPrompt />
+    alert <SignInPrompt>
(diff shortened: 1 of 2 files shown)


brian-c:
Made a little edit. Work for you?

aweiksnar:
Ah yeah that makes sense

brian-c:
Oops, I totally fouled this up somehow. Just a minute.

PR #347, opened 2015-05-15
Links shouldn't affect workflow state

brian-c:
This should fix zooniverse/Panoptes#839.

The API lib was not triggering "change" on resource refreshes, which made the save button active when no changes were made.

Also makes sure the classify page doesn't store an empty list of workflows for a project.


chrissnyder:
Works like a charm. Able to link/unlink subject sets just fine. Though it's pretty awkward to have the subject set config be auto-saved and everything else not. An optimization for later perhaps.

Nothing to complain about in the code, so merging!

Issue #337, opened 2015-05-15
Uploading a background image completely messes up the format of the project description page


mschwamb:
The background image displayed in the workflow editor at full size of the image (as does the logo) and then it pushes the text boxes for the project description out of the way and makes them smaller making it really hard to make edits. 

![screen shot 2015-05-15 at 10 44 35 am](https://cloud.githubusercontent.com/assets/753325/7645984/ae2bdd86-faef-11e4-96f4-f508c6b8440e.png)

mschwamb:
Then if you move off it goes away and you're back to this but the background image is loaded (but I didn't actually know it had succeeded in saving it and it wasn't some glitch

![screen shot 2015-05-15 at 10 50 01 am](https://cloud.githubusercontent.com/assets/753325/7646055/82014740-faf0-11e4-9b3b-62280c0ed5e6.png)

chrissnyder:
The disappearing/reappearing thing I thought we fixed yesterday, but I guess not.

What browser/OS combo are you using? Here's how the project page looks for me: http://i.imgur.com/4dcKhqU.png. Definitely shouldn't look like that.

edpaget:
@chrissnyder did you deploy those changes?

chrissnyder:
It got merged to master, so I assumed so. Maybe not then?

brian-c:
I think the flex layout behaves a bit different on Firefox than Chrome. Might just need to set a max-width somewhere on on of the column classes.

brian-c:
Also, auto-deploy seems to be down (#325). I'll push manually in a minute,

mschwamb:
I was using Firefox 37.0.2 (sorry for not mentioning that earlier)

mschwamb:
Looks like this in Chrome and Safari
![screen shot 2015-05-16 at 12 59 17 am](https://cloud.githubusercontent.com/assets/753325/7657751/06853f1e-fb67-11e4-8b7c-4cfcdb3882e8.png)
 but  like this in firefox
![screen shot 2015-05-16 at 1 01 26 am](https://cloud.githubusercontent.com/assets/753325/7657771/318f41be-fb67-11e4-8924-c296b91879ec.png)

brian-c:
Fixed in that PR.

PR #349, opened 2015-05-15
Quick fix for Firefox's handling of images in flexed containers

brian-c:
Not a perfect solution, but it's fine for where we're using this component right now.


brian-c:
Closes #337.

Issue #217, opened 2015-04-23
Hiding custom projects


perry:
We'll need to be able to hide public custom projects (eg. Old Weather, Tate) from appearing here, as they are likely to contain custom workflows that will not be supported in PFE.

brian-c:
@chrislintott vetted/promoted custom projects should still appear on zooniverse.org, right?

I'm thinking we can handle this with two project-level options:

- `"redirect": "https://www.ancientlives.org/"` would point to the external URL and redirect when a user tries to load that project

- `custom_workflow` would disable the workflow editor UI.

@edpaget Is this a reasonable use of the project config field?

edpaget:
Yeah I think that's fine.

edpaget:
I'll try to have a PR in for this by the end of today.

PR #350, opened 2015-05-15
Follow project redirects

brian-c:
Landing on a project page with `configuration: redirect: 'http://example.com'` will now take you there when you land on its project page.

Also added avatars to the project card-style links.

For half of #217.

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -116,6 +116,8 @@ module.exports = React.createClass
       </div>
 
       <div className="column">
+        {if @props.project.configuration?.redirect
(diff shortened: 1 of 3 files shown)


chrissnyder:
I assume it's intended that only we can add a redirect manually, and not exposing it anywhere?

brian-c:
Yeah, it would probably be good to have an admin-only field for this. It's not exposed in the interface, but someone determined could set it and redirect people someplace bad.

@edpaget think a public-read-only "redirect" field on the project will work?

Also, disregard this PR anyway, because the redirect should happen on the project landing page. The way this is set up will make it impossible to access Talk for custom projects.

PR #363, opened 2015-05-19
Header footer v2

### LINKED ISSUE #360: Zooniverse Discuss and Project Discuss links are confusing
On my GZ bars project:

screen shot 2015-05-19 at 17 09 12

Should we change the name of one of them?

srallen:
- Updated styles for header and footer for consistency on other pages besides home. 
- Renamed Discuss in header and footer to Talk to fix #360

### DIFF
--- a/css/main-header.styl
+++ b/css/main-header.styl
@@ -4,6 +4,9 @@
     position: absolute
     width: 100%
 
+    .main-logo
+      font-size: 2.7em
+
   a
     color: inherit
     text-decoration: none
@@ -19,7 +22,7 @@
     position: relative
 
     .account-info
-      padding-bottom: 1.5em
+      padding-bottom: 15px
 
       .avatar
         height: calc(1em + 1.5vw)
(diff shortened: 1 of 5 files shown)


css/main-header.styl
```diff
@@ -117,20 +116,18 @@
     margin: 0
 
 .main-nav-item
+  color: rgba(255,255,255,0.5) !important //ugh
```
brian-c: Ehhh, let's fix the original problem (`.main-header a`, right?) instead of this.

srallen:
Cool. I was hesitant with changing it and inadvertently changing something elsewhere like I did with the home page styles initially. I've updated and explicitly declared the anchor tag color styles.

brian-c:
Nice, thanks. The header and footer should be pretty standalone. The rest of the site is kinda a mess. Might be time to bring in something like BEM.

One more tiny thing I just remembered, sorry: The account menu appears on when hovering over the space it takes up when visible. I _think_ you can quick-fix this by applying `pointer-events: none` to `.account-menu` and switching to `pointer-events: auto` when its parent is hovered.

srallen:
Adding pointer-events fixed it.

Also agreed, BEM or something like it would be very helpful. The mess is partly my fault since I'm trying to get styles in quickly.

brian-c:
Looks good!

PR #369, opened 2015-05-20
updated logotype on home and footer

### LINKED ISSUE #365: Zooniverse logo looks weird in some browsers
Might be solved by making the whole word an image, rather than having the logo in the middle of text?

zoo-home1

srallen:
- Updated to use inline svg for entire logotype on home and in footer so it can be styled as needed.
- Added aria role and title to svg for accessibility.
- Fixed #365


srallen:
It makes sense to me to follow whatever the recommend best practice is and generating a unique id wouldn't be difficult. However, I don't have much experience with aria. I found [a discussion about aria-label vs aria-labelledby](http://stackoverflow.com/questions/19616893/difference-between-aria-label-and-aria-labelledby). It seems either could fit this use case, though, aria-labelled is generally preferred according to this discussion.

srallen:
@eatyourgreens and @brian-c, I've added a generated ID to the title and aria-labelledby attribute. This should be ready to go.

srallen:
@brian-c, ok I switched it to string interpolation.

brian-c:
Looks good to me. Reassigning to Jim, please merge if the ARIA stuff is right.

PR #379, opened 2015-05-21
Projects page styles

srallen:
- Updated styles for projects page
  - new styles for main-header on projects page
- Updated and new styles project card component


app/pages/projects.cjsx
```diff
@@ -9,7 +9,9 @@ ProjectCard = require '../partials/project-card'
 
 counterpart.registerTranslations 'en',
   projectsPage:
-    title: 'Projects'
+    title: 'All Projects'
+    showing: 'Showing'
+    found: 'found'
```
brian-c: For the future: the translator can interpolate strings like `showing: 'Showing %(count)s found'`. Probably easier to translate later.

app/partials/project-card.cjsx
```diff
@@ -23,10 +30,10 @@ module.exports = React.createClass
       'data-no-avatar': true unless @props.project.avatar
 
     <Link {...linkProps} className="project-card">
-      <svg className="project-card-space-maker" viewBox="0 0 2 1" width="100%"></svg>
+      <svg className="project-card-space-maker" viewBox="0 0 2 150px" width="100%"></svg>
       <div className="details">
         <div className="name">{@props.project.display_name}</div>
         <div className="owner">{owner?.display_name ? 'LOADING'}</div>
-        <div className="description">{@props.project.description}</div>
+        <button type="button" className="ghost-button project-card-button"><Translate content="projectsCard.button" /></button>
```
brian-c: This might not make much sense as a button since it's already inside an a tag. Tab-focus gets sorta weird.

app/pages/projects.cjsx
srallen: Cool. Didn't know about it. Thanks.

app/partials/project-card.cjsx
srallen: I discussed this with @heathv and he wants there to be some kind of indication that the project card is clickable and would prefer the card is the link, but needs to think about how to design it. For now, I've made the button's tabIndex -1 so they do not get tab-focus.

brian-c: Yeah I think making it look like a button is fine. Nesting focusables was my concern. tabindex="-1" should be fine. I'll merge this in a few minutes if that's cool with you.

srallen: Sounds good to me.

PR #388, opened 2015-05-22
Add change-password form

### LINKED ISSUE #333: Allow users to change their password
Is there an endpoint? Putting `{users: {password: 'my new password'}}` returns a 422.

brian-c:
Allows users to change their passwords from the settings page

Closes #333.

### DIFF
--- a/app/pages/settings.cjsx
+++ b/app/pages/settings.cjsx
@@ -9,6 +9,7 @@ apiClient = require '../api/client'
 putFile = require '../lib/put-file'
 
 MAX_AVATAR_SIZE = 65536
+MIN_PASSWORD_LENGTH = 8
 
 ChangePasswordForm = React.createClass
   displayName: 'ChangePasswordForm'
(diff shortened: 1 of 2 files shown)


app/pages/settings.cjsx
```diff
@@ -9,6 +9,90 @@ apiClient = require '../api/client'
...
+          </tr>
+        </tbody>
+      </table>
+
+      <p>
+        <button type="submit" className="standard-button" disabled={not @state.old or not @state.new or @tooShort() or @doesntMatch() or @state.inProgress}>Change</button>{' '}
+
+        {if @state.inProgress
+          <i className="fa fa-spinner fa-spin form-help"></i>
+        else if @state.success
```
chrissnyder: Nitpicky, but `@state.success` is never used, due to having to signOut/signIn to get a new session.

chrissnyder:
Other than the above comment, looks good to me.

app/pages/settings.cjsx
brian-c: I'd actually like to leave this. The real issue is that the auth modules needs to sign out and back in. There might be a way around that, but I don't think it's worth investing too much time in right now.

chrissnyder: Roger that.

Issue #322, opened 2015-05-14
Forgot-password link goes to old password reset page


vrooje:
It should go to the new password reset page.

brian-c:
@edpaget We'll need an endpoint for this one, too.

camallen:
@brian-c the password reset end point exists at POST `/users/password`, note it's not an API route and will respond to the `application/json` content type. You'll need to send the CSRF token and the users email address e.g. `{ users: { email: "users_email@email.com"  } }`. 

That request will send the user an email with a reset link in it, e.g `https://zooniverse.org/users/password/edit?reset_password_token=adsjhfkjsahdf` which the frontend will have to display a form to 
capture the new password and confirmation. Then PUT this back to `/users/password` with the following params, `{ user: { password: "new_password", password_confirmation: "new_password", reset_password_token: "adsjhfkjsahdf" } }` as well as the CSRF token.

We could alternatively just skin the HTML forms for the Panoptes ala https://login.zooniverse.org/ and direct all traffic there like it works now. I've got an issue setup in the backend, https://github.com/zooniverse/Panoptes/issues/768

edpaget:
Actually @camallen has already explained this. After zooniverse/Panoptes#861 is merged the email will cause rails to redirect to `panoptes-front-end-url/#/reset-password?reset_password_token=asdfasdf`

PR #419, opened 2015-05-26
Reset password page

brian-c:
@edpaget this should be okay to merge once the URLs are fixed in the email.


brian-c:
Now closes #322

PR #422, opened 2015-05-26
All metadata values are strings

### LINKED ISSUE #421: data export .csv issues
The generated .csv output classification file has some problems. 
In my manifest.csv I included a unique object ID number as well as the jpeg image name. The output.csv file ("subject_data" header) has all of my object ID numbers rounded to the nearest 100. Additionally, there were "&" characters in the jpeg file name and these were written out as "\u0026" in the output.csv file.

brian-c:
Ensures things appear as they were entered.

Closes #421.

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -214,7 +214,7 @@ EditSubjectSetPage = React.createClass
     reader.onload = (e) =>
       # TODO: Look into PapaParse features.
       # Maybe wan we parse the file object directly in a worker.
-      {data, errors} = Papa.parse e.target.result.trim(), header: true, dynamicTyping: true
+      {data, errors} = Papa.parse e.target.result.trim(), header: true
 
       metadatas = for rawData in data
         cleanData = {}


melaniebeck:
Was this implemented already? Because the issue remains with a .csv file that I downloaded today. My object ID values are still rounded to the nearest 100, etc.

edpaget:
Did you upload the subjects before or after the change? This would only affect any new subject you created after we did this. 

@brian-c is there any way to edit subject metadata? If not @melaniebeck you're welcome to send me your manifest and I'll correct the values for you.

brian-c:
There's no way to edit subject metadata. I'd re-create the subject set.

melaniebeck:
I uploaded the subject set this morning. It was the same subject set as before, however. I already wrote a script to correct the output though. Thanks, Ed.

edpaget:
Which site did you upload to the demo.zooniverse.org/panoptes-front-end or preview.zooniverse.org/panoptes?

melaniebeck:
preview.zooniverse.org/panoptes

brian-c:
Ah crap. I just deployed preview now. Sorry about that.

Should we close up demo.zooniverse and switch the auto deploy to preview?

edpaget:
We could probably do both? They're pointing at different backends still.

PR #427, opened 2015-05-27
About pages

srallen:
- About pages component
  - Added routes and updated main header nav
  - Sub-nav for about pages section
  - Added styles and content


brian-c:
Looks good. It'll need to be made a little more abstract if we ever nest things more than a level under about.

PR #430, opened 2015-05-28
Log if the deployed back end is up to date

brian-c:
Here's a neat little function that'll log how far behind master the currently deployed back end is.

It uses the new `fetch` function, but no worries if it doesn't run (i.e. IE).


chrissnyder:
Now that is clever. :+1:

PR #433, opened 2015-05-28
Replace classify links for redirected projects

brian-c:
Projects with a `redirect` field should link to their external site instead of their classify page.

### DIFF
--- a/app/pages/project/home.cjsx
+++ b/app/pages/project/home.cjsx
@@ -35,7 +35,12 @@ module.exports = React.createClass
       <div className="call-to-action-container content-container">
         <div className="description">{@props.project.description}</div>
 
-        {if @props.project.configuration?.user_chooses_workflow
+        {if @props.project.redirect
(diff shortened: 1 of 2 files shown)


chrissnyder:
Is there a project with a redirect active? (or have ideas on an easy way to test this?)

brian-c:
I don't think so. I just made a new project and ran `zooAPI._typesCache.projects._resourcesCache[231].redirect = 'https://www.example.com/'` in the console.

edpaget:
Staging has the old projects imported from Zooniverse Home. I think they may just be marked private.

edpaget:
Here's one: http://demo.zooniverse.org/panoptes-front-end/#/projects/zooniverse-admin/planet-hunters

chrissnyder:
Couldn't find the attribute on the zoo home projects, but Brian's debug code worked. Assuming actual project resources come back similarly, then :+1:

PR #434, opened 2015-05-29
Add viewing page for subjects, other talk updates

aweiksnar:
- Object page
  - View the subject
  - Favorite Button
  - Display comments mentioning that subject
  - Form to start a new discussion shown if there are boards for that project and you are logged in
    - Pre-filled the discussion comment with that image as the focus
- Some renaming
- Form fixes
- Setup subject mentioning selector to handle subject_id querys
- Default avatar

### DIFF
--- a/app/subjects/index.cjsx
+++ b/app/subjects/index.cjsx
@@ -6,6 +6,7 @@ getSubjectLocation = require '../lib/get-subject-location'
 FavoritesButton = require '../collections/favorites-button'
 ChangeListener = require '../components/change-listener'
 PromiseRenderer = require '../components/promise-renderer'
+SubjectViewer = require '../components/subject-viewer'
 NewDiscussionForm = require '../talk/discussion-new-form'
 {Navigation} = require 'react-router'
 
@@ -18,6 +19,10 @@ module?.exports = React.createClass
 
   componentWillMount: ->
     @setSubject()
+
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.params?.id isnt @props.params?.id
+      @setSubject()
 
   setSubject: ->
     subjectId = @props.params?.id.toString()
@@ -46,9 +51,7 @@ module?.exports = React.createClass
         <section>
           <h1>Subject {subject.id}</h1>
 
-          <img src={getSubjectLocation(subject).src} />
-
-          <div><FavoritesButton subject={subject} /></div>
+          <SubjectViewer subject={subject} />
 
           <PromiseRenderer promise={talkClient.type('comments').get({focus_id: subject.id})}>{(comments) =>
             if comments.length
--- a/app/subjects/index.cjsx
+++ b/app/subjects/index.cjsx
@@ -1,0 +1,81 @@
+React = require 'react'
+apiClient = require '../api/client'
+authClient = require '../api/auth'
+talkClient = require '../api/talk'
+getSubjectLocation = require '../lib/get-subject-location'
+FavoritesButton = require '../collections/favorites-button'
+ChangeListener = require '../components/change-listener'
+PromiseRenderer = require '../components/promise-renderer'
+NewDiscussionForm = require '../talk/discussion-new-form'
+{Navigation} = require 'react-router'
+
+module?.exports = React.createClass
+  displayName: 'Subject'
+  mixins: [Navigation]
+
+  getInitialState: ->
+    subject: null
+
+  componentWillMount: ->
+    @setSubject()
+
+  setSubject: ->
+    subjectId = @props.params?.id.toString()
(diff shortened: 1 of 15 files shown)


aweiksnar:
One thing I wasn't totally sure of is where to put the subject page. I created app/subjects/index, thinking that other subject specific views may come up in the future.

app/subjects/index.cjsx
```diff
@@ -0,0 +1,81 @@
...
+
+module?.exports = React.createClass
+  displayName: 'Subject'
+  mixins: [Navigation]
+
+  getInitialState: ->
+    subject: null
+
+  componentWillMount: ->
+    @setSubject()
```
brian-c: This should also get called from `componentWillReceiveProps` (`if nextProps.params?.id isn't @props.params?.id`).

app/subjects/index.cjsx
```diff
@@ -0,0 +1,81 @@
...
+
+  render: ->
+    {subject} = @state
+
+    <div className="subject talk content-container">
+      {if subject
+        <section>
+          <h1>Subject {subject.id}</h1>
+
+          <img src={getSubjectLocation(subject).src} />
```
brian-c: Maybe use components/subject-viewer here to keep subject rendering consistent and to make sure eventual non-image subjects are rendered correctly.

app/subjects/index.cjsx
```diff
@@ -0,0 +1,81 @@
...
+              </div>
+            else
+              <p>There are no comments focused on this subject</p>
+          }</PromiseRenderer>
+
+          <ChangeListener target={authClient}>{=>
+            <PromiseRenderer promise={authClient.checkCurrent()}>{(user) =>
+              if user?
+                {# these wrapping promises ensure that there are boards for a project}
+                {# & could be removed if a default board is put in place}
```
brian-c: I'm not entirely sure what's going on here. Does a project need a special discussion board set up for project comments?

aweiksnar: Users of a project can't create boards, only owners can. If a project exists that doesn't have any boards (new project, no talk setup yet, etc) I didn't want to show normal users a form that wouldn't be functional (all discussions need a board). I chatted with heath/darren about scoping here, and we agreed to keep the boards options here scoped to the subject's project to start

app/subjects/index.cjsx
aweiksnar: Ok just added that in https://github.com/zooniverse/Panoptes-Front-End/commit/b4f502a2afd711e714151da13d51a31e90c008e1

app/talk/comment.cjsx
```diff
@@ -76,7 +78,7 @@ module?.exports = React.createClass
     <div className="talk-comment">
       <div className="talk-comment-author">
         <PromiseRenderer pending={@avatarPending} promise={apiClient.type('users').get(id: @props.data.user_id).index(0)}>{(commentOwner) =>
-          <img src={commentOwner.avatar} alt={commentOwner.display_name}/>
+          <img src={commentOwner.avatar ? DEFAULT_AVATAR} alt={commentOwner.display_name}/>
```
brian-c: Eh, sorry, avatars are actually a separate link now: `user.get('avatar').then((avatar) -> avatar.src).catch(-> DEFAULT_AVATAR)`

It should also be cached manually at the component level since requesting the avatar of a user without one throws an error, which doesn't automatically cache. Avatars might be a nice thing to abstract into something like `<Avatar user={user} size="2em" />`.

app/subjects/index.cjsx
brian-c: Got it, I wasn't able to create a board on my oddly named project. Works now.

app/talk/comment.cjsx
aweiksnar: Yeah good idea- I just added in an avatar component in https://github.com/zooniverse/Panoptes-Front-End/commit/07c2429b7b5d441fe906aa785dce690f2e8e1b34. Let me know how it looks

brian-c:
Cool, thanks for that Avatar component.

Is it too early to link to the /subjects/:id page from the classification interface?

Why is linking to the tread still to-do?

chrissnyder:
Pinging @parrish to comment on thread/comment linking.

aweiksnar:
I have discussion (sans comments) linking branched [here](https://github.com/zooniverse/Panoptes-Front-End/commit/d64ccae88cacb7ca596a114cd999f4fd84fa915a#diff-88046daf66764e8c7e5f9d92342b6195R36), I can bring that in, but wanted to :eyes: into ways to improve clarity there

brian-c:
What's done looks good. Linking still to-do.

Issue #435, opened 2015-05-29
Preview site has missing project(s) error on home page


camallen:
The bottom of the home page lists some projects that we've selected. These Project id's are invalid on the prod system and an error message is displayed. 

![missing_project](https://cloud.githubusercontent.com/assets/295329/7881545/11f82eca-05fc-11e5-9609-52504aa6ccbe.jpg)

brian-c:
Closed by #440.

PR #440, opened 2015-05-29
Update project lists

brian-c:
Updates featured projects on production.

Shows appropriate (for now) projects on the projects listing page.

### DIFF
--- a/app/pages/home.cjsx
+++ b/app/pages/home.cjsx
@@ -2,13 +2,17 @@ counterpart = require 'counterpart'
 React = require 'react'
 Translate = require 'react-translate-component'
 {Link} = require 'react-router'
-
 apiClient = require '../api/client'
 PromiseRenderer = require '../components/promise-renderer'
 ZooniverseLogoType = require '../partials/zooniverse-logotype'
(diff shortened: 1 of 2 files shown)


brian-c:
Closes #435.

app/pages/home.cjsx
```diff
@@ -10,6 +9,11 @@ ProjectCard = require '../partials/project-card'
 alert = require '../lib/alert'
 LoginDialog = require '../partials/login-dialog'
 
+FEATURED_PRODUCT_IDS = if process.env.NODE_ENV is 'production'
```
chrissnyder: Why not expose the env from `app/api/config.coffee` and use whatever comes out of that?

brian-c: That'd be nice to develop with, but this way envify will turn `process.env.NODE_ENV` into `"production"`, then uglify will turn `"production" === "production"` into `true` and drop the unreachable code in the build. Not a huge optimization in this case, but probably a good habit to set.

chrissnyder: Sounds good to me.

PR #442, opened 2015-05-29
Remove talk links on production

brian-c:
For beta.

Did I get all the outward-facing Talk links, @aweiksnar?


aweiksnar:
There's one in the post-classification ui, but it only leads to #/todo/talk.

Issue #441, opened 2015-05-29
Repeat every 19th image while logged in


melaniebeck:
I've counted through a three cycles of this now. I may be off by one -- it might be 18 or 20 images before the repeat but I'm pretty sure it's 19.

melaniebeck:
Nope, it's 20. I get 19 new images and the 20th is a repeat every time.

melaniebeck:
When it repeats, the repeat image is from exactly 10 images prior.

edpaget:
Is your project private? Would you mind linking me to it or adding me so I could see if I can replicate?

melaniebeck:
It's public: http://preview.zooniverse.org/panoptes/#/projects/highzgal/gz2redo-take2

edpaget:
This should be fixed once the next deploy finishes.

brian-c:
Just deployed to preview.

PR #445, opened 2015-05-29
Throw out newly fetched subject when it's the one that's just been loaded

brian-c:
Should take care of #441. @edpaget might have a better idea down the line.

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -131,7 +131,8 @@ module.exports = React.createClass
 
     # Take the next subject in the list, if there are any.
     unless upcomingSubjects.forWorkflow[workflow.id].length is 0
-      subject = Promise.resolve upcomingSubjects.forWorkflow[workflow.id].shift()
+      subjectToLoad = upcomingSubjects.forWorkflow[workflow.id].shift()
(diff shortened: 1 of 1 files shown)


app/pages/project/classify.cjsx
```diff
@@ -142,7 +143,8 @@ module.exports = React.createClass
         sort: 'cellect' unless SKIP_CELLECT
 
       fetchSubjects = apiClient.type('subjects').get(subjectQuery).then (subjects) ->
-        upcomingSubjects.forWorkflow[workflow.id].push subjects...
+        nonLoadedSubjects = (newSubject for newSubject in subjects when newSubject isnt subjectToLoad)
```
edpaget: I may just be behind the coffeescript times, but aren't two objects never equal unless they're references to the same object, should this compare ids?

brian-c: The plumbing in the JSON API lib between the request and the resulting objects ensures that if a resource comes back that's already in the local cache, the cached object is updated and returned directly. Two `get`s for the same resource will resolve to the same object.

edpaget: sweet that's pretty nifty.

PR #447, opened 2015-05-29
Get project avatars.

### LINKED ISSUE #439: Project cards lost their avatars
@srallen mind fixing this? I should have caught this in review, sorry.

srallen:
- Refactored project card component to get project avatars
- Fixed #439

### DIFF
--- a/app/partials/project-card.cjsx
+++ b/app/partials/project-card.cjsx
@@ -12,28 +12,32 @@ counterpart.registerTranslations 'en',
 module.exports = React.createClass
   displayName: 'ProjectCard'
 
+  componentDidMount: ->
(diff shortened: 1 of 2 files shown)


brian-c:
Avatar aspect ratios were off for me, that commit fixed it.

@heathv avatars are not square in these cards. Is this intentional?

srallen:
@heathv is out today and Monday. Here's [the design](http://adobe.ly/1FktURG) for reference. They're not square.

brian-c:
Gotcha. That aspect ratio change work for you? Looks good to me otherwise.

srallen:
Yeah, that's better. Thanks.

PR #455, opened 2015-06-01
Make sure errors are reported when creating subjects

### LINKED ISSUE #438: Display subject creation errors
Where's the feedback? Noticed in #436.

brian-c:
Display errors that come up during subject creation.

Once we learn what errors people are having, maybe we can give them something more human-readable.

Close #438.

### DIFF
--- a/app/partials/subject-uploader.cjsx
+++ b/app/partials/subject-uploader.cjsx
@@ -31,6 +31,13 @@ module.exports = React.createClass
     <div className="subject-uploader">
       <p>Progress: <strong>{@state.current}</strong> / {@props.subjects.length}</p>
 
+      {unless @state.errors.length is 0
+        <ul>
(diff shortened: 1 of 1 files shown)


app/partials/subject-uploader.cjsx
```diff
@@ -75,19 +84,21 @@ module.exports = React.createClass
...
-            successes: @state.successes
-            errors: @state.errors
-
-      .catch (error) =>
-        console.error 'TODO: handle linking error', error
-
-      .then =>
-        if @isMounted()
-          @setState inProgress: false
+    unless @state.batch is 0
```
aweiksnar: Maybe `@state.batch.length ` here? It looks like `@state.batch` is always an array

brian-c: Good catch, fixed.

PR #462, opened 2015-06-02
Send project_id with registration

### LINKED ISSUE #372: Capture the project ID a user has been prompted to sign up for
Linked to https://github.com/zooniverse/Panoptes/issues/841

When prompting a user to sign in / signup after X classifications on a project, capture the project id and pass into the registrations end point as `{ project_id: Y }`.

brian-c:
If a user registers because they were prompted to while classifying, the project is now sent along with the registration.

Cleans up the check for available display_name and email before registration.

Cc: @camallen.

Closes #372.

### DIFF
--- a/app/api/auth.coffee
+++ b/app/api/auth.coffee
@@ -103,7 +103,7 @@ module.exports = new Model
         console?.error 'Failed to get session'
         throw error
 
-  register: ({display_name, email, password, credited_name, global_email_communication}) ->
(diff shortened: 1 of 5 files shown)


chrissnyder:
Should the returned user resource have this project_id included?

Front-end all seems good, and sends the project_id along.

edpaget:
I don't think we're sending back the project_id to the user. I think we're just storing it for the welcome email and maybe future analytics.

chrissnyder:
Cool, just checking in case I missed something.

PR #464, opened 2015-06-02
Home v3

srallen:
Another iteration for the homepage:
- Updated hero image
- Inserted current version of copy
- Changed rough design of the content area
- Swapped out svgs for placeholders 

There will be further style updates once a more finalized design is ready.


brian-c:
Looks good.

`line-height` of the slogan is just a bit tight:

screen shot 2015-06-02 at 3 48 56 pm

srallen:
Good catch. I bumped up the line-height on that header so it shouldn't overlap on the mobile size.

PR #466, opened 2015-06-03
Collections manager icon

aweiksnar:
- Adds an interface to manage which collections a subject is a member of
- Adds an icon to manage presenting that interface to the user
- Adds the icon to the subject-viewer


brian-c:
Looks good to me!

PR #486, opened 2015-06-04
Close alert dialog if users click underlay

aweiksnar:


brian-c:
Not sure this is totally safe. Closing a dialog might have side effects, and I don't think a stray click should e.g. abort subject creation.

Dialogs need a lot of work though, for sure.

aweiksnar:
Ok I'll close this up for now.
I typically close modals by clicking outside the main window. It might be having it as a config prop in the future for non mission-critical dialogs

Issue #482, opened 2015-06-04
On closing modal window showing upload limit errors, page becomes unresponsive.


simoneduca:
To reproduce:

1. Upload subject set when you've reached the upload limit.

![screen shot 2015-06-04 at 14 12 17](https://cloud.githubusercontent.com/assets/8989606/7984840/e3280f12-0ac3-11e5-9874-494d9dd6f408.png)
2. Close the modal window.
3. Console shows something is undefined and can't get its length property value.

![screen shot 2015-06-04 at 14 16 43](https://cloud.githubusercontent.com/assets/8989606/7984961/f35b1072-0ac4-11e5-8bc9-d1e05b70f9a2.png)

chrissnyder:
@simoneduca Can you confirm this is fixed and close if so?

simoneduca:
@chrissnyder I can't reproduce step 1 anymore. Did you lift the limit?

brian-c:
This should have been closed with #487, but I guess closing from PR comments doesn't work, sorry.

@simoneduca I believe the limit was upped to 1,000 or 10,000. Display of this number could still be better.

PR #487, opened 2015-06-04
Fix post-subject-creation counts

### LINKED ISSUE #485: Duplicate subject counts after new subject upload
I'm uploading a subject set that has 7849 subjects, each with 2 images. I'm doing this in batches of ~500 at a time. 

Today I've noticed that the subject counts are off. Example: I added 456 subjects with manifest and images, and started the download:
screen shot 2015-06-04 at 15 37 51

Then after the upload is finished, this is what displays:
screen shot 2015-06-04 at 15 44 09

So the number of subjects it says it has added is twice the number of subjects it has actually added (the subject count in smaller font at the bottom is correct), but equal to the number of images it has added.

I'm pretty certain this wasn't happening a couple of days ago when I first started uploading batches of this.

brian-c:
Keep track of subject creation separately from file uploads. Display both numbers afterward.

Closes #485.

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -119,7 +119,8 @@ EditSubjectSetPage = React.createClass
     files: {}
     deletionError: null
     deletionInProgress: false
-    creationSuccesses: []
+    successfulCreates: []
+    successfulUploads: []
     creationErrors: []
 
   render: ->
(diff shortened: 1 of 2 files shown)


brian-c:
Wait two minutes on this one, sorry.

brian-c:
Okay, now fixes #482 as well.

PR #499, opened 2015-06-05
Talk roles

### LINKED ISSUE #373: Interface for talk role assignment
An interface where users with privileged control (moderator, admin, etc...) are able to assign other users to roles of power

edpaget:
Closes #373

### DIFF
--- a/css/user-search.styl
+++ b/css/user-search.styl
@@ -1,0 +1,1 @@
+@import "default.css"
--- a/css/user-search.styl
+++ b/css/user-search.styl
@@ -1,1 +1,1 @@
-@import "default.css"
+@import "../node_modules/react-select/dist/default.css"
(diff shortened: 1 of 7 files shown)


aweiksnar:
This looks awesome! 
I'll definitely use user-search.cjsx for email style messaging in the inbox page as well

css/user-search.styl
```diff
@@ -0,0 +1 @@
+@import "default.css"
```
aweiksnar: I think it may be worth changing up this line to `@import "../node_modules/react-select/dist/default.css"` and then the css import lines in the shell scripts can be just `--include-css  \` so the css is all bundled in the same file

edpaget:
Cool I changed the CSS and added a debounce to the search function. The wait time might need a little tweaking but the 200ms felt right when I was testing it.

aweiksnar:
Sweet looks great!

PR #504, opened 2015-06-08
Keep the avatar component in sync with its user

brian-c:
This will keep avatar from making the same failing request over and over again when a user doesn't have an avatar set.


aweiksnar:
Looks great! Cleaner than what I had :v:

brian-c:
Thanks. What you had would've been fine if the back end returned missing avatars as empty resource responses instead of errors. I think media resources are gonna be weird for a while.

aweiksnar:
Ah yeah, components like this are super nice to have as things change

PR #506, opened 2015-06-08
Reset boards list when props are received

aweiksnar:
- Fixes an issue with project-linker where the boards list was not
  updated when changing between projects


brian-c:
Looks good to me.

PR #509, opened 2015-06-08
Perhaps set width via prop?

chrissnyder:
Putting this out there to see if people like the pattern. Not really ready to merge.

How about setting the width of `<ImageSelector />` via a prop? Somewhat solves the unbounded size problem when people upload big images. Also could be clever and create a mixin which tracks a component's current width as state and passes it down to this.

@brian-c @aweiksnar Thoughts?

### DIFF
--- a/app/components/image-selector.cjsx
+++ b/app/components/image-selector.cjsx
@@ -18,6 +18,7 @@ module.exports = React.createClass
     minArea: 100 # Stop reducing when there are fewer than this many pixels.
     reductionPerPass: 0.05
     onChange: Function.prototype # No-op
+    maxComponentWidth: 200
 
   getInitialState: ->
     working: false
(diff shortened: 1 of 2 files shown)


brian-c:
I don't think a prop is the right place for this. What would happen if the width were just 100% in the CSS instead of 100px? These things should probably be block-like.

I think the reason I put that 100px value on there was to quickly work around a bug in Firefox where large images with max-width: 100% would extend outside flex'd containers without an explicit width. Does that still hold with 100%? I'm not sure I checked.

If that doesn't work, maybe the ImageSelector could be responsible for itself and check its parent on load and resize and update its size if it's too big.

chrissnyder:
Width 100% is a slightly different problem, in that it expands the column to whatever the width of the image is. Basically the same symptoms though.

screen shot 2015-06-08 at 4 56 15 pm

I'll play around with checking it's parent and see how that looks.

aweiksnar:
Loosely related to this, but just wanted to put a link to css [object-fit](https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit) property out there. It may be a nice solution for displaying images of varying sizes (also in project-cards). There's a [polyfill](https://github.com/anselmh/object-fit) too

PR #511, opened 2015-06-09
Linking to single talk comments

### LINKED ISSUE #374: Provide links to invidual comments in a discussion
Coming soon, depends on https://github.com/zooniverse/Talk-Api/issues/33

aweiksnar:
- You should be able to click the ':link: Link' button under a comment to get a route for that individual comment
- The individual comment routes distinguish the comment by giving it an 'active' class (gold background)
- TalkCommentLink component updates
  - The big if/else clause in comment-link isn't my favorite piece of code, but I think it's worth isolating that project/owner params promise to project comment links
  - Takes in a comment resource prop. Should be usable for the search page too!
- Closes  #374

### DIFF
--- a/app/talk/comment-link.cjsx
+++ b/app/talk/comment-link.cjsx
@@ -64,7 +64,9 @@ module?.exports = React.createClass
                   page: pageOfComment,
                   comment: comment.id
                 }
-            <a href={projectCommentUrl}>{projectCommentUrl}</a>
+            <a href={projectCommentUrl}>
+              {@props.children ? projectCommentUrl}
+            </a>
           }</PromiseRenderer>
 
         else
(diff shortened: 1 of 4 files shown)


chrissnyder:
Looks good!

Couple things to comment on:

- I think we'd need the ability to customize the link text itself. In a perfect world, we could pass a component, but even being able to pass text feels necessary.
- I really think we need to centralize where and how we determine what section we are in and then how components adjust accordingly. I don't have a firm idea on how to do this yet (and I think it's outside the scope of this PR anyway).

aweiksnar:
Good call on customizable link text. Just pushed a commit that let's you optionally pass a child component in for that. 

In agreement on the section. I think this it the cleanest way when available, just reading it from the comment. The extra code-verbosity comes up when building links based on that. I'll work on isolating that logic to an exported function in an upcoming PR

chrissnyder:
Sounds like a plan to me.

PR #515, opened 2015-06-09
Activate moderation component

aweiksnar:
- Activates the placeholder moderation component
- Moderation component now takes in a section for context 
- Admins are moderators everywhere
- The discussion.cjsx/board.cjsx changes are just destructuring, indendation, & passing in the section prop to the moderation component


brian-c:
Nice stuff.

PR #517, opened 2015-06-09
Link to talk comments from subject page

aweiksnar:
- Link to individual talk comments from subject page
- ~~Also fix rounding error on getting a comment's page number~~

### DIFF
--- a/app/subjects/index.cjsx
+++ b/app/subjects/index.cjsx
@@ -8,6 +8,7 @@ ChangeListener = require '../components/change-listener'
 PromiseRenderer = require '../components/promise-renderer'
 SubjectViewer = require '../components/subject-viewer'
 NewDiscussionForm = require '../talk/discussion-new-form'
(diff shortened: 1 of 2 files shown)


brian-c:
Looks good to me. Just to confirm: pages are zero-indexed in Talk?

aweiksnar:
They should be one-based: https://talk-staging.zooniverse.org/comments

brian-c:
I think you were right with `Math.ceil` then. See the links for these comments on page 1:

screen shot 2015-06-10 at 10 00 33 am

(Also, I'm only getting three comments per page, is that right?)

aweiksnar:
Ah yep! Just updated to use `ceil`. 3 Per page is just for dev, but can bump that up anytime

brian-c:
Links from subject page look good.

I'm getting the wrong link when creating a new comment. Probably unrelated, I'll open an issue.

PR #523, opened 2015-06-10
Added simple file ext comparison when Windows fails to detect MIME type

### LINKED ISSUE #479: Subject drag and drop of images/metadata unresponsive
cross-posting from panoptes since the url I'm using is panoptes-front-end

Just created an account and a new project and can't seem to upload images and metadata to a subject. Perhaps I'm doing it wrong. It just doesn't do anything at all (the cursor does change to the state usually associated with the location being an acceptable spot to drop files).

This is on win 8.1 using latest chrome - same issue occurs using explorer 11

csv content is:

file,area,location
DSC03210_01.jpg,1,1
DSC03210_02.jpg,1,2
DSC03210_03.jpg,1,3
DSC03210_04.jpg,1,4
DSC03210_05.jpg,1,5
DSC03210_06.jpg,1,6
DSC03210_07.jpg,1,7
DSC03210_08.jpg,1,8

first two example images.

dsc03210_01
dsc03210_02

srallen:
- Added another if statement to `handleFileSelection` when Windows returns and empty string for file.type
  - Fixed #479 
  - When initial browser detection fails, it falls back to OS and in this case Windows uses its system registry keys. If registry key doesn't exist, it returns an empty string for file.type which is why it is failing regardless of browser in Windows. More details:
      - http://stackoverflow.com/questions/1201945/how-is-mime-type-of-an-uploaded-file-determined-by-browser
      - http://stackoverflow.com/questions/26149389/mime-type-missing-for-rar-and-tar
      - http://stackoverflow.com/questions/11182968/determining-unknown-content-types-with-the-html5-file-api

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -210,6 +210,13 @@ EditSubjectSetPage = React.createClass
       else if file.type.indexOf('image/') is 0
         @state.files[file.name] = file
         gotFile = true
+      else if file.type? # When Windows fails to detect MIME type and returns an empty string for file.type
+        allowedFileExts = ['csv', 'tsv']
+        ext = file.name.split('.').pop()
+        if allowedFileExts.indexOf(ext) > -1
+          @_addManifest file
+          gotManifest = true
+
 
       if gotFile and not gotManifest
         @forceUpdate()


brian-c:
Excellent sleuthing! Thanks.

I'm still having some trouble actually completing a cross domain upload from IE (Chrome is fine), but I think that's a separate issue.

Issue #278, opened 2015-05-05
Zoom in on selected part of a subject


aliburchard:
@camallen @eatyourgreens and I talked about this today. There seem to be a small number of projects where this is really fundamentally critical to the task (e.g. retinas), and many others where it's really helpful, if not critical (e.g. penguins, orchids, etc). 

Sounds like one complication is ensuring that marks made while zoomed in are returned in original coordinate space.

camallen:
@eatyourgreens do you have time to work on this in the next week or so?

brian-c:
Zoom is not in scope of first release.

camallen:
@brian-c we will need this for the Retina project, which we plan to run as one of the first zooniverse projects after launch. 

Are you happy to have someone else work on it?

brian-c:
Sure thing. Just want to warn that this involves completely rebuilding the subject viewer and classifier layout.

mkosmala:
We were planning to do a pseudo-zoom for Season Spotter where we simply provided a link to the original image (on a non-Zooniverse site), since zoom was out-of-scope for release. But we need to be able to specify links (Markdown?) in the metadata... See #181  

If 'zoom' changes to be in-scope for release, I'd like to know.

camallen:
@mkosmala don't count on it. Even if we do get developer effort it might be weeks after "launch" before it's available.

eatyourgreens:
I've started the [scaleable-subject](/zooniverse/Panoptes-Front-End/tree/scaleable-subject) branch to work on this, rendering the subject as an SVG image so that we can crop and zoom by changing the SVG viewbox (like Science Gossip.)  I should have time to work on it this week.

Did you have any thoughts on how the UI would work for cropping? Draw a rectangle, and the image crops down to that rectangle when the rectangle is placed?

aliburchard:
Awesome. I think we would want the user to draw the rectangle and then click "next" when they are ready to crop. That allows users to readjust the cropping area if necessary. As long as we specify in the help text that the cropping tool should be the only drawing tool in the task, I think that will be ok.

mkosmala:
I'm a bit confused. Is the point to zoom into a portion of the image? If so, I'd think that zooming would be necessary for all sorts of questions and not a question in and of itself... Maybe I don't understand exactly what the goal is here. 

I think drawing a rectangle and redrawing works. You could also do a single click and redraw with that location as the center of a zoomed-in rectangle (of predetermined dimensions).

[...]

PR #529, opened 2015-06-11
Scaleable subject

eatyourgreens:
Work towards zooming subject images for #278 
- renames SubjectViewer to SubjectAnnotator, to avoid confusion with the shared SubjectViewer component
- adds an SVG image for image subjects
- sizes SVG elements to scale with the subject image
- adds a crop method to crop the view box to x, y, width, height

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -1,7 +1,7 @@
 React = require 'react'
 HandlePropChanges = require '../lib/handle-prop-changes'
 ChangeListener = require '../components/change-listener'
-SubjectViewer = require './subject-viewer'
+SubjectAnnotator = require './subject-annotator'
 ClassificationSummary = require './classification-summary'
 tasks = require './tasks'
 {Link} = require 'react-router'
@@ -54,7 +54,7 @@ Classifier = React.createClass
         currentTask = @props.workflow.tasks[currentAnnotation?.task]
 
       <div className="classifier">
-        <SubjectViewer user={@props.user} project={@props.project} subject={@props.subject} workflow={@props.workflow} classification={currentClassification} annotation={currentAnnotation} onLoad={@handleSubjectImageLoad} />
+        <SubjectAnnotator user={@props.user} project={@props.project} subject={@props.subject} workflow={@props.workflow} classification={currentClassification} annotation={currentAnnotation} onLoad={@handleSubjectImageLoad} />
 
         <div className="task-area">
           {if currentTask?
--- a/app/classifier/subject-annotator.cjsx
+++ b/app/classifier/subject-annotator.cjsx
@@ -11,7 +11,7 @@ getSubjectLocation = require '../lib/get-subject-location'
 NOOP = Function.prototype
 
 module.exports = React.createClass
-  displayName: 'SubjectViewer' # TODO: Rename this.
+  displayName: 'SubjectAnnotator'
 
   getDefaultProps: ->
     project: null
--- a/app/classifier/subject-annotator.cjsx
+++ b/app/classifier/subject-annotator.cjsx
@@ -24,6 +24,7 @@ module.exports = React.createClass
     naturalWidth: 0
     naturalHeight: 0
     viewbox: "0 0 0 0"
+    align: "xMidYMid"
     showWarning: false
     frame: 0
     selectedMark: null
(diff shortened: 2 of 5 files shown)


edpaget:
Any reason we never did anything with this? Is it still useful?

camallen:
It's for the retina project and the start of a zoom tool that may get folded into the front end. I think @eatyourgreens threw it up for reference but probably needs a rebase now. Any ideas on merging this?

eatyourgreens:
Rebased to the latest master.

Should be safe to merge. I haven't really had time to look at adding a UI for zooming in/out of an image.

aliburchard:
@chrissnyder @edpaget @brian-c -- would you guys be able to have a look at this to okay the code for merging?

Note that this doesn't specify the retinas workflow-specific zooming, just the changes to the subject viewer that allows for the retinas-specific zoom. We still need to decide how to meet the need for zoom on a live retinas project, as there is some discussion about integrating custom code with the generalized PFE vs. having a live project exist on a different branch. Will need a conversation with @brian-c and @chrislintott for that.

brian-c:
Talked to @chrislintott about integrating this, but I'm not sure how it integrates so I'm playing with it on #/dev/classifier. Just a couple notes:

- Calling `crop` with a box not matching the aspect ratio of the original image doesn't re-size the view of the subject so the original  shows behind it. I don't think we can render both the  and the <image> at the same time.

- What actually calls `crop`? I'm imagining a crop task that stores x, y, width, and height in an annotation, and during subject render any crop-task annotations are found and viewBox is built from the last one.

eatyourgreens:
@brian-c there's an example on staging:
http://demo.zooniverse.org/panoptes-front-end/retinas/#/projects/aliburchard/ieye/classify?workflow=1039 Draw a rectangle for the third task, and the viewbox changes to that rectangle.

I can't get the original subject to show behind the SVG, no matter what shape rectangle I draw. `crop` changes the viewbox, but it shouldn't touch the height and width of the SVG element.

A crop task that defines a rectangle somewhere on the subject sounds like exactly what the Retinas project needs. At the moment, it's using an ugly hack based on task ID: 1d4d0a1f8b0dcffb20a6ee8e4e7fcd829953d721

brian-c:
Thanks @eatyourgreens. The viewbox changes to the rect, but the SVG doesn't resize so its aspect ratio is still whatever the original image's was. I can see the original behind the cropped image when I draw a tall rectangle off to the side.

screen shot 2015-08-05 at 9 19 51 am

Is Retinas going to live on www.zooniverse.org, or will it be its own deploy?

aliburchard:
I think the plan is for it to be on zooniverse.org so it is maintained
along with the platform (my conversation with @camallen makes me think this
is ideal). But also if we make this a different task type, then it is
generalizable to all projects (and we can authorize projects to access it
with the admin flag)

On Wednesday, August 5, 2015, Brian Carstensen <notifications@github.com>
wrote:

> Thanks @eatyourgreens <https://github.com/eatyourgreens>. The viewbox
> changes to the rect, but the SVG doesn't resize so its aspect ratio is
> still whatever the original image's was. I can see the original behind the
> cropped image when I draw a tall rectangle off to the side.
>
> [image: screen shot 2015-08-05 at 9 19 51 am]
> <https://cloud.githubusercontent.com/assets/464389/9087974/66c79ea4-3b53-11e5-84a9-e1f729c4601f.png>
>
> Is Retinas going to live on www.zooniverse.org, or will it be its own
> deploy?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/zooniverse/Panoptes-Front-End/pull/529#issuecomment-128013908>
> .
>


-- 
Sent from my moderately intelligent phone.

eatyourgreens:
@brian-c Got it. That is a bug. I should alter the default `xMidyMid meet` for `preserveAspectRatio` if they draw a weird rectangle.

eatyourgreens:
Background showing through should be fixed by 87dcad2 (fingers crossed)
I've deployed it to the retinas branch on demo, if anyone has some time to try and break it.

If a project really needs to crop the image down to a thin rectangle eg. lines of text, we'll need to dynamically change the subject height and width, which will probably mess with the `SubjectViewer` component. Something like the specimen labels tool in Orchid Observers, which resizes the subject based on a rectangle you draw: http://www.orchidobservers.org/#/transcribe 
but we can deal with that problem if/when it comes up.

camallen:
@eatyourgreens tried it out [here](http://demo.zooniverse.org/panoptes-front-end/retinas/#/projects/aliburchard/ieye/classify?workflow=1039), seems to work well.

eatyourgreens:
@camallen Cool. Rebased, and moved the crop functionality into a new component.

PR #540, opened 2015-06-11
Determine the Talk host the same way we determine the Panoptes host

### LINKED ISSUE #412: Preview site is using talk-staging
On the preview site, we should just turn talk off for the moment as well as point to the production talk servers (when they come online).

brian-c:
Closes #412.

CC @parrish.


aweiksnar:
LGTM

parrish:
The hostnames are good

PR #543, opened 2015-06-11
Auto save!

brian-c:
Sorry for the giant change, it's mostly indentation I swear. The gist is this: instead of updating resources with forms that save on submit, wrap inputs and buttons in an `<AutoSave resource={resourceToSave}>` component.

When a appropriate change or click or blur event bubbles up to the AutoSave, it'll save any changes to the resource (after a delay, if that makes sense). This makes direct resource changes feel more like link-changing controls that update immediately, so it'll be way less confusing for users.

It also animates its background while saving, turns green if the save was successful, and red if there was an error. Errors show on hover (in the title attribute for now). Some labels should maybe be wrapped in this as well.

I also made some changes to json-api-client to queue up consecutive saves. That shouldn't happen too often, but it can if someone is flying through inputs with the space bar or something.

I think I got everything in the Lab and the settings page. There might be some other inputs I missed. We can get those as we find them.

And please mess around with the workflow editor. I think I hooked it back up right, but it's a super hairy beast.


aweiksnar:
:shipit: Looks solid! I gave things a good test and everything worked as advertised. I'll plug this into a few spots in talk as well

PR #545, opened 2015-06-12
Links tags in sidebar to a search query

aweiksnar:


chrissnyder:
:+1:

Noticed the search input box doesn't set itself when linked to. Will fix that in another PR.

PR #546, opened 2015-06-12
Scroll to comments if being linked to

aweiksnar:
- Scrolls the window to comments if they are linked to, will be more useful as page-sizes grow


brian-c:
I'm cool with this, it works fine. But linking to specific sections on a page is typically done with the hash. What do you think of giving the comment root an `id="comment-#{@props.comment.id}"` and making the links `#/talk/1/2?page=3#comment-4`?

aweiksnar:
Agree that is a better semantic for linking to parts of the page, but I don't think it works as well with react-router currently like it does out of the box in the browser [https://github.com/rackt/react-router/issues/394 + some others]. Since these links are presumably going to be visited on hard page refreshes, perhaps it makes more sense to handle at the component level for now

brian-c:
Eh, gross, oh well.

PR #550, opened 2015-06-15
Disable owner role checkbox and add owner back

### LINKED ISSUE #548: How do you add/remove  gold standard status or talk moderator status to project owner? Owner no longer showing up in collaborator list
So on the preview site for the project I own, I no longer see myself listed. How does one add expert status to myself or remove the ability to moderate Talk for the project?

This is what I currently see in the latest version of Firefox 

screen shot 2015-06-12 at 1 32 58 pm

edpaget:
I think it was confusing not seeing the project owner on the Roles list
at all, so this just adds them back with all the checkboxes
disabled. This also disables the owner checkbox when creating new roles
since there can only be a single owner.

Closes #548

### DIFF
--- a/app/pages/lab/collaborators.cjsx
+++ b/app/pages/lab/collaborators.cjsx
@@ -66,7 +66,7 @@ CollaboratorCreator = React.createClass
           <tbody>
             {for role, _ of POSSIBLE_ROLES
               <tr>
-                <td><input id={ID_PREFIX + role} type="checkbox" name="role" value={role} /></td>
+                <td><input id={ID_PREFIX + role} type="checkbox" name="role" value={role} disabled={role is 'owner'}/></td>
                 <td><strong><label htmlFor={ID_PREFIX + role}>{ROLES_INFO[role].label}</label></strong></td>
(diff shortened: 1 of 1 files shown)


brian-c:
Should the owner's line indicate the implied roles they have?

Can an owner not also be tagged as a researcher?

edpaget:
@brian-c do you mean the explanation of the owner role? I'd agree in that case.

I'd also have no problem reenabling the owner's checkboxes since they don't have any actual affect beside what you present in the Front End.

brian-c:
I think that makes sense. Maybe a line under the owner's that says something like "As the owner of this project, you have all these privileges available. Changing these only affects how you appear to other users."

And should remove the "owner" checkbox altogether?

edpaget:
I thought it was useful to indicate who the owner was. We could also just have a line that says, 'brian-c is the owner' at the top.

edpaget:
@brian-c I'm not sure if the checking the other boxes will even change how the owner appears from #553 which makes it sound like only the most powerful role will be shown.

PR #555, opened 2015-06-15
Use patched react-select to close on select

### LINKED ISSUE #528: Auto-suggest of names in roles is buggy
When you select a name in the drop down the suggestions don't disappear right away if you're using a mouse. So the list then obscures the roles.

edpaget:
Closes #528


aweiksnar:
Sweet looks great. We should check back up on this behavior when react-select 5.0 releases to see if it's still an issue

edpaget:
Yeah I'm going to try to merge mine into master.

Issue #513, opened 2015-06-09
Conversation user search


parrish:
@aweiksnar while working on notifications, I found that the conversation started on #/users/parrish was being started with the wrong user:

`GET https://panoptes-staging.zooniverse.org/api/users?display_name=parrish`

```json
{
  "users": [
    {
      "display_name": "parrish2"
    },
    {
      "display_name": "parrish_temp"
    },
    {
      "display_name": "parrish_signup_test"
    },
    {
      "display_name": "parrishv15"
    },
    {
      "display_name": "parrish_testing"
    },
    {
      "display_name": "parrish_serengeti"
    },
    {
      "display_name": "parrish003"
    },
    {
      "display_name": "parrish124"
    },
    {
      "display_name": "parrish123"
    },
    {
      "display_name": "parrishfamily4"
    },
    {
      "display_name": "parrish_ses_test"
    },
    {
      "display_name": "parrish76543210"
    },
    {
      "display_name": "parrish_no_reply_test"
    },
    {
      "display_name": "parrish_sea"
    },
    {
      "display_name": "PARRISH1108"
    },
    {
      "display_name": "Parrishalexandria"
    },
    {
      "display_name": "ParrishCN"
    },
    {
      "display_name": "parrish001"
    },
    {
      "display_name": "parrish"
    },
    {
      "display_name": "parrish_galaxy_zoo_test_123"
    }
  ]
}
```

The exact match is 18th in the results.  Since it's doing a case-insensitive pattern match, this endpoint is not useful to select a user by name.

@camallen, @edpaget We should either provide an actual filter on the serializer and move the search to a separate endpoint or I'll provide the proper response via the talk api.

I don't have a preference, but I can't imagine this won't be a recurring problem.

edpaget:
That should be filtering on the slug parameter if it's being used in a URL.

edpaget:
To elaborate as far as I knew everything in the front-end was going identify users via the 'slug' parameter for finding them in a human-readable way, since display_names of imported users can have spaces, which we don't want in URLs. If there are uses for wanting to uniquely identify a user by display_name I'll change the endpoint back, and allow you to provide a wildcard to do the search.

parrish:
The slug filter is empty: `window.api.type('users').get({ slug: 'parrish' }) // { users: [] }`

It looks like the controller is querying user groups, but the slug for the 'parrish' user group is 'user-parrish-92'.  What chain of API calls would you have to make to arrive at that?

[...]

PR #557, opened 2015-06-15
Use User#login as username

parrish:
I'd appreciate some eyes on this before a merge.

This is related to zooniverse/Panoptes#960, zooniverse/Panoptes-Front-End#513, and  zooniverse/Talk-Api#42


app/components/user-search.cjsx
```diff
@@ -12,7 +12,7 @@ module.exports = React.createClass
       apiClient.type('users').get display_name: value, page_size: 10
         .then (users) =>
           opts = for user in users
-            { value: user.id, label: user.display_name }
+            { value: user.id, label: "@#{ user.login }: #{ user.display_name }" }
```
edpaget: Might make sense to only include the display name or reverse the order it's being shown since it's searching by display_name.

edpaget:
This looks like it has all the places to me. I'm going to do another check tonight. 

Also we'd talked about having a pop-up for users with migrated accounts giving them a chance to change their login name from whatever we assigned them.

app/components/user-search.cjsx
parrish: I was basically just mimicking most @-mention systems (e.g. github) where they show something like

*@user1* _John Doe_
*@user2* _Jane Doe_

But I guess since display names are unique, it doesn't matter that much.

edpaget:
@parrish can you rebase this. I think slugs have also been added to some of the collections URLs as well. I'll merge it once the backend migration finishes.

parrish:
@edpaget Rebased

PR #559, opened 2015-06-15
Speed up project load by including the owner

edpaget:
Reduces project load times by about a second on my machine, by
including the owner. Still working out how to do the same for the avatar

### DIFF
--- a/app/partials/project-card.cjsx
+++ b/app/partials/project-card.cjsx
@@ -1,7 +1,7 @@
 React = require 'react'
 PromiseRenderer = require '../components/promise-renderer'
 {Link} = require 'react-router'
-
+apiClient = require '../api/client'
 counterpart = require 'counterpart'
 Translate = require 'react-translate-component'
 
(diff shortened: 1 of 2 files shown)


app/partials/project-card.cjsx
```diff
@@ -12,6 +12,9 @@ counterpart.registerTranslations 'en',
 module.exports = React.createClass
   displayName: 'ProjectCard'
 
+  projectOwner: ->
+    apiClient.type(@props.project.links.owner.type).get(@props.project.links.owner.id)
```
brian-c: Is this necessary? With the include above, this resource should already be available without a new request. Accessing `links` directly makes things more fragile, since links can be defined a bunch of different ways.

edpaget: When I did project.get 'owner' after doing the include it was still making an http request. I can try fixing this in the json-api-client library since I'll need to do that anyway for the avatar.

brian-c: Hm, it hasn't been re-requesting for me.

What're you planning on doing for the avatar?

edpaget: I was thinking about having the client cache a subject by what it's constructed URL would be. So instead of looking to see if its ID already existed in the cache it would instead look at what URL it was about request and if it had already requested that.

PR #565, opened 2015-06-16
Parse subject/tag/user mentions to links

### LINKED ISSUE #556: Make mention links in markdown go to subject/user/tag

aweiksnar:
- Parses markdown in comments for subject mentions `^subject_id`, user-mentions `@user_name` and tags `#tagname`, and makes the links go to the correct location
- I'm curious on other ideas to handle the redirect to the subject page (or maybe this is fine?). I chose this way because the links could be parsed quickly as strings, and only make the api request for the project & owner if clicked
- Closes #556


edpaget:
So subjects can be a part of more than one project are their talk pages always going to be under the project they were originally uploaded for?

aweiksnar:
Oh interesting, didn't think of that. This page could also be rendered in a non project-specific context (would solve the redirect issue), but feels kinda nice to be part of talk. I think since subjects can be cross-referenced in any talk, there's no reliable way to decide which talk to direct users to. Open to other ideas if just using subject.get('project') seems problematic

edpaget:
I think it makes sense to have subject pages per project since the discussion will probably be different depending on what project they're used in. 

They could work like Github issues {owner}/{project}^{subject_id} if you want to reference a cross project subject page and just ^{subject_id} if you want to reference a subject within it's project's talk (so presumably you wouldn't need to reload the project resource).

aweiksnar:
Ah so would adding a 4th regex match for the full url to a user/project/subject work for you?

edpaget:
Yeah, but only if that makes sense to you (to have the per project subject talk pages).

aweiksnar:
I added support for that case in here https://github.com/zooniverse/Panoptes-Front-End/commit/6c0fd29535de59f5236dae8ae7a14d68173cfd28. I have a feeling that users will end up using regular markdown links for that case, since it'd be tough to remember all the specifics of the project url (at least until I put in an autocompleter in the text entry area). 
Also, that page is easy to drop in place anywhere, so if users don't find it handy it could become a modal in the future

brian-c:
What if subject IDs were always displayed with their project ID? ^123-234 and ^543-234 would technically both refer to the same subject, but the project ID would keep the context for redirecting the URL and querying the conversations.

edpaget:
I don't think we ever expose the project ids anywhere though.

brian-c:
We'd just expose it as part of the subject faux-ID. We wouldn't tell users it's the project ID, it'd just be part of the subject.

edpaget:
@aweiksnar awesome sorry for being such a pain about this.

aweiksnar:
No worries! I'm happy to sort these details out early rather than have to keep updating things as panoptes grows. I also didn't totally understand how subject sharing worked in panoptes, so this was helpful for diving into that a bit more, i.e the subject~project json-api link

brian-c:
Project names can easily change. Are we sure this is how we want to do this?

aweiksnar:
@brian-c It's expecting a project slug, which should get updated with name changes

brian-c:
But the comment won't get updated, will it?

edpaget:
We should be warning people about breaking links when they change project names anyway right?

brian-c:
Yep, just opened #603.

aweiksnar:
Ahh yeah, We'll hit the same issue if people use markdown to create links too, so a warning would be smart if this ends up as the final syntax or not. I'll think on a more robust solution for subject linking. Project IDs may be the only guarantee here

edpaget:
I really don't think we should worry about it beyond warning people about changing the name. At some point we could track old names and redirect if it really becomes an issue.

PR #576, opened 2015-06-17
Retina workflow by id not name

camallen:
Allow workflow name changes, use the id now. Also allow the workflow Id be configurable by the env.

@eatyourgreens - can you take a look?


eatyourgreens:
Looks good.

PR #580, opened 2015-06-17
Project cards are now owned cards

### LINKED ISSUE #579: Preview project listing is broken
http://preview.zooniverse.org/panoptes/#/projects

image

brian-c:
Close #579.

These new "OwnedCards" seem to be conflating some pretty unrelated stuff. Why use the same component to represent projects and collections?


edpaget:
I reused them since they work in exactly the same way. They create a link to the collection page, which is based on the same user_name/slug pattern as a project. It didn't make any sense to me to duplicate that code.

PR #582, opened 2015-06-17
Limit subject looping

brian-c:
- Renders the /dev/classifier page even though it doesn't have a project or owner
- Reduces the speed of subject frame playing
- Stops frame playing after three loops


aweiksnar:
Looks good, can you rebase with master and than I'll merge it in?

Issue #564, opened 2015-06-16
React warning for classifier/subject-viewer


eatyourgreens:
Probably nothing serious, but React dev tools throw this warning in the console for the SubjectViewer
`Warning: SubjectViewer is accessing getDOMNode or findDOMNode inside its render(). render() should be a pure function of props and state. It should never access something that requires stale data from the previous render, such as refs. Move this logic to componentDidMount and componentDidUpdate instead.`

I might have fixed this in #529

chrissnyder:
Could you isolate the problem from your PR and send a separate PR with just the fix? I'm hesitant to mess with the subject viewer  too much so close to launch.

eatyourgreens:
Sure, no problem.

eatyourgreens:
Closed by #597

PR #597, opened 2015-06-18
Don't access @refs during render()

eatyourgreens:
Store sizeRect and toolRect in component state.
Use stored sizeRect to calculate scale and details tooltip properties.
This should fix the React warning in #564


brian-c:
Works for me. This ready to merge @eatyourgreens?

eatyourgreens:
Cool. I'll merge it then.

PR #599, opened 2015-06-18
Publications team pages

### LINKED ISSUE #593: Publications page
I believe @srallen has already made progress here, but making this an explicit issue to close when that work is merged in.

srallen:
- Added publications page component with content and styles
- Updated team page so all bios are present
- Fixed #593

Note: I started to add the sort by project on the publications page, but I decided to come back to that as there are other pages of the front end that need to be worked on. I'd also appreciate someone double checking that I got all the projects (that have publications) and their publications on the page.


app/pages/about.cjsx
```diff
@@ -104,10 +102,13 @@ module.exports = React.createClass
         </div>
       </section>
       <section className="about-page-content content-container">
-        {if @state.activeTab is 'team'
-          <TeamPage />
-        else
-          <Markdown>{counterpart "about.pageContent.#{@state.activeTab}"}</Markdown>
+        {switch @state.activeTab
```
brian-c: Could we use react-router's RouteHandler component here now instead of tracking activeTab in state?

app/pages/publications-page.cjsx
```diff
@@ -0,0 +1,331 @@
...
+  updateButtonState: (currentButton) ->
+    buttons = React.findDOMNode(@refs.sideBarNav).childNodes
+    for button in buttons
+      button.classList.remove 'active'
+    currentButton.classList.add 'active'
+
+  getAvatar: (project) ->
+    apiClient.type('projects').get(slug: project.slug).get('avatar')
+      .then (avatar) => @avatarSrc = avatar.src
+    @avatarSrc
```
brian-c: This isn't going to return the expected value. Might need a PromiseRenderer in the render method.

app/pages/publications-page.cjsx
```diff
@@ -0,0 +1,331 @@
...
+
+  showPublicationsList: (navItem) ->
+    currentButton = React.findDOMNode(@refs[navItem])
+    @setState currentSort: navItem, -> @updateButtonState(currentButton)
+
+  updateButtonState: (currentButton) ->
+    buttons = React.findDOMNode(@refs.sideBarNav).childNodes
+    for button in buttons
+      button.classList.remove 'active'
+    currentButton.classList.add 'active'
```
brian-c: This should probably be handled in `render` based on the current state. Won't matter if we switch to filtering by a query param.

srallen:
I've reworked the code based on all of your comments. I also reorganized the about components into its own sub-folder too.

app/pages/about/index.cjsx
```diff
@@ -0,0 +1,95 @@
...
+
+counterpart.registerTranslations 'en',
+  about:
+    title: 'About Us'
+    nav:
+      about: 'About'
+      publications: 'Publications'
+      ourTeam: 'Our Team'
+      careers: 'Careers'
+    pageContent: '''
```
brian-c: Leftover pageContent?

brian-c:
Looks good except that extra block o' content.

I don't think sorting by project is necessary. By category should be enough, right?

Are we not interested in getting the queried state from the URL?

brian-c:
Oh, and do you mind rebasing? I'm seeing a bunch of references to `owner.slug`, which is no longer a thing.

app/pages/about/index.cjsx
srallen: Whoops. That's cleaned up now.

srallen:
Alright, I've rebased. For the sorting, doing it by project is something you can do now and it's something I can imagine someone would want to do, but it isn't something that needs to happen for first launch since so many other areas still need work. Also, we could do queried state by URL. I don't have a strong opinion either way.

PR #611, opened 2015-06-18
Add button to create subject data export

### LINKED ISSUE #610: Add subject export button
Similar to the classifications export just goes to `/projects/{id}/subjects_export`

edpaget:
Closes #610

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -5,10 +5,16 @@ PromiseRenderer = require '../../components/promise-renderer'
 ImageSelector = require '../../components/image-selector'
 apiClient = require '../../api/client'
 putFile = require '../../lib/put-file'
-moment = require 'moment'
+counterpart = require 'counterpart'
+DataExportButton = require '../../partials/data-export-button'
 
 MAX_AVATAR_SIZE = 64000
 MAX_BACKGROUND_SIZE = 256000
+
+counterpart.registerTranslations 'en',
+  projectDetails:
+    classificationExport: "Request new classification export"
+    subjectExport: "Request new subject export"
 
 ExternalLinksEditor = React.createClass
   displayName: 'ExternalLinksEditor'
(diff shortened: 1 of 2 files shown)


edpaget:
Again will need to wait for zooniverse/Panoptes#999 to be merged.

edpaget:
The backend deploy for this should be done now

app/partials/data-export-button.cjsx
```diff
@@ -0,0 +1,48 @@
...
+module.exports = React.createClass
+  displayName: 'DataExportButton'
+
+  getInitialState: ->
+    exportRequested: false
+    exportError: null
+
+  requestDataExport: ->
+    @setState exportError: null
+    apiClient.post @props.project._getURL(@props.exportType), media: content_type: 'text/csv'
```
brian-c: I don't think `project` is in props.

app/pages/lab/project-details.cjsx
```diff
@@ -110,6 +114,10 @@ module.exports = React.createClass
       .catch ->
         []
 
+    @subjectsExportGet ?= @props.project.get 'subjects_export'
```
brian-c: Might make more sense to cache this in the buttons that request them now.

edpaget:
Thanks should be fixed now.

app/partials/data-export-button.cjsx
```diff
@@ -0,0 +1,53 @@
...
+  exportGet: ->
+    @props.project.get @props.exportType
+      .catch ->
+        []
+
+  requestDataExport: ->
+    @setState exportError: null
+    apiClient.post @props.project._getURL(@props.exportType), media: content_type: 'text/csv'
+      .then =>
+        @classificationsExportGet = null
```
brian-c: I think this was left over from the old page, and it looks like the request is no longer actually being cached, so a project without these resources makes the request every time the component renders which is every time the project resource changes.

Issue #404, opened 2015-05-25
Adding a description of what the lab is when non-logged in users navigate there


mschwamb:
Currently all you get is 'You need to be signed in to use the lab' if you navigate to the lab page when not logged in. It might be worth a better description of what the lab is especially for new researchers that might be checking out the Zooniverse home page but don't yet have an account. 

![screen shot 2015-05-25 at 11 07 55 am](https://cloud.githubusercontent.com/assets/753325/7791192/5b6f637e-02ce-11e5-8d06-9c457a8ab792.png)

vrooje:
Possible text we could use for that page:


The project builder allows anyone to create a collaborative research project. 

You can use the project builder to upload your data, specify what tasks need to be completed by your team of other researchers and volunteers, and tell people about your research goals.

You need to be signed in to use the project builder. New builders might be interested in our step-by-step guide to building a project in the Zooniverse (INSERT LINK when it exists).

vrooje:
tagging @chrislintott @mrniaboc @DarrenMcRoy to check the text.

mrniaboc:
Should perhaps add a big 'sign in' button at the end?

DarrenMcRoy:
That content looks perfectly good, @vrooje -- no objections or edits here.

I agree with @mrniaboc that a big ol' sign-in button is a very good idea.

chrissnyder:
@mrniaboc You had a possible design for this already, correct? If so, could you share it with Sarah?

PR #614, opened 2015-06-19
Remove home page specific typography rules

chrissnyder:
- Home page specific typography rules causing the dialog to render incorrectly.
- Removed them so the font size baseline is the same on the home page as everywhere else.

Before:

After:

### DIFF
--- a/css/home-page.styl
+++ b/css/home-page.styl
@@ -27,7 +27,7 @@
 
     .hero-tagline
       color: inherit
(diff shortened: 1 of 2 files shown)


srallen:
@chrissnyder, I actually noticed this when I was working on #404 and have a fix in my PR for it.

chrissnyder:
Roger that, no need for this then! Closing.

PR #615, opened 2015-06-19
Add prompt to set login name for migrated users

### LINKED ISSUE #598: Prompt migrated users to set their login
When a user first logins into Panoptes with a pre-existing Zooniverse account, we should show them the login name we've given them for @-mentions and their url slug and allow them to change it if they want. This is the only time we should allow those to be changed.

edpaget:
Triggered by a user.login_prompt field on serialized response

Closes #598

### DIFF
--- a/app/partials/login-change-form.cjsx
+++ b/app/partials/login-change-form.cjsx
@@ -5,7 +5,8 @@ LoadingIndicator = require '../components/loading-indicator'
 
 counterpart.registerTranslations 'en',
   loginChangeForm:
-    updateMessage: "Other Volunteers will be able to mention you in Talk as @%(login)s and find your profile at #/users/%(login)s. You can change it below, but this is permanent"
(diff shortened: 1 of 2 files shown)


edpaget:
This depends on zooniverse/Panoptes#1000 Also I'm not sure what the text for the field should be @DarrenMcRoy any thoughts?

DarrenMcRoy:
First draft of content is in: https://docs.google.com/document/d/17JL1SHuSEng0rM3YjDthLSARoPRGhbeaaWAJ_CdnZ5A/edit

edpaget:
This should also be good to go. @aweiksnar I've just been editing https://github.com/edpaget/Panoptes-Front-End/blob/login-prompt/app/partials/login-dialog.cjsx#L48 to be `if true` for most of my testing. Let me know if you also want me to set the flags on your Panoptes account to see it that way.

aweiksnar:
Sweet! Just tested it out and looks like it works great

PR #625, opened 2015-06-19
Show open connections

brian-c:
This is just a little box that pops up when there's a server connection open that tells the user how many requests the front end is waiting on. I think it'll help reduce confusion about when things are saving. I'd like to see what folks think.

Depends on a new version of the JSONAPIClient, so `npm install`.

Ran into some nasty behavior solved by the noted timeout. I'm thinking it has to do with the router, but I couldn't track it all the way down. I think the timeout is fine for now.


aweiksnar:
Looks like it works great! I'd love to see this become the main global loading state. It may be worth playing with a short delay on it too

Issue #535, opened 2015-06-11
User profiles


alexbfree:
Pretty sure this is already planned, but since I don't see an issue, I'll create one. Currently when you click a username in Talk, it goes to a "user profiles todo" page.

srallen:
Yes, @heathv is working on several designs that I'm implementing in the user-profile branch.

edpaget:
I think we have enough for the first release.

PR #626, opened 2015-06-19
User profile

srallen:
- Reworked user profile routes, subnav, styles, and settings components for #535 
  - Placeholder components for feed and stats
  - Linked to collections and talk inbox

Work still need to be done, but can be updated next week:
  - Fetch user display name when viewing public profile and not logged in
  - Fetch user header image and display in hero area


brian-c:
Nice start.

Loading a "Messages" page errors with `Missing "name" parameter for path "/users/:name/inbox/:conversation"`, which breaks everything until reload (thanks, react-router). Will merge once that's fixed.

General questions:

- Why put settings in the profile?

- Are we not going to do a  Facebook-style one-conversation-per-pair of users?

- The :email: icon in the header goes to my profile's messages section, which makes it seem like the same page does two different things depending on who's profile it is. Like settings, I think the inbox stands alone.

edpaget:
Things are rebased and the settings and inbox are moved back to the main level.

brian-c:
Navigation's still odd to me, but I'll merge and we can fix it later. Thanks.

PR #632, opened 2015-06-22
Fill in footer links

edpaget:
Add links for everything in the footer. I removed 'Education' since we
don't have a page for that yet. And 'Other Talk' since I'm not sure what
it's for. Let me know if those should be added in and where they should
point to.


aweiksnar:
Sweet! I believe those those other links were just in there as placeholders

PR #635, opened 2015-06-22
Enable project searching

aweiksnar:
- Takes the current zooniverse wide search and uses it in the project scope


aweiksnar:
btw @chrissnyder props on making the search component work out of the box for this :v:

chrissnyder:
Related problem, but https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/comment-link.cjsx#L49 should look at the end half of the split if project sections are going to look like `project-PROJECT_ID`

aweiksnar:
Ah good call--That was my fault for not updating those in d36b2b7694f15f6c1232fef6f997895059d01db2. I'll do something similar to format how those are parsed

aweiksnar:
Just exported a function for that - Should be all under the same format inline with the api now `"project-#{id}" or "zooniverse"`

chrissnyder:
:+1:

PR #639, opened 2015-06-22
Disable name change for live projects.

### LINKED ISSUE #603: Warn that changing a project's name breaks its links
Maybe only on live projects?

edpaget:
Always indicate what a project's URL will be under the name field so
people know that editing the name will change it.

Closes #603

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -74,6 +74,20 @@ module.exports = React.createClass
     backgroundError: null
     exportRequested: false
     exportError: null
+    currentSlug: @props.project.slug
+    currentName: @props.project.display_name
+
+  projectUrl: ->
+    @props.project.get('owner')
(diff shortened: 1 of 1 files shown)


app/pages/lab/project-details.cjsx
```diff
@@ -74,6 +74,21 @@ module.exports = React.createClass
...
+    @props.project.get('owner')
+      .then (owner) =>
+        "#/projects/#{owner.login or owner.name}/#{@props.project.slug}"
+
+  undoNameChange: ->
+    @props.project.update display_name: @state.currentName
+    @props.project.save()
+    
+  warnURLChange: ->
+    console.log
```
aweiksnar: I'd take out this empty log and then looks good to go :+1:

edpaget: got 'em

PR #641, opened 2015-06-22
Prevent leaving the page while writing

brian-c:
The IOStatus component now listens on `beforeunload` to prevent the user from leaving the page in the middle of a write and potentially losing data.


edpaget:
What's supposed to cause this to trigger? I can't get any confirmation to come up.

brian-c:
It should pop up if you try to close the page while the IO status at the top has an "up" number greater than 0.

edpaget:
Ah cool got it.

PR #642, opened 2015-06-22
Update projects listing query

brian-c:
I think this is right. Projects can't be `launch_approved` without being `live`, correct?

Try `/?env=production#/projects` on a locally-running client to see this with some projects.


edpaget:
Eh there's no hard requirement on that being the case. It's up to whoever launch_approves to make sure the flags are right.

PR #643, opened 2015-06-22
Only list live projects

### LINKED ISSUE #636: Remove all demo and non-certified projects from project list
I know @chrislintott said to wait on posting all of the issues @alexbfree reported, but it reminded me that @chrissnyder actually asked me to put this particular one up before:

We need to clear out the displayed project list to go back to only certified projects. We also need to make sure that new projects don't appear automatically in that list... ? Because they need to be certified before they get on that list, right?

brian-c:
I wasn't quick enough to sneak this in.

Closes #636, too.

### DIFF
--- a/app/pages/projects.cjsx
+++ b/app/pages/projects.cjsx
@@ -20,6 +20,7 @@ module.exports = React.createClass
 
   listProjects: ->
     query =
+      live: true
       launch_approved: true
       include:'owners,avatar'
     Object.assign query, @props.query


edpaget:
You can't actually filter by that. Do we need to? I assume we'd only approved projects with the 'live' setting.

brian-c:
Oh. I thought your comment on the last one said otherwise. I'll close this.

PR #653, opened 2015-06-22
Update registration form username validation

### LINKED ISSUE #631: Registration Form should mention username/password restrictions
It should note that usernames can't have spaces and some special characters and that passwords need to be at least 8 characters.

edpaget:
Makes sure it's the same as the backend's validation. Make the
error messages a little better, and make the form bet invalid when the
username is blank.

Darken the disabled button state, which is hard to see on some monitors.

Closes #631

### DIFF
--- a/app/partials/register-form.cjsx
+++ b/app/partials/register-form.cjsx
@@ -161,7 +161,7 @@ module.exports = React.createClass
       <br />
 
       <label>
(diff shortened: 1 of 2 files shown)


edpaget:
Also made email opt-out.

aweiksnar:
Awesome looks like a good update :metal:

PR #663, opened 2015-06-22
Disable tasks until subject is loaded

### LINKED ISSUE #176: Classification allowed before subject loads
If loading is taking a while for a subject, I am still allowed to answer questions about it before it loads (and while the "please wait" planet is spinning). Better behavior would be to freeze the questions until the subject loads.

brian-c:
Closes #176.


aweiksnar:
Awesome!

PR #670, opened 2015-06-23
There should always be a reset password link

edpaget:
...on the sign in form. People should be able to click reset without
having to messup their login


aweiksnar:
:shipit: Good call

PR #671, opened 2015-06-23
Increase size of talk user name/role, closes #652

### LINKED ISSUE #652: Talk identifiers for science teams, moderators,  and Zooniverse teams is tiny and unreadable
At least on Firefox I can barely read the label that says researcher. I nearly missed it, and I think most volunteers will. Maybe it's not the final version, but I thought I'd bring it up. 

screen shot 2015-06-23 at 3 32 29 am

I also really liked the way the current non-Panoptes talk highlighted post box with a different color for each role to emphasize the fact that Zooniverse, moderators and Zooniverse were something different.

aweiksnar:


chrissnyder:
:+1:

PR #675, opened 2015-06-23
Filter talk roles by visibility

parrish:
This supports the use case of a collaborator needing a role, but not wanting to be publicly identified as that role -- e.g. a scientist needs moderation access, but only wants to be known as a researcher.


aweiksnar:
Awesome, Thanks!

PR #676, opened 2015-06-23
Discussions update

### LINKED ISSUE #613: If you navigate from a project with talk boards to one that doesn't using the project talk side bar then the old project's boards stay with new project's background
If you use the project Talk side bar to move to a project's talk without discussion boards, it leaves the previous projects boards in place 

see screenshots below:

I made this boards and threads for P4: Terrains:

screen shot 2015-06-19 at 6 05 51 am

Switched to Kitteh Zoo and the same P4: Terrains boards show

screen shot 2015-06-19 at 6 05 37 am

Same thing if I go to Snapshots at Sea

screen shot 2015-06-19 at 6 05 27 am

aweiksnar:
- Changes `current-section` to a function
  - Having this as a function works better than a mixin with Router#transitionTo, to ensure component is updated when props are received
  - Closes #613
- Also adds a `shouldComponentUpdate` to the project-linker, to prevent rerenders when changing between projects (looked a bit ugly with the loading state happening on every page change)


chrissnyder:
Works as advertised. Couple comments inline.

app/talk/index.cjsx
```diff
@@ -29,5 +29,5 @@ module?.exports = React.createClass
         </button>
       </form>
 
-      <RouteHandler {...@props} section={@state.currentSection} />
+      <RouteHandler {...@props} section={currentSection(@props.project)} />
```
chrissnyder: Couldn't this just be `<RouteHandler {...@props} section="zooniverse" />`? And then have the project-specific Talk instance use the `lib/project-section` function directly?

aweiksnar: :pensive: Yep, updated in c5e193d

chrissnyder:
LGTM

PR #678, opened 2015-06-23
Fix image uploaders

### LINKED ISSUE #655: some of the project plackards are pixelated on panoptes Zooniverse home
- chimp&see, Higgs hunters, Chicago Wildlife Watch, Orchid Observers, Science Gossip - the image is too pixelated at the current resolution being used compared to other project plackards like Planet Hunters and Galaxy Zoo 

screen shot 2015-06-23 at 4 09 56 am

screen shot 2015-06-23 at 4 11 30 am

brian-c:
Image uploader no longer tries to switch formats. It should end up with more sensible images, so this closes #655. We can re-open if it doesn't help, but then the only thing we can do is reject images that's too large, which I feel like would be annoying.

### DIFF
--- a/app/components/image-selector.cjsx
+++ b/app/components/image-selector.cjsx
@@ -2,9 +2,15 @@ React = require 'react'
 LoadingIndicator = require './loading-indicator'
 toBlob = require 'data-uri-to-blob'
 
-BASE_64_PNG_HEAD = 'data:image/png;base64,'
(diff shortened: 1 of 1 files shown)


aweiksnar:
Great! I just played around with it a bit and seems like it work well. Yeah I agree that it's cool to not have users need to check their image sizes

PR #679, opened 2015-06-23
Declare position and z-index for main header on project page

### LINKED ISSUE #667: can't access profile or signout if menu crosses a navigation bar link
Currently the sign out/profile menu covers the Spotter's Guide link when it comes down. When that happens you can't mouse over onto the  signout/profile menu

screen shot 2015-06-23 at 2 20 25 pm

When I mouse over it keeps going to the Spotter's guide link and not to profile

chrissnyder:
- Closes #667 
- I'm getting a bit lost at the purposes of all the different `on-some-page` selectors, but I think this is the only page where this would be a problem.

### DIFF
--- a/css/main-header.styl
+++ b/css/main-header.styl
@@ -1,6 +1,7 @@
 .main-header
   font-size: 10px
   line-height: 20px
+  z-index: 100
 
   .on-home-page &
     position: absolute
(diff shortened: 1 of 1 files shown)


aweiksnar:
Yeah looks better than it was. I'm thinking this may need to be updated to an onclick handler for touch devices, but I'll open a separate issue for that

PR #682, opened 2015-06-23
Use and set focus_type on comments about subjects

parrish:
This field is required if focus_id is set.  Talk will have a validation shortly to enforce this.


aweiksnar:
Thanks!

PR #686, opened 2015-06-24
Fix collection show routes.

edpaget:
Also improve collection page design


edpaget:
I need to fix some roles filtering on the backend to lock the Collection Settings and Collection Collaborators pages.

edpaget:
Actually this isn't close to ready yet.

PR #733, opened 2015-06-24
Fixing favorite button to set flag

edpaget:
Also has the favorites button make fewer requests

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -54,7 +54,7 @@ Classifier = React.createClass
         currentTask = @props.workflow.tasks[currentAnnotation?.task]
 
       <div className="classifier">
-        <SubjectViewer subject={@props.subject} workflow={@props.workflow} classification={currentClassification} annotation={currentAnnotation} onLoad={@handleSubjectImageLoad} />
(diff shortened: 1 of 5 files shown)


aweiksnar:
:+1: :+1: Much cleaner than before

edpaget:
@aweiksnar Just FYI this adds the idea of a global Favorites collection which it will add subjects to unless you explicitly pass a project property to the component.

aweiksnar:
:ok: cool, Yeah it'll be neat to see how users end up using global collections

edpaget:
Yeah I'm still making a collection scoped the project when you favorite from the classification interface. I think there are other places that button shows up though.

aweiksnar:
It's in all talk comments that mention a subject and the talk object page. They may have to remain separate for the favorite button since there's no way to safely resolve a subject's project without asking the user to do it

edpaget:
Cool yeah I was briefly looking to see if Project Talk pages had a project property I could pass down to the subject, but couldn't find it.

aweiksnar:
Yeah I'll can look into adding that in, so far it's just been reading from the url

Issue #728, opened 2015-06-24
Link tiles need to display avatars as squares


DarrenMcRoy:
Pertaining to #655 and #681 --

All avatars that are uploaded are automatically cropped to be squares.

But currently, the link tiles display as rectangles, meaning that it is literally impossible to have your image fit properly.

This needs to be resolved in one direction or the other. :-)

DarrenMcRoy:
(These tiles also need descriptions, as per #662.)

srallen:
This is fixed.

Issue #703, opened 2015-06-24
Improve pagination and navigation in project list


alexbfree:
There are a few problems with the current project list, in particular showing where you are, and how you navigate it:

Firstly, the message at the top
![showing 20](https://cloud.githubusercontent.com/assets/1473244/8332731/599d1840-1a87-11e5-9db0-3cbdbe5b54be.png)
* This gives no indication that there is more than one page, unless you scroll down. Could change it to "1 to 20 of 34" with a little Next page logo
* The wording is odd. Makes it sound like search results - which is not we want for the project list. We should give the impression this is a curated catalog, not the results of a search.

Secondly, the buttons at the bottom:
![buttons](https://cloud.githubusercontent.com/assets/1473244/8332782/960540aa-1a87-11e5-8360-c46384e55d53.png)
* These are tiny and hard to click. They look unstyled. They should be much bigger, and need some colour/styling/formatting.
* How will this handle large numbers of pages? Shouldn't we have previous/next and/or first/last buttons too. Are there "..."s when there are lots of pages?
* There is no indication of the current page. There should be a difference between the current page and the other pages.
* Are these even the right way to go? Pagination is dying out... Wouldn't an AJAX infiniscroll technique a la Tumblr be preferable?

srallen:
@alexbfree, @heathv's original design for this page used a "Show more" button at the bottom to load more like an infinite scroll rather than pagination and then the number at the top would update too. I just hadn't gotten to it yet. I agree that an infinite scroll type page is better, but am not sure if it is necessary to get done for the soft launch. At least, the pagination buttons should be styled.

alexbfree:
Thanks Sara for addressing my comments. Definitely looks a lot better! (though I would consider changing the word "found" for something like "projects" so it looks less like search results).

I agree infiniscroll is not a short term priority, maybe @chrissnyder this could be reopened and added to the post-soft-launch UI/UX list? Or maybe implementing infiniscroll per heath's design is already covered elsewhere?

Issue #688, opened 2015-06-24
Blue banner doesn't look great - neither do the get started buttons


mrniaboc:
The blue banner used on the /projects and /about pages is the exact colour you get if you highlight a section of a page in your browser, so looks quite weird. I really like the dimmed/opaque effect on the main landing page, and suggest sticking with this theme throughout.

alexbfree:
I would also add a related comment; the blue buttons that say "Get Started" on each project don't look great either... Looks rather like a bootstrap template or auto-generated button... might look better with a two-tone effect, a shadow, a gradient and/or removal of bold.

DarrenMcRoy:
I agree with these. The blue has really been throwing me off. The dimmed effect on the landing page looks much better.

aliburchard:
Agreed with all the above.

alexbfree:
Regarding the buttons, here is the current button:
![current button](https://cloud.githubusercontent.com/assets/1473244/8331700/9c4da052-1a81-11e5-9d5d-870495b30282.png)

Here's a slightly improved version: 2px line, 2px shadow, and a light blue background.

![better version?](https://cloud.githubusercontent.com/assets/1473244/8331721/b3affa92-1a81-11e5-985b-196a6d2078b7.png)

Looks good on hover too:
![hover with shadow](https://cloud.githubusercontent.com/assets/1473244/8331734/bbddbd80-1a81-11e5-94fb-6260f9501ddd.png)

edpaget:
I think the flat buttons look at lot better than ones with gradients or shadows.

mrniaboc:
All for flat design

alexbfree:
OK but even if we don't add shadows, how about at least giving it some colour? white on white is weird... and i think the button border is too thick.

DarrenMcRoy:
Assigning to @heathv to see if we can switch from the blue banner to the dimmed one?

DarrenMcRoy:
I'm still very much hoping that we can change the banner...

chrissnyder:
Not now, no.

chrissnyder:
Antiquated, as resource card design has already changed a bit.

I'm biased against these sort of meta design issues because I don't see an obvious path to close this issue. I also think we need to be careful about not designing by committee, which this issue is dangerously close to being.

mrniaboc:
I agree. The only reason I felt this case was an exception was because it
has potential to confuse visitors to the site.

Issue #662, opened 2015-06-22
Adding project descriptions to the Zooniverse panoptes home page?


mschwamb:
low priority -  On the new home page listing projects, you can't really tell what each project is about.  You have some inking for the name and the picture, but I liked the project descriptions from  old Zooniverse home tied with the picture and the project name. 

It appeared earlier in a version of the front page cards that there is a space for adding a project  description. I think adding those back in for launched projects would be really helpful for new volunteers pursuing.

DarrenMcRoy:
Is there some reason that we don't have descriptions?

I agree that it seems fairly important. Otherwise it's just a blind click on a project without knowing anything about it.

vrooje:
Yeah, I think this is probably a reasonably high priority among those issues that aren't bugs.

DarrenMcRoy:
Marking this as first-release because it seems very important to me... ?

mschwamb:
If it's helpful to know (though you might already know) - a few of the project descriptions are getting cutoff see Disk Detective, Orchid Observers, Floating Forests ,Asteroid Zoo in Firefox on a 13 inch screen

DarrenMcRoy:
I'm about to start a new issue for styling the descriptions better. :-)

PR #743, opened 2015-06-24
Updated styling for resource cards

srallen:
- Made project card avatars square for #728
- Added descriptions and removed owners for #662 
- Updated button styling for #688 
- Styled pagination buttons for #703

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -60,7 +60,7 @@ module.exports = React.createClass
                 {if meta = ownedResources[0].getMeta()
                   <nav className="pagination">
                     {for page in [1..meta.page_count]
-                      <Link to={@props.linkTo} query={{page}} key={page} className="pill-button">{page}</Link>}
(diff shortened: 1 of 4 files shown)


alexbfree:
Yay :-) Nice one @srallen !

edpaget:
We should consider keeping the owner name in the box. Projects are only unique per owner. While not a big deal as long as we're approving projects. It's totally possible for there to be a 'Galaxy Zoo' by Random Person and a 'Galaxy Zoo' by 'The Galaxy Zoo Team' and a 'Galaxy Zoo' by 'Zooniverse'.

DarrenMcRoy:
Again, my concern is that we do not want to create a delineation between "projects created in-house" and "projects created by users," especially not one that makes in-house projects look more valid than project-builder projects. The two should be as indistinguishable as possible on the projects page.

The only projects that will appear here are ones that have been approved for launch, so there's no need to worry about two having the same name.

Additionally, we've been worried about the cards getting too tall; removing one line helps at least a little with that.

We could discuss again later, but I think that the information in that box should be just what a user needs to make a decision whether or not they are interested in that project, and the username of the creator shouldn't influence that.

edpaget:
I'd argue who produced a project is a pretty important piece of information to have when deciding whether to participate in a project, but again, as long we're approving them it doesn't matter for now.

PR #751, opened 2015-06-24
Disable live workflow editing

### LINKED ISSUE #673: Can still edit live projects....
I have a live project and I just edited the workflow on it. I didn't receive any warning about editing. And when I tried the workflow, the changes had saved.

brian-c:
Here's a quick hack for disabling workflow editors on live projects.

We should go back later and disable these controls for real, or replace the editor with something else.

Closes #673.

@edpaget I think this should be enforced on the back end too.


edpaget:
I think cam has an issue on the backend for this too.

PR #766, opened 2015-06-25
Clean talk project list

### LINKED ISSUE #762: Talk sidebar needs purging
the Talk sidebar on the "Talk" page from Zooniverse home has a list of random projects in it;
this needs to be purged and updated with all relevant projects.

chrissnyder:
- Closes #762

### DIFF
--- a/app/talk/lib/project-linker.cjsx
+++ b/app/talk/lib/project-linker.cjsx
@@ -25,9 +25,11 @@ module?.exports = React.createClass
         selectedProject = project
         break
 
-    @transitionTo 'project-talk', 
-      owner: project.links.owner.login
-      name: project.slug
+    apiClient.type('users').get(project.links.owner.id)
+      .then (user) =>
+        @transitionTo 'project-talk', 
+          owner: user.login
+          name: project.slug
 
   setProjects: (metadata) ->
     query =
--- a/app/talk/lib/project-linker.cjsx
+++ b/app/talk/lib/project-linker.cjsx
@@ -25,7 +25,11 @@ module?.exports = React.createClass
           name: project.slug
 
   setProjects: (metadata) ->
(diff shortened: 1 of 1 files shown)


aweiksnar:
Looks like a better query for the sidebar list, but #goToProjectTalk isn't changing projects for me on this branch. Is it changing projects for you?

chrissnyder:
Yea, no idea why it was working for me earlier. `login` definitely doesn't exist for included `owner` within a   project resource. Had to add the call back in.

edpaget:
@chrissnyder when you include the resource it's not saved at the links property. Doing `project = apiClient.type('projects').get('1', include: 'owner')` will cache the owner so a subsequent `project.get 'owner'` won't hit the API and will return a promise with the previously included owner resource.

app/talk/lib/project-linker.cjsx
```diff
@@ -18,21 +18,31 @@ module?.exports = React.createClass
     nextState.projects isnt @state.projects
 
   goToProjectTalk: (projectId) ->
-    apiClient.type('projects').get(projectId.toString()).then (project) =>
-      project.get('owner').then (owner) =>
+    selectedProject = null
+
+    for project in @state.projects
+      if project.id is projectId
+        selectedProject = project
```
aweiksnar: I think selectedProject is set but never used, unless I'm reading this wrong

chrissnyder:
@edpaget That would do it. Query adjusted.

vrooje:
Will this also sort the projects in the sidebar by something obvious like project title? Currently it seems random.

chrissnyder:
Still random. The front end doesn't do any sorting; it just accepts however the API sends it.

edpaget:
They're sorted by most recently updated.

PR #775, opened 2015-06-25
Ensure dialogs are always above any other content

### LINKED ISSUE #753: can't close the register window/ it get blocks when going to the tmp privacy policy on firefox
Using 38.0.5 on Mac

I can't hit the x because my mouse thinks it's on the project builder link
screen shot 2015-06-25 at 7 04 09 am

then when you click the privacy link  part of the window is cut off 

screen shot 2015-06-25 at 7 04 14 am

### LINKED ISSUE #773: Can't close Help pop-up if scrollbar used to scroll to top of pop-up
On [GZ Bars](http://preview.zooniverse.org/panoptes/#/projects/vrooje/galaxy-zoo-bar-lengths/classify) I'm finding that I can replicate an issue with not being able to close the help text pop-up. If I scroll to the bottom of the help and then all the way back up, the "X" to close stops working, and if I click there I end up clicking on whatever is under that button (project builder in my case). The link loads under the pop-up, which doesn't close. So I end up with a screen that looks like:

screen shot 2015-06-25 at 12 23 24

The only way I can get the pop-up to go away is to refresh the page, which means I lose the classification I had been working on (the same subject reloads, but I have to start over).

chrissnyder:
- Closes #773
- Closes #753

### DIFF
--- a/css/dialog.styl
+++ b/css/dialog.styl
@@ -1,3 +1,7 @@
+.dialog-container
+  position: relative
+  z-index: 1
+
 .dialog-underlay
   background: rgba(black, 0.6)
   color: white
--- a/css/layout.styl
+++ b/css/layout.styl
@@ -1,10 +1,6 @@
 body
   margin: 0
 
-.panoptes-main-container
-  position: relative
-  z-index: 0
-
 .panoptes-main
   display: flex
   flex: 1 0 auto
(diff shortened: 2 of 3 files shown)


brian-c:
Sorry, that fix didn't work for me. The thing I just committed does. Want to review and merge?

chrissnyder:
:+1:

PR #776, opened 2015-06-25
Clearer owned card list count results display

### LINKED ISSUE #703: Improve pagination and navigation in project list

### LINKED ISSUE #773: Can't close Help pop-up if scrollbar used to scroll to top of pop-up

chrissnyder:
- Closes #773
- Closes #703

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -43,9 +43,14 @@ module.exports = React.createClass
       <section className="resources-container">
         <PromiseRenderer promise={@props.listPromise}>{(ownedResources) =>
           if ownedResources?.length > 0
+            meta = ownedResources[0].getMeta()
             <div>
               <div className="resource-results-counter">
-                <Translate count={ownedResources.length} content="#{@props.translationObjectName}.countMessage" component="p" />
+                {if meta
+                  pageStart = meta.page * meta.page_size - meta.page_size + 1
+                  pageEnd = Math.min(meta.page * meta.page_size, meta.count)
+                  count = meta.count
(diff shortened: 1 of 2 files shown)


aweiksnar:
:shipit: :shipit: :shipit: :shipit: :shipit:

Issue #698, opened 2015-06-24
Add a presence for Daily Zooniverse - and the zooniverse blog(s)


alexbfree:
Maybe both in top bar, and on the landing page and/or project page.

As a reminder, here's what we did in the old site:
![Daily Zooniverse flash](https://cloud.githubusercontent.com/assets/1473244/8332021/72de5a2a-1a83-11e5-81d8-6fd6ab9b676e.png)

the text is a link

mschwamb:
What about the Zooniverse blog as well ?

alexbfree:
Yes we should link to http://blog.zooniverse.org/
We should add a list of blogs for the projects too e.g. 
http://blog.oldweather.org/
http://blog.snapshotserengeti.org/
etc
(currently these are not so obvious from old zoo home - something we can improve upon)

aliburchard:
I'd vote for a daily zoo presence that is prettier than the old banner (sorry @mrniaboc)

mrniaboc:
The DZ should be linked somewhere, yes!

@aliburchard Don't apologise to me, apologise to @heathv!

heathv:
Hah, it *is* a bit goofy. But yes, I will get thinking on a way to integrate blog items and updates into the home page.

chrislintott:
What about a top level link from the main page that just links to daily.zooniverse.org or to blog.zooniverse.org?

heathv:
Sounds good to me for the short term. I'll still plan for a dedicated area on the home page.

edpaget:
Closed by #785

PR #785, opened 2015-06-25
Added Daily and Blog links

srallen:
- Added daily zoo and blog links to header and footer for #698


chrissnyder:
This pretty much maxes our horizontal space for links. Also starts looking weird at < 1100px width. We don't really support mobile that great now anyway, so I'll merge regardless, but it is something to come back to.

DarrenMcRoy:
In the long run, we probably want tabbed/nested browsing along that top header, no? I don't know the technical term... drop-downs under headers so that all the links are easily accessible.

Not for right now, anyway. :-)

PR #786, opened 2015-06-25
Builder lander copy

srallen:
This was my bad with merging #784, but these are fixes for the markdown styling.


chrissnyder:
:+1:

PR #820, opened 2015-06-26
Truncate first task select label

### LINKED ISSUE #816: very long workflow text overlaps the right-side options
The bars project has "Default" text before each question, so the workflow text is really long. This leads to this UI behavior:

screen shot 2015-06-26 at 02 10 20

(The question showing isn't the longest question in the workflow.)

I don't think the answer is to truncate the question text in the drop-down menu, as the "Default" text means all my questions start with the same thing. Can there be line breaks in each option?

chrissnyder:
- Closes #816

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -14,6 +14,9 @@ DEMO_SUBJECT_SET_ID = if process.env.NODE_ENV is 'production'
 else
   '1166' # Ghosts
 
+# This is mostly arbitrary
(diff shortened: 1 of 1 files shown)


vrooje:
Without implementing something like what's suggested in #821 I'm worried this might end up making all tasks in the drop-down list look identical. 

Maybe it could instead be 

    slice(0, half_maxlength) + ' ... ' + slice(full_length-half_maxlength, full_length-1) 

or similar?

chrissnyder:
Ah that's fair enough. I'll close this for the moment.

Issue #668, opened 2015-06-23
ZooTalk briefly flashes "no boards" before loading


DarrenMcRoy:
If you go from Zooniverse Home (http://preview.zooniverse.org/panoptes/#/) to Talk, you see a brief flash of "there are no boards" before all the boards load.

Anything we can do about that? It seems ugly and briefly disorienting?

aweiksnar:
Yep I can fix this

PR #848, opened 2015-06-26
Improve board loading state

aweiksnar:
- It currently displays `There are currently no discussions in this board.` while loading, this makes it display the loading indicator instead of that
- Same as #668 but different page


app/talk/board.cjsx
```diff
@@ -41,7 +43,7 @@ module?.exports = React.createClass
     @discussionsRequest(page)
```
brian-c: Should we set `loading` to true before this request goes out?

aweiksnar: Yeah good idea just udpated

PR #849, opened 2015-06-26
Changed transitionTo to replaceWith

### LINKED ISSUE #837: Clicking your own name goes to your inbox, and the Back button breaks
It looks like clicking someone's name in Talk currently takes them to the page where you can send that person a DM.

Clicking on your own name, however, goes to your inbox, and then the Back button on your browser can't get you out of that inbox.

I assume this will eventually change once profiles are more fleshed-out, but for now, can we at least make the back button get you back out?

srallen:
Fixed #837

### DIFF
--- a/app/pages/profile/private-message.cjsx
+++ b/app/pages/profile/private-message.cjsx
@@ -22,7 +22,7 @@ module.exports = React.createClass
     }</PromiseRenderer>
 
   redirectToMessages: ->
-    @transitionTo 'inbox'
+    @replaceWith 'inbox'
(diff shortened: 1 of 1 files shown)


brian-c:
Good change, but the whole profile page is confusing. I can no longer view my own profile?

srallen:
Yeah I agree, but that's temporary until we have something to put on it, stats or active feed content, so for now users can at least still send messages to each other.

brian-c:
Sounds good to me.

PR #855, opened 2015-06-26
Convert urls to links in md, and add breaks

### LINKED ISSUE #853: If you write out a url in Talk it doesn't become a hyperlink
lower in the queue -  Posting this before I get about it , since i saw it. Just noticed that one of the P4T moderators tried to post a link to a P4 Talk thread and it doesn't show up as a url. I know there's a good fraction of people write the full address and don't use the linking. In current Talk is identified those are links and makes them by default hyperlinks, I think  this would be handy if in the future panoptes Talk could do this as well 

http://preview.zooniverse.org/panoptes/#/projects/mschwamb/planet-four-terrains/talk/8/118?page=1&comment=216

aweiksnar:
- Closes #853

### DIFF
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -16,7 +16,7 @@ POSSIBLE_ROLES = [
 ROLES_INFO =
   owner:
     label: "Owner"
-    description: "The colleciton creator. There can only be one."
+    description: "The collection creator. There can only be one."
   collaborator:
     label: "Collaborator"
     description: "Collaborators have full access to add and remove subjects from the collection."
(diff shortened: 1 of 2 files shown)


brian-c:
The hashtag replacement breaks links with #fragments.

Might want to update the regexes to check for a `\s` or `^` before matching.

aweiksnar:
Ooh good thinking, just pushed a fix for that case

brian-c:
Sorry this is taking so long. Want to do the same thing with `@`? Email addresses are caught too.

aweiksnar:
I didn't realize linkify does emails too, sorry about not checking that. Just updated in 3de94e9

brian-c:
Sweet, thanks.

PR #865, opened 2015-06-28
Collections bugfixes

### LINKED ISSUE #827: Auto-generated collection name funky, can't change
I've favourited 1 image so far, but when I went to check out my collections, it pops up as "favorites (6)" (and /favorites-6 in the url). I can open the collection settings and click in the name box to change the name, but can't actually edit the name.

Even if the 6 referred to the number of images in my favorites collection (it doesn't), it's a bit weird in the name of the collection itself.
screen shot 2015-06-26 at 08 17 30

### LINKED ISSUE #828: Should collections on creation be default public ?  - currently set to private by default and have to uncheck to make public
I think by default we would want to make collection creation have public visibility so that volunteers and science teams can see them and peruse them like the previous way collections worked. I think the added feature of hiding a collection should be an opt in where you have to check the box

I think will tend to leave the default, and there will be more private than public collections potentially.

screen shot 2015-06-26 at 5 08 45 pm

### LINKED ISSUE #832: if you don't enter a collection name subjects get collected in a broken unnamed collection
The unnamed collection I can access when I click it, it takes my back to the collections table of contents page 
screen shot 2015-06-26 at 5 39 10 pm

### LINKED ISSUE #834: non-logged in can't access public collection get error
Get error TypeError: l is null when I click on http://preview.zooniverse.org/panoptes/#/collections/mschwamb/holey-swiss-cheese-terrain - the collection I made public in http://preview.zooniverse.org/panoptes/#/collections

### LINKED ISSUE #852: No way to delete a collection once created
Once I create a collection, if I change my mind about it I can't delete it. I saw nothing on the settings tab for deletion

### LINKED ISSUE #857: can't remove a collaborator from a collection
I added a collaborator to a collection successfully, but can't remove them. The x doesn't do anything

chrissnyder:
- Closes #857
- Closes #852
- Closes #834
- Closes #832
- Closes #828 
- Closes #827

### DIFF
--- a/app/components/user-search.cjsx
+++ b/app/components/user-search.cjsx
@@ -14,7 +14,7 @@ module.exports = React.createClass
       apiClient.type('users').get display_name: "#{value}*", page_size: 10
         .then (users) =>
           opts = for user in users
-            { value: user.id, label: "@#{ user.login }" }
+            { value: user.id, label: "@#{ user.login }: #{ user.display_name }" }
           callback null, {
             options: opts
           }
--- a/app/components/user-search.cjsx
+++ b/app/components/user-search.cjsx
@@ -14,7 +14,7 @@ module.exports = React.createClass
       apiClient.type('users').get display_name: "#{value}*", page_size: 10
         .then (users) =>
           opts = for user in users
-            { value: user.id, label: "@#{ user.login }: #{ user.display_name }" }
+            { value: user.id, label: "@#{ user.login }" }
           callback null, {
             options: opts
           }
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -1,4 +1,6 @@
 React = require 'react'
+counterpart = require 'counterpart'
+ChangeListener = require '../components/change-listener'
 PromiseRenderer = require '../components/promise-renderer'
 UserSearch = require '../components/user-search'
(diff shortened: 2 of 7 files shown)


app/components/user-search.cjsx
```diff
@@ -14,7 +14,7 @@ module.exports = React.createClass
       apiClient.type('users').get display_name: "#{value}*", page_size: 10
         .then (users) =>
           opts = for user in users
-            { value: user.id, label: "@#{ user.login }: #{ user.display_name }" }
+            { value: user.id, label: "@#{ user.login }" }
```
edpaget: This has still got to have the the display_name since that's what's actually being searched.

chrissnyder: Ah ok. I removed it because it looked a bit silly as everyone's login and display_name were identical.

Given the other methods of referencing users use login (tagging, mentions), shouldn't the search do so as well?

edpaget: See zooniverse/Panoptes#1034

edpaget: The display_name won't have changed from what it was previously so it'll probably be the easiest way to find people for now.

chrissnyder: Roger that, I'll add the display_name back into the label. Also cool to see search improvement in the pipeline.

PR #871, opened 2015-06-29
Add talk sticky discussion ordering

parrish:
This adds a toggle and ordering for talk discussion stickies.


app/talk/lib/moderation.cjsx
```diff
@@ -47,8 +47,13 @@ module?.exports = React.createClass
                 <button onClick={@toggleModeration}>
                   <i className="fa fa-#{if @state.open then 'close' else 'warning'}" /> Moderator Controls
                 </button>
-                <div className="talk-moderation-children #{if @state.open then 'open' else 'closed'}">
```
aweiksnar: Wow looks great! If you can add this `#{if @state.open then 'open' else 'closed'}">` line (used for the form css) back in looks good to merge.

parrish: Done.  Thanks!

PR #876, opened 2015-06-29
Link board preview user to latest commenter

### LINKED ISSUE #873: Most recent post link doesn't go to the most recent post
The front page of Talk, when showing recent threads, displays not the name of the most recent person to post, but instead the name of the person who started the thread.

screen shot 2015-06-29 at 1 09 01 pm

when you can see that Chris' post was actually from a long time back, and the most recent posts were mine and Janet's

screen shot 2015-06-29 at 1 10 37 pm

aweiksnar:
- Closes #873

### DIFF
--- a/app/talk/board-preview.cjsx
+++ b/app/talk/board-preview.cjsx
@@ -31,16 +31,21 @@ module?.exports = React.createClass
 
         <PromiseRenderer promise={talkClient.type('discussions').get({board_id: @props.data.id}).index(0)}>{(discussion) =>
           if discussion?
-            <div className="talk-discussion-link">
-              <PromiseRenderer promise={apiClient.type('users').get(discussion.user_id.toString())}>{(user) =>
-                <Link className="user-profile-link" to="user-profile" params={name: discussion.user_login}>
-                  <Avatar user={user} />{' '}{discussion.user_display_name}
-                </Link>
-              }</PromiseRenderer>{' '}
+            comments = discussion.links.comments
+            lastCommentId = comments[comments.length-1]
 
-              <Link to="#{@projectPrefix()}talk-discussion" params={merge({}, {board: discussion.board_id, discussion: discussion.id}, @props.params)}>{discussion.title}</Link>{' '}
-              <span>{timeAgo(discussion.updated_at)}</span>
-            </div>
+            <PromiseRenderer promise={talkClient.type('comments').get(lastCommentId)}>{(comment) =>
+              <div className="talk-discussion-link">
+                <PromiseRenderer promise={apiClient.type('users').get(comment.user_id.toString())}>{(user) =>
+                  <Link className="user-profile-link" to="user-profile" params={name: user.login}>
+                    <Avatar user={user} />{' '}{user.display_name}
+                  </Link>
+                }</PromiseRenderer>{' '}
+
+                <Link to="#{@projectPrefix()}talk-discussion" params={merge({}, {board: discussion.board_id, discussion: discussion.id}, @props.params)}>{discussion.title}</Link>{' '}
+                <span>{timeAgo(discussion.updated_at)}</span>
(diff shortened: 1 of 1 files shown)


app/talk/board-preview.cjsx
```diff
@@ -31,16 +31,21 @@ module?.exports = React.createClass
...
-                </Link>
-              }</PromiseRenderer>{' '}
-
-              <Link to="#{@projectPrefix()}talk-discussion" params={merge({}, {board: discussion.board_id, discussion: discussion.id}, @props.params)}>{discussion.title}</Link>{' '}
-              <span>{timeAgo(discussion.updated_at)}</span>
-            </div>
+            comments = discussion.links.comments
+            lastCommentId = comments[comments.length-1]
+
+            <PromiseRenderer promise={talkClient.type('comments').get({id: lastCommentId}).index(0)}>{(comment) =>
```
brian-c: Why not `talkClient.type('comments').get(lastCommentId)`?

app/talk/board-preview.cjsx
```diff
@@ -31,16 +31,21 @@ module?.exports = React.createClass
...
-              <PromiseRenderer promise={apiClient.type('users').get(discussion.user_id.toString())}>{(user) =>
-                <Link className="user-profile-link" to="user-profile" params={name: discussion.user_login}>
-                  <Avatar user={user} />{' '}{discussion.user_display_name}
-                </Link>
-              }</PromiseRenderer>{' '}
-
-              <Link to="#{@projectPrefix()}talk-discussion" params={merge({}, {board: discussion.board_id, discussion: discussion.id}, @props.params)}>{discussion.title}</Link>{' '}
-              <span>{timeAgo(discussion.updated_at)}</span>
-            </div>
+            comments = discussion.links.comments
```
brian-c: Any way we can get the last comment without accessing `links` directly? No big deal if we can't, but `links` can be expressed lots of different ways in JSON API, I hate depending on our implementation to never change.

app/talk/board-preview.cjsx
aweiksnar: Just updated to used the shorter get syntax

app/talk/board-preview.cjsx
aweiksnar: Using json-api-client links for this situation throws `Error: Value for 'discussions.id' in '/comments?discussion_id={discussions.id}' should be a string.` I felt like it was cleaner to leave it as `discussion.links.comments`

brian-c: God, I just now realized that Talk IDs aren't strings and that's why you've been `.toString()`ing all over the place.

@parrish JSON-API says `id` is always a string. Can we make these strings?

parrish:
We can, but that's a really stupid change in the spec, FWIW.

brian-c:
Must've been like that since we started using it or I wouldn't've written it like that.

parrish:
Yeah?  Restpack never serialized id's to strings.  @edpaget Did you patch something into restpack to do that?

edpaget:
Nope it's always done that.

parrish:
Right.  I'm pretty confused why I've never had that behavior from it.

edpaget:
Are we talking about the "id" field or the "user_id" fields and such? Panoptes has never included the second since that's just in the links anway.

parrish:
All of the above.

edpaget:
Huh I have no idea then.

app/talk/board-preview.cjsx
aweiksnar: `id` issues aside, any reason not to deploy this?

parrish:
By the way zooniverse/Talk-Api@8a82b80fe1 will be changing the id's to strings.  It'll be on staging when the build is ready

PR #880, opened 2015-06-29
Fix replies: ensure all lines in reply are quoted

### LINKED ISSUE #878: Replying to a post on Talk that's more than one line, only the first line gets in the quote format others look like part fo your text
screen shot 2015-06-30 at 3 16 29 am

aweiksnar:
- Closes #878


chrissnyder:
LGTM

PR #884, opened 2015-06-29
Admin mode

brian-c:
This adds an "Admin mode" checkbox to the footer when a user has `admin: true` set. While that's checked, the client will send `admin: true` as a param with every request, like:

PUT things:
```json
{
  "admin": true,
  "things": {...}
}
```


edpaget:
:smiling_imp:

Issue #738, opened 2015-06-24
Show usernames of users in conversation in inbox messages & previews


aweiksnar:
& timestamps

PR #893, opened 2015-06-30
Inbox private message form

### LINKED ISSUE #748: When your inbox is empty, it should help you figure out how to send messages.
Maybe have a sentence explaining how to send a message, or what messages are. Or some feature to find people to message....?

inbox

aweiksnar:
- Adds a form to send new private messages to the inbox page
- Closes #748
- Using user search for selection, so related to https://github.com/zooniverse/Panoptes/issues/1034

### DIFF
--- a/app/talk/inbox-form.cjsx
+++ b/app/talk/inbox-form.cjsx
@@ -1,0 +1,64 @@
+React = {findDOMNode} = require 'react'
+talkClient = require '../api/talk'
+UserSearch = require '../components/user-search'
+{Navigation} = require 'react-router'
+{getErrors} = require './lib/validations'
+subjectValidations = require './lib/message-subject-validations'
+messageValidations = require './lib/message-validations'
+
+module?.exports = React.createClass
+  displayName: 'InboxForm'
+  mixins: [Navigation]
+
+  propTypes:
+    user: React.PropTypes.object
+
+  getInitialState: ->
+    validationErrors: []
+
+  validations: (message, subject, recipients) ->
+    userErrors = if (not recipients.length) then ['Messages must have a recipient'] else []
+    subjectErrors = getErrors(subject, subjectValidations)
+    messageValidationErrors = getErrors(message, messageValidations)
+
+    validationErrors = userErrors.concat subjectErrors.concat messageValidationErrors
+    @setState {validationErrors}
+    !!validationErrors.length
+
+  onSubmitMessage: (e) ->
+    e.preventDefault()
+    form = findDOMNode(@refs.form)
+
+    recipient_ids = form.querySelector('[name="userids"]')
+      .value.split(',').map (id) -> parseInt(id)
+      .filter(Number)
+
+    title = findDOMNode(@refs.subject).value
+    body = findDOMNode(@refs.message).value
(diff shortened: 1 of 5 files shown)


aweiksnar:
If you think it's worth waiting until user search is more fleshed out I'm cool with that, but figured I'd put this out there.

chrissnyder:
Might be. Playing with it a bit now and it's difficult to separate out the two problems. Will look more after lunch.

app/talk/inbox-form.cjsx
```diff
@@ -0,0 +1,64 @@
...
+    subjectErrors = getErrors(subject, subjectValidations)
+    messageValidationErrors = getErrors(message, messageValidations)
+
+    validationErrors = userErrors.concat subjectErrors.concat messageValidationErrors
+    @setState {validationErrors}
+    !!validationErrors.length
+
+  onSubmitMessage: (e) ->
+    e.preventDefault()
+    form = findDOMNode(@).querySelector('form')
```
chrissnyder: Any reason you use this style over refs?

app/talk/inbox-form.cjsx
```diff
@@ -0,0 +1,64 @@
...
+
+    recipient_ids = form.querySelector('[name="userids"]')
+      .value.split(',').map (id) -> parseInt(id)
+      .filter(Number)
+
```
chrissnyder: Given that we are limiting recipients to one person, couldn't this just be `form.querySelector('[name="userids"]').value`?

app/talk/inbox-form.cjsx
```diff
@@ -0,0 +1,64 @@
...
+    title = findDOMNode(@refs.subject).value
+    body = form.querySelector('textarea').value
+    user_id = @props.user.id
+
+    conversation = {title, body, user_id, recipient_ids}
+
+    errors = @validations(body, title, recipient_ids)
+    return errors if errors.length
```
chrissnyder: This should just be `return errors if errors`

chrissnyder:
Separating out the validations is pretty nice. For future plans here, perhaps a `<ValidatedInput>` component would be useful so each form field handles it's own validations?

app/talk/inbox-form.cjsx
aweiksnar: Ah yeah (but it'd have to be an array). There could be multiple recipients, but I wanted to limit to one until I get to #738

chrissnyder: My bad then, didn't realize a message could have multiple recipients.

app/talk/inbox-form.cjsx
aweiksnar: You're right, refs would be cleaner here. I'll update in a sec

app/talk/inbox-form.cjsx
aweiksnar: Good call, I originally had validations return an array but forgot to update

aweiksnar:
Good idea on that validated input component, I'll look to add that soon. I just updated the commit and think I hit all of your points

chrissnyder:
:+1:

PR #900, opened 2015-06-30
Preserve the @symbol in username mentions

aweiksnar:


chrissnyder:
Mentions work either way. I assume this is just to make it more explicit that you are mentioning someone?

edpaget:
Big :+1: from me. Makes the process more discoverable.

aweiksnar:
Yeah I guess so, kinda following github's lead on this

chrissnyder:
Roger that, merging!

PR #901, opened 2015-06-30
Preserve current page when editing comments

### LINKED ISSUE #883: Editing a post on page 2 of a thread takes you back to page 1
If you post on the second page of a thread (or any other page, I assume), and then edit your post, completing the edit takes you back to page 1... this probably shouldn't happen?

aweiksnar:
- Closes #883


chrissnyder:
:+1:

PR #909, opened 2015-06-30
List popular tags in sidebar

### LINKED ISSUE #831: hashtag shows up on the Talk side bar as many times as it is used
perhaps low in the queue, but I can see it being confusing at launch - if a hashtag is used multiple times it shows up multiples times on the side bar

screen shot 2015-06-26 at 6 01 09 pm

aweiksnar:
- Closes #831
- Adds PopularTags component

### DIFF
--- a/app/talk/popular-tags.cjsx
+++ b/app/talk/popular-tags.cjsx
@@ -1,0 +1,33 @@
+React = require 'react'
+PromiseRenderer = require '../components/promise-renderer'
+{makeHTTPRequest} = require 'json-api-client'
+{Link} = require 'react-router'
+
+module?.exports = React.createClass
+  displayName: 'TalkPopularTags'
+
+  propTypes:
+    header: React.PropTypes.object             # Optional header ex: <h1>Tags</h1>
+    section: React.PropTypes.string.isRequired # Talk section
+    params: React.PropTypes.object.isRequired # URL params
+
+  tagsRequest: ->
+    talkClient.type('tags/popular').get section: @props.section, limit: 10
+
+  tag: (tag, i) ->
+    {owner, name} = @props.params
+    if owner and name
+      <span key={i}><Link params={{owner, name}} query={query: tag.name} to="project-talk-search">#{tag.name}</Link>{' '}</span>
+    else
+      <span key={i}><Link query={query: tag.name} to="talk-search">#{tag.name}</Link>{' '}</span>
+
+  render: ->
+    <div className="talk-popular-tags">
+      <PromiseRenderer promise={@tagsRequest()}>{(tags) =>
+        if tags?.length
+          <div>
+            {@props.header ? null}
+            {tags.map(@tag)}
+          </div>
+      }</PromiseRenderer>
+    </div>
--- a/app/talk/board-preview.cjsx
+++ b/app/talk/board-preview.cjsx
@@ -32,7 +32,7 @@ module?.exports = React.createClass
         <PromiseRenderer promise={talkClient.type('discussions').get({board_id: @props.data.id}).index(0)}>{(discussion) =>
           if discussion?
             <div className="talk-discussion-link">
-              <PromiseRenderer promise={apiClient.type('users').get(discussion.user_id.toString())}>{(user) =>
+              <PromiseRenderer promise={apiClient.type('users').get(discussion.user_id)}>{(user) =>
                 <Link className="user-profile-link" to="user-profile" params={name: discussion.user_login}>
                   <Avatar user={user} />{' '}{discussion.user_display_name}
                 </Link>
--- a/app/talk/comment-link.cjsx
+++ b/app/talk/comment-link.cjsx
@@ -28,7 +28,7 @@ module?.exports = React.createClass
 
   setDiscussion: ->
     {comment} = @props
-    talkClient.type('discussions').get(comment.discussion_id.toString(), {})
+    talkClient.type('discussions').get(comment.discussion_id, {})
       .then (discussion) => @setState {discussion}
 
   projectComment: ->
--- a/app/talk/comment.cjsx
+++ b/app/talk/comment.cjsx
@@ -111,7 +111,7 @@ module?.exports = React.createClass
             {if @props.data.focus_id
               <PromiseRenderer
                 promise={
-                  apiClient.type('subjects').get(@props.data.focus_id.toString())
+                  apiClient.type('subjects').get(@props.data.focus_id)
                 }
                 then={(subject) =>
                   <div className="polaroid-image">
--- a/app/talk/discussion.cjsx
+++ b/app/talk/discussion.cjsx
@@ -107,7 +107,7 @@ module?.exports = React.createClass
   onLikeComment: (commentId) ->
     user = @state.user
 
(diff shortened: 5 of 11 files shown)


app/talk/popular-tags.cjsx
```diff
@@ -0,0 +1,35 @@
...
+    params: React.PropTypes.object.isRequired # URL params
+
+  tagsRequest: ->
+    popularTags = talkClient.root + "/tags/popular?section=#{@props.section}&limit=10"
+    makeHTTPRequest 'GET', popularTags, null, talkClient.headers
+
+  tag: (tagName, i) ->
+    {owner, name} = @props.params
+    if owner and name
+      <span><Link key={i} params={{owner, name}} query={query: tagName} to="project-talk-search">#{tagName}</Link>{' '}</span>
```
brian-c: `key` attributes should be on the span, not the Link.

app/talk/popular-tags.cjsx
```diff
@@ -0,0 +1,35 @@
...
+  displayName: 'TalkPopularTags'
+
+  propTypes:
+    header: React.PropTypes.object             # Optional header ex: <h1>Tags</h1>
+    section: React.PropTypes.string.isRequired # Talk section
+    params: React.PropTypes.object.isRequired # URL params
+
+  tagsRequest: ->
+    popularTags = talkClient.root + "/tags/popular?section=#{@props.section}&limit=10"
+    makeHTTPRequest 'GET', popularTags, null, talkClient.headers
```
brian-c: Just in terms of figuring out how people are using the JSON-API lib (again, sorry), is there no other way to access this? The goal was to be able to access resources with no knowledge of the back end implementation. If the library gets in the way more than it's useful, let's stop using it. I just want this to be consistent eventually.

app/talk/popular-tags.cjsx
aweiksnar: Just updated!

app/talk/popular-tags.cjsx
brian-c: Here's a hack that I think works:

`talkClient.type('tags/popular').get section: @props.section, limit: 10`

I don't know if I feel any better about that, but at least you won't have to parse the response manually.

aweiksnar: This was the cleanest I could do. I tried `talkClient.get(...)` first, but since this endpoint returns an array of strings, the response ended up being formatted strangely (I think they were stored in an errant private key )

aweiksnar: Ah cool let me try that

brian-c: Wow, no, sorry, I didn't realized it returned an array of strings. Nevermind. This is not JSON-API. Do what you gotta do. Sorry.

brian-c: As long as I'm here: @parrish why are these strings?

aweiksnar: Not sure if that solution is ideal. Embedding within a `PromiseRenderer` yields an unresponsive script warning, and outside of it does fetch the correct data but stores the values in the `_changedKeys` array

parrish: Because that endpoint presents something that is not a resource -- it's an aggregated statistic.

brian-c: Any chance it could present resources instead? Since the rest of the API is all about resources?

parrish: I'm at a loss on it.  I could slap a fake meta hash into the response I guess.  Ultimately, it can't be interacted with in any way other than by passing section, limit, or type params.  How else could it fake a resource?

parrish: Looks like this is a fairly unaddressed issue with the spec:

https://github.com/json-api/json-api/issues/648

Since I really don't care about dogmatically following a spec, what in particular is tripping up your client?

brian-c: Tags are resources, right? I can hit /tags and get resources back. Instead of strings, return those resources.

Couldn't this be queried as something like `/tags?section=zooniverse&sort=uses&limit=10` where "uses" is whatever measure of popularity it's using?

Sorry I'm being a pain about this stuff, but I thought we decided on following JSON API so we wouldn't run into stuff like this.

edpaget: I think if you just turned the tags into objects it'd be enough for the library.

edpaget: Like ```{"name": "tag"}```, technically the only field required by the JSON-API spec we were working off is an "id" and you could put literally anything there if you want.

edpaget: If we really wanted to something more JSON-API like the URL could be `https://talk.zooniverse.org/tags?section=project-17&sort=+popular` since popularity is just a way of sorting the resources.

parrish: Here's what we're looking at with tags

id |   name | section     | user_id | comment_id | taggable_id | taggable_type
---|--------|-------------|---------|------------|-------------|--------------
1  | 'neat' | 'project-1' | 1       | 5          | 9           | 'Subject'
2  | 'cool' | 'project-1' | 1       | 6          | NULL        | NULL
3  | 'neat' | 'project-1' | 2       | 7          | 10          | 'Subject'
4  | 'cool' | 'project-2' | 1       | 8          | NULL        | NULL

If you find tags for `/tags?section=project-1` you'll get back the individual tags for project-1

If you have 10,000 of those tags, you're not going to iteratively count the uses or denormalize the counter to all of the records.

The way you do something like this is by aggregating the results in the database using something like `group by name`.  The result is that each row in the grouped result will represent an arbitrary number of actual rows.

If you need an array of tags with fake id's and NULL attributes, then I can respond with that, but it's just format consistency at that point.

parrish:
zooniverse/Talk-Api@b77e33984eded should give this a consistent format

aweiksnar:
This should be good to go. If you agree @brian-c maybe we can coordinate merging with @parrish deploying?

brian-c:
Cool, @parrish can you deploy to talk-staging and let me know?

parrish:
The new tag structure is on staging now:
``` javascript
{
  "tags": [{
    "id": null,
    "name": "good",
    "section": "zooniverse",
    "created_at": null,
    "updated_at": null,
    "user_id": null,
    "comment_id": null,
    "taggable_id": null,
    "taggable_type": null
  }, {
    "id": null,
    "name": "taggin",
    "section": "zooniverse",
    "created_at": null,
    "updated_at": null,
    "user_id": null,
    "comment_id": null,
    "taggable_id": null,
    "taggable_type": null
  }],
  "meta": {
    "tags": {
      "page": 1,
      "page_size": 10,
      "count": 2,
      "include": [

      ],
      "page_count": 1,
      "previous_page": null,
      "next_page": null,
      "first_href": "/tags/popular",
      "previous_href": null,
      "next_href": null,
      "last_href": "/tags/popular"
    }
  }
}
```

parrish:
Everything looks good on staging to me, ready to move it to production?

aweiksnar:
:+1:

brian-c:
Looks good to me.

PR #917, opened 2015-07-01
Add pagination behaviors

### LINKED ISSUE #887: Add a jump to last page icon on Talk
You have to click on the menu and select the page, it would be really nice if you could quickly jump to the end of the conversation or maybe something like is currently on Talk 2.0 - right now it's a few clicks to jump to the last page of a discussion

screen shot 2015-06-30 at 5 34 28 am

### LINKED ISSUE #906: Clicking "next" button on talk should scroll to the top of the next page
Right now when you hit the "next" button on talk, the page keeps its scroll position at the bottom of the page.  I think it would be less confusing if the scroll reset itself to the top of the page when "next" is clicked.

Tested on Mac Firefox 38.0.5 (have not tried any other browsers).

### LINKED ISSUE #912: Invisible page numbers on dark profiles (white background, white numbers)
The page numbers don't display on dark profiles. It seems you haven't set the numbers colour.

https://lh3.googleusercontent.com/-KG70pnRyr18/VZLtykB_7iI/AAAAAAAAICc/_d25tN0jF3Y/s912/ysh_2015-06-30.jpg

aweiksnar:
- 1c9bebd Closes  #887
- 982f70d Closes #906 
- 1dafd81 Closes #912

### DIFF
--- a/app/talk/lib/paginator.cjsx
+++ b/app/talk/lib/paginator.cjsx
@@ -5,11 +5,13 @@ module?.exports = React.createClass
 
   propTypes:
     pageCount: React.PropTypes.number
-    page: React.PropTypes.number # page number
+    page: React.PropTypes.number                  # page number
     onPageChange: React.PropTypes.func.isRequired # passed (page) on change
+    firstAndLast: React.PropTypes.bool            # Optional, add 'first' & 'last' buttons
 
   getDefaultProps: ->
     page: 1
+    firstAndLast: true
 
   setPage: (activePage) ->
     @props.onPageChange(activePage)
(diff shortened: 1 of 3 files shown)


brian-c:
1c9bebd looks good.
1dafd81 fixes the options but not the select itself.

Scrolling only happens when choosing from the `select` menu, `scrollTo` should probably be called from `setPage` instead of `onSelectPage`. However, shouldn't this be handled by the router automatically after a `transitionTo` call? I'd look into that first.

aweiksnar:
I've been playing with the router quite a bit and can't get that behavior to happen on any of the paginated pages, even with [ScrollToTopBehavior](http://rackt.github.io/react-router/#ScrollToTopBehavior) set. Is that behavior somthing that you have had working before?

brian-c:
Yeah, it looks like it just doesn't follow normal scroll-back-up behavior when only the query changes. That's annoying. Want to move that scrollTo call, then?

aweiksnar:
Yep just moved it over

Issue #894, opened 2015-06-30
UI weirdnesses on user Customize Profile page


rogerhutchings:
- For the avatar / background, there are no instructions on what to do other than 'Drop an image here', and the text gets hidden once an image is in there;
- There's no information on what images are okay or not (I was thinking sizes, more than content);
- There's no button to clear an existing avatar / background;
- If you upload an image, then click on the image area, then click cancel, you get an infinite loading animation (someone should try to duplicate this though)

brian-c:
A clear button can be added when zooniverse/Panoptes#1060 is resolved.

yshish:
- If you upload an image, then click on the image area, then click
   cancel, you get an infinite loading animation (someone should try to
   duplicate this though)

I had the the infinite loading animation while trying to replace the
current image with a new one.

2015-06-30 10:51 GMT+02:00 Roger Hutchings <notifications@github.com>:

chrissnyder:
Most of the original concerns were addressed, closing this is favor of a more atomic issue.

PR #936, opened 2015-07-01
Improve settings page, ImageSelector

brian-c:
This split email out the settings page.

ImageSelector components now know how to size themselves relative to their containers.

For #894:

- Added some guidance for image sizes

- Fixed rendering bug with no image selection


chrissnyder:
:+1: :+1: :+1: for the settings organization.

I couldn't test the ImageSelector as it was squished it to a pixel or two high. Unless I'm missing something on how it's expected to be used.

brian-c:
Oh boyyy, let me look. Sorry.

brian-c:
Should be solid now.

chrissnyder:
Worked like a charm.

Bit pedantic, but the file size limits don't mean too much because it will auto-shrink them for you (which is absolutely the desired behavior in my opinion).

Other than that, looks great.

PR #942, opened 2015-07-02
Remove owner and owner role options from collaborators page

### LINKED ISSUE #920: Can't remove user roles from project visibility
Unchecking a user role doesn't remove the actual role from the backend (reload page it's still there). Looks like no request if firing on the uncheck event. Related to autosave?

chrissnyder:
- Closes #920 
- I believe the root cause of this is adding additional roles to the project owner, which causes confusion as to what role should take precedence. So for the FE, I removed the row to toggle different roles on the owner.
- I also removed the owner checkbox on collaborators because it was disabled anyway.

### DIFF
--- a/app/pages/lab/collaborators.cjsx
+++ b/app/pages/lab/collaborators.cjsx
@@ -1,6 +1,7 @@
 React = require 'react'
 PromiseRenderer = require '../../components/promise-renderer'
 UserSearch = require '../../components/user-search'
+authClient = require '../../api/auth'
 apiClient = require '../../api/client'
 talkClient = require '../../api/talk'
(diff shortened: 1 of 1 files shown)


camallen:
LGTM - i tested locally and was able to remove roles this time. Does point 3 above mean i won't see who the project owner is?

chrissnyder:
Ah, that's a fair point. Didn't think about it from the perspective of a collaborator. Will rework slightly and reopen when ready.

chrissnyder:
- Added message at the top stating who the project owner is.

camallen:
@chrissnyder LGTM - tested locally and it worked well.

PR #949, opened 2015-07-02
View workflow

### LINKED ISSUE #253: "View/Preview" tab from the workflow editor
As I'm creating tasks, I'm constantly trying to see what those changes look like and how they affect the workflow. Right now, I keep having to do this by going up to "the lab" link, then selecting my current project from my list of projects and then clicking View. And then going back to "the lab" for that particular process. Could we add a tab that opens the "View Project" link directly from the editor? I realize you'll still need to save changes to a task to have them display, but I think that's fine.

brian-c:
Add a "view" link from the workflow editor to a new window.

I was adding more "view" links around the other project editor pages, but then I realized that they don't reload, so the second page will see old data. That's why there's a `reload=0` param. So for now, there's just a link from the workflow editor. It might be possible to generalize this in the future.

Closes #253.

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -216,7 +216,6 @@ EditWorkflowPage = React.createClass
       @transitionTo 'edit-project-details', projectID: @props.project.id
 
   handleTaskChange: (taskKey, path, value) ->
-    console?.log 'Handling task change', arguments...
     changes = {}
     changes["tasks.#{taskKey}.#{path}"] = value
     @props.workflow.update changes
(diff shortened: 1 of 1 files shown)


aweiksnar:
Cool! Looks like it works well. What issue does the `pointerEvents: 'all'` div address? I'm able to follow the view link both with & without it

brian-c:
Ah, that's just so I could press it on live projects where the workflow editor is disabled. Soon only parts of the editor will be disabled and we can get rid of it.

aweiksnar:
Ok cool, merging

PR #953, opened 2015-07-02
show all projects when admin user mode is active

camallen:
does what it says on the tin, allow users in admin mode to view all projects.

### DIFF
--- a/app/pages/projects.cjsx
+++ b/app/pages/projects.cjsx
@@ -20,8 +20,9 @@ module.exports = React.createClass
 
   listProjects: ->
     query =
-      launch_approved: true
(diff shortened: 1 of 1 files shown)


app/pages/projects.cjsx
```diff
@@ -20,8 +20,9 @@ module.exports = React.createClass
 
   listProjects: ->
     query =
-      launch_approved: true
       include:'owners,avatar'
+    if !apiClient.params.admin
+      query[launch_approved] = true
```
brian-c: `query.launch_approved`, no?

camallen:
updated

PR #964, opened 2015-07-03
unsubscribe all zoo emails via html form

camallen:
html form based email unsubscription for users that can't login. We may end up pulling this and just use the API GET with uniq token links but for now this will save some effort.

### DIFF
--- a/app/api/auth.coffee
+++ b/app/api/auth.coffee
@@ -251,6 +251,13 @@ module.exports = new Model
       else
         throw new Error 'Failed to sign out; not signed in'
 
+  unsubscribeEmail: ({email}) ->
+    @_getAuthToken().then (token) =>
+      data =
+        authenticity_token: token
+        email: email
+
+      client.post '/../unsubscribe', data, JSON_HEADERS
 # For quick debugging:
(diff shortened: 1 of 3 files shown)


chrissnyder:
Two questions:

- I'm only getting 404s when trying to unsubscribe. I'm assuming that's related to zooniverse/Panoptes#1068?
- Doesn't seem to matter if I'm logged in or not. The backend checks if I actually have permission to unsubscribe via email, right?

camallen:
1. that post route isn't in the backend yet
2. It's for non-logged in users mostly but it doesn't really matter, backend will unsubscribe from all emails any known email address.

chrissnyder:
Ok, was just checking that I couldn't put your email in and unsubscribe you if I was so inclined.

Code here looks fine, so merging in.

camallen:
Ah - sorry, it actually will work that way. This is a replica of the old zoo home form till the per user unsubscribe links get into emails, etc.

chrissnyder:
Just putting it on record that that seems dangerous, but if that exact functionality existed previously, then I guess it isn't any more harmful than before.

PR #991, opened 2015-07-07
Record subject dimensions

### LINKED ISSUE #986: Record subject image and view port dims on classification
Linked to https://github.com/zooniverse/Panoptes/issues/1072#issuecomment-118862047

It'd be really useful to understand what a user actually saw when they were classifying in relation to the real dimensions of subjects images.

brian-c:
This adds some metadata to classifications:

- The viewport size (on classification submission) as `metadata.viewport.width` and `.height`

- Each subject frame's actual size as `metadata.subject_dimensions.(frame index).naturalWidth` and `.naturalHeight`, and displayed size as `.clientWidth` and `.clientHeight`. This is recorded after viewing the frame; if the user never actually loads a frame of a multi-`locations` image, no dimensions are recorded for it.

Closes #986.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -133,7 +133,7 @@ Classifier = React.createClass
     {naturalWidth, naturalHeight, clientWidth, clientHeight} = e.target
 
     changes = {}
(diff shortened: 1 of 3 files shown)


camallen:
Looks good.

PR #996, opened 2015-07-07
Adds active users to talk.  Towards notification integration

parrish:


aweiksnar:
@chrissnyder and I did some testing and everything worked great :+1: Going to leave a couple line-notes

app/talk/active-users.cjsx
```diff
@@ -0,0 +1,98 @@
...
+    </li>
+
+  render: ->
+    <div className="talk-active-users">
+      <h3>{@state.total} Active Users:</h3>
+      <ul>
+        {@state.users.map(@userLink)}
+      </ul>
+      {if @state.pageCount > 1
+        <Paginator page={+@state.page} onPageChange={@onPageChange} pageCount={@state.pageCount} scrollOnChange=false />
```
aweiksnar: You can also pass `firstAndLast=false` to the paginator to avoid having to hide those buttons with css. Might be good to wrap scrollOnChange's false with curleys as well

app/api/config.coffee
```diff
@@ -14,20 +14,30 @@ TALK_HOSTS =
...
 
+isDevelopment = env isnt 'production'
+
 module.exports =
   host: hostFromBrowser ? hostFromShell ? API_HOSTS[env]
   clientAppID: appFromBrowser ? appFromShell ? API_APPLICATION_IDS[env]
-  talkHost:  talkFromBrowser ? talkFromShell ? TALK_HOSTS[env]
+  talkHost: talkFromBrowser ? talkFromShell ? TALK_HOSTS[env]
+  sugarHost: sugarFromBrowser ? sugarFromShell ? SUGAR_HOSTS[env]
+  isDevelopment: isDevelopment
```
aweiksnar: I think @brian-c argued against exposing env here [here](https://github.com/zooniverse/Panoptes-Front-End/pull/440). I don't mind either way

app/api/sugar.coffee
```diff
@@ -0,0 +1,32 @@
...
+          sugarClient.on 'response', (args...) ->
+            console.log '[SUGAR RESPONSE] ', args...
+
+        sugarClient.connect()
+      else
+        window.sugarClient?.disconnect()
+    .catch (e) ->
+      throw new Error "Failed to checkCurrent auth from sugar client"
+
+module.exports = window.sugarApiClient = sugarApiClient
```
aweiksnar: It took me a min to understand the qualified vs unqualified requires here. It may be worth exporting both and the destructuring?

app/api/config.coffee
brian-c: Right, envify turns `process.env.NODE_ENV` into `"production"`, and then uglify turns `"production" === "production"` into `true` and can remove any conditionally unreachable code at that point.

app/talk/active-users.cjsx
parrish: Yeah, with the limited space there, I just wanted to go to previous/next only. (also excluding the selector).  Good tip though

app/api/sugar.coffee
parrish: Right, that's a bit weird.  On the other hand, sugarClient(the socket client) never needs to be instantiated by external code, while the sugarApiClient(the HTTP client) does.  I can certainly change it for readability

parrish: Actually, since all the api clients are global anyway, then it really doesn't matter to me what's exported

aweiksnar: Ah yeah, I believe the `window`ed ones were set for debugging

Issue #969, opened 2015-07-04
Collecting images/creating collections doesn't work properly (Whales as indi.)


yshish:
Yesterday, I marked one image with heart icon to collect as Favourite. 
Then I created a new collection and call it *Questions / Answers* and added there one image I commented.

Today I marked another image with heart but accidentally clicked twice on the icon. I expected to see the image twice in the Favourite collection but the image was actually missing there. I tried to add it again and reloaded the page but without success.
Moreover, the *Question / Answers* collection is missing there too - I can't see it under My collections. While typing its name to search the collection, it says *No results found*.

Is it supposed to be working or not yet?

chrissnyder:
The collections interface still needs work, but this should be working.

Looking at your collections list (https://www.zooniverse.org/#/collections/yshish) I can only see created one.

Does logging in/out change what you see on that link at all?

Clicking one on the favorite icon adds it, clicking it again removes it. So that's expected behavior, though it can be communicate better. Related to #972

yshish:
Thanks, Chris.

Yes, signing in/out changes what I see under My collections. I created the WOW (visible publicly) one today when I was experimenting with them. But the one I mentioned yesterday seems to be lost completely. 

This is what I see as my collections (signed in):
https://lh3.googleusercontent.com/-iIGcsLjOzL8/VZhDpi4kFsI/AAAAAAAAIEo/m_w4xjrwq3w/s0/ysh_2015-07-04_17-36-20.jpg
(I don't understand why there are two *Favorites* collections and what the *(11)* means.) I didn't created any collection called Favorites!) But all the images collected in both of them were collected using the heart icon.

And this is what the first *Favorites(11)* looks like when I open it..
https://lh3.googleusercontent.com/-cjOPyby4wf8/VZhDp3Rx1LI/AAAAAAAAIEs/2VKSdnEc78c/s0/ysh_2015-07-04_17-36-42.jpg

Maybe an important part, when I tried clicking on the heart icon again today, it changed its colour to red which it yesterday didn't (strange!). Yesterday, it didn't display whether the image has been collected or not anyhow.

Hope this is helpful. But at least, it definitely works better today.

vrooje:
I'm having problems with collections too, specifically adding images to existing collections.

When I click to add a subject to a collection, my existing collections don't come up in the search box.

[...]

Issue #879, opened 2015-06-29
Only see Zooniverse-wide public collections if I'm not logged in


mschwamb:
When not logged in I can find all public collection (currently only the one I made):
![screen shot 2015-06-30 at 3 23 53 am](https://cloud.githubusercontent.com/assets/753325/8416327/90449c20-1ed7-11e5-9030-91bbab6ead4d.png)

When I log in, it's default to my collections and no way to see other public collections

![screen shot 2015-06-30 at 3 24 32 am](https://cloud.githubusercontent.com/assets/753325/8416344/ab222fd0-1ed7-11e5-9e4e-64aed7261eab.png)

mschwamb:
If it helps - 

So slightly different behavior now that there are more collections - you see the collections but with yours too (including ones that appear private I think). 
Not signed in
![screen shot 2015-06-30 at 11 07 13 am](https://cloud.githubusercontent.com/assets/753325/8423210/65be77c8-1f18-11e5-87b8-0495625a91c5.png)

signed in

![screen shot 2015-06-30 at 11 09 27 am](https://cloud.githubusercontent.com/assets/753325/8423219/99a6815c-1f18-11e5-8d1b-bc048b1807fe.png)

![screen shot 2015-06-30 at 11 07 35 am](https://cloud.githubusercontent.com/assets/753325/8423222/a525d9e2-1f18-11e5-858c-f6917b88a111.png)

chrissnyder:
I think I misunderstood this issue when I read it two weeks ago. Is the concern that that top level collections page at https://www.zooniverse.org/#/collections is different depending upon if you are logged in or not?

If so, I'm not sure I'm convinced this is an issue. As it stands, it could be thought of as a "Viewable Collections" page, which naturally includes your own collections if logged in.

It could be a "Public Collections" page, where everything is explicitly public, but I don't have a feel if that would be more or less intuitive than the above.

Doesn't help we named the page "All Collections" which isn't strictly accurate in either scenario.

mschwamb:
This might help - firstly if you hit collections it shows my private collections combined with all the zooniverse collections Zooniverse wide - so it shows 55 collections if I'm logged in and hit collections and if I'm not logged in 53 - I think that's confusing to any user  will wonder why would their private collections show with all the Zooniverse wide ones

mschwamb:
[...]

PR #998, opened 2015-07-08
Towards user as an app wide prop

chrissnyder:
Putting this out there for discussion, not for merging. I only made the changes to a fraction of the codebase.

This PR would remove any and all `auth.checkCurrent()` from anywhere within the app, except for at the top-level `<App/>` component and have it's currrent value passed down via props throughout the entire app.

Benefits:

- No need for any ChangeListeners, PromiseRenderers, or PromiseToSetState for the user resource anywhere in the app, except at the top. Simply make your render method depend upon @props.user and you are good.
- Formalize the user resource as app-level state, which is what it is.
- Followed correctly, would eliminate bugs where component-level user state gets out of sync with app-level user state, e.g. #969, #879, and others.

Cons:

- Have to pass user down the entire tree of components, as far as is needed.

Thoughts?

### DIFF
--- a/app/collections/create-form.cjsx
+++ b/app/collections/create-form.cjsx
@@ -1,6 +1,5 @@
 React = require 'react'
 apiClient = require '../api/client'
-auth = require '../api/auth'
 counterpart = require 'counterpart'
 Translate = require 'react-translate-component'
 
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -1,6 +1,5 @@
 React = require 'react'
 apiClient = require '../api/client'
-auth = require '../api/auth'
 getFavoritesName = require './get-favorites-name'
 alert = require '../lib/alert'
 SignInPrompt = require '../partials/sign-in-prompt'
@@ -12,10 +11,12 @@ module?.exports = React.createClass
   propTypes:
     subject: React.PropTypes.object # a subject response from panoptes
     project: React.PropTypes.object # a project response from panoptes
+    user: React.PropTypes.object # a project response from panoptes
 
   getDefaultProps: ->
     subject: null
     project: null
+    user: null
 
   getInitialState: ->
     favoritesPromise: null
@@ -70,18 +71,17 @@ module?.exports = React.createClass
       @setState favoritedPromise: Promise.resolve(true)
 
   toggleFavorite: ->
-    auth.checkCurrent().then (user) =>
-      if user?
-        Promise.all([@state.favoritesPromise, @state.favoritedPromise])
-          .then ([favorites, favorited])=>
-            if not favorites?
-              @createFavorites(user)
-            else if favorited
-              @removeSubjectFrom(favorites)
-            else
-              @addSubjectTo(favorites)
-      else
-        @promptToSignIn()
+    if @props.user?
+      Promise.all([@state.favoritesPromise, @state.favoritedPromise])
+        .then ([favorites, favorited]) =>
+          if not favorites?
+            @createFavorites @props.user
+          else if favorited
+            @removeSubjectFrom(favorites)
+          else
+            @addSubjectTo(favorites)
+    else
+      @promptToSignIn()
 
   render: ->
     <PromiseRenderer promise={@state.favoritedPromise}>{(favorited) =>
--- a/app/collections/home.cjsx
+++ b/app/collections/home.cjsx
@@ -1,12 +1,7 @@
 React = require 'react'
 CollectionCreateForm = require './create-form'
-talkClient = require '../api/talk'
 apiClient = require '../api/client'
-authClient = require '../api/auth'
-auth = require '../api/auth'
 {Link} = require 'react-router'
-ChangeListener = require '../components/change-listener'
-PromiseRenderer = require '../components/promise-renderer'
 
 module?.exports = React.createClass
   displayName: 'CollectionsHome'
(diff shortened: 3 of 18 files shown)


camallen:
This makes a lot of sense to me. Is there any way to make the user available as a top level prop on some component that is embedded in each page?

brian-c:
I'm really not crazy about trickling down app-wide state like this.

The reason for the current `<ChangeListener><PromiseRenderer>` stuff is that I wanted each component to be able to render itself without much context. In practice this was way less useful than I thought it'd be, but I still think a component should generally be able to stand on its own and not have to belong to a specific rendering hierarchy just to access the data it needs.

A sorta middle-ground I tried was re-rendering from the top on auth change and providing a direct reference to the current user from the auth module. It made changing the rendering based on the user smoother, but accessing links from the user still meant manually updating state in `componentDidMount` and `componentWillReceiveProps`. It was less verbose, but only slightly. I trashed that code but could rewrite it pretty quickly if we want to compare.

aweiksnar:
Strongly in favor of this. For the reasons Chris mentioned, and also so that the user is accessible anywhere inside the component. The `<ChangeListener><PromiseRenderer>` pattern scopes it to the children, and then requires rewriting to state/parent-props if needed outside of that. I like this because there's only one way to get a user. Also for less verbosity.

chrissnyder:
@camallen You get that for free by doing https://github.com/zooniverse/Panoptes-Front-End/commit/d7222a5b612467a6496b349e4073b0e69f3bf15a#diff-c13222b43536c8bbe0f944e7ded691c0R41

@brian-c From a page components perspective, I definitely disagree, as their place in the hierarchy is pretty well set as-is. For a more pure component's perspective, I can sort of understand, but I still disagree. I find components most reusable when they are kept as close to pure views as possible where dependencies are declared and passed as props, rather than having to require in dependencies within the component. 

More broadly, I actually like the fact that you have to pass user down throughout the whole tree. I find having an explicit dependency makes reasoning about the app easier. If the user isn't passed to a component, then I know immediately that nothing within that sub-tree depends upon the user. Same for the flip side.

It also gives us more flexibility to iterate on the auth module, as app dependency upon the current implementation is greatly lessened.

@edpaget @parrish Thoughts?

parrish:
Actually, we started off by doing it this way originally.  I can't completely remember exactly why we moved away from it.

brian-c:
```
Foo = React.createClass
  render: ->
    <Bar />
```

Here, let's say `Bar` doesn't need the current user, but its context `Foo` shouldn't have any idea if one of its eventual descendants does. If it did then every component would have to know about every other component that could be rendered under it. That's way too much responsibility.

srallen:
Passing a user down from a top level parent as a prop makes more sense to me too. I did that with Chimp&See and am starting out mammoth bats that way too. I didn't find having to pass down the user as a prop as too much responsibility, but the scale is much smaller for individual projects than it is here. I think that ```<ChangeListener><PromiseRenderer>``` are still useful and could be used in cases where a parent doesn't need to know about the user and only a specific child does (if for instance a component on an About page needed to know, but the parent About page doesn't and none of the other children need to know either). Otherwise, it's passed down as a prop from the main app component.

chrissnyder:
Messed with this a bit more last night, some thoughts below.

I converted the Collections set of routes to use this pattern, and a couple things stood out.

There were a couple instances of just passing `@props.user` down rather than actually using it in the component itself, though never more than one level skip (`<CollectionsPage/>` into `<CollectionsNav />` for example.

It also sort of depends upon how you design your components. So for a simple example, you could design `CollectionsManagerIcon` like this:

```
  displayName: 'CollectionsManagerIcon'

  toggleCollectionsManagerPopup: ->
    alert (resolve) =>
      <div className="content-container">
        <CollectionsManager user={@props.user} project={@props.project} subject={@props.subject} onSuccess={resolve} />
      </div>

  render: ->
    <button
      className="collections-manager-icon"
      onClick={@toggleCollectionsManagerPopup}>
      <i className="fa fa-list" />
    </button>
```

or you could do it like this:

```
  displayName: 'CollectionsManagerIcon'

  propTypes:
    user: React.propTypes.object
    
  getDefaultProps:
    user: null

  toggleCollectionsManagerPopup: ->
    alert (resolve) =>
      <div className="content-container">
        <CollectionsManager user={@props.user} project={@props.project} subject={@props.subject} onSuccess={resolve} />
      </div>

  render: ->
    <div>
      {if @props.user
        <button
          className="collections-manager-icon"
          onClick={@toggleCollectionsManagerPopup}>
          <i className="fa fa-list" />
        </button>}
    </div>
```

which loosens up the requirement to pass everything down. I still prefer making the eventual dependency explicit, as it's still a dependency, whether you declare it or not. Though that might be hairy in a very deeply nested page structure.

Brian, would you be up for seeing what the `ComponentDidMount` and `ComponentWillReceiveProps` looks like and comparing?

parrish:
Something to consider:

If (basically) all components have the user as a prop, they'll need to resend all of their API requests whenever the user state changes -- even if the result of the request doesn't depend on the current user.  I'm not sure how much request thrashing that would cause, but probably a fair amount.

On the other hand, we already have a decent amount of request thrashing on a logged in page load. -- e.g. we request a lot of stuff, get the user session, then have to re-request the resources.  If we make the user dependency explicit, we could defer those requests until the user session has resolved.

chrissnyder:
Like so https://github.com/zooniverse/Panoptes-Front-End/blob/top-level-user-prop/app/main.cjsx#L40-L41

brian-c:
Found my old commit: https://github.com/zooniverse/Panoptes-Front-End/compare/experimental-rerendering?w=1

Changes to the auth module make `auth.pending` and `auth.current` available. `.current` is a reference to the current user, not a promise, and because the app re-renders when the user changes it's right whenever you access it. (I'm not 100% that this is always in sync right now, I just hacked it into some reasonable spots on the existing auth module.)

The Lab index is the only page I've updated to use the new stuff, just to see if it worked.

I believe I abandoned this because it didn't reduce the amount of code needed as much as I expected it to.

edpaget:
Closed by #1094

PR #1018, opened 2015-07-09
Better user profiles

brian-c:
This does a couple things, sorry:

- Restores the feed to the user profile page. It's just a list of user comments for now. In the future it could include "Created a collection" and "Added to this collection" and that type of event, I think. I think there was an issue for this but I can't find it now.

- Adds a default `onPageChange` to the paginator, which updates the hash's `?page=n` param, which I think is what we'll want 90% of the time, right?

- The "Message" link now brings up the form to send that user a message. Before it had all kinds of redirects (to inbox and collections), which was sorta confusing. If I've misunderstood something here, let me know.


aweiksnar:
Wow this is fantastic. Great update

PR #1019, opened 2015-07-09
Add a stats page

### LINKED ISSUE #633: View/display the number of classifications you've made?
Hello,

Is it possible for a user to see how many classifications they have made in a project?

If not, might this be possible - or is there a good reason not to do this?

Thanks, Tom

brian-c:
Kinda like the old one, with a ribbon and project icons. Needs some polish, but it's a start.

Closes #633.

### DIFF
--- a/app/components/classifications-ribbon.cjsx
+++ b/app/components/classifications-ribbon.cjsx
@@ -1,0 +1,53 @@
+React = require 'react'
+
+module.exports = React.createClass
+  getDefaultProps: ->
+    projects: for i in [0...20]
+      {project: "Project #{i}", classifications: Math.floor Math.random() * 100}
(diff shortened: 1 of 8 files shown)


chrissnyder:
That's awesome!

Only two things I noticed:

- Should the "Only you can see your stats." message only be shown if the user is viewing their own stats?
- Styling a little off if a user hasn't done any classifications
screen shot 2015-07-09 at 2 19 47 pm

brian-c:
I think there should always be some text there that says stats are private. I just updated it to be a little less awkward, I think.

The empty ribbon is now styled a little more nicely.

chrissnyder:
Looks good to me.

:shipit:

PR #1059, opened 2015-07-14
Change "science case" to "research"

### LINKED ISSUE #352: science-case page: what if it's not a science project?
Even if I leave the science page content-free when designing a project, it still appears as a tab (which I think is by design as that page is where the project team and roles show up). 

But not everyone will find a page called "Science" relevant (e.g. humanitarian projects), and I can't change the name of the tab.

I suspect this isn't a first-release issue but I'm just throwing it out there that we should broaden the label of this page, or allow users to change the tab text.

brian-c:
Close #352.


edpaget:
Can you change the pageTitle property https://github.com/zooniverse/Panoptes-Front-End/blob/research/app/pages/lab/research.cjsx#L10 to "Research"?

PR #1066, opened 2015-07-14
Add autocompleting field for tagging projects

edpaget:
...in the Project Lab details page

### DIFF
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -53,9 +53,9 @@ RoleCreator = React.createClass
 
         <p className="form-help error">{errorMessage}</p>}
       <form style={style}>
-        <p>
+        <div>
           <UserSearch />
-        </p>
+        </div>
 
         <table className="standard-table">
           <tbody>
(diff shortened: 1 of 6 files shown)


edpaget:
Still waiting for design stuff for tag search on the main projects page. This just adds a field to project details for adding tags.

edpaget:
Also there's a bug were you can't delete the last tag from a project at the moment. It's a backend issue I will have a fix for soon.

brian-c:
This doesn't seem to register a change unless it autocompletes, and it doesn't autocomplete unless the tag has been used before.

Looks like tags are resources, so should these be links instead?

edpaget:
@brian-c it should complete the tag when you type a comma. I don't want to implement it as a resource link, so you don't have to search the tags resource first to find ids.

edpaget:
I will set it up so clicking out side converts whatever remains to a tag too though.

edpaget:
@brian-c changed it to add the current tag when you blur the tag input. Let me know if that works for you.

brian-c:
Cool, works for me.

PR #1067, opened 2015-07-14
Link to talk subjects with just ^

### LINKED ISSUE #882: Having subject ids mentioned automatically become a link in subject talk without mentioning owner and project name
I noticed people are using the Mention subjects with @owner/project^subject_id thing - Is it possible that like the previous Talk platforms if you mention a subject id  in a project talk if that subject number exists in that project that it automatically becomes a link to the subject page rather than this complex notation.

It was a handy feature in previous version of Talk and I think have it automatically happen would be a big help

edpaget:
...when inside a project.

Closes #882


edpaget:
Note this doesn't do anything to make sure the subject is actually in the project you make the reference from (and generally a subject's Talk page is similar).

aweiksnar:
:shipit: :shipit: :fire: :fire: LGTM 
Let me know if there are any ways I can make the subject pages more stable with any coming api updates

edpaget:
Will do.

PR #1072, opened 2015-07-15
Better behavior surrounding empty subject set names

### LINKED ISSUE #870: Funky behaviour creating new subject sets
When you go to create a new subject set, if you delete the "Untitled subject set" title, the new subject set appears in the left-hand menu as a narrow strip (because there is no text in the title). You can save a subject set without a title. 

But it's also not obviously created a new subject set - say you delete the "untitled subject set" title, but then change your mind about creating the subject set. You now have a subject set without a title, without any subjects, but that is selectable on the left hand menu.

screen shot 2015-06-29 at 13 46 26

chrissnyder:
- Closes #870
- Delete `app/pages/create-project.cjsx` which has been empty for awhile


edpaget:
:100:

PR #1073, opened 2015-07-15
Polygon hint line

### LINKED ISSUE #650: Polygon drawing tool unintuitve and unfriendly
This is probably the biggest negative comment we got in our Season Spotter beta (see #443). Users had a lot of difficulty figuring out how to use the polygon tool. In particular, they didn't realize they had to click multiple times. 

> I couldn't figure out how to draw around an area. It would just mark one place and count that as one plant.

and

> It was also a bit hard at first to understand how the tool to draw the shapes works -- I though[t] I should be holding down the mouse and drawing around the shape all at once, rather than clicking on various points around the shape.

I admit that it took me a little while to figure out how to use it myself. I'm not sure it's obvious how to finish a shape either (although having both double-click and click-on-the-original-vertex is good).

One user specifically asked for a tutorial on how to use the polygon tool!

chrissnyder:
- Adds a line that follows the mouse while marking a polygon. Useful for both training new users and allowing veterans to be more precise.
- Closes #650 maybe?

Pinging @aliburchard and @mkosmala for confirmation, demo at http://demo.zooniverse.org/panoptes-front-end/polygon-hint/#/dev/classifier

### DIFF
--- a/app/classifier/drawing-tools/polygon.cjsx
+++ b/app/classifier/drawing-tools/polygon.cjsx
@@ -33,6 +33,18 @@ module.exports = React.createClass
     forceComplete: (mark) ->
       mark.closed = true
 
+  componentWillMount: ->
+    @setState
+      mouseX: @props.mark.points[0].x
+      mouseY: @props.mark.points[0].y
+      mouseWithinViewer: true
+
+  componentDidMount: ->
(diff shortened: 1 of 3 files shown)


mkosmala:
Yes, closes #650.

marten:
LGTM

edpaget:
:shipit:

PR #1079, opened 2015-07-15
a11y fixes for dialogs

eatyourgreens:
Labels the close button so it's read as 'close', not 'times', by VoiceOver.
Adds the dialog role to the containing div. Should be announced by VoiceOver.
Fixes the keydown handler so that focus cannot leave the dialog when it is open.

### DIFF
--- a/app/components/dialog.cjsx
+++ b/app/components/dialog.cjsx
@@ -1,6 +1,6 @@
 React = require 'react'
 
-FOCUSABLES = 'input, textarea, button, select, [tabindex]'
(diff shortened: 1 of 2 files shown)


chrissnyder:
Not sure how to test the VoiceOver stuff, but the focus handling works like a charm.

In the future, do assign a PR to somebody, otherwise it's in danger of just sitting here. Brian, Alex, myself, Ed, all look at them (really anybody).

eatyourgreens:
Cool, will do.

cmd-f5 turns voiceover on/off on OSX. I usually turn it on then tab through to the control that I want to test.

I'm working on making the user account menu accessible accessible for Wellcome, so I'll port those changes over to PFE when I'm happy too. 
https://github.com/zooniverse/wellcome/compare/account-bar?expand=1

PR #1080, opened 2015-07-15
Fix collection paging

### LINKED ISSUE #1006: Pagination broken on the collections page
The collections page always shows me the first 20 collections, regardless of the value of the `page` query parameter eg. https://www.zooniverse.org/#/collections?page=3 I've been able to reproduce this in a few different browsers (Chrome, Firefox, Safari)

edpaget:
Closes #1006

### DIFF
--- a/app/pages/collections.cjsx
+++ b/app/pages/collections.cjsx
@@ -49,9 +49,10 @@ CollectionsPage = React.createClass
     'collection-show'
 
   listCollections: ->
-    query = Object.create @props.query ? {}
+    query = {}
     query.owner = @props.params.owner if @props.params?.owner?
     query.include = 'owner'
+    Object.assign query, @props.query
 
     apiClient.type('collections').get query
 


app/pages/collections.cjsx
```diff
@@ -49,9 +49,10 @@ CollectionsPage = React.createClass
     'collection-show'
 
   listCollections: ->
-    query = Object.create @props.query ? {}
+    query = {}
     query.owner = @props.params.owner if @props.params?.owner?
     query.include = 'owner'
+    Object.assign query, @props.query
```
chrissnyder: Shouldn't this be `query = Object.assign query, @props.query`?

edpaget: Nope Object.assign mutates the first parameter

http://jsfiddle.net/zkpwg3ns/ and look at your console.

chrissnyder: Ooo, fancy, didn't know that.

PR #1082, opened 2015-07-15
Hide some metadata, render the rest as Markdown

### LINKED ISSUE #181: Subject metadata hyperlink
We want to allow our users to go to a link based on the image shown. I'm thinking the best way to do this is to provide a URL in the subject metadata. Would it be possible to have this URL automatically become a hyperlink, perhaps through Markdown or some other sort of tag?

### LINKED ISSUE #285: Everything in the manifest file showing up in the metadata while classifying -
Under every subject when classifying now is a little white icon and if you click it it says metadata and appears to be spitting out everything that was in the manifest file which includes file name and In some cases ids that could be googled whie classifying. 

Example from the Galaxt Zoo 3D
image

The manga survey Galaxy ID is visible now to a classifier where likely you want to show that info on Talk 

The best example I have is Planet Hunters, if the star id was shown then people could google snd find out it is s planet candidate and they would be more likely to mark transits (even if there wasn't one in the section of data they were looking at) so I think there needs to be either the option to turn the metadata button off in the short term and the ability later to decide which columns in the manifest are showing by the  metadata button

brian-c:
Closes #285.
Closes #181.

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -129,8 +129,9 @@ module.exports = React.createClass
   showMetadata: ->
     # TODO: Sticky popup.
     alert <div className="content-container">
-      Subject metadata<br />
-      <table>
+      <header className="form-label" style={textAlign: 'center'}>Subject metadata</header>
+      <hr />
+      <table className="standard-table">
         {for key, value of @props.subject?.metadata when key.charAt(0) isnt '_'
           <tr key={key}>
             <th>{key}</th>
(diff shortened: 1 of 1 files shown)


brian-c:
Oops sorry, hold off on this please.

brian-c:
My bad, "_"-prefixed keys are never saved anyway, so I switched to a comment-style "#"- or "//"-prefix. Should be good now.

aweiksnar:
Cool LGTM :shipit:

Issue #1083, opened 2015-07-15
User role pagination


parrish:
It looks like @mrniaboc has more roles than are returned in a single page of user roles.

We should add a higher limit to the request to fetch them.

aweiksnar:
Closed via #1089

PR #1089, opened 2015-07-15
Increase role limit for user role requests

aweiksnar:
- Increase page_size of requests for a user's roles
- For #1083


chrissnyder:
the :8ball: says :shipit:

aweiksnar:
Hahahah

PR #1093, opened 2015-07-15
filter roles by project section and zooniverse wide

### LINKED ISSUE #1085: Can't see talk moderator controls for a project I'm a collaborator on
Seems my section project roles are in the next page of roles from the api and not being assigned correctly as per project talk retrieves all user roles and not all user : project roles, https://talk.zooniverse.org/roles?user_id=X

@brian-c this is the issue i was talking about on slack. Can provide more details if needed.

camallen:
closes #1085


aweiksnar:
:metal: :shipit: :metal:

PR #1096, opened 2015-07-16
Accessible account menu

eatyourgreens:
Adds an expandable button control to make the account menu accessible from the keyboard.
Enter/Space: toggle menu on/off
Up/Down keys: navigate menu when expanded.
Adds aria attributes for VoiceOver too.


chrissnyder:
:+1:

PR #1097, opened 2015-07-16
set checkbox checked value

### LINKED ISSUE #1060: "Volunteers can choose which workflow they work on" check box is broken
When you check the "Volunteers can choose which workflow they work on" box and then navigate to another aspect of Panoptes, the box resets and volunteers are not given an option of what workflow to work on when they click directly on the "Classify" tab.  Instead, users are randomly assigned a different workflow for each subject they see.  Refreshing the Project Builder edit screen, or going directly from the Project Details page to the project itself, does not resolve the issue.

camallen:
closes #1060

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -127,7 +127,7 @@ module.exports = React.createClass
           <p>
             <AutoSave tag="label" resource={@props.project}>
               {checked = @props.project.configuration?.user_chooses_workflow}
-              <input type="checkbox" name="configuration.user_chooses_workflow" checked={checked} value={checked} onChange={handleInputChange.bind @props.project} />{' '}
+              <input type="checkbox" name="configuration.user_chooses_workflow" defaultChecked={checked} defaultValue={checked} onChange={handleInputChange.bind @props.project} />{' '}
               Volunteers can choose which workflow they work on
             </AutoSave>
             <br />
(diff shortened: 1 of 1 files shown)


app/pages/lab/project-details.cjsx
```diff
@@ -126,7 +126,8 @@ module.exports = React.createClass
 
           <p>
             <AutoSave tag="label" resource={@props.project}>
-              <input type="checkbox" name="configuration.user_chooses_workflow" value={@props.project.configuration?.user_chooses_workflow} onChange={handleInputChange.bind @props.project} />{' '}
+              {checked = @props.project.configuration?.user_chooses_workflow}
+              <input type="checkbox" name="configuration.user_chooses_workflow" checked={checked} value={checked} onChange={handleInputChange.bind @props.project} />{' '}
```
edpaget: I think these should be defaultChecked and defaultValue right? https://facebook.github.io/react/docs/forms.html

camallen: looks like it!

camallen:
@edpaget updated, nice catch. I need to read the react docs.

brian-c:
Why's this uncontrolled? `handleInputChange` updates the project.

camallen:
Ok the docs point to it being controlled and it won't change, https://facebook.github.io/react/docs/forms.html#controlled-components

So your saying any `onChange` event that update the state will flow down via the render function? Explains the observed behaviour being the same. Do you want me to update a fix?

PR #1100, opened 2015-07-16
Set remember_me flag so devise holds the sesion

edpaget:


chrissnyder:
Would this replace the need to refresh the token once an hour?

edpaget:
It shouldn't affect any token stuff that's happening now. What was happening was the devise cookie (that the front-end uses to fetch the token) would expire when the browser window was closed. This just makes sure that cookie remains set for 2 weeks before expiring.

chrissnyder:
:+1:

PR #1103, opened 2015-07-16
add additional manifest description in how-to

vrooje:
Have checked the markdown but not in the actual demo site.


edpaget:
I don't think it's from anything added here, but some of the line breaks on that page seem odd. For instance happening in the middle of a sentence.

Issue #874, opened 2015-06-29
Add links/info for most recent posts to discussions


DarrenMcRoy:
![screen shot 2015-06-29 at 1 26 32 pm](https://cloud.githubusercontent.com/assets/8987864/8415166/84a7efc2-1e62-11e5-8da2-8c2a58ad95b3.png)

There's currently no way to see who has made the most recent post on a board, and when--or a link to jump right to that most recent post... only info about who started it.

This is an important part of forum UX and, I think, should be added as soon as possible.

I also think that we should get rid of the timestamps for when a board was started. And maybe even the name of the person who started the board. That stuff really isn't important... the most recent post is.

aweiksnar:
@DarrenMcRoy Your screenshot is of discussions, but comments are about boards. Assuming you mean adding the same info presented in board links to discussion links? 

Those are also sorted by latest updated too

DarrenMcRoy:
Whoops. I meant discussions.

When looking at a list of discussions in a board, you should see who left the most recent post on the discussion, and when, and be able to click through to it.

Who started the discussion is not as important, and when they started it is certainly not important.

Does that make sense? :-)

DarrenMcRoy:
A la forum UX: shows both author and most recent post, with a link to go to said recent post.

![screen shot 2015-06-30 at 10 34 05 am](https://cloud.githubusercontent.com/assets/8987864/8434861/aa58be3e-1f13-11e5-81b9-2528e9dab1c6.png)

yshish:
I think it would be very helpful if the summary said number of 'unread' posts too (the best would be a different colour so the moderators saw which thread contains unread posts by a quick view on the summary).

PR #1106, opened 2015-07-16
Use discussion link comment for board preview

### LINKED ISSUE #873: Most recent post link doesn't go to the most recent post

aweiksnar:
- Closes #873

### DIFF
--- a/app/talk/board-preview.cjsx
+++ b/app/talk/board-preview.cjsx
@@ -34,10 +34,7 @@ module?.exports = React.createClass
 
         <PromiseRenderer promise={talkClient.type('discussions').get({board_id: @props.data.id}).index(0)}>{(discussion) =>
           if discussion?
-            comments = discussion.links.comments
-            lastCommentId = comments[comments.length-1]
-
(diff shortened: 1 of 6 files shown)


edpaget:
Shouldn't the behaviour be the same on this page too? 

screenshot from 2015-07-16 16-39-26

aweiksnar:
Those discussion previews are currently correct in that they both advertise & display the time/user they were started from (originally took that line from the design).

I will display the latest comment in those though since it was requested in #874, but can do that in a separate PR

aweiksnar:
Edit: Just added in the the same behavior for that other page, so if that looks okay this'll close #874 too

edpaget:
:fist:

PR #1137, opened 2015-07-20
Wrap user settings page in a change listener

### LINKED ISSUE #1133: Display Name field in User Settings won't update on keyboard input
Both the display name and credited name fields won't visually respond to keyboard input. Copy pasting content works, but again the changes aren't visualized until the page is reloaded.

Something to do with refreshing the page I guess.

edpaget:
Closes #1133

### DIFF
--- a/app/pages/settings/index.cjsx
+++ b/app/pages/settings/index.cjsx
@@ -1,6 +1,7 @@
 React = require 'react'
 counterpart = require 'counterpart'
 Translate = require 'react-translate-component'
+ChangeListener = require '../../components/change-listener'
 {Link, RouteHandler} = require 'react-router'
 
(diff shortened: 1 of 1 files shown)


chrissnyder:
This fixes it, but it really shouldn't be needed. Is `@props.user` not correctly updating when it's changed? If not, could we put the change listener in `app/main.cjsx`?

edpaget:
No it's updating fine. There's just nothing listening to the change. React won't re-render because a sub-property of an object changed. 

The other possibility is to change the `value` property to `defaultValue`.

chrissnyder:
Ah, of course, I'm dumb.

In that instance, I think your approach does the job. `defaultValue` would work here, but then we'd need to worry about this again any other time we change the resource here.

PR #1139, opened 2015-07-20
Remove science case param

edpaget:
...since that's been moved to a separate resource


edpaget:
Pretty sure this is what's causing all the errors.

PR #1143, opened 2015-07-20
Add ability to lock discussions

### LINKED ISSUE #1039: Adding ability for moderators to lock Non-Note board discussion threads on Talk
Right now you can lock a message board and make it write only by admins, but currently not a specific thread. This feature has been used by moderators  in boards on previous Talks to end arguments between volunteers and end discussion on things that more than already played out and are just rehashing rather than civil conversation. It's not a frequent thing, but it would be helpful if this existed.

aweiksnar:
- Closes #1039


edpaget:
:+1: Looks like the reply window briefly flashes when the page first loads. It'd be sweet to take care of that, but we can handle it later.

aweiksnar:
Ah I missed that! Yeah I'll put in another PR to improve loading

PR #1146, opened 2015-07-20
Catch breadcrumbs for unpermitted boards

### LINKED ISSUE #1047: Gracefully handling  trying to go to a restricted access Talk board while non-logged in -
low in the queue - Talk should gracefully handle trying to go to a restricted access board or thread while non-logged in or for an account that doesn't have access - currently get a bunch of errors/json spit out

screen shot 2015-07-13 at 10 02 28 am

aweiksnar:
- Closes #1047

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -85,6 +85,7 @@ routes = <Route handler={App}>
     <Route name="project-classify" path="classify" handler={require './pages/project/classify'} />
     <Route name="project-talk" path="talk" handler={require './pages/project/talk'}>
       <DefaultRoute name="project-talk-home" handler={require './talk/init'} />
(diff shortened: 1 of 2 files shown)


edpaget:
I think this needs the :( Not Found error you get when you visit a project that's private.

edpaget:
Otherwise the page is a blank and user can't be sure of what error was encountered.

aweiksnar:
Ah yeah way better idea. I'll work on adding in a 'You don't belong here' type page, since apparently the react router NotFoundRoute is only indended for [poorly formed urls](https://rackt.github.io/react-router/#NotFoundRoute)

edpaget:
Couple small things left. It looks like the boards remain visible if you log out from your account but don't refresh the page. 

And if you try to visit the talk of a private project you don't have access do it still just shows a blank page instead of a not found error.

PR #1147, opened 2015-07-20
Autogenerate basic manifest when none is provided

### LINKED ISSUE #519: Option to auto-generate simple manifest file?
It occurred to me that having to generate a CSV file could be a real barrier for some users of the project builder - especially when we consider classroom usage... I realise this is a "nice to have", but why don't we add a check box "generate manifest automatically" which would create a simple one-column CSV with the filenames only, as uploaded?

### LINKED ISSUE #1144: "Upload 1 new subject" button does nothing
I add a subject (and manifest) and click the green "Upload 1 new subject" button. Nothing happens. I try clearing my cache and relogging in. Same thing -- nothing happens.

Oh, and I'm in Firefox on Windows.

edpaget:
Closes #519

Fix call to separateSubjects to make sure tooBigFiles is correctly
provided

Closes #1144


chrissnyder:
:+1:

PR #1148, opened 2015-07-20
Partially cut off final option in search fields

### LINKED ISSUE #1132: Add a visible scroll bar to the username search
Related to #1131 You can scroll in searches to see more but it's not obvious to the user see ticket  #1120 - Example looking for @chrissnyder - he's the last to come up in the search, but you have to scroll to find him. Scrolling works, but it looks from the design that are no more options. Adding a visible scroll bar would be helpful to know that there are more choices to go through 

screen shot 2015-07-19 at 5 59 17 pm

After scrolling 

screen shot 2015-07-19 at 5 59 20 pm

edpaget:
...to hint that you can scroll for more on OSes that don't include
scrollbars

Closes #1132

### DIFF
--- a/css/user-search.styl
+++ b/css/user-search.styl
@@ -13,3 +13,6 @@
     border: none !important
     box-shadow: none !important
     background: none !important
+
+  .Select-menu
+    max-height: 174px


css/user-search.styl
```diff
@@ -13,3 +13,6 @@
     border: none !important
     box-shadow: none !important
     background: none !important
+
+  .Select-menu
+    height: 174px
```
aweiksnar: Looks good, thoughts on switching to `max-height` so it can contract when underfilled?

edpaget: Yup that's totally want it should do :fire:

aweiksnar:
:+1:

PR #1167, opened 2015-07-21
Sort linked comments in latest comment previews

aweiksnar:
- Makes it so the latest comment links correctly go to the last page 
- #873

### DIFF
--- a/app/talk/board-preview.cjsx
+++ b/app/talk/board-preview.cjsx
@@ -25,7 +25,7 @@ module?.exports = React.createClass
 
         <p>{@props.data.description}</p>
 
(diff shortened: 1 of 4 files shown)


chrissnyder:
Got `Uncaught (in promise) TypeError: Cannot read property 'sticky' of undefined` in search using this. I think this unrelated to the PR, but https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/search-result.cjsx#L26 should be `<DiscussionPreview discussion={discussion} project={project} />`

aweiksnar:
Good find! just updated

chrissnyder:
:us:

chrissnyder:
Looking back on it, this does make me hate how I did the search-result component. Needs some love in the future.

PR #1173, opened 2015-07-21
Fix subject scaling for Firefox

### LINKED ISSUE #1172: Bug: Wide subjects don't display completely
Subjects which have a high width to height ratio (and are not tiny), don't display completely. Part of the left side of the image gets cut off (and the user has to scroll to see part of the right side). 

A user reports this here:
https://www.zooniverse.org/#/projects/kosmala/season-spotter-questions/talk/62/926?page=1

And I can reproduce it.

(related #1142)

brian-c:
Limit subject width in flex container (https://bugzilla.mozilla.org/show_bug.cgi?id=1113645).

Closes #1172.

### DIFF
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -8,9 +8,11 @@
     justify-content: center
     display: flex
     margin: 0 1.5vw 1em 0
+    min-width: 0
 
   .subject-container
     color: gray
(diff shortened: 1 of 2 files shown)


aweiksnar:
Looks good!

PR #1183, opened 2015-07-22
Match local files to S3 URLs by MIME type

### LINKED ISSUE #1152: Multiple files per subject: uploads assign wrong mime-type to uploaded image
Just discovered this, with @camallen, while using zooniverse/panoptes-subject-uploader but I think the same bug exists in PFE too, on this line.
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/subject-uploader.cjsx#L69

The subject.locations array in the API response is not guaranteed to be in the same order as the local files array, so matching on array index can associate the wrong local file with a newly created subject location.

Here's an example. curl https://panoptes-uploads.zooniverse.org/staging/subject_location/d270cd76-dada-42c9-af05-2d1a8aaeb5b5.jpeg brings back an uploaded text file, instead of an image.

I'm looking at how to fix this in the subject uploader, but thought I'd flag it here in case it has caused problems where subjects have multiple images associated.

eatyourgreens:
Closes #1152

### DIFF
--- a/app/partials/subject-uploader.cjsx
+++ b/app/partials/subject-uploader.cjsx
@@ -52,6 +52,15 @@ module.exports = React.createClass
     subjectData = @props.subjects[@state.current]
 
     if subjectData? and @state.inProgress
+      filesByType = {}
+      locationTypes = []
+    
+      for filename in subjectData.locations
+        type = @props.files[filename].type
+        locationTypes.push type
+        filesByType[type] ?= []
+        filesByType[type].push @props.files[filename]
+        
       locationTypes = for filename in subjectData.locations
         @props.files[filename].type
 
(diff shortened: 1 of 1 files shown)


chrissnyder:
Code looks fine. Got any manifests you were testing with?

From local testing, actually parsing the manifest happened correctly, but I'm getting 403s attempting to access the subject data after upload. Separate problem?

eatyourgreens:
Hmmm, not sure what's going on with those 403s. Here's the images I used to test it. They give 403s on the production front-end, but should work with this branch.

Github won't let me attach a csv, but the test manifest looked like this

```
subject_id, png, jpeg
1, b1799682x_0_0.png, b1799682x_0_1.jpg
```
b1799682x_0_0.png
b1799682x_0_0
b1799682x_0_1.jpg
b1799682x_0_1

chrissnyder:
Seems alright here, but I'm loath to change the subject ingestion process without another pair of eyes on it. Not very confident I'm looking at this from all the correct angles.

@edpaget Mind if you give it a look as well?

eatyourgreens:
Now that the returned locations have been fixed, in zooniverse/Panoptes/#1188, I think this PR is redundant.

chrissnyder:
Ah, a case of me being out of date then. Shalll I close this?

eatyourgreens:
Let's close this, I don't think there's anything in here that we need to add now.

PR #1184, opened 2015-07-22
Use markdown and comment editor boxes in Inbox

### LINKED ISSUE #1032: Links and Markdown don't work in private messages
If you write a url out in a private message  it doesn't become a automatic link like it does in Talk posts. Also when you write in markdown in a private message reply it doesn't work at all

Being able to add links both straight by writing it out and using markdown is pretty important for private messages. Many times volunteers are asking in a private message for help on thread or want to report an issue going on in a thread.

edpaget:
Closes #1032


aweiksnar:
:100: :shipit: Setting that max-width makes such a difference for readability. thanks!

Issue #937, opened 2015-07-01
Making the Talk message gui under your inbox the one that shows up when you click on a user profile


mschwamb:
low in the queue - with the tiny box size and such it would be nice if the mail gui that's in the inbox page shows up when you click on a username

what shows when you click on the username

![screen shot 2015-07-02 at 6 47 35 am](https://cloud.githubusercontent.com/assets/753325/8466789/2278221c-2089-11e5-8b31-baa8ee06eb3b.png)

what shows up under the inbox

![screen shot 2015-07-02 at 6 47 30 am](https://cloud.githubusercontent.com/assets/753325/8466793/2bd16ab2-2089-11e5-8224-bedee263facc.png)

aweiksnar:
Closed in #1199 - That page now has the cleaner looking comment-style markdown box

PR #1199, opened 2015-07-23
Use markdown in profile private message form

### LINKED ISSUE #1033: Size of the private message text box is extremely small
low in the queue, The current reply box for a private message is not very user friendly in terms of size. It's only big enough to view one or two lines. I'm constantly resizing it

Can it be made longer by default? 

Below is what size it looks like for me on Firefox

screen shot 2015-07-11 at 8 07 48 am

screen shot 2015-07-11 at 8 02 30 am

screen shot 2015-07-11 at 8 19 45 am

aweiksnar:
- Closes #1033

### DIFF
--- a/app/pages/profile/private-message.cjsx
+++ b/app/pages/profile/private-message.cjsx
@@ -5,7 +5,7 @@ module.exports = React.createClass
   displayName: 'PrivateMessagePage'
 
   render: ->
-    <div>
+    <div className="private-message-page">
       {if @props.user?
         <div className="content-container">
           <h2>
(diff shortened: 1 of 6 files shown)


aweiksnar:
Closes #937 too

edpaget:
Can we do a `max-width`?

aweiksnar:
I just set overflow to auto on the footer in 47bbb89 which should fix that site width issue. Let me know if that works for you

edpaget:
Sorry I meant on this page
screenshot from 2015-07-24 12-53-29

aweiksnar:
Ahh gotcha, for line width, not overflow! Yeah I'll set that

edpaget:
:ok_woman:

PR #1200, opened 2015-07-23
Decide already seen flag when component mounted

### LINKED ISSUE #1109: Already seen label pops up when you get to the summary and can hit the Talk or Next button
Now when I classify an image once I submit the classification the orange already seen banner shows up. I think this shouldn't happen. I found it confusing, and I think other volunteers will. I  think they make it it as the system thinking I classified it before but didn't tell them  ahead of time.

screen shot 2015-07-17 at 9 36 32 am

The orange label is not there when I classify the image to start with

edpaget:
Prevents already seen flag from trigger later in the subject viewer
lifecycle

Closes #1109


aweiksnar:
:+1:

PR #1209, opened 2015-07-24
Convert faked DELETE into a real one

### LINKED ISSUE #1205: Sign out API sessions via DELETE method
From  #1179 - Seems the cloudfront API proxy isn't allowing the "X-HTTP-Method-Override:'DELETE'" header through and probably shouldn't.

[Here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/api/auth.coffee#L242) should use a straight DELETE and not try to replicate Rails form delete submission override. This was probably me being stupid and advising brian to use the delete override method for some reason.

edpaget:
Use X-CSRF-Token header instead of a token in the body

Closes #1205

### DIFF
--- a/app/api/auth.coffee
+++ b/app/api/auth.coffee
@@ -7,10 +7,6 @@ JSON_HEADERS =
   'Content-Type': 'application/json'
   'Accept': 'application/json'
 
-# PhantomJS doesn't send any data with DELETE, so fake it here.
-DELETE_METHOD_OVERRIDE_HEADERS = Object.create JSON_HEADERS
-DELETE_METHOD_OVERRIDE_HEADERS['X-HTTP-Method-Override'] = 'DELETE'
-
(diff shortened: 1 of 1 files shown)


app/api/auth.coffee
```diff
@@ -236,10 +236,11 @@ module.exports = new Model
     @checkCurrent().then (user) =>
       if user?
         @_getAuthToken().then (token) =>
-          data =
-            authenticity_token: token
 
-          makeHTTPRequest 'POST', config.host + '/users/sign_out', data, DELETE_METHOD_OVERRIDE_HEADERS
```
brian-c: Could you remove the `DELETE_METHOD_OVERRIDE_HEADERS` definition at the top?

edpaget: Done

PR #1212, opened 2015-07-24
Include navigation above comments in Talk

### LINKED ISSUE #943: Page count & nav should appear at the top of the thread display too
Often I click on a post in Talk and want to go to another page without having to scroll to the bottom. Mirroring the page navigation at the top of the post displays as well as at the bottom would be really useful.

edpaget:
Closes #943


aweiksnar:
:+1:

PR #1217, opened 2015-07-24
Markdown editor in the Project Lab

edpaget:
Makes a universal editor for markdown.


app/components/markdown-editor.cjsx
```diff
@@ -17,32 +19,125 @@ module.exports = React.createClass
...
+  wrapSelectionIn: (wrapFn, opts = {}) ->
+    # helper to call markdown-insert functions on the textarea
+    # wrapFn takes / returns a string (from ./lib/markdown-insert.cjsx)
+
+    textarea = @refs.textarea.getDOMNode()
+    selection = m.getSelection(textarea)
+    {text, cursor} = wrapFn(selection)
+
+    m.insertAtCursor(text, textarea, cursor, opts)
+    @onInputChange()
```
brian-c: This method doesn't exist. Also on line 143.

brian-c:
I haven't dug into it too much, but it looks like maybe `@props.onChange` isn't being called after using the formatting buttons. I select some text, hit the Bold button, then the Preview button, and I lose the boldness.

The formatting buttons also seem to scroll the text down.

edpaget:
I think I've taken care of that now. It looks like you were right about onChange not firing when you used the buttons to insert formatting.

edpaget:
@aweiksnar I don't know if you also want to look at this since it touches a bunch of talk stuff.

aweiksnar:
Yeah I'll take a peek and make some test comments

app/components/markdown-editor.cjsx
edpaget: Added it back in to handle the onChange events.

aweiksnar:
Only thing I've noticed that isn't working for me is clicking bold/italic/etc submits the comment in talk

app/talk/comment-box.cjsx
```diff
@@ -44,68 +43,31 @@ module?.exports = React.createClass
...
         @setFeedback @props.submitFeedback
       .catch (e) =>
         @setState(error: e.message, loading: false)
 
-  onPreviewClick: (e) ->
-    @toggleComponent('preview')
-
-  onHelpClick: (e) ->
-    @toggleComponent('help')
+  onInputChange: (e) ->
```
aweiksnar: Calling e.preventDefault() here remedies that comment submitting issue

edpaget:
Ugh fix one thing create another problem.

app/talk/comment-box.cjsx
edpaget: It seems like some button is acting as `type=submit` since that shouldn't be necessary.

edpaget: Yeah none of those buttons had a type declared. I added it and it fixed that problem for me.

aweiksnar: Ah nice!

app/talk/comment-box.cjsx
```diff
@@ -159,7 +88,8 @@ module?.exports = React.createClass
             <CommentPreview header={null} content={@state.reply}/>
           </div>}
 
-        <textarea value={@state.content} onChange={@onInputChange} ref="textarea" placeholder={@props.placeholder} />
+        <MarkdownEditor className="full" value={@state.content} onChange={@onInputChange} helpText={<CommentHelp />}/>
```
aweiksnar: Can you add `placeholder={@props.placeholder}` back in here?

aweiksnar:
Cool LGT:shipit: by me

brian-c:
The only thing still stuck for me is the scrolling. I've got it fixed in `insertAtCursor` in markdown-insert.coffee by storing `{scrollLeft, scrollTop} = input` at the start and resetting them at the end, but I'm not sure why it's jumping to the end to begin with.

edpaget:
@brian-c it looks like this is an old bug in Chrome/Webkit. https://code.google.com/p/chromium/issues/detail?id=75072. I've worked around it by reading the `scrollTop` of the input before the focus event and reseting it after focus is called.

PR #1231, opened 2015-07-27
Explicitly set forms with password fields to POST

### LINKED ISSUE #1230: Login form needs default method
To avoid putting usernames and passwords in the page URL if JavaScript fails for some reason.

See zooniverse/Panoptes#1189

brian-c:
Closes #1230, but I'd love to know what's causing this.


edpaget:
The person who reported looked like they had a ton of extensions running.

PR #1243, opened 2015-07-27
Prompt to sign in on click reply if not signed in

### LINKED ISSUE #1126: If you hit reply on a post on Talk when not logged in, it should ask you to log in with a pop-up message
Currently it sends something back to the API and then nothing happens. I think the behavior should be at point to pop up saying, you're not logged in. To join a talk discussion you must be logged in either with the login window or a link to log in.

aweiksnar:
- Closes #1126


chrissnyder:
:shipit: :shipit: :shipit:

PR #1244, opened 2015-07-27
Make subject images & viewer smaller in talk

### LINKED ISSUE #1219: Making the subject smaller in the Talk note discussions and subject pages
I'm finding the subject takes up a lot of room and that you have to scroll to get down to the actual discussion parts of the thread and the subject page. I think it would be really helpful if the subject was made as large as is currently being using in 2.0 and then you could click to get enlarged version (and maybe in the future later a magnifying glass feature)

7767b158-1fde-11e5-9f06-4ee11a8284e5

screen shot 2015-07-25 at 8 59 09 am

Compared to Talk 2.0 

screen shot 2015-07-25 at 9 03 41 am

aweiksnar:
- Closes #1219

### DIFF
--- a/css/subject.styl
+++ b/css/subject.styl
@@ -2,9 +2,10 @@
   min-width: 0
 
 .subject
+  margin: 0 auto
+  max-width: 800px
   .subject-container
-    max-width: 100%
-
+    width: 100%
     img
       max-width: 100%
 
--- a/css/subject.styl
+++ b/css/subject.styl
@@ -1,0 +1,8 @@
+.subject-viewer
+  min-width: 0
+
+.subject
+  .subject-tools
+    button:not(:last-child)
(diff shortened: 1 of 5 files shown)


css/subject.styl
```diff
@@ -2,9 +2,10 @@
   min-width: 0
 
 .subject
+  margin: 0 auto
+  max-width: 800px
   .subject-container
```
chrissnyder: Is there anywhere on the site where the stack goes `.subject > .subject-container`? I think these are orphaned styles.

aweiksnar: Looks like that css class name was used elsewhere, but didn't conflict. I just renamed this instance to `.subject-page` so it doesn't come up in the future

chrissnyder:
:+1:

PR #1245, opened 2015-07-27
Store admin toggle state between refreshes

chrissnyder:
- Plops `params.admin` into localStorage for retrieving.
- Also makes the admin toggle a bit more noticeable, since it's easier to leave it on now.

### DIFF
--- a/app/partials/main-footer.cjsx
+++ b/app/partials/main-footer.cjsx
@@ -29,8 +29,13 @@ counterpart.registerTranslations 'en',
       privacyPolicy: 'Privacy policy'
 
 AdminToggle = React.createClass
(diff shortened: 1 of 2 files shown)


camallen:
Local testing when had the admin box checked, i then unchecked it and then reloaded the page via url hacking / reload button and it reset the state to be checked. Code looks good, not sure what's going on here.

eatyourgreens:
maybe delete the local storage key if the box changes from checked to unchecked?

chrissnyder:
Yea, not sure how it worked for me yesterday sufficiently to open the PR. Should be good now.

camallen:
Nice work :+1:

PR #1246, opened 2015-07-27
Update page state on conversation user changes

### LINKED ISSUE #1202: Getting redirected when you log out from a private message thread from the Talk mailbox
super low on the queue - Currently if you're in a private message thread and log out it keeps you there but just logs you out. I think we might want to redirect to Zooniverse home or Talk or something so that you're not viewing the private thread (if you log out while at the mailbox you page you do get redirected to a page that says you should log in to see the mailbox. Just not in the individual private message threads.)

aweiksnar:
- Closes #1202


chrissnyder:
:+1:

PR #1247, opened 2015-07-27
Update help text subject mentioning syntax options

### LINKED ISSUE #1221: Updating Talk Help for changes to Mentioning subjects
I believe Mention subjects with @owner/project^subject_id --> is not ^subject_id

aweiksnar:
- Closes #1221


chrissnyder:
Is that so we can point them to the project-specific Talk subject page? Pretty gross that we can't just always use `^subject_id`

aweiksnar:
Subjects can be in multiple projects, so mentioning one in zooniverse-wide talk or another project's talk can't guarantee which project to send people to, but if inside of a project that can be grabbed from the url. 

Hopefully I can put a better ui in for this type of mention at some point.

chrissnyder:
Roger that

PR #1248, opened 2015-07-27
Refactor comment links

parrish:
This PR would allow comments to be linked to without including the page number.

### DIFF
--- a/app/talk/comment-link.cjsx
+++ b/app/talk/comment-link.cjsx
@@ -29,7 +29,7 @@ module?.exports = React.createClass
 
   setDiscussion: ->
     {comment} = @props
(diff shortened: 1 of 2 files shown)


parrish:
The idea is that by not needing to link to the comment page, you'll only need project/owner info and comment/discussion ids to build the url.

That feels a little more maintainable and cuts down on the number of calls necessary for pages like search.

aweiksnar:
:+1: 
Yeah, this will cut down on pain with building links to comments. Thanks!

DarrenMcRoy:
Nice one, guys. That sounds super useful given how many places we want to have comments display. :+1:

PR #1249, opened 2015-07-27
Advance classifier to next image after navigating

### LINKED ISSUE #1239: "Already seen" banner appears on return from Talk
Perhaps this is as intended, but it confused a user:
* User classifies a subject
* User clicks Talk to comment on the subject
* User clicks browser 'back' button to go back to classifying
* User sees the image again at the "done" stage (with the classification summary and the 'Talk' and 'Next' buttons), but the image has an "Already seen" banner on it

The presence of the "Already seen" banner confused the user, as she was expecting to come back to a page that looked the same as the one she left.

edpaget:
..to Talk

Closes #1239

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -122,7 +122,7 @@ Classifier = React.createClass
 
       <nav className="task-nav">
         {if @props.owner? and @props.project?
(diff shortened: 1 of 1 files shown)


aweiksnar:
:checkered_flag:
Liking this for less clicks too

PR #1261, opened 2015-07-28
Fix comment id check

aweiksnar:
@parrish I missed this in #1248


parrish:
Is it a problem because of the style?  (assignments in conditionals probably isn't great)

aweiksnar:
Good point, I guess it'll work either way. Been debugging some links stuff today and got hung up reading that block for some reason so figured I'd PR a change

parrish:
No worries, it's more readable this way

Issue #1020, opened 2015-07-09
What to do for a subject's page?


chrissnyder:
- Related to #898, #914, #916, amongst others.

So I was going to being to tackle #898 but I ran into some conceptual problems that I hope the group can help with.

Alex built a a subject's page for Talk (ex. https://www.zooniverse.org/#/projects/aliburchard/wildebeest-watch/talk/subjects/463145), but the issue with using it for #898 is that it's scoped to a single project. It also has project-specific stuff like adding a quick note (which is a good thing).

My question to group is are we OK with me making a subjects page, at something like `#/subjects/463145` and have it pull in as much data as it can from any project that it is a part of, and then like to that page from a collection? I'm not sure what use it would be besides giving me a place to link a subject to from a collection.

Alternatives would be

- Opening the subject image in a new tab for viewing and nothing more
- Making the link goto the uploaded project's Talk subject page.

Not sure I like either of those, but wanted to put this out there for discussion.

@mschwamb, @aliburchard, @DarrenMcRoy, @DarrenMcRoy, @edpaget, etc. thoughts?

mschwamb:
I think collections, hashtags, and conversations on Talk should be gathered onto a single page just like talk 2.0 where if you go there you can see everything going on about it - It's been the most handy thing about Talk 1.0 and Talk 2.0 from a science team perspective. I've used hashtags and people's collections to do data validation and checks on Planet Four data analysis and being able to navigate through Talk has been handy because I also see people's comments. 

So I think having a separate page for collections for a subject that's different from it's Talk subject page is basically either duplicating info or cutting people off from Talk. If the page opens in a new tab I think it would be okay because you can still navigate through you collection if you choose and it's get you into Talk

mschwamb:
I think it's okay to go to the project's Talk page even if there are mixed subjects in the collection (probably as long as it opens in a new tab). I think it makes the focus about discussion and interaction - Especially if the magnifier feature that's in P4 Talk and some of the projects is implemented, I think volunteers will be happy to be redirected to Talk and it might engage them more in discussion. 

Personal opinion here, but I think this is something Talk 2.0 did right in terms of layout - I'd really like to see this duplicated or similar very similar on panoptes Talk for the subjects page.

![screen shot 2015-07-10 at 12 57 42 am](https://cloud.githubusercontent.com/assets/753325/8601467/d743774e-269e-11e5-880c-c6b77a93fcd3.png)

[...]

PR #1263, opened 2015-07-28
Link to subjects when in project and set as focus

### LINKED ISSUE #916: Adding a link to the subject pages from the notes discussions
The discussion page is keeping an index of all discussions started from clicking talk in the interface. Currently in the notes threads you only see that conversation, so there's no way to know there are other discussions. It would be helpful if there's a way to get back to the subjects page - perhaps making the title of the subject into a link

The only way currently to see the subjects page is to manually type it
Being able to access the subject page from a link from a discussion  would allow science teams to quickly see all the discussions going on about a subject

screen shot 2015-07-01 at 10 45 56 am

aweiksnar:
- Closes #916
- Right now only from projects ~ #1020
- Can be seen on demo at
  - `#/projects/alex-panoptes/test-proj/talk/42/51`
  - `#/talk/20/82`
- I've got another branch up called `link-to-subjects-in-discussion-titles` where the links happen in titles, but I think this is a bit cleaner

### DIFF
--- a/app/talk/comment.cjsx
+++ b/app/talk/comment.cjsx
@@ -16,6 +16,7 @@ talkClient = require '../api/talk'
 Avatar = require '../partials/avatar'
 SubjectViewer = require '../components/subject-viewer'
 DisplayRoles = require './lib/display-roles'
+merge = require 'lodash.merge'
 
 DEFAULT_AVATAR = './assets/simple-avatar.jpg'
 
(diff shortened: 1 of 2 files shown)


DarrenMcRoy:
Does this add a link from a Notes thread, as well as from anywhere else that a subject is linked/featured?

aweiksnar:
It does this anywhere a subject is linked/featured in a project. Link is pointed to subject page
screen shot

DarrenMcRoy:
:+1:

chrissnyder:
So it's clear, `#/talk/20/82` should or shouldn't have a link? http://i.imgur.com/cadg4OX.png

aweiksnar:
@chrissnyder That shouldn't have a link, goes back to the question of not assuming that a subject belongs to a specific project since they can be in multiple

DarrenMcRoy:
My proposed answer to that is to always link to the project where the subject was initially uploaded.

PR #1274, opened 2015-07-29
Use a generic MediaArea component for projects

brian-c:
This is phase one of the survey task editor. Surveys have images, and we needed a decent way to manage 'em, so this is a generic interface for working with resources' `linked_images`.

Here it replaces the project lab's Media page.

The Thumbnail component uses our cool little thumbnail auto-generator.

The FileButton component is like a `<select type="file" />`, but it passes its value straight into `@props.onSelect` instead of storing it.

I don't think the FileDropTarget is quite square yet, but once it is we can use it to simplify a few other components.

I can see the Thumbnail component being extended in the future to support Media resources and Subjects directly. To-do for later.

### DIFF
--- a/app/components/dropdown-form.cjsx
+++ b/app/components/dropdown-form.cjsx
@@ -18,6 +18,11 @@ DropdownForm = React.createClass
     right: 0
     top: 0
 
+  pointerStyle:
+    bottom: '100%'
+    position: 'absolute'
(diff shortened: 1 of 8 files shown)


aweiksnar:
:moneybag:  Looks really clean. I made some uploads and everything went smooth. Psyched for thumbnails too

PR #1310, opened 2015-07-31
Close polygon tool when clicking next or done

### LINKED ISSUE #1118: Enforce closure of polygons
Currently the polygon tool acts like a polyline and polygon tool at the same time. Behaviour is dependent on the closure of the line segments.

Although this adds flexibility it would be great to have polygon closure enforceable.

Currently, people can move on without closing a polygon, which might generate errors. More so as the polygon in subsequent workflows will not be rendered as a polygon but as an (open) polyline. Consequently the last vertices would not intersect with the subject, raising no alarm on part of the user.

Being able to enforce polygon closure before moving to a next task would resolve this issue.

edpaget:
Closes #1118

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -87,9 +87,9 @@ Classifier = React.createClass
               <nav className="task-nav">
                 <button type="button" className="back minor-button" disabled={onFirstAnnotation} onClick={@destroyCurrentAnnotation}>Back</button>
                 {if nextTaskKey
-                  <button type="button" className="continue major-button" disabled={waitingForAnswer} onClick={@addAnnotationForTask.bind this, nextTaskKey}>Next</button>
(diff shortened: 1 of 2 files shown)


brian-c:
I'm still getting `closed: false` on my polygon when I switch tools.

edpaget:
I thought the switch tools case was already handled by existing code and this issue was only for if you went to the next task or finished a classification.

edpaget:
I'll look into that though since I probably broke something.

brian-c:
AHHH sorry, I mistyped. I meant when I switched tasks.

brian-c:
Also I think most of the drawing-specific code directly in the classifier is a big nasty hack, and I'd like to reorganize everything in the future with some kind of task-defined on-start, on-leave system. Feel free to totally gut this thing if you're up for it.

edpaget:
Yeah looks like that'll be the only way for it.

chrislintott:
Please can we record when the tool was closed this way?

PR #1311, opened 2015-07-31
Set default avatar on error loading avatar

### LINKED ISSUE #967: Why is this user's avatar not working?
https://www.zooniverse.org/#/talk/13/102?page=1&comment=283

What's with the error message where this user's avatar should be?

edpaget:
Closes #967

### DIFF
--- a/app/partials/avatar.cjsx
+++ b/app/partials/avatar.cjsx
@@ -38,5 +38,8 @@ module?.exports = React.createClass
       .then =>
         @setState loading: false
 
+  handleError: ->
+    @setState src: DEFAULT_AVATAR
+
   render: ->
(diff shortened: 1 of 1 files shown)


edpaget:
I broken @aweiksnar's avatar on staging so you can look at it. I'll restore it after this is merged.

aweiksnar:
Ha I noticed that - username is `alex-panoptes` btw

brian-c:
Looks good to me.

DarrenMcRoy:
Does this cover all types of loading errors? It looked in #967 like there were a few different ways that avatars could go wrong.

edpaget:
yes if it fails to load an image it will replace it with the default image.

PR #1312, opened 2015-07-31
New api root

brian-c:
Changes the API subdomain and the page from which to scrape the CSRF token.

Up at https://www.zooniverse.org/temp for now.


astopy:
@brian-c I get "Failed to get auth token" in the console when I try to sign out.

edpaget:
I am unable to get any avatar images (the responses are empty when my browser requests them) on Firefox and Chrome on Arch Linux.

frankvitaleiv:
On the publication page, I am seeing the first publication for each project on the same line as the project name when my browser screen is not as wide as the descriptions are long.  

screen shot 2015-08-03 at 15 25 03
screen shot 2015-08-03 at 15 25 10

astopy:
@frankvitaleiv That's actually a problem with the live version of the site, not specific to this version. Could you open a new issue for that?

brian-c:
@astopy @edpaget I can't reproduce either of these issues. You're sure you're on https://www.zooniverse.org/temp/? Any weird browser settings?

astopy:
I'm concerned that both of those might be due to CloudFront caching, even though it shouldn't be caching those requests as far as I can see. I'll try and investigate further.

frankvitaleiv:
@astopy sure, no problem

edpaget:
@brian-c @adam I can confirm it's happening on the zooniverse.org/#/temp page. I have a bunch of extensions installed on Firefox, but Chrome is entirely stock. In my network tab, I see requests to https://panoptes-uploads.zooniverse.org/..., but with no responses.

camallen:
@edpaget change the URL to https://www.zooniverse.org/temp/#/ ?

edpaget:
@camallen I was already on that I'm fairly sure (zooniverse.org is redirected to www. right?), but all of sudden everything is working.

camallen:
Cool - i couldn't reproduce the avatar issue you were seeing but glad it's working now.

astopy:
The auth token issue on sign out appears to be specific to safari... can anyone test it in that?

eatyourgreens:
@astopy I can reproduce that auth token issue in Safari 7.1.7

brian-c:
I think this is because of the /users/sign_in to panoptes.zooniverse.org redirect when someone's already signed in.

Is there anywhere I can get a CSRF token from another domain that never redirects somewhere else?

astopy:
Would it make sense to just add a dedicated /csrf route to to panoptes, rather than relying scraping the HTML?

brian-c:
That would be sweet.

edpaget:
Also I figured out why I couldn't see avatar when testing this. I had redirected panoptes-uploads.zooniverse.org to 127.0.0.1 in my /etc/hosts file to test something out.

astopy:
OK, I think this is good to merge. Everyone happy with that?

brian-c:
Fine by me.

PR #1313, opened 2015-07-31
Return empty string if task help is blank

### LINKED ISSUE #1090: Help box doesn't clear for tasks without help text
In a workflow, I have multiple tasks. If I click on Task A, which has help text, that help text shows up in the help text box. If I then click on Task B, which does not have help text, the help text for Task A remains in the help text box.

edpaget:
Make sure the help text for a task is refreshed when the selected task
changes since a undefined property doesn't cause React to rerender

Closes #1090


edpaget:
I'm pretty this issue would only come up when someone is creating tasks outside of the task editor since the editor will create a task with an empty string as its help text which should be enough to prevent this bug from happening.

brian-c:
Yep, tricky. Good find.

PR #1315, opened 2015-07-31
Survey task editor

brian-c:
This is (once again) a total mess of so much stuff. Sorry. No rush.

Roughly in the order of the files below:

1. Split out the NextTaskSelector into its own component.

2. Make the delete task button part of the workflow editor instead of part of the task.

3. **Add a survey task editor** (import from CSVs only for now, only exposed to projects with `admin: true` owners).

4. Add a survey task summary.

5. Add a simple `<Details summary="">` component (like `<details><summary>...`, but cross-browser).

6. Improve dropdown positioning so they stay within the sides of the window.

7. Add a `metadata` filter prop to the MediaArea component. That way even though the images are associated with the workflow, we can only show the ones relevant to the task.


aweiksnar:
Fancy! I was able to upload the csvs. Data saved correctly (raw task data was helpful here) and viewable in the ui too.
Thoughts on changing the `choice.images` instances to `choice.images?etc`? I noticed that without images defined it can hang in the loading state and logs that as an undefined error in the console

brian-c:
Sorry @aweiksnar, I just saw your comment.

You might've been getting errors from the thing I just fixed, which was trying to render the wrong image addresses. Otherwise I think all uses of `choice.images` are preceded by `unless choice.images.length is 0`, unless I'm misunderstanding. Are you running into a case where `choice.images` doesn't exist?

aweiksnar:
Yeah it looks like I have a data error in the csv: one of my tasks didn't have `images` defined, but saved ok: 
```json
"DLPHN": {
  "image": [],
  "label": "Dolphin",
  "confusions": {},
  "description": "",
  "characteristics": {
    "CLR": [
      "RNG"
    ]
  },
  "confusionsOrder": []
}
```

brian-c:
Oh weird, you have a choice with an `image` instead of `images`. Is this a new task? Can you send me the CSVs you're using?

aweiksnar:
Unfortunately I've already `rm -rf`'d them. The other 2 choices have the  `images` key saved with an array of images as values, in addition to the `image` key of an empty array.

PR #1316, opened 2015-07-31
Set a flex-basis on dialog-content

### LINKED ISSUE #1191: help window width IE
The window width for the help box in internet explorer is crazy wide - full window width. This has caused some testers to try to use the back button to exit the help as it looks like a full page - obviously this screws things up.

This was based on a user with windows 7...
I've replicated this on windows 8 version 11 - a very common combination

### LINKED ISSUE #1296: Dialog Controls overlap scroll bar

edpaget:
...to restrict dialog width in Internet Explorer

Currently IE11 and below don't corrently implement flex-basis auto or
flex-basis content, so flexed content will always expand to fill the
largest space possible unless an explicity flex-basis is set.

This just sets flex-basis for the dialog-container to 800px, which seems
like a reasonable maximum.

Closes #1296; Closes #1191

### DIFF
--- a/css/dialog.styl
+++ b/css/dialog.styl
@@ -16,7 +16,6 @@
   background: white
   box-shadow: 0 1px 3px -1px black
   color: black
-  display: flex
   flex-direction: column
   left: 50%
   max-height: 90%
@@ -48,6 +47,8 @@
 
 .dialog-content
   overflow: auto
+  display: flex
+  flex: 1 0 800px
 
   img
     max-width: 100%


edpaget:
Internet Explorer is the worst.

brian-c:
Sorry, this doesn't seem to change anything for me, except now scrolling scrolls the close button too.

Story time. Dialogs are styled real tricky: Goals were:

1. The dialog wraps to fit its content. It does not make any assumptions about its content. (It's `position: absolute`, so that's automatic anyway.)

2. The dialog doesn't make any assumptions about the size of the viewport, but it never exceeds the edges. (Also easy, just set `max-width` and `-height`.)

3. The close box is visible independent of dialog scroll position.

In order for the close button to stay fixed within the dialog, the dialog can't be `scroll: auto`, so the content has to go in a container. The container can't size itself to the dialog because the dialog is trying to size itself to the container's contents at the same time.

So I set the dialog up as with `display: flex; flex-direction: column` and gave the `.dialog-controls` container a height of 0 so the whole space of the dialog is taken up by the `.dialog-content` container. Since the `.dialog-content` now has a definite size (and the dialog respects it), we can set `overflow: auto` and it'll scroll when it needs to. The close button won't scroll because it's a sibling.

`position: sticky` was actually made for stuff like this, but Chrome actually removed support for it. Awesome.

By removing `display: flex` from the dialog we lose the sticky close box. If we'd've left that, then adding `flex-basis: 800px` to the content actually limits its height, not its width, because its flex parent's direction is `columns` (but because we removed `display: flex` from the parent, it doesn't actually do anything).

In my opinion, the right fix for this would be to modify the **content** to be a readable width, not the dialog. A dialog is only responsible for being a dialog-like, not changing how its contents render. I'd maybe stick an inline `style={maxWidth: '60ch'}` on the help's `.content-container`.

Also, weird thing: Chrome is the one being weird here by adding some kinda implicit `right: 0` to the dialog and limiting its' width to 50%. IE is behaving as expected, letting it reach its `max-width` of 100%.

Aside: I just discovered CSS's `ch`aracter units and they're totally awesome and appropriate for setting line lengths.

edpaget:
Sure I'll try something something else. We can just set a max-width on markdown content  inside the dialog?

PR #1325, opened 2015-08-04
Update contact-page.cjsx

trouille:
- Removed the collaboration form and inputted an email address collab@zooniverse.org that will go to Chris Lintott and Laura Trouille directly instead. 
- For press inquires, added Laura and removed Darren.


chrissnyder:
:+1:

PR #1326, opened 2015-08-04
Optimizations

parrish:
Removes comment page calculations since comment links only require the id.
Uses embedded associations in lieu of extra API calls.

@aweiksnar Would you mind checking this out?


aweiksnar:
Looks fantastic - Only request I'd make is that the board & discussion previews continue to link to the page of the last comment and scroll to the bottom instead of highlighting the comment as if linked to. Should just be in latest-comment-link

parrish:
Hmm... I don't disagree, though I think we should probably not be linking to the last comment in every discussion preview.  For instance, the search page has a more obvious link to the last comment than the result comment.

Maybe discussion preview could accept props like `latest={comment}` or `comment={comment}` to get the right behavior.

aweiksnar:
Yeah that'd work for me. Largely for the board/discussion pages

parrish:
@aweiksnar This should be good to go.  Let me know what you think

aweiksnar:
The discussions and board updates look great. I think search results pages may need the discussion to pull the latest comment - Right now it's throwing an error

parrish:
Ah, my mistake -- the merge overwrote some changes...

parrish:
Okay, I think everything's good again

aweiksnar:
Awesome, Yeah looks really solid. merging...

PR #1329, opened 2015-08-04
Update contact-page.cjsx

trouille:
there were funny line breaks that I fixed


chrissnyder:
Could the same treatment be given to every line on that page? Looks funky only removing the CRs from those two lines http://i.imgur.com/TrPYcwd.png

PR #1330, opened 2015-08-04
Replace markdown components with markdownz library

edpaget:
This is our attempt to move some functionality into self-contained modules. 

You can find the Markdown module here: https://github.com/zooniverse-ui/markdownz


brian-c:
This works. The amount of boilerplate in the module itself is terrifying though. I'm seriously not looking forward to maintaining 20 copies of that.

I'll try breaking out the DropdownForm module (maybe split in two, as DropdownButton and ModalForm) and maybe we can compare?

edpaget:
Sorry, what do you mean by boilerplate?

brian-c:
I'm just mainly anxious about all the Gulp/Babel configuration and all those dependencies and the test harness and all its dependencies (which I'm finding there's really no way around). That seems a ton of stuff to keep up to date and in sync. Might end up not being an issue, like you said.

PR #1339, opened 2015-08-05
Fix task area at 50ch

brian-c:
Make the task area fixed width, because survey tasks show up in a three-across grid, and we can't have the width shifting around randomly between tasks.

cc: @aliburchard

### DIFF
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -76,10 +76,9 @@
     transform: rotate(-15deg)
 
   > .task-area
-    // We want this fixed at 50h, but don't completely disable flex-shrink because
-    // when it wraps to its own line it shouldn't be cropped by small windows.
-    flex: 0 0.01 50ch
+    flex: 0 0 50ch
     margin-bottom: 1em
+    max-width: 100%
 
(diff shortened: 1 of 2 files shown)


css/classify.styl
```diff
@@ -76,10 +76,10 @@
     transform: rotate(-15deg)
 
   > .task-area
-    flex: 1 1
+    // We want this fixed at 50h, but don't completely disable flex-shrink because
+    // when it wraps to its own line it shouldn't be cropped by small windows.
+    flex: 0 0.01 50ch
     margin-bottom: 1em
-    max-width: 60ch
```
aweiksnar: Looks good in chrome, but in firefox it's not shrinking to small windows. Perhaps there's a way to handle that with flexbox, otherwise adding something like this seemed to shrink it pretty well
```styl
    @media (max-width: 50ch)
      flex: none
      width: 100%
```

brian-c:
Good catch. That commit should do it.

aweiksnar:
Yep that did it :shipit:

PR #1340, opened 2015-08-05
Break out DropdownForm into its own module

brian-c:
Here's my stab at pulling out a mini module. https://github.com/zooniverse-ui/modal-form

Zuul and Sauce are pretty much exactly how I've always wanted to run tests. I wrote them in tape because I'm used to it, but I can switch to mocha. Anyway, they're super sloppy, and no Travis yet.

### DIFF
--- a/app/components/dropdown-form.cjsx
+++ b/app/components/dropdown-form.cjsx
@@ -1,84 +1,5 @@
 React = require 'react'
-
-ESC_KEY = 27
-
-DropdownForm = React.createClass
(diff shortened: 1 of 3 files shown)


edpaget:
That looks great. I don't think we need to worry about all using the same transplitation (or lack thereof) or testing libraries for each component. Tape looks pretty straightforward, and es5/es6/coffeescript are all similar enough we can switch between them depending on what we're doing.

PR #1364, opened 2015-08-07
Delegate annotation completion check to its task

brian-c:
This begins a big cleanup and refactor of the classification interface. The eventual goal is for the classifier to become a dumb subject-and-workflow display with a bunch of hooks for tasks to be able to be able to inject appropriate stuff where they need it.

This PR makes each task responsible for knowing when one of its annotations is complete. In the process I added support for the `required` property for surveys and drawing tasks, and drawing details subtasks can be marked required too. None of this is exposed in the task editor yet, since the volunteer-facing interface will need some work to make it obvious that a task isn't complete unless you do _x_.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -54,6 +54,7 @@ Classifier = React.createClass
         @props.subject.expert_classification_data
       else
         @props.classification
+      window.classification = currentClassification
 
       if currentClassification is @props.classification and not @props.classification.completed
         currentAnnotation = currentClassification.annotations[currentClassification.annotations.length - 1]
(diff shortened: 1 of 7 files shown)


app/classifier/index.cjsx
```diff
@@ -1,105 +1,119 @@
...
+      if classification.annotations.length is 0
+        @addAnnotationForTask @props.workflow.first_task
+
   render: ->
     <ChangeListener target={@props.classification}>{=>
       currentClassification = if @state.showingExpertClassification
         @props.subject.expert_classification_data
       else
         @props.classification
+      window.classification = currentClassification
```
aweiksnar: Was this just for dev?

brian-c: Yep, sorry. Deleted.

aweiksnar:
Cool, Working well for me. Tested on a few different projects/tasks. merging...

PR #1372, opened 2015-08-10
Allow discussions to be moved

### LINKED ISSUE #843: Ability to move threads in Talk
Currently not an option that I can see.

Merging threads will be an enhancement for later (mods like that), but we should have thread-moving capability sooner rather than later, especially if it's not too tricky?

aweiksnar:
- Closes #843


DarrenMcRoy:
YAYYY!!

chrissnyder:
:+1:

PR #1373, opened 2015-08-10
Break apart survey, use annotator hooks

brian-c:
- Breaks the components of the survey task out into separate files.

- Removes knowledge of the survey task from the subject annotator and instead provides a couple slots (Before and AfterSubject for now) for any task to plug itself into the interface.

I'm gonna see what I can do with the drawing task next. The longer-term goal here is to reduce the responsibility of the classifier and subject annotator so it's easier to create new types of tasks and tasks can do any weird stuff they need to without polluting the generic classification components.


aweiksnar:
Lookin good! I think breaking that up was totally worth it

PR #1377, opened 2015-08-10
Update contact-page.cjsx

trouille:
Additional line breaks fixed so it's all same format. 
(Can't wait for Alex to help me set my mac up so I do this via Emacs and not in Github directly. Whoohoo!)


chrissnyder:
:100:

PR #1380, opened 2015-08-10
Update boards on user changes

### LINKED ISSUE #1375: If you are already on the project talk page and then log in hidden boards that require moderator or team access don't appear automatically
We have a mod+team only board on Planet Four: Terrains Talks. I just noticed that if I'm on the Talk front page and the log in the front page doesn't refresh to reveal the board. I have to refresh the page for the mod board to show up. The page should automatically refresh when you log in so that all boards are shown. 

Also on the reverse if I log out on the front page of the project talk, the hidden board remains visible. I think it should disappear once you log out since you don't have permissions anymore to click on its content

aweiksnar:
- Closes  #1375

### DIFF
--- a/app/talk/init.cjsx
+++ b/app/talk/init.cjsx
@@ -24,6 +24,7 @@ module?.exports = React.createClass
 
   propChangeHandlers:
     'section': 'setBoards'
+    'user': 'setBoards'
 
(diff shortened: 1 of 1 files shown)


chrissnyder:
:+1:

Issue #1338, opened 2015-08-05
Not all thumbnails are generating on the Notes board


DarrenMcRoy:
Looking through the Planet Four: Terrains Notes and the Whales as Individuals board, not all thumbnails are generating:

https://www.zooniverse.org/#/projects/mschwamb/planet-four-terrains/talk/29
https://www.zooniverse.org/#/projects/tedcheese/whales-as-individuals/talk/31

This is true for all live projects, actually.

There are a few subjects that just aren't generating thumbnails properly. Any idea?

EDIT: It's particularly bad on Season Spotter projects.

https://www.zooniverse.org/#/projects/kosmala/season-spotter-questions/talk/62?page=1
https://www.zooniverse.org/#/projects/kosmala/season-spotter-image-marking/talk/66

mschwamb:
Does it have something to do with replies. That once someone else has replied on that thread the thumbnail then disappears? It all the cases on Planet Four is appears to be when someone has replied/posted in the notes thread beyond the initial post that created that the thread.

mschwamb:
@aweiksnar Following the above, I was able to replicate the issue. It's something to do with replies. Once a reply/posting after the initial creation of the Talk notes post. Once a second post comes in, the update on the threads page removes the thumbnail.

vrooje:
Same on GZ Bars - only the Notes threads with 1 post have thumbnails.

mschwamb:
The patch for this looks great. I'm seeing thumbnails with replies.

aweiksnar:
Great! thanks for testing @mschwamb

PR #1398, opened 2015-08-12
Get thumbs from first comment instead of latest

aweiksnar:
- This got changed at some point, Since the first comment is
  the one that sets the focus of the discussion this pulls the thumb
  from that
- #1338

### DIFF
--- a/app/talk/discussion-preview.cjsx
+++ b/app/talk/discussion-preview.cjsx
@@ -21,13 +21,14 @@ module?.exports = React.createClass
 
     <div className="talk-discussion-preview">
       <div className="preview-content">
-        {if comment.focus_id and (comment.focus_type is 'Subject')
+
+        {if discussion.focus_id and (discussion.focus_type is 'Subject')
           <div className="subject-preview">
-            <PromiseRenderer catch={null} promise={apiClient.type('subjects').get(comment.focus_id)}>{(subject) =>
+            <PromiseRenderer catch={null} promise={apiClient.type('subjects').get(discussion.focus_id)}>{(subject) =>
               <Thumbnail src={getSubjectLocation(subject).src} width={100} />
             }</PromiseRenderer>
           </div>
-        }
+          }
 
         <h1>
           {<i className="fa fa-thumb-tack talk-sticky-pin"></i> if discussion.sticky}


chrissnyder:
This was changed in the last commit on the file at https://github.com/zooniverse/Panoptes-Front-End/commit/467fd4aff98216e1985c5b37c43f1f6a2b225f6f. Without looking further, is there conflicting goals here?

aweiksnar:
I think the desired behavior for thumbnails should be the image of the focus of the first comment instead of the last. The last commit changed it to generate thumbs from the latest comment, which I believe is causing the discussions not to always have thumbs if there are replies without foci

chrissnyder:
Sounds reasonable to me. Pinging @parrish to make sure there wasn't a different consideration going on, but it looks good to merge to me.

parrish:
When display discussion previews on the subject_default board, you'd probably want the focus that the discussion was started on.

Everywhere else -- search results, other boards, etc., etc. -- you'll want to show the focus from the last comment if it exists (you're showing a preview of the last comment in the discussion, right?)

DarrenMcRoy:
No, we only want these thumbnails for the Notes board right now... which I assume is what you're calling the "subject_default" board, right?

parrish:
Yep, "Notes" is what the subject default board is generally being called.

The thumbnail will show up everywhere the discussion preview is used

aweiksnar:
It seems like the safest assumption to make is that a discussion's first comment focuses on a subject then that discussion is about it, for any board

parrish:
My proposal is:

If a discussion is started with a comment about a focus, then the discussion should store that focus.

That gets us away from N+1 queries as well as just generally making sense I think

aweiksnar:
@parrish I like that solution

parrish:
Cool, ticketed in zooniverse/Talk-Api#59

aweiksnar:
:ok_hand:

parrish:
Blah, The code is ready to go, but dockerhub is broken after their upgrade.  I'll try to get this out tomorrow.

aweiksnar:
Ok cool, I'm happy to leave this one open until it's ready and then I'll update

aweiksnar:
Just updated to use the new discussion focus id/type columns

parrish:
Looks good, but I don't think wet need this linked sorting anymore: https://github.com/zooniverse/Panoptes-Front-End/commit/fa601bbc32de148e7e9af252b04d8ee58f69fde0#diff-c56c806db68922bcecc89dda1372778bL46

aweiksnar:
Oh yep just removed the sort param too

parrish:
Awesome.  LGTM :+1:

chrissnyder:
Cool, merging.

PR #1404, opened 2015-08-13
Retinas crop tool

eatyourgreens:
@brian-c @camallen @aliburchard 
This adds 'crop', as a rectangle tool option, to the drawing tools dropdown, then changes the viewBox if a crop rectangle annotation exists when you start a new task.

Integration with the classifier is still a big hack, but this might be good enough to get the Retinas project going on PFE.

### DIFF
--- a/app/classifier/tasks/generic-editor.cjsx
+++ b/app/classifier/tasks/generic-editor.cjsx
@@ -56,7 +56,7 @@ module.exports = React.createClass
 
     <div className="workflow-task-editor #{@props.task.type}">
       <div>
-        <AutoSave resource={@props.workflow}>
+        <AutoSave resource={@props.workflow} tag="label">
           <span className="form-label">Main text</span>
           <br />
           <textarea name="#{@props.taskPrefix}.#{mainTextKey}" value={@props.task[mainTextKey]} className="standard-input full" onChange={handleChange} />
@@ -67,7 +67,7 @@ module.exports = React.createClass
 
       {unless @props.isSubtask
         <div>
-          <AutoSave resource={@props.workflow}>
+          <AutoSave resource={@props.workflow} tag="label">
             <span className="form-label">Help text</span>
             <br />
             <textarea  name="#{@props.taskPrefix}.help" value={@props.task.help ? ""} rows="7" className="standard-input full" onChange={handleChange} />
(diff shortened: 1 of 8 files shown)


brian-c:
Nicely isolated. Can we block the tool from the [list in the editor](https://github.com/zooniverse/Panoptes-Front-End/blob/retinas/app/classifier/tasks/generic-editor.cjsx#L95-L96), maybe showing it [if the project owner is an admin](https://github.com/zooniverse/Panoptes-Front-End/blob/retinas/app/pages/lab/workflow.cjsx#L108-L119)?

eatyourgreens:
I've made a start on checking the owner. Looks like staging isn't returning `owner.admin` for Ali though. I'll have another look at this tomorrow.

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -7,6 +7,34 @@ alert = require '../../lib/alert'
 DrawingTaskDetailsEditor = require './drawing-task-details-editor'
 NextTaskSelector = require './next-task-selector'
 
+DrawingToolsEditor = React.createClass
```
brian-c: Why's this piece its own module? Can't we just just show or hide the `<option>`?

If it's got to be its own module, can we include all the drawing task stuff?

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -88,15 +116,7 @@ module.exports = React.createClass
...
-                  [<div key="type" className="workflow-choice-setting">
-                    <AutoSave resource={@props.workflow}>
-                      Type{' '}
-                      <select name="#{@props.taskPrefix}.#{choicesKey}.#{index}.type" value={choice.type} onChange={handleChange}>
-                        {for toolKey of drawingTools
-                          <option key={toolKey} value={toolKey}>{toolKey}</option>}
-                      </select>
-                    </AutoSave>
-                  </div>
+                  [<DrawingToolsEditor project={@props.project} workflow={@props.workflow} name="#{@props.taskPrefix}.#{choicesKey}.#{index}.type" value={choice.type} onChange={handleChange} />
```
brian-c: Components in arrays need a `key` prop.

PR #1409, opened 2015-08-13
Set comments meta on comment submit

### LINKED ISSUE #1277: Posting a comment that starts a new page in a discussion doesn't take you to that page
If you post a comment that starts a new page in a Talk discussion, you aren't taken to that new page.

I.e., if, on a one-page discussion, I post the comment that makes it a two-page discussion (the first comment at the top of the second page), I am still on the first page after I submit, which makes it look like the comment was never posted, even though it was.

aweiksnar:
- Closes #1277

### DIFF
--- a/app/talk/discussion.cjsx
+++ b/app/talk/discussion.cjsx
@@ -74,6 +74,11 @@ module?.exports = React.createClass
             @scrollToBottomOfDiscussion()
             @shouldScrollToBottom = false
 
+  setCommentsMeta: (page = @props.query?.page) ->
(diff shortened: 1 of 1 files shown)


aweiksnar:
Easy way to test this is to make the discussion `page_size: 1` and then add a comment

chrissnyder:
That's a subtle bug indeed.

DarrenMcRoy:
I imagine finding and fixing the subtle ones is really satisfying.

PR #1423, opened 2015-08-17
Fix choice thumbnails loading, crop to the same size

brian-c:
The survey task previously loaded its choices' thumbnails in `<img` tags, set to `display: none` when not shown. They loaded anyway, for whatever reason. It ends up being like 30MB on a 50-choice project.

This switches them to be `background-image`s with a `display: none` parent, as described [here](http://timkadlec.com/2012/04/media-query-asset-downloading-results/). so they only load when they're viewed. This also allows up to crop them to all be the same size.

Used on #/projects/brian-testing/ghosts/classify for dev.


aweiksnar:
Nice! :+1:

PR #1424, opened 2015-08-17
Separate app frame and router from main

brian-c:
This is the first part of the bit I did extending the Markdown renderer. I'd still like to keep the router and app frame separated.


edpaget:
LGTM

PR #1322, opened 2015-08-03
fix line spacing issue on publications page, and add some missing links to the sitemap

alexbfree:
i noticed the line spacing is messed up on the publications page. this should fix it.

PR #1425, opened 2015-08-17
Spacing pr

edpaget:
I cleaned up Alex's git history from #1322


alexbfree:
Thanks @edpaget . I still need to learn how to clean up history safely!

edpaget:
It's always safe if you're on a separate branch, and if you mess up you can always use the [reflog](http://git-scm.com/docs/git-reflog).

PR #1428, opened 2015-08-18
Prepend replies to text instead of having separate

### LINKED ISSUE #888: Select how much of a post you want to quote in a reply
Currently, the only thing you can do as far as replying to a post is to quote the entire post in your reply, in one huge chunk, at the very beginning.

This is cumbersome and limiting, especially if you want to reply to an extremely long post, or if you want to reply to a post in bits and pieces, or if you want to reply to multiple posts.

Most forums manage this by having the "reply" button add code inline to your post, code that you can then manage, crop, break apart, etc. ... any chance we could have our system work similarly?

aweiksnar:
- Closes #888

### DIFF
--- a/app/talk/comment-box.cjsx
+++ b/app/talk/comment-box.cjsx
@@ -27,7 +27,7 @@ module?.exports = React.createClass
     header: "Add to the discussion +"
     placeholder: "Type your comment here"
     submitFeedback: "Comment Successfully Submitted"
-    content: null
+    content: ''
     subject: null
     user: null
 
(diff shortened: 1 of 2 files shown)


DarrenMcRoy:
Question: does this include an automatic permalink to the quoted comment? Would be good for it to.

aweiksnar:
Not in this PR, I'll put in an issue to add that to the api though. I believe this will solve the problem of people wanting to abridge replies  though.

DarrenMcRoy:
Okay. It's important for navigation purposes. Thanks! :+1:

edpaget:
When I hit submit after editing the quoted text, I see the original text reappear in the box. It saves correctly, but the behaviour is odd.

aweiksnar:
Yeah I'll get that fixed

aweiksnar:
@edpaget That text flash on submit should be resolved now

PR #1433, opened 2015-08-18
Fix finding the next task key

### LINKED ISSUE #1429: Next task not being picked up correctly
I am making a demo for @chrislintott to use this week at the LSST meeting. The workflow is "question: yes/no, if yes: mark, if no: next". This was working for a while, but then I tweaked the text in the questions and the help boxes, and now the conditional logic is ignored. This seems like a bug, to me!

Here's the demo: 

http://demo.zooniverse.org/panoptes-front-end/#/projects/drphilmarshall/deblendinspector

Feel free to make yourselves collaborators so you can help fix this (or ask me and Ill add you). Thanks!

brian-c:
Single-type tasks were falling back to the task's `next` when the answer's `next` was empty (which should've ended the classification). It can get to that state if "multiple" is checked at some point and a task-`next` is defined there.

Closes #1429.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -89,11 +89,12 @@ Classifier = React.createClass
 
     if TaskComponent is tasks.single
       currentAnswer = task.answers?[annotation.value]
-    # If the next task key exists, make sure the task it points to actually exists.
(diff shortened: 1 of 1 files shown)


aweiksnar:
LGTM

PR #1436, opened 2015-08-19
Let zoo-team members add other zoo-team members

aweiksnar:


brian-c:
Looks good!

Can I ask at some point in the future we replace all these `<span onClick={...}>` tags with `<button>` tags? I see there are some style conflicts to work around. Not critical right now, but I think that's an accessibility issue. I'll open an issue for it.

aweiksnar:
Yeah for sure, I'll rework that css class to work for buttons and then replace those

PR #1437, opened 2015-08-19
Added html5 video player and mp4 to readable formats. Minor layout an…

srallen:
- Added mp4 to readable formats 
- Added HTML5 video player and check for video type to subject viewer
- Hide frame play button if video is present
- Moved frame pagination out of image type to be a part of subject tools layout 
- Added a style to see active frame for the pagination

Added this since I'm working on uploading mammoth-bats data and it'd be nice to be able to play the videos in the subject viewer.

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -42,7 +42,11 @@ module.exports = React.createClass
 
     mainDisplay = switch type
       when 'image'
-        <img className="subject" src={src} style={SUBJECT_STYLE} onLoad={@handleImageLoad} />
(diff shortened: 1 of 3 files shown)


brian-c:
Whoa, sorry this got left hangin'! I left some comments last week but it looks like they never made it though. They were along the lines of:

What project can I see this on (and can you add brian-c)? I couldn't get the spinner to stop spinning (maybe onLoad isn't not firing?).

srallen:
You can check it out on mammoth-bats. I added you as collaborator to the staging project that has 5 videos you can check out in "Secondary Subject Set". Each subject for this project includes an image and a mp4 video.

brian-c:
Got it. Nice! Sorry for the delay again.

PR #1438, opened 2015-08-19
Increase tags count

### LINKED ISSUE #1420: Increasing the number of popular hashtags listed on Talk sidebar
Current Talk lists a small number of popular hash tags.  There's a lot of variety in hashtags and I think the current display, is showing too little. Currently on Planet Four: Terrains there are 10 on the list (I know there are many more hashtags being used).  Previous  version of Talk  listed 20 and I found that really useful because it allowed for easier searching for examples and such, because I could peruse the list to see what people were using. I'm finding that really hard to do with just the 10 . I'm having to use the search instead to try and found out what the hashtag is first and then search for it while writing a blog post right now. I think it would help if the 10 were increased 20 to match what was on Talk 2.0 

Planet Four: Terrains Talk
screen shot 2015-08-15 at 5 59 30 pm 2

Planet Four Talk 

screen shot 2015-08-15 at 5 58 12 pm 2

aweiksnar:
- Closes #1420
- Also changes tags to be block level so they display in a single column. I played around with 2 columns, but most longer tags didn't fit well without ellipsis
- Uses `id`s as keys for tags, since they're now a resource

### DIFF
--- a/app/talk/popular-tags.cjsx
+++ b/app/talk/popular-tags.cjsx
@@ -11,14 +11,14 @@ module?.exports = React.createClass
     params: React.PropTypes.object.isRequired # URL params
 
   tagsRequest: ->
-    talkClient.type('tags/popular').get section: @props.section, limit: 10
+    talkClient.type('tags/popular').get section: @props.section, limit: 20, page_size: 20
 
   tag: (tag, i) ->
     {owner, name} = @props.params
     if owner and name
-      <span key={i}><Link params={{owner, name}} query={query: tag.name} to="project-talk-search">#{tag.name}</Link>{' '}</span>
+      <div key={tag.id}><Link params={{owner, name}} query={query: tag.name} to="project-talk-search">#{tag.name}</Link>{' '}</div>
     else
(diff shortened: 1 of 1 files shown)


chrissnyder:
Are there any projects on staging with popular tags?

aweiksnar:
Not that I'm aware of - I changed `tags/popular` to `tags` for dev to populate the list

chrissnyder:
Ah gotcha, that works.

@parrish Totally tangental to this PR, but any reason the tag list isn't filtered to be unique by name? Or is that a filterable option on the front end? I understand the link to comment is different for each, but we just link to the same search for all of them in this use case.

aweiksnar:
@chrissnyder I believe tags/popular does that?

parrish:
I'm not sure what you're asking Chris, but maybe you missed this change: zooniverse/Talk-Api#55

It sounds like this endpoint:

> Most popular tags for subjects on a project:
> GET /tags/popular?taggable_type=Subject&section=project-123

will cover what you're looking for.  It's a unique list

chrissnyder:
Gotcha, just a consequence on testing from just the `tags` route then.

:+1:

PR #1439, opened 2015-08-19
Update auth client to handle response from superagent

edpaget:
Depends on zooniverse/json-api-client#14

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "react-select": "marten/react-select#close-after-click",
     "react-translate-component": "~0.9.0",
     "sugar-client": "^1.0.1",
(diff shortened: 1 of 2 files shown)


edpaget:
This this should all set if we want to do it.

package.json
```diff
@@ -20,6 +20,7 @@
     "react-select": "marten/react-select#close-after-click",
     "react-translate-component": "~0.9.0",
     "sugar-client": "^1.0.1",
+    "superagent": "^1.3.0",
```
chrissnyder: Don't think this is necessary.

chrissnyder:
Looks good to me after a rebase.

edpaget:
Done

PR #1444, opened 2015-08-20
Markdown project prop

aweiksnar:
- Updates to correspond with markdownz@2.0.0 release

### DIFF
--- a/app/pages/lab/project-details.cjsx
+++ b/app/pages/lab/project-details.cjsx
@@ -153,7 +153,7 @@ module.exports = React.createClass
             <AutoSave resource={@props.project}>
               <span className="form-label">Introduction</span>
               <br />
-              <MarkdownEditor className="full" name="introduction" rows="10" value={@props.project.introduction} onChange={handleInputChange.bind @props.project} onHelp={markdownHelp}/>
+              <MarkdownEditor className="full" name="introduction" rows="10" value={@props.project.introduction} project={@props.project} onChange={handleInputChange.bind @props.project} onHelp={markdownHelp}/>
             </AutoSave>
             <small className="form-help">Add a brief introduction to get people interested in your project. This will display on your landing page.</small>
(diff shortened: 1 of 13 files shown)


brian-c:
Looks like `^123` isn't picked up, and `@name/project^123` only links the user.

See #/projects/brian-testing/ghosts/talk/53/78?comment=466

aweiksnar:
Yeah it'll need an `S` prefix now to match. That was my mistake for initially matching cases without the resource type prefix.  See https://github.com/zooniverse-ui/markdownz/issues/11

brian-c:
Cool, got it now. Anything need to happen before we can use this in Talk comments?

aweiksnar:
Should be good to go with the markdowz update, Mentions will just have to look something like `S123123`

brian-c:
It doesn't look like the project is included in the comment-preview module, so `@u/p^S123` works but `^S234` just turns into "234" for some reason (where'd the "^S" go, anyway?).

aweiksnar:
I just tested locally - `^S123123` should be working ok in project talks, but in zooniverse-wide talk it needs to be project specified. It'll currently fall back to just the id so there's not a weird caret in the text if the comment happens to be rendered outside of the context of the project

brian-c:
Can I see a link to a project comment on dev? I must be doing something wrong here.

aweiksnar:
Ah so sorry! I was looking at a different preview. Now I see where  you're not seeing it:  Will push up a fix in a min

aweiksnar:
I just pushed up a change removing that `comment-preview` module in favor of just markdown, since it was a waste of a class. Deployed [here](http://demo.zooniverse.org/panoptes-front-end/markdown-project-prop/#/projects/alex-panoptes/test-proj/talk/21/80) for testing

brian-c:
Glorious now, thanks!

PR #1448, opened 2015-08-21
Separate page for favorites and links from profile

### LINKED ISSUE #1225: Making it more explciit where favorites are located
I've seen more than one comment about not being able to find favorites. I realized to me, it's not very intuitive that when you favorite that it goes into a collection under your accounts collections. Especially since using the heart is a different functionality that adding to a collection. What about adding 'Favorites' to the dropdown account menu and then it would take you to your collections page with an added favorites tab to just list your collections of favorites?

edpaget:
...and account bar

Closes #1225

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -32,7 +32,6 @@ module.exports = React.createClass
       'All'
 
   render: ->
-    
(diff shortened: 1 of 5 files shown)


chrissnyder:
:+1:

PR #1464, opened 2015-08-24
Add talk to project builder

aweiksnar:


brian-c:
Lookin' good. Should the user be able to edit and delete the boards from this place too? And can we then remove these controls from the talk page? I feel like this just spreads things out.

aweiksnar:
Yeah I can add in edit/delete there too. I'd like to keep it in talk too though, since once a project is mature folks may not want to leave talk to make those actions

brian-c:
Looks good to me.

PR #1472, opened 2015-08-26
fix old register redirect

camallen:
http://zooniverse.org/signups was busted.


edpaget:
What does the '?' do?

camallen:
matches account and accounts

edpaget:
Should we also fix the nginx redirect?

camallen:
meh

PR #1473, opened 2015-08-26
Adds a session identifier

edpaget:
That gets submitted as classification metadata to track classifiers that
log in part way through their sessions

Closes zooniverse/Panoptes#1248
Closes zooniverse/Panoptes#866

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -193,7 +193,7 @@ Classifier = React.createClass
   completeClassification: ->
     @props.classification.update
       completed: true
-      'metadata.session': getSessionID()
+      'metadata.session': getSessionID().id
       'metadata.finished_at': (new Date).toISOString()
       'metadata.viewport':
         width: innerWidth
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -5,6 +5,7 @@ ClassificationSummary = require './classification-summary'
 {Link} = require 'react-router'
 tasks = require './tasks'
 drawingTools = require './drawing-tools'
+{getSessionID} = require '../lib/session'
 
 unless process.env.NODE_ENV is 'production'
   mockData = require './mock-data'
@@ -192,6 +193,7 @@ Classifier = React.createClass
   completeClassification: ->
     @props.classification.update
       completed: true
+      'metadata.session': getSessionID()
       'metadata.finished_at': (new Date).toISOString()
       'metadata.viewport':
         width: innerWidth
--- a/app/lib/session.coffee
+++ b/app/lib/session.coffee
@@ -1,0 +1,9 @@
+generateSessionID = () ->
+  sha2 = require('crypto').createHash('sha256')
+  id = sha2.update("#{Math.random() * 10000 }#{Date.now()}#{Math.random() * 1000}").digest('hex')
+  sessionStorage.setItem('session_id', id)
+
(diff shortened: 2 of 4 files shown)


app/classifier/index.cjsx
```diff
@@ -5,6 +5,7 @@ ClassificationSummary = require './classification-summary'
 {Link} = require 'react-router'
 tasks = require './tasks'
 drawingTools = require './drawing-tools'
+{session} = require '../lib/session'
```
camallen: What doees the { session } do? also will this ever get re required from page navigation, etc?

edpaget: It requires the string from the exported object from '../lib/session' It should calculate the string on page load and remain the same until the page is reloaded.

camallen: so a full page reload would create a new session identifier.

edpaget: Yes

camallen: Ok - just wondering if this will fit @vrooje and @alexbfree 's uses cases where they want a sticky session id to track the user over time? Could we use a expiring local storage key or something?

edpaget: I don't think that'd necessarily be better. The expiry time for the key would still be arbitrary. If you want to keep tracking sessions after a refresh you could look for classifications from the same ip address after a short time.

vrooje: I think we'd like to be able to distinguish between these scenarios, each on a shared computer:

  - Person A opens browser, classifies a bit, closes browser, leaves computer. Person B sits down at computer 10 minutes later & does exactly the same thing. (Goal: we guess it's 2 different people.)
  - Person A opens browser, classifies, pauses to get a cup of coffee, comes back to computer, refreshes browser window, continues classifying. (Goal: we guess it's the same person.)
  - Person A classifies not-logged-in until prompted, then logs in, continues classifying, then closes browser. Person B sits down immediately, opens browser and does exactly the same thing. (Goal: we guess B's not-logged-in classifications don't belong to A.)

Hope this is helpful/realistic/posted in the right place?

edpaget: I don't think it's possible to distinguish between a browser being refreshed and being closed unfortunatley.

camallen: Could we combine the session id and the timestamp of the last submitted classification in local storage to look for a window period to reset the session key, perhaps say 2-5 mins?

edpaget: Yeah I can just use `sessionStorage` actually and that would work perfectly.

edpaget:
yeah session storage was definitely the right way to go?

edpaget:
Ah whoops I messed something else up.

edpaget:
Now it should be set.

camallen:
sweet - one final question. if i leave the browser open and go out for dinner come back then the session would be the same?

edpaget:
From my interpretation of the middle case I thought that would be right: https://github.com/zooniverse/Panoptes-Front-End/pull/1473#discussion_r37986037? 

I've added a ttl, but I think it's impossible to decide what it should be for all projects (unless we're constantly reseting the ttl on every user action), and we should be super explicit about communicating the limits of the session measure.

vrooje:
Agreed this will need plenty of documentation so people know what it is and isn't; but even with limitations it's still hugely useful and much appreciated.

PR #1474, opened 2015-08-26
Replace spans with buttons for sign in prompts

aweiksnar:
From conversation in #1436


edpaget:
Don't know if you're still working on this, but can you add `type=button` to the buttons make sure they never accidentally get triggered as `type=submit`.

aweiksnar:
Cool, yeah probably safer that way. Just updated and rebased

PR #1475, opened 2015-08-26
Followed discussions

parrish:
This lets a user manage subscriptions (participating and followed) for discussions.

I'm not 100% on how we want to align this on the page though.

This is good to go for staging ~~, but not production yet~~ and production.


aweiksnar:
:+1:

Issue #488, opened 2015-06-04
Allow project builders to make drawing tasks "required"


aliburchard:
Questions can be required. I think drawing tasks should be able to be set as required as well. Right now it's hard to explain why only questions can be required. If there's an instance where there is nothing to draw, then project builders can set required = false -- but on all of the drawing tasks I've made, I've made them contingent on the critter being there, so I want to require them.

chrissnyder:
I 100% agree that we need this, but I don't think we need for the imminent soft launch.

Can (and should) be a high-priority issue afterwards though. 

Happy to debate further

aliburchard:
Fair enough. I don't think it's absolutely critical if it slows down the launch, but I do think it's really important.

chrislintott:
I think we should make this very high priority - if not before then immediately after launch

On Jun 18, 2015, at 6:16 AM, aliburchard <notifications@github.com<mailto:notifications@github.com>> wrote:


Fair enough. I don't think it's absolutely critical if it slows down the launch, but I do think it's really important.

aliburchard:
@chrissnyder @brian-c I think this should be Sprint 1 - please feel free to disagree.

chrissnyder:
Removing the PFE Sprint 1 label because that doesn't make sense anymore. Likely will tackle this in the not-too-distant future anyway.

brian-c:
I've got a plan for checking required-ness across tasks. I'll plug it in as soon as the survey task is done.

aliburchard:
@brian-c how is this coming along? And is it possible to merge it with the highly requested "set min/max # markings per task" in #1564 ? This keeps cropping up as an issue across projects we're building, and would also help *enormously* with aggregation. Would it be possible to bump this up the priority list after tutorials but before the next big fancy addition?

brian-c:
At least one mark can now be required for drawing tasks. Closing this, will comment on #1564.

aliburchard:
Yay!! Thank you!

Issue #251, opened 2015-05-01
'Required' not enforced for "details" questions


mkosmala:
It would be nice to be able to require an answer in the "details" pop-ups.

aliburchard:
Similarly, with the drawing tasks, there's no "require" option. Is that intentional? (And if so, why can we "require" a question but not a drawing task?)

simoneduca:
I think the refactor mostly takes care of this. Will create a new issue, should the need arise.

PR #1476, opened 2015-08-26
Classifier refactor

brian-c:
This better isolates classifier components' responsibilities.

The Classifier now keeps track of the state of the current classification and task/annotation. Tasks can implement `onLeaveAnnotation` to react to moving on to the next annotation (e.g. The drawing task forcing any incomplete shapes to complete).

The subject annotator displays the subject and provides hooks for individual tasks to modify it however they need to. Tasks can implement `InsideSubject` (e.g. drawing task's mark initializer), `PersistInsideSubject` (the drawing task's marks renderer, which persists even when you're not on a drawing task), and `AfterSubject` (current survey choices) components to modify the subject annotator. We can easily add more hooks as we require them.

Drawing marks' sub-task details popup are now rendered by the drawing tool root using the new modal-form module, which is way, way saner than hacking the right position the tether/Tooltip module. This should be the only visible change.

This also sets things up for required drawing tasks (#488) and sub-tasks (#251) and drawing tools handling their own `forceComplete` behavior (#1310).

Big changes. Please ask if something seems weird.


app/classifier/index.cjsx
```diff
@@ -41,24 +39,30 @@ Classifier = React.createClass
...
     @setState subjectLoading: true
-    # TODO: Pre-cache subject images here and change loading state back.
+    preloadSubject subject
+      .then =>
+        if @props.subject is subject # The subject could have changed while we were loading.
+          @setState subjectLoading: false
+          @props.onLoad?()
 
   prepareToClassify: (classification) ->
     setTimeout => # TODO: Why is this here?
```
aweiksnar: Is this setTimeout still needed? I've been consistently getting a defined `classification` without it

brian-c: Looks like it's no longer needed. I've removed it.

brian-c:
Just rebased and force-pushed.

aweiksnar:
Just noticed something: It seems that some marks are now appearing offset to the cursor. This is reproducible if you go into the fossil-finder project and make some marks in the first step

brian-c:
Boo. Will investigate, thanks. I'm noticing some subject resizing issues now too. Put this on hold for a bit.

aweiksnar:
Cool will do

brian-c:
I think I got it: To reproduce consistently, go to the classify page, scroll down, go to another page, and hit the browser's Back button. That'll mount the classify page with its initial `state.sizeRect` offset by the scroll amount.

Fixed by factoring in current scroll when setting the sizeRect.

aweiksnar:
:+1:

PR #1481, opened 2015-08-27
Allow conversations to be destroyed

### LINKED ISSUE #1324: Ability to delete PMs from your inbox
Users will want to be able to delete private messages from their inbox to avoid it becoming overly cluttered, but they can't right now.

(Is there an issue for general PM inbox styling/functionality reform? I thought there was, but I can't find it.)

parrish:
Closes #1324

### DIFF
--- a/app/talk/inbox-conversation.cjsx
+++ b/app/talk/inbox-conversation.cjsx
@@ -5,12 +5,12 @@ PromiseRenderer = require '../components/promise-renderer'
 HandlePropChanges = require '../lib/handle-prop-changes'
 {Markdown} = require 'markdownz'
 CommentBox = require './comment-box'
-{Link} = require 'react-router'
+{Link, Navigation} = require 'react-router'
 {timestamp} = require './lib/time'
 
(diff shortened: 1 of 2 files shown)


app/talk/inbox-conversation.cjsx
```diff
@@ -67,6 +67,11 @@ module?.exports = React.createClass
       .then (message) =>
         @setConversation()
 
+  handleDelete: (e) ->
+    e.preventDefault()
```
aweiksnar: Looks great, only request is to add a window.confirm to the handler since it's a harsh action

parrish:
Can do.  By the way, the conversation "revives" if a user replies again, so it's not really gone until all participants delete it

aweiksnar:
Ah perfect, or we could just change it to 'Hide' or something then

parrish:
"Archive?"

aweiksnar:
:+1: Yeah that works, @vrooje was on board with 'archive' here #1324

PR #1484, opened 2015-08-28
Adds user blocking

parrish:
This lets uses block each other.

A block prevents the blocked user from messaging the user.

Additionally, a user cannot message the user they blocked.


aweiksnar:
:fist:

PR #1485, opened 2015-08-28
Add scrollX fallbacks for internet explorer

aweiksnar:
- Add fallbacks for scrollX in winternet explorer


brian-c:
Whoops, my bad. Let's just use `scrollXOffset` and drop the `scrollX` entirely.

PR #1486, opened 2015-08-29
Automate top-level className management

aweiksnar:
- This moves top-level className management to the app component, instead of having children modify the document element. It also just reads from an object now
- Also moves the lab landing screen to a new page with it's own url, and redirects there if there's no user for the lab
- The project background 'catch' sets transparency instead of via css. I believe this is a harmless change


app/pages/lab/index.cjsx
```diff
@@ -20,6 +19,9 @@ module.exports = React.createClass
     creationError: null
     creationInProgress: false
 
+  componentWillMount: ->
+    @transitionTo('lab-landing-page') unless @props.user
```
brian-c: `transitionTo` breaks the back button here. I think we want `replaceWith`.

brian-c:
* I'm not crazy about moving background-changing responsibility from the component to the top level. I think ideally we'd set up some kind of generic `<Page background="...">` container. That'd give us a good place to set page titles, too. I don't feel strongly about this though.

* Why bother moving the not-logged-in Lab route? This implementation breaks the back button (see comment), doesn't transition when you sign out from the lab page, and doesn't transition to the lab when you sign in. What's this solve?

* Maybe `gray` would be better default than `transparent`, since it's ideally expecting an image to render behind the white text?

aweiksnar:
Ok I'll make some changes that address those points.  I do think moving app-level state to a single location presents an improvement over the current system, but agree that a `<Page />` type system also makes sense.

chrissnyder:
Might depend upon how we go forward with styles like `on-secondary-page`. Whether you are on a secondary page is certainly app-level state. From a low-level consideration, I'm all for any work that removes direct DOM manipulation.

Whether it makes sense for us to organize our css around that is something else entirely, and outside the scope of this PR.

aweiksnar:
I'm going to close this one up. Seems as this issue will be better handled as part of a larger app restructuring.

PR #1493, opened 2015-08-31
Crop task

brian-c:
Re-implemented and isolated cropping as its own task after the big refactor.

No editor yet! See the `crop` task in mock-data.coffee.

Demo-able here http://demo.zooniverse.org/panoptes-front-end/crop-task/#/dev/classifier

cc: @aliburchard 

Closes #1404.


aliburchard:
@brian-c if I want to add this tool into an existing project (https://www.zooniverse.org/projects/aliburchard/ieye/), how do I go about doing that?

brian-c:
There's no editor for it yet so it has to be set manually through the API, sorry.

PR #1494, opened 2015-08-31
Hide page links from redirectable projects

edpaget:
Closes zooniverse/wildcam-gorongosa#90

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -37,6 +37,14 @@ counterpart.registerTranslations 'en',
       education: 'Education'
       talk: 'Talk'
 
+ProjectAvatar = React.createClass
+  displayName: 'ProjectAvatar'
+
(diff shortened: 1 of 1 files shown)


aweiksnar:
Cool, Looks good to me. Are all the custom projects coming out using `/#/classify`?

edpaget:
I'm not sure. I think most are (or could be modified to).

aweiksnar:
Ok cool, I'll double-check the ones in production and then merge this in :shipit:

aweiksnar:
Looks like old-weather and annotate will have different url schemas, but bats & gorongosa will be ok

edpaget:
Cool I'll open an issue for those two to have `classify` redirect to thier classification page.

aweiksnar:
Sweet. otherwise maybe this could read a classify-return-path from localstorage

PR #1501, opened 2015-09-01
Talk recents

parrish:
This adds a recent comments view at the section level as well as on board scopes.

I'd definitely like some feedback, particularly on the breadcrumb changes -- I feel like the conditional logic is getting kind of nasty there.

Also, I just dropped the link into the sidebar -- I definitely want it to be accessible, but kind of out of the way too.


DarrenMcRoy:
Any sort of recents view whatsoever has been overwhelmingly requested by the entire community. Super happy if we're going to be able to offer something soon! :+1:

aweiksnar:
I think this looks great. Only requests would be to have the link styled as an `a`, since right now it matches the other headings in the sidebar which aren't clickable. And to maybe match `/talk\/recents/` for the recents path just in case for some weird reason somebody uses that as their username or project name

aweiksnar:
:100: 
I think the breadcrumbs is definitely an improvement here, but maybe could be generalized later on. Not totally sure the best way to write that

PR #1504, opened 2015-09-01
Record auto-closing of polygons

### LINKED ISSUE #1118: Enforce closure of polygons

brian-c:
Closes #1310, which closes #1118.

### DIFF
--- a/app/classifier/drawing-tools/polygon.cjsx
+++ b/app/classifier/drawing-tools/polygon.cjsx
@@ -32,6 +32,7 @@ module.exports = React.createClass
 
     forceComplete: (mark) ->
       mark.closed = true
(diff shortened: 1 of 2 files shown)


edpaget:
:+1:

PR #1505, opened 2015-09-01
Fix comment links Navigation function

aweiksnar:
- To prevent having to click the back button twice when navigating the
  browser to single comments


aweiksnar:
@parrish - I noticed this when playing with the recents stuff (also happens in profile). Just makes it so if you follow a link to a single comment and then click back you only have to do it once.

parrish:
Good catch!

PR #1509, opened 2015-09-02
Click for account menu

brian-c:
Account menu now drops down on click instead of hover.

Also added some handsome icons.

### DIFF
--- a/app/partials/account-bar.cjsx
+++ b/app/partials/account-bar.cjsx
@@ -70,7 +70,7 @@ module.exports = React.createClass
                 <i className="fa fa-star fa-fw"></i>{' '}
                 <Translate content="accountMenu.favorites" />
               </Link>
-              <br />
+              <hr />
               <button type="button" className="secret-button sign-out-button" onClick={@handleSignOutClick}>
                 <i className="fa fa-sign-out fa-fw"></i>{' '}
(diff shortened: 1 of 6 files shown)


edpaget:
Sorry for being late to this, is it possible to do both? Won't it be a bit of an unexpected change for people if the hover doesn't work?

brian-c:
That's why the hover state is there. People'll get it. I'd prefer to not have things pop up on hover ever.

eatyourgreens:
Looks like this kills the accessibility fixes added in #1096. I can tab to the control, press enter to activate the menu, but can't enter the menu using either tab or the cursor keys.

brian-c:
Whoops, sorry about that. Should be fixed now.

eatyourgreens:
Excellent! It's working exactly like the profile button on twitter.com now.

Is it possible to pass `aria-haspopup` and `aria-expanded` through as attributes to the trigger button element? I don't think VoiceOver will read them on the span.

brian-c:
@eatyourgreens I've just made `aria-haspopup` and `aria-expanded` attributes on the trigger component in its repo. That'll have to be updated before this can be merged. Also added `menu` and `menuitem` roles.

brian-c:
Closing until new modules is finished.

PR #1492, opened 2015-08-31
Remove hash

edpaget:
This remove the need to use the hash fragement for URLs.

I'm working on a corresponding update to the nginx static server.

PR #1510, opened 2015-09-02
Fix relative path includes

edpaget:


brian-c:
This will break deploys to anywhere but the root, no?

edpaget:
Yes, but https://github.com/zooniverse/Panoptes-Front-End/pull/1492 already did, and it turns out its important for getting google to index project pages. I'll add a configuration option shortly that will restore the ability to deploy to other places.

astopy:
Might be worth adding a variable to change the root prefix for non-root deploys.

edpaget:
Yup am planning to.

PR #1511, opened 2015-09-02
Make the URL hash fragment routing optional

edpaget:
By setting NON_ROOT=true PFE will use the hash fragment for routing.

This also fixes the relative paths for the team images and automatically
changes hash fragments into normal URLs


edpaget:
This is pretty critical so hardcoded unsubscribe lines aren't broken.

edpaget:
Doing some extra stuff with this please hold off.

parrish:
LGTM. Seems to fix pagination.  I'll take a look at how comment links are being constructed next (it's still building paths like `/talk/:board_id/:discussion_id//talk/:board_id/:discussion_id/?comment=:comment_id`

edpaget:
Those paths seem like they should be fine, hmm.

parrish:
(it's doubling the params)

app/pages/profile/comment-link.cjsx
```diff
@@ -26,7 +26,7 @@ module?.exports = React.createClass
       @getCommentHREF(nextProps.comment)
 
   getCommentHREF: (comment) ->
-    @setState({href: '#'})
+    @setState({href: ''})
```
parrish: This isn't necessary is it? Just change: https://github.com/zooniverse/Panoptes-Front-End/pull/1511/files#diff-28e40cf08cbf41290f440b8f9985060bR19

PR #1523, opened 2015-09-03
add data exports page to projects

### LINKED ISSUE #1045: Putting the csv dump links in a separate data section of the project builder
low in the queue - I keep getting confused and having to remember where the links are to download the data. Could the two data dump links be put into a separate section/tab on the project builder?  Looking at the navigation links if I haven't been to the project builder in a while, I wouldn't intuitively think that project details is where I would go to get csvs of the classification database and the subject data. I think putting them in their own section would be useful especially for team members of a project who didn't really build the site but are helping with data analysis

camallen:
closes #1045 - added separate page for data dumps and linked up the workflow & contents export. 

Please check i've included only the necessary requires for the data-dump page and i haven't removed too much in the project-details one.

### DIFF
--- a/app/pages/lab/data-dumps.cjsx
+++ b/app/pages/lab/data-dumps.cjsx
@@ -1,0 +1,75 @@
+React = require 'react'
+handleInputChange = require '../../lib/handle-input-change'
+PromiseRenderer = require '../../components/promise-renderer'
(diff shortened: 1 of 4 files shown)


brian-c:
Good stuff.

PR #1527, opened 2015-09-03
Use edpaget's react-router branch

### LINKED ISSUE #1521: Paging through collections breaks
This doesn't seem exactly the same as #1284 but I'm not completely sure.

Just tried paging back and forth through a collection. Going from page 1-->2-->3 worked fine, but then when I tried to go back I ended up with a "Not found" because I was apparently directed to:

https://www.zooniverse.org/collections/vrooje/not-a-bar-but-kinda-looks-like-one/page=3?page=2

edpaget:
...until I can get it merged into master. This fixes their regex that
looks for optional trailing slashes, so it doesn't conflict with a
question mark that begins the search segment of a url.

Closes #1521


edpaget:
Actually installing via git may not work. Stupid typescript.

edpaget:
Anyway the upstream PR is in https://github.com/rackt/react-router/pull/1816/files

chrissnyder:
The PR was merged, but I don't think react-router is going to publish any new 0.13.x versions. Stick with this fork?

edpaget:
Yeah, I'm going to publish it as a namespaced npm module ('edpaget/react-router 0.13.4' or whatever), since there's a build step to convert their typescript. I'll ping back here when that's done.

edpaget:
@chrissnyder I've updated this to us my scoped release of React-Router 0.13.x on npm. Which is here: https://www.npmjs.com/package/@edpaget/react-router

chrissnyder:
I think some weird compile-step was accidentally committed.

edpaget:
Ah whoops yes it looks like I messed up my find and replace for react-router.

edpaget:
@chrissnyder This should be all set.

chrissnyder:
Looks fine from what I can test but I'm afraid to merge without another pair of eyes. @parrish could you also look at this? Relevant commit at https://github.com/edpaget/react-router/commit/53ce7f16f4241b9ea66f25b046e3f4b7f32a0418

parrish:
Missed one: https://github.com/edpaget/Panoptes-Front-End/blob/update-router/app/pages/admin.cjsx#L5

edpaget:
Updated.

parrish:
After staring at this for a while, I think I'm missing the motivation for this.

Are there currently any buggy routes?

If not, it looks like we're just replacing one workaround here: https://github.com/zooniverse/Panoptes-Front-End/pull/1527/files#diff-a50495afc02ad6142ea0b2b5538e5634R64

That aside, it seems like it works, but I can't really say for sure.

edpaget:
The motivation was the issue it marks closed #1521. The regex in react-router doesn't correctly detect the difference between routers with an optional trailing slash and routes with query params, which is fixed in my PR to that project. 

There are also 5 months worth of other unreleased react-router bugfixes contained in the version I published.

parrish:
Cool.  That makes more sense.

PR #1542, opened 2015-09-09
Replace searches fetch with a standard API resource

parrish:
After zooniverse/Talk-Api#71 deploys, this should be good to go.


aweiksnar:
Awesome, LGTM

PR #1547, opened 2015-09-10
Fetch sidebar projects list from user prefs

aweiksnar:
- Fetches sidebar projects from user prefs, or falls back to launch
  approved projects
- Short-term solution so that list isn't hard-coded anymore
- Wouldn't mind extra eyes on this in case I'm missing something ; dev projects are pretty different from production ones


chrissnyder:
:+1:

Issue #1517, opened 2015-09-03
Adding Zooniverse markdown specifics like the +tab+ to the project builder markdown help


mschwamb:
If possible, it would be really helpful to add the +tab+ syntax for making links open in new tabs and any other syntax specific to the Zooniverse markdown listed on the project builder help as an addition to the generic markdown tutorial currently linked

![screen shot 2015-09-03 at 9 06 10 am](https://cloud.githubusercontent.com/assets/753325/9648435/1a2445b4-521b-11e5-97c3-6f3d95ba78a3.png)

chrissnyder:
Good idea.

@DarrenMcRoy Can you do this and issue a PR to https://github.com/zooniverse-ui/markdownz? We'll make a separate page to point people to after the text is ready.

aweiksnar:
The helptext actually lives in PFE
- https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/markdown-help.cjsx
- https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/comment-help.cjsx

DarrenMcRoy:
Update: Working with @aweiksnar tomorrow to develop a beginner-friendly guide to all Markdown and Zooniverse-flavored Markdown capabilities (to replace the current link to http://markdownlivepreview.com/).

This can be linked from the Talk comment box as well as from the Project Builder, so we have a single location where everyone can figure out how to do whatever they need to do with Zooniverse-flavored Markdown. :-)

chrissnyder:
Hold on this. I'd like it to be the first task for new dev (coming soon!)

DarrenMcRoy:
Okay. I've done a bit of advance legwork for them, though. :-)

Here's a list of all the Markdown features that we'll want to explain.

https://docs.google.com/document/d/1I2KapyYtwY-ebRd8WLm0k_iw68Dkk4mViEJbIjsb19c/edit

PR #1549, opened 2015-09-10
Add notification preference and help text for group mentions

parrish:
I'm opening this just to get it up, but the backend is going to only be deployed to staging for the moment.

I'll comment/assign when it's ready for production.

### DIFF
--- a/app/pages/settings/email.cjsx
+++ b/app/pages/settings/email.cjsx
@@ -16,10 +16,11 @@ module.exports = React.createClass
       when 'participating_discussions' then "When discussions I'm participating in are updated"
       when 'followed_discussions' then "When discussions I'm following are updated"
       when 'mentions' then "When I'm mentioned"
+      when 'group_mentions' then "When I'm mentioned by group (@admins, @team, etc)"
       when 'messages' then 'When I receive a private message'
 
   sortPreferences: (preferences) ->
-    order = ['participating_discussions', 'followed_discussions', 'mentions', 'messages']
+    order = ['participating_discussions', 'followed_discussions', 'mentions', 'group_mentions', 'messages']
     preferences.sort (a, b) ->
       order.indexOf(a.category) > order.indexOf(b.category)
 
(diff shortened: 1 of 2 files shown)


DarrenMcRoy:
We may have to discuss this further down the road, based on some of what came out of the Best Practices seminar.

This tool will allow ANYONE to alert the researchers, when there was pretty universal agreement from the seminar that only moderators should be allowed to flag posts for researchers.

However, it's a good start!

parrish:
I don't think that's an issue at all.  Any time a comment is reported moderators/admins are notified.

If a researcher doesn't want to receive these notifications, or prefers to receive them as a digest, they can manage these notifications separately.  Abuse can be handled the same as anything else.

I think it's more appropriate to allow anybody to do this rather than have to try to push something up a chain.

If, for instance, a project team would rather it come via the moderators then they simply unsubscribe and have their community mention the moderators instead.

DarrenMcRoy:
"If, for instance, a project team would rather it come via the moderators then they simply unsubscribe and have their community mention the moderators instead."

In which case the moderators are required to manually assemble a separate digest for the researchers, instead of just clicking a button that says "researchers should look at this" -- a feature idea that was met at the workshop with *universal* acclaim from volunteers, moderators, researchers, everyone.

The entire problem in the first place is that researchers don't have time to answer all the questions asked of them, and need to know which are the important ones (and which the moderators can just handle themselves). This is why the idea of a moderator-operated flag for the researchers was so popular. Group mentions are nice, but they don't quite solve that core problem.

parrish:
Happy to reevaluate in a few months if there ends up being too much noise.  Right now, group mentions haven't even been used yet.

Right now, we should work on a better presentation of what features are currently available.

DarrenMcRoy:
Agree 100%. All I was saying is, let's discuss further down the road. :-)

chrislintott:
@parrish Hang on, this exists? Let's trial it with a science team. I nominate Galaxy Zoo.

parrish:
It's live with all Panoptes projects.  We just need to figure out a good way to present it as an available feature.

DarrenMcRoy:
I'm going to be working with a developer to build a "guide to Talk functionality" which explains everything that you can do, but @chrissnyder has asked me to hold off until we have our new developer (see: #1517)

PR #1554, opened 2015-09-11
Remove nested AutoSave

brian-c:
Unnecessary nested AutoSave element was losing its event, so it was impossible to remove external project links. cc: @aliburchard.


aweiksnar:
:+1:

PR #1561, opened 2015-09-15
Change sidebar dropdown to links, improve request

aweiksnar:
- Populates the sidebar project list with any projects that are non-redirected
  - Bit of work to get here on the frontend, but essentially keeps requesting/filtering until it gets 10 projects , unless it's at the last page
  - Updates the ui in batches of 10
- Sidebar just holds a list of collapsible project links instead of the previous dropdown, should be less clicking this way

### DIFF
--- a/app/talk/active-users.cjsx
+++ b/app/talk/active-users.cjsx
@@ -93,6 +93,6 @@ module?.exports = React.createClass
         {@state.users.map(@userLink)}
       </ul>
       {if @state.pageCount > 1
-        <Paginator page={+@state.page} onPageChange={@onPageChange} pageCount={@state.pageCount} scrollOnChange={false} />
+        <Paginator page={+@state.page} onPageChange={@onPageChange} pageCount={@state.pageCount} scrollOnChange={false} firstAndLast={false} pageSelector={false} />
       }
     </div>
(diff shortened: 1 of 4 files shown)


aweiksnar:
@edpaget - does `loadProjects` seem like a sane way to get non-redirected projects from panoptes?

parrish:
Is there any way to filter/sort that by projects the user is participating in?  I could see that list growing to be unusable pretty quickly.

aweiksnar:
@parrish It could use `user_preferences` (what it's doing now), but for most users that is only a project or 2. Another goal of that list is to get people clicking into new projects too, so hopefully putting other projects in there will help with that. I'm kinda looking at this sidebar projects list as a shorter term solution until the projects page is more navigable/searchable

parrish:
Cool.  Ready for me to merge?

DarrenMcRoy:
Yeah this is a short-term fix until we can get something more robust overall in terms of Talk framing features. I really appreciate it, Alex ... I know I've given you hell over this. :-)

This only populates with approved projects, right? "Is it Bacon" won't show?

aweiksnar:
@parrish As long as hitting that endpoint until it gets non-redirected projects is cool, I think this is good to go

parrish:
It's better than being broken, for sure.

aweiksnar:
Looks like this didn't translate to production so well

parrish:
Yeah... weird.  I'm going to leave it as is, since it's currently better than just displaying an error.

Any idea why it's behaving differently?

DarrenMcRoy:
For me, it only shows the Zooniverse Talk, nothing else.

Also -- separate issue -- you can only see it on the top level. Would be nice if it were accessible from anywhere.

aweiksnar:
@parrish Think I've got a fix here #1583. I think it was not calling set state after getting to the last page. There's 2 pages in production, and it looks like the first one is all redirected

PR #1578, opened 2015-09-17
Duplicate focuses

### LINKED ISSUE #1572: Everytime you post in a notes thread it copies the subject
@parrish and @aweiksnar I think a change in the api to make sure notes comments are associated with the subject now causes each post in a notes thread  (after the original post) to post the subject image with the reply

screen shot 2015-09-17 at 6 17 55 am

See https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/8/4820/

parrish:
Closes #1572

Renders the focus in a comment if
  - it's not a focused discussion and this comment has a focus
  - it's a focused discussion and this is the first comment
  - it's a focused discussion and this comment is about a different focus

### DIFF
--- a/app/talk/comment.cjsx
+++ b/app/talk/comment.cjsx
@@ -121,7 +121,18 @@ module?.exports = React.createClass
     </div>
 
   render: ->
-    isFirstSubjectComment = @props.index is 0 and @props.data.discussion_subject_default
+    # When the discussion is about a focus and this is the first comment on the page
+    isFirstSubjectComment = @props.index is 0 and @props.data.discussion_focus_id
+
+    # When the discussion is about a focus, but this comment is about a different focus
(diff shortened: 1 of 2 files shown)


parrish:
By the way, you can find an example of me poking around at this in http://localhost:3735/projects/brian-testing/ghosts/talk/53/78/

aweiksnar:
LGTM
Can you also add `index` to the comment propTypes? (perhaps with `.isOptional`)
Also, noticed that with this will not render focuses in the recents view. Thoughts on Prefixing `shouldRenderFocus` with `(not @props.index?) or`... so it'll always render it unless given an index prop for discussion contexts

parrish:
Good catch!

aweiksnar:
:shipit:

PR #1582, opened 2015-09-17
Synthetic replacement for hashchange events

### LINKED ISSUE #1540: New private message notification mark doesn't disappear after reading the message
When I receive a private message and open it, the red notification mark doesn't disappear. It stays displayed there even after leaving the Profile page and visiting the Talk. (I came back to the Profile page to check whether it means I've received another message and figured out it just hasn't disappear yet.) Only reloading the Profile page helps. 

This happens always.

parrish:
Closes #1540 

Since we don't have hashchange events anymore, I'm just dispatching a fake event on route changes.

I think `new Event` and `.dispatchEvent` don't have compatibility issues, but somebody may want to double check me.

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -11,6 +11,7 @@ if process.env.NON_ROOT isnt 'true' and location.hash isnt ""
   location.pathname = location.hash.slice(1)
 
 router.run (Handler, handlerProps) ->
+  window.dispatchEvent new Event 'pagechange'
   React.render(<Handler {...handlerProps} />, mainContainer);
 
 logDeployedCommit = require './lib/log-deployed-commit'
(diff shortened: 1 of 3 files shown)


aweiksnar:
@brian-c is using a package for change events here https://github.com/zooniverse-ui/modal-form/pull/6 / https://github.com/xpepermint/history-events. Assuming that'll be added into the frontend it may be worth coordinating

brian-c:
Yeah, I've got it in my click-for-account-menu branch. I'm fine with us triggering an event ourselves though, and that's a good place to do it. Depending on the history-events lib (3 stars on GitHub right now) is probably not a great idea.

Small nitpick: could we call the event `locationchange`? `window.location` is the thing that's changing, the page itself might not.

parrish:
Thanks for the feedback.  I'm not super confident in that package, but this seems like a simple enough workaround.

aweiksnar:
Cool, works great!

PR #1583, opened 2015-09-17
Improve sidebar load projects request

aweiksnar:


parrish:
Seems to work fine locally.

I am seeing a `Warning: setState(...): You passed an undefined or null state object; instead, use forceUpdate().` now though

parrish:
Think it might be this [`setState`](https://github.com/zooniverse/Panoptes-Front-End/pull/1583/files#diff-f400dfd5fa10c5405fcdb39e4a201796R41) causing it.  It calls once correctly, then again with an empty object

aweiksnar:
Good call, just updated so that `setState` will be called only after request cycle is done

parrish:
:+1:

PR #1586, opened 2015-09-17
Click for account menu

brian-c:
Another attempt!

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -1,3 +1,5 @@
+historyEvents = require 'history-events'
+
 React = require 'react'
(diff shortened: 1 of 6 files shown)


aweiksnar:
Works great! 
Only thing I noticed are console warnings `"Warning: owner-based and parent-based contexts differ ..."` when interacting with the menu

brian-c:
Rendering `Link`s outside the router context does that. No way around it until React 0.14 unfortunately.

aweiksnar:
Ah annoying! Roger that :shipit: :shipit:

PR #1589, opened 2015-09-18
Change to CustomEvent for IE 11

srallen:
Changed from `Event` to `CustomEvent` object because [IE 11 doesn't support it](http://stackoverflow.com/questions/26596123/internet-explorer-9-10-11-event-constructor-doesnt-work). Since Edge is out, I'm assuming I don't need to worry about adding the polyfill for older IE browsers. Also checked Edge and it's ok there.

I have a few modern.ie VMs setup, so ping me if anyone needs me to check support for IE/Edge.


aweiksnar:
:shipit: :100: Yeah that looks like [the one](http://caniuse.com/#feat=customevent)

PR #1593, opened 2015-09-18
CSS scrollbar hack for windows and linux browsers

### LINKED ISSUE #1296: Dialog Controls overlap scroll bar

srallen:
- Fixes #1296 in a hacky way. This gets the scrollbars in browsers in Windows and Linux to have autohide like behavior like they do in OS X.

### DIFF
--- a/css/dialog.styl
+++ b/css/dialog.styl
@@ -25,7 +25,7 @@
   position: absolute
   top: 40%
   transform: translate(-50%, -40%)
-
+    
 .dialog-controls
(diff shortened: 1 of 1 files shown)


edpaget:
Sorry just saw this, but please let's find another solution. Scroll bars aren't hidden natively in Linux Windows and users of those platforms will expect them to be there if content is scrollable.

srallen:
Sure, I'll think about another solution. I'll close this then.

PR #1606, opened 2015-09-21
Handle both subjects and recents in comment image selector

parrish:
Currently only recents can be linked.

It feels kind of hacky to have to specify the type on the resource, but it doesn't look like the api client is setting/exposing the type correctly.  This seems like the most straightforward way to conditionally handle it.

### DIFF
--- a/app/talk/comment-image-selector.cjsx
+++ b/app/talk/comment-image-selector.cjsx
@@ -16,11 +16,15 @@ module?.exports = React.createClass
 
   setRecents: ->
     @props.user.get('recents', {sort: '-created_at'})
-      .then (subjects) => @setState {subjects}
+      .then (subjects) =>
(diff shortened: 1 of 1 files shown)


aweiksnar:
Works for me. Do you think this would be a safe way to read the type instead of setting it:
 ```
  setFocusImage: (image) ->
    imageIsRecent = image.links.subject?
    if imageIsRecent
      image.get('subject').then (subject) =>
        @props.onSelectImage(subject)
    else
      @props.onSelectImage(image)
```

parrish:
Possibly.  That does break if subjects ever get a self-referential association though.  That use case is pretty likely if/when Panoptes adds functionality for more complicated workflows (e.g. pairwise)

aweiksnar:
Ok cool, merging in! Thanks!

PR #1612, opened 2015-09-21
Render null crop summary

brian-c:
Whoops, moving past a crop task without cropping anything would break at the summary screen.


aweiksnar:
LGTM :shipit:

PR #1621, opened 2015-09-22
Admin area

edpaget:
This creates an admin area. Currently this is only used for viewing project statuses and changing them. The log of changes depends on zooniverse/Panoptes#1380, but this should still work without it.

### DIFF
--- a/app/pages/admin/project-status-list.cjsx
+++ b/app/pages/admin/project-status-list.cjsx
@@ -22,11 +22,7 @@ module.exports = React.createClass
 
     query[@props.query.filterBy] = true unless query.slug?
 
-    apiClient.type('projects').get(query).then (projects) =>
-      projects
-
-  toggleActive: (project, e) ->
-    e.preventDefault()
+    apiClient.type('projects').get(query)
 
   render: ->
     <div className="project-status-page">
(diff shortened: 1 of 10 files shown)


app/pages/admin/project-status-list.cjsx
```diff
@@ -0,0 +1,57 @@
...
+      sort: '+updated_at'
+      include: 'avatar'
+
+    Object.assign query, @props.query
+
+    delete query.filterBy
+
+    query[@props.query.filterBy] = true unless query.slug?
+
+    apiClient.type('projects').get(query).then (projects) =>
```
aweiksnar: I think this can just return `apiClient.type('projects').get(query)`

app/pages/admin/project-status-list.cjsx
```diff
@@ -0,0 +1,57 @@
...
+      projects
+
+  toggleActive: (project, e) ->
```
aweiksnar: Doesn't look like this is called anywhere

app/pages/admin/project-status-list.cjsx
```diff
@@ -0,0 +1,57 @@
...
+              [owner, name] = project.slug.split('/')
+              <div key={project.id}>
+                <Link to="admin-project-status"  params={{owner, name}}>
+                  <ProjectIcon linkTo={false} project={project} />
+                </Link>
+              </div>}
+          </div>
+          <Paginator
+            page={+@props.query.page}
+            pageCount={projects[0].getMeta().page_count} />
```
aweiksnar: I'm getting `TypeError: projects[0] is undefined` if the filtered projects list is empty

aweiksnar:
Cool, LGTM. Shall I wait on 1380 or :shipit: now?

edpaget:
This should be fine to go whenever.

edpaget:
I still have to sort the broken build.

aweiksnar:
:cool:

PR #1638, opened 2015-09-25
project workflow activation on project admin page

camallen:
add ability to modify the project workflow activation setting for admin. also handle empty project lists from applied filters.


chrissnyder:
:+1:

PR #1639, opened 2015-09-25
Only disable action buttons in locked discussions

aweiksnar:
- Only disables action buttons in locked discussions instead of the
  whole thread, so people can still follow links
- Closes https://github.com/zooniverse/scribeAPI/issues/395


chrissnyder:
:+1:

Issue #1563, opened 2015-09-16
urgent - sub-task and question progression


AdrianEvans:
If a subtask answer is a required element in the workflow you can still continue to mark new features and the sub-task box will just disappear - you can also click the OK box and the sub-task will just disappear

the result is that the user can not progress to the next question - greyed out 'done' button. but there is no explanation as to why so the user doesn't get it. It isn't intuitive that you can click back on your spot and reopen the sub-task.

The code should probably not let them close the sub-task, by any method, unless they have checked a box or inserted an answer when the task is required.

many users are getting stuck and frustrated because of this and we are losing them.

aliburchard:
noticing that on projects with >1 sub-task, I find myself just clicking "ok" instead of answering all the sub-tasks, and then am stuck. I suspect this will be a wide-spread issue.

AdrianEvans:
yes. I'm thinking of unchecking 'required' then if people miss it then it
doesn't stop them progressing (obvs a temporary measure because if we
didn't want the sub-task data we wouln't ask for it)

chrislintott:
@brian-c, what say you?

Chris

brian-c:
This behavior was greatly improved in #1641, with the caveat that everyone hates the added note that change introduced (see #1666).

I left this issue open since the implementation still isn't perfect. Redesigning the classification interface to handle various levels and types of required tasks intuitively will take a fair amount of effort.

vrooje:
Should note that on my laptop screen it is still quite easy for me to place a mark on a subject in a position such that the subtask window falls off the bottom of the screen and I can't scroll further down to see the "OK" button. It's a bit harder, but still possible, to mark in a position such that the "This step is required" text is not fully visible in the sub-task window. (I'm testing this in Fossil Finder.)

Is it worth moving the required label to the top of the window?

parrish:
Is this issue still valid?

PR #1641, opened 2015-09-25
Require subtasks

brian-c:
For #1563:

This disables the "OK" button in the drawing subtasks popup until required subtasks are complete.

It does **not** prevent the user from interacting with the page (this would make it impossible for users to modify or delete the selected mark, for example) so it's still possible to create a new mark while the current subtasks are incomplete.

I believe preventing this would require a fairly big rewrite of how drawing is handled. We can do that in the future.

Instead I've darkened the page behind the subtasks popup to discourage interaction until the "OK" button is pressed, and I've added a "this task is required" note to any required tasks.


chrissnyder:
Really nice. Big step in the right direction.

PR #1415, opened 2015-08-14
Add aggregations export button

edpaget:
Also add ready state to buttons to indicate new data is being generated,
closes #1257

PR #1643, opened 2015-09-25
add aggregation data export button

camallen:
wires up aggregation exports.

** Don't merge this till @ggdhines has ironed out the kinks in the aggregation engine **


camallen:
Question, should we only enable this for internal testing at the moment? Not sure the aggregation engine is ready for public release just yet..

chrissnyder:
Duplicate of #1415, which is still waiting on the aggregation engine.

camallen:
Ha! silly me.

PR #1646, opened 2015-09-25
Treat empty string redirects as not having one

### LINKED ISSUE #1645: Sidebar project Talk links don't work
The only one that seems to be working is Galaxy Zoo: Bar Lengths.

Everything else just leaves you on the current page.

Sometimes even Bar Lengths doesn't work.

Is anyone else also experiencing this? I tested it in Firefox and Chrome.

aweiksnar:
- Closes #1645 
- Example on dev is fossil-finder

### DIFF
--- a/app/talk/lib/project-linker.cjsx
+++ b/app/talk/lib/project-linker.cjsx
@@ -47,7 +47,7 @@ module?.exports = React.createClass
     [owner, name] = project.slug.split('/')
 
     <div key={project.id}>
-      {if project.redirect?
+      {if project.redirect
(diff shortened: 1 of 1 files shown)


chrissnyder:
Under what scenario would a blank redirect string be OK here? Wouldn't that lead the redirect link to go nowhere?

aweiksnar:
This treats a redirect of `null` or `""` as not having one, so those can link to the project talk instead of main talk

aweiksnar:
Just changed the commit message to be more descriptive

chrissnyder:
Ah ok, I totally read that wrong, sorry.

aweiksnar:
:bow:

Issue #1618, opened 2015-09-22
Set a minimum size for "click and drag" annotations


CKrawczyk:
The "click and drag" annotations (box, circle, ellipse) can be added without dragging, when this happens the resulting annotation can't be resized or removed easily.  If the resulting annotation is below a certain size is should be automatically removed.

PR #1656, opened 2015-09-28
fixes issues 1618

CKrawczyk:
Fixes issues https://github.com/zooniverse/Panoptes-Front-End/issues/1618.  Circles/ellipses/boxes that fail initValid are automatically removed.


brian-c:
Good catch, thanks!

Issue #1581, opened 2015-09-17
Linking an image to a comment


parrish:
![linked-image](https://cloud.githubusercontent.com/assets/122990/9940669/da8024ba-5d36-11e5-81ac-f8cb0698cfa1.gif)

It feels like a non-obvious step to find the image, hover it, then select it.

I think we should either add some text explaining this or (ideally) automatically select the image.

@aweiksnar Thoughts?

parrish:
Also, @DarrenMcRoy have you heard any complaints/feedback about this?

DarrenMcRoy:
Nobody has brought it up that I know of.

But it might be cool that if you link an image, or mention an ID, the tool suggests that you use that as a linked image.

DarrenMcRoy:
How would you theoretically do it otherwise? Willing to defer on the UX to you if you think there's a better way.

parrish:
A couple ideas to improve it:
  - prompt the user to click the image to select (instead of hover, discover the button, then click it)
  - select the image as soon as it is found (by typing in the search box) -- click clear to remove it
  - add a button to search instead of requiring the user to hit enter (button also implies enter as an option)
  - autocomplete search results
  - pre-populate with multiple recently seen / favorited / collected subjects

aweiksnar:
Yeah agree all-around here. I may play with this in a modal too

DarrenMcRoy:
Semi-urgent: Image linking appears to have completely broken. Did we change something? @aweiksnar @parrish 

![screen shot 2015-09-18 at 10 43 55 am](https://cloud.githubusercontent.com/assets/8987864/9963661/fb48094a-5df1-11e5-9794-3a8233918f75.png)

![screen shot 2015-09-18 at 10 44 06 am](https://cloud.githubusercontent.com/assets/8987864/9963669/00b3e73c-5df2-11e5-8fa5-9035c4d6fb9e.png)

parrish:
Sort of.  If you select a recent subject, it's posting with the recent id instead of the subject id.

parrish:
@aweiksnar That code hasn't changed, but maybe the API did?  Should we just follow the links from the recents to get the subjects?  We can get the subjects via an include, but I don't know how the api client exposes included/linked records.

aweiksnar:
@parrish Just issued a PR with that. Working for me on everything except 'kitteh zoo' subjects, for which the api recents links are to subjects in a different project

parrish:
[...]

PR #1659, opened 2015-09-28
Add image search submit button, improve selection

aweiksnar:
- Some suggestions from #1581
- Adds a dedicated 'search' button to the image selector form
- If there is one match that comes back it will automatically set it as the focus image


parrish:
:+1:

PR #1660, opened 2015-09-28
Supply frame for thumbnails src in discussion link

### LINKED ISSUE #1649: Broken thumbnails for discussion preview in Talk
I'm testing subject comments for the Wellcome project. When I list comments in Talk, the discussion preview thumbnail is trying to use the `text/plain` subject location, even though there is also a location with a MIME type of `image/jpeg`.
https://www.zooniverse.org/projects/eatyourgreens/diagnosis-london/talk/100/

aweiksnar:
- Closes #1649

### DIFF
--- a/app/lib/get-subject-location.coffee
+++ b/app/lib/get-subject-location.coffee
@@ -2,9 +2,7 @@ READABLE_FORMATS =
   image: ['jpeg', 'png', 'svg+xml', 'gif']
   video: ['mp4']
 
-module.exports = (subject, frame) ->
-  frame ?= Math.floor Object.keys(subject.locations).length / 2
-
+module.exports = (subject, frame = 0) ->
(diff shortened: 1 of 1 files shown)


brian-c:
Works for me, but should we just make the default `0`?

The default was set to the middle one because in Snapshot Serengeti, an animal moving in front of the camera would presumably be most likely centered in the middle image. Not sure if that ended up being true or if it matters.

aweiksnar:
Cool, yeah that sounds good me. Just pushed to make that function have a default

brian-c:
Sorry, mind deleting the old default?

aweiksnar:
Yeah where at? I think that was the only one

aweiksnar:
Ah sorry! thought you meant in the original call itself. Just took it out

brian-c:
Thanks!

PR #1663, opened 2015-09-28
allow admin to set the project external redirect

camallen:
save api devs work...allow most of the admin stuff to be done here.

### DIFF
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -65,6 +65,45 @@ WorkflowToggle = React.createClass
       </label>
     </span>
 
+ProjectRedirectToggle = React.createClass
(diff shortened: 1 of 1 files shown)


edpaget:
I think this would actually work better with the `<Autosave>` component since that will give feed back around errors, etc.

camallen:
so i've used the autosave component https://github.com/camallen/Panoptes-Front-End/commit/5852f574da21aa4b91733eadcfa1a5081d46b2ab but this ends up firing off a bunch a project version GETs on the text input change events that update the internal project redirect state. @edpaget thoughts on restricting these to only run on load?

camallen:
@edpaget - should be using the updated version component and the autosave for redirect now.

edpaget:
:shipit:

PR #1664, opened 2015-09-28
Defer moderation loading until opened

### LINKED ISSUE #1556: Defer moderation panel content loading until opened
https://github.com/zooniverse/Panoptes-Front-End/commit/073ef7ccb47672b4435be65e337d824ae57ef794#commitcomment-13191388

aweiksnar:
- The moderation component does not hold the button / open-closed state anymore
  - It just renders children if the supplied user is a moderator for the section
- This looks like a big diff, but is mostly markup indentation
- Closes #1556

### DIFF
--- a/app/talk/board.cjsx
+++ b/app/talk/board.cjsx
@@ -32,6 +32,7 @@ module?.exports = React.createClass
     discussionsMeta: {}
     newDiscussionOpen: false
     loading: true
+    moderationOpen: false
 
   getDefaultProps: ->
     query: page: 1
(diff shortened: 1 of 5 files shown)


parrish:
https://github.com/zooniverse/Panoptes-Front-End/pull/1664/files?w=1

where `?w=1` ignores whitespace.

aweiksnar:
Heh, that's a great trick

parrish:
Seems to work for me.  I do see a couple warnings now though:

- On the board moderation: Warning: _renderNewRootComponent(): Render methods should be a pure function of props and state; triggering nested component updates from render is not allowed. If necessary, trigger nested updates in componentDidUpdate.
- On the discussion moderation: Warning: Use the \`defaultValue\` or \`value\` props on \<select\> instead of setting \`selected\` on \<option\>

parrish:
If we're not concerned about that I'll go ahead and merge.  (not sure if we care)

aweiksnar:
Just fixed up that select warning. I traced the `_renderNewRootComponent` warning to how I'm generating the plain-text comment previews [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/latest-comment-link.cjsx#L16-L19). I'll tackle that in another PR so I can think a bit on the best way to do that or maybe add something to `markdownz` for it

parrish:
:+1:

aweiksnar:
Thanks!

PR #1670, opened 2015-09-29
Gold standard mode toggle

brian-c:
This adds a section to the classification interface below the task for expert users to change expert classification settings.

Currently gold-standard mode is the only option. A demo/throwaway mode should be coming soon.

@aliburchard: In demo mode, did you want the classifications saved with a `throwaway: true` flag, or should they not be saved at all?

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -6,6 +6,7 @@ ClassificationSummary = require './classification-summary'
 tasks = require './tasks'
 {getSessionID} = require '../lib/session'
 preloadSubject = require '../lib/preload-subject'
+TriggeredModalForm = require 'modal-form/triggered'
 
 unless process.env.NODE_ENV is 'production'
   mockData = require './mock-data'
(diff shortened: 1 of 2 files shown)


aliburchard:
@brian-c I would lean towards throwing them away, since that's explicitly why we are in the demo mode, and it might make it easier to cut ties with "throwaway" subject sets that were uploaded specifically for testing? But I could be convinced otherwise.

aweiksnar:
Looks nice! Definitely a good thing that it resets in between classifications

brian-c:
Ha, yeah, I had it persisting at first but then I realized that at some point somebody's going to forget it's set and everything'll be ruined forever. Whoops.

aweiksnar:
Haha yeah that'd be no fun

aliburchard:
Does it reset between every classification? The use cases I'm thinking of have experts doing a batch of, say 50 or so gold standard classifications in a row - having to set "gold mode" in between every classification would be...prohibitive. Can we instead add a MASSIVE alert saying "you are in gold mode" when someone is classifiying in gold mode?

aliburchard:
@brian-c I actually can't find this toggle anywhere....perhaps because owners can't seem to set themselves as experts on projects, which I'll open as another issue.

PR #1671, opened 2015-09-30
ui update for aggregation experimental feature

camallen:
remove admin only view for aggregation dumps, instead change the UI to signal this a new experimental feature.

### DIFF
--- a/app/pages/lab/data-dumps.cjsx
+++ b/app/pages/lab/data-dumps.cjsx
@@ -10,7 +10,7 @@ isAdmin = require '../../lib/is-admin'
 counterpart.registerTranslations 'en',
   projectDetails:
     classificationExport: "Request new classification export"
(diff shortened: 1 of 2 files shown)


edpaget:
screen shot 2015-09-30 at 9 07 32 am

I think the experimental UI maybe a little extreme...

camallen:
Yeah - what do you think about just removing the blue highlight on "Experimental new feature"? I want to signal that this may not work properly and that it applies to the aggregation button...

edpaget:
How about adding an *experimental* to the button?

aweiksnar:
http://fontawesome.io/icon/flask/ may work here too

camallen:
thanks @aweiksnar used a gear icon..look better?

edpaget:
:ok_hand:

PR #1672, opened 2015-09-30
Board positions

### LINKED ISSUE #1665: All the Talk message boards have changed order both on Zooniverse Talk and the project talks
I'm not sure if this was supposed to happened but all the Talk boards are in different places. I had set up of P4: Terrains so that the Help thread was first so people saw it without having to scroll if they had smaller screen. Nowt the help board is on the bottom and the notes thread is the top. 

Same thing on Zooniverse Talk the boards are all in a different order. Looks like the boards are being list by the last time there was a post. 

@parrish and @aweiksnar if this a new feature, is there a way to pin the boards so their order doesn't change? I find the reordering very confusing.

parrish:
This allows boards to be ordered by most recent activity or in a static/manual order.

Closes #1665.

This is currently broken in Firefox until zooniverse-ui/drag-reorderable#1 is merged.

Also, touch support is currently missing.

### DIFF
--- a/app/talk/init.cjsx
+++ b/app/talk/init.cjsx
@@ -16,6 +16,7 @@ require '../api/sugar'
 ZooniverseTeam = require './lib/zoo-team.cjsx'
 alert = require '../lib/alert'
 AddZooTeamForm = require './add-zoo-team-form'
+DragReorderable = require 'drag-reorderable'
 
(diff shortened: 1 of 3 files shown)


aweiksnar:
:+1: :shipit: :+1:

PR #1673, opened 2015-09-30
Throwaway classifications

brian-c:
Admins and project owners and collaborators can now enter "demo mode", where classifications aren't actually saved.

I moved the demo mode and gold standard toggles into their own dropdown in the classification interface.

Demo mode persists across classifications (presumably presenters will want to go through a few), but there's a little warning in the interface where you can turn it off.


aweiksnar:
LGTM

Issue #1669, opened 2015-09-29
Error when setting retirment limit using front-end


CKrawczyk:
I get a "422 (Unprocessable Entity)" when I try to change the retirement limit for a project using the front-end:

<img width="1420" alt="screen shot 2015-09-29 at 5 13 51 pm" src="https://cloud.githubusercontent.com/assets/6557526/10170591/8b7ce0ec-66cd-11e5-820a-c51f48301cba.png">

I see the same error in both Firefox and Chrome.

edpaget:
Can you look at the response body in the Network panel?

CKrawczyk:
The response is:

{"errors":[{"message":"{\"retirement\"=\u003e\"contains additional properties [\\\"classification_count\\\"] outside of the schema when none are allowed\"}"}]}

edpaget:
Looks like the front-end is sending back retirement criteria incorrectly.

CKrawczyk:
I was playing around with this today and I noticed that I get the same error message when trying to change the retirement limit directly with the API:

```js
workflow.update({'retirement.classification_count.count': 10}).save()
```
gives
```js
Error: {"retirement"=>"contains additional properties [\"classification_count\"] outside of the schema when none are allowed"}
```
Is the API expecting this limit to be set in a different way?  Or could this be a bug in the API that is preventing this limit from being set?

camallen:
what does the network payload look like, the api is expecting something like 
`{ workflows: { retirement: { criteria: "classification_count",  options: { count: 10 } } } }`

CKrawczyk:
the network payload is
```js
{workflows: {retirement: {classification_count: {count: 10}}}}
```
OK, I see the issue now, the workflow is retrieved in a different format than the one it expects to take in.

edpaget:
This was fixed via https://github.com/zooniverse/Panoptes/pull/1402

PR #1679, opened 2015-10-01
Fix 1669

CKrawczyk:
This fixes https://github.com/zooniverse/Panoptes-Front-End/issues/1669

This makes the front-end work as expected, but it does not address the root cause of this problem.  The reason this is happening is when the API creates a new workflow the retirement section is not formatted correctly.  It returns
```js
{workflows: {retirement: {classification_count: {count: 15}}}
```
instead of
```js
{workflows: {retirement: {criteria: "classification_count", options: {count: 15}}}}
```
This should probably be addressed in the back-end code.


camallen:
@CKrawczyk thanks, I've issued a PR to fix the default format in the backend.

edpaget:
Fixed via https://github.com/zooniverse/Panoptes/pull/1402

CKrawczyk:
https://github.com/zooniverse/Panoptes/pull/1402 only fixes this issue for newly created workflows, any workflow made before the fix will still fail.  Either a patch similar to this one is needed to check on the front end if the workflow is formatted correctly, or a one time code needs to run to update all the older workflows.

camallen:
@CKrawczyk - it should fix the default serialized format so all workflows should be fixed. Have you got an example workflow that is still broken?

CKrawczyk:
When the workflow editor is working again I will give it a try.

CKrawczyk:
Yep, you are right, everything works as intended now.

PR #1683, opened 2015-10-01
Revert "Update auth client to handle response from superagent"

chrissnyder:
Reverts zooniverse/Panoptes-Front-End#1439

Merged too quickly. With additional testing, json-api-client is not passing errors back correctly, resulting in confusing or otherwise incorrect error messages. Try registering a user with a username thats already taken, or inputting a login and compare to live.


chrissnyder:
I cancelled the deploy so it's not going to break production. Working on a fix now.

PR #1689, opened 2015-10-01
Moved close button to outside of dialog

### LINKED ISSUE #1296: Dialog Controls overlap scroll bar

srallen:
- Fixed #1296 by moving the close button to the right of the dialog container

### DIFF
--- a/css/dialog.styl
+++ b/css/dialog.styl
@@ -3,9 +3,12 @@
   z-index: 1
 
 .dialog-underlay
+  align-items: center
   background: rgba(black, 0.6)
   color: white
+  display: flex
   height: 100%
+  justify-content: center
   left: 0
   position: fixed
   top: 0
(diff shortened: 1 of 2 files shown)


brian-c:
Now the close button disappears on mobile.

Can we not solve this by just nudging it over to the left like, half an em?

srallen:
I could do that, but it looks odd for dialogs that don't need to scroll.

I just updated the PR and added a media query so the fixed positioning on the close button only applies to screens larger than 768px. Also, fixed a responsive issue in the footer. Overall a lot could be done and the dialogs still don't look right at least in iOS safari since the header responsiveness needs fixing as well as dialogs showing tables (i.e. the metadata dialog in the classifier), but it makes sense to do a pass on those things in separate PRs.

brian-c:
This still allows for situations where the close box is positioned offscreen. How about just moving it to the top side?

srallen:
Ok, update PR. I didn't see where that was happening, but I guess I missed it.

brian-c:
Works for me. Thanks!

PR #1693, opened 2015-10-02
Added pagination for email prefs. Fixed styling on project pagination

srallen:
- Added pagination to email preferences in settings modeled after how it's done in the project builder and on talk
- Fixed a styling issue for the project page pagination when on the first page


brian-c:
Looks good to me.

PR #1702, opened 2015-10-05
Use local node modules for testing, rather than assuming global installs

itsravenous:
`npm test` requires browserify, testling and tap-spec to be available in the users `PATH` (i.e. installed globally), but this isn't mentioned in the README. I suggest using the local modules, which has two benefits:

- No global installs required
- Control over the version of the testing tools

I'm not sure how commonplace this is in node development, but it's a pattern I've recently adopted for the above reasons.

If this isn't to the project's taste, however, I can always create a PR adding the install lines to the README.

(Confession: I haven't managed to get the tests to run properly in any case, but I think that's because I haven't set up the config to point to a local backend)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
     "stage": "NON_ROOT=true NODE_ENV=staging npm run build && publisssh ./build \"demo.zooniverse.org/panoptes-front-end/$DEPLOY_SUBDIR\"",
     "stage-branch": "NON_ROOT=true DEPLOY_SUBDIR=$(git symbolic-ref --short HEAD) npm run stage",
     "start": "./bin/serve.sh",
-    "test": "browserify ./test/runner.coffee --extension .coffee --extension .cjsx --transform coffee-reactify --transform envify --ignore-transform coffeeify | testling \"${TESTLING_ARGS:---pass}\" | tap-spec",
+    "test": "node node_modules/browserify/bin/cmd ./test/runner.coffee --extension .coffee --extension .cjsx --transform coffee-reactify --transform envify --ignore-transform coffeeify|node node_modules/testling/bin/cmd \"${TESTLING_ARGS:---pass}\" | node node_modules/tap-spec/bin/cmd",
     "test-mac": "TESTLING_ARGS=\"-x open --new -a Google\\ Chrome --args --incognito\" npm run test"
   },
   "config": {


camallen:
@itsravenous did you think about getting the local packages available via PATH? E.g. http://stackoverflow.com/questions/9679932/how-to-use-package-installed-locally-in-node-modules

itsravenous:
Ooh, I hadn't come across `npm bin`; that's useful and feels cleaner. I think you need to restate `$PATH` for each piped command though, so the cmd would be:

```
PATH=$(npm bin):$PATH browserify ./test/runner.coffee --extension .coffee --extension .cjsx \
    --transform coffee-reactify --transform envify --ignore-transform coffeeify | \
    PATH=$(npm bin):$PATH testling "${TESTLING_ARGS:---pass}" | PATH=$(npm bin):$PATH tap-spec
```

Alternatively, backticks would provide brevity:

```
`npm bin`/browserify ./test/runner.coffee --extension .coffee --extension .cjsx \
    --transform coffee-reactify --transform envify --ignore-transform coffeeify | \
    `npm bin`/testling "${TESTLING_ARGS:---pass}" | `npm bin`/tap-spec
```

I don't think either of the above will work on Windows, but I don't know whether that's a priority for you guys. If it is, my original proposal should work cross-platform.

brian-c:
Commands in package.json's scripts list should run with add the node_modules/.bin directory to the PATH automatically. Are you finding this is not the case?

Windows has never been a priority, but of course it'd be nice if it worked.

itsravenous:
More things I didn't know! That makes perfect sense too. I mistakenly thought the `ENOENT` error I got when running `npm test` was down to it not finding the commands, but that obviously doesn't make sense. Closing this; apologies for taking up people's time but thanks for the education ;)

edpaget:
Yeah @itsravenous the last time I tried getting this running on windows, I was unable to (and now longer have a computer with windows to try it out).

itsravenous:
Hmm, just tried in my [Modern IE VM](https://dev.modern.ie/tools/vms/linux/) and npm install fails because the `get-fallback-polyfills` script fails:

```
> panoptes@0.0.2 postinstall C:\Users\IEUser\dev\Panoptes-Front-End
> npm run get-fallback-polyfills && npm run get-font-awesome


> panoptes@0.0.2 get-fallback-polyfills C:\Users\IEUser\dev\Panoptes-Front-End
> curl "$npm_package_config_fallbackPolyfillsURL" > ./public/fallback-polyfills.js

curl: (6) Couldn't resolve host '$npm_package_config_fallbackPolyfillsURL'
```

Looks like the variable doesn't expand on Windows (this is in Git Bash). I can open a separate issue for this and investigate if Windows support is needed in any capacity?

edpaget:
You're welcome to if you want to maintain the windows support. I think everyone here develops on OS X or Linux, so we'd probably just break it again without someone checking.

itsravenous:
I think I'll pass then (using Linux here).

PR #1707, opened 2015-10-07
Use response meta field over query param for collection pagination

### LINKED ISSUE #1284: Pagination broken in collections
If I hit the next button it takes me to the top of the page, but the page is still the first page. If instead I manually change the number of the page to the next one it works and then the next and back buttons work.

chrissnyder:
Closes #1284


aweiksnar:
:horse_racing: :shipit:

PR #1711, opened 2015-10-07
Show newest subject comments first

### LINKED ISSUE #1706: Flipping the chronology on the subject page so the most recent comment shows up first or posting a few most recent comments on the top
Checking a subject page every notes post from different notes threads are individually posted in chronological order with oldest first, but since people quickly glance at this, they don't see if the research team corrects terminology or something recently posted.

Although it is paginated, I don't think people will read through the entire discussion when classifying. Notes as I see it is for people to quickly comment on a subject. So I see the behavior as more of a quick glance at comments and then writing in the box and going back to classify the next image. 

With the 140 character comments on the previous version of Talk, the posts had the most recent listed first, and I found that helpful because if someone said 'there's tar- life on Mars', I could write 'no, the dark matches are X. Read more here', and the next person classifying and on Talk would see that. 

If you don't want to completely flip the order, would it be possible to post the 2-5 most recent comments, and then the rest in order from oldest to newest?

Example [here](https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/subjects/490275/)

aweiksnar:
- Closes #1706

### DIFF
--- a/app/subjects/index.cjsx
+++ b/app/subjects/index.cjsx
@@ -55,7 +55,7 @@ module?.exports = React.createClass
 
   setComments: (subject = @state.subject, page = @props.query.page ? 1) ->
     talkClient.type('comments')
-      .get({focus_id: subject.id, focus_type: 'Subject', page})
+      .get({focus_id: subject.id, focus_type: 'Subject', page, sort: '-created_at'})
       .then (comments) =>
         commentsMeta = comments[0]?.getMeta()
         @setState {comments, commentsMeta}


chrissnyder:
Works, but it's pretty much guaranteed that someone will complain that they liked the prior behavior better. What's our stance on why we need this?

aweiksnar:
This will make the subject page have fresh content that continually updates, instead of it displaying the first 10 comments for its lifetime. That will make it a more interesting page to check. Also helps differentiate it from the subject discussions, which can be quite similar if the subject is not discussed much elsewhere

chrissnyder:
Sounds good to me.

PR #1713, opened 2015-10-07
Admin allow to set user limits

camallen:
started on an admin user page, it only currently allows subject_limits to be fetched / updated (in admin mode). 

One note: and you can't edit yourself as it got weird when updating props on the logged in user (avatar req's fired off and i couldn't type into the limit box. Any thoughts on this appreciated and happy to mould this any way you like.

### DIFF
--- a/app/pages/admin.cjsx
+++ b/app/pages/admin.cjsx
@@ -8,7 +8,7 @@ counterpart.registerTranslations 'en',
   userAdminPage:
     header: "Admin"
     nav:
-      createAdmin: "Manage Administrators"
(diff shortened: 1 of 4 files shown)


chrissnyder:
Getting a 404 to the users PUT route. Panoptes issue?

camallen:
shouldn't be since https://github.com/zooniverse/Panoptes/pull/1413, you'll need to have admin mode on though.

chrissnyder:
:neutral_face: Of course, I'm dumb.

PR #1715, opened 2015-10-07
Provide fallback card image

aweiksnar:
Adds a fallback for projects without an avatar. Noticed these went away at some point but couldn't find anything indicating why in the git history


chrissnyder:
:+1:

Issue #1226, opened 2015-07-27
My headshot displays sideways on iPad/iPhone Safari


DarrenMcRoy:
Presented without comment:

![sideways_darren](https://cloud.githubusercontent.com/assets/8987864/8897866/dec6b0e4-33e1-11e5-8e59-541a03f45598.png)

A user reported spotting this on her iPad Safari, and I can replicate it in my iPhone Safari. Any idea?

PR #1717, opened 2015-10-08
Small CSS fix to ensure vertical alignment of each member's image…

alexbfree:
...alongside their row (important for 4-line descriptions)

### DIFF
--- a/css/team-page.styl
+++ b/css/team-page.styl
@@ -16,6 +16,8 @@
         border-radius: 4px
         height: 80px
         width: 80px
+        margin-top: auto
+        margin-bottom: auto
 
       .team-member-details
         flex-basis: 75%


chrissnyder:
Without playing designer, I disagree this is needed. I like the current behavior better.

alexbfree:
OK well let's get a designer's opinion then - @jwbmartin what do you think? My view is that since there is no delineation between rows, the spacing looks messed up when descriptions flow onto four lines... e.g. see Chris Doogue's row:
screenshot

jwbmartin:
I do agree that it is a little bothersome from a design perscpective, and this spacing changes depending on the size of the screen too. But on the list of things that need to be fixed this is probably not an extremely pertinent issue.

DarrenMcRoy:
If it does ever get changed, maybe someone could look at #1226 too? :-)

PR #1731, opened 2015-10-09
Changing all instances of 'beta' to 'review'

mrniaboc:


chrissnyder:
Could https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/lab-policies.cjsx be updated as well?

mrniaboc:
Ah, well spotted! Will do.

PR #1736, opened 2015-10-09
Notifications

parrish:
This is a somewhat tentative PR to add notifications.

I'll be adding line notes shortly.

@brian-c @aweiksnar Feedback welcome!


app/components/notification-viewer.cjsx
```diff
@@ -1,53 +0,0 @@
-React = require 'react'
```
parrish: I'm removing the popups for now to reduce confusion, but they may reappear when we get to it.

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+
+  propTypes:
+    project: React.PropTypes.object
+    user: React.PropTypes.object
+
+  getDefaultProps: ->
+    query: page: 1
+
+  getInitialState: ->
+    firstMeta: { }
```
parrish: Tracks meta for the top and bottom paginator separately.

This lets us increase the bottom page and top page independently.

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+    lastMeta: { }
+    notificationsMap: { }
```
parrish: Maps notification id to notification resource to allow an easy way to check if a notification is already delivered when marking it as read

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+
+  componentWillMount: ->
+    @getNotifications() if @props.user
+
+  componentWillUnmount: ->
```
parrish: Mark the current top and bottom page of notifications as read when leaving this page.

This solves the problem of "reading" the first and last page of notifications

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+
+  getNotifications: (page = @props.query.page) ->
+    page or= 1
+    query = {page}
+    query.section = "project-#{ @props.project.id }" if @props.project
+    query.section = @props.params.section if @props.params.section
+    talkClient.type('notifications').get(query).then (newNotifications) =>
+      meta = newNotifications[0]?.getMeta() or { }
+      notifications = @state.notifications or newNotifications
+      meta.notificationIds = (n.id for n in newNotifications)
```
parrish: Working with ids instead of objects here to prevent reference copies

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+      notificationsMap = @state.notificationsMap
+
+      for notification in newNotifications
+        notificationsMap[notification.id] = notification
+
+      {firstMeta, lastMeta} = @state
+
+      if meta.page > @state.lastMeta.page
+        lastMeta = meta
```
parrish: Got a new bottom page, it's appended to the list

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+        notifications.push newNotifications...
+      else if meta.page < @state.firstMeta.page
+        firstMeta = meta
+        notifications.unshift newNotifications...
```
parrish: Got a new top page, it's prepended to the list

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+      else
+        firstMeta = lastMeta = meta
+
+      @setState {notifications, notificationsMap, firstMeta, lastMeta}
+
+  markAsRead: (meta) ->
+    =>
```
parrish: Allows a closure to determine which paginator page we're marking as read.

This feels a bit hacky, but react does some funky binding stuff that I had to work around.

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+      ids = @state["#{ meta }Meta"].notificationIds
+      ids = (id for id in ids when not @state.notificationsMap[id].delivered)
```
parrish: Pull out a list of notification ids for the paginator that is getting marked as read

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+      return if ids.length is 0
+      talkClient.put '/notifications/read', id: ids.join(',')
+      for notification in @state.notifications when notification.id in ids
+        notification.update delivered: true
```
parrish: Since the API calls operates on a batch of records, I'm just updating the attribute on the resource, but not sending the PUT

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+
+  markAllAsRead: ->
+    talkClient.put '/notifications/read'
+    for notification in @state.notifications
+      notification.update delivered: true
+    @forceUpdate()
```
parrish: `/notifications/read` without any ids specified marks them all as read.

For some reason, react doesn't always pick up a change here, so I'm forcing the update.

app/pages/notifications.cjsx
```diff
@@ -0,0 +1,145 @@
...
+          {@title()}
+        </h1>
+
+        {if @props.user?
+          <ChangeListener target={@props.user}>{ =>
+            if @state.notifications?.length > 0
+              <div>
+                  <div className="centering">
+                    <div className="talk-module inline-block">
+                      {if @state.firstMeta.page > 1
```
parrish: Only render the top paginator if we're not starting on the first page of notifications

app/pages/notifications/data-request.cjsx
```diff
@@ -0,0 +1,35 @@
...
+
+  getInitialState: ->
+    projectName: null
+
+  componentWillMount: ->
+    apiClient.type('projects').get(@props.notification.project_id).then (project) =>
+      @setState projectName: project.display_name
+
+  stopPropagation: (e) ->
+    e.stopPropagation()
```
parrish: This is necessary as I think react was trying to handle the click event.

app/pages/notifications/message.cjsx
```diff
@@ -0,0 +1,51 @@
...
+  getInitialState: ->
+    message: null
+    messageUser: null
+    conversation: null
+
+  componentWillMount: ->
+    talkClient.type('messages').get(@props.notification.source_id, include: 'conversation,user').then (message) =>
+      message.get('user').then (messageUser) =>
+        message.get('conversation').then (conversation) =>
+          @setState {message, conversation, messageUser}
```
parrish: This is actually only one API call since the user and conversation is sideloaded.

If there's an easier way to do this with the client, I'd be happy to change it

app/pages/notifications/moderation.cjsx
```diff
@@ -0,0 +1,84 @@
...
+    reports: []
+
+  componentWillMount: ->
+    talkClient.type('moderations').get(@props.notification.source_id).then (moderation) =>
+      comment = moderation.target or moderation.destroyed_target
+      @setState {moderation, comment}
+
+      promises = []
+      for report in moderation.reports then do (report) =>
+        promises.push apiClient.type('users').get(report.user_id.toString(), { }).then (user) =>
```
parrish: Since a destroyed comment is just an embedded JSON copy of what the comment was, it doesn't get formatted by the serializer -- the id's are going to be integers

app/pages/notifications/notification.cjsx
```diff
@@ -0,0 +1,32 @@
+React = require 'react'
+CommentNotification = require './comment'
+DataRequestNotification = require './data-request'
+MessageNotification = require './message'
+ModerationNotification = require './moderation'
+
+module?.exports = React.createClass
```
parrish: Basically just a dispatcher to render the right type of notification

app/partials/main-header.cjsx
```diff
@@ -56,6 +57,7 @@ module.exports = React.createClass
           <Link to="projects" className="main-nav-item"><Translate content="mainNav.projects" /></Link>
           <Link to="about" className="main-nav-item"><Translate content="mainNav.about" /></Link>
           <Link to="talk" className="main-nav-item"><Translate content="mainNav.talk" /></Link>
+          <Link to="notifications" className="main-nav-item"><Translate content="mainNav.notifications" /></Link>
```
parrish: This links to notifications from all projects/sections.

I imagine we'd eventually integrate this into the personalized sign in page, but until we get to it I'm guessing it's a top level link.

app/router.cjsx
```diff
@@ -44,6 +44,7 @@ module.exports =
     <Route name="project-research" path="research/?" handler={require './pages/project/research'} />
     <Route name="project-results" path="results/?" handler={require './pages/project/results'} />
     <Route name="project-classify" path="classify/?" handler={require './pages/project/classify'} />
+    <Route name="project-notifications" path="notifications/?" handler={require './pages/notifications'} />
```
parrish: per-project notifications

app/router.cjsx
```diff
@@ -59,6 +60,9 @@ module.exports =
     <Route name="project-education" path="education/?" handler={require './pages/project/education'} />
   </Route>
 
+  <Route name="notifications" path="notifications/?" handler={require './pages/notifications'} />
```
parrish: all notifications

app/router.cjsx
```diff
@@ -59,6 +60,9 @@ module.exports =
...
+  <Route name="section-notifications" path=":section/notifications/?" handler={require './pages/notifications'} />
```
parrish: non-project section notifications

app/talk/latest-comment-link.cjsx
```diff
@@ -9,15 +9,10 @@ merge = require 'lodash.merge'
 
 PAGE_SIZE = require('./config').discussionPageSize
 
-truncate = (string, ending = '', length = 80) ->
+truncate = (string = '', ending = '', length = 80) ->
   return string if string.trim().length <= length
   string.trim().slice(0, (length - ending.length)) + ending
 
-latestCommentText = (discussion) ->
```
parrish: This was removed because mutating the DOM during a render was doing some really weird stuff.

In some cases, it would cause a re-render infinite loop

app/talk/latest-comment-link.cjsx
```diff
@@ -31,15 +26,29 @@ module?.exports = React.createClass
...
 
+  componentWillMount: ->
+    comment = @props.comment or @props.discussion?.latest_comment
+    return unless comment
+    apiClient.type('users').get(comment.user_id).then (commentUser) =>
+      @setState {commentUser}
+
+  componentDidMount: ->
+    latestCommentText = @refs?.markdownText?.getDOMNode()?.textContent
+    @setState({latestCommentText}) if latestCommentText
```
parrish: Instead of manually creating a DOM node, I'm just rendering the component invisibly and pulling the `textContent` from that.

app/talk/latest-comment-link.cjsx
```diff
@@ -49,11 +58,16 @@ module?.exports = React.createClass
...
-        <PromiseRenderer promise={apiClient.type('users').get(comment.user_id, {})}>{(user) =>
-          <Link className="user-profile-link" to="user-profile" params={name: user.login}>
-            <Avatar user={user} />{' '}{user.display_name}
-          </Link>
-        }</PromiseRenderer>{' '}
+        <div ref="markdownText" className="hidden-markdown">
+          <Markdown content={comment.body} />
+        </div>
+
+        {if @state.commentUser?
```
parrish: The PromiseRenderer was the source of the re-render loop.

app/talk/lib/paginator.cjsx
```diff
@@ -39,6 +39,8 @@ module?.exports = React.createClass
     firstAndLast: true
     pageSelector: true
     scrollOnChange: true
+    previousLabel: <span><i className="fa fa-long-arrow-left" /> Previous</span>
```
parrish: These are just changes to allow some customization and integration points with the paginator.

app/talk/search-result.cjsx
```diff
@@ -24,14 +24,10 @@ module.exports = React.createClass
...
     <div className="talk-search-result talk-module">
       <CommentLink comment={comment}>{comment.discussion_title}</CommentLink>
       <Markdown content={comment.body} project={@props.project} />
-      {if section is 'zooniverse'
-        <DiscussionPreview {...@props} discussion={discussion} comment={comment} />
-      else
-        [owner, name] = comment.project_slug.split('/')
-        <DiscussionPreview {...@props} discussion={discussion} owner={owner} name={name} comment={comment} />
-      }
+      <DiscussionPreview {...@props} discussion={discussion} owner={owner} name={name} comment={comment} />
```
parrish: Refactor that crept in from when I was tracking down a bug.

I can remove/resend this separately if we care.

app/pages/notifications.cjsx
aweiksnar: Would this work as `do =>`?, since it looks like the closure is always called without args

parrish: How would you set that up?  Another nested function? `do` calls the defined function immediately, right?

That would end up looking like

```coffeescript
markAsRead: (meta) ->
    do (meta) =>
      =>
        # do stuff
```

aweiksnar: I think you could then just treat it as a normal function then, so the invocations could be `@markAsRead('meta')` instead of `@markAsRead('meta')()`  (unless I'm missing something else happening here...)

parrish: If the function was
``` coffeescript
markAsRead: (meta) ->
    do (meta) =>
      alert 'called'
```

It would run on page load since `do` is immediately calling the defined function to create the closure.

aweiksnar: Oh whoops  gotcha :ok_hand: Thanks. I missed that since it's called on mount too

aweiksnar:
Overall looks good to me, Only issue I'm having right now is that I can't get notifications to become marked as 'read'

parrish:
Cool.  I think all of yours were already read.  I just marked them unread.

parrish:
It's also maybe not obvious enough.  It's just the border-color

mark_as_read

aweiksnar:
Ahh okay, It only on DMs that I'm not getting the border, but working great for discussions :+1: . Did you leave those out just for the envelope dot?

parrish:
Ahhh.... as it turns out the account bar is requesting all conversations in your inbox (individually) which marks them as read.

aweiksnar:
Yeah that makes sense now. I think I had made that initially just request the latest 1 since it's just an on/off state
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/account-bar.cjsx#L40-L44

brian-c:
This works for me, although I really don't like that we're adding yet another pagination style instead of fixing the one we've started standardizing.

parrish:
Do you mean talk/lib/paginator, or do you mean pagerz?

brian-c:
Looks like most things use talk/lib/paginator. @aweiksnar is pagerz an abstract version of that we're going to switch to at some point?

parrish:
Okay.  I'm using talk/lib/paginator, I'm just relabeling/hiding some of the elements.

A page transition felt really jarring on this view, but that could've just been me.

aweiksnar:
@brian-c Yeah pagerz will be very similiar to talk/lib/paginator, but will support different UI styles of paging from the same api. It still needs a bit time to finish everything up though

aweiksnar:
@parrish Mind rebasing with master and also possibly hiding or deactivating the mark-all-as-read button if there are no notifications, and then I'll merge this in :ship:

parrish:
Rebased.  The button won't show up if you don't have notifications:

``` coffeescript
{if @props.user?
  <ChangeListener target={@props.user}>{ =>
    if @state.notifications?.length > 0
      # ...
      <button onClick={@markAllAsRead}>
        Mark all as read
      </button>
      # ...
```

aweiksnar:
Cool thanks. So the mark-as-read button should go away whenever clicked? I was thinking it'd only need to show up if any notifications where `delivered` is `false` were present

parrish:
Yeah, the UI doesn't know how many are unread unless it's paged through all of them.

I could add a count to the metadata though.  I'll take a look at doing that.

parrish:
Good suggestion!  Updated

aweiksnar:
:shipit: Let's do this

parrish:
:100: :star:

PR #1741, opened 2015-10-12
Misc fixes

saschaishikawa:
Addresses the following issues:
1. change text "Scientist" to "Researcher"
2. parse markdown in classification summary
3. include timestamp in moderation logs

### DIFF
--- a/css/moderations.styl
+++ b/css/moderations.styl
@@ -5,9 +5,6 @@ hover-button(color)
     background: lighten(color, 5%)
     border-color: color
 
-.moderations-timestamp
-  font-size 12px
-
 .moderations
   section button
     text-transform: capitalize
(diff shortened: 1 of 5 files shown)


css/moderations.styl
```diff
@@ -5,6 +5,9 @@ hover-button(color)
     background: lighten(color, 5%)
     border-color: color
 
+.moderations-timestamp
+  font-size 12px
+
```
chrissnyder: Not used. Should be dropped (or add the class).

chrissnyder:
Cool, looks good.

For housekeepings sake, make sure the relevant issues are closed (not sure if they'll pick up on the commit message), and add "closes X" to the PR description. Like so: https://github.com/zooniverse/Panoptes-Front-End/pull/1639

saschaishikawa:
@chrissnyder: github pays attention to issue numbers referenced in commit messages, so we're good :)

chrissnyder:
That, but it's also a matter of convention. We've followed the unintentional pattern of putting which issues a PR closes in the PR description. Just make sure that pattern is followed in the future.

PR #1742, opened 2015-10-12
Fix talk moderation display bugs.  Closes #1385

### LINKED ISSUE #1385: moderation: weirdness with record of deleted posts
In testing the delete functionality on Talk moderator controls, I noticed that every time you delete a post the moderation controls link to a completely different post afterward (perhaps because the old post is gone?).

Before:

screen shot 2015-08-11 at 17 56 32

After (the above deleted post, and a different one, also on the GZ Bar Lengths boards; I just didn't get a screenshot of it):

screen shot 2015-08-11 at 17 57 16

My report comment is still the same, but the post linked to is on a different board for a different project.

parrish:
- Renders destroyed comments correctly
- Removes invalid state parameter 'all'

### DIFF
--- a/app/talk/moderations.cjsx
+++ b/app/talk/moderations.cjsx
@@ -57,6 +57,7 @@ module?.exports = React.createClass
       {page: page, page_size: PAGE_SIZE, state: @props.query.state},
       if section then {section} else {}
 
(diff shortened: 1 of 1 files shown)


aweiksnar:
Nice!

PR #1743, opened 2015-10-12
Tag search

### LINKED ISSUE #1648: Can't search by hashtag on Talk, returning back a string search of all content when you click on hashtag link
When you click on the hyperlink for a Talk hashtag it takes you to a search result, but that search isn't actually doing any search by the hashtag it returns back any comment when the text minus the # of the hashtag is present. For example if you click on the '#ice' hashtag from the side bar of P4: Terrains talk the search results include these responses which don't have any hashtags.

screen shot 2015-09-26 at 7 12 24 am

screen shot 2015-09-26 at 7 10 36 am

Also other Talk posts that have hashtags which include part of the word are returned rather than the exact '#ice' hashtag

screen shot 2015-09-26 at 7 09 51 am

Advanced users use the hashtag search to do their own reviews and vetting of subjects flagged by other volunteers. I use the hashtag search to find examples and to find what hashtags are useful to get images we use for data quality checks.  It's harder to sift through the results when other content is in there. 

Would it be possible that when you click on the hashtag hyperlink it returns back only results with the hashtag and also allow in the Talk search that if the string starts with the pound sign it indicates you're searching for hashtags and not a string search?

Searching by clicking on the '#mound' and this was returned in the search

screen shot 2015-09-26 at 7 13 23 am

parrish:
Provides a subject search by tag.  Closes #1648

I'm linking project popular tags to this (subjects tagged with #tagname)

Non-project popular tags still link to text search.

I *think* that's probably the right thing to do.

### DIFF
--- a/app/router.cjsx
+++ b/app/router.cjsx
@@ -52,6 +52,7 @@ module.exports =
       <Route name="project-talk-moderations" path="moderations/?" handler={require './talk/moderations'}/>
       <Route name="project-talk-subject" path="subjects/:id/?" handler={require './subjects'}/>
       <Route name="project-talk-board-recents" path="recents/:board/?" handler={require './talk/recents'} />
+      <Route name="project-talk-tags" path="tags/:tag/?" handler={require './talk/tags'} />
       <Route name="project-talk-board" path=":board/?" handler={require './talk/board'} />
       <Route name="project-talk-discussion" path=":board/:discussion/?" handler={require './talk/discussion'} />
     </Route>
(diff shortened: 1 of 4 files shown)


aweiksnar:
Sweet, Looks solid to me. Would you be able to set a max-width on the tag results subject images? example of a large image on dev at `/projects/alex-panoptes/test-proj/talk/tags/hashtags/`

parrish:
Added

aweiksnar:
:shipit:

PR #1745, opened 2015-10-12
Modify error message when adding a collaborator.  Closes #1331

### LINKED ISSUE #1331: "Not allowed to create this role" when adding a collaborator
I just tried to add @vvh to AnnoTate as a collaborator, and got the error `Error: not allowed to create this Role` in the front end.

The console had a 401 error for loading a resource from https://talk.zooniverse.org/roles, and a 400 error from https://panoptes.zooniverse.org/api/project_roles

After refreshing, it looks like she had actually been added.

parrish:


app/pages/lab/collaborators.cjsx
```diff
@@ -119,7 +119,12 @@ CollaboratorCreator = React.createClass
         @props.onAdd? arguments...
 
       .catch (error) =>
-        @setState error: error
+        if error.message.match /not allowed to create this role/i
+          error.message = 'Your account status on this project is still being setup. Please try again later.'
+
+        creating = false
+        @setState {error, creating}
```
chrissnyder: Shouldn't need to set `creating` as the following then block will run.

app/pages/lab/collaborators.cjsx
```diff
@@ -119,7 +119,12 @@ CollaboratorCreator = React.createClass
...
+        @fetchAllRoles()
```
chrissnyder: Doesn't exist in this component. Fire a callback instead?

chrissnyder:
Any trick to testing this/simulating heavier API load? Look fine in abstract, just curious if I'm missing something.

parrish:
Updated.  To simulate the condition of having the panoptes role, but not the talk role, I just removed the talk role manually.

screen shot 2015-10-12 at 2 05 41 pm

chrissnyder:
:+1:

PR #1750, opened 2015-10-13
Add max-width to column under edit-workflow-page

### LINKED ISSUE #1725: Having the task text wrap rather than extending in width in one line in the workflow editor
I thought this had been a ticket, but I can't find it. I'm finding the text I wrote from the working version of the decision tree for my new project is pushing the rest of the workflow editor out such that the scrollbar is covering the resize handle for the help box. Can the text wrap around so it doesn't change the width of the page? 

It looks like it kind wraps, but something about my string is causing it issues

screen shot 2015-10-09 at 2 57 27 pm

screen shot 2015-10-09 at 3 07 10 pm

srallen:
Fixed #1725 by adding max-width to column class when under edit-workflow-page class. Flex-shrink is a bit buggy in Firefox, so this may pop up elsewhere.

### DIFF
--- a/css/workflow-editor.styl
+++ b/css/workflow-editor.styl
@@ -78,3 +78,6 @@
 .edit-workflow-page
   select
     max-width: 100%
+    
+  .column
+    max-width: 500px


chrissnyder:
Does something like `flex-grow: 1` not work? Would like a more general solution than specifying an exact pixel value.

srallen:
It's already inheriting flex-grow and flex-shrink properties because the `column` class is being generically used throughout front end: https://github.com/zooniverse/Panoptes-Front-End/blob/master/css/layout.styl#L31-L35

I'm following the suggestion [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1086218) and [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1114904) by Mozilla and specifying a max-width. 

I could change it to a percentage or viewport width rather than pixel value, but I've is scoped specifically to its use on the workflow editor.

chrissnyder:
Of course, my bad for not looking first.

Limiting width seems fine to me. This might be a case where others will come back with differing opinions, but it's easy to change if folks don't like it.

Issue #1052, opened 2015-07-13
"Not allowed to create this Comment" on Talk


vrooje:
Just got this error message on [a post on Zoo Talk](https://www.zooniverse.org/#/talk/18/549):

![screen shot 2015-07-13 at 17 40 14](https://cloud.githubusercontent.com/assets/2357126/8655045/7635601c-2986-11e5-9f60-024d601dfab2.jpg)

I'd just loaded the thread so it shouldn't have been a timeout, and I reloaded the page, pasted the exact same text, submitted, and it all went fine. Not sure what happened but I thought I should flag it.

DarrenMcRoy:
That's oddly unspecific. @aweiksnar -- any idea what could have happened? Not to pull attention away from the sprint, but in case it's something obvious... ?

mschwamb:
I got that as well - if I tried to subject it again it wouldn't work. I reloaded the page and it worked fine. I had posted in a previous thread right before it. It was a note thread, but it looks like @vrooje's was. 

![screen shot 2015-07-25 at 3 30 52 pm](https://cloud.githubusercontent.com/assets/753325/8888502/96674d5e-32e2-11e5-8ae8-6dad57f2b213.png)

aweiksnar:
It seems like this has been sorted? Error-handling should at least now deliver more descriptive messages here

mschwamb:
Thanks @aweiksnar for looking into it. I haven't seen this error message again.  It might be the same thing as the tokens issues #1318 . I'll reopen if it pops up again.

mschwamb:
Whoops - had thought i had created this issue. @vrooje did. So I'll let her close this. Sorry about that.

vrooje:
Alas, I have just gotten this again, on [Zooniverse Talk](https://www.zooniverse.org/#/talk/18/115?scrollToLastComment=true&page=2)...

![screen shot 2015-08-20 at 17 17 42](https://cloud.githubusercontent.com/assets/2357126/9388785/8db2299a-475f-11e5-938f-450ab812c9b1.jpg)

parrish:
Just wanted to double check if this is still an issue?

vrooje:
Not for a while... let's try closing it again and seeing what happens...

DarrenMcRoy:
Oof, unfortunately, I have to reopen this.

Just hit this.

[...]

PR #1753, opened 2015-10-13
Use apiClient headers as talkClient headers

aweiksnar:
- Fix for #1052, which *seems* to be due bearer-tokens syncing between clients
- I tested this out by refreshing the token every second and am consistently able to submit comments
- This is how the client was originally setup. It seems the move away from this style may have been intended to fix an unrelated problem


parrish:
I can't seem to break it.  Looks good to me.

PR #1754, opened 2015-10-13
Route tag searches to tags results page

### LINKED ISSUE #1748: Trigger tag search in Talk if you have # as the first character
A suggestion for how to trigger the tag search in the current Talk setup. The volunteers who pointed out the Tag search wasn't there, were automatically typing in the search bar #hashtag . Perhaps that would be an easy way to trigger the tag search manually and keep the same search interface for now?

aweiksnar:
- Searches are triggered by having a '#' as the first character
- Closes #1748


parrish:
:+1:

PR #1761, opened 2015-10-14
Add a component to prevent multiple form submissions

### LINKED ISSUE #1759: Disable submission buttons while submitting
To prevent duplicate requests.  As reported in #1681

parrish:
Closes #1759 

This is probably generally useful too.

Only caveat is that when the button is a `<button type="submit">` it still needs an `onClick` handler to call the form submission.  (Apparently react kills the form submit event if there's a click event handler)

### DIFF
--- a/app/components/single-submit-button.cjsx
+++ b/app/components/single-submit-button.cjsx
@@ -1,0 +1,31 @@
+React = require 'react'
+
+module.exports = React.createClass
+  displayName: 'SingleSubmitButton'
+
+  getDefaultProps: ->
+    className: ''
+    disabled: false
+    type: 'button'
+    onClick: Function.prototype
+
+  getInitialState: ->
+    disabled: @props.disabled
+
+  onClick: (e) ->
+    @setState disabled: true
+    promise = @props.onClick e
+    if promise
+      promise.then =>
+        @setState disabled: false
+    else
+      @setState disabled: false
+
+  render: ->
+    <button
+      type={@props.type}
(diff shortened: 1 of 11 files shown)


app/talk/comment-image-selector.cjsx
```diff
@@ -19,6 +20,7 @@ module?.exports = React.createClass
       .then (subjects) =>
         subject.type = 'recent' for subject in subjects
         @setState {subjects}
+      .catch =>
```
aweiksnar: Were these empty catch functions added for anything specific?

app/components/single-submit-button.cjsx
```diff
@@ -0,0 +1,31 @@
...
+  displayName: 'SingleSubmitButton'
+
+  getDefaultProps: ->
+    className: ''
+    disabled: false
+    type: 'button'
+    onClick: Function.prototype
+
+  getInitialState: ->
+    disabled: @props.disabled or false
```
aweiksnar: I think `or false` can be removed since the default prop is false

app/talk/comment-image-selector.cjsx
parrish: Since the button is waiting for the promise to be resolved, if a subject isn't found it never resolves.

app/components/single-submit-button.cjsx
parrish: Good catch!

aweiksnar:
:shipit:

Issue #1681, opened 2015-10-01
"This comment has been deleted" comments


mkosmala:
Do regular users see comments that have been deleted as comments that say "This comment has been deleted" or just moderators?

aweiksnar:
That will appear for all users

mkosmala:
Can we hide the comments that have been deleted? They're rather unsightly. Plus sometimes we get lots of deleted comments in a row, which looks really weird.

aweiksnar:
The reasoning for this choice was that so comments after deleted comments wouldn't appear completely out of the blue if they were in response to the deleted comment

mkosmala:
Hmm... that makes sense, too. But we have many deleted comments that were just because someone accidentally submitted a comment twice (or more) in a row. And none where someone replied to a comment and then that comment was deleted...

aweiksnar:
If you are a moderator on that talk you can report them and then delete from the moderator-controls 'view reported comments' page and it will do a hard delete

mkosmala:
Oh. Hmm... that's a lot of work. Let me see if I can think of a nice way to take into account both needing to keep some deleted comments and wanting to be rid of others...

mkosmala:
@chrissnyder Actually, once a comment has been deleted, I can no longer report it... So I don't have a way to hard-delete "deleted" comments. Re-open this?

chrissnyder:
Not sure I agree there is a problem here. Leaving a "this comment has been
deleted" placeholder is pretty common behavior across other discussion
platforms, and for good reason.

It is a bit weird that mods can't delete those comments, but I also don't
think it's a good idea to encourage mods to hard-delete every comment that
someone has deleted themselves.

mkosmala:
On new projects, there can be more deleted comments than regular ones, as people learn the interface. It makes the boards look really ugly and discourages conversation. See, for example:
https://www.zooniverse.org/projects/kosmala/season-spotter-questions/talk/62/1787?comment=3886&page=1

Tagging @DarrenMcRoy and @aliburchard who have been on Talk boards of multiple projects and have a broader view of the subject.

DarrenMcRoy:
It's important to have a record of a comment that has been deleted, for volunteers and for us. However, currently, the deleted comments take up as much space as a regular one, and also look exactly like a regular one. (Heck, I could submit a comment that says "this comment has been deleted" and it would look identical to a deleted comment!)

[...]

PR #1762, opened 2015-10-14
Reduce the height and opacity of deleted comments

parrish:
As discussed in #1681

deleted_comment

I reduced the avatar image size to match the height of the comment box without the toolbar buttons.


aweiksnar:
:shipit:

PR #1766, opened 2015-10-14
Hide the new discussion link on subject default (notes) boards

### LINKED ISSUE #872: Remove "create discussion" from Notes board
It should not be possible to create a new discussion in the Notes board, because said board should be reserved entirely for subject threads. Having anything else able to go there would be confusing and counterproductive.

parrish:
Closes #872

### DIFF
--- a/app/talk/board.cjsx
+++ b/app/talk/board.cjsx
@@ -185,7 +185,9 @@ module?.exports = React.createClass
         </div>
         }
 
-      {if @props.user?
+      {if @state.board.subject_default
+        <span></span>
+      else if @props.user?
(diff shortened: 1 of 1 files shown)


DarrenMcRoy:
Oh, thanks, dude! This is a really helpful tweak for keeping threads organized.

chrissnyder:
That'll cause the "sign in" message to appear on the notes board if the user is actually signed in.

parrish:
Oops, good catch.  Updated

chrissnyder:
LGTM :shipit:

Issue #1017, opened 2015-07-09
having active users list show up on instances there's a talk side bar


mschwamb:
maybe this isn't the permanent home for this feature, but it would be nice if the active user list showed up on this side bar as well 

![screen shot 2015-07-09 at 9 06 29 pm](https://cloud.githubusercontent.com/assets/753325/8595887/806cb8b0-267e-11e5-9210-1f8836dfd6b9.png)

mschwamb:
Also if the hastags could show up as well on the side bar everywhere - I think that would be extremely helpful

vrooje:
+1
and +1 :)

mschwamb:
@aweiksnar would it be easy to add the active users and popular hashtags to the Talk side bar on the Talk side bar that shows when you click on the board like [here](https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/10/)?

 I'm finding I have a hard time remember what the evolved hashtag for a feature is and having it somewhere not just on the Talk front page would help. I also love the list of active users and so if it could show up there too, it would give the feeling there are other people around. 

I ask because the toolbar looks to be exactly the same as the one on the main Talk page in design that this seemed maybe easy to replicate.

parrish:
Duplicate of #892

mschwamb:
@parrish Well slightly different I think. Currently the side bar that exists on the boards doesn't show active users or hashtags. The other ticket is asking for the side bar to be everywhere which I imagine is more work moving things around. Since I thought this was an easy change to make on the side bars of the discussion boards I made it a different ticket.

mschwamb:
@parrish Likes look you made it all standard which is awesome. thanks

PR #1767, opened 2015-10-14
Talk sidebar

### LINKED ISSUE #892: Having the talk sidebar show up everywhere in Talk
low in the queue, but this might be a simple adjustment so I'm suggesting it now - For helping promote hashtags, I was wondering if the side bar you see on the front Talk page could be everywhere  in Talk so that when you're in a discussion you can see the most popular hashtags like you can on Talk 2.0. Also would help with fast transitions between project Talks

screen shot 2015-06-30 at 11 30 01 am

compared to

screen shot 2015-06-30 at 11 36 37 am

parrish:
Organizes talk sidebars and makes them consistent throughout talk.

Closes #892 
Related to #1017

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -52,11 +52,11 @@ ProjectPage = React.createClass
     project: null
 
   componentDidMount: ->
-    sugarClient?.subscribeTo "project-#{ @props.project.id }"
+    sugarClient.subscribeTo "project-#{ @props.project.id }"
     document.documentElement.classList.add 'on-project-page'
 
   componentWillUnmount: ->
-    sugarClient?.unsubscribeFrom "project-#{ @props.project.id }"
+    sugarClient.unsubscribeFrom "project-#{ @props.project.id }"
     document.documentElement.classList.remove 'on-project-page'
 
   getPageTitles: (page) ->
--- a/app/talk/board.cjsx
+++ b/app/talk/board.cjsx
@@ -17,6 +17,9 @@ merge = require 'lodash.merge'
 talkConfig = require './config'
 SignInPrompt = require '../partials/sign-in-prompt'
 alert = require '../lib/alert'
+PopularTags = require './popular-tags'
+ActiveUsers = require './active-users'
+ProjectLinker = require './lib/project-linker'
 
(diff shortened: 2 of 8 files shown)


parrish:
Believe it or not, the `min-width: 1px` is actually required.

In Firefox (and IE I think), `max-width` is broken inside `display: flex` containers.

Don't ask me ***why*** that fixes it though.  :rage4: 

flex

parrish:
Also, this does feel like a bit of duplication, but I'm not sure how to approach refactoring this.

The flex layout from `talk-list-content` needs to come from the layout using the sidebar.  Also, different views have different items (or project contexts) available -- e.g. recent comments.

srallen:
@parrish, this might be the same issue I encountered with flex that fixed in #1750. According to the bugzilla reports I linked to, Mozilla devs are saying this is how the flex spec should be working and that a min-width or max-width should be specified:

> Basically: flex items will refuse to shrink below their minimum intrinsic width, unless you explicitly specify "min-width" or "width" or "max-width" on them.

> You get the expected result if you add "min-width: 0;"

parrish:
Wacky.  Most of the issue came from the image tag.

> flex items will refuse to shrink below their minimum intrinsic width

I don't think this is *entirely* an issue with the intrinsic width of the flex child.  I think it had more to do with how that width propagated through the child elements (and vice-versa through the percentages).

On the image:

  - `max-width: 100%` worked, but used the wrong width (larger than the parent), so it broke the flex layout
  - `width: 100%` worked great, except for images that were smaller than the parent.

On the flex children:
  - `width: 100%` and `max-width: 100%` didn't affect the image.
  - `min-width: whatever` caused the `max-width` on the image to work correctly.

In any case.  Yay CSS.  :broken_heart:

aweiksnar:
Yeah agree that the sidebar doesn't work as part of the talk base view component. I'm hoping that in the next talk design iteration the sidebar format will be changed up, but this looks great until then.

Do the sugarClient subscriptions need be moved to the top level route-handlers now? `/talk/index` & `/pages/talk` so they don't unsubscribe when the board view unmounts

parrish:
Good catch.  I'll look at a more general fix for sugar subscriptions

parrish:
Should be fixed now.  Thanks!

aweiksnar:
:shipit:

PR #1773, opened 2015-10-15
Fix discussion move

### LINKED ISSUE #1765: Thread moving appears to have broken?
This thread showed up in the wrong board, and I wanted to move it to Project Building, but I couldn't select anything from the drop-down except the board it's already in:

https://www.zooniverse.org/talk/14/11842/?scrollToLastComment=true&page=1

Can anyone else move it, or has that capability broken?

parrish:
Closes #1765 

I think this might be a regression from a5c465aad1

This also fixes the transition to the new board.


aweiksnar:
:shipit: :shipit:

PR #1779, opened 2015-10-16
fixes broken link to Simpson et al paper

mrniaboc:


chrissnyder:
New link appears broken too?

camallen:
classic miller

mrniaboc:
There ya go

PR #1784, opened 2015-10-16
Allow owners to make gold standard classifications

### LINKED ISSUE #1770: Owners can't set themselves as experts (and thus can't enter gold standard data)
I feel like we fixed a similar issue previously. When I go to the collaborators page for a project that I own, under the "project owner" heading, all I get is text saying "you are the project owner." I can assign other people roles of collaborator, expert, researcher, moderator, tester, etc, but not myself. I think this is a high priority, as it affects gold standard data entry as well as how project owners appear on talk.

brian-c:
In addition to experts, closes #1770.


aweiksnar:
:shipit: :shipit:

PR #1787, opened 2015-10-16
Queue failed classification saves

brian-c:
If saving a classification fails, the classification's data is stored in localStorage and saved when the next classification save is successful.

Test with a query string like `?simulate-classification-save-failure=1`.


aweiksnar:
:shipit: :+1: 
Thanks for putting in simulateSaveFailure, made this nice & easy to test out on dev

PR #1794, opened 2015-10-19
Add links to full size subjects in talk

### LINKED ISSUE #979: Adding a magnifier feature from Talk 2.0 into panoptes Talk
future feature request - I've seen a few request on P4T Talk to have the magnifier that exists on P4 Talk on the new panoptes talk. I don't think the magnifier is implemented in all of Talk 2.0 but for some projects like P4. 

Basically you moused over the subject and a zoom in window showed up that could move around the image to see different parts at higher resolution 

Since you can't see my mouse in a screenshot, I faked a mouse pointer in the image

screen shot 2015-07-06 at 9 26 06 am

aweiksnar:
- Adds links to full res subject images in discussions and subject page
- Closes #979

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -41,6 +41,7 @@ module.exports = React.createClass
     onLoad: NOOP
     defaultStyle: true
     project: null
+    linkToFullImage: false
 
   getInitialState: ->
     loading: true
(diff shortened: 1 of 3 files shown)


app/components/subject-viewer.cjsx
```diff
@@ -103,6 +104,10 @@ module.exports = React.createClass
               <FavoritesButton project={@props.project} subject={@props.subject} user={@props.user} />
               <CollectionsManagerIcon project={@props.project} subject={@props.subject} user={@props.user} />
             </span>}
+          {if type is 'image' and @props.linkToFullImage
+            <a href={src} title="Subject Image" >
```
chrissnyder: Extra space before closing >

chrissnyder:
Thoughts on having it open in a new window? It'd be the only action button below the subject that moves the user somewhere else.

chrissnyder:
Other than that, LGTM

app/components/subject-viewer.cjsx
aweiksnar: Good catch just updated

aweiksnar:
Yeah agree that new window is better :+1: Just updated

chrissnyder:
:shipit:

PR #1796, opened 2015-10-19
Added back to inbox link

### LINKED ISSUE #1034: Stuck in private message thread
When you click on the new message envelope icon on the top bar, it takes you to your inbox. If you click on the message thread with the new message, you got the private message thread between you and the other user but then there's no way to go back into the mailbox other than by clicking the envelope icon on the top right

I think there should be some link on top of the message thread that can take you back to your mailbox

srallen:
Fixed #1034 by adding "Back to Inbox" link.


DarrenMcRoy:
Oof, thanks a million, Sarah. This was a real annoyance. Great work.

aweiksnar:
:shipit:

PR #1805, opened 2015-10-19
Only run search request when query has changed

aweiksnar:
- Fixes an error where the search results would sometimes refresh itself


chrissnyder:
LGTM

aweiksnar:
:cookie:

PR #1808, opened 2015-10-20
Workflow visualization

CKrawczyk:
This adds a workflow visualization tool to the front-end (accessed via the workflow editor).  Right now this is just a viewer, but I plan to turn it into a full editor soon.  It also might be a good idea to test this on staging before pushing it to production, just to make sure it does not break anything.


chrissnyder:
We aren't the best at picking up PRs without a prompt to do so, so if you'd like someone to take a look at your PR, hit them up with @ mention.

chrissnyder:
Couple things, so you have some feedback to work off of.

First off, that's pretty cool and probably very useful to teams. Screenshot below!



Some code critiques:

- Why is jsPlumb both a vendor and node module dependency? Can we just pick one?
- I personally really don't like the inclusion of jQuery or jQuery UI. The only thing I see you using it for is making nodes resizable. Feels like you could just write that yourself and ditch the dependencies.
- I've yet to see a scenario where setting component properties directly like you do around https://github.com/zooniverse/Panoptes-Front-End/pull/1808/files#diff-ea4d83b06fd016cdfa834ccebed80d10R24 (and much more so within that component) is a good idea. You are using them as state and they are referenced within the render function, thus they are component state and you should use `setState`. If the number of things you are tracking gets to be too unwieldy, consider breaking the component up.

CKrawczyk:
Thanks for the feedback.  Just to respond to your points:

+ The jsPlumb node module was not written very well and does not support being "Required" (it looks like it was converted into a node module late on in its development and was not ported correctly, see this [issue](https://github.com/sporritt/jsPlumb/issues/437) for more info).  To work around this I actually copy the distribution source from the node package to the public folder so it can be included in the HTML (See [this line](https://github.com/CKrawczyk/Panoptes-Front-End/blob/workflow-visualization/package.json#L60)).  I could using something like [borwserfy-shim](https://github.com/thlorenz/browserify-shim) to include it in main.js file if you feel that is a better solution.
+ Yeah, that is the only thing I was using jQuery and jQuery-UI for.  A while back I spent a bit of time trying to write a resize code that played nicely with jsPlumb and did not have much luck so I used this as a quick work around.  I will give it another shot.
+ I see what you are saying, I will change this around.

eatyourgreens:
On the plus side, jQuery UI usually gives you accessibility for free, particularly keyboard support and ARIA states and roles.

CKrawczyk:
Just addressed points 2 and 3 above.  I ended up using [react-component-resizable](https://github.com/nrako/react-component-resizable) to handle the node resize without jQuery or jQuery-UI.

CKrawczyk:
As an additional note the new resize method is done via CSS so it will not work in IE or Edge.

app/pages/lab/workflow-viewer/answers.cjsx
```diff
@@ -0,0 +1,58 @@
+React = require 'react'
+Sty = require './jsplumb-style'
```
chrissnyder: Generally speaking, it's not a great idea to abbreviate variable names outside of the most obvious cases, like array indexes. Full names act as great inline documentation for the next dev to come along. Could you go through the PR and convert shorthand names into their full name, e.g. `Sty` to `JSPlumbStyle`, `wf` to `workflow`, `pos` to `position`, etc.

app/pages/lab/workflow-viewer/answers.cjsx
```diff
@@ -0,0 +1,58 @@
...
+
+  render: ->
+    if @props.inputs.type == 'drawing'
+      icon = <span className='drawing-tool-icon-vis' style={color: @props.inputs.item.color}>{icons[@props.inputs.item.type]}</span>
+    else
+      icon = undefined
+    <li className='answer-item' id={@props.inputs.listId}>
+      <div className='lab'>
+        {icon}
+        <Markdown className='markdown-vis'>{@isEmptyStr(@props.inputs.item.label)}</Markdown>
```
chrissnyder: Is `isEmptyStr` necessary? `Markdown` should display either way just fine.

app/pages/lab/workflow-viewer/answers.cjsx
chrissnyder: `Sty` looks to be unused here, can be removed.

app/pages/lab/workflow-viewer/answers.cjsx
```diff
@@ -0,0 +1,58 @@
...
+      listId: @props.inputs.uuid.get(idx)
+      type: @props.task.type
+
+    <AnswerItem jp={@props.jp} key={"Al_#{idx}"} inputs={inputs} eps={@props.inputs.eps} />
+
+  render: ->
+    if @props.task.type == 'drawing'
+      answers = (@createAnswer(idx, item) for idx, item of @props.task.tools)
+    else
+      answers = (@createAnswer(idx, item) for idx, item of @props.task.answers)
```
chrissnyder: I think this section would be a bit easier to understand if you moved the conditional into `createAnswer`, especially cause it already reaches into props rather than spreading the logic across both render and createAnswer. Maybe adding `@props.task` as an argument to the function?

app/pages/lab/workflow-viewer/index.cjsx
```diff
@@ -0,0 +1,51 @@
...
+    [owner, name] = @props.project.slug.split('/')
+    viewParams = {owner, name}
+    viewQuery = workflow: @props.workflow.id, reload: 0
+    @makeHref 'project-classify', viewParams, viewQuery
+
+  render: ->
+    <div>
+      <div style={pointerEvents: 'all'}>
+        <span className='standard-button' onClick={@onSort}>Sort Task Nodes</span>
+      </div>
```
chrissnyder: Any reason this couldn't be a `button`?

app/pages/lab/workflow-viewer/index.cjsx
```diff
@@ -0,0 +1,51 @@
...
+        <a href={@workflowLink()} className="standard-button" target="from-lab" onClick={@handleViewClick}>Test this workflow</a>
+      </div>
+    </div>
+
+module.exports = React.createClass
+  displayName: 'WorkflowVisPageWrapper'
+
+  getDefaultProps: ->
+    params:
+      workflowID: ''
```
chrissnyder: I don't think this means much here, considering the page requires a workflow ID to do anything. Maybe using `React.PropTypes.shape` with a required param of `workflowID`?

app/pages/lab/workflow-viewer/jsplumb-style.cjsx
```diff
@@ -0,0 +1,104 @@
...
+
+commonA =
+  connector: ["Flowchart",
+    stub: 30
+    cornerRadius: 5
+    alwaysRespectStubs: false
+    midpoint: 0.5]
+  #connector: ["Straight"]
+  #connectior: ["Bezier", { curviness: 150 }]
+  #connectior: ["State Machine"]
```
chrissnyder: Leftover dev code?

app/pages/lab/workflow-viewer/jsplumb-style.cjsx
```diff
@@ -0,0 +1,104 @@
...
+
+commonA_open =
+  connector: ["Flowchart",
+    stub: 30
+    cornerRadius: 5
+    alwaysRespectStubs: false
+    midpoint: 0.5]
+  #connector: ["Straight"]
+  #connectior: ["Bezier", { curviness: 150 }]
+  #connectior: ["State Machine"]
```
chrissnyder: Leftover dev code?

app/pages/lab/workflow-viewer/task-helpers.cjsx
```diff
@@ -0,0 +1,58 @@
...
+# Render the task name box
+TaskName = React.createClass
+  displayName: 'TaskName'
+
+  isEmptyStr: (str) ->
+    str.replace(/^\s+|\s+$/g, '').length == 0
+
+  render: ->
+    task = @props.wf[@props.taskKey]
+    task_text = tasks[task.type].getTaskText(task)
```
chrissnyder: Another note on code style. Javascript variable/function naming convention is camelCase. To keep things consistent with the rest of the codebase, we should try to stick with that.

(Take my code style comments with a grain of salt, as folks have different styles and to each their own. In this case, keeping things consistent with the rest of the app helps other devs jump in when necessary.)

app/pages/lab/workflow-viewer/task-helpers.cjsx
```diff
@@ -0,0 +1,58 @@
...
+    style =
+      'zindex': 'inherit'
+    <div className='answer-item' id={@props.plumbId + '_name'} style={style}>
+      <Markdown className='lab markdown-vis'>{"#{task_text}"}</Markdown>
```
chrissnyder: Should be able to just do `<Markdown className='lab markdown-vis'>{task_text}</Markdown>` yea?

app/pages/lab/workflow-viewer/task.cjsx
```diff
@@ -0,0 +1,313 @@
...
+    task: @props.wf[@props.taskKey]
+    endpoints: []
+    zIndex: @props.taskNumber + 2
+    uuid: 0
+    uuids: []
+    me: null
+    width:
+      min: 0
+      max: 0
+      val: 0
```
chrissnyder: Couple of things here. First, there are few are far between instances where you should immediately move a prop into a state variable. Doing that means you are now tracking app state in two spots (this component and the parent component), which can lead to state synchronization bugs. You should be able to replace every instance of `@state.zIndex` with `@props.taskNumber + 2` or if you want to be more explicit about it, using a function like so:

```
computeZIndex = (taskNumber) ->
  taskNumber + 2
```

Same comments as above on variable naming.

Be careful about duplicate keys within the object you are returning. In this case, only one of the `zIndex` keys matters (doesn't really matter here, but might somewhere else).

app/pages/lab/workflow-viewer/task.cjsx
```diff
@@ -0,0 +1,313 @@
...
+          eps.push(ep)
+        @setUuid(null, uuid, uuids)
+      if @state.task.type != 'single'
+        ep = @props.jp.addEndpoint(@props.plumbId+'_name', Sty.commonA, {uuid: @props.plumbId + '_next'})
+        ep.canvas.style['z-index'] = @state.zIndex
+        eps.push(ep)
+    computedStyle = window.getComputedStyle(me)
+    new_state =
+      endpoints: eps
+      me: me
```
chrissnyder: No reason to do this. If you need the component DOM node, access it like you do above.

app/pages/lab/workflow-viewer/task.cjsx
```diff
@@ -0,0 +1,313 @@
...
+      if parseInt(style.width) > @state.width.max
+        style.width = @state.width.max + 'px'
+      if parseInt(style.width) < @state.width.min
+        style.width = @state.width.min + 'px'
+    else
+      style = @props.pos
+      style['zIndex'] = @state.zIndex
+    if @state.zIndex > max_z
+      max_z = @state.zIndex
+    style
```
chrissnyder: If you need to track left, top, width, etc as state, set them as state rather than saving them indirectly on the style object of the component.

Pseudo-code below, not tested but might look like this:

```
getStyle: ->
  width = if @state.width > @state.maxWidth
    @state.maxWidth
  else if @state.width < @state.minWidth
    @state.minWidth
  else
    @state.width

  if @state.zIndex > currentMaxZIndex
    currentMaxZIndex = @state.zIndex

  left: @state.left
  top: @state.top
  width: width
  zIndex: @state.zIndex
  ```

File: app/pages/lab/workflow-viewer/task.cjsx [hunk 13]
```diff
@@ -0,0 +1,313 @@
...
+        endpoints: @state.endpoints
+      uuid:
+        set: @setUuid
+        get: @getUuid
+
+  # set positon
+  moveMe: (pos, update=true) ->
+    me = @state.me
+    me.style.left = pos.left + 'px'
+    me.style.top = pos.top + 'px'
```
chrissnyder: Indirectly setting state here.

app/pages/lab/workflow-viewer/workflow.cjsx
```diff
@@ -0,0 +1,226 @@
...
+            wf[tmp].subtask = true
+            L += 1
+            # set 'next' to be the next subtask in the list (if it exists)
+            if tool.details[sdx + 1]?
+              wf[tmp].next = 'S' + L
+          # replace details with a list of keys
+          wf[k].tools[tdx].details = subList
+    # make sure the keys are in number order
+    keys.sort()
+    {init, keys, wf, taskStateSet}
```
chrissnyder: What's the reason for reformatting the workflow itself? Seems pretty brittle and would break as and when we add/change tasks.

app/router.cjsx
```diff
@@ -95,6 +95,8 @@ module.exports =
     <Route name="edit-project-collaborators" path="collaborators/?" handler={require './pages/lab/collaborators'} />
     <Route name="edit-project-media" path="media/?" handler={require './pages/lab/media'} />
     <Route name="edit-project-workflow" path="workflow/:workflowID/?" handler={require './pages/lab/workflow'} />
+    <Route name="edit-project-workflow-visualization" path="workflow/:workflowID/visualize/?" handler={require './pages/lab/workflow-viewer'} />
+    <Route name="edit-project-workflow-visualisation" path="workflow/:workflowID/visualise/?" handler={require './pages/lab/workflow-viewer'} />
```
chrissnyder: Took me a minute to figure out why you did this. :us: :gb:

views/index.ejs
```diff
@@ -93,6 +93,8 @@
 
     <script src="https://cdn.polyfill.io/v1/polyfill.min.js?features=default,es6,Promise,fetch&amp;flags=gated"></script>
     <script>('assign' in Object) || document.write('<script src="/fallback-polyfills.js"></'+'script>');</script>
+    <!-- Include jsPlumb jquery and jquery-ui -->
```
chrissnyder: Extra comment

package.json
```diff
@@ -14,6 +14,7 @@
     "ejs": "^2.3.3",
     "express": "^4.13.3",
     "json-api-client": "~0.4.0",
+    "jsplumb": "1.7.9",
```
chrissnyder: Totally cool with just pulling down js-plumb after install as a vendor dependency. No need for this.

chrissnyder: Oops, nevermind, I see you move the dist out of node_modules.

css/workflow-visualizer.styl
```diff
@@ -0,0 +1,130 @@
...
+  margin-left: 10px
+  text-weight: 500
+
+.drag-handel
+  padding-bottom: 10px
+  cursor:  move
+  border-top-left-radius: 15px
+  border-top-right-radius: 15px
+  background-color: #000
+  background-color: rgba(0,0,0,.3)
```
chrissnyder: rgba syntax is only not supported in IE < 9, which we don't support anyway. Can remove duplicate entries like this.

css/jsplumb.styl
```diff
@@ -0,0 +1,20 @@
...
+  z-index: 0
+
+/** DISABLE TEXT SELECTION (SET ON BODY WHEN DRAGGING IS OCCURRING) **/
+._jsPlumb_drag_select *
+  -webkit-touch-callout: none
+  -webkit-user-select: none
+  -khtml-user-select: none
+  -moz-user-select: none
+  -ms-user-select: none
+  user-select: none
```
chrissnyder: Stylus auto vendor prefixes stuff, so you can just do `user-select: none` here.

app/pages/lab/workflow-viewer/workflow.cjsx
CKrawczyk: The reason I am doing this here is because I need to extract the sub-tasks as full nodes.  For the visualization bit this is not that important, but when I extend this into a full editor it will become important (i.e., when the user disconnects a sub-task the node should stick around as a normal task in the workflow, or a normal task can be connected up as a sub-task).

css/jsplumb.styl
CKrawczyk: Nice, I did not know that :smiley:

app/pages/lab/workflow-viewer/answers.cjsx
CKrawczyk: A div with an empty line renders as a different height (nothing vs. single space).  Now that I think about it I can do the same thing with a min-height in the CSS.

app/pages/lab/workflow-viewer/task-helpers.cjsx
CKrawczyk: Yep, left over from when I was pre-pending the name with the task number (I ended up moving that to the drag handle instead).

app/pages/lab/workflow-viewer/task.cjsx
CKrawczyk: Nice catch, did not mean to put `zIndex` in two times.  Also thanks for the code, that is exactly what I need for this, `zIndex` needs to be initialized by props, but can be changed when a node is dragged (i.e., the node being dragged is always on top).

app/pages/lab/workflow-viewer/answers.cjsx
CKrawczyk: Yep, forgot I moved the bit that used `Sty` in this file.

CKrawczyk:
@chrissnyder thanks for all the comments.  The latest commit has addressed them.

brian-c:
Looks solid. Mind rebasing from master? Should be fine to merge after that.

CKrawczyk:
There we go, rebase done.

brian-c:
Sorry @CKrawczyk , looks like it needs one more rebase!

CKrawczyk:
@brian-c rebase is done.

brian-c:
Thanks!

PR #1809, opened 2015-10-20
transitionTo lab index page if signed out of project builder.

### LINKED ISSUE #1125: When you log out the project builder taking you to the nicer page asking you to sign in or return to project
low in the queue - when you log out while in the project builder you get taken to this
screen shot 2015-07-18 at 3 25 41 pm

I think it should take you back to the same page you go to if you click the project builder linker not logged in. I think it looks more polished
 
screen shot 2015-07-18 at 3 25 53 pm

srallen:
Fixed #1125 by adding a transitionTo to the lab index page if on a lab editor page with projectID when signing out.

### DIFF
--- a/app/pages/lab/project.cjsx
+++ b/app/pages/lab/project.cjsx
@@ -7,6 +7,7 @@ HandlePropChanges = require '../../lib/handle-prop-changes'
 apiClient = require '../../api/client'
 counterpart = require 'counterpart'
 ChangeListener = require '../../components/change-listener'
+Router = require '@edpaget/react-router'
 
 DEFAULT_WORKFLOW_NAME = 'Untitled workflow'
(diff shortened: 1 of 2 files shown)


chrissnyder:
I don't think that's the right place for this to go. It makes account-bar.cjsx aware of specific logic we want to have happen on the labs page, which feels like a conflation of concerns.

Could we try doing something similar, but on https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/project.cjsx instead? The page should react to the presence or not of a user, rather than any specific URL scheme.

srallen:
Yeah, that's a better approach. I've moved the `transitionTo` to `EditProjectPageWrapper`

chrissnyder:
LGTM

PR #1810, opened 2015-10-20
Prevent css font color conflict in board preview

### LINKED ISSUE #1786: Removing or lightening the grey box around the username role on the post preview on Talk
Love the user role ole on the Talk post preview. On firefox I'm seeing that there's a darker gray box with the role text that makes it hard to see. Maybe just leave it clear with black text? Whatever you think is best. Thanks for adding this. 

screen shot 2015-10-16 at 4 58 00 pm

aweiksnar:
- Decreases specificity so the role p's don't conflict here
- Closes #1786

### DIFF
--- a/css/project-roles.styl
+++ b/css/project-roles.styl
@@ -12,7 +12,7 @@
     color: MAIN_HIGHLIGHT
 
   &.translator
-    color: gold
+    color: goldenrod
 
   &.zoo-team
     color: green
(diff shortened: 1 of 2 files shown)


parrish:
I'd suggest we just remove the background color on the label because it's still *really* hard to read.  Here's the difference between COPY_GREY_MID and none:

role | label
-----|-----
team | green
admin | navy
translator | gold
owner | owner

I'd also say we could switch translator from gold to goldenrod for a bit better contrast

goldenrod

aweiksnar:
@parrish Yeah I agree, just updated, let me know what you think. I left the grey background untouched outside of talk since I think it's only used overlaid on an image background

parrish:
:+1:

PR #1811, opened 2015-10-20
Add hamburger nav menu, improve homepage collapse

aweiksnar:
- Adds a :hamburger: style mobile navigation for smaller screens

- I'm not crazy about having a fixed breakpoint vs fluid nav, but I
  think this is a nice compromise with the way the css/global state is
  right now, and will be a usability improvement


chrissnyder:
:hamburger:

parrish:
Looks good, but we ought to change the hover color on the links (white on white)

https://github.com/zooniverse/Panoptes-Front-End/pull/1811/files#diff-55a3d0fdd1ec3f1fd723e5e76364d149R139

aweiksnar:
Oof, Yeah good catch, Just updated

parrish:
:+1:

PR #1813, opened 2015-10-21
Uniquely filter roles based on display name

### LINKED ISSUE #1806: on some project Talks, researchers have two stripes
I'm not sure if this is a bug or a difference between projects with Talk created before launch of the new Zoooniverse api and new Talk and those created after. 

Noticed this when I was checking something on Talk compared to Fossil Finder Talk 

screen shot 2015-10-20 at 4 04 30 am

On P4: Terrains talk the research team has one  stripe.

chrissnyder:
Closes #1806 

I probably missed something here, but putting it out there anyway.

This PR does two things:

1. Remove section filtering from the roles display component. Bit worried about this part, but it doesn't look like anything was actually being filtered out?
2. Uniquely filters the given roles by whatever display name would be output.

### DIFF
--- a/app/talk/lib/display-roles.cjsx
+++ b/app/talk/lib/display-roles.cjsx
@@ -1,4 +1,5 @@
 React = require 'react'
+uniq = require 'lodash.uniq'
 
 zooniverseTeamRole = (role) ->
   role.section is 'zooniverse' and [ 'admin', 'team' ].indexOf(role.name) isnt -1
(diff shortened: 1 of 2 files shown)


aweiksnar:
:+1: Yeah this is a safer way to set this up in the case of duplicate roles. Removing section filtering will be ok since it currently always happens in the requested data passed in. 
:shipit: :flags:

PR #1816, opened 2015-10-21
Pass params to moderation notifications

parrish:
This fixes a bug that is causing moderation report notifications on projects to break.


aweiksnar:
:+1:

PR #1817, opened 2015-10-21
Paginate boards list, increase request size

### LINKED ISSUE #1815: (Old Weather) Talk is only displaying the first 10 boards
The Old Weather Talk community is telling me that they've created 13 or 14 boards, but only the first 10 are displaying.

If we're not going to have any way to nest boards, then we really can't put limits on the number of top-level boards.

https://www.zooniverse.org/projects/zooniverse/old-weather/talk/

aweiksnar:
- Closes #1815

### DIFF
--- a/app/talk/init.cjsx
+++ b/app/talk/init.cjsx
@@ -17,6 +17,7 @@ ZooniverseTeam = require './lib/zoo-team.cjsx'
 alert = require '../lib/alert'
 AddZooTeamForm = require './add-zoo-team-form'
 DragReorderable = require 'drag-reorderable'
+Paginator = require './lib/paginator'
 
 module?.exports = React.createClass
   displayName: 'TalkInit'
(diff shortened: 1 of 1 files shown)


DarrenMcRoy:
I don't know about paginating the boards list. I would rather they just all displayed on one page. I know that might increase load time, but I've never been to a forum where boards are paginated.

(a la the [old-school Old Weather forum](http://forum.oldweather.org/) where they have 17 boards all displaying on the front page -- and then more nested beyond that)

aweiksnar:
@DarrenMcRoy It'll only show them as pages if there are more than 20 boards. It'll now request 20 at a time

DarrenMcRoy:
I can live with that for now, then. Especially since you can manually order them. Good work.

chrissnyder:
:hamster:

PR #1818, opened 2015-10-21
Convert mobile nav icon to react component

aweiksnar:
- SVG image assets aren't working correctly with the asset paths in production
- This is probably nicer to work with as a component too


chrissnyder:
:hamburger:

aweiksnar:
:bow: Thanks!

PR #1819, opened 2015-10-21
Fix notifications page when not logged in

aweiksnar:
- Notifications#markAsRead only needs to be called if a user is logged
  in


chrissnyder:
:+1:

PR #1820, opened 2015-10-22
Fix file upload regex check in admin mode

rogerhutchings:
At the moment, uploading subjects doesn't work if you have admin mode turned on. What's happening is the regex looking for the file field in the manifest can't find a valid entry, tries to treat every field as an image location, and can't find any subjects.

This matches anything that looks like a filename with a 2-4 character extension and uses that.

Thanks to @camallen for helping me solve this one :)

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -16,7 +16,7 @@ isAdmin = require '../../lib/is-admin'
 NOOP = Function.prototype
 
 VALID_SUBJECT_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif', '.svg']
-INVALID_FILENAME_CHARS = ['/', '\\', ':']
+INVALID_FILENAME_CHARS = ['/', '\\', ':', ',']
 MAX_FILE_SIZE = 600000
 
 announceSetChange = ->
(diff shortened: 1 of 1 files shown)


chrissnyder:
Got a gist or something with an example manifest that causes the problem?

rogerhutchings:
https://gist.github.com/rogerhutchings/d31c836b3ef0a06a0135

app/pages/lab/subject-set.cjsx
```diff
@@ -273,8 +273,9 @@ EditSubjectSetPage = React.createClass
   _findFilesInMetadata: (metadata) ->
     filesInMetadata = []
     for key, value of metadata
-      extensions = if isAdmin() then '' else "(?:#{VALID_SUBJECT_EXTENSIONS.join '|'})"
+      extensions = if isAdmin() then '\\.\\w{2,4}' else "(?:#{VALID_SUBJECT_EXTENSIONS.join '|'})"
       filesInValue = value.match? ///([^#{INVALID_FILENAME_CHARS.join ''}]+#{extensions})///gi
+      console.log ///([^#{INVALID_FILENAME_CHARS.join ''}]+#{extensions})///gi
```
chrissnyder: Could the log be removed?

chrissnyder:
LGTM

app/pages/lab/subject-set.cjsx
rogerhutchings: done

PR #1822, opened 2015-10-22
private collections

### LINKED ISSUE #1812: Favorites say they are public but really they are private
Not sure if this front end or backend since it might just be appearance. 

screen shot 2015-10-21 at 3 50 48 am

As far as I can tell favorite collections are private by default, but if you look the properties of a default public collection it says it is public, though I can't see it in my list of collections or user profile. Is this the intended behavior?

When I go to https://www.zooniverse.org/favorites/mschwamb/ it says 'No Collections Found'

### LINKED ISSUE #1821: collections settings page visible when collection is set to public
Super low priority as the API will reject unauthorized updates but the UI doesn't check for relevant roles before showing collection settings. Best to check the authenticated user has relevant roles to edit and modifying rendered content. 

E.g. https://www.zooniverse.org/collections/camallen/favorites-6/settings/

camallen:
closes #1812 closes #1821

turns out most collection private settings are null (will add a default in the api) and were displaying as public but weren't. modified the display here so users can decide. Also added a role check for collection settings / collab pages before render.

### DIFF
--- a/app/collections/settings.cjsx
+++ b/app/collections/settings.cjsx
@@ -77,6 +77,12 @@ module.exports = React.createClass
         <CollectionDeleteDialog {...@props} onComplete={resolve} />
       </div>
 
+  public_collection: ->
+    if @props.collection.private?
+      not @props.collection.private
+    else
+      @props.collection.private
+
   render: ->
     <div className="collection-settings-tab">
       <ChangeListener target={@props.collection}>{=>
--- a/app/collections/settings.cjsx
+++ b/app/collections/settings.cjsx
@@ -4,6 +4,7 @@ DisplayNameSlugEditor = require '../partials/display-name-slug-editor'
 apiClient = require '../api/client'
 alert = require '../lib/alert'
 SetToggle = require '../lib/set-toggle'
+CollectionRole = require '../lib/collection-role'
 ChangeListener = require '../components/change-listener'
 
 CollectionDeleteDialog = React.createClass
@@ -43,7 +44,7 @@ CollectionDeleteDialog = React.createClass
 
 module.exports = React.createClass
   displayName: "CollectionSettings"
-  mixins: [Navigation, SetToggle]
+  mixins: [Navigation, SetToggle, CollectionRole]
   setterProperty: 'collection'
 
   getDefaultProps: ->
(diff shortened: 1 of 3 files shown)


app/lib/collection-role.coffee
```diff
@@ -0,0 +1,12 @@
+apiClient = require '../api/client'
+
+module.exports =
+  hasSettingsRole: ->
+    if @props.user?
+      apiClient.type('collection_roles').get(collection_id: @props.collection.id, user_id: @props.user.id)
+        .catch ->
+          false
+        .then ([userRoles]) ->
+          userRoles.roles.some (role) -> role.match ///owner|collaborator///i
```
chrissnyder: `userRoles` might be an empty array. Or did you mean to put the catch after the then?

app/collections/settings.cjsx
```diff
@@ -63,47 +64,50 @@ module.exports = React.createClass
...
-        .then ([roles]) ->
-          roles?
-
   confirmDelete: ->
     alert (resolve) =>
       <div className="confirm-delete-dialog content-container">
         <CollectionDeleteDialog {...@props} onComplete={resolve} />
       </div>
 
+  public_collection: ->
```
chrissnyder: Super nitpicky style comment, but could this be camelCase?

app/lib/collection-role.coffee
camallen: won't the .then only run on success? if it's response is a 200 and an empty array then it won't match and will return false, no?

app/collections/settings.cjsx
camallen: yeah for sure, i'm always changing this after i see them...default style for me.

app/lib/collection-role.coffee
camallen: sorry, did you mean something like this before i check for the roles.. `return false if userRoles.length == 0`

chrissnyder: In this instance, .then will run every time. catch only catches an error that happens before it in the chain, in this case if the API call itself threw. The chain continues to the next `then` statement.

There isn't any matching per se that's going on here. `.then ([userRoles]) ->` will run everytime, and attempt to destructure the first argument passed back and assign `userRoles` as the first element of the first argument. If you don't have any roles on the collection, it'll be assigned `undefined`, leading to an error when you try to access the `role` property.

camallen:
updated.

chrissnyder:
Crap, sorry @camallen, one more thing. Since `hasSettingRole` returns a promise, you have to delay rendering the component until after that promise returns. `PromiseRenderer` does the trick here. See https://gist.github.com/chrissnyder/791e611d80473c6f521c for an example that would work here

camallen:
my bad, on it..

chrissnyder:
:shipit:

PR #1847, opened 2015-10-22
Show drawing tool details when the shape is not in progress

### LINKED ISSUE #1804: sub-tasks should appear on mouseup rather than on click
We're getting a lot of feedback from beta testers that the subtasks popping up are making it hard to complete the main drawing tasks. I agree with this. It makes more sense to me to have the sub-task window pop up only after the shape is finished.

brian-c:
Adds an `_inProgress` property to marks. When it's false, the details will show. Closes #1804.

### DIFF
--- a/app/classifier/drawing-tools/circle.cjsx
+++ b/app/classifier/drawing-tools/circle.cjsx
@@ -20,11 +20,17 @@ module.exports = React.createClass
       r: 0
       angle: 0
 
+    initStart: ->
+      _inProgress: true
+
     initMove: ({x, y}, mark) ->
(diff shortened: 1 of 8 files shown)


chrissnyder:
From what I can tell, sub tasks still show up immediately upon mousedown.

brian-c:
@chrissnyder I can't reproduce. What project are you looking at? What browser are you using?

chrissnyder:
Double-checked and realized I hadn't had the proper branch checked out. Sorry for my sloppiness.

LGTM

PR #1853, opened 2015-10-23
changes banner image on About page

mrniaboc:


mrniaboc:
Any idea why the banner image commit didn't seem to work @chrissnyder ?

chrissnyder:
Was cached, because we only cache-bust our JS and CSS. Cleaned out the cache and it looks to work now.

mrniaboc:
Thanks Chris.

PS: I have no idea what you just said.

PR #1855, opened 2015-10-23
Organize subject components

parrish:
The subject index page was getting pretty huge, so before I added anything else to it, I wanted to try to start breaking out some components.

I also tried to clean up and refactor a bit too.

PR #1856, opened 2015-10-23
Subject discussion list

parrish:
Depends on #1855 (merge that one first).

I'm also not attempting to paginate the subject discussion list.  Aside from having to dance around multiple page query params, I doubt we really care about anything but the 20 most recent discussions about a subject.  Happy to change later on if that becomes an issue.

### DIFF
--- a/app/subjects/comment-list.cjsx
+++ b/app/subjects/comment-list.cjsx
@@ -43,10 +43,10 @@ module?.exports = React.createClass
             <Comment key={comment.id} data={comment} locked={true} linked={true} />}
         </div>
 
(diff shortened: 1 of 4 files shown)


DarrenMcRoy:
I'll keep an eye on if the 20-discussion display limit becomes an issue, but that should be plenty for 99.9% of cases.

aweiksnar:
Looks solid, can you rebase to master and then I'll merge

parrish:
Updated

aweiksnar:
:shipit:

PR #1825, opened 2015-10-22
Add notification links to the sidebar

parrish:

PR #1857, opened 2015-10-23
Fix render return value on sidebar notifications

parrish:
From #1825

`Uncaught (in promise) Error: Invariant Violation: SidebarNotifications.render(): A valid ReactComponent must be returned. You may have returned undefined, an array or some other invalid object.(…)`

### DIFF
--- a/app/talk/lib/sidebar-notifications.cjsx
+++ b/app/talk/lib/sidebar-notifications.cjsx
@@ -32,7 +32,7 @@ module?.exports = React.createClass
       'Notifications'
 
   render: ->
-    return null unless @props.user and @state.unreadCount?
+    return <span></span> unless @props.user and @state.unreadCount?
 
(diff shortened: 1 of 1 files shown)


aweiksnar:
Any steps needed to reproduce this? I'm not seeing the error in console. Returning null/false from render should be [okay](https://facebook.github.io/react/docs/component-specs.html#render)

parrish:
Can't reproduce it now.  It definitely seemed specific about `SidebarNotifications.render()` and changing the return value did fix it when it was occurring.

For what it's worth, it was on any discussion page I had tried.  Nothing else was affected.

Close or merge as you like I guess.  :confused:

aweiksnar:
Interesting I'll keep an eye out for it and if either of us notices it's happening consistently let's reopen/merge

PR #1858, opened 2015-10-23
Subject mentions

parrish:
Depends on #1855 and #1856 (merge those first)

Same deal as the discussion list, I'm just grabbing the 20 most recent.


DarrenMcRoy:
This one might be more important to paginate -- I think there will be more subject mentions than discussions. But I'm okay to wait and see.

parrish:
Rebased for a cleaner diff.

aweiksnar:
:shipit: :shipit:

PR #1859, opened 2015-10-23
Bump modal form version

### LINKED ISSUE #1631: Secondary subtasks on a drawing task are off the page / not reachable
I created a drawing task with 4 subtasks. Only the first question is visible, because the balloon has a fixed size with no scrollbar. Also I am not sure if there is supposed to be a button at the bottom of the balloon, because I can't see beyond the first question.

A less critical but related issue, is that the balloon defaults to position below the drawn box. If you draw from top-left to bottom-right, as most Westerners will, this pretty much guarantees you will have to scroll the page down to see what that balloon you saw a brief flash of, was.

screenshot

brian-c:
Adds modal form scrolling. Closes #1631.

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "lodash.pick": "~3.1.0",
     "lodash.uniq": "^3.2.2",
     "markdownz": "^3.0.0",
-    "modal-form": "~1.1.0",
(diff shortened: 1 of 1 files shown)


aweiksnar:
:shipit:

PR #1870, opened 2015-10-26
Tutorial editor

brian-c:
This switches the tutorial from being baked into the project's config to being linked as a separate resource.

There's also a rudimentary editor in place. I'm working with this to figure out a good way to organize components and data with Redux in the future.

The editor can be activated in a project's admin settings.

### DIFF
--- a/app/classifier/tutorial-button.cjsx
+++ b/app/classifier/tutorial-button.cjsx
@@ -1,0 +1,31 @@
+React = require 'react'
+apiClient = require '../api/client'
+Tutorial = require '../lib/tutorial'
+
+module.exports = React.createClass
+  getDefaultProps: ->
+    user: null
+    project: null
+
+  getInitialState: ->
+    tutorial: null
+
+  componentDidMount: ->
+    @fetchTutorialFor @props.project
+
+  componentWillReceiveProps: (nextProps) ->
+    unless nextProps.project is @props.project
+      @fetchTutorialFor nextProps.project
+
+  fetchTutorialFor: (project) ->
+    apiClient.type('tutorials').get project_id: @props.project.id
+      .then ([tutorial]) =>
+        @setState {tutorial}
+
+  render: ->
+    if @state.tutorial? and @state.tutorial.steps.length isnt 0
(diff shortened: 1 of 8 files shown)


app/classifier/tutorial-button.cjsx
```diff
@@ -0,0 +1,31 @@
...
+
+  componentDidMount: ->
+    @fetchTutorialFor @props.project
+
+  componentWillReceiveProps: (nextProps) ->
+    unless nextProps.project is @props.project
+      @fetchTutorialFor nextProps.project
+
+  fetchTutorialFor: (project) ->
+    apiClient.type('tutorials').get project_id: @props.project.id
```
aweiksnar: Should `@props.project.id` be `project.id` here?

app/components/media-card.cjsx
```diff
@@ -9,6 +9,7 @@ module.exports = React.createClass
     src: ''
 
   render: ->
+    console.log 'src', @props.src
```
aweiksnar: Extra log

brian-c:
Fixed both.

aweiksnar:
Thanks looks good :+1:  I added a tutorial to one of my dev projects and works great. Not sure how I feel about the _boundForceUpdate pattern, but happy to wait until the upcoming architecture changes on that

brian-c:
I bound `forceUpdate` instead of using a `ChangeListener`. With the current architecture, how would you go about updating a component when a resource changes?

aweiksnar:
Could resource updates happen in the parent and then have a shouldComponentUpdate in the child? 
Proposed client edits could be passed up and the top level would trigger saving the new resource

brian-c:
The Controller is the top level, (aside from the Fetcher, which isn't concerned with anything other than making sure the resource is available).

In redux-speak, I'm making the Controller a "smart component" and everything below it a "dumb component", but because we're not using a global store, the Controller re-renders when its resource changes. With redux there'll be an app-level Provider, which I believe is essentially just a app-level forceUpdate when the store changes, right?

I'm under the impression that `shouldComponentUpdate` is mainly for performance optimization, not really necessary here.

aweiksnar:
Alright, Yeah I guess it's just a style preference then. The docs recommend avoiding forceUpdate , so wanted to bring it up if there was another way

PR #1873, opened 2015-10-27
Constrain width of images in notifications

### LINKED ISSUE #1864: Constraining images that are too big in the Talk notifications window
screen shot 2015-10-24 at 4 21 53 am

aweiksnar:
- Closes #1864

### DIFF
--- a/css/talk.styl
+++ b/css/talk.styl
@@ -19,6 +19,9 @@ COPY_GREY_LIGHT = #afaeae
   box-sizing: border-box
   color: COPY
 
+  img
+    max-width: 100%
+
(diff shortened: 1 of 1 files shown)


chrissnyder:
Any danger to just setting that at the top level and being done with it?

aweiksnar:
This is on talk-module, which is close to top-level (comments, board previews, discussion previews, sidebar etc). 
I didn't want to make it global since it could potentially mess with 3rd party maps / other weird things if we ever add those :confused: but yeah it'd probably be safe for 99.9% of cases

chrissnyder:
Word. This'll do for now, was just throwing it out there.

Issue #1722, opened 2015-10-09
got a  Error: 401 Unauthorized  while trying to upload subjects - subject loader should stop rather than continuing on to try and upload the next subject


mschwamb:
The subject loader just kept trying to go forward, so now I don't know if it succeeded on some or failed, and need to reload the whole thing. I did the stupid thing of trying to do the whole 6000 subjects in one go since I had the time for my computer to sit.  It died past the 2000 mark. Sorry I don't have a console of what happened. 

Even if the subject uploader failed and was able to recover on a different image, it's very easy for someone to miss there was a problem. I think the subject loader should stop. It would at least enable recovery easier, because you could see the last subject on the list uploaded check it and then upload the rest.

camallen:
seems like your authenticated session ended if you're getting 401 unauthorized on that route. Also i think the subject unloader reports the final state after it's finished and you can review.

Finally should we make this text more obvious and explicit? We say it for a reason...
> We recommend uploading subjects in batches of 500 - 1,000 at a time.

mschwamb:
@camallen it sounded more to me as suggestion  for speed/time it takes rather than a things will time out and very bad things will happen warning.My fault, I'm dealing with it, but a stricter message (bolded) or an actual restriction of number of subjects uploaded in one go I think would help (the latter suggestion I think has already been ruled out - see #1723 )

vrooje:
I think it would be hugely useful for the site to check after an upload which subjects were *actually* uploaded, and possibly provide a list of subjects that failed, i.e. "here's the new manifest you should feed into the project builder when you try again and don't want to end up with duplicate subjects".

aliburchard:
+100 to @vrooje's suggestion. I know it's possible to suss that out by downloading the subject export and comparing it to the original manifest, but it's not always straightforward to do that. I would *love* to be provided with a list of subjects that failed as the "new manifest"

camallen:
Pretty sure this already happens as per [here](https://github.com/zooniverse/Panoptes-Front-End/blob/62c9eff713b756f82e6c33a3b487e6a3b605a470/app/partials/subject-uploader.cjsx#L36) but i haven't actually tested the front end to see the report.

mschwamb:
@camallen only if you pause from my experience. if you stop it which is my first instinct when I see lots of red errors flying by it doesn't report out.

mschwamb:
This has gotten better over the past year. So closing.

PR #1879, opened 2015-10-29
Emphasize the subject set size recommended for upload and clarifying private data doesn't show on Talk as well on subject set upload page

mschwamb:
Adding bold to text and ‘strongly’ to the upload subject set recommendation so it stands out more against the rest of the text of the page, so project research teams pay more attention to it  and don't ignore it like me. 

Mention in the place where you drop the manifest file and subjects that private data won’t show up on talk either.

This doesn't close any issue though related to  #1722


chrissnyder:
:+1:

PR #1886, opened 2015-11-02
only update selected collections from options array

### LINKED ISSUE #1846: Can't find my collections when I try to add to them
I definitely have a collection that I've made:
screen shot 2015-10-22 at 21 35 42

But when I try to add a subject to it I cannot get it to show up:
screen shot 2015-10-22 at 21 36 50
screen shot 2015-10-22 at 21 37 03

Not sure why this is happening, but is there a workaround in the meantime?

camallen:
closes #1846 and also remove unused project_id filter and cleaned up some unsued code.

### DIFF
--- a/app/collections/manager.cjsx
+++ b/app/collections/manager.cjsx
@@ -18,10 +18,10 @@ module?.exports = React.createClass
     project: React.PropTypes.object
 
   addToCollections: ->
-    options = @refs.search.options()
-    return unless options
+    collections = @refs.search.getSelected()
+    return unless collections.length > 0
 
-    promises = for { collection } in options
+    promises = for { collection } in collections
       collection.addLink('subjects', [@props.subject.id])
 
     Promise.all(promises)
@@ -41,7 +41,6 @@ module?.exports = React.createClass
           ref="search"
           multi={true}
           project={@props.project}
-          user={@props.user}
           onChange={@handleChange} />
         <button type="button" className="standard-button search-button" disabled={if @state.hasCollectionSelected then true else false} onClick={@addToCollections}>
           Add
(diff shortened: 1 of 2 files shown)


app/components/collection-search.cjsx
```diff
@@ -8,40 +8,50 @@ module.exports = React.createClass
...
+            collection: collection
+          }
 
         callback null, {
           options: opts
         }
 
-  options: ->
-    return @refs.collectionSelect.state.options
+  addSelected:(collIds) ->
```
chrissnyder: `react-select` gives the currently selected options as the second argument to this callback. Maybe use that instead of trying to back into it yourself?

app/collections/manager.cjsx
```diff
@@ -18,10 +18,10 @@ module?.exports = React.createClass
     project: React.PropTypes.object
 
   addToCollections: ->
-    options = @refs.search.options()
-    return unless options
+    collections = @refs.search.getSelected()
```
chrissnyder: This feels weird to me. Components shouldn't reach directly into each other's state, as that introduces coupling that could get messy later on.

Since we need to know what collections are selected in the `CollectionsManager`, maybe that array should be state here, and not in `CollectionSearch`, having it passed up via callback? Or better yet, maybe just move the submit button down into `CollectionSearch`, since that state isn't used anywhere else other than to fire the request off.

app/components/collection-search.cjsx
camallen: Yeah, i did. But it only passes the current state of the selected options so is one behind on each onChange event and would require n+1 calls to get to n. I thought there would be a nice way to just call `selectComponent.selected()` instead of hassling with `.value() manipulations, etc.`

app/collections/manager.cjsx
camallen: Got it, will look to build a component that searches and adds.

chrissnyder:
Thinking about this more, going to merge this now since it fixes a bug. I'll look at my code style nitpicks in another PR.

:shipit:

PR #1891, opened 2015-11-05
Gives Grant his proper full name

mrniaboc:
It has been an absolute pleasure @chrissnyder. Thought I'd give you one last chance to use the "won't fix" label.


chrissnyder:
Or just an excuse to spam emojis :100: 

:8ball: says to :shipit: 

All aboard the :shipit: :train: !

PR #1904, opened 2015-11-11
Add clear button to user profile media

### LINKED ISSUE #1903: Add ability to remove profile pic
-- in addition to being able to upload a new/different one, have the ability to just remove it.

brian-c:
@aweiksnar mind giving this a shot? It works on a new account I just created, but on my usual dev account deleting my avatar _appears_ to replace it with an older avatar resource. @camallen any idea off hand why that might be happening?

Closes #1903.

### DIFF
--- a/app/components/image-selector.cjsx
+++ b/app/components/image-selector.cjsx
@@ -11,7 +11,7 @@ module.exports = React.createClass
     accept: 'image/*'
     maxSize: Infinity # In bytes
     ratio: NaN # Width / height
-    defaultValue: ''
+    src: ''
     placeholder: ''
     minArea: 300 # Stop reducing when there are fewer than this many pixels.
     reductionPerPass: 0.05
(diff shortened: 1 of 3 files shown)


aweiksnar:
This is working great for me - Tested pretty thoroughly on a few of my dev accounts and my ouroboros-created one and all were clearing the avatar when needed.

The only thing I noticed is dragging/dropping an avatar over a pre-existing one and then clicking clear deletes it in the api but doesn't clear the image-selector afterwards

camallen:
@brian-c nope, PM me the login of the account that's problematic.

brian-c:
Thanks @aweiksnar, fixed by removing the local image state from the ImageSelector component. It now relies on one to be passed down as `src` instead of `defaultValue`. That way there's nothing to get out of sync.

aweiksnar:
Awesome yeah that did it :shipit:

PR #1905, opened 2015-11-11
Text task

saschaishikawa:
* Create new task to transcribe text
* Add text tasks to mock-data workflow
* Include text task as experimental feature


app/pages/lab/workflow.cjsx
```diff
@@ -105,6 +106,14 @@ EditWorkflowPage = React.createClass
...
                     </button>
                   </AutoSave>{' '}
+                  {if @canUseTask(@props.project, "text")
+                    <AutoSave resource={@props.workflow}>
+                      <button type="submit" className="minor-button" onClick={@addNewTask.bind this, 'text'} title="Text tasks: the volunteer writes free-form text into a dialog box.">
+                        <i className="fa fa-pencil fa-2x"></i>
+                        <br />
+                        <small><strong>Text</strong></small>
+                      </button>
+                    </AutoSave>}
```
brian-c: Can we keep the spaces characters (`{' '}`) between these buttons? Otherwise they're inconsistently jammed together.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+    expanded: false
+
+  getInitialState: ->
+    expanded: @props.expanded
+
+  render: ->
+    <div className="classification-task-summary">
+      <div className="question">
+        {@props.task.instruction}
+        {if @state.expanded
```
brian-c: We can probably lose these buttons, and the expanded state altogether.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+          <button type="button" className="toggle-more" onClick={@setState.bind this, expanded: false, null}>Less</button>
+        else
+          <button type="button" className="toggle-more" onClick={@setState.bind this, expanded: true, null}>More</button>}
+      </div>
+      <div className="answers">
+      {if @props.annotation.value?
+        <div className="answer">
+          <i className="fa fa-check-circle-o fa-fw"></i>
+          <Markdown>{@props.annotation.value}</Markdown>
```
brian-c: The value should probably be considered plain text, not Markdown, eh?

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+
+  statics:
+    Editor: GenericTaskEditor
+    Summary: Summary
+
+    getDefaultTask: ->
+      type: 'text'
+      instruction: 'Enter an instruction.'
+      help: ''
+      answers: []
```
brian-c: I think `answers` got left over from another task.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+
+    getTaskText: (task) ->
+      task.instruction
+
+    getDefaultAnnotation: ->
+      value: null
+
+    isAnnotationComplete: (task, annotation) ->
+      annotation.value? or not task.required
```
brian-c: The combination of the default `value` (from `getDefaultAnnotation`) and the question mark here means that an empty string is "complete". I think I'd make the default value `''` (so the value is always a string) and test `annotation.value isnt '' or not task.required` here.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+
+  getDefaultProps: ->
+    task: null
+    annotation: null
+    onChange: NOOP
+
+  render: ->
+    answers =
+      <label className="answer">
+        <textarea ref="textInput" onChange={@handleChange} />
```
brian-c: Text inputs and textarea elements can take `className="standard-input full"` to fill their available width. Maybe give set `rows="5"` or so, too.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,74 @@
...
+      </label>
+
+    <GenericTask question={@props.task.instruction} help={@props.task.help} answers={answers} required={@props.task.required} />
+
+  handleChange: (index, e) ->
+    @props.annotation.value = @refs.textInput.getDOMNode().value
```
brian-c: `getDOMNode ` is deprecated. Use `React.findDOMNode(@refs.textInput).value` here.

saschaishikawa:
@brian-c: thanks for the feedback. those comments should now be addressed. lemme know if there's anything i missed

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,65 @@
...
+
+  getDefaultProps: ->
+    task: null
+    annotation: null
+    onChange: NOOP
+
+  render: ->
+    <GenericTask question={@props.task.instruction} help={@props.task.help} required={@props.task.required}>
+      <label className="answer">
+        <textarea className="standard-input full" rows="5" ref="textInput" onChange={@handleChange} />
```
brian-c: Missed this, sorry: the textarea needs to display the value of the annotation. `value={@props.annotations.value}` should do it.

app/classifier/tasks/text/index.cjsx
```diff
@@ -0,0 +1,65 @@
...
+      help: ''
+
+    getTaskText: (task) ->
+      task.instruction
+
+    getDefaultAnnotation: ->
+      value: ''
+
+    isAnnotationComplete: (task, annotation) ->
+      annotation.value? or not task.required
```
brian-c: Now that the value is `''` by default, this function will always return true. This line should probably be `annotation.value isnt '' or not task.required`

brian-c:
Lookin' fine, thanks!

PR #1906, opened 2015-11-11
Move external links in main nav to triggered modal

aweiksnar:
- Real-estate is getting pretty tight in the main nav. The external
  links are probably not essential to always be visible, so this moves them to a
  dropdown modal


brian-c:
Text in the popup is a bit giant on project pages (this might need a lower level fix, the root font size shouldn't be changing depending on what page you're on) and seems to disappear on hover.

aweiksnar:
@brian-c I took a stab at a lower level fix / cleaning up some of the css that was causing different secondary pages to behave differently. 

I think it makes sense to set the top level em font to the main size used on the page and scale from there instead of using 10px and accounting for it. Let me know if the changes look okay

brian-c:
Sizing looks solid now, thanks.

Links in the popup still turn white on hover though.

aweiksnar:
Oh good catch, wasn't getting that in FF. Just pushed an update

brian-c:
Awesome, thanks!

PR #1907, opened 2015-11-11
Preload assets used on main pages

aweiksnar:


brian-c:
Why load these on page load instead of as they're needed?

aweiksnar:
I think it provides a smoother experience for users changing between pages on the site for the first time. Would you rather not load these ahead of time?

brian-c:
The page already super heavy at load. I don't think we should be loading anything we don't have to until we need it.

aweiksnar:
Okay that's fair

PR #1909, opened 2015-11-13
Survey task styles

srallen:
- Added lots of styles and changed some markup for the survey task that'll be used for Wildlife Watch USA (aka Snapshot Wisconsin). 
- It was mostly borrowed from what I did for Wildcam. 
- :+1:  by @jwbmartin.

### DIFF
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -69,64 +69,63 @@ module.exports = React.createClass
     <div className="survey-task-choice">
       {unless choice.images.length is 0
         <ImageFlipper images={@props.task.images[filename] for filename in choice.images} />}
+      <div className="survey-task-choice-content">
+        <div className="survey-task-choice-label">{choice.label}</div>
+        <div className="survey-task-choice-description">{choice.description}</div>
 
-      <div className="survey-task-choice-label">{choice.label}</div>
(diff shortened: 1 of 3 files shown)


brian-c:
Can you point me to an example project?

srallen:
I've added you as a collaborator on the wildlife watch project and also to my test project on staging: http://demo.zooniverse.org/panoptes-front-end/#/projects/srallen086/survey-test/classify/?workflow=1905&reload=0

brian-c:
Thanks. Looks like layout/image display gets a little wonky as the list narrows down:

screen shot 2015-11-16 at 3 23 32 pm

screen shot 2015-11-16 at 3 24 03 pm

screen shot 2015-11-16 at 3 24 29 pm

srallen:
Cool, I just missed grabbing those styles fixing that in wildcam. Also got rid of the `white-space: nowrap`. It made sense in wildcam, but doesn't here.

brian-c:
Thanks, looks good, but I'm still only getting thumbnails when `data-breakpoint="20"`. Should be there in 10 and 5 too.

srallen:
Ok, I changed the markup to use a `img` tag instead of `span` and now am seeing images in the other breakpoints.

brian-c:
Great! Thanks for all that.

PR #1910, opened 2015-11-16
Styles for slide tutorial

srallen:
- Added a couple of styles to the `tutorial-dialog` class: 
  - `border-radius` to copy how the slide tutorial looks in older custom projects. 
  - `max-width` to keep the modal from being _slightly_ larger than the tutorial slide content. The modal form was a few decimal points larger in width causing a little white border on the right side.


brian-c:
Cool, I'm actually not sure what that ~400px is calculated from.

PR #1915, opened 2015-11-18
Stage to preview

brian-c:
Switch `stage` script from pushing to demo.zooniverse.org/panoptes-front-end to preview.zooniverse.org/panoptes-front-end.

Heads-up to all @zooniverse/developers.

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -13,7 +13,7 @@
   </head>
 
   <body>
-    <!--Last build: <%= Date.now() %>-->
+    <!--Last build: <%= new Date() %>-->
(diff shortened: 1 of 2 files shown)


views/index.ejs
```diff
@@ -13,7 +13,7 @@
   </head>
 
   <body>
-    <!--Last build: <%= Date.now() %>-->
+    <!--Last build: <%= new Date() %>-->
```
astopy: @brian-c Did you mean to include this change in this PR?

brian-c: Yep, sorry, should have mentioned it.

PR #1916, opened 2015-11-18
Set default save prefix to "~", switch existing dependencies

brian-c:
Think this is a good idea? I just did battle with npm trying to install an older React 0.13 compatible version of a now-React 0.14-compatible lib.


aweiksnar:
Yeah LGTM for sure, especially with auto-deploys :ship:

PR #1920, opened 2015-11-18
Remove subject default option from subject page

### LINKED ISSUE #1919: Remove ability to create discussion in Notes board from subject page
We've removed the ability to generate a new discussion in the Notes board from the Notes board itself...

... but you can still do it from the subject page.

screen shot 2015-11-18 at 3 50 55 pm

The default-threads board -- Notes or its equivalent -- shouldn't be listed as an option for a place to start a new discussion.

aweiksnar:
- Closes #1919


parrish:
:+1: :shipit:

PR #1927, opened 2015-11-19
Add project card hover/reveal effect

aweiksnar:
:copyright: @jwbmartin's wireframes

### DIFF
--- a/app/partials/owned-card.cjsx
+++ b/app/partials/owned-card.cjsx
@@ -52,8 +52,8 @@ module.exports = React.createClass
         owner: owner
         name: name
 
-    <div className="card" ref="ownedCard">
-      <FlexibleLink {...linkProps}>
+    <FlexibleLink {...linkProps}>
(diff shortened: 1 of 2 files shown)


brian-c:
Looks nice! A couple things:

`.card`'s `overflow: hidden` kills the keyboard focus ring.

Mobile/keyboard users won't see content on `:hover`.

jwbmartin:
Would it be possible to turn off :hover styling in mobile?

aweiksnar:
Just added card focus-ability in 6af10ee, That look okay or think it's worth revealing on `:active` too?

brian-c:
Just fixed focus on cards. Couldn't find a quick fix for mobile. @jwbmartin I suggest not using hover states to reveal information. Mobile won't get to see it, or worse, they'll see it for the split second before the page changes.

Looks good to me otherwise. @aweiksnar go ahead and merge if you're happy with my change.

PR #1934, opened 2015-11-23
Classification feedback

brian-c:
When the Classifier loads a subject and workflow, it'll also check for a gold standard classifications for the pair.

When the classification is submitted, it's given a "quality" score (which is how closely it matches the gold standard). All task modules have a new `testAnnotationQuality` comparison function. These can probably use some iteration.

The summary screen either says "great job" or "keep going" (we can play with wording here) depending on the quality. The summary UI will be updated in the future, for now it allows the user to flip between their classification and the expert's.

@chrislintott Current phrasing is: "Keep at it, all classifications are useful!" and "Keep up the good work!" Is that fine for now? I don't think too many projects have gold standard data, so it'll probably be super rare that anyone sees these at this point.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -1,5 +1,6 @@
 React = require 'react'
 apiClient = require '../api/client'
+testClassificationQuality = require '../lib/test-classification-quality'
 ChangeListener = require '../components/change-listener'
 SubjectAnnotator = require './subject-annotator'
 ClassificationSummary = require './classification-summary'
@@ -31,6 +32,7 @@ Classifier = React.createClass
   getInitialState: ->
     subjectLoading: false
     expertClassification: null
+    classificationQuality: NaN
     showingExpertClassification: false
     selectedExpertAnnotation: -1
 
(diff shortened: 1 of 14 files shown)


mschwamb:
I was thinking of starting gold standard data for Planet Four: Terrains, but I'll hold off if this is going to be the default behavior. 

Can there eventually be an option included to not have the message change based on the gold standard data? I'm not sure every project would want this. I use the gold standard as validation to what's being done by the volunteers collectively to show how accurate the entire project is, so then using it to encourage/discourage or change behavior changes how I could use it in analysis since now its part of the feedback. It would be helpful if there was an option to provide a set of data like spacewarps for influencing behavior that was separate (but could also be if projects wanted) the gold standard data.

brian-c:
@mschwamb this is made optional in a separate PR (gold-standard-import), coming real soon.

@aweiksnar I think this is fine to review and merge, It shouldn't actually affect anything yet, and we can always change wording later. If you're busy and you'd like to re-assign to someone else please do.

aweiksnar:
**Very** cool, `test-shape-closeness` looks nice too :+1:

PR #1951, opened 2015-11-30
collection search list defaults

### LINKED ISSUE #1950: fix the 10 collection limit for what's displayed on the dropdown menu for adding a subject to a collection and removing the favorite collections from this list
When I try and add to collections it's not listing my entire list of collections. New collections aren't showing on the list, also it's now listing favorites which is confusing because there is a favorites button. According to @camallen the front end is cutting at 10 items in the list, but that means people have to remember the names of their collections, which isn't so easy if you're working on many projects. 

More detailed discussion on this closed backend ticket https://github.com/zooniverse/Panoptes/issues/1513#issuecomment-160588043

Example - 

My  collections:

7745e932-9424-11e5-8f59-9a49dcc7530a

What is showing up in the interface when I try and add to a collection:

90a9d262-9424-11e5-8d6b-e5ac42340095

95b8bc6e-9424-11e5-95cf-090db857d0cc

camallen:
closes #1950 - bump default collection page size and exclude fav's by default

@aweiksnar perhaps we should add a checkbox to include / exclude favs (default to exclude) as the fav button makes sense for fav collecting.

### DIFF
--- a/app/components/collection-search.cjsx
+++ b/app/components/collection-search.cjsx
@@ -15,7 +15,8 @@ module.exports = React.createClass
 
   searchCollections: (value, callback) ->
     query =
-      page_size: 10
+      page_size: 20
+      favorite: false
       current_user_roles: 'owner,collaborator'
     query.search = "#{value}" unless value is ''
 


aweiksnar:
:shipit: Looks good. I'll post up a separate issue for favorites toggling :+1:

Issue #1958, opened 2015-12-01
Add customizable message box for PFE projects


DarrenMcRoy:
A feature that has been tossed around for a while, recently brought up again by @chrislintott -- we should have some sort of open text box, ideally somewhere on *both the home landing page and in the classification interface*, that allows for research teams to deliver an announcement or a message.

Sort of like a static intervention that appears consistently (maybe is hideable by the user?) that the project owner could change/update within the Project Builder.

This could be used for everything from announcing challenges to displaying paused/completed messages to inviting people to search for and tag a particularly important object... the potential uses are legion.

Thoughts?

brian-c:
Closed by #2308

PR #1959, opened 2015-12-01
Add a banner to completed projects

brian-c:
This is a quick little "All done" message for projects that've run out of subjects.

Here's how you can see it locally on a production project: http://localhost:3735/projects/vrooje/galaxy-zoo-bar-lengths/classify?env=production

### DIFF
--- a/app/pages/project/finished-banner.cjsx
+++ b/app/pages/project/finished-banner.cjsx
@@ -1,0 +1,56 @@
+React = require 'react'
+
+module.exports = React.createClass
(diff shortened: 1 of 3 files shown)


DarrenMcRoy:
Does it appear on the classification page, the home page, or both?

brian-c:
It appears on every project page.

DarrenMcRoy:
Brilliant. Love it. Might be a good fix for #1958 in the long run.

PR #1963, opened 2015-12-01
Revert "Add a banner to completed projects"

brian-c:
Reverts zooniverse/Panoptes-Front-End#1959


brian-c:
- Link to results page is broken because routes sometimes end in slashes and sometimes don't. (Fix the routes, things should have one right URL).

- Non-PFE projects (like Old Weather) that use PFE talk are confused. (Maybe check for existence of workflows? `redirect` field? Needs more research).

camallen:
@brian-c what does point 2 mean, when you say confused?

PR #1965, opened 2015-12-02
Subject viewer frame wrapping

itsravenous:
- [x] Handle undefined `FrameWrapper` prop
- [x] Shift decision to show flipbook toggle to a prop, allow parent component to make that decision (when classifying it depends on the workflows config, but e.g. talk pages probably want to always show it)
- [x] Related to above: don't rely on `@props.workflow` existing - use existential operator
- [x] Play nicely with new video controls


itsravenous:
This is all good to go as far as I can see - could probably use some more testing but functionality wise I'm happy. One thing to note: The layout of the images is controlled by a project builder option on the workflow. This applies when you have multiple videos too, though the option currently doesn't make that clear. As videos are still admin-only at the moment I'm thinking just leave the wording on the option to stay referring to images, rather than trying to cover video in the same sentiment. Thoughts @eatyourgreens, @vrooje and anyone else?

eatyourgreens:
I'm having a hard time imagining a case where you would want to watch multiple videos as part of the same subject. Some sort of comparison task, perhaps?

I'm okay with leaving the wording as it is for the moment. Easy enough to change it when we need it.

vrooje:
Maybe in a "which CME was bigger" version of Solar Stormwatch? 

Either way I'm happy with leaving the wording as-is...

itsravenous:
Great! I'm done here then - let me know if anything else is required before merging.

app/classifier/frame-annotator.cjsx
```diff
@@ -0,0 +1,148 @@
...
+
+  handleChildLoad: (e) ->
+    @props.onLoad? e, @props.frame
+    {naturalWidth, naturalHeight} = e.target
+    unless @props.naturalWidth is naturalWidth and @props.naturalHeight is naturalHeight
+      @setState {naturalWidth, naturalHeight}, =>
+        @updateSize()
+
+  handleFrameChange: (frame) ->
+    @setState {frame}
```
eatyourgreens: Should frame changes be handled at a higher level, by a component which has access to the whole subject? I'm confused here, because you have an immutable frame number via `props.frame` but also have a frame number stored in `state.frame`, which can be changed. Which frame should this component display?

eatyourgreens:
Can you squash some commits on this? Maybe squash the 'remove debug' commits and the one-line fixes?

app/classifier/frame-annotator.cjsx
itsravenous: Sorry - this is an accidental leftover. The frame should only be in `@props` - `handleFrameChange` and `@state.frame` aren't used. Will clean this up when I'm on zoowork tomorrow :)

itsravenous:
Will do tomorrow.

itsravenous:
@eatyourgreens at one point (0a3d687) I merged in master because I wasn't smart enough to understand why rebase wasn't doing what I wanted. This is making squashing commits difficult - any hints (I'm pretty new to rebasing)?

eatyourgreens:
Are you rebasing against master? `git rebase -i master`

That should only list the commits that are new in this branch (listed in the PR here.)

app/classifier/frame-annotator.cjsx
```diff
@@ -0,0 +1,148 @@
...
+          <AfterSubject {...hookProps} />}
+      </div>
+    </div>
+
+  handleChildLoad: (e) ->
+    @props.onLoad? e, @props.frame
+    {naturalWidth, naturalHeight} = e.target
+    unless @props.naturalWidth is naturalWidth and @props.naturalHeight is naturalHeight
+      @setState {naturalWidth, naturalHeight}, =>
+        @updateSize()
```
eatyourgreens: Has this been accidentally left in from the Subject Viewer? Looks like `naturalHeight` and `naturalWidth` should be updated elsewhere and passed back in as props.

itsravenous: Must have been, as it's not referred to at all, though I can't find it in SubjectViewer now either.

eatyourgreens: Looks like it was there to handle the case where each frame has a different size.

PR #1971, opened 2015-12-03
Return of the finished project banner

brian-c:
Let's try again.

This time I'm skipping `redirect`ed projects and projects with no active workflows, and using the proper (for now) `Link` component.

It now only appears on the landing and classify pages.

There's also a "dismiss" button that'll hide it for three days, which is arbitrary but sane, I think.


camallen:
Had a play works well.

PR #1977, opened 2015-12-07
bugfix: listen to the correct event when video loads

### LINKED ISSUE #1974: Loading indicator remains after video files have loaded
I think the video element should be using `onLoadedData`, not `onLoad`, on this line:
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/subject-viewer.cjsx#L58

eatyourgreens:
Closes #1974

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -47,6 +47,9 @@ module.exports = React.createClass
     loading: true
     playing: false
     frame: @props.frame ? 0
+  
+  componentDidMount: ->
+    @refs.videoPlayer?.getDOMNode().addEventListener 'canplaythrough', @handleLoad
 
   render: ->
     {type, format, src} = getSubjectLocation @props.subject, @state.frame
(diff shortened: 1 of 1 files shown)


brian-c:
I haven't been able to get videos to load on my end. I'm trying an MP4 from http://www.sample-videos.com/, should that work?

This way of handling the event looks good, but, would it make more sense to listen for `canplaythrough`?  `loadeddata` only waits for the first frame.

Also, can we remove the listener once one the handler fires?

eatyourgreens:
Good call. I've updated to use `canplaythrough`.

I'm not sure what's going on with the sample MP4. Codec issues maybe? @trouille has a project with a test mp4 video. That works for me in Chrome and Firefox.

eatyourgreens:
Scratch that: the test mp4 works in Chrome but Firefox isn't downloading it.

brian-c:
I get the audio of the sample MP4 to play through Chrome, but not the video, and Firefox says "Video can't be played because the file is corrupt".

Mind sending me your example video?

eatyourgreens:
@brian-c just sent you a link in Slack to the test project.

Issue #14, opened 2014-10-30
Faster dev builds


brian-c:
@aweiksnar @parrish re: the faster-dev-builds branch.

Cool on Michael for finding the watchify fork with the delay option. I'll keep an eye on substack/watchify#24 to see when we can switch this to the normal release.

Any objection to pulling React from a CDN? It's shimmed into browserify, so `require('react')` still works, but `require('react/lib/merge')` (for example) no longer will.

PR #1982, opened 2015-12-07
Skip classification summary option

itsravenous:
Adds a lab option to enable disabling the classification summary. Talk is made accessible via an extra "Done & Talk" button.

Ali S mentioned this might need to be an "experimental" option, looking at how to stipulate that now but rest is good to go.


itsravenous:
(Conflict is just the new tutorial button in `app/classifier/index`)

aliburchard:
Thanks for pushing this forward @itsravenous!  I'm making the uneducated guess that the conflict is because the "done & talk" button overlaps the new "tutorial" button? Would it be possible to get around this by making them smaller?

@brian-c I know you're not thrilled with a "done" and a "done and talk" button as a general solution, but we're looking to implement this on a couple of very specific "collaborative filtering" style projects as a test case. Most immediately, @chrislintott and @mrniaboc and I are hoping to set this up on Snapshots at Sea this week so we can launch the first of a series of yes/no mini-projects. Are there any obstacles to wiring this functionality up to that specific project (https://www.zooniverse.org/lab/14/)? I've added you and Tom as collaborators on it just now.

Thanks!

itsravenous:
@aliburchard meant conflict only in the Git sense :) At a glance the new tutorial button seems to fit alongside ok (it's the ? in a circle, right?)

brian-c:
@ali Not a huge fan of this solution for two reasons, which we talked about in Oxford:

- It should not be an option. Projects with near-identical UIs should not behave differently based on the project author's preference. This has nothing to do with a project's science and everything to do with the UI of the workflow, which should be predictable regardless of the project.

- Having both "Done" and "Done & Talk" buttons is just super inelegant. I thought we were going to go with a Kelp-style image of the previous (few?) subject(s?) under the classifier.

chrislintott:
@brian-c Re your first point - rather than making it an option, what if projects which have only a single question (or a single layer to their task) behaved like this? It's this use case that I'm worried about. 

Re your second - I think the Kelp style thing is too big a change from standard functionality and UI. I'm comfortable with this, though we might need to relabel...

mschwamb:
@chrislintott one note - Planet Four: Terrains is one question and we're getting a lot out of Talk including finding new regions to target HiRISE. It's been extremely productive for us. Obviously the needs of the wide platform outweigh a single project, but if possible we would like to still be able to keep the summary page entrance to Talk even though we only have a single question (multiple answer) in our decision tree.

aliburchard:
@brian-c I totally agree that this shouldn't be a publicly available _option_, but it's also something we really need for *specific* projects that use what we've been calling a "collaborative filtering" approach. These projects will have just a single question (typically a yes/no question), and so it's really important that users are able to move rapidly onto the next subject with just one click. I really like @chrislintott's suggestion of ultimately making this "skip summary" the default behaviour for one-click projects (and agree with @mschwamb that we need to provide volunteers with a route into talk).

I know done/done&talk isn't the most elegant solution, it is less of a UI change than Kelp and straightforward to implement. Providing access to the "recents" feed came up as an option when @itsravenous was working on this, but we realized that that solution only works for users who already know that 'Talk' exists. 

What if we flag this as an experimental feature and wire up this "skip summary" functionality just to a specific project (Snapshots at Sea) for testing? There won't be any way for project builders to implement this in the project builder, so we don't have to worry about project builders using this before we've figured it out fully.  This way we can track whether folks click on the Done&Talk button and continue to brainstorm alternative, long-term solutions.

mschwamb:
Original Planet Four the design actually stops most people from ever getting to Talk. We have a next and Talk button and everyone clicks next since it's above the Talk button. We've gotten way more traffic on  Talk on Planet Four: Terrains because of the current summary page so I do worry about the rapid click changing that behavior for us, and since we're looking for weird stuff to get high resolution orbiter images of  Talk has been super handy. 

 I understand that the needs for more projects outweigh any input from the Planet Four team. I like @aliburchard suggestion as an experimental feature as it would be keep Terrains the same but allow you to build your rapid review projects like Sunspotter where Talk might not be important.

aliburchard:
@brian-c what do you think about these suggestions?

aliburchard:
inspired by conversation with @chrislintott and @rogerhutchings what if the buttons were labeled "Done & Next" (with smaller text below saying "move on to the next subject") and "Done & Talk" (with small text below saying "discuss this image on talk"

brian-c:
Just spoke with @chrislintott, will merge as-is for now.

@itsravenous, mind rebasing from master? Thanks!

itsravenous:
@brian done! Had to fetch and merge my remote branch afterwards, which I'm confused about. I haven't done rebasing in a long time, so let me know if I've done anything stupid. Tested merging to latest master and it fast-forwarded so I think all is well.

mschwamb:
so question - that stops Planet Four Terrains from having a Talk button - what do I do to fix that?

brian-c:
@mschwamb This is opt-in, it won't affect anything except the project we activate it for.

mschwamb:
Great. Thanks for the clarification @brian-c

aliburchard:
Thanks so much @brian-c and @itsravenous! 

One last question...how do I get it set up for Snapshots at Sea? (Project #14)

itsravenous:
Once it's deployed you should be able to enable it as an experimental feature. Then enable it for a particular workflow by ticking "Hide classification summaries"

aliburchard:
Got it, thanks :). Alright, so I've got it up and running and it looks great! (Have a thought for down the line - if we only enable it for 1-task projects, then we can get rid of the "back" button.) But I've noticed one tiny issue: the Done&Talk button doesn't respect the "required" flag on questions. It shouldn't be clickable until the user has answered a question that is marked as retired in the project builder:

screen shot 2015-12-17 at 10 27 49

itsravenous:
Good catch! Logged at #2037.

PR #1987, opened 2015-12-08
tag select added to project builder

CKrawczyk:


CKrawczyk:
@alexbfree can you merge this in

PR #1991, opened 2015-12-09
Miniztm

amyrebecca:
Sorting and searching work on the projects page from MiniZTM


alexbfree:
Thanks Amy! looks good :-)

PR #1896, opened 2015-11-09
ignore artifacts from build process on linux

camallen:

PR #1995, opened 2015-12-10
Fix Linux scripts

brian-c:
Closes #1896. From what I can tell, in Linux `npm run-script` uses `sh` instead of the user's shell. I just pulled out the script the uses brace expansion and explicitly made it a bash script (and tweaked another one).


camallen:
:shipit:

PR #1997, opened 2015-12-11
Scroll up between project list pages

itsravenous:
Added this as a callback on the card list - i.e. the parent component/page decides what to do on page change - I can imagine cases where you wouldn't want to jump to the top of the document when changing the list page (e.g. a list buried halfway down a long page)

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -31,6 +31,9 @@ module.exports = React.createClass
     else
       'All'
 
+  onPageChange: (page) ->
+    @props.onPageChange? page
+
   render: ->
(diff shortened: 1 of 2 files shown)


brian-c:
Components generally shouldn't concern themselves with managing scroll position on navigation.

Looks like something's been written to deal with this, probably worth looking into: https://github.com/rackt/scroll-behavior

amyrebecca:
May want to consider this issue as part of #2149 ?

marten:
Shall we either close or merge this?

srallen:
Let's close this. The project page overhaul will be done soon and this functionality can be added via the scroll-behavior router plugin mentioned above in a PR later.

PR #2000, opened 2015-12-11
Revert "Workflow visualization"

brian-c:
Reverts zooniverse/Panoptes-Front-End#1808

Deploy doesn't copy the jsPlumb file correctly.


brian-c:
@CKrawczyk why's the jsPlumb file being copied into /public? Can't it be `require`d?

CKrawczyk:
There is a bug in the jsPlumb npm build of the package and it does not require properly.  The copy was a workaround.

PR #2001, opened 2015-12-11
Allow running postinstall in docker

astopy:
It wasn't running before because of a permissions issue (npm drops privs
by default, whereas Docker builds as root by default).

Also make sure postinstall.sh fails on errors.

**This is dependent on the [pfe-base build currently running on Docker Hub](https://hub.docker.com/r/zooniverse/pfe-base/builds/bz9x3hfdaxxbiqrx9neuufg/). Don't merge until that's finished.**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,4 +4,4 @@ WORKDIR /src/
 
 ADD . /src/
 
-RUN npm install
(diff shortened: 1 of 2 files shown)


astopy:
@brian-c You should be good to reinstate #1808 now.

PR #2002, opened 2015-12-11
Ensure user search callback is called

aweiksnar:
- Prevents the user search box from hanging in the empty state


parrish:
Good catch!  :+1:

aweiksnar:
Thanks!

Issue #2004, opened 2015-12-12
Can no longer get the link to the last thread in a discussion thread from the Talk front pages


mschwamb:
Now no matter what you click on,  it takes you to the start first page of the Talk thread for the most recent discussions on the project Talk front pages. I tried clicking the title and the time stamp and both have in the header that same thing. There's nothing in there in the address directing the api to bring up the most recent post. This is definitely a bug. I've tried in Firefox and Chrome. 

I'm talking about the links on the front page of Talk and those on the individual discussion board lists like https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/8

@aweiksnar and @parrish is this something you could fix on Monday. I think this is an extremely used feature. I know I use it frequently.  I'm not 100% sure when the changed happened as I've been spending more time on Comet Hunters talk which since the project is in beta has much shorter discussion threads. I only noticed it when dealing with a post on Planet Four: Terrains

mschwamb:
Looks like it's bigger - no links work for notifications it now has a typo it doesn't add the / after the zooniverse.org - did the redirect change or something?

https://www.zooniverse.orgprojects/mschwamb/planet-four-terrains/talk/10/20781?comment=42625

aweiksnar:
@mschwamb Just opened a PR to fix the latest-comment-linking issue, Will try and hunt down the notifications link later today!

mschwamb:
Thanks for doing this on a weekend @aweiksnar - enjoy your Saturday!

aweiksnar:
@mschwamb :bow: Yeah we shipped a big dependency update on Friday afternoon so I figured something would have gone under the radar

parrish:
PR #2004 opened for the notification links.

mschwamb:
Thanks @parrish and @aweiksnar

mschwamb:
Especially with it being the weekend - thanks for responding.

PR #2005, opened 2015-12-12
Add query back to latest-comment-link

aweiksnar:
- Links to last comment/page of discussions instead of the first, mentioned in #2004


parrish:
:shipit:

PR #2006, opened 2015-12-12
Fixes links from comment notifications.  Closes #2004

### LINKED ISSUE #2004: Can no longer get the link to the last thread in a discussion thread from the Talk front pages

parrish:
This is the only place I found a broken link, so I'm pretty sure this should close the issue.

### DIFF
--- a/app/talk/comment-link.cjsx
+++ b/app/talk/comment-link.cjsx
@@ -24,7 +24,7 @@ module?.exports = React.createClass
   projectCommentUrl: ->
     {comment} = @props
     [ownerName, projectName] = comment.project_slug.split('/')
-    href= "projects/#{ownerName}/#{projectName}/talk/#{comment.board_id}/#{comment.discussion_id}?comment=#{comment.id}"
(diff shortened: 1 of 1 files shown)


aweiksnar:
:ship: Thanks!

Issue #1972, opened 2015-12-03
After using the preview button and submitting the post, it doesn't come back to the writing mode


yshish:
When you use the preview button, and submit your post, the text field stays in the preview mode for the next comment too. So it looks like you aren't allowed to type any more. It shouldn't stay in the preview mode for the next post but should return to the writing mode automatically.

itsravenous:
The comment boxes use [markdownz](https://www.npmjs.com/package/markdownz) for their editor, which doesn't expose its preview state for manipulation. Looking into what needs to happen to change that.

itsravenous:
Markdownz issue sorted and merged, just need to wait for #1981 as markdownz depends on React 0.14

PR #2014, opened 2015-12-14
Reset commentbox to edit mode after submission

itsravenous:
Addresses #1972.


aweiksnar:
Clever use of that loading state  :shipit: :100:  I like that it disables editing while submitting too :+1:

PR #2015, opened 2015-12-14
Fix broken tag select

### LINKED ISSUE #2010: Can't see anything you're typing in the tags line on the project details page
You can't see any of the text you type in the line for tags in the project builder now at least for Firefox. It definitely wasn't like that a few months ago. I can't see any text and if I have a typo and hit delete it deletes the text and the tag before

eatyourgreens:
onInputChange passes the input value, not an event.
Closes #2010

### DIFF
--- a/app/components/tag-search.cjsx
+++ b/app/components/tag-search.cjsx
@@ -11,15 +11,13 @@ module.exports = React.createClass
 
   searchTags: (value, callback) ->
     if value is ''
-      callback null, { options: [] }
+      { options: [] }
     else
       apiClient.type('tags').get search: "#{value}", page_size: 10
         .then (tags) =>
           opts = for tag in tags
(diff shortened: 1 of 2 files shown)


eatyourgreens:
This seems to fix the broken text input, but I'm still getting `Uncaught (in promise) TypeError: Cannot read property 'options' of undefined` from `react-select`. That error gets thrown on the user search autocomplete too.

aweiksnar:
Thanks :shipit: Looks like a few api changes with the newest `react-select`. I'll send a PR back your way for the async warning

eatyourgreens:
Cheers. I think the second commit should take care of the async warning.

aweiksnar:
Yep it does :beers: - I just added in #2016 too, just to ensure so the callback is called so it doesn't have the loading in spinner in the empty state

PR #2026, opened 2015-12-15
Fix admin project links

### LINKED ISSUE #2025: Admin panel - project card links are broken
When I want to update a project's status - say, re-LIVE-ing Whales as Individuals after the PI has added new data, I go to the admin panel and click on that project's card. However, instead of taking me to that project's admin page, it goes here: https://www.zooniverse.org/admin/project_status/%7Bowner%7D/%7Bname%7D

Looks like the owner/project variables aren't being parsed correctly into the url? Assume this is a front-end issue @brian-c but could also be a back-end one @camallen or @marten?

eatyourgreens:
Fixes #2025


aweiksnar:
:shipit:

PR #2029, opened 2015-12-15
Uses optimized /projects?homepage=true endpoint and eliminates unnecessary API calls

parrish:
:warning: This depends on zooniverse/Panoptes#1552.  Don't merge until that's live.

### DIFF
--- a/app/pages/projects.cjsx
+++ b/app/pages/projects.cjsx
@@ -22,7 +22,7 @@ module.exports = React.createClass
     query = {include: 'avatar'}
 
     if !apiClient.params.admin
-      query.homepage = true
+      query.simple = true
 
     apiClient.type('projects').get Object.assign {}, query, @props.location.query
 
--- a/app/pages/projects.cjsx
+++ b/app/pages/projects.cjsx
@@ -22,14 +22,12 @@ module.exports = React.createClass
     query = {include: 'avatar'}
 
     if !apiClient.params.admin
-      query.launch_approved = true
+      query.homepage = true
 
     apiClient.type('projects').get Object.assign {}, query, @props.location.query
 
   imagePromise: (project) ->
-    project.get('avatar')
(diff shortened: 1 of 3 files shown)


camallen:
api for this should be deployed now, be nice to see how much faster it is.

parrish:
Do you want it to automatically use zooniverse/panoptes#1552?

If not, then we should remove the homepage query param and just let it come from the URL if we want to test it.

camallen:
it'd be nice to just test the different end points manually and then decide which one is better. I was just letting folk know the staging api is ready for testing this.

camallen:
this is really fast, i've tested on both homepage=true and cards=true. card needs the `http://` prefix added to avatar_src for reference.

parrish:
> card needs the http:// prefix added to avatar_src for reference

Protocol relative urls ought to be fine, shouldn't they?  I had left it relative since it was broken in a HTTP local environment.  Not terribly picky here though.

camallen:
you're right - i should just be prefixing the avatar url like you are in the fast serializer with a `//`

app/pages/projects.cjsx
```diff
@@ -22,14 +22,12 @@ module.exports = React.createClass
     query = {include: 'avatar'}
 
     if !apiClient.params.admin
-      query.launch_approved = true
+      query.homepage = true
```
brian-c: What's `homepage` mean here? This is the projects page. Still looks like the expected response though.

camallen: an api param override to select the fast serializer. there is another that i've got called `?cards=true` as well.

parrish: Yeah, it's a dumb query param name.

parrish:
Now just waiting on zooniverse/Panoptes#1581 to merge/deploy

camallen:
this should be good to go now

PR #2035, opened 2015-12-16
Validate markdown help table nesting

mcbouslog:
Clean-up markdown table nesting issues.

### DIFF
--- a/css/markdown-editor.styl
+++ b/css/markdown-editor.styl
@@ -62,6 +62,8 @@
 .markdown-editor-help
   margin: 2em
   table
+    margin: auto
+    width: 100%
     border-collapse: collapse
   td
     border: 1px solid black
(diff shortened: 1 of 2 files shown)


css/markdown-editor.styl
```diff
@@ -62,6 +62,7 @@
 .markdown-editor-help
   margin: 2em
   table
+    margin: auto
```
aweiksnar: Thoughts on adding in `width: 100%` here too? So it'll align the tables vertically and fill the container

mcbouslog: Yeah I like that, looks better!  I think I properly updated, fixuped and forced push back, but let me know if not.

aweiksnar: :white_check_mark:

aweiksnar:
:surfer: Cool cool, this definitely takes care of those react nesting warnings

PR #2038, opened 2015-12-17
Confirm workflow deletion

brian-c:
Looks like something got screwed up with workflow deletion at some point and confirmation and state/error reporting was removed. This restores it. Also entirely possible it was never in there.

Cc @DarrenMcRoy.


eatyourgreens:
:+1: looks good here.

PR #2039, opened 2015-12-17
Video control fixes

### LINKED ISSUE #2036: New tutorial button pushes the toggle  (& collection button in Firefox) into a werid spot
@brian-c small fix we need.  The newTutorial button looks great and it very clear. Strange thing happened, it moved the toggles for multiple images in a subject and the collection button  into a weird spot for FIrefox and just moved the toggle for Chrome and Safari.  Can you or someelse adjust that? Not a huge deal since we've got the play button working fine

Looks like from non-logged in that is just slid the toggle to the right - looks different in fire fox than chrome

Firefox:

screen shot 2015-12-17 at 9 01 15 am

Chrome:

screen shot 2015-12-17 at 9 06 48 am

Safari:

screen shot 2015-12-17 at 9 08 21 am

Here's what it used to look like: 

comet_hunters_blinking4

So it's only in Firefox that the collection button is impacted and the toggle is on two lines. The toggle used to in the middle, however you want to adjust this fine by us  and I'll update the tutorial if need be

eatyourgreens:
Use a range slider for the scrubber.
Fix inline styles in the subject tools markup.
Fixes #2036

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -47,9 +47,11 @@ module.exports = React.createClass
     loading: true
     playing: false
     frame: @props.frame ? 0
-    time: 0
     playbackRate: 1
   
+  componentDidMount: ->
+    @refs.videoScrubber?.value = 0
+    
   componentDidUpdate: ->
     @refs.videoPlayer?.playbackRate = @state.playbackRate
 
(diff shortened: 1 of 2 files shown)


aweiksnar:
:shipit:

PR #2043, opened 2015-12-17
Clean up collections manager react-select

### LINKED ISSUE #2040: When you select a collection from the drop down menu it's name changes to a number
I'm not sure why this is happening, but when you select the collection from the drop down menu the form autofills to a number. I find it a bit confusing at first since no good way to know if you've accidentally selected the wrong collection. 

screen shot 2015-12-17 at 11 10 24 pm

screen shot 2015-12-17 at 11 10 28 pm

aweiksnar:
- Closes #2040

### DIFF
--- a/app/components/collection-search.cjsx
+++ b/app/components/collection-search.cjsx
@@ -10,9 +10,6 @@ module.exports = React.createClass
     multi: false
     project: null
 
-  getInitialState: ->
-    collections: []
-
   searchCollections: (value, callback) ->
     query =
       page_size: 20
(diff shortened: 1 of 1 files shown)


aweiksnar:
mention-bot is alive! :octocat:

camallen:
Nice one mention-bot! @aweiksnar so the reason i initially used the onChange hook was the selected state was always 1 behind what it should have been, see [here](https://github.com/zooniverse/Panoptes-Front-End/pull/1886#discussion_r43690069). I assume that i just missed the correct API call from react-select but can you make sure that when adding multiples the selected state is reported correctly?

aweiksnar:
Ah gotcha! Yeah I think reading the values state from the react-select will work here too, been testing things out and the selected state seems to always be in sync with single/multiple selections

camallen:
LGTM - i won't have time to test till tomorrow, anyone else?

PR #2051, opened 2015-12-21
Flexible survey

amyrebecca:
This adds a new survey-related task to the system, "flexibleSurvey." FlexibleSurvey works like a regular survey, but adds the ability to specify which questions belong to which choices by the way that an additional field is filled out in the questions csv file.

It's of course silly to have two survey tasks, but it was considered too risky to modify the existing task directly because we have deployments who are consuming it.

I'm looking to merge this onto master tomorrow so that we can deploy it for WiscWildWatch

[sample spreadsheet](https://github.com/zooniverse/Panoptes-Front-End/files/69062/surveyTask-spreadsheets-inclHumanNnothing-multiQuestions-v2.xlsx)

### DIFF
--- a/app/classifier/tasks/flexible-survey/annotation-view.cjsx
+++ b/app/classifier/tasks/flexible-survey/annotation-view.cjsx
@@ -1,0 +1,35 @@
+React = require 'react'
+
+module.exports = React.createClass
+  displayName: 'SurveyAnnotationView'
+
+  getDefaultProperties: ->
+    task: null
+    classification: null
+    annotation: null
+
+  render: ->
+    <div>
+      {for identification, i in @props.annotation.value
+        identification._key ?= Math.random()
+
+        answersByQuestion = @props.task.questionsOrder.map (questionID) =>
+          if questionID of identification.answers
+            answerLabels = [].concat(identification.answers[questionID]).map (answerID) =>
+              @props.task.questions[questionID].answers[answerID].label
+            answerLabels.join ', '
+        answersList = answersByQuestion.filter(Boolean).join '; '
+
+        <span key={identification._key}>
+          <span className="survey-identification-proxy" title={answersList}>
+            {@props.task.choices[identification.choice].label}
+            {' '}
+            <button type="button" className="survey-identification-remove" title="Remove" onClick={@handleRemove.bind this, i}>&times;</button>
+          </span>
+          {' '}
+        </span>}
+    </div>
+
+  handleRemove: (index) ->
+    @props.annotation.value.splice index, 1
+    @props.classification.update 'annotations'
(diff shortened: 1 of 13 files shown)


amyrebecca:
Brian, I know you're busy, but would it be possible for you to look at this? Everyone suggested you because you'd written the original survey task!

app/classifier/tasks/flexible-survey/annotation-view.cjsx
```diff
@@ -0,0 +1,36 @@
...
+    task: null
+    classification: null
+    annotation: null
+
+  render: ->
+    <div>
+      {for identification, i in @props.annotation.value
+        identification._key ?= Math.random()
+
+        #TODO: ARB: what is happening here
```
camallen: stray comment?

app/classifier/tasks/flexible-survey/choice.cjsx
```diff
@@ -0,0 +1,153 @@
...
+          for index in [0...@props.images.length]
+            <span key={@props.images[index]}>
+              <button type="button" className="survey-task-image-flipper-pip" disabled={index is @state.frame} onClick={@handleFrameChange.bind this, index}>{index + 1}</button>
+              {' '}
+            </span>}
+      </span>
+    </span>
+
+  renderPreload: ->
+    <div style={@PRELOAD_STYLE}>
```
camallen: should we be using css here instead?

camallen:
This is a pretty big chunk of work, i won't have time to test it out properly sorry.

app/classifier/tasks/flexible-survey/utility.coffee
```diff
@@ -0,0 +1,8 @@
+module.exports =
+  getQuestionIDs: (task, choiceID) ->
+    return [] if task.choices[choiceID].noQuestions
+    return task.questionsOrder unless task.questionsMap? and choiceID of task.questionsMap
+    task.questionsMap[choiceID]
+  getQuestions: (task, choiceID) ->
+    @getQuestionIDs(task, choiceID).map (idx ->
```
aweiksnar: I think this function is unused, but if I'm wrong `idx` would need to be wrapped in parens

app/classifier/tasks/flexible-survey/choice.cjsx
brian-c: This style is required for the module to function right, so I think including it in the module is the right call.

aweiksnar:
:cool: :cool: Also for github history's sake: running `diff app/classifier/tasks/survey/ app/classifier/tasks/flexible-survey/` gives a better representation of the files changed from the original survey task

PR #2054, opened 2015-12-22
Ensure workflow on mount

### LINKED ISSUE #2048: Snapshots at Sea problem - Get no buttons to classify if I click talk and done on a subject and then classify again and submit a reply interface stops repsonding
The whole interface stops responding and all you see is the image if you classify an image as done & talk and then go back to classify and click done or done and talk cc @aliburchard @brian-c @chrislintott 

screen shot 2015-12-21 at 11 51 35 pm

itsravenous:
Fixes #2048. @vrooje @eatyourgreens this addresses the issue in a non-hacky way, and retains the same subject (and any unfinished classification) when returning to "classify" from any other project route (research, talk, etc).

As [mentioned](https://github.com/zooniverse/Panoptes-Front-End/issues/2048#issuecomment-166705181) the fix is pretty simple: `@state.workflow` is restored in `componentDidMount` (which is called when hopping back to the classify page). Maybe this is helpful in general too.

Also addressed is @eatyourgreens concern that the Done & Talk calls `@loadAnotherSubject` without waiting for `@saveClassification`. Would have separated these but my laptop was dying and just needed to push the changes up before it went completely (writing this on other half's Windows laptop)!

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -76,6 +76,13 @@ module.exports = React.createClass
     project: 'loadAppropriateClassification'
     query: 'loadAppropriateClassification'
 
+  componentDidMount: () ->
+    @getCurrentWorkflowID().then (id) =>
+      @getWorkflow @props.project, id
+        .then (workflow) =>
+          @setState
+            workflow: workflow
+
   loadAppropriateClassification: (_, props = @props) ->
     # To load the right classification, we'll need to know which workflow the user expects.
     # console.log 'Loading appropriate classification'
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Deployed to http://preview.zooniverse.org/panoptes-front-end/itsravenous-preserve_workflow_in_classify_page/?env=production

Looks like the bug is fixed, so I'm going to merge this, but I do get this warning from React about the ProjectLinker.
`Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op. Please check the code for the ProjectLinker component.`

itsravenous:
Hmm, I don't recall touching the ProjectLinker, but very possible I've caused that indirectly.

camallen:
@itsravenous can you file an issue to investigate this warning and put some thoughts down? be nice not to lose track of any issues the PR introduced.

PR #2060, opened 2015-12-28
Bump history verison to allow installation

brian-c:
Something was complaining enough to prevent installation, react-router I think.

I believe this is because we prefix our dep versions with `~` instead of `^`, but I'm still not convinced that's wrong. I don't know, I could go either way at this point.


aweiksnar:
:shipit:

PR #2061, opened 2015-12-28
Fix mobile nav

### LINKED ISSUE #2057: Main navigation doesn't work on iOS
Trying to use the site on my phone, I can click on Projects, About, Talk etc in the popup menu, but the site stays on the home page.

brian-c:
Closes #2057.

We could look into making this a more elegant/general thing, but then it'd probably all change in the next version of react-router anyway.

### DIFF
--- a/app/partials/main-header.cjsx
+++ b/app/partials/main-header.cjsx
@@ -1,6 +1,6 @@
 counterpart = require 'counterpart'
 React = require 'react'
-{IndexLink, Link} = require 'react-router'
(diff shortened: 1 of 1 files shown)


eatyourgreens:
:+1: Sorry it's taken me a while to get to this. Just catching up after christmas.

eatyourgreens:
:+1: Sorry it's taken me a while to get to this. Just catching up after christmas.

PR #2062, opened 2015-12-28
Session storage fallback

### LINKED ISSUE #2058: Zooniverse.org doesn't display content on Safari in Private browsing
Issue:
* Accessing zooniverse.org in Safari in Private browsing on OSX causes the content to fail to load.
* Header and footer show, but main content is blank.
* User can Sign In normally, after which the website works more-or-less normally.

Analysis:
* When attempting to load zooniverse.org in Safari in Private browsing (and not logged in), the following error is logged on the console: `QuotaExceedError: DOM Exception 22: An attempt was made to add something to storage that exceeded the quota`
* Ostensibly, this is because Safari in Private browsing lacks access to the `localStorage` feature, even though it's available in the global namespace. More details: http://stackoverflow.com/questions/14555347/html5-localstorage-error-with-safari-quota-exceeded-err-dom-exception-22-an

Notes:
* Tested on Safari 9 on OSX

brian-c:
Closes #2058, and fixes session ID lookup (looks like classifications' metadata.session was never actually set to anything).

It'd also be super cool to remove the 90KB `crypto` module too, so let me know if `Math.random()` is good enough.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -322,7 +322,7 @@ Classifier = React.createClass
   completeClassification: ->
     @props.classification.update
       completed: true
-      'metadata.session': getSessionID().id
+      'metadata.session': getSessionID()
       'metadata.finished_at': (new Date).toISOString()
       'metadata.viewport':
         width: innerWidth
--- a/app/lib/session.coffee
+++ b/app/lib/session.coffee
@@ -1,4 +1,4 @@
-stored = sessionStorage.getItem('session_id')
+stored = sessionStorage?.getItem('session_id')
 
 generateSessionID = () ->
   sha2 = require('crypto').createHash('sha256')
--- a/app/lib/session.coffee
+++ b/app/lib/session.coffee
@@ -1,18 +1,22 @@
+stored = sessionStorage.getItem('session_id')
+
 generateSessionID = () ->
   sha2 = require('crypto').createHash('sha256')
   id = sha2.update("#{Math.random() * 10000 }#{Date.now()}#{Math.random() * 1000}").digest('hex')
   ttl = tenMinutesFromNow()
   stored = {id, ttl}
-  sessionStorage.setItem('session_id', JSON.stringify(stored))
+  try
+    sessionStorage.setItem('session_id', JSON.stringify(stored))
   stored
 
 getSessionID = () ->
-  {id, ttl} = JSON.parse(sessionStorage?.getItem('session_id'))
+  {id, ttl} = JSON.parse(sessionStorage.getItem('session_id')) ? stored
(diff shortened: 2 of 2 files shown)


camallen:
LGTM - see my comments in #2058 about requiring a hash function of some sort.

shaunanoordin:
I'm getting an error trying to compile and run the `session-storage-fallback` branch.

Error: 
* `ReferenceError: sessionStorage is not defined`
* at `app/lib/session.coffee`, line 4: `stored = sessionStorage.getItem('session_id')` (added in commit 292d589)
* upon running `npm start`; prevents website from running.

Action:
* I'm currently trying to see why my setup chokes on this; `sessionStorage` should theoretically be a global object. (Actually, if a lack of a reference to sessionStorage was the problem, shouldn't this fail on the client-side browsing, not during server-side compile?)

brian-c:
Good catch @shaunanoordin, fixed now. I think the app initially runs once in Node in dev to get routing working.

shaunanoordin:
Thanks @brian-c ! I'll re-test now.

shaunanoordin:
Good news: We can now access PFE while in Safari Private mode.
(Minor?) Bad news: We cannot sign in to our Zooniverse accounts while in Safari Private mode.
Still good news: PFE signing in is pretty much confirmed functional for every other major desktop browser (Private mode or otherwise). There are no issues when Safari is not in Private mode.

Error:
* Signing In on Safari Private mode will fail, showing the following error message on the Sign In dialog box:
* `Error: invalid_grant The provided authorization grant is invalid, expired, revoked, does not match the redirection URI used in the the authorization request, or was issued to another client.`

Actions:
* I'm trying to trace this error to see where it goes.
* (I'm aware that the Stargazing Live event is coming in 2 weeks however, so I won't dig too deep into this edge case if it introduces any questions to Sign In stability riiight before Stargazing.)

Tested on:
* FF43 / Edge25 / IE11 + Win10 (Normal & Private browsing)
* Chrome47 / FF43 / Safari 9 / Opera 34 + OSX (Normal & Private browsing)
* Attempted browsing PFE, then signing in, then browsing PFE while logged in.

(Yes, IE11 works fine.)

shaunanoordin:
EDIT: The latest `invalid_grant` error may be complete unrelated to the SessionID in Private Safari issue (which seems resolved) If so, this PR seems safe to merge while I spin off another investigation into the `invalid_grant` error.

shaunanoordin:
Yup, `invalid_grant` seems to be a separate issue which may be only relevant to Safari Private mode on localhost. I'm merging this PR now. I'll revisit this later to address the hash function brought up by @camallen . Thanks all!

PR #2063, opened 2015-12-28
Field guide

brian-c:
Some changes from @jwbmartin's design:

- It's on the right, so it doesn't cover the subject during classification

- It's got a tab instead of Chimp&See's field guide icon, that way it can live on any of the project pages (and people can reference it while in Talk, etc.)

Preview at http://preview.zooniverse.org/panoptes-front-end/field-guide (under brian-testing/ghosts).

Editor coming real soon.


amyrebecca:
I reviewed this pull request and see no problems with it, but I'm not sure whether I should merge during stargazing. If I shouldn't, then I'll come back to this as soon as that's over and hit the big green button.

PR #2064, opened 2015-12-28
Add selection task

aweiksnar:
- Adds a selection task: a task that contains 1 or more select menus with options
  - Code & data shape should be similiar to other tasks/editors, but let me know if I missed anything key here
- Adds an Editor for adding select boxes and options 
  - mainly just inputs for adding/deleting key-value pairs
  - also drag&drop to change up the order of the selects (`drag-reorderable` is slick :banana: btw)
- Useful for Notes From Nature panoptes port, also other projects with 2-d lists of items to choose from
- Should be quick to setup a project on dev in the lab, but also viewable on `/projects/alex-panoptes/my-project/classify`

### DIFF
--- a/app/classifier/tasks/selection/index.cjsx
+++ b/app/classifier/tasks/selection/index.cjsx
@@ -61,14 +61,17 @@ module?.exports = React.createClass
       <div className="selection-task">
 
         {selectBoxes.map (name, i) =>
-          <select key={i} defaultValue="" ref="select-#{name}" onChange={@onChangeSelect.bind(@, selectBoxes)}>
-            <option key="_title" value="" disabled>{answers[name].title}</option>
+          <div>
+            <div>{answers[name].title}</div>
+            <select key={i} defaultValue={@props.annotation.value[name] ? ""} ref="select-#{name}" onChange={@onChangeSelect.bind(@, selectBoxes)}>
+              <option key="_title" value="" disabled>--</option>
 
-            {answers[name].values.map (option, i) =>
-              <option key={i} value={option.value}>
-                {option.label}
-              </option>}
-          </select>
+              {answers[name].values.map (option, i) =>
+                <option key={i} value={option.value}>
+                  {option.label}
+                </option>}
+            </select>
+          </div>
           }
 
       </div>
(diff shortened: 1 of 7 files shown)


app/classifier/tasks/selection/index.cjsx
```diff
@@ -0,0 +1,84 @@
...
+
+  render: ->
+    {answers, answersOrder} = @props.task
+    selectBoxes = if answersOrder.length then answersOrder else Object.keys(answers)
+
+    <GenericTask question={@props.task.instruction} help={@props.task.help} required={@props.task.required}>
+      <div className="selection-task">
+
+        {selectBoxes.map (name, i) =>
+          <select key={i} defaultValue="" ref="select-#{name}" onChange={@onChangeSelect.bind(@, selectBoxes)}>
```
brian-c: `value` here should come from the annotation. As written, going back to a selection task after completing it will not display the annotation's value.

app/classifier/tasks/selection/index.cjsx
```diff
@@ -0,0 +1,84 @@
...
+            <option key="_title" value="" disabled>{answers[name].title}</option>
```
brian-c: Could we take the label out of the menu so it's not lost once a selection is made? Maybe set his up as a table.

app/pages/lab/workflow.cjsx
```diff
@@ -159,6 +160,15 @@ EditWorkflowPage = React.createClass
                         <small><strong>Crop</strong></small>
                       </button>
                     </AutoSave>}{' '}
+
+                  {if @canUseTask(@props.project, "selection")
```
brian-c: Mind adding "selection" to the list of experimental features in pages/admin/project-status?

aweiksnar: Yep - added in 55a66f659126b5b6a76ce58c70ff73a93e4ac186

app/classifier/tasks/selection/index.cjsx
aweiksnar: Ah good call I was only testing with a single task. Added in 55a66f659126b5b6a76ce58c70ff73a93e4ac186 and also testable in `/projects/alex-panoptes/bobbys-from-nature/classify`

brian-c:
I'm having a hard time using the editor, it works quite a bit differently than the others (e.g. user-defined keys, "Save workflow"), and I keep getting it into states where nothing works anymore (`Cannot read property 'map' of undefined`). Seems like there's enough there to get out a demo though. If we're in a big hurry for this I can merge and we can iterate on it later since it's behind an admin flag. That sound okay @trouille?

aweiksnar:
There's no rush to ship if you're not happy. Let's chat more about the editor tomorrow morning and then I'll push up some more changes

Issue #1936, opened 2015-11-24
user search & debounced calls on first change event


camallen:
When searching for users the debounced function fires on first invocation with small input and doesn't wait until the user has finished typing to submit the full search param. I know it may not ever be a full search as it's dependent on user input but firing multiple times with small inputs is putting extra load on the api nodes (something i've got to try and fix as well).

Looks like the debounced function will run on first invocation and then after every X wait period too. Not sure if react-select handles change events with a delay either (e.g. [here](https://github.com/JedWatson/react-select/issues/614)). Is there a way to ensure just the last event fires after finishing typing?

<img width="901" alt="screen shot 2015-11-24 at 10 27 49" src="https://cloud.githubusercontent.com/assets/295329/11364226/f244732c-9296-11e5-9584-f19f29bccf3b.png">

camallen:
closing as i think this was revoled in #2071

PR #2069, opened 2015-12-31
fix busted user search by removing debounce function

camallen:
Using debounce with react-select gives me a component that is stuck in search
screen shot 2015-12-31 at 15 49 18

This is not ideal and just removes the debounce function since it breaks the react-select async component similar to what’s described here, https://github.com/JedWatson/react-select/issues/614#issuecomment-159513545. I think the debouncer just returns a busted promise or none-at all unless it meets the criteria to run. 

I'd be supper happy to see this one closed for a working throttle / debouncer that feeds nicely to react-select in the context of #1936.

### DIFF
--- a/app/components/user-search.cjsx
+++ b/app/components/user-search.cjsx
@@ -27,4 +27,5 @@ module.exports = React.createClass
       searchPromptText="Type to search Users"
       className="search standard-input"
       closeAfterClick={true}
-      asyncOptions={debounce(@searchUsers, 200)} />
+      matchProp={'label'}
+      asyncOptions={@searchUsers} />


brian-c:
Wow, yeah, I just messed with this for an hour thinking it'd be easy to fix but react-select is really fighting me on it.

camallen:
Right..I was thinking that the promise never fires so the callback for react-select never runs. Could we add a manual debounce into the `@searchUsers` function and just return an empty options array if we don't fire the promise? I think that should fulfil the expected interface for react-select.

@brian-c was this the sort of thing you were trying?

PR #2071, opened 2015-12-31
Manually debounce user search

brian-c:
As suggested in #2069, here's the user search manually debounced.

I was trying to store the options in the UserSearch component where they belong and pass them as `options` to the Select component (I still think this is the proper way to do it, I'm not sure why react-select wants to handle async data itself). This worked fine, except the Select lost whatever text was typed into it whenever the parent updated. To work around that I tried storing that text and passing it as a `value` into the Select's `inputProps` (again, mostly correct, in my opinion), but the Select wouldn't know to clear its placeholder. Sooo: yeah, this works!

### DIFF
--- a/app/components/user-search.cjsx
+++ b/app/components/user-search.cjsx
@@ -1,23 +1,34 @@
 React = require 'react'
 Select = require 'react-select'
 apiClient = require '../api/client'
-debounce = require 'debounce'
+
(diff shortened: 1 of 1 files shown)


camallen:
this works well on my testing, especially against this PR https://github.com/zooniverse/Panoptes/pull/1565

PR #2072, opened 2016-01-04
Workflow visualization take 3

CKrawczyk:
I have fixed the case sensitive file name issue and added support for the 'flexibleSurvey' task type.  This should be good to go, although we might want to do a test deploy to preview just to make sure it won't break anything else during the deploy process.


brian-c:
This works for me. @astopy what's the best way to make sure this deploy doesn't fail? And `set -e` in postinstall.sh should prevent this from breaking the site if it does fail, right?

astopy:
Yes, adding `set -e` should do it. I'd recommend deploying to staging using docker, since that's how we're deploying to production. This should be all you need to run (assuming you already have Docker installed):

`docker build -t pfe . && docker run -ti --rm --env-file=<(env) pfe npm run-script stage`

CKrawczyk:
@astopy just to clarify, is that docker command something I can run on my end (i.e. a local build) or something you or @brian-c should run (i.e. needs AWS access keys)?  Also `set -e` is already in the postinstall.sh script.

astopy:
@CKrawczyk Ah, yes you won't be able to publish to preview yourself, but you can run this instead to make sure it at least builds:

`docker build -t pfe . && docker run -ti --rm pfe npm run-script build`

CKrawczyk:
Local docker build had no errors :smiley:

CKrawczyk:
Just rebased, @brian-c @astopy this is ready to merge.

brian-c:
Looks like the script tag needs a dot in NON_ROOT mode.

`<script src="<% if (process.env.NON_ROOT === "true") { %>.<% } %>/jsPlumb.min.js"></script>`

CKrawczyk:
Script tag updated.

brian-c:
Workin' fine now, thanks!

PR #2073, opened 2016-01-04
Field guide editor

brian-c:
This adds an editor for project field guides.

I'm pretty pleased with how things are organized here. Pretty much anything the user can do that interacts with the back end is defined in an `actions` module, and the interface just calls those actions.

Any questions, just ask!


amyrebecca:
I reviewed this pull request and see no problems with it, but I'm not sure whether I should merge during stargazing. If I shouldn't, then I'll come back to this as soon as that's over and hit the big green button.

brian-c:
Yeah, let's wait until next week before we merge anything non-Stargazing related. Thanks.

PR #2076, opened 2016-01-05
Bump page size for checking lab permissions

brian-c:


camallen:
Solid - i think we can filter by type for just collab / owner roles too. I'll check on that.

PR #2083, opened 2016-01-07
More optimizations

parrish:
Mainly cutting down on redundant API calls on `/`, `/about/publications`, and `/collections`.

`/collections` could probably use some server-side improvements.

`/projects` now uses the `cards=true` param for consistency.

@camallen If we merge this, then we can just settle on that serializer and remove the other.

### DIFF
--- a/app/pages/about/publications-page.cjsx
+++ b/app/pages/about/publications-page.cjsx
@@ -2,7 +2,7 @@ counterpart = require 'counterpart'
 React = require 'react'
 Translate = require 'react-translate-component'
 apiClient = require '../../api/client'
-PromiseRenderer = require '../../components/promise-renderer'
+Loading = require '../../components/loading-indicator'
 
 counterpart.registerTranslations 'en',
   publications:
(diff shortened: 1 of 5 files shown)


app/pages/about/publications-page.cjsx
```diff
@@ -293,41 +306,43 @@ module.exports = React.createClass
...
     @setState currentSort: navItem
 
-  getAvatar: (project) ->
-    <PromiseRenderer promise={apiClient.type('projects').get(slug: project.slug).get('avatar')} pending={null}
-      catch={=> <img src="/assets/simple-pattern.jpg" />}
-      then={([avatar]) =>
-        <img src={avatar.src} alt="Project Avatar" />} />
+  avatarFor: (project) ->
+    src = if project?.avatar_src
+      "//#{ project.avatar_src }"
```
camallen: Should the API return the `//` url prefix?

app/pages/home.cjsx
```diff
@@ -42,17 +42,27 @@ counterpart.registerTranslations 'en',
...
+    src = if project.avatar_src
+      "//#{ project.avatar_src }"
+    else
+      '/assets/simple-avatar.jpg'
+    Promise.resolve src
+
   render: ->
     <div className="featured-projects">
-      <PromiseRenderer promise={apiClient.type('projects').get(FEATURED_PRODUCT_IDS)}>{(projects) =>
+      <PromiseRenderer promise={apiClient.type('projects').get(id: FEATURED_PRODUCT_IDS, cards: true)}>{(projects) =>
```
camallen: will this use the index route? If it's a show route, i'll have to make the cards context expandable to index / show.

app/pages/about/publications-page.cjsx
parrish: Either way. It's probably better if it returned a URL with a protocol though

app/pages/home.cjsx
parrish: Yeah, it uses the index action with an id filter.

parrish:
@brian-c I can't really find any more easy optimizations.  This should be good to check out again.  Thanks!

brian-c:
No avatars are showing up on the Publications page. Looks good otherwise.

PR #2084, opened 2016-01-07
Flexible survey

amyrebecca:
These are a set of small behavior and style changes to support the flexible survey tool requested by Wisconsin Wildlife Watch. The effects of these changes should be contained to the flexible survey tool, an experimental task that lets different "choices" in the first step of a survey result in different "questions" in the second step of a survey.

1. Users of the flexible survey tool must make at least one choice before submitting a classification
2. Users must answer at least one question in order to complete an identification, even if none of the individual questions are marked as required
3. The choices should be laid out by column, not by row.

### DIFF
--- a/app/classifier/tasks/flexible-survey/choice.cjsx
+++ b/app/classifier/tasks/flexible-survey/choice.cjsx
@@ -64,6 +64,20 @@ module.exports = React.createClass
           return false
     true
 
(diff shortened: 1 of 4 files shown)


amyrebecca:
I just assigned this to Mark because it's pretty quick and we didn't want to bother anyone else.

PR #2087, opened 2016-01-08
only include launch approved projects by default

camallen:
and always use the cards route for pages, even when admin.


parrish:
Good catch!

PR #2089, opened 2016-01-09
avoid the api project roles owner calls

camallen:
Cuts down on unnecessary api calls, especially when there are large numbers of collabs on a project. Instead just compare the logged in user id to the project_roles linked owner id to find map the roles they have.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -253,14 +253,9 @@ Classifier = React.createClass
     if @props.project?
       getUserRoles = @props.project.get 'project_roles'
         .then (projectRoles) =>
-          getProjectRoleHavers = Promise.all projectRoles.map (projectRole) =>
-            projectRole.get 'owner'
-          getProjectRoleHavers
-            .then (projectRoleHavers) =>
-              (projectRoles[i].roles for user, i in projectRoleHavers when user is @props.user)
-            .then (setsOfUserRoles) =>
-              [[], setsOfUserRoles...].reduce (set, next) =>
-                set.concat next
+          projectRoles.map (projectRole) =>
+            if @props.user.id == projectRole.links.owner.id
+              projectRole.roles
 
       <PromiseRenderer promise={getUserRoles}>{(userRoles) =>
         if isAdmin() or 'owner' in userRoles or 'collaborator' in userRoles or 'expert' in userRoles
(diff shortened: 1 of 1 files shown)


camallen:
Also this should probably have a large page size until we add specific role filtering on this end point. E.g. pulsar-hunters issue grant had the other day.

camallen:
Actually this should probably use the filter by user_id project roles end to point too, I'll update.

app/classifier/index.cjsx
```diff
@@ -250,17 +250,11 @@ Classifier = React.createClass
...
-          getProjectRoleHavers = Promise.all projectRoles.map (projectRole) =>
-            projectRole.get 'owner'
-          getProjectRoleHavers
-            .then (projectRoleHavers) =>
-              (projectRoles[i].roles for user, i in projectRoleHavers when user is @props.user)
-            .then (setsOfUserRoles) =>
-              [[], setsOfUserRoles...].reduce (set, next) =>
-                set.concat next
+          projectRoles.map (projectRole) =>
+            projectRole.roles
```
brian-c: The mapped array needs to be reduced/flattened instead, or this resolves to an array-of-arrays of roles.

app/classifier/index.cjsx
```diff
@@ -250,17 +250,11 @@ Classifier = React.createClass
     </div>
 
   renderExpertOptions: ->
-    if @props.project?
-      getUserRoles = @props.project.get 'project_roles'
+    if @props.project? && @props.user?
+      getUserRoles = @props.project.get('project_roles', user_id: @props.user.id)
```
brian-c: Adding a query param here means the JSON-API client no longer caches the value locally, so it'll re-request every time the component renders (which is very often). We'll need to cache this value in a property of the component, and clear it when the project or user changes.

camallen: that explains why i was seeing strange API request caching behaviour. Why won't it cache the API call results if it's the same API url?

brian-c: Bad design on my part. I expected most resources to be fetched as links of other resources, so I conflated making specific queries with wanting fresh data. Clearly not the case.

Ideally when we encounter problems like this we'll use it as an opportunity to stop using a `PromiseRenderer`.

camallen:
closing in favour of #2096

PR #2091, opened 2016-01-10
Lab user roles testing

camallen:
Be more specific about which user's roles we are testing where we can. Also added a larger page size to the collaborators page for projects with large numbers of users with roles. This should probably just have paging put on it.


marten:
LGTM

camallen:
This was a pretty old PR i had open. It should avoid the problem when lots of roles exist on a project and some users aren't in the default first page. Also will target just the owner role via the user_id param.

PR #2094, opened 2016-01-11
proper sort params for the api

camallen:
use the sort indexes on the api, restpack supports - for desc and nothing for asc sort orders. https://github.com/RestPack/restpack_serializer#sorting

### DIFF
--- a/app/pages/home.cjsx
+++ b/app/pages/home.cjsx
@@ -79,7 +79,7 @@ module.exports = React.createClass
     document.documentElement.classList.remove 'on-home-page'
 
   lastFourProjects: ->
-    @props.user.get("project_preferences", page_size: 4, sort: '+updated_at')
+    @props.user.get("project_preferences", page_size: 4, sort: '-updated_at')
 
   render: ->
     aboutItems = ['contribute', 'explore', 'collaborate', 'discover']
--- a/app/pages/admin/project-status-list.cjsx
+++ b/app/pages/admin/project-status-list.cjsx
@@ -13,7 +13,7 @@ module.exports = React.createClass
   getProjects: ->
     query =
       page_size: 24
-      sort: '+updated_at'
+      sort: '-updated_at'
       include: 'avatar'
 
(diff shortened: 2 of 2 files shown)


app/pages/home.cjsx
```diff
@@ -79,7 +79,7 @@ module.exports = React.createClass
     document.documentElement.classList.remove 'on-home-page'
 
   lastFourProjects: ->
-    @props.user.get("project_preferences", page_size: 4, sort: '+updated_at')
+    @props.user.get("project_preferences", page_size: 4, sort: 'updated_at')
```
brian-c: I'm not sure that's right. To me it looks like this (and `-updated_at`) returns four oldest. `+updated_at` returns the four most recent.

camallen: I just tested on my dev api node to see if the sort order was added using +, its not due to this [line](https://github.com/RestPack/restpack_serializer/blob/master/lib/restpack_serializer/options.rb#L65) leaving a space in the + param spot. Basically as the docs say, use a `-` to activate desc order otherwise if it's `sort=field` it'll be asc

camallen: Also that is just a by product of the db default result order, not an explicit order by clause on the query.

parrish: Pretty sure the confusion here is the difference in the JSON API spec and the implementation.  (`+fieldname` is the explicit form of ascending, `fieldname` is the implicit ascending)

camallen: Yeah - I was surprised the index wasn't being used and I investigated the query, it didn't have an order by at all from the sort param scope being set.

brian-c: Gotcha, we want newest first, so should these be prefixed with `-`?

camallen: Indeed - i should look at the results that came back...not just the sql scope! Updated.

brian-c:
Solid, thanks.

Issue #2090, opened 2016-01-09
Project classify page calling the same resources multiple times


camallen:
Project classify page calls workflows twice, once for active and other for not. Is this required?

<img width="730" alt="screen shot 2016-01-09 at 19 29 55" src="https://cloud.githubusercontent.com/assets/295329/12217945/e0494586-b707-11e5-9127-906668979c8e.png">
Note the image has the preflight OPTIONS and the GET so it's listed as 4 times.

camallen:
So i've done a bit more debugging and it looks like the workflow calls above are due to the way we fetch a specified workflow (if a query param, ignoring active flags) OR we fetch a random active one. We could just do one of the other in this case, not both as it currently does. 

Also i've found the classify page is behaving somewhat strangely that is causing the same API request to be fired multiple times. It seems that it tries to render the summary page on classification start (i.e. no annotations on the first task), which fires the user roles and gold_standard promises. It then renders all the tasks (not just the active one) which again fires the user roles promise, this repeats for each progression through the interface. Finally on summary it renders the summary page which calls user_roles and gold_standard again. All this is resulting in multiple duplicated queries to the API that could have been stored state at the start or at least cached at the API client layer.

Thoughts?

simoneduca:
@camallen Not sure I can reproduce. Did it happen on submitting a classification (pressing 'Next')?

camallen:
PM'ing you a logging patch

simoneduca:
This is a screenshot showing Cam's description above (2nd paragraph)

![screen shot 2016-01-11 at 12 19 11](https://cloud.githubusercontent.com/assets/8989606/12233269/b82bdf94-b85d-11e5-81eb-04a556f0fc73.png)

camallen:
closed by #2096

PR #2096, opened 2016-01-11
Optimize classify

parrish:
This is an attempt to cut down on the number of API requests on the classify page.

Most of the requests will still end up being resent on each classification because of the relationship between the Classifier and ClassifierWrapper components (Classifier is unmounted/remounted between each classification).

Project roles and the requests for each user with a role on the project have been filtered and limited down to one request per page load.  Duplicate workflow requests have been cut down on initial page load.

Project preferences, the tutorial, 2 workflows requests, favorite collection, and favorite collection subjects will still run for each classification.  I'm hesitant to change that since everything except the tutorial could change between classifications.

I'm open to suggestions.

cc @camallen

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -32,7 +32,6 @@ Classifier = React.createClass
 
   getInitialState: ->
     subjectLoading: false
-    expertClassifier: null
     expertClassification: null
     classificationQuality: NaN
     showingExpertClassification: false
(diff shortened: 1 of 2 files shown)


brian-c:
Looks good, works for me. Thanks Michael!

@camallen This should take care of #2089, right? Also #2090?

camallen:
@parrish, looks good!

PR #2112, opened 2016-01-14
Improve the word "retired" on the red flash to the more meaningful ...

alexbfree:
... "finished" to address science team concerns about people wasting effort and still classifying.. Wording agreed with @ali @grant @camallen @astopy @alexbfree et al


alexbfree:
Thanks @brian-c !

PR #1684, opened 2015-10-01
Revert "Revert "Update auth client to handle response from superagent""

edpaget:
This reverts commit 38d0dc9e2bfebe70e03c8fbf3f2e8bbcc312f5d1.

PR #2113, opened 2016-01-15
Use external panoptes-client lib

brian-c:
**To-do:** update with new version after zooniverse/panoptes-javascript-client#20 is merged.

~~This keeps the old files as proxies to the new lib. That's probably something we should clean up as we go along instead of replacing a whole bunch of `require`s all at once.~~ Never mind, they're all replaced now. Sorry for the big diff.

Closes #1684 (no longer relevant).


srallen:
I'm seeing `Uncaught (in promise) TypeError: Cannot read property 'slug' of undefined` on the publications page and the publications list fails to load.

brian-c:
Good catch @srallen, fixed in zooniverse/json-api-client#25 and published. Thanks!

rogerhutchings:
- (Dev classifier)[http://preview.zooniverse.org/panoptes-front-end/external-panoptes-client/dev/classifier] throws `Uncaught (in promise) TypeError: Cannot read property 'id' of null`

- When moving from 'All' to 'My Collections' in the Collections page, I get `Uncaught (in promise) TypeError: Cannot read property 'locations' of undefined` (I don't have anything in my collections, mind, so it could be a result of that)

- Trying out a project: going to iEye throws `Uncaught (in promise) TypeError: Cannot read property 'get' of undefined`

eatyourgreens:
`Uncaught (in promise) TypeError: Cannot read property 'get' of undefined` is probably failing to get a field guide for a project. I'm seeing that error on zooniverse.org projects as well. I think Mammoth Bats is the only project that has a field guide resource at the moment.

brian-c:
@rogerhutchings I think these are all present in the current version, so outside the scope of this PR. Looks like Jim's right about the field guide issue, and the collections thing appears to be due to a collection not having anything it. And I think the mock data is missing a reference for the dev classifier.

I've added these to my quick-fixes list.

rogerhutchings:
Grand, then LGTM.

PR #2137, opened 2016-01-23
Combo task

brian-c:
@mcbouslog, @simensta: Here's the start of a combo task.

Demo at http://preview.zooniverse.org/panoptes-front-end/combo-task/ (and there's a link to the dev classifier in the footer under Projects, direct links still don't work on preview).

Some refactoring was needed to support this: Task components and drawing tools no longer modify their annotations directly, but call an `onChange` function with the updated annotation. That means those components no longer have to know anything about what data they're modifying (annotation vs drawing subtask vs combo-task task), they just fire the onChange function and their parent does what it needs to do.

In general, most React components should know as little about the data they're modifying as possible. The responsibility of directly modifying resources should live in one top-level component. In this case I'm trying to move modification of the classification (and its annotations) out of the tasks and drawing tools and into the Classifier as much as possible. That's what'll allow us to use the same task components in different contexts (normal workflow, drawing subtask, and now combo-task).

There's still a bit left to implement: The editor, the summary view, the `isAnnotationComplete` and `testAnnotationQuality` functions. ~~Tasks that hook into the SubjectAnnotator (drawing, crop, survey) currently can't be used in a combo because the SubjectAnnotator only looks at the hooks for the current task type. It should be possible to set up the combo task to proxy the hooks of its subtasks.~~ Drawing is now the only task not supported.

@aliburchard, @chrislintott: This is something we thought we'd try as an alternative to having a text-input task and a new multiple-text-inputs task. A side effect of making this generic is that we'll be able to use this to show a survey task and a checkbox task at the same time, which covers the Serengeti-style "Fire"/"Nothing" use case. We can talk about this in Oxford next week, but I think this can replace the out-of-workflow-task feature we were planning (which has been sitting half finished in the workflow-task branch).


eatyourgreens:
Pinging @itsravenous, since he's been working on the classifier recently.

Do we need to merge #1965 first before refactoring the classifier?

itsravenous:
As this contains only minimal changes to `SubjectAnnotator` and no changes to `SubjectViewer`, we should be OK on this one.

brian-c:
@mcbouslog I think there's enough here to review and merge. A basic editor is in place. I think the only thing left is a summary, which will require some major refactoring further up the chain, and frankly I'm not sure that feature is sticking around for long anyway. This should at least be enough to get started with the plan for NfN.

@aliburchard I've added a feature that allows the combo task to be considered complete if **at least one** of its subtasks is complete. This'll allow a combo of a survey and a checkbox to behave like Serengeti, where either an animal must be selected or a checkbox must be checked. I'm not entirely sure I like this solution, but we can play with it while it's behind the admin flag.

mcbouslog:
per our discussion on Slack, here are a few notes for follow-up
- allowing a next task be set to follow a combo task
- low priority design note - in the editor, maybe change the dots in the combo task child tasks list (on the right after adding child tasks) to the icon with horizontal lines that (to me anyway) implies things are moveable/drag-able?

brian-c:
Thanks @mcbouslog. I've added the next-task selector. I also fixed the cursor of the draggable items so it's more obvious they're draggable (and I'm going to add this to the drag-reorderable library in a minute). I wanna give some more thought to a proper visual indication and implement it when this is more polished.

PR #2141, opened 2016-01-25
text tool option for subTask

simensta:
This PR adds the text tool as an option when creating subTasks in the workflow editor. This option is only available for projects with the experimental text tool enabled.

@brian-c, @mcbouslog: thoughts?

### DIFF
--- a/app/classifier/tasks/drawing-task-details-editor.cjsx
+++ b/app/classifier/tasks/drawing-task-details-editor.cjsx
@@ -4,6 +4,10 @@ AutoSave = require '../../components/auto-save'
 module.exports = React.createClass
   displayName: 'DrawingTaskDetailsEditor'
 
+  propTypes:
+    project: React.PropTypes.object.isRequired
+    task: React.PropTypes.object.isRequired
+
   getDefaultProps: ->
     workflow: null
     task: null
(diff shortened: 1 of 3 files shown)


brian-c:
Looks nice, thanks! I'll probably be able to do a proper review and merge tomorrow.

app/classifier/tasks/drawing-task-details-editor.cjsx
```diff
@@ -4,6 +4,10 @@ AutoSave = require '../../components/auto-save'
 module.exports = React.createClass
   displayName: 'DrawingTaskDetailsEditor'
 
+  propTypes:
```
brian-c: This is generally unnecessary in a component with such a specific instance (very handy in more reusable ones, though). If you do define `propTypes`, I'd define all of them (this is missing `toolIndex`).

brian-c:
Looks good. A couple UI things to maybe clean up:

- In the GenericEditor, can we only display the "Choices" label `if choicesKey?`? This isn't something you added, but until now it hasn't mattered.

- Maybe wrap the "Question" and "Text" buttons in a TriggeredModalForm, like in `lab/workflow.cjsx`? They're a bit unwieldy here. Or maybe just lose the icons. Also, there's no verb associated, so it's not immediately clear what's going to happen when you click them.

app/classifier/tasks/drawing-task-details-editor.cjsx
brian-c: Also (sorry!), I've never found `React.PropTypes.object` to ever be super meaningful. Maybe check out `React.PropTypes.shape` to specify the required property types.

simensta:
@brian-c: thanks! thoughts on the following changes?

GenericEditor:
* displays the "Choices" label conditionally

DrawingTaskDetailsEditor:
* removed propTypes
* the "Question" and "Text" buttons are wrapped in a TriggeredModalForm
subtask_tools

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -47,10 +47,12 @@ module.exports = React.createClass
           <small className="form-help">Add text and images for a window that pops up when volunteers click “Need some help?” You can use markdown to format this text and add images. The help text can be as long as you need, but you should try to keep it simple and avoid jargon.</small>
         </div>}
 
-      <hr />
+      {if choicesKey?
+        <hr />
 
-      <span className="form-label">Choices</span>
-      {' '}
+        <span className="form-label">Choices</span>
```
brian-c: Little JSX gotcha: an CoffeeScript `if` expression like:

```jsx
{if true
  <OneThing />
  <Another />}
```

can only resolve to a single value, so this becomes `<Another />`, where `<OneThing />` is skipped.

This particular file works around the problem with the convention of resolving to an array, (and each item has an arbitrary `key` property), but that's probably a little convoluted. I'd just wrap these in a container `<div>` so they'll both display.

brian-c:
Looks good, other than that skipped `<hr />`.

I noticed you've removed the bottom "Close" button. Was that intentional? I think it's fine, since there's a close box (and I'm generally in favor of giving the user one right way to do things).

app/classifier/tasks/generic-editor.cjsx
simensta: good to know!

simensta:
Fixed the skipped `<hr />` by wrapping those items in  a `<div />`. I did intentionally remove the "Close button" because we already provide a close box.

brian-c:
Great. Thanks @simensta!

PR #2145, opened 2016-01-27
Tasks now pass up updated annotation through onChange

brian-c:
(Changes from #2137) Tasks are now responsible for passing updated annotations up to their parent, instead of updating the annotation in place and telling the classifier that it changed.

This allows task components to be used in contexts other than as direct children of the classifier.

cc: @itsravenous


eatyourgreens:
This looks good to me but I think #1965 has created some conflicts in the classifier and annotator components. Can you rebase to bring in that PR?

brian-c:
I believe this is good to review @eatyourgreens. I've also updated the dev classifier with (I think) all the new features.

eatyourgreens:
:+1: looks good to me.

PR #2147, opened 2016-01-27
Min and max drawing tool marks

brian-c:
Allow setting a min and max number of marks per drawing tool.

Involved some refactoring of answer buttons and their css to show the user the right status.

screen shot 2016-01-27 at 4 26 39 pm

Cc: @aliburchard, this'll be available when merged, no admin flag or anything. Existing projects won't be affected.


aliburchard:
yay! thanks.

Issue #2140, opened 2016-01-25
field guide - create field guide doesn't work right away


aliburchard:
Have to reload the page before the fields visible to edit.

trouille:
Semi-related to this issue - The field guide remains even after you remove it (e.g., for Mammoth Bats, https://www.zooniverse.org/lab/378/, which Brian you're a collaborator on), I decided best to remove the field guide functionality (by un-clicking the choice on the admin project-status page), and so you don't see the field guide as a tab option anymore in the Project Editor, but the field guide icon still shows up on the project site itself - https://www.zooniverse.org/projects/zooniverse/mammoth-bats/

brian-c:
@trouille I just added your issue to my list-o'-quick-things, will close this when done. Thanks.

PR #2148, opened 2016-01-27
Load field guide after creation

brian-c:
Fixes a bug where creating a project's field guide wouldn't actually load it until the page was reloaded. Also adds a "loading" state and cleans up some other visual bits.

For #2140

### DIFF
--- a/app/pages/lab/field-guide/actions.js
+++ b/app/pages/lab/field-guide/actions.js
@@ -5,8 +5,8 @@ var projects = apiClient.type('projects');
 var guides = apiClient.type('field_guides');
 
 var DEFAULT_ITEM = {
-  title: 'Untitled',
-  content: 'Here’s everything you need to know about the great **Untitled**...'
+  title: 'An example',
+  content: 'Here’s everything you need to know about the great **Example**...'
 };
 
 var actions = {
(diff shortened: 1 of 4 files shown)


simensta:
As soon as I enable the field guide feature for a project, I am able to create and edit entries. However, I came across the following issues:
* the field guide is rendered on the project page as soon as the feature is enabled, even if there are no entries. It seems to me that opting for access to an experimental feature for a project shouldn't automatically instantiate that feature in the project.
* when the field guide feature is disabled the field guide continues to render on the project page.

brian-c:
Sorry, totally forgot to respond to this: 

I've got a separate PR coming today to handle the first issue.

I don't think the second is worth writing a check for, since we shouldn't really be turning features on and off like that. The other admin flags are for exposing features in the editor, not the project, and this is consistent with that behavior.

aliburchard:
@brian-c is the second one (disabling a feature and not having it display) something we want to consider as a way for when project owners want to edit a feature? I suppose versioning the field guides will mostly address this. I wonder if there is ever a use case for project owners to want to "remove" a field guide but not delete it.

brian-c:
Versioning will address this. If we want to offer an on/off switch for field guides (I don't think we should), we should add it explicitly (as a property of the project configuration) so as not to confuse the intention of the admin flag.

simensta:
@brian-c: LGTM -- I'll take a look at #2191 after merging this PR. Thanks!

PR #2154, opened 2016-01-28
Fix SVG dimensions for videos

eatyourgreens:
Use `offsetHeight` and `offsetWidth` when `naturalHeight` and `naturalWidth` are undefined.

Should fix the drawing surface dimensions being set to 0 for video subjects.

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -143,10 +143,12 @@ module.exports = React.createClass
     scrubber.value = player.currentTime
 
   handleLoad: (e) ->
+    width = e.target.videoWidth ? e.target.naturalWidth
+    height = e.target.videoHeight ? e.target.naturalHeight
     @setState
       loading: false
       frameDimensions:
-        width: e.target.naturalWidth
-        height: e.target.naturalHeight
+        width: width ? 0
+        height: height ? 0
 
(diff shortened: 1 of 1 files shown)


brian-c:
`offsetWidth` and `Height` are dependent on the display of the subject image/video, which means different classifications might be made against different dimensions. If we can't determine the *actual* video dimensions, we shouldn't allow marking at all.

Haven't tried it, but it looks like maybe there's `videoWidth` and `Height`: https://html.spec.whatwg.org/multipage/embedded-content.html#concept-video-intrinsic-width

eatyourgreens:
Good point. I'll have a look at that. MDN has a bit of documentation.
https://developer.mozilla.org/en/docs/Web/API/HTMLVideoElement

Not sure why they didn't keep `naturalWidth` and `naturalHeight` for compatibility with the image element API :unamused:

brian-c:
It sounds like videos can contain multiple different video tracks (I have zero experience with this), so if I had to guess I'd say it was because naturalWidth is static but videoWidth might change depending on the current track. Weird anyway.

eatyourgreens:
Rebased to use `videoHeight` and `videoWidth`. If I've got the logic right, this should fall back to 0 when the dimensions are unknown. Relies on #2162 before it will work in Firefox.

PR #2156, opened 2016-01-28
Add optional keys to the frame viewer

eatyourgreens:
Should fix React warnings when the new side-by-side image viewer is actually used.

My bad: I completely missed this when reviewing #1965


eatyourgreens:
Rebased to bring in #2145

brian-c:
Solid, thanks.

PR #2162, opened 2016-01-29
Fix video loading in Firefox

### LINKED ISSUE #2135: Loading cover never disappears for video subjects in Firefox
E.g. on [Mammoth Bats](https://www.zooniverse.org/projects/zooniverse/mammoth-bats/classify). It seems the `canplaythrough` event isn't being fired, meaning `SubjectViewer#handleLoad` (which removes the loading cover) never gets called. Works fine in Chrome.

Update: OK, it does disappear once you've clicked play, but this is definitely has potential to confuse users.

(If it helps: Ubuntu 15.10, Firefox 43.0.4)

eatyourgreens:
Closes #2135 

Uses `canPlay` rather than `canPlayThough` and sets `preload=auto` to force Firefox to download the mp4 file.

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -58,7 +58,7 @@ module.exports = React.createClass
         </div>
       when 'video'
         <div className="subject-video-frame">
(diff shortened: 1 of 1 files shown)


eatyourgreens:
I'm not convinced about forcing the browser to download a large file, but couldn't think of another way to get this to work in Firefox. Ideally, we'd want to get rid of the loading frame without downloading the entire file.

PR #2174, opened 2016-02-02
Started discussions preference

parrish:
This allows users to subscribe to new discussions on a board.

### :warning: The API for this is only deployed on staging at the moment :warning: 

We should coordinate deploys before merging.

This is currently staged at http://preview.zooniverse.org/panoptes-front-end/started_discussions-preference/

@srallen, would you mind taking a look?

### DIFF
--- a/app/pages/notifications/notification.cjsx
+++ b/app/pages/notifications/notification.cjsx
@@ -3,6 +3,7 @@ CommentNotification = require './comment'
 DataRequestNotification = require './data-request'
 MessageNotification = require './message'
 ModerationNotification = require './moderation'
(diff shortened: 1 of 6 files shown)


brian-c:
Front end looks good, but I didn't get a notification when I started a new discussion from another account. This is on staging.

parrish:
I got the notifications.  Looks like brian-testing has the subscription and it goes to a mailinator address with an email frequency of 'never', which is the default.

brian-c:
It should still show up on /notifications though right?

brian-c:
Never mind, there it is! Sorry.

PR #2175, opened 2016-02-03
Fix undefined commentsMeta after deleting a comment

### LINKED ISSUE #2171: Delete own post not working in Talk
I tried to delete my own post, it appeared to work, but when I went back to the board, the thread was still there. In the end @marten had to use moderator controls to delete the post, which worked.

eatyourgreens:
It was possible to set commentsMeta to undefined after deleting a post, which then prevents the discussion page from re-rendering.
Closes #2171

### DIFF
--- a/app/talk/discussion.cjsx
+++ b/app/talk/discussion.cjsx
@@ -78,11 +78,19 @@ module?.exports = React.createClass
   setComments: (page = @props.location.query?.page) ->
     @commentsRequest(page)
       .then (comments) =>
-        commentsMeta = comments[0]?.getMeta() ? {}
-        @setState {comments, commentsMeta}, =>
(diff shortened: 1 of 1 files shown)


parrish:
While this does work, it should probably redirect back to the board when there aren't any comments left.

Right now, it just reloads an empty discussion and throws an error

eatyourgreens:
No error here. I get a 204 response to the DELETE request, then there's a GET request which returns an empty comments array.

But, yeah, redirecting back to the parent board would be much better if the discussion has gone.

alexbfree:
Sorry for the delay in reviewing this; I didn't get the notification as I only fixed my github notifications after this was assigned. Looks good, thanks for fixing.

PR #2181, opened 2016-02-03
External panoptes lib for real this time

brian-c:
Everything from #2113 but the envify transform is applied globally, so the panoptes-client library is transformed. Massive confusion here came from me testing the deploy with an `npm link`ed version of the lib, which got transformed (I guess that's a bug in browserify?) and the auto-deploy happening with a normal `npm install`ed copy. My bad.

I also removed coffeeify, which was only used when other libs tried to import modules from this one.


rogerhutchings:
Nice one - I'm happy to merge this, but let's leave it until you're online so we can catch any errors.

PR #2184, opened 2016-02-04
Ensure SVG stretches over original subject size when cropping. Fixes #2180

### LINKED ISSUE #2180: crop-tool shows the cropped image overlaying the original image
After selecting an area with the crop tool, only the SVGImage is updated, not the underlying image rendered in the [FrameViewer](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/frame-viewer.cjsx#L52). In other words, after cropping an image, the user sees the cropped image overlaying the original image.

itsravenous:
When I did the multi frame stuff, I migrated the SVG-stretching styles from the component to CSS, and absentmindedly switched from `top: 0; left: 0; width: 100%; height: 100%` to `top: 0; left: 0; bottom: 0; right: 0;`. I had no idea there was any difference, but you learn something new every day.

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -12,14 +12,6 @@ generateRadioRef = (name, subject, frame) =>
 module.exports = React.createClass
   displayName: 'FrameViewer'
 
-  statics:
-    overlayStyle:
-      height: '100%'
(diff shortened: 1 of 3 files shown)


eatyourgreens:
LGTM!

PR #2189, opened 2016-02-05
Hide the back button for single task workflows

brian-c:
...and swap the done and done-and-talk buttons, as requested by @aliburchard.


app/classifier/index.cjsx
```diff
@@ -162,7 +162,11 @@ Classifier = React.createClass
       <hr />
 
       <nav className="task-nav">
-        <button type="button" className="back minor-button" disabled={onFirstAnnotation} onClick={@destroyCurrentAnnotation}>Back</button>
+        {if Object.keys(@props.workflow.tasks).length > 1
+          <button type="button" className="back minor-button" disabled={onFirstAnnotation} onClick={@destroyCurrentAnnotation}>Back</button>}
```
parrish: `disabled={onFirstAnnotation}`

Isn't that always true?

parrish:
Ah, misread that.  Looks good.

PR #2191, opened 2016-02-05
Better handling of empty field guides

brian-c:
To merge after #2148.

This prevents existing field guides from showing up if they have no items and allows fully deleting a field guide with no items.

I'm not sure the user should have to think about the difference between having a field guide with no items  and not having a field guide at all. Might be worth revisiting in the future.

### DIFF
--- a/app/components/field-guide.cjsx
+++ b/app/components/field-guide.cjsx
@@ -36,7 +36,7 @@ module.exports = React.createClass
   renderItem: ({icon, title, content, items}) ->
     <div className="field-guide-item">
       <header>
-        {if icon?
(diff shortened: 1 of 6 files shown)


simensta:
The field guide won't render if I upload an icon for an entry. `@props.icons[icon]` appears to be undefined at [L58](https://github.com/zooniverse/Panoptes-Front-End/blob/more-robust-field-guide/app/components/field-guide.cjsx#L58). 

The rest of the field guide functionality looks good!

brian-c:
@simensta Should be fixed now, thanks.

simensta:
LGTM -- thanks!

PR #2194, opened 2016-02-08
Fix survey editor display of confused pairs

brian-c:
Survey editor was unable to display confused pairs.

### DIFF
--- a/app/classifier/tasks/survey/editor.cjsx
+++ b/app/classifier/tasks/survey/editor.cjsx
@@ -134,7 +134,7 @@ module.exports = React.createClass
           <strong>{choice.label}</strong>
         }>
           <Markdown content={choice.description} />
-          <p>
+          <div>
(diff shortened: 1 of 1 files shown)


amyrebecca:
Can you say more about what needs to be done here? May as well roll it into #2208 while that's delayed.

brian-c:
The original code has a `<div>` inside a `<p>`, which totally freaks out the browser (I guess because `<p>` doesn't _require_ a closing tag in HTML, so the presence of a `<div>` implicitly ends the paragraph, and then it ends up with a bunch of leftover stuff that doesn't belong to any node), so it'd just refuse to render. Feel free to fix in your PR and close this one.

eatyourgreens:
`<p>` only allows inline content in HTML, so a block level tag like `<div>`
is invalid inside a paragraph. That might be why the browser freaks out.

--
James O'Donnell
email: jim@zooniverse.org
mobile: 07811 771 352
twitter: @pekingspring

PR #2195, opened 2016-02-08
Env cleanup

brian-c:
This simplifies making references other files (mainly images) and fixes inconsistencies between local dev, staging, and production environments.

Replaces `NON_ROOT` environment variable with a `<base href="/deployed/dir/" />` element.

  - All file references are now relative to root, exactly the same as when we used hash routing for the app. The app doesn't need to know where it's deployed.

  - Router `<Link>` elements are rooted from the router, same as always.

Any `<a>` elements pointing inside the router need to be absolute from from the base (See things like `href: document.baseURI.slice(0, -1) + href` below). These should be replaced with real `<Link>`s over time.

The dev server now respects the `DEPLOY_SUBDIR` variable, making it possible to check how things will look on staging.

Also cleans up dependencies and all the weird little unused server scripts.


brian-c:
I think this is good to go. Any volunteers to review and merge?

@astopy: This'll affect deploys. I've run the `stage-with-docker` script successfully, but there might be more you'd like to test.

@zooniverse/developers: Just a heads-up that all file references should now be relative to index.html, and all in-app links should be made with the `<Link>` component (and if that's impossible for some reason, just take `document.baseURI` into consideration).

parrish:
This all seems to work fine for me, but I'll wait for somebody else to confirm before merging.

Minor issue: The SVG logo now isn't visible on /projects in firefox.

astopy:
That's fine with me. If `stage-with-docker` works that should be all we need to test.

camallen:
Is this deployed somewhere we can all test?

astopy:
http://preview.zooniverse.org/panoptes-front-end/env-cleanup/

PR #2197, opened 2016-02-10
Split out dev classifier, show current annotations

brian-c:
Dev classifier now shows data for annotation(s).


app/router.cjsx
```diff
@@ -121,7 +121,12 @@ module.exports =
 
   <Route path="todo" component={-> <div className="content-container"><i className="fa fa-cogs"></i> TODO</div>} />
   <Route path="dev/workflow-tasks-editor" component={require './components/workflow-tasks-editor'} />
-  <Route path="dev/classifier" component={require './classifier'} />
+  <Route path="dev/classifier" component={
+    if process.env.NODE_ENV is 'production'
+      React.createClass render: -> <p>Env is {process.env.NODE_ENV}</p>
```
simensta: What do you think of using `require './pages/not-found'` instead of `<p>Env is {process.env.NODE_ENV}</p>`?

brian-c:
Good call, fixed.

simensta:
LGTM -- thanks!

Issue #2130, opened 2016-01-20
Add pan and zoom feature


mcbouslog:
Add pan and zoom feature.
@brian-c @simensta

mschwamb:
#2122 partly a duplicate

simensta:
@brian-c, @eatyourgreens, @mcbouslog: 
There is now a [staging branch with Pan & Zoom](https://preview.zooniverse.org/panoptes-front-end/pan-zoom) functionality under the dev classifier. 

Operate Pan & Zoom via Keyboard:
1. `tab` to focus the subject `<img>`.
2. While the `<img>` is in focus, use keys `-` and `+` to zoom and the arrow keys to pan.

Operate Pan via Trackpad:
1. Click the multiple arrows button to engage panning. While panning is enabled, you can use the trackpad to move the subject image. Panning by trackpad is not enabled by default to avoid conflicts with drawing tasks. 

Todos:
- [ ] Conditionally display Pan & Zoom, if enabled under the Admin panel.

- [ ] Test with mouse

simensta:
@aliburchard: Pan & Zoom would enable users to access and delete misplaced marks that fall outside the viewport (as described in #2125).

simensta:
This is now an admin-enabled tool.

PR #2199, opened 2016-02-10
pan & zoom feature for images

simensta:
This PR address Issue #2130 by introducing a pan and zoom feature for images in the subject-viewer. An admin will need to opt-in to this feature on a project-by-project basis. 

**WIP:**
- [x] viewable in dev/classifier
- [x] scroll wheel to zoom in/out
- [x] keyboard event listeners available when the pan control is activated
- [x] add limits to prevent moving off the image
- [x] continuous zoom on button press

### DIFF
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -135,7 +135,7 @@ module.exports = React.createClass
           <rect ref="sizeRect" width={@props.naturalWidth} height={@props.naturalHeight} fill="rgba(0, 0, 0, 0.01)" fillOpacity="0.01" stroke="none" />
           {if type is 'image'
             <Draggable onDrag={@props.panByDrag} disabled={@props.disabled}>
-              <SVGImage src={src} width={@props.naturalWidth} height={@props.naturalHeight} />
+              <SVGImage className={"pan-active" if @props.panEnabled} src={src} width={@props.naturalWidth} height={@props.naturalHeight} />
             </Draggable>
           }
 
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -133,10 +133,11 @@ module.exports = React.createClass
           <BeforeSubject {...hookProps} />}
         <svg style=svgStyle viewBox={createdViewBox} {...svgProps}>
           <rect ref="sizeRect" width={@props.naturalWidth} height={@props.naturalHeight} fill="rgba(0, 0, 0, 0.01)" fillOpacity="0.01" stroke="none" />
-
           {if type is 'image'
-            <SVGImage src={src} width={@props.naturalWidth} height={@props.naturalHeight} />}
-
+            <Draggable onDrag={@props.panByDrag} disabled={@props.disabled}>
+              <SVGImage src={src} width={@props.naturalWidth} height={@props.naturalHeight} />
+            </Draggable>
+          }
           {if InsideSubject?
             <InsideSubject {...hookProps} />}
 
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -138,7 +138,8 @@ module.exports = React.createClass
               <SVGImage src={src} width={@props.naturalWidth} height={@props.naturalHeight} />
             </Draggable>
           }
-          {if InsideSubject?
+
+          {if InsideSubject? && !@props.panEnabled
             <InsideSubject {...hookProps} />}
 
           {for anyTaskName, {PersistInsideSubject} of tasks when PersistInsideSubject?
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -90,41 +90,12 @@ module.exports = React.createClass
   toggleWarning: ->
     @setState showWarning: not @state.showWarning
 
-  zoom: (change) ->
-    newNaturalWidth = @state.naturalWidth * change;
-    newNaturalHeight = @state.naturalHeight * change;
-    
-    newNaturalX = @state.naturalX - (newNaturalWidth - @state.naturalWidth)/2;
-    newNaturalY = @state.naturalY - (newNaturalHeight - @state.naturalHeight)/2;
-    
-    @setState
-      naturalWidth: newNaturalWidth, 
-      naturalHeight: newNaturalHeight,
-      naturalX: newNaturalX,
-      naturalY:newNaturalY
-
-  zoomReset: ->
-    @setState
-      naturalWidth: @props.naturalWidth, 
-      naturalHeight: @props.naturalHeight,
-      naturalX: 0,
-      naturalY: 0
-
-  panHorizontal:(direction) ->
-    return if this.state.naturalX == 0 || this.state.naturalY == 0   
-    @setState
-      naturalX: @state.naturalX * direction
-
-  panVertical:(direction)->
-    @setState
-      naturalY: @state.naturalY * direction
-
   render: ->    
     taskDescription = @props.workflow.tasks[@props.annotation?.task]
     TaskComponent = tasks[taskDescription?.type]
     {type, format, src} = getSubjectLocation @props.subject, @props.frame
     
-    createdViewBox = "#{@state.naturalX} #{@state.naturalY} #{@state.naturalWidth} #{@state.naturalHeight}"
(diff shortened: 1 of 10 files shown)


eatyourgreens:
Panning/zooming with the keyboard is nice and smooth (testing on Chrome). Maybe add limits to prevent moving off the image? Something like https://github.com/zooniverse/Diaries/blob/master/app/lib/zoom_surface.coffee#L62-L65

I'd have the keyboard event listeners also available when the pan control is activated, so that I can click that button and start zooming in/out with the keyboard.

Maybe have a continuous zoom in/out when the mouse is held down on the +/- buttons, like OWD and consistent with the behaviour of the +/- keys. I've found this really useful in OWD on the iPad.

People might expect the scroll wheel to zoom in/out to, so perhaps add a listener for that when zooming is active, which hooks into the same code as the +/- keys.

app/components/frame-viewer.cjsx
```diff
@@ -26,13 +26,31 @@ module.exports = React.createClass
...
     frameDimensions: {}
+    viewBoxDimensions: {
+      x: 0,
+      y: 0,
+      width: 0
+      height: 0
+    }
+    panEnabled: false
+    zooming: false
+    zoomingTimeoutId: 0
```
eatyourgreens: Should this be `zoomingTimeout`?

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+      viewBoxDimensions:
+        width: width ? 0
+        height: height ? 0
+        x: 0
+        y: 0
+
     @props.onLoad? e, @props.frame
+
+  canZoomOut: ->
+    if @state.frameDimensions.width == @state.viewBoxDimensions.width && @state.frameDimensions.height == @state.viewBoxDimensions.height
```
eatyourgreens: `return @state.frameDimensions.width == @state.viewBoxDimensions.width && @state.frameDimensions.height == @state.viewBoxDimensions.height` should work here instead of testing and returning true if true.

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+      true
+
+  continuousZoom: (change) ->
+    return if change == 0
+    @setState zooming: true
+
+    zoomNow = =>
+      # return if change == 0
```
eatyourgreens: comment left over from development?

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+      @zoom(change)
+      clearTimeout @state.zoomingTimeout
+      @setState zoomingTimeout: setTimeout(zoomNow, 200)
+    
+    zoomNow()
+
+  zoom: (change) ->
+    return if change == 0 || @state.zooming == false
```
eatyourgreens: `!@state.zooming` instead of `@state.zooming == false`

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+        height: @state.frameDimensions.height,
+        x: 0,
+        y: 0
+
+
+  toggleZoom: ->
+    @setState zooming: !@state.zooming
+
+  togglePanOn: ->
+    unless @state.panEnabled == true
```
eatyourgreens: you don't need to test Booleans for `true` or `false`

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+      @setState panEnabled: true, =>
+        if @state.panEnabled then this.refs.subjectImage.focus()
```
eatyourgreens: Won't this `if` always return `true`?

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+  
+  togglePanOff: ->
+    @setState panEnabled: false
+  
+  toggleKeyPanZoom: ->
+    @setState keyPanZoomEnabled: !@state.keyPanZoomEnabled, =>
+      if @state.panEnabled then this.refs.subjectImage.focus()
+
+  panByDrag: (e, d) ->
+    return if @state.panEnabled == false
```
eatyourgreens: `return if @state.panEnabled`

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+
+    @setState
+      viewBoxDimensions:
+        x: Math.max(minumumX, Math.min(changedX, maximumX))
+        y: Math.max(minimumY, Math.min(changedY, maximumY))
+        width: @state.viewBoxDimensions.width
+        height: @state.viewBoxDimensions.height
+
+  frameKeyPan: (e)->
+    return if @state.panEnabled == false
```
eatyourgreens: `return unless @state.panEnabled`

app/classifier/frame-annotator.cjsx
```diff
@@ -23,8 +23,10 @@ module.exports = React.createClass
     onChange: Function.prototype
 
   getInitialState: ->
-    naturalWidth: 0
-    naturalHeight: 0
+    naturalWidth: @props.naturalWidth
+    naturalHeight: @props.naturalHeight
```
eatyourgreens: props should be immutable, so storing `naturalWidth` and `naturalHeight` as props and state looks wrong to me. If they don't change, then you can remove the state variables.

app/components/frame-viewer.cjsx
```diff
@@ -86,9 +104,34 @@ module.exports = React.createClass
...
-        {frameDisplay}
-      </FrameWrapper>
+      <div>
+        <FrameWrapper frame={frame} naturalWidth={@state.frameDimensions?.width or 0} naturalHeight={@state.frameDimensions?.height or 0} panByDrag={@panByDrag} viewBoxDimensions={@state.viewBoxDimensions or "0 0 0 0"} workflow={@props.workflow} subject={@props.subject} classification={@props.classification} annotation={@props.annotation} loading={@state.loading} onChange={@props.onChange} panEnabled={@state.panEnabled} >
+          {frameDisplay}
+        </FrameWrapper>
+        {if ( @props.project? && 'pan and zoom' in @props.project?.experimental_tools)
+          <div className="pan-zoom-controls" >
+            <div className="draw-pan-toggle" >
+              <div onClick={@togglePan} className={if @state.panEnabled then "" else "active"} >
```
eatyourgreens: `onClick` should be avoided on divs. Better to use it with a link or a button.

app/components/frame-viewer.cjsx
```diff
@@ -86,9 +104,34 @@ module.exports = React.createClass
...
+                <button title={"draw"} className={"fa fa-mouse-pointer"} title={"annotate"} onClick={@togglePanOff}/>
```
eatyourgreens: I'm not sure why you've wrapped one `onClick` handler in a second `onClick` handler here.

app/components/frame-viewer.cjsx
simensta: I think the initial state value of `zoomingTimeoutId` is misleading. Instead of `0`, the value should be `null`. `zoomingTimeoutId` refers to the id returned from `setTimeout()`, so I would like to keep the property name, but change the property's initial value. Thoughts?

eatyourgreens: Ok. In that case, you just need to change it to `zoomTimeoutId` when you assign it later.

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+
+  continuousZoom: (change) ->
+    return if change == 0
+    @setState zooming: true
+
+    zoomNow = =>
+      # return if change == 0
+      @zoom(change)
+      clearTimeout @state.zoomingTimeout
+      @setState zoomingTimeout: setTimeout(zoomNow, 200)
```
eatyourgreens: Should be `zoomingTimeoutId` on these two lines.

app/components/frame-viewer.cjsx
```diff
@@ -144,4 +187,151 @@ module.exports = React.createClass
...
+    
+    zoomNow()
+
+  zoom: (change) ->
+    return if change == 0 || @state.zooming == false
+    clearTimeout @state.zoomingTimeout
```
eatyourgreens: Should be `zoomingTimeoutId`

app/components/frame-viewer.cjsx
simensta: Ah I see. Definitely.

Issue #2085, opened 2016-01-08
issues with new vertical sorting


aliburchard:
@amyrebecca @brian-c looks like vertical sorting went into effect for all projects using the original camera survey task (in response to #2024), but there seem to be some issues with distributing species among the columns (whatever code determines the appropriate #choices per column seems to be wonky):

see: 
https://www.zooniverse.org/projects/aliburchard/camera-catalogue/classify
https://www.zooniverse.org/projects/aliburchard/emammal/classify

and more, but can't actually get on Zooniverse just now.

In the camera-catalogue project, for example, instead of the list running off the page, you have to scroll right to see the other choices, which I wouldn't have guessed.

brian-c:
Not sure how you ended up doing this @amyrebecca. Seems like there should be a way to do it with flexbox, but if you end up needing to manually re-sort the items here's the function that does it in Serengeti:

https://github.com/zooniverse/Serengeti/blob/master/app/lib/columnize.coffee

amyrebecca:
I was able to use flexbox for this! The only problem with the flexbox implementation is a browser bug where i can't use max-height properly and have to set height instead.

The fact that it's affecting the regular survey task is just my mistake and I've already merged a PR for it. In general we'll have to talk about really really long lists of options and how they should scroll. There's not really a good way to handle that and also provide columns. Given the option, flexbox will just make a reallllllly long single-column list, so you have to restrict the height; and once the height is restricted, you're stuck scrolling to the right to get options that don't fit.

aliburchard:
following up as per our chat the other day. Is it possible (perhaps using the code Brian suggested above) to create your own function to bypass the flexbox bug? If you write a function that counts the # of choices and distributes those choices evenly across three columns and calculates the #choices per column? Then you just hand that number back to flexbox.

aliburchard:
Just thinking more about this -- wondering if @brian-c has any insights into possible workarounds for this?

brian-c:
I messed around with some flexbox today and it seems like the only way to force a certain number of columns is to divide the number of items and try to calculate a height manually, which sorta defeats the point. I'd probably just go with the function to reorder the items.

aliburchard:
Yay!!!!!!! thanks @amyrebecca and @brian-c !!

PR #2208, opened 2016-02-12
Combine survey and flexible-survey tasks

amyrebecca:
These were only implemented separately in order to reduce risk of regressive impact in the short term. We're now ready to combine the two, as well as fix some outstanding bugs.


amyrebecca:
Ali decided that if we couldn't get all of the columns to be equal length, it was better to just stick with horizontal wrapping, so the last commit removes all of the style changes for drawing the choices in columns

brian-c:
Is the `columnize` function from Serengeti not working in this case? I thought we were gonna try that.

https://github.com/zooniverse/Serengeti/blob/master/app/lib/columnize.coffee

amyrebecca:
No, the only guidance I got on this was that it should be do-able with flex box, by using the data-breakpoint stuff you had set up

aliburchard:
Pretty sure we talked about writing custom functions to handle the dynamic lists: https://github.com/zooniverse/Panoptes-Front-End/issues/2085

amyrebecca:
weird, i see the discussion but i never got alerts about those messages.

unfortunately, i've got to move on to the geordi integration tasks, so the columns are not going to make it, certainly not for a launch on thursday :/

srallen:
@brian-c, do you have time today or tomorrow to review and merge? Wisconsin Wildlife Watch needs to get its second season launched this week.

app/pages/lab/workflow.cjsx
```diff
@@ -55,7 +55,8 @@ EditWorkflowPage = React.createClass
     @history.createHref("/projects/#{owner}/#{name}/classify", viewQuery)
 
   canUseTask: (project, task)->
-    task in project.experimental_tools
+    # task in project.experimental_tools
+    true
```
brian-c: Is this left over from dev?

brian-c:
Checking this out now.

@amyrebecca or @aliburchard, what's the main difference with the flexible task? Looks like most of the changes are in the CSV import functions, has the format of the CSVs changed?

brian-c:
I think I've got it all sorted now. Just a couple minor things:

- Some sub-lists in the CSVs are comma separated (choices' images, characteristics' icons) and some are semicolon separated (questions' answers, includes, excludes). We should pick one.

- `questionsMap` seems to map a question to the choices that need that question. Could that information not be stored on the question itself? Minor, but it'll make updates easier.

@aliburchard I'll get the columns sorted in a separate PR after this one is merged.

amyrebecca:
Thanks for the review! To reply to your concerns:

1. The commented out line and the "return true" was indeed from debugging; I reverted it
2. Good point; I've made this change
3. questionsMap is separate from the individual question because it not only lists which questions map to a particular choice, but also which order they should be presented in, so I think this is probably okay but I will change it if you feel strongly.

brian-c:
Looks good, thanks!

Sounds like #3 is fine if we want to customize question order.

Just had one more thought: are there any projects still using `flexibleSurvey`? We could prevent them from breaking by setting `flexibleSurvey: require './survey'`, but ideally we've converted them already.

amyrebecca:
We are going to be stuck launching WWW Season 2 as flexible survey for various reasons so I think including that line will probably be wise; should that line you suggest be in tasks/index.coffee ?

brian-c:
Yep, tasks/index. Since these are combined now, shouldn't updating WWW just involve change the task `type`, which we can do manually? Or are there considerations beyond the front end (e.g. aggregation) we need to take into account?

amyrebecca:
There aren't other considerations afaik, I just haven't tampered with the production database yet so this seems like the more natural fix :)

aliburchard:
Hey @brian-c and @amyrebecca what's the status of this? Just now getting caught up, but looking forward to having the tasks merged.

brian-c:
Will merge this morning, and then I'm gonna have a stab at implementing Serengeti's sorting function.

PR #2213, opened 2016-02-15
FrameViewer should pass frame up on load

brian-c:
Fixes zooniverse/Panoptes#1670. The frame wasn't being passed up to the classifier on load, which broke recording the image dimensions in the classification.


eatyourgreens:
:+1: Looks good to me.

PR #2214, opened 2016-02-16
Specify LTS Node and npm versions

brian-c:
I've set `engines` to the versions of Node/npm I'm currently running, but anything in the LTS range should work.


brian-c:
Please hold off on this a moment, sorry.

PR #2215, opened 2016-02-17
Move pfe-base back into main Dockerfile

astopy:
There's finally a binary version of phantomjs, which was the only reason
we separated these before.

@brian-c I'm assuming phantomjs 2.1 is OK (we were previously using 2.0)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,18 @@
-FROM zooniverse/pfe-base
+FROM ubuntu:14.04
 
-WORKDIR /src/
+ENV DEBIAN_FRONTEND noninteractive
+
+WORKDIR /src
+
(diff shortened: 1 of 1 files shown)


astopy:
Deployed to https://preview.zooniverse.org/panoptes-front-end/remove-pfe-base/ for testing.

This *shouldn't* make any difference to the build, but I removed some unused apt-get packages so it'd be good to make sure that didn't break anything.

brian-c:
This looks good to me. Not sure we technically need phantom anymore, since tests have moved to panoptes-client, but let's leave it so we can figure out how to get those tests running before deploy anyway.

Will make a note to merge tomorrow when we're both online, @astopy.

Issue #2155, opened 2016-01-28
inconsistent annotations with surveys vs. other tasks


ggdhines:
For whales as individuals (just to take an example project that does not use surveys), we have the following workflows and associated tasks:
84
T2
T3
T1

9
T4
T5
T2

An example annotation for Whales as Individuals is:
[{u'task': u'T4', u'value': 0}, {u'task': u'T5', u'value': 0}, {u'task': u'T2', u'value': [{u'y': 11.739976644608797, u'x': 36.41231883008673, u'tool': 0, u'frame': 0, u'details': []}, {u'y': 12.913974309069676, u'x': 936.8684843283519, u'tool': 0, u'frame': 0, u'details': []}]}, {u'task': u'T3', u'value': [{u'y': 197.23160762942777, u'x': 460.22545539445923, u'tool': 0, u'frame': 0, u'details': []}]}]
The annotations are given as dictionaries for each task - each dictionary has two keys: "task" and "values". The value of "task" is the task id - for example I can easily see that the first annotation corresponds to the first task in workflow 9. This is as it should be.
Now with Wildcam Gorongosa which uses the survey task: we have 
338
T1

so one task. An example annotation:
[{u'task': u'survey', u'value': {u'answers': {}, u'filters': {}, u'choice': u'NTHNGHR'}}]
Here the value associated with "task" should be "T1" but its not. "Survey" is not the task name, its the task type. For example, with Whales, its as if we had "task":"single". 
So someone confused task id with task type. Since my code needs to run on any project, I need consistency. For now I'll hard code in some exceptions.

brian-c:
Just a fun note: the Wildcam Gorongosa survey task is not the Panoptes-Front-End survey task.

This bug does not affect the Panoptes-Front-End survey task.

ggdhines:
alright - this will only be a problem for future custom projects which there really shouldn't be any. Other than being slightly annoyed nothing to do.

srallen:
@ggdhines, here is the relevant line in Wildcam's code base and could be easily changed: https://github.com/zooniverse/wildcam-gorongosa/blob/master/app/stores/classification-store.coffee#L41

However, I'm not sure if this has any implications mid-project if it is changed. We should also define what the convention is for this since one hasn't been. If it's follow whatever PFE is doing, that's fine, but we need it documented.

eatyourgreens:
I'm following a similar convention for Wellcome
[...]

PR #2220, opened 2016-02-20
Annotation documentation

brian-c:
Detailed documentation of annotations.

Also some general README.md updates and cleanup.

Viewable here: https://github.com/zooniverse/Panoptes-Front-End/blob/annotation-documentation/README.md#format-of-annotation-values

For #2155

Cc: @ggdhines, @chrislintott, @srallen, @eatyourgreens

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,11 @@
 
 Requires Node.js.
 
-Run `npm install` to get started.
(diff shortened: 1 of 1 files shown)


simensta:
Looks good -- thanks @brian-c!

PR #2227, opened 2016-02-23
Don't fetch field guide images unless the guide exists

parrish:
Just stumbled across this when trying to track down a different bug


brian-c:
Thanks for the catch.

PR #2257, opened 2016-03-01
Survey sort into columns

### LINKED ISSUE #2024: Filter Task: alphabetize vertically instead of horizontally?
@eatyourgreens and I noticed recently that while our original custom camera trap projects -- Snapshot Serengeti and Chicago WIldlife Watch -- sort the species list in columns and _then_ rows, PFE sorts the species in rows first. 

Vertically:
screen shot 2015-12-15 at 11 52 18

Horizontally:
screen shot 2015-12-15 at 11 52 43

I personally find the vertical sorting much more intuitive, and think we should be consistent across projects. With a lot of PFE camera trap projects getting close to launching (especially USA WW), I think this is a decision to make *soon* in order to maintain consistency. That being said, Wildcam Gorongosa is our only horizontally sorted project, and I don't know what a change to PFE would mean for that interface.

Really curious to hear from @trouille, @amyrebecca, and @srallen for thoughts on sorting for USAWW, and from @jwbmartin @brian-c @mrniaboc @vrooje @chrislintott or anyone else who might have thoughts on whether it's better to sort vertically v horizontally.

### LINKED ISSUE #2085: issues with new vertical sorting

### LINKED ISSUE #2245: Fix vertical sorting in survey task

I think this may be already addressed and deployed per #2024 and #2085.
@amyrebecca is it correct that these are "done"  now?

@aliburchard Can you confirm if there is anything left to do here or do we have what we need?

brian-c:
This sorts items in the survey task into columns using the sort function from Serengeti (which I've pulled out into a separate module).

In the process, I've explicitly decoupled thumbnail size from column count to simplify the CSS a bit.

Cc @aliburchard

Closes #2245
Closes #2085
Closes #2024

### DIFF
--- a/app/classifier/tasks/survey/chooser.cjsx
+++ b/app/classifier/tasks/survey/chooser.cjsx
@@ -2,8 +2,6 @@ React = require 'react'
 TriggeredModalForm = require 'modal-form/triggered'
 sortIntoColumns = require 'sort-into-columns'
 
-THUMBNAIL_BREAKPOINTS = [Infinity, 40, 20, 10, 5, 0]
-
 module.exports = React.createClass
   displayName: 'Chooser'
 
@@ -26,10 +24,18 @@ module.exports = React.createClass
       else
         choiceID
 
+  whatSizeThumbnails: ({length}) ->
+    if length <= 5
+      'large'
+    else if length <= 10
+      'medium'
+    else if length <= 30
+      'small'
+    else
+      'none'
+
   howManyColumns: ({length}) ->
-    if length is 0
-      0
-    else if length < 5
+    if length < 5
       1
     else if length < 20
       2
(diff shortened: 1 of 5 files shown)


aliburchard:
@brian-c I can't remember - are we holding off on merging this for some reason? (I vaguely rememeber something being buggy, but would love to see this merged if we can.) Or is it just waiting for review?

amyrebecca:
brian was waiting for me to get a chance to review this, mainly. i'm looking at it right now.

amyrebecca:
Wow so all of the logic went out of the CSS and back into the javascript. That makes me feel better about not being able to find a CSS solution to this problem!

PR #2258, opened 2016-03-01
transparent underlay style for subTask

simensta:
In master, the subTask modal is render with a non-transparent underlay. This PR proposes to make that underlay transparent in order to give participants the clearest and most consistent view of a subject during a subTask. 

Example with non-transparent underlay:
screen shot 2016-03-01 at 1 03 29 pm

Example with transparent underlay:
screen shot 2016-03-01 at 1 02 47 pm

This PR will not change the underlay style for the "Need some help" modals.

### DIFF
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -10,7 +10,7 @@ SEMI_MODAL_FORM_STYLE =
 
 SEMI_MODAL_UNDERLAY_STYLE =
   pointerEvents: 'none'
-  backgroundColor: 'rgba(0, 0, 0, 0.3)'
+  backgroundColor: 'rgba(0, 0, 0, 0)'
 
 module.exports = React.createClass
   displayName: 'DrawingToolRoot'


brian-c:
I believe the visually-darkened-but-still-passes-through-clicks treatment here was added when we allowed sub-tasks to be required. The idea was to give the appearance that the user needed to complete the dropdown before they could do other stuff, but still allow them to modify the shape they drew.

I'm afraid if we remove the underlay entirely, users will ignore required sub-tasks. Any thoughts?

Might be fun to build in something similar to how the old Zootorial library focused on things (see the Seregeti tutorial for an example). Probably more work than we can afford right now though.

simensta:
I think the current visual cue of greying-out the 'OK' button until an answer is provided is sufficient to communicate that a subTask is required. Because a user may need to refer to the underlaying image to address the question in the subTask modal, I think darkening the image is more hinderance than help. When demoing projects with transcription subTasks, I've received feedback that darkening the image makes completing a classification more difficult. 

I do like the way that the old Zootorial library focuses on items. Until we can darken parts of an image with this precision during a subTask, I think it is better not to darken the whole image.

Side-note: when I was looking into required tasks, I realized that a text task cannot yet be required, so I created Issue #2278.

PR #2263, opened 2016-03-04
Stop the subject SVG from filling its parent

### LINKED ISSUE #2262: need minor tweak to camera survey "delete" button
The button to delete an accidental species classification on the survey tasks is being obscured by the subjects. Normally you can see the species name and an "x" by it beneath the task, but now the pictures are bigger and overflowing onto that.

In this example from WWW, you can sort of see the little tiny white "x" in the bottom left corner, but not the name of the species.  Also noticing this on other projects now, so wondering if there was a change to the container for the subject, @amyrebecca or @brian-c ?

screen shot 2016-03-03 at 12 17 37

brian-c:
Fixes #2262.

Not sure what the intention was to have the SVG grow as big as it can. In general it should maintain the proportions of the subject image. If I've broken something unknowingly here, let me know.

(Seems I can't assign this to @itsravenous. Mind taking a look Tom? Thanks!)

### DIFF
--- a/css/frame-annotator.styl
+++ b/css/frame-annotator.styl
@@ -6,6 +6,4 @@
     position: absolute
     left: 0
     top: 0
-    width: 100%
-    height: 100%
     overflow: hidden


itsravenous:
I remember there being a reason, but it's likely not applicable any more - I've tested this with multi-image and multi-video subjects (my work on those is when it was introduced) and all seems well. Sorry about the breakage.

eatyourgreens:
Wasn't this added so that the SVG resizes correctly after the subject image is cropped?

itsravenous:
Ah, I think you may be right there...


*Tom Jenkins*
Ravenous - Web and Mobile development

*@itsravenous*
*itsravenous.com <http://itsravenous.com>*

Registered in England as It's Ravenous Ltd.
Ltd. Company No: 8781815
Registered office (no post) Holland House, 1-5 Oakfield, Sale, Cheshire M33
6TT

itsravenous:
Yes, @eatyourgreens is correct - first screenshot is with the width and height rules, second is without, for the same crop.
screenshot from 2016-03-05 10-14-29

eatyourgreens:
This might affect pan-and-zoom in #2199 too.

aliburchard:
@itsravenous @eatyourgreens @brian-c Any thoughts on when we might be able to push this fix through? I'm really worried about how #2262 is affecting the data on WWW.

aliburchard:
Ah, okay @eatyourgreens just explained this to me. So, I understand that this fix will break crop and pan&zoom tools (which will affect retinas, and @simensta's stuff). 

It looks like a quick fix (suggested by jim) that wouldn't break anything else would be to display the species ID tags *over* the SVG and subject. It will be ugly, but I'm really worried about the quality of data right now, so it seems like the best possible solution for now. @brian-c any way you'd be able to do that? thanks!

eatyourgreens:
Adding `position:relative` to `.survey-annotator-proxy` might be a quick fix. Haven't checked it in anything except Chrome.

eatyourgreens:
That fix works in IE and Chrome, and shouldn't affect the crop task, so I've opened #2268

brian-c:
Closed by #2268

PR #2265, opened 2016-03-07
Fix: dev/classifier images don't load over HTTPS

eatyourgreens:
Quick fix for Mixed Content warnings that stop the example images from loading.


brian-c:
I'm pretty sure lorempixel serves over https, [their server is just down right now](https://twitter.com/lorempixel/status/706863159395614720). Maybe you had the images served over http cached?

eatyourgreens:
Ah right, my bad, but we should fix those Mixed Content warnings by removing the http: URLs anyway?

brian-c:
Oh, right on! Pre-coffee. I totally read the diff backward, sorry!

PR #2266, opened 2016-03-07
Don't set base href to DEPLOY_SUBDIR

astopy:
No longer necessary with new preview URLs: https://remove-base-href.pfe-preview.zooniverse.org

### DIFF
--- a/views/index.ejs
+++ b/views/index.ejs
@@ -4,7 +4,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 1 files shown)


views/index.ejs
```diff
@@ -4,7 +4,7 @@
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <base href="<%= process.env.DEPLOY_SUBDIR === undefined ? '/' : '/' + process.env.DEPLOY_SUBDIR + '/' %>" />
+    <base href="/" />
```
eatyourgreens: Better to delete the `<base>` tag completely?

PR #2270, opened 2016-03-08
Fix default images

### LINKED ISSUE #2267: Survey task - breaks classification with no explanation
@brian-c @amyrebecca I'm trying to rebuild several of the camera trap projects with the new merged tool, but when I go to test out the workflow I'm just getting a spinning gyroscope and the summary classification screen. Any ideas? 

If it helps, I've made you collaborators on my test project that I'm using to troubleshoot, https://www.zooniverse.org/lab/2083/workflow/1639 -- note that the workflow built with the old survey works, just not this one. (*note: amy, i can't find your Z username, but brian you're a collab.)

Thanks!

screen shot 2016-03-08 at 12 00 27

brian-c:
Choices were being created with an `image` property instead of `images`. This would break if no images were specified to override it.

Closes #2267.


aliburchard:
Thank you!

Issue #2202, opened 2016-02-11
On the How To page, markdown overflows its boxes


alexbfree:
https://www.zooniverse.org/lab-how-to

![screenshot 2016-02-11 15 48 36](https://cloud.githubusercontent.com/assets/1473244/12981083/15602a3a-d0d7-11e5-854d-570a3be7e467.png)
![screenshot 2016-02-11 15 48 47](https://cloud.githubusercontent.com/assets/1473244/12981082/155feff2-d0d7-11e5-8b97-b98fb1e8f2e4.png)

srallen:
Fixed in #2271

PR #2271, opened 2016-03-09
Add Overflow to Markdown Boxes

wgranger:
Closes issue #2202


css/secondary-page.styl
```diff
@@ -53,6 +53,7 @@
       background-color: #E8E8E8
       border-radius: 4px
       padding: 1em
+      overflow: auto
```
srallen: Tiny quibble, but our convention is to put css properties in alphabetic order (you can find this info in the repo readme).

PR #2280, opened 2016-03-10
Add Component Mount to Set First Page

wgranger:
"page" default was returning NaN. The componentDidMount function in turn does what the default intended to do by setting the default first page through setPage.

### DIFF
--- a/app/talk/lib/paginator.cjsx
+++ b/app/talk/lib/paginator.cjsx
@@ -28,6 +28,9 @@ module?.exports = React.createClass
     previousLabel: <span><i className="fa fa-long-arrow-left" /> Previous</span>
     nextLabel: <span>Next <i className="fa fa-long-arrow-right" /></span>
 
+  componentDidMount: ->
+    @setPage(1)
+
   mixins: [History]
 
   setPage: (activePage) ->


parrish:
Hey, I must've missed this from when I was on vacation.  I'll take a look now.

parrish:
This is probably a little counter intuitive, but this is working a bit differently than you might expect.

Adding logging to `componentDidMount`, `componentWillReceiveProps`, and `render`

Here's how it works before this change:
```
render page NaN
componentDidMount (do nothing)
componentWillReceiveProps 2
render page 2
```

Here's how it works after this change:
```
render page: NaN
componentDidMount setPage(1)
componentWillReceiveProps page: 2
render page: 2
componentWillReceiveProps page: 1
render page: 1
```

In other words, this causes it to always show page 1 when navigating directly to the url (rather than clicking the paginator buttons).

The reason the paginator originally renders with `page: NaN` is because the parent component passes the prop with something like `page: +@state.meta.page` which will be `+undefined` (`NaN`) before the ajax request is finished.  Then it will be `page: 2`, which triggers another `render`.

Other than deferring the Paginator rendering until resolving the API request, this is pretty well unavoidable, but I don't think it's causing any problems.  I guess we could always have something like

``` coffeescript
render: ->
  {page, pageCount} = @props
  return <div /> unless page
```

Happy to dig deeper if there is a bug here somewhere.

wgranger:
@parrish It seems by this process the page number is set by default to one after the initial render, but I wasn't able to find a way it interfered with pagination otherwise. Should I look into deferring the rendering until after the ajax request?

parrish:
Apparently, React's default props don't override falsey prop values (like NaN in this case).

Since I don't think it's really changing behavior to render with NaN before the ajax request completes, it's probably fine.

To cover all the unwanted values, we could check in render with

``` coffeescript
render: ->
  {page, pageCount} = @props
  return <div /> unless page > 0 # covers undefined, null, and NaN cases
```

But, that's probably not necessary.

wgranger:
I noticed when the page is rendered with NaN, the user is unable to click "next" to see a new list of projects. The page jumps back to the top without the projects changing. This happens when you view all projects under the "set project status" section under the Admin bar.

PR #2287, opened 2016-03-14
Subject viewer styling

### LINKED ISSUE #2285: favorite, collections, and metadata button shifted over from edge of image in classification interface
@brian-c  it looks like a recent change creates extra space that pushes over the task text and tools on all projects. I think it's also causing the image to display a bit smaller than it would normally would if the edge of the collection button lined up with the edge of the image  Can someone look into that?

screen shot 2016-03-13 at 4 50 37 pm

screen shot 2016-03-13 at 4 50 15 pm

same thing in chrome and happens whether you're logged in or not

screen shot 2016-03-13 at 4 53 45 pm

eatyourgreens:
Changes pan-and-zoom to use flexbox.
Restores `subject` class on SVG marking surface.
Fixes #2285

### DIFF
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -127,7 +127,7 @@ module.exports = React.createClass
       <div className="subject-area">
         {if BeforeSubject?
           <BeforeSubject {...hookProps} />}
-        <svg style=svgStyle viewBox={createdViewBox} {...svgProps}>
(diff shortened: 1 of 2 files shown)


eatyourgreens:
@itsravenous @simensta can you take a look to see if this will break any of the work that you've been doing on the layout of the subject viewer?

simensta:
@eatyourgreens, look good to me. Thanks!

PR #2301, opened 2016-03-15
Mini course editor

srallen:
In anticipation of the mini course being added, zooniverse/panoptes#1726, I've abstracted out the tutorial editor into `project-modal-editor` in partials and added it to the list of experimental features. The only thing I really changed from the original tutorial editor was adding a `type` prop. How it will look and function in a project itself will come in a separate PR.

### DIFF
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -16,6 +16,7 @@ EXPERIMENTAL_FEATURES = [
   'combo'
   'tutorial'
   'field guide'
+  'mini-course'
   'hide classification summaries'
   'pan and zoom'
 ]
--- a/app/pages/lab/mini-course.cjsx
+++ b/app/pages/lab/mini-course.cjsx
@@ -7,10 +7,10 @@ module.exports = React.createClass
   render: ->
     <div>
       <div>
-        <p>Here’s some text all about the tutorial editor. Lorem ipsum dolor sit amet, etc.</p>
+        <p>Here’s some text all about the mini-course editor. Lorem ipsum dolor sit amet, etc.</p>
         <p><strong>Work in progress.</strong></p>
       </div>
       <div>
-        <ProjectModalEditor project={@props.project} type="tutorials" />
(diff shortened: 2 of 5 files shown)


brian-c:
How similar are the tutorial and mini courses?

If they are basically exactly the same, could we base them on the same resource type? Is a tutorial just a mini course that everybody gets on first visit (alternatively, is a mini course just a tutorial that pops up at specific times)?

If they are slightly different (or there's a chance that they will be in the future), could we keep the front end code for working with them separate?

I'm concerned that this could turn into another OwnedCards type of thing, or the generic task editor, where things look similar enough on the surface but are really conceptually separate, and then we end up with UI code with all kinds of twisty conditional branches.

srallen:
They are very similar. The back-end resource will be identical to the tutorial, but separate. Since we need to support projects having both tutorials and mini-courses, it made sense to have a separate resource on the back end.

You can think of the mini course as a tutorial that pops up at specific times. The differences with the mini- course are:

- It pops up after X number of classifications. 
- It starts with the next slide content each time. We can track last slide seen in the user's project preferences
- Users can opt out entirely. This also can be tracked in user preferences. 
- Users can restart the mini-course later if they change their mind about opting out.

There will be a/b testing to determine the frequency of the pop ups. It's possible the results will tell us that researchers should control the frequency, which would necessitate adding this to the editor and the resource.

For now my plan was that they would be separate resources on the back-end, would share an abstracted out editor, but would have slightly different logic in the display with hard-coded pop up frequencies (that may vary based on the experiment) like the sign up prompt.

brian-c:
Alright, cool.

Looks like this is just missing the /lab/:id/mini-courses route.

Also: I totally forgot I've got a branch with some slightly nicer tutorial editor style updates. I'll rebase and PR it before you get too far on this.

srallen:
Whoops. The route is in there now. The page will have an error loading the editor, though, since the resource doesn't exist yet in Panoptes.

And thanks for the style updates, that'll be really helpful. 

This should be ready to merge now. I'll be working on the logic for how it displays in the projects in a separate PR once the resource exists.

Issue #2100, opened 2016-01-12
Slide tutorial: close button is inaccessible on small screens


eatyourgreens:
When the viewport height is smaller than the height of the slide tutorial form element, the close button can get positioned above the top of the page, and can't be reached by scrolling upwards.

Also: why is the slide tutorial using a form? It doesn't contain any form input elements or submit any data.

alexbfree:
I noticed a similar issue on iPhone 4 / iOS 9.2 - The close button and part of the tutorial window is off the right hand side of the screen and can't be reached.

srallen:
Fixed in #2303

PR #2303, opened 2016-03-15
Adjust Margins on Tutorial Modal

wgranger:
Close issue #2100 

Default left margin caused the tutorial modal to be pushed too far to the right on small screens. Increasing the top margin pushed the modal down far enough to reveal the close button.

### DIFF
--- a/css/tutorial.styl
+++ b/css/tutorial.styl
@@ -4,8 +4,8 @@
   padding: 0 0 1.5em
 
   @media screen and (max-width: 400px)
-    margin-left: 0vh
-    margin-top: 5vh
+    margin-left: 0em
+    margin-top: 1.75em
 
 .tutorial-steps
   display: flex
--- a/css/tutorial.styl
+++ b/css/tutorial.styl
@@ -2,6 +2,8 @@
   border-radius: 8px
   max-width: 400px
   padding: 0 0 1.5em
+  margin-left: 0px !important
+  margin-top: 28px !important
 
 .tutorial-steps
   display: flex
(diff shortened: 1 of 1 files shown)


srallen:
A few things: 
- Consider using `em` or [viewport units](https://developer.mozilla.org/en-US/docs/Web/CSS/length) instead of fixed pixel sizes. Either are better for responsiveness. Also, for consistency with our other styles.
- `!important` should be generally avoided and I also don't think it's necessary in this case. The tutorial.styl is further down the cascade.
- Also remember our alphabetic ordering of style properties convention.
- Consider a media query since this is only a issue for small viewports.

wgranger:
Good call on the styling adjustments. Need to get in the habit of keeping things alphabetical.

css/tutorial.styl
```diff
@@ -3,6 +3,10 @@
   max-width: 400px
   padding: 0 0 1.5em
 
+  @media screen and (max-width: 400px)
+    margin-left: 0vh
```
srallen: `vh` is viewport height, so it doesn't make sense to use `vh` here. To be consistent with our other styles, `em` would be best. From MDN:

> The CSS properties line-height, font-size, margin-bottom and margin-top often has a value expressed in em.

css/tutorial.styl
```diff
@@ -3,6 +3,10 @@
...
+    margin-top: 5vh
```
srallen: To be consistent with our other styles, `em` would be best. From MDN:

> The CSS properties line-height, font-size, margin-bottom and margin-top often has a value expressed in em.

Issue #1053, opened 2015-07-13
Panoptes - allow a user to temporarily 'hide' lines / objects they have drawn on a image.


Capella05:
As a future improvement would it be possible to add an additional button that allows a user to toggle between the original image they have been given to classify, and their input?

When classifying on 'GZ: Bars' I am finding it very difficult to determine if the lines I have drawn are accurate to the galactic bar itself, as my input is masking the feature itself. I have found myself deleting the lines I have drawn just to get a second look!

Functionality wise, it would not be too dissimilar to the invert toggle button.

vrooje:
This is related to/similar to #204 and #648 - clearly it's a feature that everyone will find really useful!

The other reason to add this feature is that when marks are very close together it's sometimes impossible to mark where you need to if there's another mark adjacent. Workarounds include drawing elsewhere and then moving the new mark, but that's pretty awkward and it would be better to just hide the other marks and draw the new one in a blank space.

chrissnyder:
:+1: On the radar.

vrooje:
Bumping this - this is going to be a big problem for the disaster relief stuff, exemplified by this image:
![screen shot 2016-03-12 at 5 01 59 pm](https://cloud.githubusercontent.com/assets/2357126/13726272/ddeff2aa-e874-11e5-8958-edf696bc84a3.jpg)

If I've already marked e.g. structural damage with a rectangle I can't add a different marking inside the rectangle without drawing it outside first and then dragging it in. The only action possible within a previously-drawn shape is to select that shape, even if I have a different marking tool selected. That's a huge limit on how people can classify.

We are supposed to release this project imminently and I know this is going to be a big obstacle. How can we coordinate to come up with a solution here? Is anyone actively working on this?

Tagging @saschaishikawa @itsravenous @brian-c @chrislintott @aliburchard @mrniaboc ... others?

aliburchard:
Yes, this is definitely an obstacle, and I'd love to see it fixed. I'm not sure how hard it is to fix, but it certainly comes up in projects trying to mark things that are close together.

chowjunfat:
Ditto to Capella05 on Galaxy Zoo Bars.

[...]

Issue #648, opened 2015-06-22
Hard to draw adjacent polygons


mkosmala:
We had feedback from our Season Spotter beta (see #443) that it's hard to draw adjacent polygons. This is because the user ends up selecting the original shape when trying to draw a new one next to it. Not sure there's an easy fix. Would welcome ideas.

aliburchard:
Should note this isn't just related to polygons (though the multi-click nature of polygons makes them the most problematic). You can't, for example, mark a point over a rectangle, because clicking over the rectangle highlights the rectangle instead of creating a new point (if these tools are in the same task).

camallen:
Sounds like the hide old marks issue again.

chrissnyder:
Hiding old marks does solve it, as long as the volunteer clicks it before they try to make the new mark.

Do we think this is sufficient to solve this problem?

Alternative solutions?

- Only allow teams to ask for one type of mark per task
- "Freeze" all other marks when in the middle of creating a new mark.

mkosmala:
Hiding old marks works for me. 

"Only allow teams to ask for one type of mark per task" doesn't work because we ask, e.g., for them to mark the outline of trees. If two trees are next to each other, it's the same marking tool, just two different marks. 

Freezing works so-so. If the user starts the second shape near the first one, it will be unclear to the GUI whether the user wants to make a new shape or select the old one. This would require the user to start the shape elsewhere. Not terrible, but not very user-friendly, either.

alexbfree:
Just adding a couple of other alternative solutions:
1) have a "hide all other shapes" toggle button so that people can temporarily hide other shapes if they are "in the way"
2) allow vector shapes to be selected by their edges (like in Powerpoint, Illustrator etc) rather than by clicking on the contained area....

eatyourgreens:
Closing this because "Hide previous marks" is an option now for projects.

PR #2304, opened 2016-03-16
Hide previous marks

saschaishikawa:
Creates a checkbox to hide all marks a user has created up until the checkbox is toggled (resolves issue #1053)


vrooje:
Closes #648 as well, I think?

alexbfree:
This looks great @saschaishikawa . I think this solves a really important problem, well done! 

One small suggestion, I wonder if it would waste less space to align the tool alongside the help button.. And maybe if doing so, it would be better before than after, like this:

screenshot 2016-03-16 11 11 24

Another small thing... I think it jars slightly that this is a checkbox but the label is a verb. Usually verbs are on buttons and checkboxes describe state. So I would be inclined to tweak the label to say something like "Previous marks hidden" - or else to change it to a button (that changes its label to Show all marks once pressed).

app/classifier/tasks/generic.cjsx
```diff
@@ -48,6 +48,22 @@ module.exports = React.createClass
...
+      {if @props.toggleHideMarksBefore
+        <div>
+          <p>
+            <small>
+              <strong>
+                <label className="hide-previous-marks-toggle">
+                  <input type="checkbox" onClick={@props.toggleHideMarksBefore}>
+                    Hide previous marks
+                  </input>
+                </label>
```
simensta: Hey @saschaishikawa, I would move the <Label /> text out of the `<Input />`, like so:
```
                <label className="hide-previous-marks-toggle">
                  <input type="checkbox" onChange={@props.toggleHideMarksBefore}>
                  </input>
                  {"Hide previous marks"}
                </label>
```
This will get rid of console warning:
```
 `Warning: input is a void element tag and must not have `children``
```

saschaishikawa: Thanks, @simensta. I was wondering why I kept getting that warning! I'll take care of that when I get a chance

eatyourgreens: `<input>` shouldn't have opening and closing tags in HTML either.

simensta:
Hey @saschaishikawa, this is looking good. Can you make the `hide-previous-marks` feature admin enabled or make the feature opt-in at the workflow editor level?

saschaishikawa:
Added `hide previous marks` as an experimental feature. I think that takes care of the remaining issues.

app/classifier/frame-annotator.cjsx
```diff
@@ -83,13 +83,16 @@ module.exports = React.createClass
     y = (e.pageY - @state.sizeRect?.top) / scale.vertical || 0
     {x, y}
 
-  render: ->    
+  toggleWarning: ->
```
brian-c: I don't think this method lives here anymore. Maybe a rebasing mistake?

app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -65,7 +64,10 @@ module.exports = React.createClass
                 onDestroy: @handleDestroy.bind this, annotation, mark
 
               ToolComponent = drawingTools[toolDescription.type]
-              <ToolComponent key={mark._key} {...toolProps} {...toolEnv} {...toolMethods} />}
+
+              if i < @props.hideMarksBeforeIndex then null # hide previous marks
```
brian-c: This `if`/`else` formatting is pretty weird. Consider breaking these lines and indenting.

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -50,13 +50,21 @@ module.exports = React.createClass
...
 
-      {if choicesKey?
+      {if choicesKey
         <div>
           <hr />
           <span className="form-label">Choices</span>
         </div>}
       {' '}
 
+    {if choicesKey is 'tools' and 'hide previous marks' in @props.project.experimental_tools # NOTE: assumes 'tools' is unique to drawing! --STI
```
brian-c: Maybe `if @props.type.task is 'drawing'`?

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -50,13 +50,21 @@ module.exports = React.createClass
...
+      [<label key="hide-previous-marks" className="pill-button">
```
brian-c: This array-of-one (and its `key` prop) isn't necessary (probably left over from a rebase, I think I remember working on this a few days ago).

brian-c:
- Ideally, the classifier should not have any direct knowledge of the implementation of any task. Logic for the drawing task should be contained in the drawing task module. Instead of keeping the `hideMarksBeforeIndex` in the classifier's state, maybe it can belong to the annotation.

- Is it really necessary to keep this behind an admin flag? Seems like that's our default when adding something, but things never make it out of that mode. We've done this in other projects, we know it's good stuff. Thoughts, @vrooje?

vrooje:
I agree we should have better pathways in place for evaluating whether a feature should come out of admin mode... but I do like the principle of first testing on projects where we know the team is aware it's an early phase of the tool.

With this particular tool the main thing I'm wondering about is whether people are going to forget they've hidden previous markings and re-draw things, and I'd like to use my own project as a test of that before releasing this to the wider world without any oversight.

Related question: will we be able to track whether someone has ticked this box during a classification (e.g. with Geordi)?

saschaishikawa:
Re: admin flag, what if we flip the logic and make it opt-out instead? Otherwise, I'm struggling to find a use case where someone wouldn't want to hide previous marks (other than it makes the interface needlessly complicated)

saschaishikawa:
@brian-c, good call on removing task-specific stuff from Classifier (I was struggling to find a way around that). I've made a few changes that should address this.

aliburchard:
@saschaishikawa just to clarify - you mean make the presence of that tick-box automatic and opt-out, right? So previous marks aren't automatically hidden, but the *ability* to hide them is there by default, right? (not being pedantic, just want to make sure I *can* see all my already marked wildebeest on https://www.zooniverse.org/projects/dl-j/serengeti-wildebeest-count by default).

brian-c:
What should be happening to the hidden marks when we move on to the next task? Currently it just sticks to whatever was set, with no way to change it afterward. Kinda weird.

Maybe this should actually be stored classification-level, and a `PersistBeforeSubject` hook could be added for the control. Then it'd apply to _all_ drawing marks, not just those made as part of the current task. Seems like it'd be more useful that way anyway.

saschaishikawa:
@aliburchard correct. I was only referring to whether or not the "Hide previous marks" checkbox appears to the user. As you put it, their _ability_ to hide previous marks, should they choose to use it.

vrooje:
@brian the marks made in the previous task are "fixed" when moving on to the next task, right? So if I drew a huge rectangle around the whole image in task 1 and then moved on to task 2, I could still mark within the T1 rectangle during T2 even if it wasn't hidden. 

If that's correct then I think we should consider revealing all marks when a user advances to the next task, so they don't forget they have things hidden.

aliburchard:
@vrooje, yah, i like that approach.

saschaishikawa:
@brian-c, added an AfterTask hook to render the 'Hide previous marks' checkbox. Also set it up to auto-disable hiding marks when the an annotation is added, which brings in a potentially weird side-effect: if you hide marks and switch drawing tools in the same task, the checkbox gets unchecked. I thought this was OK.

Re: checkbox location, @brian-c expressed preference for moving the checkbox/button above the subject. I had some difficulty creating a hook to make that happen. I figured we'd get the behavior right before worrying about location.

brian-c:
Looked into figuring this out a bit more this morning. The organization of the subject side of the classifier has sort of disintegrated a bit with all the recent changes, and there's really no appropriate place to hook into right now, so I'm fine with sticking this on the task side.

Having an `AfterTask` hook doesn't really make sense (you should just put that content in the task itself), but a `PersistAfterTask` (rendered by the classifier between the task component and the task-nav) would work since we want this control to be available throughout the whole classification process. See the implementation of the `PersistInsideSubject` hook rendered by the `FrameAnnotator`.

Also, in my opinion the setting should stick between tool selections and between tasks.

saschaishikawa:
@brian-c: so replace `AfterTask` with `PersistAfterTask` along these lines? Gonna try to figure out how to keep the checkmark state between tool selections

app/components/subject-viewer.cjsx
```diff
@@ -44,6 +44,11 @@ module.exports = React.createClass
     frameDimensions: {}
     inFlipbookMode: @props.allowFlipbook
 
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.annotation.task isnt @props.annotation?.task # task changed!
+      @props.classification._hidePreviousMarks = false
```
saschaishikawa: I don't think this should be in subject-viewer, but I couldn't find another way determine when the task changes.

brian-c:
Implemented in #2408.

PR #2305, opened 2016-03-16
Prevent images from overlapping the frame in Firefox

### LINKED ISSUE #2285: favorite, collections, and metadata button shifted over from edge of image in classification interface

alexbfree:
This fixes #2285. Tested in Chrome & Firefox & Safari. Also checked Pan & Zoom tool still aligned correctly in Dev Classifier

### DIFF
--- a/css/frame-annotator.styl
+++ b/css/frame-annotator.styl
@@ -1,6 +1,7 @@
 .frame-annotator
   // Layout
   position: relative; // Contain absolutely-positioned children
+  max-width: 100%;
 
   svg
     position: absolute


camallen:
@alexbfree you happy that this fixes it?

alexbfree:
OK @camallen I will merge it.

eatyourgreens:
:+1:

PR #2308, opened 2016-03-16
Project announcements banner

brian-c:
This just adds a super basic banner for project announcements.

In the future we might integrate this with Talk/notifications/project subscriptions, but for now it's just some text stored in the project configuration.


parrish:
:shipit:

PR #2309, opened 2016-03-16
Explicit flex prop on child items. Fixes subject viewer in IE 11

srallen:
Got notified by WWW team that the subjects in the classifier weren't displaying at all in IE. Explicitly stating the `flex` prop on the child items fixes this. Tested in Chrome, Firefox, Safari, IE 11, and iOS Safari and this doesn't seem to break anything. 

I looked in MS Edge, but it appears to have some other issues that I'll have to take a look at later.

### DIFF
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -18,6 +18,9 @@
 .classifier
   display: flex
   justify-content: center
+  
+  > *
+    flex: 1 1 auto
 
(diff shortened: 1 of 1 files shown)


alexbfree:
Confirmed, tested in IE11 and it makes the subject viewer appear again. Merging.

alexbfree:
Just confirmed, this is now deployed on zooniverse.org and working in IE11

PR #2311, opened 2016-03-16
Pan Zoom Revisions: Scale drawn marks when zooming

simensta:
This PR:
- scales drawn marks correctly when zooming in on a subject
- removes some unnecessary state variables, methods, and setTimeout calls.

I think the next step is to move pan & zoom into its own component, but this PR will fix the above issues in the meantime.

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -150,8 +150,8 @@ module.exports = React.createClass
       clearTimeout @state.zoomingTimeoutId
 
   zoom: (change) ->
-    return if !@state.zooming
     @clearZoomingTimeout()
+    return if !@state.zooming
     newNaturalWidth = @state.viewBoxDimensions.width * change
     newNaturalHeight = @state.viewBoxDimensions.height * change
   
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -78,13 +78,13 @@ module.exports = React.createClass
               </div>
             </div>
             <div>
-              <button title={"zoom out"} className={"zoom-out fa fa-minus" + if @canZoomOut() then " disabled" else "" } onMouseDown={ @continuousZoom.bind(this, 1.1 ) } onMouseUp={@stopZoom} />
+              <button title={"zoom out"} className={"zoom-out fa fa-minus" + if @cannotZoomOut() then " disabled" else "" } onMouseDown={ @continuousZoom.bind(this, 1.1 ) } onMouseUp={@stopZoom} />
             </div>
             <div>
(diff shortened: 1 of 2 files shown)


app/components/frame-viewer.cjsx
```diff
@@ -119,10 +116,10 @@ module.exports = React.createClass
               </div>
             </div>
             <div>
-              <button title={"zoom out"} className={"zoom-out fa fa-minus" + if @canZoomOut() then " disabled" else "" } onMouseDown={ @continuousZoom.bind(this, 1.1 ) } />
+              <button title={"zoom out"} className={"zoom-out fa fa-minus" + if @canZoomOut() then " disabled" else "" } onMouseDown={ @continuousZoom.bind(this, 1.1 ) } onMouseUp={@stopZoom} />
```
eatyourgreens: Confused by this. Disable the button if you can zoom out?

app/components/frame-viewer.cjsx
```diff
@@ -119,10 +116,10 @@ module.exports = React.createClass
...
             </div>
             <div>
-              <button title={"zoom in"} className="zoom-in fa fa-plus" onMouseDown={@continuousZoom.bind(this, .9)} onMouseUp={@toggleZoom} />
+              <button title={"zoom in"} className="zoom-in fa fa-plus" onMouseDown={@continuousZoom.bind(this, .9)} onMouseUp={@stopZoom} />
```
eatyourgreens: Ditto re. mouse event handlers.

app/components/frame-viewer.cjsx
```diff
@@ -200,17 +197,18 @@ module.exports = React.createClass
 
   continuousZoom: (change) ->
     return if change == 0
```
eatyourgreens: Do you need to check for a timeout and clear it, before you return here?

app/components/frame-viewer.cjsx
```diff
@@ -200,17 +197,18 @@ module.exports = React.createClass
...
-    zoomNow = =>
-      @zoom(change)
-      clearTimeout @state.zoomingTimeoutId
-      @setState zoomingTimeoutId: setTimeout(zoomNow, 200)
-    
-    zoomNow()
+    @setState zooming: true, =>
+      zoomNow = =>
+        # if !@state.zooming, we don't want to continuously call setTimeout
+        return if !@state.zooming
```
eatyourgreens: Not quite sure what this check does, since you set `state.zooming` to true before calling this function.

simensta:
@eatyourgreens -- thanks for the feedback! Do the above changes address your concern about corresponding key event handlers?

eatyourgreens:
Looks good to me. :+1:

Issue #2144, opened 2016-01-26
Video task: annotate events as they occur during playback


eatyourgreens:
This is basically porting the Solar Stormwatch 'What's That?' task to PFE. A task which allows you to record the times at which defined events occur during video playback. Stormwatch implemented this as a set of pause/record buttons, displayed after the video player. Pressing a button records the event type and current time of the video. Multiple events can be recorded for each type, so the annotation returned will probably be an array of arrays.

![image](https://cloud.githubusercontent.com/assets/59547/12580913/3de22f48-c42b-11e5-809a-5f49ddddc6c8.png)

eatyourgreens:
Pinging @brian-c and @itsravenous since this probably overlaps with #1965 and #2137

itsravenous:
Yep, so #1965 modifies `<SubjectViewer />` so that it delegates the rendering of a single image/video (and any drawing tools for it) within the subject to `<FrameViewer />`.

I think your controls need to be per-video, in case you have two videos in one subject (or am I over thinking this?), so they would end up being injected into `<FrameViewer />` just as drawing tools currently are (the pause button method would work fine - you could e.g. have a popup come up when you click pause to choose the event to record at that timestamp).

brian-c:
Should I create a separate PR with just the classifier changes needed for the combo task? The combo task itself shouldn't step on anybody's toes, but there are some changes to how tasks report their changes.

eatyourgreens:
@itsravenous I think the marking controls can probably be rendered by the task's `AfterSubject` hook, in which case they should work fine with your branch. Does your branch pass the frame number through to that hook, or does each frame just render the same `AfterSubject` call?

eatyourgreens:
@brian-c that sounds like a sensible approach to me. :+1:

itsravenous:
@eatyourgreens yep, frame number [is included](https://github.com/itsravenous/Panoptes-Front-End/blob/subject_viewer_frame_wrapping/app/classifier/frame-annotator.cjsx#L102) in `hookProps` which is passed to `AfterSubject`. I misspoke before, I meant "injected into `<FrameAnnotator />`".

trouille:
@eatyourgreens Molly Wick of the Environmental Protection Agency in the U.S. is interested in setting up a video project. She asked, "would it be possible to have users watch the video clips and then clicking on a fish when it appears as the video is playing so that the presence, location of the fish, and the timestamp of the video are recorded."

[...]

PR #2319, opened 2016-03-17
Video player

eatyourgreens:
Work towards #2144 

Moves the video player into its own component and tidies up the flipbook code for subjects with multiple images.

Might affect projects that use video or projects that use the flipbook slideshow for subject images, or projects that use pan-and-zoom.


simensta:
@eatyourgreens, looks good to me. Thanks!

PR #2323, opened 2016-03-17
Auto delete out-of-bounds marks

### LINKED ISSUE #2125: bug: coordinates outside of image frame
We're going through the first tranche of Fossilfinder data and we perplexed to see coordinates for use of the point tool falling outside of the image. At first we thought that it could be something to do with image size as seen but quickly realised that negative numbers wouldn't make sense at all. And we also thought that perhaps users could click and mark areas of the webpage outside of the image area. It seems to be neither of those things.

A user can click a point on the image with the drawing point tool, where it will drop the marker. But then they can drag this marker outside of the image field. It disappears but on completion the summary page shows that this marker is still registers and when you click on more it shows the coordinates to be outside of the frame. This is not uncommon - we have a few thousand examples in our database.

Could a check be put in place to ensure that markers outside of the coordinate system of an image are discarded? Presumably users are dragging the markers off the image when they have changed their mind in these cases, rather than clicking on the x to delete the marker.

brian-c:
This adds a check after the "main" drag of each drawing tool to see if the its element's bounding box has left the bounds of the subject image. If so, its associated mark is deleted from the annotation.

The same behavior is added to the `initValid` check of the point tool, which is the only tool that can be _initialized_ out of bounds.

Closes #2125.

### DIFF
--- a/app/classifier/drawing-tools/circle.cjsx
+++ b/app/classifier/drawing-tools/circle.cjsx
@@ -3,6 +3,7 @@ EllipseTool = require './ellipse'
 DrawingToolRoot = require './root'
 DragHandle = require './drag-handle'
 Draggable = require '../../lib/draggable'
+deleteIfOutOfBounds = require './delete-if-out-of-bounds'
 DeleteButton = require './delete-button'
 
 MINIMUM_RADIUS = 5
@@ -62,7 +63,7 @@ module.exports = React.createClass
       {if @props.selected
         <line x1="0" y1="0" x2={@props.mark.r} y2="0" strokeWidth={GUIDE_WIDTH / ((@props.scale.horizontal + @props.scale.vertical) / 2)} strokeDasharray={GUIDE_DASH} />}
 
-      <Draggable onDrag={@handleMainDrag} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
         <ellipse rx={@props.mark.r} ry={@props.mark.r} />
       </Draggable>
 
--- a/app/classifier/drawing-tools/delete-button.cjsx
+++ b/app/classifier/drawing-tools/delete-button.cjsx
@@ -12,8 +12,6 @@ CROSS_PATH = "
   L 0 #{RADIUS * 0.7 }
 "
 
-DESTROY_TRANSITION_DURATION = 300
-
 module.exports = React.createClass
   displayName: 'DeleteButton'
 
@@ -21,6 +19,7 @@ module.exports = React.createClass
     x: 0
     y: 0
     rotate: 0
+    destroyTransitionDuration: 300
 
   render: ->
     transform = "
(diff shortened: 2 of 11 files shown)


simensta:
If I drag the drawing tool totally outside the bottom bound of certain subjects, `delete-if-out-of-bounds.coffee` returns false. This seems to only happen on images with greater height. Check out `projects/simenstad-tests/practice-project-3-16-2016/classify` on staging to see this in action.

brian-c:
@simensta Please give brian-testing access to that project. Thanks!

simensta:
@brian-c: no problem, brian-testing should have access now.

brian-c:
Yep, something weird's going on when getting shapes' bounding rects. I'll see if I can figure it out soon. Sorry.

brian-c:
There was a really dumb y-instead-of-x typo I just fixed (which of course took forever to track down), and I've applied the scale to the point tool's `initValid` check as you suggested (thanks!). Scale shouldn't be necessary on the other drag-end checks, since it's dealing directly with the element's dimensions as calculated by the DOM, which is already scaling the parent SVG. Care to take another look?

simensta:
@brian-c: Some combination of the values returned for `boundingBox` and `tool.props.containerRect` in `delete-if-out-of-bounds.coffee` are wrong for images that need to be scaled to fit in the subject-viewer. 

To replicate, drag a rectangle mark such that the top line of the rectangle is aligned with the bottom of the SVG viewport.
example rectangle mark on svg
At this point, I would predict that the value `rect.top` should be close to the value of `bounds.top` + `bounds.height`. The actual values are as follows:
`rect.top` = ~ 148.2 
`bounds.top` + `bounds.height` = ~ 403.4

I re-added an example subject to `projects/simenstad-tests/practice-project-3-16-2016/classify` on staging to demonstrate the issue.

alexbfree:
Just had a play with this! So glad we are doing something about this, it's bugged me for a while.

Is there any way could restrict the box as-you-drag? or at least when you drop the final corner?
Even with this new fix, we can still get into a mess with off-screen boxes shifting subtask popups into weird positions or becoming non-draggable/non-deleteable.

brian-c:
@simensta Looks like I was missing the scroll position of the window, which would cause that behavior. Huge thanks for checking this so thoroughly.

@alexbfree That behavior might make sense for rectangles, but it wouldn't for any other shapes (ellipses, polygons, etc.) so I'd rather keep it consistent. Good point on the popup positioning though. Currently a popup just attaches to another element, but since that element might be clipped, maybe we should have a way to limit it to not go beyond another. I don't think there's a way to programmatically detect clipping by a parent's overflow, but I'll research.

alexbfree:
@brian-c Ah yes, good point about different shape types... So are you deleting the whole shape when it's something like a point or ellipse? And adjusting the shape when it's rectangle/poly/line?

Regarding the positioning of pop-up: Judging by the feedback from the Snapshot Serengeti: Computer Vision beta, people really don't like the use of the popup for the subtasks... we might want to think longer term about reworking the way people answer subtasks anyway (perhaps now we have the ability to hide/show marks, we could do something like, hide all but one mark, ask them subtask questions about that, then show the next one, ask questions about that, etc...). Anyway that's a side point but just thought I would mention it.

simensta:
@brian-c : Good to know, the PR looks good to me. Thanks!

@alexbfree: The auto-delete behavior introduced in this PR is not dependent on mark type (e.g., point, ellipse, rectangle). If all of a mark's points are dragged outside the SVG viewport, the mark is deleted. If the mark is partially inside the SVG viewport, the mark isn't deleted or adjusted.

PR #2330, opened 2016-03-21
Feature detect IE11 and only apply flex prop when detected

### LINKED ISSUE #2329: subject viewer now displaying really wide for all projects
I suspect this was introduced to fix some of the display issues, but I'm not a fan.

screen shot 2016-03-21 at 11 58 05
screen shot 2016-03-21 at 11 57 31
screen shot 2016-03-21 at 11 59 25

srallen:
Fixed #2329 by feature detecting for IE11 and only applying `flex: 1 1 auto` to the subject viewer when it is IE11. Flex-grow must be set to 1 for the subject viewer to display correctly in IE11. I tried out ideas with setting max-width to the parent element or possibly setting max-width of the viewer to the naturalWidth of the subject image, but both of those didn't entirely solve the issue or had unintended consequences.

Deployed as a feature branch: https://preview.zooniverse.org/panoptes-front-end/fix-2329
@alexbfree or @eatyourgreens care to take to look to make sure I'm not inadvertently breaking something somewhere. I checked chrome, FF, IE11, safari, and iOS safari and it looked alright to me. IE11 will still possibly have the black background behind subjects with large viewports, but here's hoping our IE support days will be behind us soon :weary:

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -5,6 +5,7 @@ alert = require '../lib/alert'
 getSubjectLocation = require '../lib/get-subject-location'
 CollectionsManagerIcon = require '../collections/manager-icon'
 FrameViewer = require './frame-viewer'
+classnames = require 'classnames'
 
 NOOP = Function.prototype
 
@@ -17,7 +18,6 @@ subjectHasMixedLocationTypes = (subject) ->
         allTypes.push type
   allTypes.length > 1
 
-ROOT_STYLE = display: 'block'
 CONTAINER_STYLE = display: 'flex', flexWrap: 'wrap', position: 'relative'
 
 module.exports = React.createClass
@@ -51,9 +51,13 @@ module.exports = React.createClass
         inFlipbookMode: allowFlipbook
 
   render: ->
-    rootClass = 'subject-viewer'
-    if @props.workflow?.configuration?.multi_image_layout then rootClass += ' subject-viewer--layout-' + @props.workflow.configuration?.multi_image_layout
-    if @state.inFlipbookMode then rootClass += ' subject-viewer--flipbook'
+    rootClasses = classnames('subject-viewer', { 
+      'default-root-style': @props.defaultStyle
+      'subject-viewer--flipbook': @state.inFlipbookMode 
+      "subject-viewer--layout-#{@props.workflow.configuration?.multi_image_layout}": @props.workflow?.configuration?.multi_image_layout
+    })
+    # Feature detect IE11 and apply flex prop. Revisit or remove when IE11 is no longer supported.
+    rootStyle = flex: "1 1 auto" if "ActiveXObject" in window or window.ActiveXObject isnt undefined
     mainDisplay = ''
     {type, format, src} = getSubjectLocation @props.subject, @state.frame
     if @state.inFlipbookMode
@@ -91,7 +95,7 @@ module.exports = React.createClass
               </span>}
           </span>
 
-    <div className={rootClass} style={ROOT_STYLE if @props.defaultStyle}>
+    <div className={rootClasses} style={rootStyle}>
       {if type is 'image'
         @hiddenPreloadedImages()}
       <div className="subject-container" style={CONTAINER_STYLE} >
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -18,9 +18,6 @@
 .classifier
   display: flex
   justify-content: center
-  
-  > *
-    flex: 1 1 auto
 
   @media (max-width: 900px) // This is between iPad landscape and portrait.
     flex-wrap: wrap
@@ -38,7 +35,7 @@
     max-width: 100%
     > *
       display: flex
-      flex: 1 1 auto
+      flex: 1 1 auto // Fix for IE 11. Revisit or remove when IE11 is no longer supported.
 
   .subject
     border-radius: 5px
--- a/css/subject.styl
+++ b/css/subject.styl
@@ -1,6 +1,9 @@
 .subject-viewer
   min-width: 0
   max-width: 100%
+  
+  .default-root-style
+    display: block
 
   .subject-frame-pips
     .subject-frame-pip


eatyourgreens:
I've just done a couple of Whales As Individuals classifications in IE11 on Win8 and this branch seemed fine. @aliburchard are you okay with the black margins persisting in IE, but vanishing in Chrome, Firefox etc?

brian-c:
This is a seriously critical piece of the app, and if the 0,0 point goes off somehow, or the image display shifts from that point, classifications will be wrong.

I'd be in favor of massively simplifying the behavior of the subject view (especially with all the recent activity around side-by-side view and pan and zoom) before we start forking it by browser.

srallen:
I could set a max-width based on naturalWidth of the original source image for IE11 browsers too just for now. Previously, the subject viewer did not display at all; it had 0 width without specifying flex-grow to be 1. I agree, I'd rather not feature detect for it, but I wasn't sure of a faster way to get a fix out there.

Or we could ask folks not to use IE11?

eatyourgreens:
The top left of the image has been registering as (0,0) on the images I've tried marking. Ditto the bottom right corner shows the correct coordinates.

aliburchard:
I'm not sure I follow the issues driving the margins/implication of the current fix to weigh in effectively. I don't like having the subject viewer display differently in different browsers, and do worry what that means for classifications made in the different browsers. What does a "proper fix" entail vs a quick browser-specific quick fix?

eatyourgreens:
@aliburchard The classifications look the same to me in all the browsers I've checked, though I'd be more comfortable if someone more familiar with them checked that.

I'm also okay with waiting on merging this, since it affects the look of the site but not the actual marking data that's being gathered.

brian-c:
Off the top of my head, I recall the black bars are the background of the SVG. The SVG is sized to 100% the width and height of its container, which was an image tag (which would maintain its proportions when resized) but now includes the `Before` and `AfterSubject` hooks. I _think_ moving those hooks up out of the SVG's parent should be a fix for the black bars, but I haven't actually dug into it.

aliburchard:
I know that we do need a longer term fix that involves rewriting the subject viewer, but would be good to improve it in the meanwhile if that fix is a long way off. I'd like to confirm that the coordinates aren't affected though, and don't have much intuition with marking data.

I've marked a rectangle in the upper left corner using the main branch and this fix. Wondering if @vrooje @mschwamb or @mkosmala might have better intuition about whether those markings look right (based on the summary output)? (e.g. x = 1.2 vs 2.9 -- is that just how much I'm off by across the two classifications?)

screen shot 2016-03-22 at 15 23 11
screen shot 2016-03-22 at 15 23 03

eatyourgreens:
@brian-c 100% height/width got taken off in response to #2262 and #2285 The SVG is using the `subject` class now, so it should be sized using viewport units. I guess the problem is that it always expands to fill the width of its container, rather than sizing proportionally to the aspect ratio of the child `image` tag,

aliburchard:
I made a test project where I marked the same image in the main branch and the fix branch. I think the difference is just my inability to mark exactly?

screen shot 2016-03-22 at 15 49 27
screen shot 2016-03-22 at 15 45 51

mschwamb:
I don't have a drawing project. That's up to y'all to check. Even if the markings are okay, the views are different, and I'd say it has to be fixed ASAP. 

I might not be understanding what's going on, but here's my thoughts on this.

I'm more concerned that different people are seeing different things in the classifier which may influence behavior.  For example in my view and size of browser window on Firefox I don't have any black around Planet Four: Terrains, Comet hunters or the whales project, but @aliburchard  you seem to. 

If this all being caused by IE 11 and IE 11 is a small percentage of classifiers (which it's been since the start of Planet Hunters) why not stop IE support until this can be properly solved and make sure the viewer is working fine in the other three main browsers and stick up a message saying to switch to a modern brower if you have ie 11. This is what happened with Planet Hunters originally because IE was too buggy to handle the interface and it wasn't worth the time to fix it. A few people complained, it was fine as IE is used by an incredibly small fraction of Zooniverse participants. 

If this is being caused by the zoom feature, why not roll it back on the production version or make that experimental until there's a proper fix since there's no live project using that feature?

eatyourgreens:
@aliburchard looks good to me. I'm happy to merge.

eatyourgreens:
@mschwamb I don't think we can guarantee uniformity of look and feel across all browsers on mobile, tablets, desktop etc. That's just a feature of modern web development. We can work to minimise differences across browsers though.

aliburchard:
Great. Yup, I think we are essentially not fixing this for IE. 

I'm not 100% sure where this was introduced (whether it's related to the side-by-side image display (@itsravenous?) or pan & zoom (@simensta?) or something else. I think we all agree that this is a quick fix, and that a more complete solution is needed. It sounds like that complete solution will be a lot of work, and I don't know what the best approach is to tackle that.

eatyourgreens:
@aliburchard if we trace this back, we'll find scaleable subject images came in originally for the crop task, but there's been a lot of work since then. Definitely not a simple revert.

aliburchard:
aha! thanks @eatyourgreens.

eatyourgreens:
Also, I love the idea of having a test card image to check marking tasks after changes to the classifier.

aliburchard:
:) I'll make anyone collabs that needs to be (it's project # 962 and we can/should use it for testing all the things)

mschwamb:
@eatyourgreens agreed. I'm only asking for as consistent as is realistically possible. Sounds like this does that and that there's no simple way to revert back the changes that caused the issue in the first place, so this seems reasonable then.

 thanks to all involved for fixing this.

mkosmala:
@aliburchard I haven't started looking at my markings classifications yet. But @khufkens has written scripts to deal with the markings in Floating Forests. He may have advice/opinion here.

khufkens:
The black boarder existed in Google Chrome as well, I hope this does not influence the markings because this would be seriously messed up. I'll check my data later (when did this start?).

vrooje:
Oops, late to this, but @aliburchard your marks looked right to me in both cases. Should I be checking Galaxy Zoo Bar Lengths before/after a particular date to see if marks are shifted?

brian-c:
Rendering the interface from the classification data is direct, so there's very little chance we'll screw that up without it being immediately obvious to everybody. By design, the drawing tools themselves always assume that 0,0 is the top-left of the image, and its dimensions are the natural dimensions of the image. (There is some code to scale stroke thickness appropriately, but this is not affected by classification data).

Turning the mouse cursor position into classification data is where things get trickier, and in situations where we modify the image's 0,0 point (cropping and zooming) or display dimensions there is potential for the mouse coordinates to be offset from the image coordinates. [Here's something we can try](http://stackoverflow.com/a/10298843) to probably make this more robust.

But to reiterate: if the classification data is offset from the mouse position, it will be obvious. Marks always match the data, even if the mouse doesn't.

Issue #2248, opened 2016-03-01
Ability for Zooniverse staff project builders to use black-on-white text in project pages


alexbfree:
One of the tasks we identified for enabling the move of Snapshot Serengeti to Panoptes.

@aliburchard has specified that this is a must have for the move, as the current pages (being white on a photographic background) are too hard to read.

We need to either be able to force black text on a white background, or else have a very basic means to style pages. This doesn't need to be available in the project builder for all users.

@brian-c and @aliburchard and maybe @chrislintott probably need to discuss/agree the way forward for this.

marten:
I say we just change the default style for everyone, to either a flat white background or a slightly translucent white. I'd also like to put in a max-width on the content area while we're at it.

alexbfree:
good ideas Marten. We could also look at increasing line spacing to improve readability too.

alexbfree:
I suppose one thing about using the plain white style on all the project pages is, where do we draw the line? do we lose the "background image for a project" completely? Will there be a smooth transition from backgrounded project homepage to white project pages? I guess this ties into #2249 which Grant is going to discuss with Jordan.

marten:
We'd keep the background for the header and you'd still see it behind the homepage and the classify page.

aliburchard:
Definitely need to keep the background for landing pages! Not sure about the classify pages, since it's often cluttered (great one for @jwbmartin and @mrniaboc).

For the text/about pages: maybe slightly translucent white background with max width over...the image? OR just white all around like the talk pages.

jwbmartin:
Hey all as far as project builder things I have been playing around with some things. I have a few rough ideas to increase legibility below. Let me know what you think.

![classify_page-01](https://cloud.githubusercontent.com/assets/14332849/13467429/dc10dc06-e063-11e5-9c09-a9d4d4177af9.png)

![2b833914-b47c-11e5-8018-8f7f77873894](https://cloud.githubusercontent.com/assets/14332849/13467432/de7bbba0-e063-11e5-9c92-57cac8d66c01.JPG)

alexbfree:
@jwbmartin Nice work! I especially like the first screenshot! 

Only one small tweak i would make, the Option buttons have the same grey as the background grey - maybe something brighter or darker would make them stand out more?

[...]

PR #2331, opened 2016-03-22
Make sure the background behind text is really dark

marten:
Main goal for this is to enhance readability of text and the make images stand out better. I know we're working towards a larger redesign, but perhaps we can take this small step in the meantime?

/cc @aliburchard @alexbfree @jwbmartin 

# Before

screenshot_05

# After

screenshot_06

### DIFF
--- a/css/project-page.styl
+++ b/css/project-page.styl
@@ -140,6 +140,9 @@
   font-size: .75em
   color: #CDCDCD
 
+.project-text-content
+  background: rgba(0,0,0,0.8)
+  padding: 1.5em 3vw
 
 .project-metadata
   padding: 1.5em 3vw


aliburchard:
This is pretty awesome. I know a redesign is on the horizon, but this would go a long long way to making projects more readable in the meanwhile (e.g. see below)! Thanks @marten!

screen shot 2016-03-22 at 12 01 33

alexbfree:
It looks good, but I would like to propose an alternative...

```
.project-text-content {
    background: rgba(255,255,255,0.35);
    padding: 1.5em 3vw;
    color: #000000;
    border-radius: 3px;
}
```

looks like this. 

screenshot 2016-03-24 14 01 47

I think black on white (or a translucent white) is more readable, and white-on-black is a bit of a niche taste (geeks & devs etc) whereas black-on-white is more mainstream.

Also I think a very slight curving of the corners helps the right angles be less jarring, without introducing a new "curved" style.

CKrawczyk:
+1 for dark text on light background (actually better for you eyes in the long run).

srallen:
Pinging @jwbmartin since project builder projects redesign is something she's working on.

jwbmartin:
Looking much better with the white on black! I think the padding could be more though just so the line length of text isn't quite so long and maybe even taking the alpha up to 0.8 or 0.9 just to make it even more legible.

jwbmartin:
I realize now what I said may be confusing so what I mean is I like the white background with black text...

marten:
I'll try and find some time to update this this afternoon.

marten:
I had a look at this yesterday, and going from dark to light background has many more things that will need to change (for instance, the "OWNER" label has bright blue font color, which won't be readable). I propose we stick to the dark BG for now, and postpone that switch to another PR. I'm going to push a commit with tweaked padding though.

alexbfree:
I don't think that's the right way to go.. If we are all agreeing that white background is better we should take the time to do that. I can help with the implementation - should be pretty quick if it's just CSS colours etc. Can you enumerate the list of things that will also need changing if we go for white background? So we can get a sense of how much work it is?

srallen:
I agree with @marten to do a small change for now. The projects will have their styling overhauled by @jwbmartin so it makes sense for big changes to wait for those designs.

alexbfree:
OK I guess I am motivated by knowing we need to do it for #2248. @aliburchard @jwbmartin care to weigh in?

aliburchard:
+1 to @srallen and @marten suggestions. In the interest of making the pages readable for the current projects, let's make marten's changes as a quick and easy improvement, and then implement @jwbmartin's big changes when they're ready.

alexbfree:
i'm a bit confused. Admittedly I don't know what "big changes" @jwbmartin has in store, but changing to a white background and fixing a few labels that don't look right is still a very small change, not a big change, just as Marten's black background change is.

I just think it might look a bit schizophrenic to users if we change the background to black, only to change it to white some days later.... Happy to provide a small PR now to do the change to white background if that helps resolve this!

aliburchard:
I think this is turning into something much more complicated than it should be. Marten's fix is a very, very small change from what exists that makes the pages a lot more readable. It's not a big change from what currently exists (already white on black), and it's 100% ready to merge. It's meant to be temporary until we can implement @jwbmartin's complete overhaul. But my understanding is that Jordan's proper overhaul is still weeks (maybe months?) from being implemented. Could we not in the meanwhile implement this very small change that makes the project pages much more readable?

marten:
I'm fine with @alexbfree implementing a white BG, we might want to check that no projects have any transparent images that'll be unreadable on a white/light background.

alexbfree:
OK just spoke to Ali - clarified that the white bg change is also tiny and can be done right now, she has agreed we can go ahead with the white bg change for now, and wait for jordan's change later.

I will create a separate PR, link it here, then close this one.

And good point to check images. I'll check the pages for all the public projects. (FWIW I think that check would be needed with black background too)

alexbfree:
OK here we go. Closing this PR > let's move to https://github.com/zooniverse/Panoptes-Front-End/pull/2374

Thanks @marten for kicking this off!

Issue #2326, opened 2016-03-18
simple stats page


mschwamb:
I found for checking whether Planet Four: Terrains was close to running out of data, I was calling for the subjects and annotations csv file when  I wanted the number of subjects completed in the subject set and how many classifications per day.  I think it would be really handy to have some stats like Ouroboros projects have without having to download the subject csvs and classifications. 

What about having a page in the project builder that had stats that ran once a day/ looked at current project counters?  Number of registered users for the project, and then for each work flow: the number of total classifications and number of classifications per day for the past N days. For each subject set: whether its live or not, how many subjects are in there and how subjects are retired and % completion for each workflow. 

The Ouroboros stats graphs are really nice, but even being able to get the classifications numbers per day would be helpful for managing a project and help new science teams figure out how their project is progressing. I could see this for now being a query that runs once a day and stores the number is structure with those values being reported on the stats page. 

cc @aliburchard , @mrniaboc , and @chrislintott

camallen:
this has been on the radar for a while, @CKrawczyk any progress on this?

CKrawczyk:
I have been getting new images ready for GZ and this project has fallen on the back burner.  I will pick it back up next week.  Some more work needs to go into presenting the graphs for long running project (the hourly view for becomes difficult to navigate).  I also want to add the plotted time interval as a URL parameter for easier sharing of stats information.  To give a sneak peak:

![screen shot 2016-03-18 at 2 43 57 pm](https://cloud.githubusercontent.com/assets/6557526/13881232/ef98fb30-ed17-11e5-9fef-d8b0246e1d87.png)

Note: the big gap in the plot from mid-Jan to the start of Feb is due to the stats server not collecting classification data (not an issue with the main site).

mschwamb:
Looks really nice and super handy. I hope part or all of this is implemented in the coming weeks. 

 @CKrawczyk  one suggestion. It would be nice if it you could also report or have a drop  down menu for the builder to be able to check the percentage complete and needed classifications per subject set as well for the top circle graphs. It would be nice to check and see if we're through 20% of the new data just uploaded to the site. The numbers and graphs are useful for the full dataset, but if I upload new subjects in preparation for being out of data, it seems like that circle plot would change and the percentage complete would go down.

aliburchard:
!!! omg @CKrawczyk this is so freaking exciting. You can't tease with us pictures of such awesomeness. That's like making kids wait for Christmas presents.

[...]

PR #2333, opened 2016-03-22
Public project dashboard

CKrawczyk:
This branch is dependent on the configuration of the stats server in this PR: https://github.com/zooniverse/panoptes-javascript-client/pull/29

This adds a public stats page for all PFE projects (will close https://github.com/zooniverse/Panoptes-Front-End/issues/2326).  It provides stats for each workflow and the number of classifications/talk posts per hour, day, week, or month.

Things left to do:
- [x] Add URL prams to indicate the range being viewed on the bar graphs (that way people can link to a particular time range)
- [x] Add a link to the stats page someplace on the project page (located with the counters on the landing page).
- [x] Add and ETA based on current classification rate (or maybe add code to model a "GMB"s exponential falloff)
- [x] Allow bar charts to be filtered by workflow

Note 1: the stats server only collects data for the projects on production, to access this page make sure `?env=production` is in the URL.  E.g. `http://localhost:3735/projects/zooniverse/pulsar-hunters/stats/?env=production`

Note 2: the gap in the classification data from Jan 13 to Feb 7 was due to an error in the stats server, not an issue on the main site.

Note 3: navigating the bar graphs for long running projects when viewed by hour is not the smoothest thing in the world, but I think it is good enough for now.


mschwamb:
If this is going to add a stats link to the top navigation bar can the padding be reduced between the elements to fit more on in one line. If you've got two external links even with short titles for most reasonable sizes of the browser on my 13inch screen for both of my live projects the navigation bar it barely fits on one line. If I add any more characters to planet four: Terrains site guide link beyond 'Guide' it pushes the blog link to a second row and looks awkward.

CKrawczyk:
@mschwamb the more I think about it the less I like the idea of putting the link in the top bar.  It might be more at home on the project's landing page near the current counters.

mschwamb:
@CKrawczyk FWIW I really like that idea

marten:
@CKrawczyk me too

package.json
```diff
@@ -19,9 +21,12 @@
     "markdownz": "~4.0.2",
     "modal-form": "~2.3.0",
     "moment": "~2.9.0",
-    "panoptes-client": "~2.0.0",
+    "panoptes-client": "git://github.com/CKrawczyk/panoptes-javascript-client.git#add-stats-server",
```
eatyourgreens: Panoptes client version should be 2.2 to pick up the stats server config.

app/pages/project/stats/stats.cjsx
```diff
@@ -0,0 +1,213 @@
...
+        <div>
+          {retirement}
+        </div>
+        <div>
+          <span className="progress-stats-label">Images retired:</span> {@props.workflow.retired_set_member_subjects_count.toLocaleString()} / {@props.workflow.subjects_count.toLocaleString()}
+        </div>
+        <div>
+          <span className="progress-stats-label">Classifications:</span> {@props.workflow.classifications_count.toLocaleString()} / {(@props.workflow.subjects_count * @props.workflow.retirement.options.count).toLocaleString()}
+        </div>
+        <PromiseRenderer promise={@statCount()}>{(statData) =>
```
eatyourgreens: We're trying to remove the `PromiseRenderer` component. See @brian-c's comments in the [PFE readme](https://github.com/zooniverse/Panoptes-Front-End/blob/master/README.md)

Better to store the stat data in component state and render when it updates.

package.json
brian-c: Why is the stats server defined in panoptes-client if it's not used there? Should an interface with the stats server interface be part of panoptes-client?

app/pages/project/stats/stats.cjsx
brian-c: Yep, and it looks like the `statCount` method(s) would probably be the right thing to pull out into the panoptes-client library.

package.json
CKrawczyk: Ah, forgot to change that back.  Thanks @eatyourgreens

app/pages/project/stats/stats.cjsx
```diff
@@ -0,0 +1,213 @@
...
+        if w is null
+          areNulls = true
+      if not areNulls
+        @setState({workflowsLoaded: true})
+
+  shouldComponentUpdate: (nextProps, nextState) ->
+    return (@props.by isnt nextProps.by) or (@props.workflowId isnt nextProps.workflowId) or (@state isnt nextState)
+
+  statCount: (period, type) ->
+    query = if @props.workflowId then "workflow_id=#{@props.workflowId}" else "project_id=#{@props.projectId}"
```
eatyourgreens: I'll have a look today at moving this into the Panoptes javascript client.

eatyourgreens:
@CKrawczyk @brian-c My first pass at a really simple stats client:
https://github.com/zooniverse/panoptes-javascript-client/blob/stats-client/lib/stats-client.js

I've tested it locally by replacing the `statCount` method with something like
```coffeescript
statsClient
      .update
        projectId: @props.projectId
        workflowId: @props.workflowId
      .statCount period, type
```

eatyourgreens:
@CKrawczyk here's a PR that removes the Promise Renderers and pulls the stats counts in from the API client. https://github.com/CKrawczyk/Panoptes-Front-End/pull/2

CKrawczyk:
@eatyourgreens I just added in your changes and hooked it up so the dropdown menus fetch the new stats when changed.

CKrawczyk:
The api calls have been updated to reflect the code changes in https://github.com/zooniverse/panoptes-javascript-client/pull/39.

eatyourgreens:
Great. I think all that's left now is to update `package.json` when the new version of the API client is published.

eatyourgreens:
Small style change in https://github.com/zooniverse/panoptes-javascript-client/commit/99739177ccf84d1348e8ffaf73b8fd2212d2959d The API params should be `projectID` and `workflowID`.

brian-c:
panoptes-client 2.3.0 has been published with the stats client.

CKrawczyk:
OK, changes made.  I also pushed latest changes to [preview](https://project-dashboard-rebase.pfe-preview.zooniverse.org/?env=production) and everything seems to be in order.

package.json
```diff
@@ -5,12 +5,14 @@
...
     "babelify": "~6.4.0",
     "classnames": "~2.2.3",
+    "chartist": "~0.9.5",
     "counterpart": "~0.17.0",
     "data-uri-to-blob": "0.0.4",
     "debounce": "~1.0.0",
     "drag-reorderable": "~0.1.0",
     "fast-levenshtein": "~1.0.7",
     "history": "~1.17.0",
+    "json-api-client": "~3.1.0",
```
eatyourgreens: This can go now that we've moved the API call to the Panoptes client.

eatyourgreens:
Excellent. If you can remove `json-api-client` from `package.json`, I think this is ready to go.

CKrawczyk:
Do you think I should add a footnote in about the gaps in the stats data explaining that they were caused by a bug in our event stream and not issues with the main site?  I get the feeling if this is not included users might mistakenly think classifications were lost.

Since the gaps are during known times I can make sure the footnote only shows up if the project was launched before either of the gaps started.

CKrawczyk:
@eatyourgreens I added in the following footnotes to inform users about the gaps in the stats data (they only show up if the project launched before these gaps happened):
> †The gap in the classification data from Jan-13-2016 to Feb-07-2016 was caused a bug in our event notification system. **No** classifications were lost in this time.

>‡The gap in the talk data from Feb-18-2016 to Apr-07-2016 was caused a bug in our event notification system. **No** talk comments were lost in this time.

Do these look reasonable?

@brian-c I noticed you reverted the version bump of the panoptes client, since this branch is dependent on that version should we hold off merging this until the issues is resolved?

brian-c:
Yes, please wait, should be fixed again today.

eatyourgreens:
Looks like this is good to go now.

Issue #2337, opened 2016-03-22
Suddenly, subjects are not displaying in Talk


mkosmala:
Subjects normally display fine in the Notes section of Season Spotter Questions Talk. But today they're not there:
![talk_no_image](https://cloud.githubusercontent.com/assets/1066340/13966241/ce475fde-f047-11e5-80d7-75de28150449.png)

No subject is. Even old posts (say, > 20 days) that previously had subjects displaying no longer display subjects.

Note that the images still show up in the table of contents. Just not in the posts themselves:
![talk_index](https://cloud.githubusercontent.com/assets/1066340/13966351/5a6ae45e-f048-11e5-82ad-932d75c9a5d3.png)

eatyourgreens:
This was fixed by #2338

PR #2338, opened 2016-03-22
A workflow isn't passed into the subject viewer for talk

brian-c:
Quick fix for #2337.

More generally, design-wise: the SubjectViewer shouldn't be aware of a workflow, that's the SubjectAnnotator's job. Might look into this in the future.


eatyourgreens:
Looks good to me.

PR #2341, opened 2016-03-23
Paginate /admin/project_status via the response metadata.

parrish:
Sorry for the delay, but #2280's root cause is project-status-list attempting to use a url param that doesn't exist rather than an issue with the paginator.


wgranger:
Wow. Yeah, that seemed to do it. I guess the meta data was the main source of this problem.

PR #2345, opened 2016-03-24
Fix the title and numbering for recent Planet Hunters paper. 

willettk:
Boyajian+15 was initially submitted to the arXiv as "Planet Hunters X", when it should have been "Planet Hunters IX" (only the ninth in the series). I also changed the link to point to the correct MNRAS version of the paper (which is open access), rather than the arXiv preprint.


marten:
Content update I can't really say anything about. @willettk @simensta can we push this PR along?

simensta:
Definitely -- looks good. The changes in this PR duplicate the changes in #2352, so I will close #2352 after merging this PR.

PR #2349, opened 2016-03-29
Remove Tooltip class

brian-c:
Replaces the last Tooltip instance with a TriggeredModalForms to keep things more consistent and remove the dependency on tether.


eatyourgreens:
Looks good to me. Now that state has been removed from `WarningBanner` should it use the new syntax for stateless components?

parrish:
@eatyourgreens I imagine there are a lot of components that could be rewritten.  If we want to change syntax, it's probably best to have a separate PR for that.

PR #2358, opened 2016-03-31
Stateless component for WarningBanner

eatyourgreens:
Trying out the new syntax for stateless components in React 0.14.


simensta:
Looks good, thanks @eatyourgreens!

PR #2370, opened 2016-04-01
[project page branch] Improve the sort labels on project page

alexbfree:
I am proposing this change so that we improve the sort labels to have consistent capitalisation, and some more meaningful names / useful sorts to users:

Specifically: 

* "Participants" => "Most Popular"
* "last updated" => "Most Recently Updated"
* "Launch Date" => "Most Recently Launched"
* "Activity" => "Most Active"
* "Completeness" => Inverted so that least complete not most complete is first, and changed to "Most Help Needed"
* "Default" => removed because it's not useful to anyone * 

also

* "Sort By" => "Sort by" (since it looks weird with capital B)

Also I noticed there was a "project-sorts" include file that was not being used, so I put the updated values into there and ensured the include file is loaded instead of having a duplicate declaration.

*Note: the default sort is still active when you first arrive, we probably should just change the API to choose a more useful default sort order - then change this page to deliberate select something useful such as "most help needed" or "most recently launched" when you first arrive, and have that sort preselected


srallen:
@alexbfree Sorry I haven't looked at this until now, but it looks like a rebase is needed before it can be merged.

alexbfree:
Looks like this PR has got messed up as the base has changed since i created it.

Note to self. The only changes we need are https://github.com/alexbfree/Panoptes-Front-End/compare/0d454b96d7c6a2903b3404bdce57dc98f2dd7c7e...alexbfree:1f84f10fa70d31e68dbf4506c1f47138c5802f2f . I will create a new branch and PR with just those and close this one.

alexbfree:
Superseded by #2417

PR #2379, opened 2016-04-04
allow admin subject limit field to be updatable

camallen:
use defaultValue to set the initial value

### DIFF
--- a/app/pages/admin/user-settings.cjsx
+++ b/app/pages/admin/user-settings.cjsx
@@ -96,7 +96,7 @@ UserLimitToggle = React.createClass
     <div>
       <AutoSave resource={@props.editUser}>
         Subject Limit:{' '}
-        <input type="number" name="subject_limit" min="1" ref="subjectLimit" defaultValue={@props.editUser.max_subjects} onBlur={@updateLimit} onChange={handleInputChange.bind @props.editUser} />
+        <input type="number" min="1" ref="subjectLimit" defaultValue={@props.editUser.subject_limit} onBlur={@updateLimit} onChange={handleInputChange.bind @props.editUser} />
         <span>{ @errorMessage() }</span>
       </AutoSave>
     </div>
--- a/app/pages/admin/user-settings.cjsx
+++ b/app/pages/admin/user-settings.cjsx
@@ -96,7 +96,7 @@ UserLimitToggle = React.createClass
     <div>
       <AutoSave resource={@props.editUser}>
         Subject Limit:{' '}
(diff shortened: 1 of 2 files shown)


brian-c:
Looks like there's a conflict between `name="subject_limit"`and `value={@props.editUser.max_subjects}`. If `max_subjects` is what gets updated, the `name` attribute is wrong.

camallen:
the api attribute is subject_limit and the serialized api_user field is max_subjects [ala here](https://github.com/zooniverse/Panoptes/blob/686bcabf5a7494514e9993d496a2e455bec3cf3c/app/serializers/user_serializer.rb#L19) - looks like some naming was changed but the serialized attribute wasn't in the api - can update if needed.

camallen:
updated but is waiting on https://github.com/zooniverse/Panoptes/pull/1771

brian-c:
Cool, thanks. Only remaining concern is `onBlur` and on`Change` both modifying the user, but differently. I think onBlur is unnecessary.

camallen:
I think it was updating incrementally (mouse wheel / up + down keys on that input) without the differentiation (it's been a while). So instead of running the save it updated the state and when you lost focus it would run the save...feel free to change it up.

camallen:
@brian-c the api change to support this are ready to be deployed today - remind me to get this out.

app/pages/admin/user-settings.cjsx
```diff
@@ -96,7 +96,7 @@ UserLimitToggle = React.createClass
     <div>
       <AutoSave resource={@props.editUser}>
         Subject Limit:{' '}
-        <input type="number" name="subject_limit" min="1" ref="subjectLimit" value={@props.editUser.max_subjects} onBlur={@updateLimit} onChange={handleInputChange.bind @props.editUser} />
+        <input type="number" min="1" ref="subjectLimit" defaultValue={@props.editUser.subject_limit} onBlur={@updateLimit} onChange={handleInputChange.bind @props.editUser} />
```
brian-c: This input needs a name.

I recommend using `value` instead of `defaultValue` and removing the `onBlur` handler which conflicts with the AutoSave parent.

camallen:
is this better?

brian-c:
Not sure what happened here. Can we use `value` instead of `defaultValue`? `defaultValue` is for when you want the input to manage its own state, and we're using it to reflect the state of the user resource.

camallen:
updated - so me using the convoluted event handlers to update state was the weirdness with the value update originally. This should be good to go now.

brian-c:
Perfect, thanks!

PR #2374, opened 2016-04-01
"Quick fix" to improve readability

alexbfree:
Based on https://github.com/zooniverse/Panoptes-Front-End/pull/2331, which this PR replaces

Key features:

* White background (85% opacity) and black text on all project pages
* Small padding (5px) above and below embedded images
* Change owner/collaborator to tags to be readable on white background (slight increase to padding, change colours)

I have reviewed this with all project pages on all published-to-the-homepage public project builder projects, all seem readable.

## Before (FAQ with images)

screenshot 2016-04-01 17 56 15

## After (FAQ with images)

screenshot 2016-04-01 17 56 47

## Before (Research Page with Team Role markers)

screenshot 2016-04-01 17 56 24

## After (Research Page with Team Role markers)

screenshot 2016-04-01 17 56 58

(Happy to change colours of role markers if anyone has better suggestions!)

PR #2381, opened 2016-04-04
Make links on the white page contents slightly darker.

marten:
Following up on #2374 I found the links slightly to light. I've switched them over to the same colour as the owner tag in the sidebar.

## Before:

image

## After:

image

### DIFF
--- a/css/project-page.styl
+++ b/css/project-page.styl
@@ -152,7 +152,10 @@
     padding: 0.1em 0.4em
 
     &.owner
-      background-color: #1277fd
+      background-color: #0072ff
+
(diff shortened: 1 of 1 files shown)


rogerhutchings:
LGTM, let's get @jwbmartin's blessing first though

marten:
@rogerhutchings I've now switched both the owner tag and the links over to the same colour as used in Talk for links there. Still LGTY?

eatyourgreens:
:+1:

alexbfree:
Good move, I think that looks much better!

PR #2383, opened 2016-04-04
Fix survey display

### LINKED ISSUE #2380: Survey task - choices display suddenly misbehaving
@amyrebecca @brian-c This is a new issue, which I imagine was introduced with vertical sorting, but the thumbnails are now displaying way too early on the survey tasks, meaning that the text is often split across two lines for some choices and not others. I assume there's normally some threshold at which point the icons display, say, when <15 species are available.  Now they are displaying when way more species are available (for western shield and camera trap test, this is before anything has even been filtered.

screen shot 2016-04-04 at 14 43 09

screen shot 2016-04-04 at 14 46 18

This is a new issue that only appeared in the last few days. I think it might also be compounded by the subjects appearing massive again, thus limiting the amount of screen real estate available to the choices list itself.   For example, on WWW, we see the text split across lines and the subjects enormous from the start
screen shot 2016-04-04 at 14 48 17

and then displaying icons with too many choices still left exacerbates this
screen shot 2016-04-04 at 14 51 21

brian-c:
Closes #2380:

- Widens the task to 60 characters so items in Wisconsin Wildlife Watch don't wrap

- Thumbnails only appear when there are three columns

- Thumbnail aspect ratio is fixed to 1.25.

CC: @aliburchard

### DIFF
--- a/app/classifier/tasks/survey/chooser.cjsx
+++ b/app/classifier/tasks/survey/chooser.cjsx
@@ -29,7 +29,7 @@ module.exports = React.createClass
       'large'
     else if length <= 10
       'medium'
-    else if length <= 30
+    else if length <= 20
       'small'
     else
       'none'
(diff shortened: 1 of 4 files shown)


eatyourgreens:
:+1: looks good to me

aliburchard:
Thanks!!

Issue #2132, opened 2016-01-21
Survey task: focus is lost when selecting an animal with the keyboard


eatyourgreens:
Tab through the classification interface until you've selected an animal, then select that choice by pressing Enter. Focus jumps back out to the browser window and you have to tab back through the whole interface to interact with the next screen.

Keyboard focus should move to the survey task choice component when it mounts after being selected from the previous screen.

eatyourgreens:
Closed by #3563

PR #2384, opened 2016-04-04
autofocus for text tool fixes #2377

### LINKED ISSUE #2377: autofocus for text tasks
When we have a text task, can we have it "autofocus" on the text box - would save people from clicking. (I've created a project with just one task - entering text. Having to click each time really kills the rhythm.)

simensta:
This PR adds an autofocus property to the text tool, as suggested by @ggdhines in #2377. @mcbouslog, this will be relevant for the NfN project. Thoughts?

### DIFF
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -1,4 +1,5 @@
 React = require 'react'
+ReactDOM = require 'react-dom'
 {Markdown} = require 'markdownz'
 GenericTask = require '../generic'
 GenericTaskEditor = require '../generic-editor'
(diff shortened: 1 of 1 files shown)


eatyourgreens:
I'm not sure about autofocus on first load. Another component, such as the tutorial, may have focus and you won't want to steal it. 

It does make sense to have the task area focus when a new task loads after pressing the Done button. I added some logic to readymade, to do this for orchidobservers.org and sciencegossip.org. I'm not 100% sure it's applicable here but maybe worth looking at?

mcbouslog:
definitely relevant to NfN, thank you for doing!

i did a quick test locally, and it looks like the tutorial is overriding the textarea autofocus (even after closing the tutorial, in which case the textarea would have to be clicked in). have you guys had the same experience? if so, i think that's ok then?

other items i noted:
1. for text tool used as subtask in drawing task, this pulls cursor to textarea, regardless where textarea is in order of subtasks (i.e. if textarea third subtask, pulls cursor there, see Dev Classifier - Drawing Task - Circle)
2. does 1. with combo task as well

i'm trying to think of potential solutions to items noted, and will post if come up with anything.

otherwise this doesn't really matter, but if we're in the task tinkering i think with ReactDOM we can shorten the value capture to `value = @refs.textInput.value`.

eatyourgreens:
One other thought, only one element on the page can have autofocus, which limits the use of this component. See the notes on `autofocus` here. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input

eatyourgreens:
For reference, here's the autofocus behaviour I came up with for classifying quickly on readymade projects. https://github.com/zooniverse/zooniverse-readymade/commit/90a5c06cc9e3a113c6528317702291f6d848bebb

Autofocus when components mount is an issue for other tasks. See #2132

simensta:
Thanks for the feedback and resources! It seems to me that we need to implement focusing the task area when a new task loads after pressing the Done button, before implementing text tool specific focus. I think this will help resolve some of the order-of-focus issues @mcbouslog noted above.

@eatyourgreens, I'll start a new branch to add the behavior you cited from readymade. In the meantime, I will close this PR.

Issue #2256, opened 2016-03-01
update help text for survey task questions.csv


aliburchard:
Now that the survey tasks are combined, we need to update the help text to reflect the addition of include and exclude columns, and to reflect that the csv accepts or requires ; instead of c between possible answers.

*not sure how the "question sets" for making "one of these questions" required is handled; I'm okay leaving that out of the docs if we want to keep it as a hidden feature because it's really confusing.

aliburchard:
@amyrebecca (or maybe @brian-c ) can you also please clarify the additional changes made to formatting for these csvs? We're having a lot of issues with projects trying to launch. It looks like the reference photos for the "image" column now need semicolons as well, but maybe also the headers for the characteristics file?

Are there any other changes to the required formatting?

aliburchard:
@chrislintott good news is that this seems to be the only lingering *incorrect* in-line help text on the editor because @marten very kindly helped me edit the markdown for the others.  

@mcbouslog - we'll probably want to revisit some of the help text with the fancy new editor -- wondering if you'd be game to tackle this in the process? (Maybe we can do it all together when I'm in Chicago next month? Because I'm not actually leaving, no matter how hard I try...)

mcbouslog:
Definitely want revisit the help text, and I'd be happy to help! If I can't get to sooner, while you're in Chicago would be a great time to work on together.

aliburchard:
sweeeeet!!

PR #2387, opened 2016-04-05
Update help for choices.csv

### LINKED ISSUE #2368: Correct the help text for camera survey characteristics.csv file
people are having a lot of problems uploading their image icons because they think they need to specify the icon URL (which is indicated in the instructions).
screen shot 2016-04-01 at 10 33 09

@amyrebecca or @brian-c could one of you please update that help pop-up to indicate that the filename should be "filename.jpg" and not "http//......"

aliburchard:
fixes #2368 , #2256

### DIFF
--- a/app/classifier/tasks/survey/editor-help.cjsx
+++ b/app/classifier/tasks/survey/editor-help.cjsx
@@ -39,20 +39,20 @@ module.exports =
   </div>
 
   characteristics: <div className="content-container">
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/survey/editor-help.cjsx
```diff
@@ -10,21 +10,25 @@ module.exports =
           <th>Name</th>
           <th>Description</th>
           <th>Images</th>
+          <th>No auestions</th>
```
eatyourgreens: Typo here?

PR #2389, opened 2016-04-05
Fix emoji sizes in Talk comments

eatyourgreens:
Reverts a change from #2374 which has broken the display of smileys in Talk.
cc. @camallen

### DIFF
--- a/css/project-page.styl
+++ b/css/project-page.styl
@@ -154,9 +154,6 @@
     &.owner
       background-color: #1277fd
 
-  img
-    padding: 5px 0
-
 .project-metadata
   padding: 1.5em 3vw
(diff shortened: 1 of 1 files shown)


camallen:
so this looks find to me but we lose @alexbfree padding on the project images. Should we be styling talk images with the compounded project img css?

eatyourgreens:
I couldn't see a way to distinguish between the two classes of image, but maybe someone more familiar with Talk would know.

I'm happy to merge this and come back to image layout as part of any redesign of these pages, anyway.

camallen:
LGTM - reverting the padding won't break anything else from my understanding.

PR #2399, opened 2016-04-07
Workflow tutorials

brian-c:
Tutorials are now looked up by workflow first and fall back to lookup by project. (@marten, just to confirm:  getting `/tutorials?workflow_id=123` and `/tutorials?project_id=234` will both work and will continue to in the future, correct?)

Completed tutorials are now recorded by tutorial ID instead of project ID.

Related to zooniverse/Panoptes#1747. That deploy and this merge should probably be at the same time.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -40,11 +40,13 @@ Classifier = React.createClass
   componentDidMount: ->
     @loadSubject @props.subject
     @prepareToClassify @props.classification
-    Tutorial.startIfNecessary @props.user, @props.project
+    {workflow, project, user} = @props
+    Tutorial.startIfNecessary {workflow, project, user}
 
   componentWillReceiveProps: (nextProps) ->
     if nextProps.project isnt @props.project or nextProps.user isnt @props.user
-      Tutorial.startIfNecessary nextProps.user, nextProps.project
+      {workflow, project, user} = nextProps
+      Tutorial.startIfNecessary {workflow, project, user}
     if nextProps.subject isnt @props.subject
       @loadSubject subject
     if nextProps.classification isnt @props.classification
@@ -178,7 +180,7 @@ Classifier = React.createClass
       <p>
         <small>
           <strong>
-            <TutorialButton className="minor-button" user={@props.user} project={@props.project} title="Project tutorial" aria-label="Show the project tutorial" style={marginTop: '2em'}>
+            <TutorialButton className="minor-button" user={@props.user} workflow={@props.workflow} project={@props.project} title="Project tutorial" aria-label="Show the project tutorial" style={marginTop: '2em'}>
               Show the project tutorial
             </TutorialButton>
           </strong>
--- a/app/classifier/tutorial-button.cjsx
+++ b/app/classifier/tutorial-button.cjsx
@@ -4,27 +4,28 @@ Tutorial = require '../lib/tutorial'
 
 module.exports = React.createClass
   getDefaultProps: ->
-    user: null
+    workflow: null
     project: null
+    user: null
 
   getInitialState: ->
     tutorial: null
 
   componentDidMount: ->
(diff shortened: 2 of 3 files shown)


marten:
Yep, both those URLs will keep working.

I assume this'll change further to support multiple kinds of tutorials later on? In theory `/tutorials?workflow_id=123` could return multiple tutorials with different `kind` values.

marten:
Actually, I think `/tutorials?project_id=123` will just return a list of *all* tutorials for that project across workflows, since we've associated all existing tutorials with all workflows in the backend. You shouldn't need to check for a generic non-workflow tutorial really.

brian-c:
> I assume this'll change further to support multiple kinds of tutorials later on? In theory /tutorials?workflow_id=123 could return multiple tutorials with different kind values.

Ah, I forgot about `kind`. Would `/tutorials?workflow_id=123&kind=classifier` work for normal classifier tutorials? I'll set that on creation, although it seems `kind` currently is not currently updatable: `{"schema"=>"contains additional properties [\"kind\"] outside of the schema when none are allowed"}`

> we've associated all existing tutorials with all workflows in the backend

Currently, new tutorials are still associated with the project on creation. I think this is still the desired behavior from a product design perspective. Per-workflow tutorials are a special case we're implementing for a couple specific projects right now.

srallen:
This looks good and is fetching by workflow, but the editor still needs rework to be able to create tutorials by workflow. I started some work on that on the mini-course branch that I haven't pushed up yet. I can merge this now and finish up the editor with the mini-course, if that's good with you, @brian-c?

brian-c:
For now I'm thinking per-workflow tutorials will be created by hand. This functionality is not exposed to project builders. It may be in the future. (@chrislintott, @trouille, thoughts on this?)

marten:
I guess the UI for it would just be the same as we do for associating workflows to subject sets. Would that be reusable at all? If it's not much more work, I'd really prefer we push features to completion rather than people needing our help.

srallen:
@marten, currently subject sets are associated on the workflow editor page and I'm thinking now that may be the most straight forward way for now to associate tutorials/mini-courses. I was initially thinking it'd be in the tutorial editor, but I like your idea better.

I would also prefer to complete the feature now too.

trouille:
@srallen @marten I agree with you both on this:
1) associating tutorials/mini-courses to a workflow similar to the way subject sets are associated. 
2) completing the feature now to save us time in the long run

srallen:
@brian-c, @marten I rebased my mini-course branch against the workflow-tutorial branch and did a pass on how the editor could work in the [mini-course branch](https://github.com/zooniverse/Panoptes-Front-End/tree/mini-course) and deployed here: https://mini-course.pfe-preview.zooniverse.org

The editor has been refactored to allow the creation of multiple tutorials or mini-courses. When they're initially created the `kind` property is set, but is only associated to the project. Linking to workflows is in the workflow editor similar to subject sets. Currently I still need to work on:

- Getting the workflow link to work. I've modeled the functions after what's happening with the subject sets, but I'm getting an error: `POST https://panoptes-staging.zooniverse.org/api/workflows/1817/links/tutorials 404 (Not Found)`
- UI improvement with possibly adding a display name to the tutorials. This would just make it easier to know which tutorial or mini course you are linking in the workflow editor view. But maybe not necessary for now.
- Still need to get the mini course working in the project itself.
- Existing tutorials that still have `kind: null` will be displayed in both the mini course editor and the tutorial editor since the editor is shared. Two routes I'm thinking in how to deal with this: allow those with null values to only display in the tutorial editor since mini-courses are new or retroactively set the value in the existing tutorials.

Let me know if anyone wants to be added as a collaborator to the project I've been developing against.

srallen:
Closing this and workflow tutorials will be superseded by an upcoming pull request for mini-courses.

PR #2408, opened 2016-04-08
Quick hide marks toggle

saschaishikawa:
Supercedes #2304

### DIFF
--- a/app/classifier/tasks/drawing/hide-previous-marks-toggle.cjsx
+++ b/app/classifier/tasks/drawing/hide-previous-marks-toggle.cjsx
@@ -21,12 +21,17 @@ HidePreviousMarksToggle = React.createClass
     @props.onChange @props.classification
 
   render: ->
+    annotations = @props.classification.annotations
+    currentAnnotation = annotations[annotations.length-1]
+    return null unless @props.workflow.tasks[currentAnnotation.task].enableHidePrevMarks
+
+
     DrawingTaskComponent = require '.' # Circular
 
     currentlyHidingMarks = @props.classification._hideMarksBefore > 0
 
     marksCount = 0
-    @props.classification.annotations.forEach (annotation) =>
+    annotations.forEach (annotation) =>
       taskDescription = @props.workflow.tasks[annotation.task]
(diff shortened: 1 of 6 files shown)


app/classifier/tasks/drawing/hide-previous-marks-toggle.cjsx
```diff
@@ -21,12 +21,17 @@ HidePreviousMarksToggle = React.createClass
     @props.onChange @props.classification
 
   render: ->
+    annotations = @props.classification.annotations
+    currentAnnotation = annotations[annotations.length-1]
+    return null unless @props.workflow.tasks[currentAnnotation.task].enableHidePrevMarks
```
brian-c: We're back to only displaying this on the drawing task?

app/classifier/tasks/generic-editor.cjsx
```diff
@@ -57,6 +57,14 @@ module.exports = React.createClass
         </div>}
       {' '}
 
+      {if 'hide previous marks' in @props.project.experimental_tools
```
brian-c: I truly do not believe this should be optional. But if we're going to make it optional, and we persist the checkbox across tasks, this should be a property of the workflow. If it's only on the drawing task, this is fine.

saschaishikawa: By "property of the workflow", you mean it should be like `@props.workflow.enableHidePrevMarks` instead of `@props.task.enableHidePrevMarks`? If so, the back-end won't let me call `update()` on that.

brian-c:
Merging for use tomorrow. Some things to decide:

- Should the checkbox be available on all tasks once there are marks on the image? I'd argue yes, because answering questions might be easier if you can see the original image. If so, let's move the `enableHidePrevMarks` property from the task to the workflow configuration. If not, we should take the checkbox out of the persistent hook and put it in the task itself.

- Should this be optional for the project owner? I'll argue no, because I think tasks should be consistent and familiar across projects.

- When does stuff get out from behind the experimental flag? We've had this feature on other projects. I don't think it's necessary here.

Issue #841, opened 2015-06-26
Add button to clone workflows within (or across) projects


aliburchard:
Just needed to do this for whales to make a subtle change for different subject sets. Think it would be really useful for projects in general.

ezzjcw:
I would like to be able to do this too. At the moment I am doing it all manually. I feel greedy asking for it, but do you plan to add a clone/duplicate workflow button to Panoptes?

camallen:
This just needs a button to wire it up as the api already offers it, http://docs.panoptes.apiary.io/#reference/projects/project-create-links/add-a-link

> Workflow and Subjects added in this way will be copied and their id will be returned as part of the complete Project representation response.

srallen:
This is done.

PR #2411, opened 2016-04-11
add copy workflow button

mcbouslog:
issue #841 

not sure the workflow.cjsx lines 71 and 72 are necessary?
```
@props.project.uncacheLink 'workflows'
@props.project.uncacheLink 'subject_sets' # An "expert" subject set is automatically created with each workflow.
```

added to my ToDo list - wire original `New Workflow` button into new workflow.js action. holding off on that due to possible subsequent pull request related to allowing addition and editing of inactive workflows while project is live.


camallen:
This looks like overkill, the api allows you to do this already via the project add_links method, http://docs.panoptes.apiary.io/#reference/projects/project-create-links/add-a-link

> Workflow and Subjects sets added in this way will be copied and their id will be returned as part of the complete Project representation response.

The api will copy the workflow object completely here. Note: subjects sets will only copy over if not in the same project, see code [here](https://github.com/zooniverse/Panoptes/blob/a204079e00aa7e3b26fe9fa9e0d43160dee8b150/app/controllers/api/v1/projects_controller.rb#L209).

brian-c:
@camallen: The goal here is to create a new workflow with all the tasks and settings of an existing workflow in the same project. Are you saying we should post the new workflow data **without** `links: {project: '1'}` to `/projects/1/links/workflows` instead of posting it to `/workflows` **with** `links: {project: '1'}`?

@mcbouslog: I do think it's overkill to make tasks, configuration, and retirement rules optional in the dialog. Just ask for a new name and take the user to the editor, and they can make their changes there.

Also, how about hooking the dialog up to the "New workflow" button?

camallen:
@brian-c yeah - that linking code on projects should do a copy the workflow & contents and save it as a new one, fancy trying it out?

brian-c:
Sorry, am I supposed to post the copied data, or post something else (looks like maybe `{workflows: [originalID]}`?) and the back end copies it for me? What's the advantage over just creating a new resource?

camallen:
as per the docs, post to `api/project/id/links/workflows` with a payload of 
```
{
  "workflows": ["1"]
}
```
Up to you but leverage the code we've built (and tested) to do this and push the responsibility to the api to do the data wrangling for you (if indeed it does what you want / need).

brian-c:
Eh, I just posted `{"workflows": ["622"]}` to `/projects/231/links/workflows` intending to create a copy of workflow 622 in the project. Instead it deleted 622 and replaced it with a new one (2262). I don't think that should have happened.

Generally, I find working with links real confusing for anything but fetching related resources (and even there there are a few different ways to do it). Can we just use the standard verbs here?

camallen:
not sure what you mean by standard verbs, these are resource links. 

Looks like a bug in implementation, maybe the original intention was to copy workflows from existing project to new ones.

brian-c:
I just mean posting workflow data to the /workflows endpoint is easier to reason about than posting another workflow's ID to the /projects/id/links/workflows endpoint and getting a workflow back, then modifying it and putting it back to the /workflows/id endpoint.

camallen:
I'll have a look at this tomorrow on both counts, fixing the /projects copy and add as well as looking at adding the projects route links via workflows.

mcbouslog:
simplified the dialog, refactored the New Workflow button to include the dialog and workflowActions, and moved the create or copy logic to the action (though not sure that's a good idea, and could easily be moved back to dialog if so preferred). 

reasoning for moving create or copy logic to action was that with a little (not yet done) refactoring the dialog would be an agnostic get-a-new-title-and-save-or-copy component - maybe eventually useful for creating/copying tutorials, potentially other aspects.

mcbouslog:
I think I can clean this up a bit more by making the new component take in projectID, workflow (or other item to copy aspects from) and button styling, and move the action and show/hide/handleWorkflowDialog functions into the component as well. not sure that's worth doing though, regardless, i'll hold off until we reach a conclusion about how to handle things with the API? which i embarrassingly am trying to follow, but not sure i entirely understand the back and forth regarding.

mcbouslog:
@camallen, any update regarding the /projects copy and add, or adding the projects route links via workflows? or could we add this copy workflow button as is and refactor later to work with the API as you've noted, once the API is updated? 

of course no worries if no update or we should hold off, not trying to unnecessarily rush anything, just checking what i will or won't have available to demo to a few of the NfN science team members tomorrow, but this demo is absolutely no reason to rush anything (i'd hope they can figure out a copy button on their own if it's merged later!).

camallen:
i haven't gotten back to it, it's on my list though to fix the bug to make sure it doesn't kill existing workflows.

I'm not too phased if you all want to do this client side, we do have a feature for this on the api and if it's required for workflow copy links (e.g. not via a project) then maybe we should add it.

app/pages/lab/workflow-create-dialog.cjsx
```diff
@@ -0,0 +1,58 @@
...
+    workflow: null
+
+  getInitialState: ->
+    busy: false
+    error: null
+
+  handleSubmit: (e) ->
+    e.preventDefault()
+
+    if not @refs.newDisplayName.value
```
brian-c: Use `<input required />` when possible instead of checking for form values manually.

app/pages/lab/workflow-create-dialog.cjsx
```diff
@@ -0,0 +1,58 @@
...
+ReactDOM = require 'react-dom'
+counterpart = require 'counterpart'
+
+WorkflowCreateDialog = React.createClass
+  getDefaultProps: ->
+    onCancel: ->
+    onSubmit: ->
+    onSuccess: ->
+    projectID: ''
+    workflow: null
```
brian-c: This prop name is a bit confusing, Maybe call it `workflowToClone`?

app/pages/lab/actions/workflow.js
```diff
@@ -0,0 +1,30 @@
+var apiClient = require('panoptes-client/lib/api-client');
+
+var workflows = apiClient.type('workflows');
+
+var workflowActions = {
+  createWorkflow: function(workflowData, sourceWorkflow) {
```
brian-c: The "create" action shouldn't also be responsible for knowing how to copy something. Leave that to the component that accepts a workflow to copy.

app/pages/lab/workflow-create-dialog.cjsx
```diff
@@ -0,0 +1,58 @@
+React = require 'react'
+ReactDOM = require 'react-dom'
+counterpart = require 'counterpart'
+
+WorkflowCreateDialog = React.createClass
```
brian-c: This name is a little confusing; it's not a dialog, it's a form that gets put into a dialog elsewhere.

app/pages/lab/actions/workflow.js
brian-c: And this is subjective, but as much as possible I've tried to avoid UI components directly manipulating resources' `links` properties, since that's really just an implementation detail of how they're serialized. Maybe this signature could be `createWorkflowForProject(projectID, workflowData)`, and the action can wire up the link?

mcbouslog:
i think i addressed items noted, let me know what you think.

brian-c:
Looks good to me! Thanks.

PR #2412, opened 2016-04-11
Revert owned card styles to master

eatyourgreens:
@alexbfree commented that these files possibly shouldn't be changed by the project page PR.

Do we have different styles now for project cards and owned cards?

### DIFF
--- a/css/owned-card.styl
+++ b/css/owned-card.styl
@@ -1,18 +1,3 @@
-.discipline-section
-  h2.discipline-title
-    padding-right: 20px
-    padding-top: 20px
-    padding-bottom: 20px
-    padding-left: 25px
-    display: flex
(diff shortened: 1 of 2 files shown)


alexbfree:
@eatyourgreens I think @brian-c and @amyrebecca took the decision to separate owned cards (collections/favorites etc) from project cards. Perhaps they can add detail as to why/if they need to be separate.

eatyourgreens:
Incidentally, we've removed the changes to the owned cards markup in #2405, so I _think_ the CSS should stay unchanged too.

amyrebecca:
You're right, the owned-card markup shouldn't have changed at all. I think these changes were probably made before the branch had separated owned cards from project cards and just was never reverted.

PR #2418, opened 2016-04-12
[project-page branch] PR: Update all hover, focus and active states 

### LINKED ISSUE #2373: [project page branch] "Current Page" state not shown correctly
When we originally designed this, the current page state is supposed to be shown like this (in this example we are currently filtered to just look at History projects):

screenshot 2016-04-01 15 24 06

(This correct behaviour can still be seen on https://preview.zooniverse.org/panoptes-front-end/miniztm/#/projects and miniztm branch)

What we have now is this

screenshot 2016-04-01 15 23 07

Tasks:

1. Ensure All Disciplines is only blue background when it's active, white other times
2. Ensure each Discipline's colour stays when it's focussed, not just on hover
3. Remove the blue highlight square box around the panel on :focus

UPDATE:
On latest version at https://project-page.pfe-preview.zooniverse.org/?env=production, we see slightly different behaviour:

* Active elements are now keeping their colour, so task 2 no longer needed.
* All Disciplines is highlighting blue only when clicked, but not keeping its blue when active, which it should.

Re: the blue highlight square, see discussion with @eatyourgreens below, where I suggest an alternative solution.

alexbfree:
This fixes #2373. This also fixes an issue which was stopping "social science" from highlighting properly when it is the active filter ("social science"!="social-science").

@srallen once @jwbmartin has confirmed she is happy with https://project-page-states.pfe-preview.zooniverse.org/projects?env=production you can merge this.


css/disciplines-slider.styl
```diff
@@ -43,81 +49,133 @@ DISCIPLINE_COLOR_SOCIAL_SCIENCE = #7EA8CE
...
+    &:focus
+      color: DISCIPLINE_COLOR_PHYSICS
+      border-color: DISCIPLINE_COLOR_PHYSICS
 
   .filmstrip--disciplines__discipline-card--discipline-social-science
-    &:hover,
+    &:active,
     &.filmstrip--disciplines__discipline-card--active
-      background-color: DISCIPLINE_COLOR_SOCIAL_SCIENCE
+      background-color: DISCIPLINE_COLOR_SOCIAL_SCIENCE !important
```
srallen: Is this `!important` necessary?

css/disciplines-slider.styl
```diff
@@ -16,20 +16,26 @@ DISCIPLINE_COLOR_SOCIAL_SCIENCE = #7EA8CE
...
-    padding: .5em
+    padding: .2em .5em .5em
     color: #006DFF
     background-color: white
     min-width: 100px
-    &:hover,
+    &:active,
     &.filmstrip--disciplines__discipline-card--active
-      color: white
+      color: white !important
```
srallen: Is there any way to avoid this `!important`?

srallen:
Just a general convention, but the style props have been put into alphabetical order in our other stylus files. Could you do that here for consistency?

css/disciplines-slider.styl
alexbfree: Ah good spot, that was something I tried while debugging the social science issue that I left in by mistake. will remove.

css/disciplines-slider.styl
alexbfree: Hmm that's a tricky one. It was there because the active font colour has to override the hover/focus font colour, and it has the same specificity so the hover/focus colour wins.

Managed to remove it by moving this block to the end, but I am not sure if that is bad practice in a stylus file as the `.filmstrip--disciplines__discipline-card` is now repeated...

```
.filmstrip--disciplines__discipline-card
 &:active,
 &.filmstrip--disciplines__discipline-card--active
   color: white
   outline: none
```

alexbfree:
Have alphabetized everything in the file. Am assuming for alphabetization purposes that child elements always come after properties at that level even if alphabetically earlier, and also have assumed that for alphabetization purposes, `:`, `.` etc are ignored. Let me know if that is not the right convention.

alexbfree:
@srallen I think i have addressed all review comments now, so it's ready for you to take another look. Cheers!

css/disciplines-slider.styl
srallen: IMO, this is preferable. Thanks.

srallen:
As noted in #2373, Firefox still has its own focus border. This should fix it and scope it only to the filmstrip since we do not have focus styles defined for other buttons yet:

```
.filmstrip
	padding: 0
	display: flex
	margin-bottom: 1.5em
	max-width: 1200px
	margin: 0 auto

	button::-moz-focus-inner
		border: 0
```

For some reason, Firefox was ignoring the `border: 0` if it was on the `.filmstrip__nav-btn` class.

css/disciplines-slider.styl
alexbfree: ok glad you prefer it. thanks

alexbfree:
OK. good spot. I have a meeting to run to now but will try and fix that before tomorrow a.m.

jwbmartin:
Looking good for me!

srallen:
Double checked Safari and IE. Looks fine in IE. Safari still has blue outline, but up to the OS X system setting, so probably best to ignore for now unless Safari users complain.

Issue #2423, opened 2016-04-13
[project page branch] Proper handling of zero-result discipline filters


alexbfree:
As @srallen pointed out, when you click the Humanitarian filter there are no projects. There was one when we first coded this, but it has since been archived/removed.

Also you get an ugly output
![screenshot 2016-04-13 10 00 24](https://cloud.githubusercontent.com/assets/1473244/14488145/943db11c-015e-11e6-83fc-faaf5ff1db5c.png)

We need to design a better solution, which may include one or some combination of:

1. Clean up the display when there are no results.
2. When no results are found, remove the filter from the filmstrip and revert back to All Disciplines
3. Find some way to pre-emptively remove filters we know to be empty (if we can find a way to do this without running every filter query! Maybe add support at back end to provide a list of "currently active discipines"?)

Whatever solution we come up with should account for the fact that getting zero results could happen on any of the queries. Though we could possibly rely on hardcoding our knowledge of what exists in the short term.

alexbfree:
Note, the linked PR #2425 implements 1, but not 2 or 3, so we need to keep this ticket open after that PR is merged.

srallen:
For the sake of getting the project page done for now, probably the most simplest thing to do now is to comment out this line: https://github.com/zooniverse/Panoptes-Front-End/blob/project-page/app/components/disciplines.cjsx#L8

A more dynamic approach could come in a later PR.

alexbfree:
@srallen I agree with the pragmatic approach @srallen . Will do that now.

alexbfree:
Short term fix was committed in https://github.com/zooniverse/Panoptes-Front-End/pull/2432

Issue #2362, opened 2016-03-31
[project page branch] Clean up pagination when only 1 page


alexbfree:
We can probably do something a bit nicer than this:
![screenshot 2016-03-31 17 50 12](https://cloud.githubusercontent.com/assets/1473244/14183622/149e546a-f769-11e5-946e-dbbffa3a582f.png)

Either remove numbers entirely, or a simple "Page 1 of 1" text.

alexbfree:
could be fixed as part of #2361

alexbfree:
Fixed by #2425

PR #2425, opened 2016-04-13
[project page branch] PR: Pagination fixes

alexbfree:
To fully address #2362 and to partially address #2423

* Remove unused duplicate message translations
* Ensure uniform colouring and styling on both nav bars
* Improve pagination message
* Make pagination conditional - if no results, or only one page, don't show buttons.

### DIFF
--- a/app/components/project-card-list.cjsx
+++ b/app/components/project-card-list.cjsx
@@ -229,7 +229,7 @@ ProjectFilteringInterface = React.createClass
       .then (projects) =>
         pages = projects[0]?.getMeta()?.page_count
         project_count = projects[0]?.getMeta()?.count
-        project_count = 0 if !project_count
+        project_count ?= 0
         @setState {projects, pages, project_count}
       .catch (error) =>
         @setState {error}
--- a/app/components/project-card-list.cjsx
+++ b/app/components/project-card-list.cjsx
@@ -175,9 +175,10 @@ PageSelector = React.createClass
 
   render: ->
     <nav className="pagination">
-      {for page in [1..@props.total]
-        active = (page is +@props.current)
-        <button onClick={@handleChange.bind this, page} key={page} className="pill-button" style={border: "2px solid" if active}>{page}</button>}
+      {if @props.total>1
+         for page in [1..@props.total]
+           active = (page is +@props.current)
+           <button onClick={@handleChange.bind this, page} key={page} className="pill-button" style={border: "2px solid" if active}>{page}</button>}
     </nav>
 
 ProjectFilteringInterface = React.createClass
@@ -228,6 +229,7 @@ ProjectFilteringInterface = React.createClass
       .then (projects) =>
         pages = projects[0]?.getMeta()?.page_count
         project_count = projects[0]?.getMeta()?.count
+        project_count = 0 if !project_count
         @setState {projects, pages, project_count}
       .catch (error) =>
         @setState {error}
@@ -259,9 +261,13 @@ ProjectFilteringInterface = React.createClass
         <div className="resource-results-counter">
           <SearchSelector />
           <SortSelector value={@props.sort} onChange={@handleSortChange} />
-          {pageStart = @props.page * 20 - 20 + 1
-          pageEnd = Math.min(@props.page * 20, @state.project_count)
-          <p className="showing-with-link-para"><Translate pageStart={pageStart} pageEnd={pageEnd} count={@state.project_count} content="projectsPage.countMessage" /></p>}
+          {if @state.project_count>0
+             pageStart = @props.page * 20 - 20 + 1
+             pageEnd = Math.min(@props.page * 20, @state.project_count)
+             showingMessage = "projectsPage.countMessage"
+           else
+             showingMessage = "projectsPage.notFoundMessage"
+          <p className="showing-with-link-para"><Translate pageStart={pageStart} pageEnd={pageEnd} count={@state.project_count} content={showingMessage} /></p>}
           <PageSelector current={@props.page} total={@state.pages} onChange={@handlePageChange} />
         </div>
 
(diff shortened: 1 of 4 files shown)


css/project-list-pages.styl
```diff
@@ -43,6 +43,10 @@
     .pill-button
       margin-right: 10px
       padding: 1em
+      color: rgb(64,64,64)
+      font-weight: normal
+      border-color: rgb(64,64,64) !important
```
alexbfree: I'm aware the use of !important here is sub-optimal, but I spent over half an hour trying to get to the bottom of why the upper nav bar has different styles in effect, and couldn't quite figure it out. This works for now so would prefer to go with this for now, and if someone who understands what is going on here a bit better wants to fix it later, they can!

app/pages/projects-disciplines.cjsx
```diff
@@ -1,15 +1,6 @@
-counterpart = require 'counterpart'
 React = require 'react'
 TitleMixin = require '../lib/title-mixin'
 apiClient = require 'panoptes-client/lib/api-client'
-OwnedCardList = require '../components/owned-card-list-by-discipline'
```
alexbfree: removing unused include

app/components/project-card-list.cjsx
```diff
@@ -228,6 +229,7 @@ ProjectFilteringInterface = React.createClass
       .then (projects) =>
         pages = projects[0]?.getMeta()?.page_count
         project_count = projects[0]?.getMeta()?.count
+        project_count = 0 if !project_count
```
eatyourgreens: `project_count ?= 0`

app/pages/projects-disciplines.cjsx
```diff
@@ -1,15 +1,6 @@
...
-
-counterpart.registerTranslations 'en',
```
alexbfree: this set of messages already exists in `projects.cjsx` and in fact that is the one used, this one is unused so can be deleted.

css/project-list-pages.styl
```diff
@@ -43,6 +43,10 @@
...
+      box-shadow: rgb(64, 64, 64) 0px 0px 1px 0px inset !important
```
eatyourgreens: Haven't we agreed no drop shadows in the new design @srallen @jwbmartin

css/project-list-pages.styl
eatyourgreens: House style is to avoid `!important`. Is it necessary here?

app/components/project-card-list.cjsx
alexbfree: I knew there had to be a simpler way! thanks!

app/pages/projects-disciplines.cjsx
eatyourgreens: Didn't you delete this file in another PR, or am I getting confused?

css/project-list-pages.styl
alexbfree: Bottom line, I spent so long trying to solve this problem, I can't afford to spend any more time on it. Something very odd is going on with the CSS inheritance on this page and so this is the best quick fix I can do for now. Technically it's probably not necessary and there's a better way. Pragmatically, it's necessary for now.

app/pages/projects-disciplines.cjsx
alexbfree: That's true, I have a line in the other PR that removes the OwnedCardList dependency. The removal of the translation stuff is new tho. Will edit so we don't have the same change in 2 PRs.

css/project-list-pages.styl
alexbfree: Can we handle that separately? For now I'm just making the top pagination buttons look the same as the bottom, not adding any new style

eatyourgreens: Could that be fixed by removing the extra `nav` element that's wrapped around the bottom pagination buttons? Might be a quicker fix than writing new CSS. Apart from that element, it looks like both navigation blocks have the same markup.

css/project-list-pages.styl
alexbfree: maybe i can cheat and make a rule that is more specific than the other. will see if i can do that as a quick fix/compromise.

eatyourgreens: I had a quick look and I've found the problem. There's an error in the markup that I submitted as part of this commit 2786572acef9bccba20e45f074b7700d115c7878

I'm happy to fix the markup (the pagination stuff needs moving out of the `resource-results-counter` div) but you  don't need to submit any CSS changes in this PR.

css/project-list-pages.styl
alexbfree: You're right that that double nested nav shouldn't be there.
However, i just tried removing it in inspector and it makes no difference to styling. 

Besides which, it's the styling of the bottom element that we want, not the top one (which is kind of greyed out and bolded and thin-boxed)

The markup does look the same - I think the problem is that the "correct" look is happening through some accidental inheritance rather than being explicitly specified. 

I'm not saying we shouldn't remove the box-shadow, just saying let's not handle that issue as part of this PR.

eatyourgreens: See my comment above. I've figured it out. The markup for the top pagination buttons is in the wrong place, and inheriting grey colours from its parent.

css/project-list-pages.styl
alexbfree: wow - good spot! it picked up so much of the styling it didn't occur to me it might have been the wrong element type. OK I will remove all attempts to modify button styling from this PR. Will you take care of the nested nav at the bottom as well, or shall I do that?

alexbfree:
OK good to merge now I think @eatyourgreens ?

eatyourgreens:
One sec. I'll issue a PR with the markup changes.

eatyourgreens:
Here you go https://github.com/alexbfree/Panoptes-Front-End/pull/1

alexbfree:
Great - thanks @eatyourgreens . Merged and tested. All seems good now if you would like to proceed with the merge.

eatyourgreens:
👍

PR #2427, opened 2016-04-13
resolves frozen subTask editor - fixes #2422

### LINKED ISSUE #2422: Tool tip task not working
@simensta Trying to use the tool-tip task (i.e., the sub-task on a marking task) on this project on either Chrome or Firefox on a Mac OSX 10.9.5. If I remember correctly, you created this tool, so I thought best to direct this question to you.

When I click on question (see image here), nothing happens. Can you take a look and see what might be causing this? I was able to use this tool a month or so ago with no problem. 

I've added you as a collaborator on the project so you can try as well. https://www.zooniverse.org/lab/2501/workflow/1758

tool-tip-error

simensta:
The `<GenericTaskEditor />` editor needs to check for a `@props.project` before checking if `'hide previous marks' in @props.project.experimental_tools`, otherwise the subTask editor freezes as documented in Issue #2422.

### DIFF
--- a/app/classifier/tasks/generic-editor.cjsx
+++ b/app/classifier/tasks/generic-editor.cjsx
@@ -57,7 +57,7 @@ module.exports = React.createClass
         </div>}
       {' '}
 
-      {if 'hide previous marks' in @props.project.experimental_tools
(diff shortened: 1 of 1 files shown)


brian-c:
Good catch, thanks.

PR #2429, opened 2016-04-13
cleanup variables, markdown-help, other misc

mcbouslog:
nothing major here, just a little clean-up, but please take a look and let me know if any questions. more substantial updates (CSV import, presets, etc.) will be separate pull requests.

### DIFF
--- a/app/classifier/tasks/dropdown/dropdown-dialog.cjsx
+++ b/app/classifier/tasks/dropdown/dropdown-dialog.cjsx
@@ -8,11 +8,10 @@ DropdownDialog = React.createClass
     selects: []
     initialSelect: {}
     related: []
-    onSave: ->
     onCancel: ->
+    onSave: ->
 
   getInitialState: ->
-
     editSelect: JSON.parse JSON.stringify @props.initialSelect
     optionsKeys: {}
     conditionalSelects: []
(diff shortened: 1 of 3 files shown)


app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -28,6 +27,9 @@ DropdownDialog = React.createClass
       conditionalSelects.reverse()
       @setState conditionalSelects: conditionalSelects
 
+  handleFocus: (e) ->
+    e.target.select()
```
brian-c: This isn't really standard behavior for text inputs.

mcbouslog:
Yeah that's probably overkill, the autofocus is enough, I'll delete/update.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -84,7 +83,7 @@ module?.exports = React.createClass
       return true if not requiredSelects.length
 
       select = (i) ->
-        return i if annotation.value[i]?.value isnt null
+        return i if annotation.value[i]?.value isnt null and annotation.value[i]?.value isnt undefined
```
brian-c: CoffeeScripts' `?` operator checked `null` and `undefined` for you, so this could be `annotation.value[i]?.value?`.

But we can actually simplify this whole function quite a bit:

```coffee
isAnnotationComplete: (task, annotation) ->
  task.selects.every (select, i) ->
    not select.required or annotation.value[i]?.value?
```

mcbouslog:
the simplified isAnnotationComplete looks amazing, thanks! i wasn't familiar with .every but will keep in mind.

PR #2430, opened 2016-04-13
Persist recent subject notes toggle

### LINKED ISSUE #1869: Preserve state of hiding subject notes in Recents
I've noticed that whenever you return to a Recents page on Talk, the "subject notes" is always clicked, so you always have to manually unclick it if you just want to see non-subject discussions.

Could we perhaps preserve the state of not displaying those, either on a per-Talk basis or across all of Talk?

parrish:
Extracts/refactors query param updating used by the paginator.

Also refactors the paginator to accept a key param to allow paginating multiple resources on the same page.

Closes #1869.

### DIFF
--- a/app/talk/lib/paginator.cjsx
+++ b/app/talk/lib/paginator.cjsx
@@ -1,12 +1,6 @@
 React = require 'react'
 {History} = require 'react-router'
-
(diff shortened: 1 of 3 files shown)


brian-c:
Looks good, thanks.

PR #2436, opened 2016-04-14
Fix projects page filters

### LINKED ISSUE #2372: [project page branch] When you change the sort order, or filters, page must be reset to 1
At the moment, if you're on say page 3, then change the sort order, you end up on page 3 of the new sort order, which is confusing.

Probably easiest to just delete page parameter when adding a sort parameter.

### LINKED ISSUE #2419: [project page branch] Clicking 'All Disciplines' doesn't remove active filter

eatyourgreens:
Resets the page to 1 when discipline changes.
Resets the discipline filter for All Disciplines.
Fixes #2419
Fixes #2372

### DIFF
--- a/app/components/project-card-list.cjsx
+++ b/app/components/project-card-list.cjsx
@@ -239,7 +239,8 @@ ProjectFilteringInterface = React.createClass
         @setState loading: false
 
   handleDisciplineChange: (discipline) ->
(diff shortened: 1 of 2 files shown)


alexbfree:
looks good. i didn't like my string based solution anyway, just didn't know about how to use the object param.

Issue #2424, opened 2016-04-13
[project page branch] Decide default sort order and set it


alexbfree:
Per #2417, even though we have removed default sort from the dropdown, it is still in effect when you first arrive on the page. Need to do something about this, either updating backend to return a more sensible order, or activating one of the more useful sorts by default.

This is more of a To-Do than a bug, just noting it here so we don't forget. Not a blocker for launch.

camallen:
@alexbfree the backend has a "sensible" project sort order (via the `launch_approved` and `beta_approved` query params), [code here](https://github.com/zooniverse/Panoptes/blob/6237b1df643c8464ecb1cab868088a2bb3ada083/app/controllers/api/v1/projects_controller.rb#L55). 

![image](https://cloud.githubusercontent.com/assets/295329/14489557/d9e6c706-0165-11e6-9b81-2df4a4e3d19f.png)

The results may not be what you expect but the sort order is specified and configurable.

eatyourgreens:
Default sort order is set in the front-end here, I think, as activity descending:
https://github.com/zooniverse/Panoptes-Front-End/blob/project-page/app/components/project-card-list.cjsx#L188

eatyourgreens:
@alexbfree I've started a branch for the sort issues: https://github.com/zooniverse/Panoptes-Front-End/tree/project-page-sort

alexbfree:
@camallen I don't disagree that that is theoretically sensible, but in practice the actual data is coming up in a much more arbitrary order (perhaps this is a data issue?) . I've no strong views whether we address this front end or back end or both, but the current default order we're actually getting with the production data isn't useful. 

There was some discussion on slack https://zooniverse.slack.com/archives/dev/p1459518622000517 where we basically agreed that for whatever reason current project order isn't coming up in a predictable or useful order (see grant's screenshot).

alexbfree:
@eatyourgreens which sort issues exactly? Do I need to move the sort changes in https://github.com/zooniverse/Panoptes-Front-End/pull/2417 over there?

eatyourgreens:
@alexbfree I'm confused. Isn't #2417 already merged into that branch?

camallen:
my point is the current default sort order is not a code issue and not related to the front end at all (unless to specify the sort order), instead some data wrangling in the api layer.

eatyourgreens:
[...]

PR #2438, opened 2016-04-14
Sort by project name

eatyourgreens:
Adds sort by project name to the list of options.
Moves the sort options from app/components to app/lib.

### DIFF
--- a/app/components/project-card-list.cjsx
+++ b/app/components/project-card-list.cjsx
@@ -6,7 +6,7 @@ Select = require 'react-select'
 Translate = require 'react-translate-component'
 debounce = require 'debounce'
 Filmstrip = require '../components/filmstrip'
-PROJECT_SORTS = (require './project-sorts').PROJECT_SORTS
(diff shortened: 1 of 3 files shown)


app/lib/project-sorts.coffee
```diff
@@ -1,5 +1,6 @@
 module.exports =
   PROJECT_SORTS: [
+    {value: 'display_name', label: 'Project Name'}
```
alexbfree: Maybe 'Alphabetical' would be a better description of this ordering?

app/components/project-card-list.cjsx
```diff
@@ -186,7 +186,7 @@ ProjectFilteringInterface = React.createClass
   getDefaultProps: ->
     discipline: ''
     page: 1
-    sort: '-activity'
+    sort: 'display_name'
```
alexbfree: Good - better than `-activity` - but see #2424 - not sure this is the best default order. Let's go with this for now though, then if Chris wants something different we can change it under #2424.

alexbfree:
Despite my comments, I see no reason not to merge this now.

PR #2441, opened 2016-04-14
[project page branch] Rename "Astronomy" category to "Space"

alexbfree:
... so that it encompasses "Planetary Science" too. Per @mschwamb's suggestion in https://github.com/zooniverse/Panoptes-Front-End/issues/2373 - @mrniaboc and I both agreed it is a good idea - and Space is what it used to be called on old Zoo home.


mschwamb:
Thanks!

PR #2445, opened 2016-04-14
Fix issues with talk moderations

### LINKED ISSUE #2166: Moderation: Filters over reported comments behave inconsistently/intermittently return inaccurate results
I just made a test project, with two boards, and a couple of test comments.
Was playing with moderation controls, so I deleted one comment and ignored the other.

https://www.zooniverse.org/projects/alexbfree/cat-spotter/talk/moderations?state=all

Now when I click between All Reports - Opened - Ignored - Deleted, I see varying results:

- "All reports" is consistently empty once all reports processed - perhaps rename to "New Reports" or "requiring review" or something?
- "Opened" sometimes shows one comment (at random), sometimes shows two (all). BTW shouldn't this be called "Moderation Log" or "Moderated Reports" or something like that?
- "Ignored" sometimes shows "There are not currently any reports that require moderation." and sometimes shows the one correct ignored message
- "Delete" sometimes shows "There are not currently any reports that require moderation." and sometimes shows the one deleted message

### LINKED ISSUE #2167: Moderation: Deleted comments preview show a broken but still clickable link
After deleting a comment, it can still be viewed. the breadcrumbed link is still generated, but takes you to an undefined URL. Probably we should not generate the link for deleted comments. 

screenshot 2016-01-29 16 56 45

screenshot 2016-01-29 16 58 09

parrish:
This closes #2166 and also closes #2167.

I also tried to cleanup code sprawl by extracting components to isolate state.

The change to pages/profile/comment-link are to prevent linking to deleted boards or discussions as mentioned in #2167.

Feedback welcome.

### DIFF
--- a/app/pages/profile/comment-link.cjsx
+++ b/app/pages/profile/comment-link.cjsx
@@ -51,20 +51,23 @@ module?.exports = React.createClass
   render: ->
     <div className="profile-feed-comment-link">
       <header>
-        <span className="comment-timestamp"title={moment(@props.comment.created_at).toISOString()}>
+        <span className="comment-timestamp" title={moment(@props.comment.created_at).toISOString()}>
           {moment(@props.comment.created_at).fromNow()}
-        </span>{' '}
-        in{' '}
-        <a href={@state.href}>
-          {if @state.project? and @state.owner
-            <span>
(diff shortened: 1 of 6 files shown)


brian-c:
Thanks @parrish!

PR #2451, opened 2016-04-15
Make talk-recents request after component mounts

### LINKED ISSUE #1869: Preserve state of hiding subject notes in Recents

brian-c:
Fixes #1869.


parrish:
Good catch

PR #2464, opened 2016-04-18
setting up context

amyrebecca:
user, project, workflow, and subject are now exposed to the context by the PanoptesApp object; the child components that are responsible for them use callbacks (also exposed on the context) to keep these values in sync

this means that anywhere in the application can inspect these values by adding

contextTypes:
  user: React.PropTypes.object

and then looking at the value @context.user

this way all of these things will be available for all of our logging.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -390,6 +390,9 @@ Classifier = React.createClass
 module.exports = React.createClass
   displayName: 'ClassifierWrapper'
 
+  contextTypes:
+    updateSubject: React.PropTypes.func
+
   getDefaultProps: ->
     user: null
     classification: {}
@@ -414,6 +417,9 @@ module.exports = React.createClass
 
     unless nextProps.classification is @props.classification
       @loadClassification nextProps.classification
+
+  componentWillUpdate: (nextProps, nextState) ->
+    @context.updateSubject nextState.subject if nextState.subject isnt @state.subject
 
   loadClassification: (classification) ->
     @setState
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -63,6 +63,7 @@ module.exports = React.createClass
 
   contextTypes:
     updateWorkflow: React.PropTypes.func
+    updateClassification: React.PropTypes.func
 
   getDefaultProps: ->
     query: null
@@ -86,6 +87,7 @@ module.exports = React.createClass
 
   componentWillUpdate: (nextProps, nextState) ->
     @context.updateWorkflow nextState.workflow if nextState.workflow isnt @state.workflow
+    @context.updateClassification nextState.classification if nextState.classification isnt @state.classification
 
   loadAppropriateClassification: (_, props = @props) ->
     # To load the right classification, we'll need to know which workflow the user expects.
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -61,6 +61,9 @@ module.exports = React.createClass
 
   title: 'Classify'
 
+  contextTypes:
+    updateWorkflow: React.PropTypes.func
+
   getDefaultProps: ->
     query: null
     project: null
@@ -80,6 +83,9 @@ module.exports = React.createClass
   componentDidMount: () ->
     @getCurrentWorkflow().then (workflow) =>
       @setState {workflow}
+
+  componentWillUpdate: (nextProps, nextState) ->
+    @context.updateWorkflow nextState.workflow if nextState.workflow isnt @state.workflow
 
   loadAppropriateClassification: (_, props = @props) ->
     # To load the right classification, we'll need to know which workflow the user expects.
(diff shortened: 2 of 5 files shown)


alexbfree:
I'm happy to merge this into the geordi-integration branch - I think @brian or @parrish or @camallen who know the code better than I should review before this gets merged to master tho.

brian-c:
I have concerns about setting up this many app-level context properties (these effectively global variables) for the purpose of logging.

If the end goal is to keep a record about the last-viewed project/subject/workflow and last-created classification, what about just keeping an app-level `log` function in context, and giving that a method to change the record? E.g. `@context.log.record({project: '123'})` and later `@context.log('User did a thing')`. That way we only expose one global, the logger can keep its information private to itself, and components won't be able to rely on any unnecessarily global properties.

I think app-wide log function (as well as the current user), is an appropriate use of app context.

amyrebecca:
it's fine with me if we only want to expose a single global. i do think that being able to ask about what user is logged in from wherever we are in the tree could be useful, but i know that it also reduces the ability to reuse components so...

brian-c:
Yeah, `user` makes sense, since it's part of the current state of the whole app. There are likely to be components that don't care about the current user sitting between the app root and some component that does. For example in a structure like App/Header/LoginBar, the LoginBar needs to know the user, but the Header itself probably doesn't. We'd currently solve this by passing the user down through the Header anyway (meaning it has to keep track of a property it doesn't actually use), but that's only because we wrote all this before React's context was publicly documented.

alexbfree:
I agree with your general concerns about global variables Brian and yes they are to be avoided.. 

One thing I would note is this context is not just needed for logging - I would like to make use of it in the Talk improvements I am working on too - having more knowledge of the current context can help build better UIs and more contextual navigation options.... 

So in general it would be nice to have awareness of context throughout the UI.

This would also let us do enhancements such as (theoretical example, not saying we would actually do this) having a Talk link that takes you to your current project talk while you're in a project, but general Talk when you're not, or such as keeping track of your current project while viewing collections, favourites or user profiles clicked on from Talk.

Actually @amyrebecca I just realised, assuming we can convince the others to go ahead with this, given I will want to use the context stuff elsewhere too, we should probably create a separate branch for the context stuff and get it merged sooner, before geordi-integration. Otherwise I would have to cherry pick certain commits, or wait until geordi integration is finished, neither of which is ideal.

amyrebecca:
the context work i did is simple enough if we need it, but i just spent the past few hours stripping the interface down to what @brian-c suggested, so i'm going to leave it that way for now.

alexbfree:
just changed my mind a bit so updated my comment above.

alexbfree:
>just spent the past few hours stripping the interface down to what @brian-c suggested

which suggestion? to only store `user` ? or to store a global `log` variable?

amyrebecca:
Yeah, both. Going to have a user object on the context as well as a log method.

alexbfree:
OK. I'm not 100% convinced about the log method approach, though I can see it has a certain elegance. Let me know when there is something to review.

brian-c:
@alexbfree Every UI state should be able to be described strictly with the URL and the current user. The same URL should not render two different ways depending on any previous behavior.

amyrebecca:
I mean unless there is some bug, the project on the context and the project on the URL should be identical, and it COULD save some db lookups. However, keeping the same piece of information in two places IS inelegant and a good way to introduce bugs.

alexbfree:
Sorry if my comments were a bit unclear @brian-c, I was sort of thinking it through in writing as I typed that: Let me clarify:

I completely agree that "the same URL should not render two different ways".
That's not quite the same as making additional options available in a certain context
(For example, making the Admin menu available when admin mode is active, or having a sign out link when a user is logged in)

Overall though I think it's more a question of whether the needed information is available when rendering that URL. Current user is a good example of something that isn't, and it seems we're all agreed we can store that context globally. Hopefully everything else will be solvable without need to store anything else in context - whether or not we need the "log method" storage time will tell! But I think we are basically agreed.

amyrebecca:
Tbh in a lot of ways the practice of stuffing a lot of data onto the context and then just asking for it isn't terribly different from the redux model, but there is the important difference that we don't provide dispatching, so that we can easily get stuck in callback hell with componentWillReceiveProps loops, etc

eatyourgreens:
Could the geordi client follow the same pattern as the Panoptes client ie. require it into components where needed, then call methods on the client to update project/workflow/subject data and log that data during component updates?

We experimented with using Context in https://github.com/zooniverse/wildcam-gorongosa-education but dropped it since updating data in two places introduced bugs, and Facebook says context should not be used to transfer model data between components: https://facebook.github.io/react/docs/context.html

srallen:
Perhaps we can discuss during miniztm since redux is on the agenda?

amyrebecca:
It's hard to see what the possible point of context COULD be beyond transferring model data between components.

amyrebecca:
Also +1 for Redux, I've built web apps with it and it's pretty nice. Keeping all your application state in one place has a lot of things to recommend it, especially if it's protected by a dispatch mechanism.

alexbfree:
@srallen sounds good to discuss next week. Since I can't be there in person, please could I participate in that discussion via Skype/hangout etc if at all possible? I'll understand if times don't work out. Cheers

PR #2472, opened 2016-04-19
Fix conditional for redirect

### LINKED ISSUE #2470: Project page search results don't link to projects
Clicking the name of a project in the search results list does to the home page rather than the project's page.

srallen:
Fixed #2470. The issue is that the `project.redirect` prop is a an empty string in the production db whereas it's null in dev. I've fixed the conditional to check for both. This will fix search result redirects for projects that do not have an external redirect set.

Branch can be viewed here: https://fix-project-page-search.pfe-preview.zooniverse.org/projects


eatyourgreens:
👍🏼

PR #2473, opened 2016-04-19
Enable WorldWide Telescope

### LINKED ISSUE #2295: Design UI elements for WWT integration

### LINKED ISSUE #2296: ClassificationSummary plugin architecture

### LINKED ISSUE #2297: Figure out how to calculate inputs to WWT API for sample images

### LINKED ISSUE #2298: API call to store a crop of an image on AWS

### LINKED ISSUE #2299: Compute WWT API call from classification

### LINKED ISSUE #2300: Project Builder changes to enable / disable WWT in summary

wgranger:
closes #2300 closes #2299 closes #2298 closes #2297 closes #2296 closes #2295 

This PR allows WorldWide Telescope (WWT) to be used as an experimental feature for the Old Astronomy project. WWT must be enabled as an experimental feature, then turned on in the project workflow. At the end of a classification, a button will appear allowing the user to view her/his classification with the WWT. Use the "Old Astro-test" project in staging for testing.

The WWT button on the classification page will likely be re-styled in the future.


amyrebecca:
Please feel free to loop me into any questions about this pull request. Will and I worked together on it a lot, which is why we need someone other than me to review it.

brian-c:
Thanks, will check this out first thing tomorrow.

brian-c:
@wgranger I don't have access to an "Old Astro-test" project. Please add brian-testing as a collaborator.

wgranger:
@brian-c done

app/classifier/index.cjsx
```diff
@@ -240,6 +241,16 @@ Classifier = React.createClass
     <div>
       Thanks!
 
+      {if @props.workflow.configuration.external_api is 'world_wide_telescope'
+        <strong>
+          <WorldWideTelescope
```
brian-c: The prop names `classification` and `user` don't really match their value here.

app/pages/lab/workflow.cjsx
```diff
@@ -261,6 +261,22 @@ EditWorkflowPage = React.createClass
...
 
+          {if 'worldwide telescope' in @props.project.experimental_tools
+            <div>
+              <div>
+                <AutoSave resource={@props.workflow}>
+                  <span className="form-label">Use World Wide Telescope API</span><br />
+                  <small className="form-help">Allow user to view subject in the WWT after classifying.</small>
+                  <br />
+                  <input type="checkbox" id="world_wide_telescope" onChange={@handleSetWorldWideTelescope} defaultChecked={@props.workflow.configuration?.external_api}/>
+                  <label htmlFor="world_wide_telescope">WorldWide Telescope</label>
```
brian-c: Convention is to wrap the input in its label. Defining an `id` doesn't make sense when building reusable components.

brian-c:
Can I recommend that instead of a `external_api: 'world_wide_telescope` configuration property, we implement this with a more generic key and a more descriptive value like `customizations: ['world_wide_telescope_summary']`.

That way we can implement other per-project customizations the same way, e.g. the Pulsar Hunters feedback. Having a generic way to decouple per-project customizations from the main UI code will keep things a little simpler in the long run.

Less important: please try to follow formatting style of the repo, specifically with regard to spacing around operators, line breaks around functions, multi-line conditional expressions, and filename case.

amyrebecca:
external_api is an awful name for a flag, honestly, and i do think that we need to make sure that it can accept multiple values.

does it seem wise to have a single array contain all possible customizations, or is it more logical to separate them by applicability, for example custom_summary: ['wwt'], custom_feedback: ['pulsar'] etc?

app/classifier/world_wide_telescope.cjsx
```diff
@@ -93,32 +109,40 @@ class Plate
...
   getCropUrl: ->
     url = @url.replace(/^https?\:\/\//i, "")
     # TODO: we need to account for the fact that the size of the image might be different than
     # the size that it is displayed
     "http://imgproc.zooniverse.org/crop/#{@starChart.width}/#{@starChart.height}/#{@starChart.x}/#{@starChart.y}?u=#{url}"
+
   computeRotation: ->
-    if @starChart.xAxis == Axis.RA || @starChart.xAxis == Axis.GLAT then 0 else 90
+    if @starChart.xAxis == Axis.RA || @starChart.xAxis == Axis.GLAT
+    then 0
```
amyrebecca: I believe this introduces a different behavior than the original code because of indenting rules?

brian-c:
@wgranger I'm not able to get this to show the new summary since your changes. Also, as @amyrebecca mentioned, the customizations value should be an array.

@amyrebecca I could see that kind of organization being super useful, since then each section of the classifier could look at its own key. Sound good @wgranger?

wgranger:
Sounds good. I'll take a look at that bug and will add custom_summary as an array value.

amyrebecca:
Is there anything else needed in order to merge this pull request? We know that it's somewhat brittle because of the reliance on the structure of the classification, but it's difficult to know how to avoid that given the lack of symbolic names.

brian-c:
Looks good to me.

trouille:
@wgranger When you implement the fix for correctly aligning within WWT, also think about how you could set things up so that when you click on World Wide Telescope, it opens a new tab or window, rather than staying in the same window. The reason for this is that if there are 3 images you've labeled, you want to be able to click and view all 3 in sequence, but right now it replaces our zooniverse site with WWT on the first image you click, and you can't go back to view the others. Let me know if that makes sense.

wgranger:
@trouille That makes sense. I think it would be nicer if the WWT popped up in another tab so the user isn't distracted and taken away from the workflow. Shouldn't be hard too make the change.

Issue #2107, opened 2016-01-13
Set up metadata on all Zooniverse pages so that Facebook/Twitter/G+ sharing previews work correctly


alexbfree:
At the moment when you share a Panoptes Zooniverse link on social media, the generated preview looks like this:

![screenshot 2016-01-13 12 47 02](https://cloud.githubusercontent.com/assets/1473244/12294711/5041efb6-b9f4-11e5-9dde-6b06af26edcc.png)

By setting the metadata tags and/or facebook open graph tags in each page, we can ensure the generated previews include a thumbnail (for the project) and a description. It's generally accepted that links with images and descriptions get more clicks than plain text ones.

I may get chance to look at this myself but logging the ticket for now, anyone else can feel free to pick it up.

eatyourgreens:
I think this is going to need server-side rendering, since the FB and Twitter crawlers won't execute the javascript that renders these pages. http://www.michaelbromley.co.uk/blog/171/enable-rich-social-sharing-in-your-angularjs-app

alexbfree:
Since server-side rendering isn't likely to happen in the short term, and after discussion with @rogerhutchings, I'm suggesting we should go for an intermediate fix of having a basic Zooniverse logo and site description appear for all Zooniverse links... to be replaced by project-specific previews later, when it is technically easier to do so.

Related links (thanks @eatyourgreens ):
>Here’s facebook’s page where you can test URLs with their crawler. https://developers.facebook.com/tools/debug/
> Here’s what their crawler sees for a Zooniverse project. https://developers.facebook.com/tools/debug/og/echo?q=https%3A%2F%2Fwww.zooniverse.org%2Fprojects%2Ftedcheese%2Fsnapshots-at-sea

It would also be possible to add a Share to Facebook button to handle this - but that may be a slippery slope as we would need to think about all the different social networks people might want to share to....

We probably need to think about not just sharing the Zoo homepage or specific Zoo projects but also sharing subjects, sharing collections, and sharing Talk threads/posts....

eatyourgreens:
`index.html` has OpenGraph tags now. https://github.com/zooniverse/Panoptes-Front-End/blob/master/views/index.ejs#L14-L17

PR #2477, opened 2016-04-20
Add default OG tags

rogerhutchings:
@alexbfree Adds some default open graph tags so sharing links on FB is slightly richer. 

Ref #2107 

screen shot 2016-04-20 at 16 08 38


rogerhutchings:
don't know why the image is off-centre like that, it's exactly at the aspect ratio FB specifies

rogerhutchings:
@alexbfree bump

alexbfree:
Yay! Sorry I just saw this. Nice one Roger.

alexbfree:
PS maybe it's not the aspect ratio, but the image size, that's wrong? maybe try generating one exactly the right pixel size?

PR #2484, opened 2016-04-25
added in help text for the discipline tags 

mschwamb:
I added in help text for the discipline tags since there was nothing currently in the project builder telling you what the discipline tags do. I think I've added in the current structure in the correct place to add in a help line. 

@alexbfree I assume you would be the right one to review it as long as @aliburchard and or @mrniaboc don't have objections to the text


alexbfree:
Good spot! Must have been copy and pasted.

I would change it to:
"Enter or select one or more discipline tags to identify which field(s) of research your project belongs to"

mschwamb:
@alexbfree better?

alexbfree:
Oops just realised I was looking at the bottom comment the first time, not the top one. But yes, this is good. Merging.

PR #2486, opened 2016-04-25
Check engines before starting dev server or building

brian-c:
This'll fail hard when trying to use incompatible Node versions.


alexbfree:
Good idea! Same approach we did on Serengeti

PR #2492, opened 2016-04-25
render favorites button unless workflow configuration states otherwise

saschaishikawa:
Checks `workflow.configuration.disable_favorites` and renders the Fav button only if the value isn't `true`

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -117,7 +117,8 @@ module.exports = React.createClass
             <button type="button" aria-label="Metadata" title="Metadata" className="metadata-toggle" onClick={@showMetadata}><i className="fa fa-info-circle fa-fw"></i></button>}
           {if @props.subject? and @props.user? and @props.project?
             <span>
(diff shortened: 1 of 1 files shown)


app/components/subject-viewer.cjsx
```diff
@@ -117,7 +117,7 @@ module.exports = React.createClass
             <button type="button" aria-label="Metadata" title="Metadata" className="metadata-toggle" onClick={@showMetadata}><i className="fa fa-info-circle fa-fw"></i></button>}
           {if @props.subject? and @props.user? and @props.project?
             <span>
-              <FavoritesButton project={@props.project} subject={@props.subject} user={@props.user} />
+              {<FavoritesButton project={@props.project} subject={@props.subject} user={@props.user} /> unless @props.workflow.configuration.disable_favorites}
```
brian-c: Sorry, I may have misspoke. How about:

```
{unless @props.workflow.configuration.disable_favorites
  <FavoritesButton project={@props.project} subject={@props.subject} user={@props.user} />}
```

Issue #2239, opened 2016-02-29
adding intro text in Project Builder: make character limit explicit


vrooje:
I didn't realize there was a 1500-character limit on the introduction text for a project. I'm working on a project where there's a lot of branding images and required "this project doesn't express the official opinion of..." etc. text that needs to show on the front page. I went over the limit and the text wouldn't save, and I didn't know why until I happened to mouseover the box.

I think we should message this a bit more explicitly.

aliburchard:
should we just take off that character limit?

alexbfree:
It does seem rather arbitrary

vrooje:
Yeah, I'd be a fan of that, or increasing it...

vrooje:
@brian-c or others ( @marten ? @camallen ? @parrish ?), can we remove the character limit, at least temporarily? It's basically impossible to use the introduction for anything useful in our Ecuador earthquake response project because the character limit is only barely high enough to include all the logos of partners -- and we need a few paragraphs of text in there besides.

I can put it in the Research tab but that is really not ideal as there's other stuff that needs to go there... so if this is possible it would be fantastic to do this very very soon.

camallen:
Pretty sure its a front end restriction only. Submitting the data via the client directly would circumvent it as a workaround

brian-c:
@camallen Looks like a back-end thing, sorry:

https://github.com/zooniverse/Panoptes/blob/master/app/models/project_content.rb#L8

Presumably we could just delete this line?

vrooje:
Ah, is that why I got a 400 trying to update the text via the API?

camallen:
Ha, thanks @brian-c. Should we have a limit and if so what size?

alexbfree:
Note that Roger has added a character counter, which has been merged in #2493. This should at least make it easier to see what the problem is. We definitely need to increase this limit though.
4000? 5000?

vrooje:
I'd be in favor of a higher limit like 5k; perhaps with a warning right
below the counter that says something like "The longer your intro is, the
less likely people are to read it."

parrish:
I think this might be a similar problem to the Talk character limit.

>  a lot of branding images

In cases where a user is entering a lot of links or images, the amount of markdown or URL characters can cause the character length to grow far faster than the actual text content.

If we move the validation client-side, we could find the visible character count from the rendered markdown.

[...]

PR #2493, opened 2016-04-25
Add char limit info to fields on project builder

rogerhutchings:
These are hardcoded in at the moment, but I wonder if we could do something clever with an options request to get these

related to #2239


alexbfree:
Nice job @rogerhutchings ! 

BTW out of curiosity, how did you generate a gif like that?

rogerhutchings:
http://www.cockos.com/licecap/

alexbfree:
@rogerhutchings Thanks!

PR #2497, opened 2016-04-26
add focus to dropdowns

mcbouslog:
adds focus for dropdown task

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -97,6 +97,9 @@ module?.exports = React.createClass
     else
       @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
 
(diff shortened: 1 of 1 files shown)


mcbouslog:
need to incorporate @props.autoFocus

mcbouslog:
the dropdown task utilizes module react-select, which currently doesn't have an autofocus attribute (though eventually will once version 1.0.0 is out of beta), so currently focus is assigned with componentDidMount. 

however, the current issue is that when there's multiple dropdowns, and a selection is made on a dependent dropdown, focus jumps back to the first dropdown.

simensta:
Autofocus for the dropdown is behaving as expected in Chrome and Firefox. In Safari, I can see the autoFocus being set in the correct input. But, after the rest of the page loads focus is back on the window. @eatyourgreens, any ideas on what could be causing this?

Issue #2491, opened 2016-04-25
It seems that my classifications aren't being recorded (param is missing)


yshish:
I noticed some seriously looking errors after finishing a classifications on Western Shield Camera Watch
Screenshot:
![](https://lh3.googleusercontent.com/-HwqEZwGr6EE/Vx5l-Nc1xPI/AAAAAAAAI2s/w_lGIKIJo089tOxdxRajhXTURaOrsI7mgCCo/s0/ysh_2016-04-25_20-44-32.png)

Does it mean that my classifications aren't being recorded? (But the counter is counting..)

camallen:
@yshish that does not look good, i can't reproduce on my end. Is this still happening?

yshish:
Yes, it is still happening. But the classification counter hasn't stopped counting. 

Working on Chromium Version 45.0.2454.101 Ubuntu 15.04 (64-bit).

eatyourgreens:
@camallen I've seen this before, but can't remember where. Classifications that failed to post correctly get saved locally and posted again with incomplete metadata. Related to https://github.com/zooniverse/panoptes-javascript-client/issues/25 (the api client doesn't pass back the 422 status code.)

camallen:
@yshish try incongnito mode / another browser to see if it's still an issue. Not sure how some broken classifications got built and stored to localstorage...do they ever expire?

eatyourgreens:
@camallen I _think_ those failed classifications will try to resend forever. I don't see anything in the javascript to expire them. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/classify.cjsx#L287-L303

yshish:
I tried Firefox and got some error which looks different to me (just can't tell whether it is related to it or not).
![](https://lh3.googleusercontent.com/-7cycBf6iUxA/Vx_Q24wsvHI/AAAAAAAAI3E/lbeNCZjKoTUKdL3g8v6wyuiZ77yUzcw4ACCo/s0/Screenshot_2016-04-26_22-26-46.png)

I tried incognito mode in Chromium and got some errors which look different too.
![](https://lh3.googleusercontent.com/-NcjNU4bOk7I/Vx_RvlCT5iI/AAAAAAAAI3Y/d1yi7OxHakg7Vm_dqZqcj8zph2BCGho5wCCo/s0/ysh_2016-04-26_22-33-05.png)

I'm sorry if it isn't helpful. Can you tell?

[...]

PR #2500, opened 2016-04-26
Don't access local state when queueing classifications

brian-c:
For #2491

Still to come: dealing with malformed classifications in the queue.


eatyourgreens:
Looks good to me.

PR #2509, opened 2016-04-29
Remove erroneous top margin

alexbfree:
The first question on all workflows on all projects isn't properly aligning with the top of the subject viewer. This should fix most cases.

### DIFF
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -150,7 +150,7 @@
 
       > :last-child
         margin-bottom: 0
-        
+
     .answer-button-icon-container
       align-self: flex-start
       flex-shrink: 0
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -125,7 +125,7 @@
     margin-top: 2em
 
   .markdown:first-child
-    p:first-child
+    :first-child
       margin-top: 0
 
   // Only show the "required" warning when a task is outside the normal flow.
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -146,7 +146,11 @@
 
       > :last-child
         margin-bottom: 0
-
+      
+    &:first-child
+      p:first-child
+        margin-top: 0
+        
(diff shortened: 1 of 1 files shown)


alexbfree:
Before:
screenshot 2016-04-29 14 51 25

After:
screenshot 2016-04-29 14 50 38

css/classify.styl
```diff
@@ -124,6 +124,10 @@
   & + .workflow-task
     margin-top: 2em
 
+  .markdown:first-child
```
srallen: Maybe use `.question:first-child` instead of `.markdown` so it has a more targeted scope? Not sure if markdown comes in somewhere else in the workflow task where a top margin is desirable.

css/classify.styl
```diff
@@ -124,6 +124,10 @@
...
+    :first-child
```
srallen: Duplicate?

srallen:
You may want to consider a media query. It looks a bit crowded now without it on small viewports.

css/classify.styl
alexbfree: no that was intentional... it looks for the first occurrence of `.markdown`, and then within that for the first child, regardless of whether it's a `p` or `img` or something else

css/classify.styl
alexbfree: Good idea. Will change that.

alexbfree:
ok will do those improvements then reassign

alexbfree:
closed will not fix? or closed because no longer needed / handled elsewhere? I have this in a "might get to this at some point" pile....

srallen:
Closed because it was unfinished and now is so old there are merge conflicts to resolve. It'd be easier to reintroduce this in a new branch and PR.

alexbfree:
ok good to know. I guess i should have kept it open as an issue rather than a PR

PR #2519, opened 2016-05-03
Less obnoxious sign-in suggestion

brian-c:
This removes the "You should sign in" pop-up between _n_ classifications and replaces it with a button under the subject. When the subject changes, the button jiggles tantalizingly for a couple seconds.

jiggle

Right?

In the process, I cleaned up the look of all the under-subject buttons by removing default system style and fixing icon widths.

Also removed a `willReceiveProps` method in the SubjectViewer that was broken and never called anyway. Probably a typo (it would've been `componentWillReceiveProps`).

### DIFF
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -87,13 +87,14 @@ module?.exports = React.createClass
   render: ->
     <PromiseRenderer promise={@state.favoritedPromise}>{(favorited) =>
       <button
-        className="favorites-button"
+        className="favorites-button #{@props.className ? ''}"
         type="button"
         title={if favorited then 'Unfavorite' else 'Favorite'}
         onClick={@toggleFavorite}>
         <i className="
           fa fa-heart#{if favorited then '' else '-o'}
           #{if favorited then 'favorited' else ''}
+          fa-fw
         " />
       </button>
     }</PromiseRenderer>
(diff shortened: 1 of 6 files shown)


mschwamb:
@brian-c I'm not sure I would notice the animated icon below the image if I was classifying. For your consideration, what about the way planet hunters puts up a blue bar in the space above the image and the text question with an x allowing you to close it or just keep it and continue going? I really liked that when it was added in for PH2 launch.

screen shot 2016-05-03 at 5 40 10 pm

brian-c:
@mschwamb Totally valid, this might not be perfect but once it's in when can iterate on it when we have some spare cycles.

amyrebecca:
we're going to need to follow up with the syracuse team about this because of their geordi logging requests. i lost a half-day yesterday to the projects page rewrite, as well. are there any other areas of the app i need to be wary about?

saschaishikawa:
@brian-c the PR looks good to me. I'm inclined to merge. @amyrebecca how does this affect what you're currently doing?

amyrebecca:
some of the UI affordances people wanted us to log have been removed. i have no objection to the PR, just a general gripe i guess :)

saschaishikawa:
@amyrebecca that brings up a good point. How do we know which UI elements we're logging? Is there a config file somewhere, a google doc even?

amyrebecca:
there IS a google doc, i think it's linked over on the geordi-integration milestone or in one of the issues associated

PR #2520, opened 2016-05-03
[WIP] Text task prev answers access

mcbouslog:
per Notes from Nature conversion, legacy version includes Ctrl-M shortcut key combination to access previous answer provided for text tasks (which include annotating scientific name, location, habitat, collector and collector number). Often a collection of subjects will have the same scientific name or collector, so this feature makes annotating that attribute significantly easier, increasing volunteer participation and satisfaction.

this Pull Request adds the following functionality to the **text task**:

1. Ctrl-M recalls most recent answer provided (from local storage), pressing again cycles through in chronological order (from most recent to least recent, then back to most recent)
2. Ctrl-K clears answer shown (from local storage) and from being accessed in the future
3. answers are stored to local storage associated with a key defined as `"#{window.location.pathname}-#{question}"`, thereby preventing a volunteer from pressing Ctrl-M in a Scientific Name field and provided Collector names previously provided

[branch staged for testing](https://text-task-prev-answers.pfe-preview.zooniverse.org)

eventually shortcut keys (and how we note what shortcut keys are available) will be improved site-wide, and this text task functionality will be included in those improvements.

### DIFF
--- a/app/classifier/tasks/generic-editor.cjsx
+++ b/app/classifier/tasks/generic-editor.cjsx
@@ -48,6 +48,11 @@ module.exports = React.createClass
             <MarkdownEditor name="#{@props.taskPrefix}.help" onHelp={-> alert <MarkdownHelp/>} value={@props.task.help ? ""} rows="7" className="full" onChange={handleChange} />
           </AutoSave>
           <small className="form-help">Add text and images for a window that pops up when volunteers click “Need some help?” You can use markdown to format this text and add images. The help text can be as long as you need, but you should try to keep it simple and avoid jargon.</small>
+          {if @props.task.type is 'text'
+            <div>
+              <br />
+              <small className="form-help">For text tasks consider adding the help message "Use Ctrl-M to cycle through previous answers provided and Ctrl-K to clear previous answers."</small>
+            </div>}
         </div>}
 
       {if choicesKey?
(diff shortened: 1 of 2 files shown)


mcbouslog:
people that may be interested: @simensta, @eatyourgreens, @brian-c

brian-c:
Why control-M? Isn't that probably taken on Windows? Maybe the up and down arrows (only if the entire input is selected) would be more discoverable.

brian-c:
I'd key this on the workflowID and task key instead of the URL and task instruction. More stable and predictable.

mcbouslog:
Control-M what currently (per legacy NfN) used. Per [this Microsoft list](https://support.microsoft.com/en-us/kb/126449) the Ctrl-M and Ctrl-K combinations don't appear to be currently used, but please let me know if there's a better reference to check. Otherwise we thought about other key options, but any suggestion will have pros/cons and thought might be simplest to stick with current setup?

I like your suggestion regarding workflowID and task key - will make that change pending other comments/edits.

mcbouslog:
[this branch staged for testing](http://text-task-arrow-answers.pfe-preview.zooniverse.org/) is set so the up and down arrows (no Ctrl key pressing required) cycle through previous answers. Ctrl + K is still set to clear an answer, but thoughts on what that should be if going a non-alpha-character route? the code is on [a branch in my forked PFE repo here](https://github.com/mcbouslog/Panoptes-Front-End/tree/text-task-arrow-answers) if interested in details, but generally similar.

otherwise, general thoughts on [simply up and down arrows](http://text-task-arrow-answers.pfe-preview.zooniverse.org/) vs [Ctrl-M](https://text-task-prev-answers.pfe-preview.zooniverse.org/)?

@brian-c , @eatyourgreens , @simensta (tagging a few that may be interested, but no rush to respond)

brian-c:
I've sketched out a little component that works the way I think this should work:

https://historied-input-poc.pfe-preview.zooniverse.org/ (then click "Dev classifier", and it's just stuck on top there.)

historied-input

This can render as a text input or textarea (with `tag="textarea"`). It just needs a `historyID` prop (in this case that'd probably be `user.id + workflow.id + taskKey`.

When the selection is at the end of the input, the up and down keys cycle through history, filtered by any text before the selection like a simple shell-style history. If all the text is selected, option-delete removes that entry from the history (I assume this is more of a power user thing?).

https://github.com/zooniverse/Panoptes-Front-End/blob/historied-input-poc/app/components/historied-input.cjsx

mcbouslog:
Closing pull request to rebase and rebuild based on "historied input" example.

PR #2521, opened 2016-05-03
A more pleasant project-status admin page

srallen:
Adds an `admin-page` stylus file with a few styles for the project status admin page and some better markup. Started to use BEM here too. The experimental feature list is getting unwieldily; this makes it a bit more pleasant to look at.

### DIFF
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -106,7 +106,7 @@ ProjectExperimentalFeatures = React.createClass
       <AutoSave resource={@props.project}>
         <div className="project-status__section--table">
           {EXPERIMENTAL_FEATURES.map (task) =>
-            <label key={task}>
+            <label key={task} className="project-status__section--table--row">
               <input type="checkbox" name={task} checked={@setting(task)} onChange={@updateTasks.bind @, task} />
               {task.charAt(0).toUpperCase() + task.slice(1)}
             </label>}
@@ -156,8 +156,8 @@ VersionList = React.createClass
   render: ->
     <PromiseRenderer promise={@props.project.get 'versions'}>{ (versions) =>
       vs = versions.sort()
+      <h4>Recent Status Changes</h4>
       <ul className="project-status__section--list">
-        <h4>Recent Status Changes</h4>
         {vs.map (version) =>
           key = Object.keys(version.changeset)[0]
           from = version.changeset[key][0].toString()
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -104,9 +104,9 @@ ProjectExperimentalFeatures = React.createClass
     <div className="project-status__section">
       <h4>Experimental Features</h4>
       <AutoSave resource={@props.project}>
-        <div className="project-status__section--table">
+        <div className="project-status__section-table">
           {EXPERIMENTAL_FEATURES.map (task) =>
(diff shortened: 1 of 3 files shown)


css/admin-page.styl
```diff
@@ -0,0 +1,13 @@
+.admin-page
+  .project-status
```
brian-c: With BEM none of these need to be nested.

You could still imply the hierarchy with the Stylus `&` feature:

```styl
.project-status
  &__section
    margin: ...

    &--list
      margin: ...
```

app/pages/admin/project-status.cjsx
```diff
@@ -152,7 +156,8 @@ VersionList = React.createClass
   render: ->
     <PromiseRenderer promise={@props.project.get 'versions'}>{ (versions) =>
       vs = versions.sort()
-      <ul className="project-status-changes">
+      <ul className="project-status__section--list">
+        <h4>Recent Status Changes</h4>
```
brian-c: An H4 child of a UL is odd, maybe swap these lines?

css/admin-page.styl
```diff
@@ -0,0 +1,13 @@
...
+    .project-status__section
+      margin: 1em 0
+      
+      .project-status__section--list
+        margin: 0 0 1em 0
+
+      .project-status__section--table
+        display: table
+        
+        label
```
brian-c: Give this a class to avoid nesting.

brian-c:
Much nicer, thanks! Left some comments inline re: BEM implementation.

srallen:
Thanks for the suggestions. Got those in.

app/pages/admin/project-status.cjsx
```diff
@@ -101,13 +101,16 @@ ProjectExperimentalFeatures = React.createClass
...
-    <div>
+    <div className="project-status__section">
+      <h4>Experimental Features</h4>
       <AutoSave resource={@props.project}>
-        {EXPERIMENTAL_FEATURES.map (task) =>
-          <label key={task}>
-            <input type="checkbox" name={task} checked={@setting(task)} onChange={@updateTasks.bind @, task} />
-            {task.charAt(0).toUpperCase() + task.slice(1)}
-          </label>}
+        <div className="project-status__section--table">
```
brian-c: Sorry, I don't want to nitpick, this is a new way of doing things and I want to make sure we' start out on the same page.

Should this be the element `section-table`? This implies it's another nested `section` with a `table` modifier.

app/pages/admin/project-status.cjsx
```diff
@@ -101,13 +101,16 @@ ProjectExperimentalFeatures = React.createClass
...
+          {EXPERIMENTAL_FEATURES.map (task) =>
+            <label key={task} className="project-status__section--table--row">
```
brian-c: And is this `section-table-row`?

srallen:
No worries about the nit-picks. I think I need to look at more BEM implementation examples since it's not entirely clear to me when you decide when something is a block vs element vs modifier. Anyway, I've updated the PR.

brian-c:
Looks solid, thanks.

My understanding is: Blocks are totally independent and should be able to render properly anywhere, elements won't make any sense outside the context of their block, and modifiers are just different states of a block or element like "has-errors" or "needs-attention". I think the built-in pseudo-classes like `:active` and `:disabled` can be used in place of a modifier as well.

PR #2523, opened 2016-05-03
Allow collections manager dialog overflow

### LINKED ISSUE #2056: Drop down for the list of existing collections doesn't work properly
I've started using Panoptes collections in big and noticed a bug:

When I click on the drop-down to open the list of existing collections, it closes immediately. It only stays opened as long as I keep the mouse button down on the drop-down arrow. Once I put my finger up from the button, the list closes/disappears. The only way to keep it opened is to click and hold the button down while moving the cursor to the left, on the 'Collection name' text. Then it stays opened. (Not sure it is the problem, but after clicking on the drop-down a scroll bar appears on the right side of it which makes the space for the head line *Add Subject to Collection* text too narrow and so it divides into two lines - it moves the *collection* word down. It also moves the drop down for opening the list down, so the cursor doesn't stay on it. Could this be causing closing of the list?

Why this is so irritating: When the list is longer and I need to scroll down to choose a collection which is not displayed on the first four positions, it is a big problem. I'm using a track point so I can't scroll other way then by using the scroll bar on the side of the window. When I know the names of the collection which are hidden, then I can type their names. But since there are all collections from all Panoptes projects displayed in the same list, it will become a problem soon. 

And one more thing.. The list of collections doesn't display what project they are related to, so I'm afraid I will put the image into a 'wrong' one. I would like to keep the collections separated for each project. When we have such similar projects like Whales as indi and Snapshot at sea, it will be difficult to recognize what project is my Humpback collection related to since we see humpbacks in both projects, unless I add the project name into the name of the collection. Is this a final solution for the list of collections, or are you about to change that somehow? Please, consider that :) I believe I am not the only one who would like to keep the collections for individual projects.

Hope it makes sense.
Working on Chromium, Ubuntu at this moment, unable to test other browsers today.

brian-c:
Extending #2522, this switches to modal-form Dialog component to allow the collections dropdown to extend beyond the dialog bounds.

Closes #2522, which closes #2056.

### DIFF
--- a/app/collections/manager-icon.cjsx
+++ b/app/collections/manager-icon.cjsx
@@ -13,15 +13,21 @@ module?.exports = React.createClass
   getInitialState: ->
     open: false
 
+  open: ->
+    @setState {open: true}
+
+  close: ->
+    @setState {open: false}
+
   render: ->
     <button
       className="collections-manager-icon"
       title="Collect"
-      onClick={@setState.bind this, {open: true}, null}>
+      onClick={@open}>
       <i className="fa fa-list" />
 
       {if @state.open
-        <Dialog tag="div">
-          <CollectionsManager user={@props.user} project={@props.project} subject={@props.subject} onSuccess={@setState.bind this, {open: false}, null} />
+        <Dialog tag="div" closeButton={true} onCancel={@close}>
+          <CollectionsManager user={@props.user} project={@props.project} subject={@props.subject} onSuccess={@close} />
         </Dialog>}
     </button>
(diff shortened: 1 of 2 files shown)


parrish:
Works well, but it looks like we need to add/implement a close button:

``` coffeescript
close: ->
  @setState open: false

render: ->
  # ...
  <Dialog tag="div" closeButton={true} onCancel={@close}>
    # ...
  </Dialog>
```

brian-c:
Yes indeed, I've just added it.

parrish:
👍

PR #2524, opened 2016-05-04
pfe with webpack

amyrebecca:
Do not merge until we've had some discussion, but here's the webpack work.

npm start runs the server in dev mode, with hot reloading etc
npm run build builds a static version of the site
npm run serve serves a static version of the site
npm run stage-branch, etc. work as they used to

right now refresh is broken on branches on preview, but i think that was already the case. i'm auto-generating a .htaccess file that should fix that, but publisssh doesn't publish dotfiles. i've created issue brian-c/publisssh#7 to address that, and then we can test the .htaccess fix.

this looks like an awful lot of changed files, but many of them were just changing the way that we referred to markdownz and can be glanced over very quickly.


camallen:
> right now refresh is broken on branches on preview, but i think that was already the case

what url are you using? From memory it only works via `branch-name-panoptes-front-end.zooniverse.org`

amyrebecca:
i didn't actually know you could access it that way! i've been using preview.zooniverse.org/panoptes-front-end/branch-name

brian-c:
Have only glanced at this, but real quick:

- Looks like this needs a little rebase on master.

- Isn't .htaccess just an Apache thing? What's it do here?

- Lots of duplication between the dev and production configs, can we require (parts of) one into the other?

Will take a look at this further this afternoon. Awesome work, this will be great. Thanks!

amyrebecca:
Thanks for checking it out! 

* Does it need rebased? It says that there are no conflicts so I'm wondering what I missed

* I wasn't sure if we were serving with nginx or apache so this was just a stab at the "refresh = 404" problem we get there, we can probably drop it if this won't help

* I'll look at how to consolidate the configs, but I sort of liked the idea of keeping them independent so that developers don't mangle the config we use for production accidentally. Of course there is the risk of the config not getting updated then...

Also, I know 75 files is a LOT but most of them are just changes to the way we require markdownz, so that should reduce the review burden some hopefully!

brian-c:
I suggested a rebase because it looks like there's some changes being reverted. See world_wide_telescpe.cjsx and project/classify.cjsx for example. Rebasing manually shows conflicts in package.json.

amyrebecca:
oh weird, you're right. i'm not sure how to deal with those exactly? especially considering the fact that i have some merge commits in there...

the package.json file has changed a ton so the conflicts there are no surprise

amyrebecca:
I guess I can just checkout those files from the appropriate revisions and re-commit them, would that work?

brian-c:
I think a rebase on origin/master will do the trick.

amyrebecca:
i'll give it a shot :)

amyrebecca:
that worked, i note that the number of files modified has decreased by two :3

amyrebecca:
i'm going to pull out the .htaccess stuff since it looks like nginx doesn't support them (even for compatibility) but how do you feel about keeping the config separate. i actually quite liked them that way but it's fine if you want me to combine them.

brian-c:
I don't understand how we benefit from having two separate configs. Why make it possible to break something and not realize it until it deploys? Shouldn't the dev environment be as close as humanly possible to production?

amyrebecca:
well the reason that i kept them separate is that they have two different concerns:

* the dev config has a lot of stuff that's specific to file-watching, hot reloading, debugging helpers, etc
* the build config has a lot of stuff that's specific to deployment of the transpiled code, optimization, etc

i mean i get that it's always iffy to have two different configs but even if i consolidated them into a file there'd be a big switch statement based on whether or not your NODE_ENV is development or staging/production.

brian-c:
Finally actually gave this a try. I got this error when trying to load the page after `npm start`:

```
Error: no such file or directory
 at MemoryFileSystem.readFileSync (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/webpack-dev-middleware/node_modules/memory-fs/lib/MemoryFileSystem.js:107:10)
 at /Users/brian/Zooniverse/Panoptes-Front-End/server_dev.js:43:35
 at Layer.handle [as handle_request] (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/lib/router/layer.js:95:5)
 at trim_prefix (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/lib/router/index.js:312:13)
 at /Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/lib/router/index.js:280:7
 at Function.process_params (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/lib/router/index.js:330:12)
 at next (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/lib/router/index.js:271:10)
 at SendStream.error (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/node_modules/serve-static/index.js:120:7)
 at emitOne (events.js:77:13)
 at SendStream.emit (events.js:169:7)
 at SendStream.error (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/node_modules/send/index.js:245:17)
 at SendStream.onStatError (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/node_modules/send/index.js:356:12)
 at next (/Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/node_modules/send/index.js:658:28)
 at /Users/brian/Zooniverse/Panoptes-Front-End/node_modules/express/node_modules/send/index.js:666:23
 at FSReqWrap.oncomplete (fs.js:82:15)
```

Anything I'm missing?

amyrebecca:
that's weird, i'll take a look at it and figure out what's wrong. i assume you already did `npm install` or else i'd think that was the reason

brian-c:
Yeah, I trashed node_modules and `npm install`ed a couple times. When webpack is done bundling, there's no dist directory created, which I think is wrong. Creating one manually doesn't help. I've got Node 4.2.3 and npm 2.15.3.

amyrebecca:
good idea to check my versions! i also have an idea why dist/ might not be getting created and it's not hard to solve :)

amyrebecca:
also if you have a chance, can you clarify whether we should be worried about the behavior of

https?://branch-name-panoptes-front-end.zooniverse.org/

or

https?://preview.zooniverse.org/panoptes-front-end/branch-name/

because i was working on the latter the whole time and it was quite difficult to make it work! and i haven't tested to see whether the former behaves properly or not

brian-c:
Ha, looks like staging URLs have changed again since I last used them. Currently they're `http://branch-name.pfe-preview.zooniverse.org/`

I do know that the site now expects to *only* ever be deployed to a server root. This was not always the case.

amyrebecca:
THAT would explain why I had so much trouble getting it to work from anywhere other than server root!

amyrebecca:
Is there a default location for staging to build if I'm not building a branch?

brian-c:
I _believe_ this is the current behavior of our deploy scripts:

| Command | Bucket | Path prefix | Resulting URL |
| --- | --- | --- | --- |
| `npm run stage` | zooniverse-static | preview.zooniverse.org | https://pfe-preview.zooniverse.org/
| `npm run stage-brach` | zooniverse-static | preview.zooniverse.org | https://branch.pfe-preview.zooniverse.org/
| `npm run deploy-branch` | zooniverse-static | www.zooniverse.org/branch | https://www.zooniverse.org/branch **BROKEN**
| `npm run deploy` | zooniverse-static | www.zooniverse.org/ | https://www.zooniverse.org/

amyrebecca:
I didn't want to test deploy without folks around since like... it could break everything. And deploy-branch still doesn't seem to work, but i think that's a configuration issue and not a code issue? But local stuff works, stage works, stage-branch works, etc. I think this should be ready to go.

amyrebecca:
to use webpack: npm start
to check precompiled stuff: npm run prep && npm run serve
to stage a branch: npm run stage-branch
to build staging: npm run stage
to deploy a branch: npm run deploy-branch
to deploy to production: npm run deploy

views/index.ejs
```diff
@@ -2,14 +2,14 @@
 
 <html>
   <head>
+    <base href="<%= htmlWebpackPlugin.options.baseUrl %>" />
```
brian-c: Duplicated `<base />` tag. I suggest we drop all this extra `baseUrl` and `useBasePath` stuff for now. I think it's worth fixing, but this doesn't do it and this PR is the wrong place for it anyway.

.gitignore
```diff
@@ -9,4 +9,7 @@
 /public/vendor.js
 /public/jsPlumb.min.js
```
brian-c: Is the public dir still built to? I think we can clear a lot of these out.

app/pages/project/home.cjsx
```diff
@@ -30,6 +30,7 @@ module.exports = React.createClass
 
   render: ->
     [owner, name] = @props.project.slug.split('/')
+    console.log Markdown
```
brian-c: Log slipped through.

amyrebecca:
Working now to address notes from brian-c

brian-c:
(Needs a rebase, sorry. To make things easier in the future, maybe we use branches instead of forks so the review can push up minor changes like that.)

This looks good to me, but I'd love some more eyes on it before we merge. @srallen @eatyourgreens @parrish?

srallen:
I can do a pass at reviewing this tomorrow.

parrish:
Yep, I'll give it a shot too.

eatyourgreens:
Do we still need the `deploy-branch` script? Branches can be tested against production data with the `env` query parameter  `https://branch-name.pfe-preview.zooniverse.org?env=production`

brian-c:
Agreed, `deploy-branch` is misleading to keep around since it doesn't do anything useful.

amyrebecca:
i'll rebase and remove the deploy-branch stuff :)

srallen:
Question about the build: do we want to consider adding image minification? I don't think this was happening before, but can easily be added into the build config now.

amyrebecca:
I think it'd definitely be a good idea, but maybe something we could build subsequent to this? It's already such a huge PR!

srallen:
@amyrebecca Agreed that it can be a PR later.

srallen:
After deleting `node_modules` and re-installing all of the modules a couple of times, I'm getting a console error: `Uncaught TypeError: history.listen is not a function` throwing on https://github.com/amyrebecca/Panoptes-Front-End/blob/master/app/main.cjsx#L19

Looks like `history` version 1.17.0 installed successfully.

srallen:
I think the issue is that I don't have `main.js` in the `dist` folder. It should be compiled and copied into `dist, yes?

amyrebecca:
That's weird, I've never seen that. If you're running it in dev mode, then it should be serving that stuff from the in-memory filesystem rather than from disk, so it's not necessarily a problem that the file is missing for the moment. I'll dig into this.

parrish:
I can confirm, I see the same `history.listen` error.

I'm on node v4.3.0 (npm v2.14.12) if that's helpful.

amyrebecca:
Wouldn't be surprised if something got murdered in a rebase, etc. I'll let y'all know what I find and I appreciate the repro steps

amyrebecca:
It's fixed. I messed up a commit where I was removing some of the stuff I'd put in to try to get deploy-branch working. Sorry to waste your time; it should be fine now.

parrish:
Yep.  It's working now.  I'll spend some time tomorrow working through it.

srallen:
Cool, `npm start` works now. Checking `npm run prep && npm run serve`, I get this error: `pfePrefix is not defined` on line 18 in the `webpack.build.js` file.

parrish:
This is working well for me!

##### (begin opinion)

Livereload is a problem for me though
  - It tends to create a fragile dev environment
  - It intermittently requires refreshes or server restarts to get accurate builds
    - You lose time by not finding bugs or by chasing bugs from livereload rather than the app
  - I'd rather leave that tool choice up to the individual developer

##### (end opinion)

Also, in case anybody wants to start testing ES6 from CoffeeScript, you'll need to backtick the import:

``` coffeescript
# some-file.cjsx
`import Test from './test';`
```

``` javascript
// test.jsx
import React from 'react';

export default class Test extends React.Component {
  render() {
    return <p>Works!</p>;
  }
}
```

amyrebecca:
i fixed the run prep issue, thank you for noticing that. if we want to have a discussion about the hot reloading, i understand, but it was a big part of the reason most of us thought it was worth migrating in the first place. 

disabling it should be as simple as removing line 21 from server_dev.js and removing line 11 from webpack.dev.js

srallen:
I think this is close. The only thing I'm seeing now is that `check-engines` is missing from this? Do we still want to run that as a part of `npm start`? 

For live reloading, I can definitely see it potentially be more of a hinderance than a help. Perhaps we can give it a trial run and then check back in a month or two for a consensus on it? We could consider switching to [using the CLI](https://webpack.github.io/docs/webpack-dev-server.html#hot-module-replacement-with-inline-mode-on-cli) and a separate npm script for it instead of being in the config?

amyrebecca:
we can probably generate configs that don't use live reload, use those for npm start, and then use npm run dev for live reload. i keep removing check_engines inadvertently because it keeps throwing errors on my machine and then i forget to change it back before i commit, lol

brian-c:
> i keep removing check_engines inadvertently because it keeps throwing errors on my machine and then i forget to change it back before i commit

Is it because you're using the wrong Node version, or is something broken?

In your last commit: `check-engines` should be called with `&&` so it stops the rest of the script from running if the check fails.

amyrebecca:
i'm not really sure which node version i'm running, i don't pay attention to it. as long as the check is advisory, it doesn't give me any trouble, which is why i used ;

brian-c:
It's not really advisory, it's deliberately set so we're all running the code against same version.

amyrebecca:
ah, i guess i'll have to take some time and get nvm or something working then. i didn't realize we were concerned about that because node's usually pretty good about forward/backward compatibility.

srallen:
Many of us have multiple versions of node installed and use a version manager; Most older Ouroboros projects use node 0.10.x or 0.12.x. It's helpful to have a check to make sure you're switched to the correct version.

srallen:
I think this is ready now. LGTM, unless anyone else sees anything?

package.json
```diff
@@ -42,34 +41,60 @@
...
+    "copy-webpack-plugin": "~2.1.3",
+    "css-loader": "~0.23.1",
     "ejs": "~2.3.3",
-    "envify": "~3.4.0",
     "express": "~4.13.3",
-    "nib": "~1.1.0",
-    "publisssh": "~1.0.0",
+    "extract-text-webpack-plugin": "~1.0.1",
+    "file-loader": "~0.8.5",
+    "fs": "0.0.2",
```
brian-c: What is this?

package.json
```diff
@@ -42,34 +41,60 @@
...
+
+    "build": "check-engines && webpack --config ./webpack.build.js --progress --profile --colors",
+    "clean": "rimraf dist",
+    "prep": "export NODE_ENV=staging; npm run clean && npm run build",
+    "publish": "publisssh ./dist \"zooniverse-static/${BUCKET}/$DEPLOY_SUBDIR\" $PUBLISH_FLAGS",
+    "serve": "export NODE_ENV=staging; check-engines && node server.js",
+    "start": "export NODE_ENV=development; check-engines && babel-node server_dev.js",
+
+    "_deploy": "export BUCKET=www.zooniverse.org; npm run clean && npm run build && npm run publish",
+    "deploy": "export NODE_ENV=production; export DEPLOY_SUBDIR=panoptes-front-end; export PUBLISH_FLAGS=\"-d\" npm run _deploy",
```
brian-c: Isn't `-d` `--dry-run`? Probably left over. (Let's use long flags when they're available.)

package.json
```diff
@@ -42,34 +41,60 @@
...
-    "postinstall": "./bin/postinstall.sh",
-    "deploy": "npm run build && publisssh ./build zooniverse-static/www.zooniverse.org/$DEPLOY_SUBDIR",
-    "deploy-branch": "export DEPLOY_SUBDIR=$(git symbolic-ref --short HEAD); npm run deploy",
-    "stage": "export DEPLOY_SUBDIR=${DEPLOY_SUBDIR:-panoptes-front-end}; export NODE_ENV=staging; npm run build && publisssh ./build \"zooniverse-static/preview.zooniverse.org/$DEPLOY_SUBDIR\"",
-    "stage-branch": "export DEPLOY_SUBDIR=\"panoptes-front-end/$(git symbolic-ref --short HEAD)\"; npm run stage;  echo \"Deployed to http://$(git symbolic-ref --short HEAD).pfe-preview.zooniverse.org/\";",
-    "stage-branch-with-docker": "./bin/run-through-docker.sh 'npm run-script stage-branch'",
+
+    "build": "check-engines && webpack --config ./webpack.build.js --progress --profile --colors",
+    "clean": "rimraf dist",
+    "prep": "export NODE_ENV=staging; npm run clean && npm run build",
```
brian-c: When is the prep script used?

package.json
```diff
@@ -42,34 +41,60 @@
...
+    "publish": "publisssh ./dist \"zooniverse-static/${BUCKET}/$DEPLOY_SUBDIR\" $PUBLISH_FLAGS",
+    "serve": "export NODE_ENV=staging; check-engines && node server.js",
```
brian-c: When is the serve script used?

brian-c:
Just snuck in some last minute issues before merge. The non-built-in `fs` module is incredibly odd. I don't think it should be there. And the deploy script is set to `--dry-run`.

brian-c:
@amyrebecca I think 864a7769e0957c130f5f2fcde10f53d4e7cd9eaa is missing the `-d` removal from package.json. We all good after that?

PR #2525, opened 2016-05-04
Clean up display of owned cards (collections, favourites etc)

alexbfree:
When project cards and owned cards were separated, the CSS was not cleanly split and still shared the same CSS class. Possibly because of this, or for another reason, owned cards are wrongly stretched vertically to 175px. This change separates out the CSS classes to avoid accidental collisions like this in future, and stops the 175px height from applying to owned cards

Also adding some space below collection titles as the text was too close to the button.

Changes owned cards from this
screenshot 2016-05-04 11 54 58

to this
screenshot 2016-05-04 11 54 24

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -55,7 +55,7 @@ module.exports = React.createClass
                   count = meta.count
                   <Translate pageStart={pageStart} pageEnd={pageEnd} count={count} content="#{@props.translationObjectName}.countMessage" component="p" />}
               </div>
-              <div className="card-list">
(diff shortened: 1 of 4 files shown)


srallen:
@jwbmartin Just to check in, are you working on any designs for collections or have any comment about this?

alexbfree:
I think there are definitely more improvements we can do when showing collections, think it should be fine to merge this as a small incremental change in the meantime though....

srallen:
@alexbfree LGTM, but can you rebase? Thanks.

alexbfree:
Rebased, merging per @srallen's approval above.

Issue #2507, opened 2016-04-28
dropdown task - preset options


mcbouslog:
Continues #2376...

This issue is to discuss the dropdown task preset options.
See Issue #2385 for discussion related to the dropdown task options import.

@brian-c, @simensta, @aliburchard

mcbouslog:
The following preset options have been added and are available to be applied when building a dropdown task's options:

- Countries
- US states
- Canada provinces
- Mexico states
- months
- days (in month)

Currently I'm working on a branch in my forked PFE repo, located [here](https://github.com/mcbouslog/Panoptes-Front-End/tree/dropdown-presets). 
I've also staged the branch [here](https://dropdown-presets.pfe-preview.zooniverse.org/). Go to the Project Builder and try building a dropdown task (after enabling via Admin experimental options). 

Please let me know what you think, thanks!

PR #2527, opened 2016-05-04
add preset options to dropdown task

mcbouslog:
The following preset options are available to be applied when building a dropdown task's options:

-Countries
-US states
-Canada provinces
-Mexico states
-months
-range of numbers (could be used for years, days, etc.)

Code is on a branch in my forked PFE repo, located [here](https://github.com/mcbouslog/Panoptes-Front-End/tree/dropdown-presets). 
Branch is staged [here](https://dropdown-presets.pfe-preview.zooniverse.org/). Enable the experimental dropdown task in a project to test adding preset dropdown options.

I'm anticipating some edits, please let me know what you think, no recommendation too small (like renaming `data` folder to `presets`?). Thanks!

closes Issue #2507


app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -182,6 +230,34 @@ DropdownDialog = React.createClass
 
       <br />
 
+      <div>
+        <h2 className="form-label">Presets</h2>
+        <select ref="preset" onChange={@handlePresetChange}>
```
brian-c: We're already setting state for `numberRange`, why not just keep the value of this select as state, since we eventually change it directly?

app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -182,6 +230,34 @@ DropdownDialog = React.createClass
...
+          <option value="Countries">Countries</option>
+          <option value="statesUSA">States - United States</option>
+          <option value="provincesCanada">Provinces - Canada</option>
+          <option value="statesMexico">States - Mexico</option>
+        </select>
+
+        {if @state.numberRange
+          <div>
+            <label>Top number:
+              <input type="number" ref="topNumber" min="0" max="999999" />
```
brian-c: Why set a min/max on these?

app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -97,6 +105,46 @@ DropdownDialog = React.createClass
...
+    @setState editSelect: select
+    @setState numberRange: false
+    @refs.preset.value = ""
+
+  handleNumberRange: (select, optionsKey) ->
+    topNumber = @refs.topNumber.value
+    bottomNumber = @refs.bottomNumber.value
+
+    options = []
+    for num in [topNumber..bottomNumber]
```
brian-c: The `value` of an HTML input is always a string, so this is a range between two strings. It works because JavaScript coerces the values into numbers, but it's prettyyy weird. Maybe `parseFloat` those values first.

brian-c:
Yeah I think "presets" makes more sense. Also style-wise, file names are dash-separated, and variable names starting with a capital letter are reserved for classes.

I've left some comments in-line. In general, modifying the DOM directly (i.e. `@refs.someInput.value = ''`) should be avoided in favor of setting state and rendering with the right value.

mcbouslog:
thanks Brian!

1. i think i properly changed the preset select to use state instead of @refs, please take a look and confirm. i'll add a comment on line 89 regarding a different, but similar scenario.
2. removed min/max, no good reason to have
3. parseFloat appears to work great
4. renamed folders, files and variables

app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -77,7 +85,7 @@ DropdownDialog = React.createClass
     @setState editSelect: select
 
   onClickAddOption: ->
-    @addOption(@refs.optionInput.value)
+    @addOption @refs.optionInput.value
     @refs.optionInput.value = ''
```
mcbouslog: should this also be updated to use state instead of refs? i'm fairly certain i know how to, but in my head it looks like more code so i'm not sure if this is a scenario where editing the ref/DOM directly is ok? editing state instead of the ref/DOM directly makes total sense though.

brian-c: Sorry for the delay. I think this is fine here. Manipulating the DOM directly in a React component is very unusual, I'd avoid it in the future.

PR #2528, opened 2016-05-04
manage, edit and create workflows in Project Builder rgdless if project live

mcbouslog:
1) adds same workflow active/inactive management currently in admin settings of a project to Visibility section within Project Builder
2) allows editing of inactive workflows
3) allows creation or copying (within project) of new workflows, which are set as active if project in development (current setup) or inactive if project live.

a new workflow active status is set within the workflow-create-form.cjsx. i believe it's duplicative to set the workflowActiveStatus prop's default to false ([line 12](https://github.com/mcbouslog/Panoptes-Front-End/blob/467d8c28c147b39587febd6bd7bae5c8eef24ca7/app/pages/lab/workflow-create-form.cjsx#L12)) and set the active property to false if the prop is undefined ([line 39](https://github.com/mcbouslog/Panoptes-Front-End/blob/467d8c28c147b39587febd6bd7bae5c8eef24ca7/app/pages/lab/workflow-create-form.cjsx#L39)), but maybe it's worth it to make sure a workflow isn't unintentionally added as active?

also, i considered removing the workflow management from admin project settings, but because all admins have quick access to all projects' admin settings, but only collaborators have access to the Project Builder, Visibility section, i thought ok to leave in both places so an admin unassociated with a project can quickly manage workflows if need be, without being added as a project collaborator.

per Laura's suggestion, mentioning @aliburchard and @camallen who may be interested in this?

branch staged for review [here](https://edit-inactive-workflows.pfe-preview.zooniverse.org)

### DIFF
--- a/app/components/workflow-toggle.cjsx
+++ b/app/components/workflow-toggle.cjsx
@@ -1,0 +1,29 @@
+React = require 'react'
+SetToggle = require '../lib/set-toggle'
+
+module.exports = React.createClass
+  displayName: "WorkflowToggle"
+
+  mixins: [SetToggle]
+
+  getDefaultProps: ->
+    workflow: null
+    project: null
+    field: null
+
+  getInitialState: ->
+    error: null
+    setting: {}
+
+  setterProperty: 'workflow'
+
+  render: ->
+    workflow = @props.workflow
+    setting = workflow[@props.field]
+    <span>
+      { workflow.id } - { workflow.display_name}:
+      <label style={whiteSpace: 'nowrap'}>
+        <input type="checkbox" name={@props.field} value={setting} checked={setting} onChange={@set.bind this, @props.field, not setting} />
+        Active
+      </label>
+    </span>
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -57,33 +57,6 @@ ProjectToggle = React.createClass
       </label>
     </span>
 
-# WorkflowToggle = React.createClass
-#   displayName: "WorkflowToggle"
-#
-#   mixins: [SetToggle]
-#
-#   getDefaultProps: ->
-#     workflow: null
-#     project: null
-#     field: null
-#
-#   getInitialState: ->
-#     error: null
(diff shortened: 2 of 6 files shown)


aliburchard:
@mcbouslog omg this is freaking exciting. I can't actually test it, but I'm excited to see it in action.

alexbfree:
Excited to see this. Could you deploy to a staging URL so we can try it?

mcbouslog:
just re-staged, the link Andrea posted is the latest and greatest

simensta:
@mcbouslog, this PR looks good to me. Nice work! 


RE: Defaulting `workflowActiveStatus` to `false` in [`handleSubmit`](https://github.com/mcbouslog/Panoptes-Front-End/blob/467d8c28c147b39587febd6bd7bae5c8eef24ca7/app/pages/lab/workflow-create-form.cjsx#L39). 
The default value provided in `getDefaultProps` will be used if the parent component passes `undefined` as the value for `workflowActiveStatus`. However, the default value will not be used if the parent component defines `workflowActiveStatus` as `null`. I believe [Line 39](https://github.com/mcbouslog/Panoptes-Front-End/blob/467d8c28c147b39587febd6bd7bae5c8eef24ca7/app/pages/lab/workflow-create-form.cjsx#L39) will set `workflowActiveStatus` to `false`, if `@props.workflowActiveStatus` is either `null` or `undefined`.

alexbfree:
Tested quickly on the staging branch, looks good.

I do have one small concern. Now that we have three different types of "state" exposed to project builders : Dev/Live , Public/Private and Active/Inactive, I wonder if project builders may find this confusing.

I think we should seek out some specific feedback on this from project builders (though of course that could be done after we roll it out).

mcbouslog:
Thanks! And I totally agree, while it's nice to give people control/access to so many states, it's also confusing, and I'd imagine especially so for someone just starting to use the Project Builder. I'm not sure what the solution is though, but thinking just like you mentioned - we wait and see what feedback we get.

simensta:
Okay, I think this is ready to merge.

PR #2533, opened 2016-05-05
Set window object to variable

srallen:
Apparently assigning the window object the require directly was throwing a `zooAPI is undefined` error in the console.


brian-c:
I'm fine with this but I'd like to understand what the cause was. How can I reproduce that error?

srallen:
I'm seeing it on zooniverse.org in the browser console using Chrome when you try to use the apiClient:

screen shot 2016-05-05 at 11 58 50 am

brian-c:
Hm, that's super strange. That happens on master for you? This change shouldn't really do anything at all. Might be worth some further debugging later.

PR #2538, opened 2016-05-06
Big cleanup of the root project page

brian-c:
Just realized this branch is totally named wrong, sorry. This is a refactor of the root project page. It removes any `ChangeListener` and `PromiseRenderer` components and makes accessing resources more explicit.

Also set up a redirect for /projects/owner/name**/home** back to /projects/owner/name, so there's only a single URL per page. Feels kinda hacky but I just could not get react-router to behave.

Just FYI @srallen, this is in preparation for my change for storing the last selected workflow as a state on the project root page.

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -1,16 +1,24 @@
 counterpart = require 'counterpart'
 React = require 'react'
-ChangeListener = require '../../components/change-listener'
-PromiseRenderer = require '../../components/promise-renderer'
 Translate = require 'react-translate-component'
-{Link} = require 'react-router'
+{IndexLink, Link} = require 'react-router'
 {Markdown} = require 'markdownz'
-TitleMixin = require '../../lib/title-mixin'
-HandlePropChanges = require '../../lib/handle-prop-changes'
-apiClient = window.api = require 'panoptes-client/lib/api-client'
-{sugarClient} = require 'panoptes-client/lib/sugar'
-LoadingIndicator = require '../../components/loading-indicator'
 PotentialFieldGuide = require './potential-field-guide'
+TitleMixin = require '../../lib/title-mixin'
+apiClient = require 'panoptes-client/lib/api-client'
+{sugarClient} = require 'panoptes-client/lib/sugar'
+
+counterpart.registerTranslations 'en',
+  project:
+    loading: 'Loading project'
+    disclaimer: "This project has been built using the Zooniverse Project Builder but is not yet an official Zooniverse project. Queries and issues relating to this project directed at the Zooniverse Team may not receive any response."
+    nav:
+      research: 'Research'
+      results: 'Results'
+      classify: 'Classify'
+      faq: 'FAQ'
+      education: 'Education'
+      talk: 'Talk'
 
 SOCIAL_ICONS =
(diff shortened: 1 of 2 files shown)


app/pages/project/index.cjsx
```diff
@@ -27,164 +35,222 @@ SOCIAL_ICONS =
...
-            <Translate className="project-disclaimer" content="project.disclaimer" component="p" />}
-
-          <PotentialFieldGuide project={@props.project} />
-        </div>
-      }</PromiseRenderer>
-    }</ChangeListener>
-
-module.exports = React.createClass
-  displayName: 'ProjectPageWrapper'
-  mixins: [TitleMixin, HandlePropChanges]
```
srallen: Are we deprecating `HandlePropChanges`? I was starting to try that out in the workflow assignment feature I'm working on.

app/pages/project/index.cjsx
```diff
@@ -27,164 +35,222 @@ SOCIAL_ICONS =
...
-ProjectAvatar = React.createClass
-  displayName: 'ProjectAvatar'
-
-  render: ->
-    <PromiseRenderer promise={@props.project.get 'avatar'} then={(avatar) =>
-      <img src={avatar.src} className="avatar" />
-    } catch={null} />
 
 ProjectPage = React.createClass
-  displayName: 'ProjectPage'
```
srallen: Did you mean to remove the `displayName` here?

app/pages/project/index.cjsx
```diff
@@ -27,164 +35,222 @@ SOCIAL_ICONS =
...
+      <PotentialFieldGuide project={@props.project} />
+    </div>
+
+
+ProjectPageController = React.createClass
+  displayName: 'ProjectPageController'
+
+  mixins: [TitleMixin]
 
   title: ->
```
srallen: Is this being used?

srallen:
Router change looks good to me. I have a few inline questions on the project index.

app/pages/project/index.cjsx
brian-c: Yeah, I'd like to avoid it. It's an unnecessary abstraction that makes it harder to follow the component lifecycle. Any reason to keep it?

app/pages/project/index.cjsx
brian-c: Yep, this is set automatically by the JSX parser when a class is assigned to a variable.

app/pages/project/index.cjsx
brian-c: `TitleMixin` calls this to update the document title. (`TitleMixin` will eventually be replaced with a wrapper component.)

app/pages/project/index.cjsx
srallen: Nope. I was trying to be consistent with what I was seeing be done in the files I was working in. Good to know.

app/pages/project/index.cjsx
srallen: Ahh, for some reason I thought that was something Babel did with JSX.

app/pages/project/index.cjsx
srallen: Cool. Thanks.

Issue #2536, opened 2016-05-06
Need to strip images from tool labels when using in classification summary


alexbfree:
As you can see in this screenshot from https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/, we end up with something rather ugly 

![screenshot 2016-05-06 12 09 45](https://cloud.githubusercontent.com/assets/1473244/15071449/8fd53c62-1383-11e6-876c-7cc2fb4685a1.png)

We could render the image, but TBH i think just stripping them out would be better

Any project that uses images in labels, will suffer the same issue. Her is another: https://www.zooniverse.org/projects/klmasters/galaxy-zoo-for-schools-outreach/classify

alexbfree:
While we're in there, let's clean up the text a bit:

How about:
`Animal facing to the left (2 rectangles)`
rather than
`rectangle Animal facing to the left (2)`

alexbfree:
OK I found a few minutes, i have a fix for this... PR here https://github.com/zooniverse/Panoptes-Front-End/pull/2560/files

alexbfree:
A note about the fix: it requires you to specify your links like this:

```
![](https://panoptes-uploads.zooniverse.org/production/project_attached_image/9caf156d-5952-45f1-8787-81451ac4eb7f.png)
```
not this:
```
![animal_facing_left_1.png](https://panoptes-uploads.zooniverse.org/production/project_attached_image/9caf156d-5952-45f1-8787-81451ac4eb7f.png)
```
or you will still see the filename (ie link text) in the label

eatyourgreens:
@alexbfree so remove any alt text from the images? I'm not sure about that.

alexbfree:
@eatyourgreens No - we are entirely removing the images.

Removing the link label ensures that nothing survives of the images - which is what we want in this context. (without that, the image link in full markdown gets replaced with just the link label)

alexbfree:
@eatyourgreens you can try it out here on this dummy project > https://cleanup-drawing-summary.pfe-preview.zooniverse.org/projects/alexbfree/untitled-project-5-10-2016-4-22-18-pm/classify

eatyourgreens:
I understand, but your markdown example will render as `<img alt="" src="https://panoptes-uploads.zooniverse.org/production/project_attached_image/9caf156d-5952-45f1-8787-81451ac4eb7f.png">` in places where the image is rendered as HTML ie. in the classification questions.

[...]

PR #2560, opened 2016-05-10
Remove any markdown from tool labels when summarising classification

alexbfree:
Fixes https://github.com/zooniverse/Panoptes-Front-End/issues/2536 and improves classification summary for drawing tasks.

### DIFF
--- a/app/classifier/tasks/drawing/summary.cjsx
+++ b/app/classifier/tasks/drawing/summary.cjsx
@@ -11,6 +11,7 @@ module.exports = React.createClass
     expanded: false
 
   stripMarkdownFromLabel: (label) ->
+    label = label.replace /\!\[[^\]]*\]\([^)]*\)/g, ""
     remark.process(label)
 
   getInitialState: ->
--- a/app/classifier/tasks/drawing/summary.cjsx
+++ b/app/classifier/tasks/drawing/summary.cjsx
@@ -1,4 +1,6 @@
 React = require 'react'
+strip = require 'strip-markdown'
+remark = (require 'remark').use(strip)
 
 module.exports = React.createClass
   displayName: 'DrawingSummary'
@@ -7,6 +9,9 @@ module.exports = React.createClass
     task: null
     annotation: null
     expanded: false
+
+  stripMarkdownFromLabel: (label) ->
+    remark.process(label)
 
   getInitialState: ->
     expanded: @props.expanded
@@ -31,7 +36,7 @@ module.exports = React.createClass
         toolMarks = (mark for mark in @props.annotation.value when mark.tool is i)
         if @state.expanded or toolMarks.length isnt 0
           <div key={tool._key} className="answer">
-            {tool.type} <strong>{tool.label}</strong> ({[].concat toolMarks.length})
+            <strong>{@stripMarkdownFromLabel(tool.label)}</strong> ({[].concat toolMarks.length} {tool.type}s marked)
             {if @state.expanded
               for mark, i in toolMarks
                 mark._key ?= Math.random()
--- a/app/classifier/tasks/drawing/summary.cjsx
+++ b/app/classifier/tasks/drawing/summary.cjsx
@@ -10,6 +10,9 @@ module.exports = React.createClass
     annotation: null
     expanded: false
 
+  getCorrectSingularOrPluralOfDrawingType: (type, number) ->
+    if number>1 then "#{type}s" else type
+
   stripMarkdownFromLabel: (label) ->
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/drawing/summary.cjsx
```diff
@@ -1,4 +1,6 @@
 React = require 'react'
+strip = require 'strip-markdown'
+remark = (require 'remark').use(strip)
```
eatyourgreens: Could we do the processing with our existing Markdown library, rather than importing a second Markdown parser?

alexbfree: fair point. which markdown parser do we use? i will have to figure out what .use is doing.

alexbfree: I looked into this. Looks like `strip-markdown` is a specific plugin written for `remark`. We use https://github.com/zooniverse/markdownz which is built upon https://github.com/markdown-it/markdown-it. While `markdown-it` does have [plugins](https://www.npmjs.com/browse/keyword/markdown-it-plugin), there doesn't seem to be one equivalent to `strip-markdown`. So there isn't an easy way to avoid using `remark` without writing our own version of `strip-markdown` to work on `markdown-it`. I think we will have to put up with inconvenience of loading `remark` for now. It's only for this one single `.cjsx` file, not the whole classifier....

alexbfree:
code is deployed to staging and can be tested using Computer Vision: Serengeti here : https://cleanup-drawing-summary.pfe-preview.zooniverse.org/projects/alexbfree/computer-vision-serengeti/classify?env=production

alexbfree:
Did a quick update so that it will say
"1 rectangle marked" or "2 rectangles marked" rather than always having the 's'
Deployed to the staging branch too.

app/classifier/tasks/drawing/summary.cjsx
eatyourgreens: I had a play around with passing the `zero` option to `markdown-it` before parsing the question, but that just passes back the raw markdown text. Not hugely useful.

Seems a pain to require browsers to download `remark` just for this one line of code. Long term, I'd like to look at extending `markdownz` to return the plain text content of a markdown string.

app/classifier/tasks/drawing/summary.cjsx
```diff
@@ -8,6 +10,13 @@ module.exports = React.createClass
     annotation: null
     expanded: false
 
+  getCorrectSingularOrPluralOfDrawingType: (type, number) ->
+    if number>1 then "#{type}s" else type
+
+  stripMarkdownFromLabel: (label) ->
+    label = label.replace /\!\[[^\]]*\]\([^)]*\)/g, ""
+    remark.process(label)
```
eatyourgreens: This will return empty if an answer for a question is just an image, so we may have to revisit this in future.

app/classifier/tasks/drawing/summary.cjsx
srallen: Agreed. I don't think we should add a dependency just for this. Could we consider just rendering the markdown in the summary, too?

eatyourgreens: Sorry, saw this after I'd merged the PR. Another option would be to render markdown in the summary, then style summary images with `display:none`. Or just leave them in.

app/classifier/tasks/drawing/summary.cjsx
alexbfree: Good point. Didn't think of that. Let's wait and see if that ever happens. Thanks for the merge.

app/classifier/tasks/drawing/summary.cjsx
alexbfree: i am not a fan of rendering the markdown for the tool label in the summary. I think it is much less succinct. Though I realise this is a subjective opinion. Of course I am thinking of my primary use case with Computer Vision: Serengeti and how I am using graphics in tool labels there. Other people might use them differently but in my case the graphic is really help for the classify page, not part of the tool's name.
I like the succinctness of "Animal facing to the left (2 rectangles marked)" in the summary.

I think this is a bit different from when there is a graphic in the answer - as seen in https://www.zooniverse.org/projects/klmasters/galaxy-zoo-for-schools-outreach/classify (notice we render the image in the summary there).

I agree the point that using another dependency isn't great. However neither is writing a custom parser.

I suppose now that we have the regex to remove the image tags, I can remove the calls to strip markdown, since that was only left in to remove other formatting. Which in most cases isn't half as important. I'll do another PR to do that.

PR #2561, opened 2016-05-10
Improve solution to cleanup drawing summary

alexbfree:
Following feedback on #2560 / #2536 - Remove the third party dependency. More details in the commit comment


alexbfree:
assigning to @eatyourgreens for review/merge - am addressing @srallen's comment and am removing 3rd party dependency. We don't really need it in this case anyway now we have the regex.

PR #2508, opened 2016-04-28
autofocus task navigation

simensta:
This PR improves the keyboard navigation of the following tasks in the `/classify` page:
- text tool
- single answer
- drawing
- multiple answer

This should apply to regular tasks, as well as subTasks. Here is the example branch with a [simple workflow ](https://autofocus-task-navigation.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?workflow=2304)and a [combo-task workflow](https://autofocus-task-navigation.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?workflow=2305). Fixes #2377. 

~~WIP~~
- [x] issues in Safari (solution: turn on keyboard navigation in system preferences)

@eatyourgreens, @mcbouslog -- thoughts?

PR #2564, opened 2016-05-11
[Do not merge] Testing revert of "autofocus task navigation"

### LINKED ISSUE #2563: First option button now has a blue halo around the white check part - looks like the interface is recommending a default answer
The interface is now putting a blue halo around the white part of the button on check bock for the first response now for every question task. It makes it look like there's a default or preferred/recommended answer which is not true in my case that the first is typical especially for the multiple choice/multiple answer ones. Can that please be taken off, I think this will bias people's classifications with people thinking that the interface is hinting at the default or typical option

screen shot 2016-05-10 at 4 33 15 pm
screen shot 2016-05-10 at 4 32 58 pm

screen shot 2016-05-10 at 4 33 06 pm

screen shot 2016-05-10 at 4 36 06 pm

only when you select a response does the blue outline in the white area go away for the first answer button

screen shot 2016-05-10 at 4 37 02 pm

alexbfree:
Just testing a revert of zooniverse/Panoptes-Front-End#2508 to see if it fixes #2563

### DIFF
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -66,7 +66,7 @@ module.exports = React.createClass
           {for detailTask, i in toolProps.details
             detailTask._key ?= Math.random()
             TaskComponent = tasks[detailTask.type]
-            <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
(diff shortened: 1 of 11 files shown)


alexbfree:
Confirmed by testing https://revert-2508-autofocus-task-navigation.pfe-preview.zooniverse.org/projects/mschwamb/comet-hunters/classify that this PR will indeed undo the changes and fix #2563 if we want to merge it.

alexbfree:
Cancelling this, @simensta has a better approach coming soon in a new PR.

PR #2566, opened 2016-05-11
removes autoFocus on first answer for single and multiple answer tasks

simensta:
This PR address Issue #2563 by removing the autoFocus attribute from the `<SingleChoiceTask />` and the `<MultipleChoiceTask />`.

Examples:
- [Sample workflows with multiple task types](https://limit-autofocus-fix-2563.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project)

The 'Next' button,  not the 'Talk' button, is still focused at the end of the classification. Does that also need to be changed?

CC: @eatyourgreens, @alexbfree, @aliburchard, @mschwamb


alexbfree:
FWIW I think focussing Next at the end is OK. Given a choice between classifying more and going to talk, we'd probably prefer the former.

Issue #2534, opened 2016-05-05
Workflow assignment


srallen:
- [x] Add setting to make a workflow default, written to project config
- [x] If workflow is default, load it and write to user preferences, otherwise continue to random selection
- [x] Load user project preferences, add listener for changes
- [x] If preferences include a workflow assignment that is not the current workflow, load pop up dialog to allow user to choose
- [x] Add workflow assignment preference to user project preferences if user switches
- [x] Show button choice on project home page so user can switch at a later time and according to their level
- [ ] Add option to opt out in user's profile

srallen:
Everything is implemented but the opt out, but that can be a separate issue if the project team thinks we need it.

PR #2572, opened 2016-05-11
Store workflow in prefs

### LINKED ISSUE #2565: Keep track of the last selected workflow
###### Defining some new workflow selection behavior:

When the user loads a project landing page
- If the user's project preferences have a `selected_workflow`, the classify link will point to `/classify?workflow=123`.
- Else if the project has a `default_workflow` configuration, the classify link will point to `/classify?workflow=456`.
- Otherwise it'll point to `/classify`.

The individual workflow links on the landing page (if available via the `user_chooses_workflow` project configuration) will set the user's `selected_workflow` preference.

When the user loads `/classify?workflow=123`, only subjects in that workflow are loaded.

When the user loads `/classify` (with no `?workflow=` specified)  a random workflow is chosen for each new subject.

@mrniaboc Does this sound like what you've requested?

Spoke with @srallen re: #2534, I don't think we're in conflict, but let's not work on them at the same time.

brian-c:
~~WIP for~~ Closes #2565.

The user's workflow selection from the project landing page is now stored in their project preferences and passed into the "Classify" link in the project nav. Not-signed-in users get a temporary project preferences resource that's thrown out when they switch projects or reload the page.

~~I want to play with it a little more, but~~ I think this is just about ready. Any thoughts @mrniaboc? @srallen, is this something you can work with for #2534?

### DIFF
--- a/app/pages/project/home.cjsx
+++ b/app/pages/project/home.cjsx
@@ -28,9 +28,11 @@ module.exports = React.createClass
   getInitialState: ->
     workflows: []
 
-  render: ->
-    [owner, name] = @props.project.slug.split('/')
+  handleWorkflowSelection: (workflow) ->
+    @props.onChangePreferences 'preferences.selected_workflow', workflow?.id
+    undefined # Don't prevent default Link behavior.
 
+  render: ->
     <div className="project-home-page">
       <div className="call-to-action-container content-container">
         <FinishedBanner project={@props.project} />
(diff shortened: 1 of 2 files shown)


srallen:
@brian-c I think this will work really well with what we're going to build for Gravity Spy. @amyrebecca could you also look this over let me know if there's any red flags you might see with working with Nero that I might not be considering?

mrniaboc:
This looks like it will solve the issue with the Pattern Perception project. When can we get it merged?

brian-c:
I'm happy with it if everybody else is.

amyrebecca:
This looks like it should suit our purposes for Nero really well

PR #2574, opened 2016-05-12
Bump panoptes-client again

brian-c:
With correct OAuth hosts for same-domain and cross-domain access.


brian-c:
Just kicked off a manual deploy on Jenkins. Will revert if there's any funny business again. Will let you know when it's safe to send an email, @mrniaboc.

PR #2575, opened 2016-05-12
Add Shaun to the Team page

shaunanoordin:
It's been 6 months, so it's about darn time I added my goofy face to this team I'm so proud to be part of.


shaunanoordin:
Thanks, Jim!

PR #2577, opened 2016-05-12
Mini-course and workflow tutorials

srallen:
Apologies for the size of this PR. This superceded #2399 and brings in the workflow tutorial functionality as well as mini-courses

- Mini-course pops up every X number of classifications and uses similar logic to the old sign-in prompt on the `classify` project page. Currently the mini-course pops up every 2 classifications until it is complete for easy testing.
- Mini-course has its own view in `lib` with it logic to track the which slide to start on, whether or not someone has opted out, and when the mini-course is completed. Sometime in the future, some logic may be added to have the mini-course pop up to non-signed-in users, but that was punted for later. This is why there are many `if user?` type checks
- Project builders can create multiple tutorials or mini-courses if those have been enabled by admin and linking to tutorials is done on the workflow editor page. Only one tutorial of a kind can be linked to a workflow
- I would prefer if we only fetched tutorials by workflow id, however, I left in the logic to fetch by either workflow or project since I'm not sure if the tutorial has been enabled for projects in the last month or so that this was being worked on. 
- Improved UI and styling will come in a later PR for the mini-course. Still getting feedback from folks about that, but I wanted to get the functionality done. Notes for Nature need the workflow tutorial feature next week sometime. @trouille when does this need to be in place for NfN next week?

Mini-courses and workflow tutorials can be tested at: https://mini-course.pfe-preview.zooniverse.org/projects/srallen086/srallen-testing/classify
You must be signed-in for the mini-course to work. Let me know if you want me to add you as a collaborator on the test project.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -20,7 +20,6 @@ WorldWideTelescope = require './world_wide_telescope'
 MiniCourse = require '../lib/mini-course'
 MiniCourseButton = require './mini-course-button'
 
-
 PULSAR_HUNTERS_SLUG = 'zooniverse/pulsar-hunters'
 
 Classifier = React.createClass
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -220,7 +220,7 @@ Classifier = React.createClass
       <p>
         <small>
           <strong>
-            <MiniCourseButton className="minor-button" user={@props.user} project={@props.project} title="Project Mini-course" aria-label="Show the project mini-course" style={marginTop: '2em'}>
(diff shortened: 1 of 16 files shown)


trouille:
@brian-c @srallen @mcbouslog: In an ideal world, we'd like for the Notes from Nature team to be able to use this functionality (specifically the tutorial by workflow functionality) starting this Monday (5/16), so that they have Monday & Tuesday to get the tutorials in place for applying for Beta Review on Wednesday.

brian-c:
The tutorials side of this looks real good. I understand we're still working on the minicourses. Merging now, might leave some more comments this weekend.

PR #2580, opened 2016-05-13
Adds username and tag autocompletion to talk

### LINKED ISSUE #1884: Adding the full list of popular hashtags to the side bar on the Talk subject pages
One thing that the moderators are noticing is more variability in hashtags because people don't have any guides of what the collective lexicon is becoming. I know autocompletion is in the works which would help, but also having the same list of the popular hashtags  that shows on the front page of Talk show up on the subject page sidebar. I think it give people words to chose from if they want to hashtag which  think has been a big help in previous version of Talk.

parrish:
This closes #1884

Usernames are autocompleted by login or display name with priority based on
  1. Frequently mentioned users
  2. Frequently messaged users
  3. Group names (e.g. moderator, researcher, admin)
  4. All users

By default it will show top results from priorities 1 - 3.

Tags are autocompleted by name (trigram fuzzy searching) with priority based on
  1. Similar suggested tags
    - Separate PR ~~in the works~~  in #2582 for managing suggested tags.
  2. Similarity scaled by the number of unique user usages

By default it will show the suggested tags.

The changes to comment-box are just wrapping the component in a relatively positioned container and adding the suggester.  [Cleaner diff here](https://github.com/zooniverse/Panoptes-Front-End/pull/2580/files?w=0#diff-3a7c3905094f87ae89efe8121197052a)

Also, the username completion will be pretty slow until zooniverse/panoptes#1800 is merged.

### DIFF
--- a/app/talk/comment-box.cjsx
+++ b/app/talk/comment-box.cjsx
@@ -8,6 +8,7 @@ Loading = require '../components/loading-indicator'
 SingleSubmitButton = require '../components/single-submit-button'
 alert = require '../lib/alert'
 {Markdown, MarkdownEditor} = require 'markdownz'
+Suggester = require './suggester'
 
 module?.exports = React.createClass
   displayName: 'Commentbox'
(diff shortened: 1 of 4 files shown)


brian-c:
Working through this now, but it seems to break any input on the textarea. `@state.textArea` of the Suggester isn't defined for some reason. I can see it being set on mount. Any ideas?

parrish:
Updated to fix the click event listener.  Still trying to track down the missing `textArea`

parrish:
So, it looks like `handleInput` was being called before the initial state was set [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2580/files#diff-2b566c53ad0f4451825dba7fb8bb25f9R35) as all the various props were being set [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2580/files#diff-3a7c3905094f87ae89efe8121197052aR98).  I just put a guarding `if @state.textArea` around it to ensure that the component has been setup first.

brian-c:
This is real slick. The only error I'm getting now is that `ids` is a number in `ActiveUsers::userIdsOnPage` where it expects an array. Can't reproduce it consistently though. Maybe just wrap a `[].concat(ids)` around it?

parrish:
Thanks.

Can't figure how `active-users` is impacted by this change...

I think I'll spend some time today trying to clean up various warnings/errors getting thrown these days.

If you think it's related to this, I'm happy to trap it in this PR though.

brian-c:
Cool, I'll merge this then, thanks.

parrish:
Cool!  Happy to see this go.  😎

PR #2582, opened 2016-05-13
Adds suggested tags to the lab page

parrish:
Probably doesn't make sense until #2580 is merged


app/pages/lab/talk.cjsx
```diff
@@ -13,11 +14,13 @@ module?.exports = React.createClass
   getInitialState: ->
     boards: []
     editingBoard: null
+    suggestedTags: []
 
   section: ->
     projectSection(@props.project)
 
   componentWillMount: ->
```
brian-c: Make requests in component**Did**Mount to make sure it's actually rendered.

parrish: Fixed.  Good catch, I didn't notice that when I added it.

brian-c:
Ha, just noticed I should have merged the other one first. This one stands alone well enough though, will work on the other one next.

Issue #2363, opened 2016-03-31
Add discipline tagging capability to Admin project settings page


alexbfree:
There is a need for admins to be able to apply discipline tags to projects (as currently you have to be a collaborator or nag project owners/collaborators to do it) - we should add the same new dropdown/tag field that's in the Project Builder following the merge of the new project page, to the Admin Project Settings page 
( https://www.zooniverse.org/admin/project_status/username/projectname )

alexbfree:
PR is ready (#2586)

Can be tested at http://discipline-filter-from-admin-page.pfe-preview.zooniverse.org/

srallen:
Don't think this is necessary as this is probably not a regular task we will be doing for project builders. If it does need to be done by one of us, we can do via API directly or get to lab page using admin mode.

Issue #2150, opened 2016-01-28
Admins cannot access projects they do not own or collaborate on in Project Builder


alexbfree:
This is something that came up in a recent call with Laura/Alex/Cam/Adam/Amy.

There are actually several use cases for this (some of these are just off the top of my head):
- Assigning a correct Project Discipline or other label (per the new Discipline-based project filtering)
- Generating a data export for research purposes (either for ourselves or external researcher)
- Admin may need to check workflow configuration for debugging
- Developer fixing an HTML/Formatting error on one of a project's pages, without having to bug project owner
- Quickly updating some critical error on a launched project
- ....

camallen:
looks like a dup of #1780

mschwamb:
Why does an admin page need to be added for assigning the appropriate discipline?. Can't that be checked during the review stage and confirmed when the project applies for Zooniverse approved status? If the tag is changed to a drop down menu there would be a limited set of options.  It seems to me there would be no need to have an admin to go into a project builder and  change that value

alexbfree:
thanks for linking @camallen . Had completely forgotten about raising this ticket. Linking in @brian-c who just made a comment over on #2586 suggesting this should be possible.

alexbfree:
quoting @brian-c from #2586:

>I'd be in favor of fixing the lab page for admins over duplicating functionality. It probably checking if the user is an owner or collaborator, and it just needs to check if they're an admin.
>
>Admins in admin-mode can just about anything, in case of emergencies.

Looks like that would make this a bug then, not a feature enhancement

mschwamb:
@alexbfree you've seen #1780?

alexbfree:
Thanks @mschwamb, seems we need to have some internal discussion on this to get our strategy lined up. Watch this space!

mschwamb:
I think is solved by what closed #1780?

srallen:
Yes, I think this can close.

PR #2586, opened 2016-05-16
Add Discipline filter to Admin page

alexbfree:
Fixes https://github.com/zooniverse/Panoptes-Front-End/issues/2363
Uses @amyrebecca's code from https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/project-details.cjsx

### DIFF
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -4,9 +4,12 @@ apiClient = require 'panoptes-client/lib/api-client'
 SetToggle = require '../../lib/set-toggle'
 {Link} = require 'react-router'
 moment = require 'moment'
+TagSearch = require '../../components/tag-search'
 ChangeListener = require '../../components/change-listener'
 ProjectIcon = require '../../components/project-icon'
 AutoSave = require '../../components/auto-save'
+{DISCIPLINES} = require '../../components/disciplines'
+Select = require 'react-select'
 handleInputChange = require '../../lib/handle-input-change'
 
 EXPERIMENTAL_FEATURES = [
@@ -29,6 +32,26 @@ ProjectToggle = React.createClass
   displayName: "ProjectToggle"
 
   mixins: [SetToggle]
+
+  handleDisciplineTagChange: (value) ->
+    newTags = if value is '' then [] else value.split(',')
+    @setState disciplineTagList: newTags
+    allTags = newTags.concat @state.otherTagList
+    @handleTagChange(allTags)
+
+  handleOtherTagChange: (value) ->
+    newTags = if value is '' then [] else value.split(',')
+    @setState otherTagList: newTags
+    allTags = @state.disciplineTagList.concat newTags
+    @handleTagChange(allTags)
+
+  handleTagChange: (value) ->
+    event =
+      target:
+        value: value
+        name: 'tags'
+        dataset: {}
+    handleInputChange.call @props.project, event
 
   getDefaultProps: ->
     project: null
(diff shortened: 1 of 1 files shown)


app/pages/admin/project-status.cjsx
```diff
@@ -92,6 +95,43 @@ ProjectExperimentalFeatures = React.createClass
...
+        value: value
+        name: 'tags'
+        dataset: {}
+    handleInputChange.call @props.project, event
+
+  splitTags: (kind) ->
+    disciplineTagList = []
+    otherTagList = []
+    for t in @props.project.tags
+      if DISCIPLINES.some((el) -> el.value == t)
```
camallen: Is this duplicated code somewhere? Could it be in a mix-in?

app/pages/admin/project-status.cjsx
```diff
@@ -110,7 +150,30 @@ ProjectExperimentalFeatures = React.createClass
...
+          multi={true}
+          onChange={@handleDisciplineTagChange} />
+        <small className="form-help">Enter or select one or more discipline tags to identify which field(s) of research your project belongs to. These tags will determine the categories your project will appear under on the main Zooniverse projects page, if your project becomes a full Zooniverse project. </small>
+      </AutoSave>
+      <br />
+      <AutoSave resource={@props.project}>
+        <span className="form-label">Other Tags</span>
+        <br />
+        <TagSearch name="tags" multi={true} value={@state.otherTagList} onChange={@handleOtherTagChange} />
+        <small className="form-help">Enter a list of additional tags to describe your project separated by commas to help users find your project.</small>
```
camallen: maybe some wording on how you can use spaces in your tags - i found the interface slightly non-intuitive to use spaces / commas for delimiting.

camallen:
minor comments, Looking good.

app/pages/admin/project-status.cjsx
alexbfree: i haven't designed this, it's a direct copy of what we have in the project builder. Which isn't to say that isn't a valid observation, just that this PR shouldn't be held up by that. Can we make a separate issue/suggestion for that?

app/pages/admin/project-status.cjsx
alexbfree: yes, it is the same as https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/project-details.cjsx. I am not actually sure how to do react mixins but will take a look if it can be done easily.

alexbfree:
thanks

brian-c:
Is this just duplicating the same feature on two different pages? Admins should be able to use the usual interface to make changes to any project, is that not the case?

alexbfree:
If that's true @brian-c That's news to me!

alexbfree:
@brian-c Just tried it. Nope.

screenshot 2016-05-17 14 41 18

alexbfree:
Are admins really supposed to be able to go into project builder for *any* project and make changes? Presumably just by going direct to the lab/<project-id> url since there wouldn't be a link anywhere?
That would be handy if true.

camallen:
linking - https://github.com/zooniverse/Panoptes-Front-End/issues/2150

brian-c:
I'd be in favor of fixing the lab page for admins over duplicating functionality. It probably checking if the user is an owner or collaborator, and it just needs to check if they're an admin.

Admins in admin-mode can just about anything, in case of emergencies.

alexbfree:
OK @brian-c, I agree with your reasoning. Looks like we have a bug then, so I will close this and we can use #2150 to track the bug

Issue #2532, opened 2016-05-05
geordi-logging Talk interface


wgranger:
- [x] talk-view event
- [x] enter key/ search button
- [x] back button
- [x] subject metadata
- [x] subject favorite
- [x] subject unfavorite
- [x] add to collection button
       - [x] add-to-collection
       - [x] make-collection-private
       - [x] make-collection-public
       - [x] add-collection
       - [x] search-collection
- [x] subject image button
- [ ] toolbar clicks
- [x] add comment button
- [x] notifications link
- [x] participant click
- [x] change projects button
- [x] hashtag clicks
- [x] subject name link click
- [x] subscribe to thread button
- [x] unsubscribe from thread button
- [x] edit post title button
- [x] insert linked image button
- [x] click on image to link it button
- [x] post links
       - [x] like-post
       - [x] reply-post
       - [x] link-post
       - [x] report-post
       - [x] edit-post
       - [x] delete-post
- [x] update comment button
- [x] pagination clicks
- [x] New discussion button
- [x] discussion title click
- [x] profile click under a post title
- [x] time click under a post
- [x] Recent Comments click
- [x] Breadcrumb click

wgranger:
Note: accessing the buttons for "toolbar clicks" is hidden in Markdown and not in PFE.

alexbfree:
Let's have me do a QA cycle before we merge the PR & close these tickets.

Issue #2531, opened 2016-05-05
geordi-logging classification interface


amyrebecca:
- [x] scroll down on project home
- [x] get started button
- [x] view subject
- [x] task interaction events
    - [x] draw
    - [x] drag
    - [x] resize
    - [x] delete
    - [x] select-option
    - [x] task-end
- [x] help with this task button
- [x] subtask answer click
- [x] subtask "ok" click
- [x] step completion button
- [x] exit to talk at end of task
- [x] cog menu
    - [x] classify-cog
    - [x] demo-mode-on
    - [x] demo-mode-off
    - [x] gold-standard-mode-on
    - [x] gold-standard-mode-off
    - [x] demo-mode-query
    - [x] gold-standard-mode-query
- [x] "classify" event when next pressed at end of task

alexbfree:
Let's have me do a QA cycle before we merge the PR & close these tickets.

srallen:
PR is merged.

Issue #2529, opened 2016-05-05
geordi-logging general interface


amyrebecca:
- [x] Top menu nav links
- [x] Footer menu nav links
- [x] Submenu clicks
    - [x] Project menu nav links
    - [x] About menu nav links
    - [x] Collect menu nav links
    - [x] Globe menu nav links
    - [x] User menu nav links
- [x] Clicks within collections
    - [x] view-collection
    - [x] view-favorite
    - [x] settings-collection
    - [x] settings-favorite (doesn't appear to exist)
    - [x] collab-favorite (doesn't appear to exist)
    - [x] collab-collection
- [x] send message screen
    - [x] username-search
    - [x] send-message
- [x] sign in / sign out
    - [x] login
    - [x] logout
- [x] registration form
    - [x] register
- [x] please register popup
    - [x] please-register
    - [x] register-no-thanks
    - [x] register-sign-in
    - [x] register-register

alexbfree:
+1 !! looks like you've been busy @amyrebecca !

amyrebecca:
several of these have been broken by subsequent merges :/

alexbfree:
Let's have me do a QA cycle before we merge the PR & close these tickets.

PR #2590, opened 2016-05-16
Geordi integration

amyrebecca:
Geordi is a small API for logging user activity. These commits add a geordi logging object to the global context for all components underneath PanoptesApp and add logging of a host of activities requested by our research team at Syracuse.

### DIFF
--- a/app/classifier/drawing-tools/column-rectangle.cjsx
+++ b/app/classifier/drawing-tools/column-rectangle.cjsx
@@ -10,8 +10,6 @@ MINIMUM_WIDTH = 25
 
 module.exports = React.createClass
   displayName: 'RectangleTool'
-  contextTypes:
-    geordi: React.PropTypes.object
 
   statics:
     initCoords: null
@@ -42,18 +40,11 @@ module.exports = React.createClass
 
   initCoords: null
 
-  dragEnd: (target) ->
-    @context.geordi?.logEvent type: 'drag-column'
-    deleteIfOutOfBounds target
-
-  logResize: ->
-    @context.geordi?.logEvent type: 'resize-column'
-
   render: ->
     {x, width} = @props.mark
 
     <DrawingToolRoot tool={this}>
-      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind null, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
         <rect ref="rect" x={x} y={0} width={width} height={@props.containerRect.height / @props.scale.vertical} />
       </Draggable>
 
@@ -61,8 +52,8 @@ module.exports = React.createClass
         <g>
           <DeleteButton tool={this} x={x + width + 20} y={15} />
 
-          <DragHandle x={x} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleLeftDrag} onEnd={@logResize}/>
-          <DragHandle x={x + width} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleRightDrag} onEnd={@logResize} />
+          <DragHandle x={x} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleLeftDrag} />
+          <DragHandle x={x + width} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleRightDrag} />
         </g>}
     </DrawingToolRoot>
 
--- a/app/classifier/drawing-tools/column-rectangle.cjsx
+++ b/app/classifier/drawing-tools/column-rectangle.cjsx
@@ -10,6 +10,8 @@ MINIMUM_WIDTH = 25
 
 module.exports = React.createClass
   displayName: 'RectangleTool'
+  contextTypes:
+    geordi: React.PropTypes.object
 
   statics:
     initCoords: null
@@ -40,11 +42,18 @@ module.exports = React.createClass
 
   initCoords: null
 
+  dragEnd: (target) ->
+    @context.geordi?.logEvent type: 'drag-column'
+    deleteIfOutOfBounds target
+
+  logResize: ->
+    @context.geordi?.logEvent type: 'resize-column'
+
   render: ->
     {x, width} = @props.mark
 
     <DrawingToolRoot tool={this}>
-      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
         <rect ref="rect" x={x} y={0} width={width} height={@props.containerRect.height / @props.scale.vertical} />
       </Draggable>
 
@@ -52,8 +61,8 @@ module.exports = React.createClass
         <g>
           <DeleteButton tool={this} x={x + width + 20} y={15} />
 
-          <DragHandle x={x} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleLeftDrag} />
-          <DragHandle x={x + width} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleRightDrag} />
+          <DragHandle x={x} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleLeftDrag} onEnd={@logResize}/>
+          <DragHandle x={x + width} y={(@props.containerRect.height / @props.scale.vertical) / 2} scale={@props.scale} onDrag={@handleRightDrag} onEnd={@logResize} />
         </g>}
     </DrawingToolRoot>
 
@@ -64,7 +73,7 @@ module.exports = React.createClass
   handleLeftDrag: (e, d) ->
     if @props.mark.width - d.x / @props.scale.horizontal >= MINIMUM_WIDTH
       @props.mark.x += d.x / @props.scale.horizontal
-      @props.mark.width -= d.x / @props.scale.horizontal 
+      @props.mark.width -= d.x / @props.scale.horizontal
       @props.onChange @props.mark
 
   handleRightDrag: (e, d) ->
--- a/app/classifier/drawing-tools/column-rectangle.cjsx
+++ b/app/classifier/drawing-tools/column-rectangle.cjsx
@@ -53,7 +53,7 @@ module.exports = React.createClass
     {x, width} = @props.mark
 
     <DrawingToolRoot tool={this}>
-      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind null, this} disabled={@props.disabled}>
         <rect ref="rect" x={x} y={0} width={width} height={@props.containerRect.height / @props.scale.vertical} />
       </Draggable>
 
--- a/app/classifier/drawing-tools/ellipse.cjsx
+++ b/app/classifier/drawing-tools/ellipse.cjsx
@@ -14,9 +14,6 @@ DELETE_BUTTON_ANGLE = 45
 module.exports = React.createClass
   displayName: 'EllipseTool'
 
-  contextTypes:
-    geordi: React.PropTypes.object
-
   statics:
     defaultValues: ({x, y}) ->
       x: x
@@ -56,13 +53,6 @@ module.exports = React.createClass
     x: @props.mark.rx * Math.cos theta
     y: -1 * @props.mark.ry * Math.sin theta
 
-  dragEnd: (target) ->
-    @context.geordi?.logEvent type: 'drag-ellipse'
-    deleteIfOutOfBounds target
-
-  logResize: ->
-    @context.geordi?.logEvent type: 'resize-ellipse'
-
   render: ->
     positionAndRotate = "
       translate(#{@props.mark.x}, #{@props.mark.y})
@@ -79,15 +69,15 @@ module.exports = React.createClass
           <line x1="0" y1="0" x2="0" y2={-1 * @props.mark.ry} strokeWidth={guideWidth} strokeDasharray={GUIDE_DASH} />
         </g>}
 
-      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
         <ellipse rx={@props.mark.rx} ry={@props.mark.ry} />
       </Draggable>
 
       {if @props.selected
         <g>
           <DeleteButton tool={this} x={deletePosition.x} y={deletePosition.y} rotate={@props.mark.angle} />
-          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'x'} x={@props.mark.rx} y={0} scale={@props.scale} onEnd={@logResize}/>
-          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'y'} x={0} y={-1 * @props.mark.ry} scale={@props.scale} onEnd={@logResize}/>
+          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'x'} x={@props.mark.rx} y={0} scale={@props.scale} />
+          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'y'} x={0} y={-1 * @props.mark.ry} scale={@props.scale} />
         </g>}
     </DrawingToolRoot>
 
--- a/app/classifier/drawing-tools/ellipse.cjsx
+++ b/app/classifier/drawing-tools/ellipse.cjsx
@@ -14,6 +14,9 @@ DELETE_BUTTON_ANGLE = 45
 module.exports = React.createClass
   displayName: 'EllipseTool'
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
   statics:
     defaultValues: ({x, y}) ->
       x: x
@@ -53,6 +56,13 @@ module.exports = React.createClass
     x: @props.mark.rx * Math.cos theta
     y: -1 * @props.mark.ry * Math.sin theta
 
+  dragEnd: (target) ->
+    @context.geordi?.logEvent type: 'drag-ellipse'
+    deleteIfOutOfBounds target
+
+  logResize: ->
+    @context.geordi?.logEvent type: 'resize-ellipse'
+
   render: ->
     positionAndRotate = "
       translate(#{@props.mark.x}, #{@props.mark.y})
@@ -69,15 +79,15 @@ module.exports = React.createClass
           <line x1="0" y1="0" x2="0" y2={-1 * @props.mark.ry} strokeWidth={guideWidth} strokeDasharray={GUIDE_DASH} />
         </g>}
 
-      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
         <ellipse rx={@props.mark.rx} ry={@props.mark.ry} />
       </Draggable>
 
       {if @props.selected
         <g>
           <DeleteButton tool={this} x={deletePosition.x} y={deletePosition.y} rotate={@props.mark.angle} />
-          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'x'} x={@props.mark.rx} y={0} scale={@props.scale} />
-          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'y'} x={0} y={-1 * @props.mark.ry} scale={@props.scale} />
+          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'x'} x={@props.mark.rx} y={0} scale={@props.scale} onEnd={@logResize}/>
+          <DragHandle onDrag={@handleRadiusHandleDrag.bind this, 'y'} x={0} y={-1 * @props.mark.ry} scale={@props.scale} onEnd={@logResize}/>
         </g>}
     </DrawingToolRoot>
 
--- a/app/classifier/drawing-tools/point.cjsx
+++ b/app/classifier/drawing-tools/point.cjsx
@@ -13,6 +13,9 @@ DELETE_BUTTON_ANGLE = 45
 
 module.exports = React.createClass
   displayName: 'PointTool'
+  
+  contextTypes:
+    geordi: React.PropTypes.object
 
   statics:
     defaultValues: ({x, y}) ->
@@ -40,6 +43,10 @@ module.exports = React.createClass
     x: (SELECTED_RADIUS / @props.scale.horizontal) * Math.cos theta
     y: -1 * (SELECTED_RADIUS / @props.scale.vertical) * Math.sin theta
 
+  dragEnd: (target) ->
+    @context.geordi?.logEvent type: 'drag-point'
+    deleteIfOutOfBounds target
+
   render: ->
     averageScale = (@props.scale.horizontal + @props.scale.vertical) / 2
 
@@ -58,7 +65,7 @@ module.exports = React.createClass
       <line x1="0" y1={crosshairSpace * selectedRadius} x2="0" y2={selectedRadius} strokeWidth={crosshairWidth} />
       <line x1={crosshairSpace * selectedRadius} y1="0" x2={selectedRadius} y2="0" strokeWidth={crosshairWidth} />
 
-      <Draggable onDrag={@handleDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
         <circle r={radius} />
       </Draggable>
 
--- a/app/classifier/drawing-tools/point.cjsx
+++ b/app/classifier/drawing-tools/point.cjsx
@@ -13,9 +13,6 @@ DELETE_BUTTON_ANGLE = 45
 
 module.exports = React.createClass
   displayName: 'PointTool'
-  
-  contextTypes:
-    geordi: React.PropTypes.object
 
   statics:
     defaultValues: ({x, y}) ->
@@ -43,10 +40,6 @@ module.exports = React.createClass
     x: (SELECTED_RADIUS / @props.scale.horizontal) * Math.cos theta
     y: -1 * (SELECTED_RADIUS / @props.scale.vertical) * Math.sin theta
 
-  dragEnd: (target) ->
-    @context.geordi?.logEvent type: 'drag-point'
-    deleteIfOutOfBounds target
-
   render: ->
     averageScale = (@props.scale.horizontal + @props.scale.vertical) / 2
 
@@ -65,7 +58,7 @@ module.exports = React.createClass
       <line x1="0" y1={crosshairSpace * selectedRadius} x2="0" y2={selectedRadius} strokeWidth={crosshairWidth} />
       <line x1={crosshairSpace * selectedRadius} y1="0" x2={selectedRadius} y2="0" strokeWidth={crosshairWidth} />
 
-      <Draggable onDrag={@handleDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
         <circle r={radius} />
       </Draggable>
 
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -101,7 +101,7 @@ Classifier = React.createClass
   prepareToClassify: (classification) ->
     classification.annotations ?= []
     if classification.annotations.length is 0
-      @addAnnotationForTask classification, @props.workflow.first_task, true
+      @addAnnotationForTask classification, @props.workflow.first_task
 
   render: ->
     <ChangeListener target={@props.classification}>{=>
@@ -343,15 +343,9 @@ Classifier = React.createClass
       </nav>
     </div>
 
-  logEvent: ->
-    @context?.geordi?.logEvent arguments...
-
   renderExpertOptions: ->
     return unless @props.expertClassifier
-
-    logWrapper = @logEvent
-
-    <TriggeredModalForm triggerProps={{onClick: -> logWrapper type: 'classify-cog' }} trigger={
+    <TriggeredModalForm trigger={
       <i className="fa fa-cog fa-fw"></i>
     }>
       {if 'owner' in @props.userRoles or 'expert' in @props.userRoles
@@ -360,7 +354,7 @@ Classifier = React.createClass
             <input type="checkbox" checked={@props.classification.gold_standard} onChange={@handleGoldStandardChange} />{' '}
             Gold standard mode
           </label>{' '}
-          <TriggeredModalForm triggerProps={{onClick: -> logWrapper type: 'gold-standard-mode-query' }} trigger={
+          <TriggeredModalForm trigger={
             <i className="fa fa-question-circle"></i>
           }>
             <p>A “gold standard” classification is one that is known to be completely accurate. We’ll compare other classifications against it during aggregation.</p>
@@ -373,7 +367,7 @@ Classifier = React.createClass
               <input type="checkbox" checked={@props.demoMode} onChange={@handleDemoModeChange} />{' '}
               Demo mode
             </label>{' '}
-            <TriggeredModalForm triggerProps={{onClick: -> logWrapper type: 'demo-mode-query' }} trigger={
+            <TriggeredModalForm trigger={
               <i className="fa fa-question-circle"></i>
             }>
               <p>In demo mode, classifications <strong>will not be saved</strong>. Use this for quick, inaccurate demos of the classification interface.</p>
@@ -384,7 +378,6 @@ Classifier = React.createClass
   # Whenever a subject image is loaded in the annotator, record its size at that time.
   handleSubjectImageLoad: (e, frameIndex) ->
     @context.geordi?.remember subjectID: @props.subject?.id
-    @context.geordi?.logEvent type: 'view'
 
     {naturalWidth, naturalHeight, clientWidth, clientHeight} = e.target
     changes = {}
@@ -396,10 +389,7 @@ Classifier = React.createClass
     classification.update 'annotations'
 
   # Next (or start):
-  addAnnotationForTask: (classification, taskKey, noLog) ->
-    noLog = (typeof noLog is 'boolean') && noLog
-    @context.geordi.logEvent type: 'task-end' unless noLog
-
+  addAnnotationForTask: (classification, taskKey) ->
     taskDescription = @props.workflow.tasks[taskKey]
     annotation = tasks[taskDescription.type].getDefaultAnnotation taskDescription, @props.workflow, tasks
     annotation.task = taskKey
@@ -412,7 +402,6 @@ Classifier = React.createClass
     @props.classification.update 'annotations'
 
   completeClassification: ->
-    @context.geordi?.logEvent type: 'step-end'
     @props.classification.update
       completed: true
       'metadata.session': getSessionID()
@@ -429,11 +418,9 @@ Classifier = React.createClass
     @props.onComplete?()
 
   handleGoldStandardChange: (e) ->
-    @logEvent type: if e.target.checked then 'gold-standard-mode-on' else 'gold-standard-mode-off'
     @props.classification.update gold_standard: e.target.checked || undefined # Delete the whole key.
 
   handleDemoModeChange: (e) ->
-    @logEvent type: if e.target.checked then 'demo-mode-on' else 'demo-mode-off'
     @props.onChangeDemoMode e.target.checked
 
   toggleExpertClassification: (value) ->
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -443,7 +443,7 @@ Classifier = React.createClass
     @setState backButtonWarning: true
 
   warningToggleOff: ->
-    @setState backButtonWarning: false 
+    @setState backButtonWarning: false
 
   renderBackButtonWarning: ->
     <p className="back-button-warning" >Going back will clear your work for the current task.</p>
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -24,6 +24,9 @@ PULSAR_HUNTERS_SLUG = 'zooniverse/pulsar-hunters'
 Classifier = React.createClass
   displayName: 'Classifier'
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
   getDefaultProps: ->
     user: null
     workflow: null
@@ -54,6 +57,13 @@ Classifier = React.createClass
       @loadSubject subject
     if nextProps.classification isnt @props.classification
       @prepareToClassify nextProps.classification
+
+    @context.geordi.remember subjectID: @props.subject?.id
+
+  componentWillUnmount: () ->
+    try
+      @context.geordi?.forget ['subjectID']
+
 
   loadSubject: (subject) ->
     @setState
@@ -91,7 +101,7 @@ Classifier = React.createClass
   prepareToClassify: (classification) ->
     classification.annotations ?= []
     if classification.annotations.length is 0
-      @addAnnotationForTask classification, @props.workflow.first_task
+      @addAnnotationForTask classification, @props.workflow.first_task, true
 
   render: ->
     <ChangeListener target={@props.classification}>{=>
@@ -136,7 +146,7 @@ Classifier = React.createClass
     # Should we disabled the "Back" button?
     onFirstAnnotation = classification.annotations.indexOf(annotation) is 0
 
-    
+
 
     # Should we disable the "Next" or "Done" buttons?
     if TaskComponent.isAnnotationComplete?
@@ -333,9 +343,15 @@ Classifier = React.createClass
       </nav>
     </div>
 
+  logEvent: ->
+    @context?.geordi?.logEvent arguments...
+
   renderExpertOptions: ->
     return unless @props.expertClassifier
-    <TriggeredModalForm trigger={
+
+    logWrapper = @logEvent
+
+    <TriggeredModalForm triggerProps={{onClick: -> logWrapper type: 'classify-cog' }} trigger={
       <i className="fa fa-cog fa-fw"></i>
     }>
       {if 'owner' in @props.userRoles or 'expert' in @props.userRoles
@@ -344,7 +360,7 @@ Classifier = React.createClass
             <input type="checkbox" checked={@props.classification.gold_standard} onChange={@handleGoldStandardChange} />{' '}
             Gold standard mode
           </label>{' '}
-          <TriggeredModalForm trigger={
+          <TriggeredModalForm triggerProps={{onClick: -> logWrapper type: 'gold-standard-mode-query' }} trigger={
             <i className="fa fa-question-circle"></i>
           }>
             <p>A “gold standard” classification is one that is known to be completely accurate. We’ll compare other classifications against it during aggregation.</p>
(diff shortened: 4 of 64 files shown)


amyrebecca:
@srallen, any chance you'd have time to look at this? i know you're very busy

srallen:
@amyrebecca Can it wait till Friday? If not, I can punt to someone else.

amyrebecca:
@srallen yeah i'm sure it can, there's no deadline we're trying to hit

alexbfree:
Wow well done @amyrebecca and @wgranger . I definitely would like to review this before merge, but I am away on a business trip today and not working Monday. If there's any chance you can wait until Tuesday I can review this then. 

Have you got this deployed on staging somewhere? logging to geordi staging?

alexbfree:
PS does this include everything in the spreadsheet, or is this just a first pass?

srallen:
Something odd has happened with this branch. I can only check it out as a remote branch in a detached head state like if this feature branch was on a fork. Don't think this really affects anything, but fwiw, thought to note it here.

srallen:
@alexbfree I'm guessing #2529, #2531, #2532 is the list of logging that was added.

IMO, in the future, could issues be [closed by the PR](https://github.com/blog/1506-closing-issues-via-pull-requests) rather than ahead of it so we can have a reference to the original issue(s) and then have it auto close when merged?

package.json
```diff
@@ -69,7 +71,7 @@
     "stage-branch": "export DEPLOY_SUBDIR=\"panoptes-front-end/$(git symbolic-ref --short HEAD)\"; npm run stage;  echo \"Deployed to http://$(git symbolic-ref --short HEAD).pfe-preview.zooniverse.org/\";",
     "stage-branch-with-docker": "./bin/run-through-docker.sh 'npm run-script stage-branch'",
     "stage-with-docker": "./bin/run-through-docker.sh 'npm run-script stage'",
-    "start": "check-engines && ./bin/serve.sh",
```
srallen: Was taking out `check-engines` intentional?

app/talk/init.cjsx
```diff
@@ -27,6 +27,13 @@ module?.exports = React.createClass
   propTypes:
     section: React.PropTypes.string # 'zooniverse' for main-talk, 'project_id' for projects
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
+  logTalkClicks: (talkItem) ->
```
alexbfree: Why is the method name plural?

app/talk/index.cjsx
```diff
@@ -8,21 +8,30 @@ TalkFootnote = require './footnote'
 module?.exports = React.createClass
   displayName: 'Talk'
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
+  logSearchBox: (searchItem) ->
```
alexbfree: the method name and event type don't seem consistent? "search box" or "talk view" ?

app/talk/inbox-form.cjsx
```diff
@@ -27,7 +30,13 @@ module?.exports = React.createClass
     @setState {validationErrors}
     !!validationErrors.length
 
+  logEvent: (evtType) ->
+    @context.geordi?.logEvent
+      type: evtType
+      data: null
```
alexbfree: is setting data to null necessary?

app/talk/discussion-preview.cjsx
```diff
@@ -15,21 +15,25 @@ module?.exports = React.createClass
   propTypes:
     discussion: React.PropTypes.object
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
   discussionLink: ->
+    logClick = @context.geordi?.makeHandler? 'view-discussion'
```
alexbfree: Why is this done with a different approach than some of the other files below? (no `makeHandler` in some of the others)

app/talk/comment-image-selector.cjsx
```diff
@@ -46,10 +49,11 @@ module?.exports = React.createClass
         @props.onSelectImage subject
 
   imageItem: (data, i) ->
+    logClick = @context.geordi?.makeHandler? 'select-image-to-link'
     <div key={data.id} className="talk-comment-image-item">
       <img src={getSubjectLocation(data).src} />
       <div className="image-card-select">
-        <button onClick={=> @setFocusImage(data)}>Select</button>
+        <button onClick={=> @setFocusImage(data); logClick?('')}>Select</button>
```
alexbfree: is the empty string as parameter correct?

app/talk/comment-box.cjsx
```diff
@@ -31,6 +31,7 @@ module?.exports = React.createClass
     content: ''
     subject: null
     user: null
+    logSubmit: false
```
alexbfree: this needs a comment i think to explain what it represents

alexbfree:
One comment . There seem to be a tonne of methods of the format
```
logNewDiscuss: ->
 +    @context.geordi?.logEvent
 +      type: 'new-discussion'
```
couldn't we centralize these to a common included file to avoid repetition and having so many almost-the-same-but-not-quite methods?

app/talk/search.cjsx
```diff
@@ -26,6 +26,14 @@ module.exports = React.createClass
   displayName: 'TalkSearch'
   mixins: [History]
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
+  logBackClick: (linkName) ->
+    @history.goBack()
```
alexbfree: Method should be renamed - it does more than just log, it actually executes the Back

app/talk/popular-tags.cjsx
```diff
@@ -23,6 +26,7 @@ module?.exports = React.createClass
     talkClient.type('tags/popular').get query
 
   tag: (talkTag, i) ->
+    logClick = @context.geordi?.makeHandler? 'hashtag-sidebar'
```
alexbfree: not sure this is being used?

alexbfree:
Also, I'm seeing at least three different approaches to Geordi logging:

```
@context.geordi?.logEvent type: 'drag-circle'
```

```
 logBackClick: (linkName) ->
    @context.geordi?.logEvent
      type: linkName

onClick={@logBackClick.bind this, 'search-back'}
```

```
logClick = @context.geordi?.makeHandler? 'hashtag-sidebar'
```

I think we should be more consistent.

app/partials/main-header.cjsx
```diff
@@ -70,19 +76,19 @@ module.exports = React.createClass
 
   links: ->
     <nav className="main-nav #{@mobileClass()}">
-      <Link to={"/projects"} className="main-nav-item" activeClassName="active"><Translate content="mainNav.projects" /></Link>
-      <Link to={"/about"} className="main-nav-item" activeClassName="active"><Translate content="mainNav.about" /></Link>
-      <Link to={"/talk"} className="main-nav-item" activeClassName="active"><Translate content="mainNav.talk" /></Link>
-      <Link to={"/notifications"} className="main-nav-item" activeClassName="active"><Translate content="mainNav.notifications" /></Link>
-      <Link to={"/collections"} className="main-nav-item" activeClassName="active"><Translate content="mainNav.collect" /></Link>
+      <Link to={"/projects"} className="main-nav-item" activeClassName="active" onClick={@logClick?.bind(this, 'mainNav.projects')}><Translate content="mainNav.projects" /></Link>
```
alexbfree: these event type keys have camel case and dots, where a lot of the others have all lower case and dashes. Up to now all Geordi event types have been lower case and dashes. I think we should keep them consistent unless we have a good reason not to.

alexbfree: Or, if it's because we need scoping, then we should change ALL of them to this format

app/partials/app.cjsx
```diff
@@ -3,15 +3,81 @@ auth = require 'panoptes-client/lib/auth'
 IOStatus = require './io-status'
 MainHeader = require './main-header'
 MainFooter = require './main-footer'
+GeordiClient = require 'zooniverse-geordi-client'
 {generateSessionID} = require '../lib/session'
 
+class GeordiLogger
```
alexbfree: I'd like to see more commenting on this class. what it represents and what its methods do.

app/partials/app.cjsx
```diff
@@ -3,15 +3,81 @@ auth = require 'panoptes-client/lib/auth'
...
+          type: eventType
+          data: "\"#{eventData}\""
+
+  logEvent: (logEntry) ->
+    geordi = @instance()
+    newEntry = Object.assign {}, logEntry, @keys
+    geordi?.logEvent newEntry
+    console.log 'No logger available for event ', JSON.stringify(logEntry) unless geordi?.logEvent
+
+  remember: (keyVals) ->
```
alexbfree: the scope and purpose and lifetime of this needs explaining

app/partials/app.cjsx
```diff
@@ -3,15 +3,81 @@ auth = require 'panoptes-client/lib/auth'
...
+
+  geordi: null
+
+  instance: =>
+    @geordi = @geordi || @makeGeordi @keys?.projectToken
+
+  makeGeordi: (projectSlug) ->
+    new GeordiClient
+      server: @state?.env
+      projectToken: projectSlug || ''
```
alexbfree: i wonder if we should also capture workflow ID if it's available

app/partials/app.cjsx
```diff
@@ -3,15 +3,81 @@ auth = require 'panoptes-client/lib/auth'
...
+    for key in forgetKeys
+      rebuild = true if key == 'projectToken'
+      delete @keys[key]
+
+    @geordi = @makeGeordi '' if rebuild
+
 module.exports = React.createClass
   displayName: 'PanoptesApp'
 
+  geordiLogger: null
```
alexbfree: explain this variable

app/pages/project/talk.cjsx
```diff
@@ -9,13 +9,21 @@ TalkFootnote = require '../../talk/footnote'
 module.exports = React.createClass
   displayName: 'ProjectTalkPage'
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
+  componentWillMount: ->
+    @context.geordi?.logEvent
```
alexbfree: why does this use a different approach than that below in the render method?

app/pages/project/metadata.cjsx
```diff
@@ -38,4 +57,5 @@ module?.exports = React.createClass
           <div>Retired Subjects</div>
         </div>
       </div>
+      <div ref="belowStats"></div>
```
alexbfree: what is this added for? is that really part of the Geordi PR?

app/pages/project/index.cjsx
```diff
@@ -47,19 +47,25 @@ ProjectPage = React.createClass
...
     pages: []
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
   componentDidMount: ->
     document.documentElement.classList.add 'on-project-page'
     @fetchInfo @props.project
     @updateSugarSubscription @props.project
+    @context.geordi?.remember projectToken: @props.project?.slug
```
alexbfree: Nice approach! I like it!

app/partials/app.cjsx
alexbfree: Hmm actually I see it is captured here https://github.com/zooniverse/Panoptes-Front-End/pull/2590/files#diff-83addd0f5535ee9ffe878b365f61712bR89 - where is it being stored in Geordi?  (sorry i think I am just missing this!)

app/components/user-search.cjsx
```diff
@@ -16,12 +16,14 @@ module.exports = React.createClass
...
   searchUsers: (value) ->
     clearTimeout @queryTimeout
+    onSearch = @props.onSearch
 
     if value is ''
       Promise.resolve options: []
     else
       new Promise (resolve) =>
         @queryTimeout = delayBy @props.debounce, =>
+          onSearch() if onSearch
```
alexbfree: This doesn't look like a Geordi related change?

app/collections/show-list.cjsx
```diff
@@ -21,10 +21,13 @@ module?.exports = React.createClass
...
   mixins: [History]
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
   componentDidMount: ->
     @fetchCollectionSubjects pick @props.location.query, VALID_COLLECTION_MEMBER_SUBJECTS_PARAMS
 
-  componentWillReceiveProps: (nextProps) ->
+  componentWillReceiveProps: (nextProps, nextContext) ->
```
alexbfree: where is that used?

app/classifier/tasks/generic.cjsx
```diff
@@ -50,6 +56,7 @@ module.exports = React.createClass
     </div>
 
   showHelp: ->
+    @logClick? null
```
alexbfree: is that right that we are passing null?

app/classifier/tasks/drawing/marking-initializer.cjsx
```diff
@@ -80,6 +83,9 @@ module.exports = React.createClass
     mark = @props.annotation.value[@props.annotation.value.length - 1]
     MarkComponent = drawingTools[taskDescription.tools[mark.tool].type]
 
+    toolName = taskDescription.tools[mark.tool].type
+    @context.geordi?.logEvent type: "draw-#{toolName}"
```
alexbfree: efficient, i like it

app/classifier/index.cjsx
```diff
@@ -333,9 +342,15 @@ Classifier = React.createClass
...
     </div>
 
+  logEvent: ->
+    @context?.geordi?.logEvent arguments...
+
   renderExpertOptions: ->
     return unless @props.expertClassifier
-    <TriggeredModalForm trigger={
+
+    logWrapper = @logEvent
```
alexbfree: just curious, why do we need the wrapper?

app/classifier/drawing-tools/ellipse.cjsx
```diff
@@ -53,6 +56,13 @@ module.exports = React.createClass
     x: @props.mark.rx * Math.cos theta
     y: -1 * @props.mark.ry * Math.sin theta
 
+  dragEnd: (target) ->
+    @context.geordi?.logEvent type: 'drag-ellipse'
+    deleteIfOutOfBounds target
+
+  logResize: ->
+    @context.geordi?.logEvent type: 'resize-ellipse'
+
```
alexbfree: I'm slightly concerned about the volume of data all these drag and resize clicks will generate!

app/classifier/drawing-tools/column-rectangle.cjsx
```diff
@@ -40,20 +42,27 @@ module.exports = React.createClass
...
+
+  logResize: ->
+    @context.geordi?.logEvent type: 'resize-column'
+
   render: ->
     {x, width} = @props.mark
 
     <DrawingToolRoot tool={this}>
-      <Draggable onDrag={@handleMainDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
+      <Draggable onDrag={@handleMainDrag} onEnd={@dragEnd.bind this, this} disabled={@props.disabled}>
```
alexbfree: Thinking of maintainability, is there anything we could do to make `@dragEnd.bind this, this` a bit clearer? Hard to understand why it's passing the same param twice...

alexbfree:
OK I have finished code review. Generally it looks good, with a few comments:

1. We should standardize our event type naming, not use a mix of scoped-camelcase and lower-case-with-dashes. This'll be a design change i guess not an implementation change. But maybe it's be good for everything to be scoped.

2. I think we should standardize the way in which we log to Geordi. So many different ways being used!

3. I would like to see a little bit more commenting, especially in the Geordi class.

4. Ensure no code for other features is present here. A couple of changes looked non-Geordi related.

Before we merge this, I think we should do a full QA cycle. But maybe you have already done some of that. I'm happy to be tester, I'll sync up with you on Slack and discuss.

srallen:
This probably should be reviewed by multiple people since this PR touches so much of the code, but @alexbfree if you want to take the lead, feel free to reassign this PR to yourself.

Looks like this PR also needs a rebase.

package.json
amyrebecca: no, it just annoys me and i forget to re-add it before committing. i'll make sure that's not in what gets merged

app/talk/init.cjsx
amyrebecca: it's a react feature, you'd have to ask them

amyrebecca:
1. i used the event names specified by the syracuse document. i really have better things to do than go back through and fix the capitalization but if there's no alternative i suppose we can push back the things i'm working on right now.
2. this is probably because a couple of us were working on it and also because about half the time we are logging additional data besides the event type and about half the time we are not. i don't see the value-add from re-writing all of this just because we use a few different method patterns
3. the geordi class is just a wrapper, it doesn't really do anything. if there's something in particular you think is unclear please add more detail
4. these changes are all geordi-only, i'm not sure what you thought you were seeing

amyrebecca:
i've finished the rebase and put check-engines back into package.json. this code is ready for merge pending any further requested changes.

srallen:
@alexbfree zooniverse/geordi-client#2 really probably should be merged and the client be patch version bumped before this gets merged.

srallen:
@alexbfree instead of manually testing with the Geordi server, could you look into adding integration tests to the geordi client? I opened zooniverse/geordi-client#3 for this and I would be a lot more comfortable with merging this if there were tests in place.

app/classifier/drawing-tools/column-rectangle.cjsx
srallen: React [autobinds for the mouse event](http://facebook.github.io/react/blog/2013/07/02/react-v0-4-autobind-by-default.html) so you can call the event handler and get what you'd expect back in a normal DOM event. However, if you want to do something with the React class instead, you have to bind it: `handler.bind(null, this)` and `this` will be the React class rather than the mouse event. Using `handler.bind(this)` will throw a warning in the console because the React library thinks you're trying to do the same thing as the autobinding. I'm not sure why this was changed to `@dragEnd.bind this, this`, though, @amyrebecca any clarification? FWIW, it seems like they accomplish the same thing.

app/classifier/drawing-tools/ellipse.cjsx
srallen: Can Geordi handle it? Maybe this event logging should be added later and in the interim do some performance testing?

app/classifier/index.cjsx
```diff
@@ -55,6 +58,13 @@ Classifier = React.createClass
     if nextProps.classification isnt @props.classification
       @prepareToClassify nextProps.classification
 
+    @context.geordi.remember subjectID: @props.subject?.id
+
+  componentWillUnmount: () ->
+    try
+      @context.geordi?.forget ['subjectID']
+
```
srallen: Should there be a `catch` here?

srallen:
Re: point 2, I also don't think abstracting out the method for logging is worth the effort. It may be slightly repetitious, but we have so many contributors to PFE now, I'd prefer some repetition that offers clarity of what's happening within a component over an abstraction. Perhaps later if it makes sense, it can be refactored.

app/collections/show-list.cjsx
srallen: When `contextTypes` are defined, they also become available in the lifecycle methods. While not explicitly used, it makes it clear it's now available in the lifecycle: https://facebook.github.io/react/docs/context.html#referencing-context-in-lifecycle-methods

app/components/subject-viewer.cjsx
```diff
@@ -201,6 +204,7 @@ module.exports = React.createClass
     @props.onFrameChange frame
 
   showMetadata: ->
+    @logSubjClick "metadata"
```
srallen: This is throwing `TypeError: this.logSubjClick is not a function` when I tried clicking on the metadata button in the subject viewer.

srallen: I don't see `logSubjClick` defined anywhere in the `SubjectViewer`.

app/pages/project/home.cjsx
```diff
@@ -33,6 +36,10 @@ module.exports = React.createClass
     undefined # Don't prevent default Link behavior.
 
   render: ->
+    logClick = @context.geordi?.makeHandler? 'get-started'
+    [owner, name] = @props.project.slug.split('/')
+
+  render: ->
```
srallen: render is being called twice.

app/pages/project/home.cjsx
```diff
@@ -33,6 +36,10 @@ module.exports = React.createClass
     undefined # Don't prevent default Link behavior.
 
   render: ->
+    logClick = @context.geordi?.makeHandler? 'get-started'
+    [owner, name] = @props.project.slug.split('/')
```
srallen: Don't think this is necessary. The component is using `@props.project.slug` where necessary and I don't see `owner` or `name` being used anywhere anyway.

app/pages/project/index.cjsx
```diff
@@ -211,6 +219,13 @@ ProjectPageController = React.createClass
       owner: null
       preferences: null
 
+  fetchProject: (owner, name) ->
+    @setState
+      loading: true
+      error: null
+      project: null
+      owner: null
+
```
srallen: The added lines 222-228 should be removed. It's breaking the project from loading. Possibly mistake in rebase?

app/collections/create-form.cjsx
```diff
@@ -55,14 +67,14 @@ module?.exports = React.createClass
         <input className="collection-name-input" ref="name" onChange={@handleNameInputChange} placeholder="Collection Name" />
         <div className="collection-create-form-actions">
           <label>
-            <input ref="private" type="checkbox" defaultChecked={false}/>
+            <input ref="private" type="checkbox" onClick={@logPrivacy.bind this} defaultChecked={false}/>
```
srallen: Binding here is throwing a React warning: `You are binding a component method to the component. React does this for you automatically in a high-performance way, so you can safely remove this call.` I don't think the binding here is necessary.

app/components/collection-search.cjsx
```diff
@@ -39,6 +43,7 @@ module.exports = React.createClass
       ref="collectionSelect"
       multi={@props.multi}
       name="collids"
+      onChange={@logSearch.bind this}
```
srallen: Don't think the binding here is necessary. Also it's throwing a React warning.

app/pages/project/metadata.cjsx
srallen: It's being used in the logging event on lines 19 and 20.

app/pages/project/metadata.cjsx
```diff
@@ -1,12 +1,31 @@
...
+
+  scrollDone: false
+
+  handleScroll: ->
+    return if @scrollDone
+
+    @scrollDone = visible.inViewport(@refs.belowStats)
+    @context.geordi?.logEvent type: 'scroll-down' if @scrollDone
+
+  componentDidMount: ->
```
eatyourgreens: You might want to throttle this method with a timeout, otherwise scrolling may become jerky on mobiles.

eatyourgreens: Oops, that should have referred to `handleScroll`. Fat fingers on a mobile phone!

app/partials/app.cjsx
```diff
@@ -3,15 +3,81 @@ auth = require 'panoptes-client/lib/auth'
...
+    new GeordiClient
+      server: @state?.env
+      projectToken: projectSlug || ''
+      zooUserIDGetter: () => @state.user?.id
+      subjectGetter: () => @keys?.subjectID
+
+  makeHandler: (defType) ->
+    instance = @instance
+    (eventData, eventType) ->
+        eventType = defType if typeof linkType isnt 'string'
```
eatyourgreens: `linkType` isn't defined in this method. Should this be `eventType`? What happens here if `defType` isn't a string either?

camallen:
Without reading the code (sorry) this will result in zooniverse.org event tracking == more gerodi server api requests (especially under media / promotion). I know @alexbfree did some basic load tests on the api server and it looked ok but stuff breaks...

so will this integration handle a dead geordi server gracefully? In my mind that is either just die quickly or flag geordi is dead for a timeout period (session storage) and then try api calls again after this expires. 

Thoughts?

app/partials/app.cjsx
eatyourgreens: @alexbfree why do we need to instantiate a new Geordi client each time that the project slug changes? Shouldn't the client be instantiated once, then updated with a new project ID when the slug changes in the URL?

Other API clients (Panoptes, Talk, Sugar, Stats) are in `panoptes-client`, for convenience. Should the Geordi Panoptes client code live there too?

eatyourgreens:
Staged now at https://geordi-integration.pfe-preview.zooniverse.org/ by the way.

camallen:
thanks @eatyourgreens looks like the current integration breaks the app :(

image

eatyourgreens:
@camallen Yeah, @srallen pointed out an outstanding PR on the Geordi client which will fix that https://github.com/zooniverse/Panoptes-Front-End/pull/2590#issuecomment-221317800

There's also a bug, also pointed out by Sarah, that prevents projects from loading when you click through from the projects page.

srallen:
@camallen Thanks, you more clearly ask about a point I brought up in a line comment. I'm also pretty concerned that the geordi-client or geordi itself do not have any tests.

camallen:
@alexbfree my comment [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2590#issuecomment-221536633) will be fixed by making geordi only run over https (which should be the default)

alexbfree:
@eatyourgreens has pointed out to me that this code as it is, will instantiate a different Geordi client for every project. This is the only way to do it with the Geordi client as is - but Geordi client was built for Ouroboros where there is only one project - the right thing for us to do here is to update the Geordi client such that there can be one client, but multiple loggers, one for each project. So you would do something like GeordiClient.createLogger("project-name") and use that logger instead of the Geordi client directly.

It was always my intent that we would make necessary updates to Geordi and the Geordi client as part of its integration with PFE, not code to what's there as an unchangeable service. Storing the panoptes project, workflow, etc as a distinct field is one of the other things we should do in this vein. However unfortunately as the integration has proceeded without my involvement I haven't had chance to help make these things happen.

I really think what needs to happen before we proceed with this integration is that we need to properly sync up and work on this together. there needs to be a strategy across PFE, Geordi client and Geordi, and some proper integration testing as @srallen points out. Using Geordi at the scale we are going to do here is new ground for Geordi, and while I have done extensive load testing of the Geordi server (see https://github.com/zooniverse/geordi/tree/master/loadtest ), we have not tested the client in this kind of set up, and we need to do it properly.

For now, I'm tagging this PR Do not merge, as I think significant work is needed before we merge.

app/talk/init.cjsx
alexbfree: I meant `logTalkClicks` not `contextTypes`

app/partials/app.cjsx
alexbfree: @eatyourgreens I agree. Geordi client was built for Ouroboros where there is only one project and it was always my intent we would update geordi client as part of this integration to meet the needs of PFE usage - this is a good example of where we need to do that, it shouldn't be bound to a project. Unfortunately I haven't been included in the development effort so this was missed. I will sync up with @wgranger and @amyrebecca and make sure we update geordi client, and this integration, accordingly.

alexbfree: PS @eatyourgreens That sounds sensible about `panoptes-client`. Where is that, I don't immediately see it? 
We shouldn't tie Geordi client to Panoptes, as it is used with Ouroboros too - and could be used elsewhere too - but in terms of its specific integration to PFE that should certainly be consistent with other clients used in PFE.

alexbfree:
@srallen re your comments on tests: 

The Geordi server does have load tests https://github.com/zooniverse/geordi/tree/master/loadtest but you're right that it doesn't have functional tests. That's mainly because it has almost no code - it's just a configuration of Loopback REST API server. The only real code is this one file: https://github.com/zooniverse/geordi/blob/master/geordi/common/models/event.js . Yes we could add some tests for that, perhaps more important since I added sequencing. However it's immediately obvious from a single post whether it's working or not - there aren't a lot of edge cases - that's why this hasn't been done up to now.

Geordi client, on the other hand, probably does need some tests adding. This will be the first time it has been used in production and I think we need to create the integration tests as you suggest as part of this PFE integration effort.

camallen:
@eatyourgreens is the issue with the client object instantiation a real blocker? Seems like we could merge and iterate on this to make it better with geordi lib releases and updated pfe integration.

app/partials/app.cjsx
eatyourgreens: @alexbfree `panoptes-client` is https://github.com/zooniverse/panoptes-javascript-client It bundles together the various clients used by PFE for convenience. Might be something to consider for the future.

eatyourgreens:
@camallen Agreed. Something to consider for a future PR, if we decide we want to keep all our web service clients in one place.

alexbfree:
I take your point @camallen , that particular thing is not a blocker. But I don't think this PR is ready to merge, for other reasons (such as broken UI, and inconsistent calls). That's why I have marked it Do not merge. Am hoping to catch up with Amy and Will next week to figure out how best to proceed.

wgranger:
I'm currently looking over these comments and making changes on the Talk logging, which was the code I touched. Once these changes are made I'll push up my commits and see where we stand with the rest of the logging. I also hope to meet with @alexbfree early next week to touch base.

app/classifier/index.cjsx
amyrebecca: i figured if it fails we don't really care, we just don't want to throw an error to the app

amyrebecca:
unfortunately i've been rebasing this thing daily for about two weeks, and along the way it's acquired some breakage because i can't test 100+ logging sites every time i do it. the longer we hold on to this, the more developer time we waste standing still.

app/partials/app.cjsx
amyrebecca: this is a rebase error, and you're correct

app/pages/project/metadata.cjsx
amyrebecca: good point.

app/pages/project/index.cjsx
amyrebecca: most likely

app/classifier/drawing-tools/ellipse.cjsx
alexbfree: Geordi can handle it in terms of accepting the data - we've done load testing for that. It's more of an ops issue in terms of how big do we want the database to get. My gut feeling is we shouldn't capture drag/resize events except for specific experiments/cases.

app/pages/project/metadata.cjsx
alexbfree: Per my comment about drag/resize above, I'm not sure we should be capturing scroll events across the board, only for specific experiments/situations. it will generate too much data.

alexbfree:
>the longer we hold on to this, the more developer time we waste standing still.
I don't think we are wasting time. I know you've put a lot of work in and are keen to see the back of it, but It's better to take the time to dot the I's, cross the T's and thoroughly do this properly rather than rush it in when it's not fully baked.

camallen:
@amyrebecca i completely agree re the rebase effort (it really sucks). You could squash the commits down (or rerere) to reduce the effort.

amyrebecca:
@alexbfree didn't mean to say making the changes is wasting time, just that i spend between a half hour and an hour a day rebasing so that this doesn't diverge too badly from the master (because so many files were touched). i'm going to try @camallen suggestion of squashing the commits in the hopes that this will speed things up.

app/partials/main-header.cjsx
wgranger: This info `mainNav.collect`, etc. is actually being passed in as "data." The event type for each of these remains `top-menu` as defined in the `logClick` function on line 41.

app/pages/project/talk.cjsx
wgranger: They look a bit different because one is on a click while the other executes with the page load. They could both probably be abstracted into the same function though.

app/components/user-search.cjsx
wgranger: This is related to Geordi because we are only trying to log geordi in certain instances of UserSearch. In this instance, the onSearch prop, which is a function, is passed in on line 59 of `inbox-form.cjsx`.

app/classifier/drawing-tools/ellipse.cjsx
wgranger: This was actually added according to the documentation we received as to what was to be logged in the classification interface. Looks like the drawing events we're currently targeting are draw, drag, resize, delete, select-option, and task-end.

eatyourgreens: These handlers are being called on drag end, rather than during the tool resize. I don't think that will generate a huge volume of data.

app/collections/manager.cjsx
```diff
@@ -17,7 +17,12 @@ module?.exports = React.createClass
     subject: React.PropTypes.object
     project: React.PropTypes.object
 
+  logModalChange: (clickedItem) ->
+    @props.contextRef?.geordi?.logEvent
```
eatyourgreens: Why is Geordi passed in via props here when it's passed by context everywhere else?

wgranger: Probably because this is in a modal. Modals don't have access to context. There are a couple instances of this in the Geordi logging.

app/pages/project/metadata.cjsx
wgranger: @eatyourgreens I assume using debounce would accomplish this? Instead of running `handleScroll` with every tiny scroll, we could control how many milliseconds pass until `handleScroll` is called.

app/classifier/drawing-tools/point.cjsx
```diff
@@ -40,6 +43,10 @@ module.exports = React.createClass
     x: (SELECTED_RADIUS / @props.scale.horizontal) * Math.cos theta
     y: -1 * (SELECTED_RADIUS / @props.scale.vertical) * Math.sin theta
 
+  dragEnd: (target) ->
+    @context.geordi?.logEvent type: 'drag-point'
+    deleteIfOutOfBounds target
+
```
eatyourgreens: This throws `Uncaught TypeError: Cannot read property 'getBoundingClientRect' of undefined`. Looks like the React component is being passed in as `target`, but `deleteIfOutOfBounds` expects a DOM element.

eatyourgreens: Ignore that, sorry. I think that error is unrelated to this PR.

app/collections/manager.cjsx
amyrebecca: will is correct. this was a real head scratcher for me for a day or two!

app/partials/main-header.cjsx
alexbfree: OK

app/classifier/drawing-tools/ellipse.cjsx
alexbfree: Following the discussion, we have agreed that what we need to do for now is limit it so that these scrollEnd, dragEnd etc events are only logged for the Gravity Spy project ID.  In future we should improve this with a project preference. I will create a separate issue for that.

app/partials/app.cjsx
alexbfree: Following the meeting we've agreed that this should be done, but as a separate ticket. I will raise a ticket.

app/pages/project/metadata.cjsx
alexbfree: ok

app/components/user-search.cjsx
alexbfree: OK

app/collections/show-list.cjsx
alexbfree: ok

alexbfree:
I have now merged Adam's pull request to add HTTPS support and published a new geordi-client version zooniverse-geordi-client@1.3.7. And updated this branch to use 1.3.7

app/collections/manager.cjsx
alexbfree: should we maybe use the `makeHandler` pattern here as well then? (fine if not)

eatyourgreens: Could do, but we'd still need to call `geordi.makeHandler` somehow and `geordi` is undefined if passed by context.

alexbfree:
OK @eatyourgreens @wgranger and I had a meeting to determine next steps. Notes [here](https://docs.google.com/document/d/1lH4wfa1bU1ajq-nyIm8X6E16VSXDRR2pQseqizj8a3I/edit) but the actions we agreed before this should be merged are:

**Alex**

- Update Geordi Server to store workflow ID as its own column. Then notify Will
- Create ticket to move PFE's Geordi integration code to panoptes-javascript-client
- Create ticket to create project preferences for Geordi.
- Review second requirements doc against the main spreadsheet and check no gaps. (if gaps, pass to Will to implement)
- Create ticket and discuss with ops about Geordi periodic archiving approach/CSV exports per project for researchers.
- Conduct final QA against requirements once everything else is done and prior to merge.

**Will**

- Add a project Slug for Zoo-wide talk (zooTalk) and Zoo home (zooHome) so that there is always a project slug in all PFE Geordi logger calls
- (once workflow column available in Geordi) Update Geordi client and PFE Geordi client code to pass workflow ID through to Geordi when logging (in the classifier). Check every time subject changes.
- Update Geordi client calls to not set project
- Add methods to Geordi client to `setProjectToken` and `setWorkflowId` and call these where appropriate (respectively especially where changing project, loading subject in classifier, etc) (optionally, using Jim's refactor idea)
- Add a limit so that ALL PFE Geordi logging only happens for specific Project IDs. It should be an array, but for now, only Gravity Spy should be included. I would have two arrays - one for extended logging (drawing events etc), the other for general Geordi logging.
- Add integration tests to Geordi Client per Sarah's suggestion.
- Address any remaining comments in PR.

We agreed the following things are fine as is:

- The three different types of calling Geordi (logEvent via a bound function, logEvent called direct, use of makeHandler)
- Consistent Event Types - turns out the eventTypes are all fine. I was looking at the data (however we did identify a separate issue there, see above).
- Storing project slug as project Token.

eatyourgreens:
Rebased this to latest master (with webpack.) Should be okay but there were a couple of small conflicts (in `/app/pages/projects/home.cjsx`, `/app/classifier/index.cjsx` and `package.json`.)

eatyourgreens:
We're pretty sure this is ready to go for Gravity Spy now.

eatyourgreens:
Staged at https://geordi-integration.pfe-preview.zooniverse.org/

app/classifier/drawing-tools/column-rectangle.cjsx
```diff
@@ -64,7 +64,7 @@ module.exports = React.createClass
   handleLeftDrag: (e, d) ->
     if @props.mark.width - d.x / @props.scale.horizontal >= MINIMUM_WIDTH
       @props.mark.x += d.x / @props.scale.horizontal
-      @props.mark.width -= d.x / @props.scale.horizontal 
+      @props.mark.width -= d.x / @props.scale.horizontal
```
alexbfree: no change?

app/partials/app.cjsx
```diff
@@ -3,15 +3,83 @@ auth = require 'panoptes-client/lib/auth'
...
+    @geordi = @geordi || @makeGeordi @keys?.projectToken
+
+  makeGeordi: (projectSlug) ->
+    new GeordiClient
+      env: @state?.env
+      projectToken: projectSlug || @keys?.projectToken
+      zooUserIDGetter: () => @state.user?.id
+      subjectGetter: () => @keys?.subjectID
+
+  makeHandler: (defType) -> # Once defined, efficiently logs different data to same event type
```
alexbfree: much nicer!

app/partials/main-footer.cjsx
```diff
@@ -75,6 +75,19 @@ AdminToggle = React.createClass
...
 
+  contextTypes:
+    geordi: React.PropTypes.object
+
+  logMenuClick: (linkName) ->
+    @context.geordi?.logEvent
+      type: 'footer-menu'
+      data: "\"#{linkName}\""
+
+  loggableLink: (aLink, linkSymbolicName) ->
```
alexbfree: nice!

app/partials/main-header.cjsx
alexbfree: I know i briefly thought otherwise... but i really do think that we should stick to the original Geordi design intent - if it's a string, we should store it in relatedID not data. We should keep data as JSON. We can do this this way first if it's too much to change: 
https://zooniverse.slack.com/archives/geordi-into-panoptes/p1465224533000008

app/talk/init.cjsx
alexbfree: We should rename this to `logTalkClick`

app/classifier/drawing-tools/column-rectangle.cjsx
wgranger: no change

app/talk/popular-tags.cjsx
alexbfree: @wgranger can you check this? delete if not being used, confirm used and we can ignore this comment if it is.

wgranger:
@alexbfree Concerning the `makeHandler` function, all `data` received was in the form of a string, so it's now being sent to `relatedID`. There was only one other instance where `data` was receiving a string.

app/partials/app.cjsx
alexbfree: @wgranger since it's a public exported variable, I think we should have a comment saying what this is/what it offers to a class that includes this file

app/partials/main-header.cjsx
alexbfree: @wgranger did you see this comment?

app/talk/popular-tags.cjsx
wgranger: This is being used. In an earlier commit, the hashtag call was missing `data` for a project-specific talk. The newest commit resolves this.

app/partials/main-header.cjsx
wgranger: Yes. This has been resolved with the most recent commit. Strings are now passed to relatedID through the `makeHandler` function. When `data` is passed in through another format, often an object, it uses the `logEvent` function.

app/partials/app.cjsx
wgranger: This comment would explain what geordi is?

app/talk/popular-tags.cjsx
alexbfree: ok. resolved.

app/partials/main-header.cjsx
alexbfree: ok sounds good. subject to testing, this comment thread is resolved.

app/partials/app.cjsx
alexbfree: @will Something simple like "An instance of the geordi client that can be used by the consumer of this class to log events to the Geordi server and google analytics"

Actually could probably go even simpler than that, it's a bit wordy. I just think since the class is effectively offering an interface to the rest of the codebase it warrants at least a short comment.

app/components/user-search.cjsx
alexbfree: resolved

app/collections/show-list.cjsx
alexbfree: resolved

app/components/subject-viewer.cjsx
alexbfree: @wgranger please check this is resolved.

app/classifier/drawing-tools/column-rectangle.cjsx
alexbfree: resolved

app/partials/app.cjsx
eatyourgreens: An interface that allows a Panoptes to impersonate an Ouroboros project to the Geordi client?

It's essentially a wrapper so that `zooUserIDGetter` and `subjectGetter` work in the client.

eatyourgreens: Or maybe "maintains project and subject context for the Geordi client"? Something along these lines anyway.

app/components/subject-viewer.cjsx
wgranger: This is resolved.

app/partials/app.cjsx
will: Is Geordi like a Star Trek thing?

alexbfree:
Tested, inspected. LGTM. Well done to Amy, Will and Jim for a lot of hard work that went into this.

I am ready to merge, just waiting for the OK from @eatyourgreens before I proceed.

alexbfree:
@eatyourgreens has approved, so now we merge!!

parrish:
Looks like this introduced some broken code [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2590/files#diff-39).  It might be a good idea to re-review the changes here to make sure we don't have any other regressions.

app/partials/main-footer.cjsx
```diff
@@ -89,44 +102,44 @@ module.exports = React.createClass
...
             <Translate component="h6" content="footer.about.title" />
-            <Link to="/about"><Translate content="footer.about.aboutUs" /></Link>
-            <Link to="/about/education"><Translate content="footer.about.education" /></Link>
-            <Link to="/about/team"><Translate content="footer.about.ourTeam" /></Link>
-            <Link to="/about/publications"><Translate content="footer.about.publications" /></Link>
-            <Link to="/about/contact"><Translate content="footer.boilerplate.contact" /></Link>
+            {@loggableLink <Link to="/about" />, 'footer.about.aboutUs'}
+            {@loggableLink <Link to="/about/education" />, 'footer.about.education'}
+            {@loggableLink <Link to="/about/team" />, 'footer.about.ourTeam'}
+            {@loggableLink <Link to="/about/education" />, 'footer.about.education'}
```
parrish: Replaced publications link with a duplicate education link

app/talk/latest-comment-link.cjsx
```diff
@@ -50,13 +57,15 @@ module?.exports = React.createClass
     @setState({latestCommentText}) if latestCommentText
 
   discussionLink: (childtext = '', query = {}, className = '') ->
+    if className = "latest-comment-time"
```
parrish: Accidental assignment (fixed in 8602426a)

parrish:
Okay, double checked and fixes are in #2712

Issue #2593, opened 2016-05-17
ETC on Stats page seems to  significantly underestimate the time to completion? Says 12 days I get 115 days


mschwamb:
@CKrawczyk I had to calculate time to completion for Planet Four: Terrains where we do have all the previous completed subject sets linked in the workflow so the stats should be correct. 

Using the database I find that P4T is in need of 80734 classifications to complete the current live dataset. Using the retired flags from the database csv dump 6007 subjects are still live

Of those subjects that are still live, P4T needs 80734 classifications to get all the subjects live to the 20 retirement limit. I'm just summing up 20- classification count for those subjects that do not have the retirement flag set. 

We're doing about 700 classifications per day, so that's 115 days to completion (80734/700.0) not 12 days as the stats page is suggesting

One thing to note is that for all projects launched in June with panoptes launch, there was a glitch in the retirement on the backend and subjects weren't being retired so there are some subjects that got 400 classifications, so if you're just doing number of classifications the project ever received divided by the total needed to finish all the current data it will always give very skewed results for Planet Four: Terrains and those other projects.

mschwamb:
Maybe this is also sorted by zooniverse/Panoptes#1804 ?

CKrawczyk:
The current calculation takes the number of classifications for the past 14 days (excluding the current day) and calculates how long it would take to retire the current workflow via:

`(14 days) * ((number of subjects * retirement limit) - number of classifications done) / (number of classifications in the past 14 days)`

But as you said the glitch early on means `number of classifications done` is skewed.  Unfortunately with the information currently available via the public api (I can't access the original database since that is not public) this is the best estimate I can think of.  As you siad https://github.com/zooniverse/Panoptes/issues/1804 will likely help with this.

camallen:
just thinking about this, when science teams link new / unlink old data that number of subjects can change but the total number of classifications will never decrease here. Not sure how fixing the retired counters will help with a situation like this.

mschwamb:
so this also means that if I change the retirement count on the project this would also be incorrect?  I've changed the retirement count several times depending on stargazing live and other criteria in the past on projects.  Really unfortunate that these stats are not useful. 

I'd love to advertise the stats page with a blog post I'm not sure what to say since for one of my projects it says we're out of data and the other it says we're going to run out in 14 days which isn't true. I guess I'll hold off for now.  Hopefully in the future there will be a better way of reporting the actual status of the project.

[...]

Issue #2591, opened 2016-05-16
Stats page completion plot is incorrect if you don't keep all your previous subject sets active


mschwamb:
This might be a backend issue.  Because I don't have completed subject sets still checked on Comet Hunters I get a completed count that is higher than the  number of available subjects which makes the plot on the stats page say that Comet Hunters is 100% complete and out of data when it's not. 

This might be more of a backend issue, but basically unless a project owner keeps everything checked then the completion plot isn't useful at all. Perhaps instead it can do the count for the live subject sets and do the query on retired versus live and show that in the plot

cc @CKrawczyk @camallen

aliburchard:
I think this is a failure to message properly on our part, but we know that if you unlink subject sets from workflows then the counters are wrong, so I've been telling everyone to keep all subject sets linked. What's the intention behind delinking subject sets? @camallen thinking we either need to more clearly message the relationship between subjects, subject sets, and workflows or else change the way we calculate stats...

camallen:
https://github.com/zooniverse/Panoptes/issues/1804

Agree, I think we should fix the counters

aliburchard:
yay! that makes more sense and is how people intuitively want to use the subject set linking.

mschwamb:
thanks

srallen:
This looks fixed.

mschwamb:
Re-opening - A temp band-aid solution was implemented. Comet Hunters still says 100% complete one of the workflows  because I thinks unchecked and beta test classifications.  All that was added was a note to see the ETC might be wrong. Currently it makes the stats page not helpful and I still need to compute the completion number myself. I think the backend might now have the feature but the stats page hasn't been updated.

CKrawczyk:
@camallen is there a value on the backend that would be more accurate for this than `workflow.classifications_count` (i.e. classification count for live subject sets in a workflow)?

camallen:
I think we need to change the workflow completeness metric to only use the current linked set state instead of the global counts, e.g. [here](https://github.com/zooniverse/Panoptes/blob/4f4fcd6398a1cf1a0d92483057b28827f53e6cca/app/workers/calculate_project_completeness_worker.rb#L32) should figure out how many of the current set is retired instead of taking the whole workflow lifetime count OR we change the count to be the active count and add a new `lifetime_retired_subjects_count` for the total count. That way the value is precomputed (might be a little difficult though).

mschwamb:
see #3145

PR #2594, opened 2016-05-17
message possibly incorrect ETA problem

camallen:
linked to #2591 and #2593

message to users that the ETA may be wrong - happy to change the text on this one.


CKrawczyk:
Looks reasonable to me, I was just thinking about adding something like this myself.

camallen:
let's get it out there then and we can change the text later.

PR #2595, opened 2016-05-17
Only fetch workflow tutorials

srallen:
Now that workflow tutorials are in production and Marten reran the rake task yesterday to link all of the pre-existing tutorials, we can now remove the fallback call to fetch tutorials by project id only. This prevents the possible scenario of a project builder creating two tutorials, but since they are unlinked to a workflow, we don't know which they intended to have pop up. The call to fetch by project id will just fetch the first one in the response array. I also bolded the instructions on the editor to go to the workflow editor for linking.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -44,12 +44,12 @@ Classifier = React.createClass
     @loadSubject @props.subject
     @prepareToClassify @props.classification
     {workflow, project, user} = @props
-    Tutorial.startIfNecessary {workflow, project, user}
+    Tutorial.startIfNecessary {workflow, user}
 
   componentWillReceiveProps: (nextProps) ->
     if nextProps.project isnt @props.project or nextProps.user isnt @props.user
       {workflow, project, user} = nextProps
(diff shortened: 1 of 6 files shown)


brian-c:
There's an unchanged line that might not make sense anymore:

```coffee
.then ([tutorial]) ->
  tutorial if tutorial?.kind is 'tutorial' or tutorial?.kind is null
```

Will return the first tutorial if it's the right kind. Should that actually be something like:

```coffee
.then (tutorials) ->
  nonMinicourseTutorials = tutorials.filter (tutorial) ->
    tutorial.kind in ['tutorial', null]
  nonMinicourseTutorials[0]
```

That way it'll still find the tutorial if the first one is a minicourse-kind.

A couple other tiny things:

- Updating the `find` signature to just `({workflow}) ->` so it's clear that `project` is never used

- Remove the `awaitTutorialForWorkflow =` assignment since we're returning the result directly

srallen:
I've updated `find` for both the tutorial and mini-course.

brian-c:
Grand, thanks!

PR #2601, opened 2016-05-17
Cleans up errors in Talk

parrish:
As mentioned in #2580.  Fixes duplicate key errors in popular tags and traps the oddly intermittent bug in `ActiveUsers#userIdsOnPage`


brian-c:
Works for me, but does this mean tag IDs are not unique for some reason?

Issue #2253, opened 2016-03-01
Ensuring Preservation of Project Context


alexbfree:
One of the tasks we identified for enabling the move of Snapshot Serengeti to Panoptes.

All views of text and subject content need to be, by default, filtered to Snapshot Serengeti content (or whatever project you are currently viewing the Talk for). 
At the moment this is true for hashtags and boards/threads, but not for collections, recents, and in some cases search results. 

We acknowledged the desire to make Zooniverse community more open and not completely hide other Talk content, therefore we propose an "on-by-default" filter for the current project which can be removed.

We decided that Collections should *not* be limited to only hold and only show subjects for the current project - however we agreed that for usability reasons, if technically doable, we would like to be able to filter lists of collections to only show collections that apply to the current project, or view all.

In general we agreed that for "minimum acceptable" filtering to maintain project context should be best effort - if we have technical difficulties in doing it at this stage we should just do as much as we can for now.

This will also be helped by fixing #2212.

This ticket strongly overlaps with #2252 and #2251 and they may end up getting developed in tandem.

alexbfree:
Notes to self:

There are actually two recents pages right now:

1. project-specific e.g. `https://www.zooniverse.org/projects/aliburchard/wildebeest-watch/talk/recents`
2. user-specific e.g. `https://www.zooniverse.org/users/alexbfree`

Both use the same API endpoint (e.g. `https://talk.zooniverse.org/comments?sort=-created_at&page_size=20&section=project-6`) so shouldn't be too hard to unify. 
Right now, the project-specific one has a checkbox that includes/excludes subject comments. This adds/removes the subject_default=false endpoint from the When included, images are shown inline to the posts.  

Talk API doesn't seem to have a means to retrieve only subject posts, it can only retrieve all posts, or all non-subject posts. In the short term might be able to work around this by just getting the combined response and parsing it, but probably should modify the API slightly so you can specify the opposite of `subject_default=false` ie. subjects only. This would also allow separate paging/load more of subject vs. of posts.

Navigationally, will need to ensure the user's profile maintains some sense of current project context.
Currently this is outside of project context (per its URL) so will probably need to build a small tracking object in local storage or similar to keep track of current project ID. This would allow filters to be included (and enabled by default) on the recents page, as well adding nav options to return to the project's nav menus...

eatyourgreens:
[...]

PR #2604, opened 2016-05-17
Add project context to subject page comments

### LINKED ISSUE #2366: Subject link made with ^S appears in the Talk comment on the discussion page but doesn't show when the comment is displayed on the subject page
See the top comment on this subject page

https://www.zooniverse.org/projects/mschwamb/comet-hunters/talk/subjects/1287322

screen shot 2016-04-01 at 11 09 56 am

And then see the corresponding comment in the Talk notes thread

screen shot 2016-04-01 at 11 09 53 am

https://www.zooniverse.org/projects/mschwamb/comet-hunters/talk/84/35934?comment=88863&page=1

It would be helpful if the subject links could work on the subject page view. Make it easier for research times to go directly go to the intended subject the volunteer is referring to, without having to go to the discussion directly to see if the volunteer did make it the hyperlink.

parrish:
This is something of a temporary fix.  If subjects are actually used from multiple project contexts, this becomes a bit funky unless we maintain the project path in the route.

This fixes #2366.  It triages some of #2253.

### DIFF
--- a/app/subjects/comment-list.cjsx
+++ b/app/subjects/comment-list.cjsx
@@ -40,7 +40,7 @@ module?.exports = React.createClass
         <h2>Comments:</h2>
         <div>
           {for comment in @state.comments
-            <Comment key={"comment-#{comment.id}"} data={comment} locked={true} linked={true} />}
+            <Comment {...@props} key={"comment-#{comment.id}"} data={comment} locked={true} linked={true} hideFocus={true} />}
         </div>
 
         {if +@state.meta.page_count > 1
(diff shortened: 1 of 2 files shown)


parrish:
Feel free to reassign if you want to.

brian-c:
Is there any situation where we wouldn't be able to derive the project from the route?

alexbfree:
linking to #2253 . The new design I am implementing for #2253 also relies on getting the project from the route in all cases so it should be fine.

PR #2613, opened 2016-05-19
Changed 'Volunteers' to 'Registered Volunteers' on stats page

mschwamb:
To make the stats page consistent with the landing page, I changed 'Volunteers'--> 'Registered Volunteers' since the number quoted in the stats page is the number of logged-in sessions. The value doesn't reflect the number of unique non-logged  sessions, so I relabeled it in this PR so it's clear to research teams and volunteers what the number reported means


CKrawczyk:
Looks good to me.

Issue #2587, opened 2016-05-16
Improve graph labelling on stats page


alexbfree:
@CKrawczyk and I discussed this idea, raising it as a ticket here as suggested.

When you hover over a bar, show its count either in hover text, or by highlighting and labelling the bar on hover..

eatyourgreens:
Any info accessed via hover should also be available via focus, so make the bars focusable and show the labels on focus too.

PR #2615, opened 2016-05-19
add hover and focus labels for bar graphs

CKrawczyk:
Closes https://github.com/zooniverse/Panoptes-Front-End/issues/2587

Adds on-focus and on-hover labels to the bar charts on the stats page.  The labels are displayed along the top to make sure no bars are drawn over the top of them.

@eatyourgreens if you have time could you take a look at this.

### DIFF
--- a/app/pages/project/stats/charts.cjsx
+++ b/app/pages/project/stats/charts.cjsx
@@ -124,7 +124,17 @@ Graph = React.createClass
         if data.index % numberBars
           data.element.attr({style: "display: none"})
     else if data.type == 'bar'
-      data.element.attr({style: "stroke-width: #{100 / length}%"})
+      data.element.attr(
+        style: "stroke-width: #{100 / length}%"
+        focusable: true
+        tabindex: 0
+      )
+      data.group.elem('text', {
+          x: data.x1
+          y: 15
+          class: "ct-label ct-tooltip"
+        }
+      ).text(data.series[data.index])
 
   onDrawSmall: (data) ->
     if data.type == 'bar'
--- a/css/project-stats-page.styl
+++ b/css/project-stats-page.styl
@@ -7,6 +7,9 @@ zoo-orange = #f78d27
       .ct-bar, .ct-slice-donut
         // stroke-width: 4.5%
         stroke: zoo-blue
+        &:hover + .ct-tooltip
+        &:focus + .ct-tooltip
+          fill: #000
     .ct-series-b
       .ct-bar, .ct-slice-donut
         stroke-width: 4.5%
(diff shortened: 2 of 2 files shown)


eatyourgreens:
Code looks good to me. Staged at https://stats-graph-labelling.pfe-preview.zooniverse.org/?env=production if @alexbfree or @mschwamb has any comments.

I did have a bit of trouble at first noticing the labels when they popped up.

app/pages/project/stats/charts.cjsx
```diff
@@ -124,7 +124,17 @@ Graph = React.createClass
         if data.index % numberBars
           data.element.attr({style: "display: none"})
     else if data.type == 'bar'
-      data.element.attr({style: "stroke-width: #{100 / length}%"})
+      data.element.attr(
+        style: "stroke-width: #{100 / length}%"
+        focusable: true
```
eatyourgreens: Out of interest, is this an SVG attribute?

CKrawczyk: I actually don't know, just found it during a google search, removing it does not seem to hurt anything in Chrome.  Also I just noticed focus does not work at all in Firefox.

CKrawczyk:
Yeah, I had to put them at the top of the graph since SVG does not support z-index, and the bar to the right of a label covers it up.

mschwamb:
same feedback as @eatyourgreens maybe you could bold them change the color so it's more noticeable?

app/pages/project/stats/charts.cjsx
eatyourgreens: Sounds like Firefox hasn't added support for `tabindex` on SVG elements. I'll have a look to see if I can find anything about SVG support in the various browsers.

CKrawczyk: Looks like `focusable` is used by IE and Edge instead of `tabindex` according to http://allyjs.io/data-tables/focusable.html.

eatyourgreens: No harm in having both. If a parser encounters an unknown attribute, it should just ignore it and move on (it may add the attribute as a property to the DOM node.)

CKrawczyk:
How does bold and blue look?
screen shot 2016-05-19 at 1 05 13 pm

eatyourgreens:
Looks good to me. I'm happy to merge/deploy this and iterate on the styling later if that's required.

Also, it looks like Firefox never implemented support for focus on SVG elements, so I wouldn't worry about the keyboard not working in Firefox.

mschwamb:
That looks great to me.

css/project-stats-page.styl
```diff
@@ -19,8 +18,7 @@ zoo-orange = #f78d27
     fill: #000
   .ct-tooltip
     text-anchor: middle
-    color: #000
-    fill: rgba(0,0,0,0)
+    fill: alpha(zoo-blue, 0)
```
eatyourgreens: Missing a `font-weight: bold` here?

CKrawczyk: Yep, added it to my browser to test and never put in the css, fixed now.

eatyourgreens:
:+1:

PR #2617, opened 2016-05-19
dropdown bug fixes

### LINKED ISSUE #2544: dropdown options dialog - select conditional list not readable - Firefox
Firefox issue at minimum, select conditional list ok on Chrome and Safari

mcbouslog:
fixes the following issues:

1. in editor dialog - options for upstream conditional answers not showing labels
2. showing task title and dropdown label when the same
3. improperly keeping conditional dropdown disabled when custom answer provided and allow create enabled
4. adding placeholder text for when dropdown is purposefully disabled

closes #2544

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -118,10 +118,13 @@ module?.exports = React.createClass
     select.options[optionsKey]
 
   getDisabledAttribute: (i) ->
-    select = @props.task.selects[i]
+    {selects} = @props.task
+    select = selects[i]
+    condition = selects.filter (filterSelect) => filterSelect.id is select.condition
+    conditionIndex = selects.indexOf(condition[0])
     optionsKeys = @state.optionsKeys
 
-    if select.condition? and not optionsKeys[select.condition]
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -143,13 +146,16 @@ module?.exports = React.createClass
       <div>
         {selectKeys.map (i) =>
           options = @getOptions(i)
+          disabled = @getDisabledAttribute(i)
           <div key={Math.random()}>
-            <div>{selects[i].title}</div>
+            {if selects[i].title isnt @props.task.instruction
+              <div>{selects[i].title}</div>}
```
brian-c: Is this `if` checking for something specific? Seems kinda weird to just leave out the select title.

mcbouslog:
It's for a situation where you'd have multiple individual dropdown tasks within a combo task, like Month, Day and Year, where each has a task.instruction that's the same of the select.title. [Here's an example of what we're looking to avoid](https://pfe-preview.zooniverse.org/projects/markb-panoptes/nfn-task-test/classify?reload=0&workflow=2379).

brian-c:
Ah, I gotcha. Looks good then, thanks!

PR #2619, opened 2016-05-19
Use the talk-api projects resource for talk project links

### LINKED ISSUE #2138: Custom projects don't get listed in the Talks navigation bar
Projects like Shakespeare's World and  Wildcam Gorongosa  don't show up in the Talk navigation side bar
screen shot 2016-01-24 at 8 58 22 pm

parrish:
Simplifies the project listing in talk by using the talk resource instead of Panoptes.

Panoptes doesn't differentiate between projects with off-site urls and projects without .org talk (e.g. Ouroboros projects).

The Talk endpoint only lists projects that are launched and have Talk setup.

This closes #2138

### DIFF
--- a/app/talk/lib/project-linker.cjsx
+++ b/app/talk/lib/project-linker.cjsx
@@ -1,91 +1,52 @@
 React = require 'react'
-apiClient = require 'panoptes-client/lib/api-client'
+talkClient = require 'panoptes-client/lib/talk-client'
 {Link} = require 'react-router'
 Loading = require '../../components/loading-indicator'
-FEATURED_PRODUCT_IDS = require '../../lib/featured-projects'
-
-take = (n, arr) -> arr.slice(0, n)
-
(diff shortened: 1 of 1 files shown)


camallen:
>Panoptes doesn't differentiate between projects with off-site urls and projects without .org talk (e.g. Ouroboros projects).

We serialize the migrated field [here](https://github.com/zooniverse/Panoptes/blob/8de236503c5c44f20a9d47911f31a167faef061a/app/serializers/project_serializer.rb#L11) which are legacy / ouroboros projects and can if needed add a filter on that.

parrish:
Ah.  Hadn't seen that.  I think this route is better though since we can limit the results to launched projects with talk.

camallen:
Sounds like a plan

parrish:
API is good to go on this now.

Feel free to reassign.

PR #2620, opened 2016-05-19
Fix comment editing

### LINKED ISSUE #2618: Unable to edit posts using # or @ symbol since the keywords/username suggestions appeared
When I try to edit my post, it doesn't let me write # or @ symbol. The symbol disappears right after being typed and my cursor jumps at the end of the comment automatically. 

This means I am not able to add a hashtag after the correct species is confirmed by a scientist, etc. Very unhelpful. 

It is not a problem on my side. 

When it happens (everytime I type one of those symbols) the browser console shows this error:
>main.8fad6311fda1e98f6a9dc3b1f3ae728f.js:28 Uncaught SyntaxError: Failed to execute 'querySelector' on 'Document': '#mimic-textarea-0-2-0-2-3-0-3-5-0-0-$108232-1-3-0-3-0-0-1-0' is not a valid selector.

More: https://www.zooniverse.org/projects/birgus2/western-shield-camera-watch/talk/276/54454?comment=109135&page=1

parrish:
`document.querySelector` was breaking in cases where textarea-mimic-div ids contained non alphanumeric characters.  One of these cases was triggered by editing a comment.

This closes #2618

### DIFF
--- a/app/talk/lib/get-caret-position.cjsx
+++ b/app/talk/lib/get-caret-position.cjsx
@@ -38,9 +38,8 @@ getStyleOf = (element) ->
     element.currentStyle
 
 # Creates a div to mimic the textarea
-mimic = (element) ->
-  reactId = element.getAttribute 'data-reactid'
-  id = "mimic-textarea#{reactId}".replace(/\./g, '-')
+mimic = (element, id) ->
+  id = "mimic-textarea#{id}"
   div = document.querySelector "##{id}"
 
   unless div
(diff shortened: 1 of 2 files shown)


brian-c:
Ooo, now that I read this again I remember that the next version of React does away with `data-reactid` entirely. Probably best not to rely on another library's implementation details anyway. Can we access the element another way?

parrish:
Barf.  Yeah, I was just using it as a readymade unique id.  I'll just pass an identifier down from Suggester.

parrish:
Just pushed up a fix that just uses pseudo-unique date-based ids instead.

brian-c:
Nice, looks good, thanks.

Issue #2611, opened 2016-05-18
Warn about Deleting Annotations on Back button


simensta:
It can be unclear to the user that once the back button is hit, the annotations for the current task will be lost. This is especially surprising in back-to-back drawing tasks, when I user goes back to adjust a mark from a previous task, only to loose their current annotations.

Possible Solutions:
- Allow project builders to completely disable/remove the Back button from the Classify page.
- Warn users that pressing the back button will delete their current annotation(s). Maybe something like [this](http://codepen.io/simensta/pres/0528889d386044c10b14bd25b4c3b820)?

mschwamb:
@simensta I'd suggest changing the wording., especially if this wording/text shows on non-drawing projects as well.  I don't think volunteers know what is meant by annotations are. I think that's jargon the Zooniverse and research teams use, but isn't really known much outside of that. 

I'd suggest 'Going back will delete your work for the current task' or something along those lines

eatyourgreens:
How about something like "Going back will clear your answers for this question."? Just in case 'task' isn't clear to the volunteers either.

mschwamb:
I really like your text @eatyourgreens except for the word 'question'. Not all tasks in the workflow are questions sometimes they are directives like 'mark all the galaxy's bar'  I think 'task' covers  all scenarios, but likely other word choices too.

I'll have no complaints with whatever you and @simensta iterate and decide upon. Thanks for considering my opinion.

eatyourgreens:
I'm happy with either version too, so whichever @simensta thinks is best. If volunteers do have trouble understanding the text, it's an easy thing to change.

simensta:
@mschwamb, good point. I am thinking a combination of the suggestions above: "Going back will clear your work for the current task." I will update the PR momentarily.

mschwamb:
@simensta I think that works well and is very clear.

PR #2621, opened 2016-05-19
warning onFocus or hover for task back button

simensta:
Fixes Issue #2611. This PR warns users that pressing the back button deletes the annotations for the current task. The warning will only show when a user hovers over or focuses the back button. See it in action [here](https://back-button-warning.pfe-preview.zooniverse.org/).

screen shot 2016-05-19 at 2 47 56 pm

@mrniaboc, @aliburchard: If merged, this PR will introduce a slight change to the classify interface. What messaging would need to happen for the project builder? Also, opinions about copy?

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -39,6 +39,7 @@ Classifier = React.createClass
     classificationQuality: NaN
     showingExpertClassification: false
     selectedExpertAnnotation: -1
+    backButtonWarning: false
 
   componentDidMount: ->
     @loadSubject @props.subject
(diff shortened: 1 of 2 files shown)


app/classifier/index.cjsx
```diff
@@ -414,6 +416,15 @@ Classifier = React.createClass
...
     @setState showingExpertClassification: value
 
+  warningToggleOn: ->
+    @setState backButtonWarning: true
+
+  warningToggleOff: ->
+    @setState backButtonWarning: false 
+
+  renderBackButtonWarning: ->
+    <p className={"back-button-warning"} >{"Going back will delete all annotations for the current task."}</p>
```
eatyourgreens: I'm not sure that you need the curly brackets around the class name and paragraph text here.

eatyourgreens:
Minor comment about inserting strings as variables. Otherwise looks good to me.

simensta:
Thanks @eatyourgreens -- unnecessary curly brackets removed! @mrniaboc, do we need to notify project builders about this addition before merging?

eatyourgreens:
I'm going to assume @mrniaboc has no problem with this, and merge.

PR #2622, opened 2016-05-19
Default workflow feature

srallen:
Part of completing #2534 and building on top of #2572, this allows project builders to set a default workflow on the workflow edit page. I also fixed a bug I found where the state of the checkbox of the new default workflow input as well as the hide classification summaries input did not clear when switching between different workflow edit pages.

@brian-c feel free to punt to someone else if you're busy.

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -54,11 +54,6 @@ EditWorkflowPage = React.createClass
     deletionError: null
     workflowCreationInProgress: false
 
-  componentWillReceiveProps: (nextProps) ->
-    if nextProps.workflow isnt @props.workflow
-      ReactDOM.findDOMNode(@refs.defaultWorkflow).defaultChecked = nextProps.project.configuration.default_workflow is nextProps.workflow.id
-      ReactDOM.findDOMNode(@refs.hideClassificationSummaries).defaultChecked = nextProps.workflow.configuration.hide_classification_summaries
-
   workflowLink: ->
     [owner, name] = @props.project.slug.split('/')
     viewQuery = workflow: @props.workflow.id, reload: @state.forceReloader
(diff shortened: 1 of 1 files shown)


app/pages/lab/workflow.cjsx
```diff
@@ -53,6 +54,11 @@ EditWorkflowPage = React.createClass
     deletionError: null
     workflowCreationInProgress: false
 
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.workflow isnt @props.workflow
+      ReactDOM.findDOMNode(@refs.defaultWorkflow).defaultChecked = nextProps.project.configuration.default_workflow is nextProps.workflow.id
```
brian-c: I don't think `componentWillReceiveProps` is necessary here, since `checked` can all be set in `render` for both of these inputs.

srallen: I added this in to fix the bug I described. When switching between workflow editors, if the first workflow had these options enabled, when switching to the another workflow that do not have them enabled, the inputs would stay checked giving the false impression that these options were enabled for the second workflow.

brian-c: OH I get it now! Sorry! I think there's a neater way to handle that, let me play with it a minute...

brian-c: Okay, yeah, I'm seeing that now. I think fixing that is just a matter of using `checked` instead of `defaultChecked`. The `render` method is looking at the correct resources, but using `defaultChecked` makes the input [uncontrolled](https://facebook.github.io/react/docs/forms.html), which means it's not bound to anything and won't change to match new data.

srallen: doh! Gotta always check those docs. I'll have a change up shortly.

srallen:
Just updated to use `checked`.

brian-c:
Looks great, thanks.

PR #2631, opened 2016-05-24
Mini course ui

### LINKED ISSUE #2597: Style mini-course
Add basic styling for the mini-course per @jwbmartin design.

srallen:
Closes #2597. This adds styling for the mini-course as well as a refactored markup and a refactoring to use the preferences passed down as props now.

Can be previewed here: https://mini-course-ui.pfe-preview.zooniverse.org/projects/srallen086/mini-course-one-column/classify

There are 16 slides in the mini-course that is to be setup for Gravity Spy and is setup in the test project in the staged feature branch too. I still have the classifier hard coded to have it increment every two classifications (this will be changed in a later PR, but left as 2 for easy testing).

I think you had some comments about mini-courses with the previous workflow tutorials PR, so I'm assigning to you.


app/classifier/index.cjsx
```diff
@@ -220,7 +219,7 @@ Classifier = React.createClass
       <p>
         <small>
           <strong>
-            <MiniCourseButton className="minor-button" user={@props.user} project={@props.project} workflow={@props.workflow} title="Project Mini-course" aria-label="Show the project mini-course" style={marginTop: '2em'}>
+            <MiniCourseButton className="minor-button" user={@props.user} preferences={@props.preferences} project={@props.project} workflow={@props.workflow} title="Project Mini-course" aria-label="Show the project mini-course" style={marginTop: '2em'}>
```
brian-c: `title` and `aria-label` are probably not necessary on text buttons.

app/lib/mini-course.cjsx
```diff
@@ -17,81 +16,73 @@ module.exports = React.createClass
 
   statics:
     find: ({workflow}) ->
-      # Prefer fetching the tutorial for the workflow, if a workflow is given.
+      # Prefer fetching the tutorial for the workflow, so we know which one to fetch if multiple exist.
```
brian-c: Comment out of sync ("tutorial").

app/lib/mini-course.cjsx
```diff
@@ -17,81 +16,73 @@ module.exports = React.createClass
...
       if workflow?
         apiClient.type('tutorials').get workflow_id: workflow.id, kind: "mini-course"
           .then ([minicourse]) ->
             minicourse
```
brian-c: Why are multiple minicourses allowed per workflow if only the first is ever selected?

srallen: Technically mini-courses are just a tutorial resource with `kind: 'mini-course'` set and we allow multiple tutorials (in the back end) per workflow. However, in the project builder interface, we are only allowing one tutorial and/or one mini-course per workflow to be linked. With this limitation in the builder UI, I can be assured it's the first tutorial that is mini-course kind that is returned.

app/classifier/index.cjsx
srallen: They're being used on the "Show the project tutorial" button too, so I assume they were useful for something. Should they be removed from both?

css/mini-course-dialog.styl
```diff
@@ -1,3 +1,71 @@
...
+        a
+          color: MINI_COURSE_BLUE
+        
+  .media-card-header
+    background-color: #111111
+    
+  .media-card-media
+    display: block
+    margin: 0 auto
+    max-height: 33vh
```
brian-c: How about a `max-width: 100%` so a wide image doesn't cause the dialog to scroll?

app/classifier/index.cjsx
brian-c: Ah, yeah, that used to be just an icon button. Not a big deal, we can clean it up later.

app/lib/mini-course.cjsx
brian-c: Ahhh, I get it. Can we make the mini-course selection control out of radio buttons then?

Looks like it's the same for tutorials.

srallen: I added a conditional to display the input as a checkbox if there's only one and a radio if there are many to choose from in the workflow editor. I originally tried using radio buttons for all, but was getting these errors in the console when clicking the input: 
`Failed to load resource: the server responded with a status of 422 (Unprocessable Entity)`
`Uncaught (in promise) Error: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_workflow_tutorials_on_workflow_id_and_tutorial_id"`

srallen:
I think I addressed all of your comments. Let me know if there's anything else.

brian-c:
Very nice, thank you!

PR #2633, opened 2016-05-25
Do not create a default workflow task

### LINKED ISSUE #934: Tasks default to "questions" that need to be deleted for "drawing" workflows
When I create a new workflow in the project builder, the default first task is a "question". However, if I want my first task to be "drawing", the initial untitled and unpopulated "question" task remains and I have to manually delete it. This is rather annoying. Please auto-delete the first "question" task if it hasn't been modified from the default when the user chooses to start with a "drawing" task.

brian-c:
New workflows start with an empty task list. Closes #934.


eatyourgreens:
Looks good to me 👍

PR #2638, opened 2016-05-26
Bump modal-form

brian-c:
Bump to modal-form 2.4.0, which takes overflow into account when determining target dimensions.


simensta:
LGTM -- thanks @brian-c!

PR #2639, opened 2016-05-27
update focus and when image pan-able

mcbouslog:
addresses focus and when pan-able, keeping the following use cases in mind:

1. keyboard only - tab to pan & zoom controls to pan or zoom (or attempt to >:( on Safari) then tab away. once tabbed away from pan & zoom controls pan-able should be disabled.
2. mouse with drag and wheel - click cross-arrows so can pan & zoom, drag label to center, mousewheel to zoom in, maybe drag a bit more to re-center. once clicked away from pan & zoom controls pan-able should be disabled.
3. mouse with drag and click +/- buttons - click cross-arrows so can pan & zoom, drag label to center, click + button to zoom in (either hold down or click multiple times), maybe drag a bit more to re-center. once clicked away from pan & zoom controls pan-able should be disabled.

[branch staged here](https://pan-zoom-focus.pfe-preview.zooniverse.org/)

@simensta @eatyourgreens 

oh the draw button has two `title`'s, i just scooted together, i doubt hurting anything but just wanted to see if intentional or should get rid of one?

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -75,30 +75,38 @@ module.exports = React.createClass
           <div className="pan-zoom-controls" >
             <div className="draw-pan-toggle" >
               <div className={if @state.panEnabled then "" else "active"} >
-                <button title={"draw"} className={"fa fa-mouse-pointer"} title={"annotate"} onClick={@togglePanOff}/>
+                <button title={"draw"} title={"annotate"} className={"fa fa-mouse-pointer"} onClick={@togglePanOff}/>
               </div>
               <div className={if @state.panEnabled then "active" else ""}>
-                <button title={"pan"} className={"fa fa-arrows"} title={"pan"} onClick={@togglePanOn}/>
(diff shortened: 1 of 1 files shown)


eatyourgreens:
:+1: I do like the way you override the cursor keys if the zoom buttons are selected, but not if drawing/annotating.

By the way, the staging link should be https://pan-zoom-focus.pfe-preview.zooniverse.org

eatyourgreens:
I'm pretty sure that the second title will override the first on the drawing button.

By the way, I think you can replace `attribute = {"a constant string"}` with `attribute = "a constant string"` in all of your tags.

mcbouslog:
Oops, fixed the staging link.

Sounds good regarding attributes and strings, I'll edit accordingly before merge.

mcbouslog:
@eatyourgreens - cleaned up the attribute strings, let me know if there's anything else I can do. thanks,

eatyourgreens:
👍

PR #2641, opened 2016-05-31
Reorder tutorial slides

### LINKED ISSUE #2629: Add ability to reshuffle the tutorial cards
My collaborator just misunderstood how to replace an image in the tutorial and in the project we're building . He hit the x thinking that would delete the image (re-reading the help text I can see where he might have gotten confused). Since it was the 3rd step of out 7, I now need to add a new step and move all the tutorial text down by one by copying and pasting and then reupload the images. It would be nice for situations like this or when you're in development that you could move the different cards in the tutorial into a different order.

srallen:
Closes #2629. This adds `drag-reorderable` to the project modal editor that handles the tutorial and mini-course editing interface. This also makes the UI a bit more consistent with the field guide editor. 

Staged at https://reorder-tutorial-slides.pfe-preview.zooniverse.org/

### DIFF
--- a/app/partials/project-modal-editor.cjsx
+++ b/app/partials/project-modal-editor.cjsx
@@ -4,6 +4,9 @@ putFile = require '../lib/put-file'
 FileButton = require '../components/file-button'
 {MarkdownEditor} = (require 'markdownz').default
 debounce = require 'debounce'
+DragReorderable = require 'drag-reorderable'
+CroppedImage = require '../components/cropped-image'
+classnames = require 'classnames'
 
 ProjectModalStepEditor = React.createClass
   getDefaultProps: ->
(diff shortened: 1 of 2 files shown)


brian-c:
Great stuff. Only thing is I was super confused that nothing was reordering as I dragged, but I just figured out that the label on the button is the item index, which changes while the drag is in progress.

Maybe on render we could set each item's `_key` to its index, and then on drop clear out all the `_key`s and let them get reset at the next render?

srallen:
I addressed your commits and it has a more visual indication of a change in the dragging action. Sorry about the commits. I'm not sure why those duplicated on github. I just rebased and pushed up and my local git log looks fine. Let me know if you want me to close this and reopen a new PR to have a clean commit history (not sure how to fix this when it's fine locally).

brian-c:
Found a couple more bits, sorry:

- Adding a step should select it.

- The "Delete tutorial" button doesn't seem to work. No delete request is sent.

srallen:
@brian-c I've reverted back to `Math.random()` for the step key. Using the index potentially caused errors with it not being unique if you deleted a step between any number of steps and then tried to add a new one. I've added instead some outline styling which I think makes the drag and drop more visible. I've also fixed the other two items. I left a comment in the code about the fix for the tutorial deletion.

srallen:
Oh and it looks like the duplicated commits on github fixed itself.

brian-c:
Looks good, thanks!

PR #2642, opened 2016-05-31
Style long survey choices

### LINKED ISSUE #2471: Display of choices wonky on mobile for survey tasks
any row where a column got two lines now has black spaces above and below the one-line columns. presumably, the cell with the "choice" inside should expand for all columns in those rows.

img_1063

brian-c:
Closes #2471.

Also fixes the dev classifier on iOS. I don't entirely understand the problem, so I'm going to look into it a bit more, but this'll at least work around it.

### DIFF
--- a/app/pages/dev-classifier/index.cjsx
+++ b/app/pages/dev-classifier/index.cjsx
@@ -20,7 +20,7 @@ ClassificationViewer = React.createClass
     @_boundForceUpdate = null
 
   ignoreUnderscoredProperties: (key, value) ->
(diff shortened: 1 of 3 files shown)


brian-c:
Calling
```
JSON.stringify(['a'], function(key, value) {
  console.log(typeof key);
  return value;
});
```
Chrome and Firefox log "string" twice (the first is an empty string used as a key for the whole first argument) and Safari logs "string" and then "number".

Safari does however correctly return "string" from `typeof Object.keys(['a'])[0]`.

PR #2511, opened 2016-04-29
Updates publications page

mrniaboc:

PR #2653, opened 2016-06-01
Publication page open access

### LINKED ISSUE #1850: Publications page needs additional field: open-access link
Not all journals (etc) our researchers submit to will be open-access, but all our publications should be accessible to everyone. Mostly researchers get around this by having open-access versions somewhere on the web (arXiv for physics, but other solutions exist for other fields). Sometimes they're identical, but more often they're slightly different re: typesetting and such.

We should have a field for an open-access version of every publication, so we don't have to choose between linking to the official publication and linking to an open-access version.

srallen:
Closes #1850. Adds a `openAccess` property to the publications object and displays it next to the original href. Staged at: https://publication-page-open-access.pfe-preview.zooniverse.org/about/publications

@willettk, @mrniaboc is this ok? This supercedes #2511


willettk:
Thumbs up. Thanks, @srallen.

PR #2656, opened 2016-06-02
Use Webpack!

brian-c:
Lots of hard work from @amyrebecca and some rebasing and petty tweaking of scripts from me.

`npm run stage` now stages to the branch subdomain. Having everybody stage to the same place was pretty silly.

Continued-from and closes #2524 (had to make a new branch, so it gets a new PR, sorry).

PR #2657, opened 2016-06-02
Makes hot-reload optional

parrish:
Yay, a contentious PR!

Since it's really just down to developer preference, this allows hot-reloading to be optionally enabled.

`npm start` to serve without it and `npm run start-dev` to serve with it.  Happy to flip those around or rename based on preference.

cc/d everybody assigned from #2656.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,6 +3,11 @@
   "env": {
     "development": {
       "plugins": [
+        ["transform-react-jsx"]
+      ]
+    },
+    "hot-reload": {
+      "plugins": [
         ["react-transform", {
           "transforms": [{
             "transform": "react-transform-hmr",
--- a/dev-server.js
+++ b/dev-server.js
@@ -18,7 +18,10 @@ const middleware = webpackMiddleware(compiler, {
 });
 
 app.use(middleware);
-app.use(webpackHotMiddleware(compiler));
+
+if (process.env.BABEL_ENV === 'hot-reload') {
+  app.use(webpackHotMiddleware(compiler));
+}
 
 app.use(express.static(path.join(__dirname, 'dist')));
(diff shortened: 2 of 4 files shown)


parrish:
It's also worth noting that [react-transform-hmr](https://github.com/gaearon/react-transform-hmr) is deprecated now.

amyrebecca:
I like the idea of leaving it up to developer preference honestly although when bugs appear we'll need to make sure we clarify when the behavior appears (if there were no chance of bugs being introduced by hot reload--and I'll admit I haven't seen any in a long time but you're right that there might be--then there'd be no need to let people choose).

react-transform-hmr IS deprecated, but as of the time I was doing this PR, its "replacement" wasn't ready for prime time yet :) We should track its progress and drop it in when it's in at least a beta state.

brian-c:
I dig it. Are we overloading the NODE_ENV variable though? We should be able to run a local hot server against production if we really wanted to for some reason.

And what about `serve-hot` for the script, to line up better with `serve-static`, since they're both alternatives to the normal start script.

brian-c:
Also, I don't seem to be getting the "webpack: bundle is now VALID." message anymore, on either `npm start` or `npm run start-dev`.

parrish:
> We should be able to run a local hot server against production if we really wanted to for some reason.

It looks like the alternative to setting it in the environment is to inject the reload plugin into every `config.module.loaders` that needs it, which I suppose would work if we wanted to make it env independent.

> And what about serve-hot for the script, to line up better with serve-static

Sounds good

parrish:
Updated.  Apparently the easiest approach here is to set `BABEL_ENV` instead -- it's preferentially used over `NODE_ENV` by babel.

PR #2663, opened 2016-06-03
Clean up a bit

brian-c:
A couple little things that don't justify their own PRs.

- Logo and logotype components render their own shapes instead of trying to reuse a single global instance. These were getting lost when starting from a non-root page, and the optimization is unnecessary anyway. Also cleaned up their titles and title IDs.

- Styled the subject-flag button to match the others

- Removed some unused/outdated files


parrish:
👍

PR #2666, opened 2016-06-06
Updates to the Geordi client

eatyourgreens:
Not ready to merge yet.

### DIFF
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -27,13 +27,11 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
       subjectGetter: () => @keys?.subjectID
 
   makeHandler: (defType) -> # Once defined, efficiently logs different data to same event type
-    instance = @instance
-    (eventData, eventType) ->
+    (eventData, eventType) =>
         eventType = defType if typeof eventType isnt 'string'
-        if GeordiLogger.tokens.indexOf(instance().projectToken) > -1
-          instance()?.logEvent
-            type: eventType
-            data: "\"#{eventData}\""
+        @logEvent
+          type: eventType
+          data: "\"#{eventData}\""
 
   logEvent: (logEntry) -> # Accepts key/values to make appropriate Geordi logging
     newEntry = Object.assign {}, logEntry, @keys
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -11,7 +11,8 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
 
-  keys: {}
+  keys: 
+    projectToken: 'zooHome'
 
   geordi: null
 
@@ -20,8 +21,8 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
 
   makeGeordi: (projectSlug) ->
     new GeordiClient
-      server: @state?.env
-      projectToken: projectSlug || ''
+      env: @state?.env
+      projectToken: projectSlug || @keys?.projectToken
       zooUserIDGetter: () => @state.user?.id
       subjectGetter: () => @keys?.subjectID
 
@@ -35,24 +36,23 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
             data: "\"#{eventData}\""
 
   logEvent: (logEntry) -> # Accepts key/values to make appropriate Geordi logging
-    geordi = @instance()
     newEntry = Object.assign {}, logEntry, @keys
     if GeordiLogger.tokens.indexOf(newEntry.projectToken) > -1
-      @geordi?.logEvent newEntry
-    console.log 'No logger available for event ', JSON.stringify(logEntry) unless @geordi?.logEvent
+      @instance().logEvent newEntry
+      console.log 'No logger available for event ', JSON.stringify(logEntry) unless @instance().logEvent
 
   remember: (keyVals) ->
-    rebuild = keyVals?.projectToken? && (keyVals?.projectToken != @geordi?.projectToken)
-    @geordi = @makeGeordi keyVals.projectToken if rebuild
+    reset = keyVals?.projectToken? && (keyVals?.projectToken != @geordi?.projectToken)
+    @instance().update {projectToken: keyVals.projectToken} if reset
     @keys = Object.assign {}, @keys, keyVals
 
   forget: (forgetKeys) ->
-    rebuild = false
+    reset = false
     for key in forgetKeys
-      rebuild = true if key == 'projectToken'
+      reset = (key == 'projectToken')
       delete @keys[key]
 
-    @geordi = @makeGeordi '' if rebuild
+    @instance().update {projectToken: 'zooHome'} if reset
 
 module.exports = React.createClass
   displayName: 'PanoptesApp'
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -70,7 +70,7 @@ module.exports = React.createClass
   getEnv: ->
     reg = /\W?env=(\w+)/
     browser_env = window?.location?.search?.match(reg)
-    @state?.env || browser_env || 'staging'
+    @state?.env || browser_env?[1] || 'staging'
 
   getInitialState: ->
     user: null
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Staged at https://geordi-panoptes-client.pfe-preview.zooniverse.org/

eatyourgreens:
Forgot to add, this depends on zooniverse/geordi-client#8

alexbfree:
I did some testing and it looks good. "zooHome" is logged in DB, and I don't see "unspecified" anymore. 

A couple of comments:
- I think that we should have a separate project ID for Zoo-wide talk, "zooTalk" - at the moment it is now logging "zooHome" for these.
- Unrelated to this PR, but just noting here something i noticed in testing (cc @wgranger) - The data field is not properly/consistently populated when clicking around in talk - boards are OK but otherwise it's sometimes null and sometimes empty string.

app/partials/app.cjsx
```diff
@@ -32,23 +33,22 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
           data: "\"#{eventData}\""
 
   logEvent: (logEntry) -> # Accepts key/values to make appropriate Geordi logging
-    geordi = @instance()
     newEntry = Object.assign {}, logEntry, @keys
-    geordi?.logEvent newEntry
+    @instance().logEvent newEntry
     console.log 'No logger available for event ', JSON.stringify(logEntry) unless geordi?.logEvent
```
wgranger: Line 38 should probably say `unless @instance().logEvent`

eatyourgreens:
OK @wgranger @alexbfree I think this is good to go, once zooniverse/geordi-client#8 is merged.

app/partials/app.cjsx
```diff
@@ -72,7 +70,7 @@ module.exports = React.createClass
   getEnv: ->
     reg = /\W?env=(\w+)/
     browser_env = window?.location?.search?.match(reg)
-    @state?.env || browser_env || 'staging'
+    @state?.env || browser_env[1] || 'staging'
```
alexbfree: could we add a comment to explain what browser_env[1] is?

app/partials/app.cjsx
```diff
@@ -20,39 +21,36 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
...
+          data: "\"#{eventData}\""
 
   logEvent: (logEntry) -> # Accepts key/values to make appropriate Geordi logging
-    geordi = @instance()
     newEntry = Object.assign {}, logEntry, @keys
     if GeordiLogger.tokens.indexOf(newEntry.projectToken) > -1
-      @geordi?.logEvent newEntry
-    console.log 'No logger available for event ', JSON.stringify(logEntry) unless @geordi?.logEvent
+      @instance().logEvent newEntry
+      console.log 'No logger available for event ', JSON.stringify(logEntry) unless @instance().logEvent
```
alexbfree: I suggest console.warn instead of console.log

Also can we say "No Geordi logger available for event" so it's really clear what we are talking about?

app/partials/app.cjsx
```diff
@@ -11,7 +11,8 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
 
-  keys: {}
+  keys: 
+    projectToken: 'zooHome'
```
alexbfree: I don't think the defaulting to zooHome should be part of geordi-client. That is PFE-specific and should be passed in. Remember the client should be usable on non-PFE sites.

package.json
```diff
@@ -40,7 +40,7 @@
     "swipe-js-iso": "~2.0.1",
     "test-shape-closeness": "0.0.1",
     "visible-element": "~1.0.1",
-    "zooniverse-geordi-client": "~1.3.7"
+    "zooniverse-geordi-client": "zooniverse/geordi-client#panoptes-client"
```
alexbfree: I think since we already update geordi-client that we can update this now to 1.3.10 ? Maybe it should be ^1.3.10 ?

app/partials/app.cjsx
```diff
@@ -20,39 +21,36 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
...
   logEvent: (logEntry) -> # Accepts key/values to make appropriate Geordi logging
-    geordi = @instance()
     newEntry = Object.assign {}, logEntry, @keys
     if GeordiLogger.tokens.indexOf(newEntry.projectToken) > -1
-      @geordi?.logEvent newEntry
-    console.log 'No logger available for event ', JSON.stringify(logEntry) unless @geordi?.logEvent
+      @instance().logEvent newEntry
+      console.log 'No logger available for event ', JSON.stringify(logEntry) unless @instance().logEvent
 
   remember: (keyVals) ->
```
alexbfree: we could maybe add a comment to this and the method below to explain what it does and what the format of keyVals is

app/partials/app.cjsx
eatyourgreens: It will be the first matched substring from the regular expression, `(\w+)` in the line above. Maybe it would be better to use a more reliable method to get the environment?

app/partials/app.cjsx
```diff
@@ -11,7 +11,8 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
```
alexbfree: What is the purpose of this? is it to ensure that we cannot post any other token to Geordi than one of these? Will we need to maintain a hardcoded list of valid tokens in future? That's probably not viable once we support logging on all / arbitrary projects.

package.json
eatyourgreens: 1.3.10 doesn't have the `update()` method used here but this line should update once that change is available in the npm module.

alexbfree:
Generally looks good. Thanks. I will wait and see how you respond to comments before i merge.

app/partials/app.cjsx
alexbfree: I'm not sure, I think @wgranger wrote this. Comments Will?

app/partials/app.cjsx
eatyourgreens: From what I can tell, this stores tokens like `subjectID` so that they're available to be added to event data. I'm not 100% sure.

package.json
alexbfree: Sorry, i have just seen the other PR  for geordi-client - I remembered already merging one but that was functional tests. Will look now. then will merge and create 1.3.11 then we can update here.

app/partials/app.cjsx
eatyourgreens: Yeah, this is a filter to log Gravity Spy. I have no better ideas for filtering logging so that we aren't recording data for every project.

app/partials/app.cjsx
eatyourgreens: Oh no, I wrote it. I'm just not sure if this is the best method for getting the environment.

package.json
alexbfree: OK 1.3.11 is now available, please update to `^1.3.11`

app/partials/app.cjsx
alexbfree: or alternatively, how do we do this elsewhere?

eatyourgreens: It's basically copied from the Panoptes client here: https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js#L76

eatyourgreens:
@alexbfree Thanks! I've pushed some changes in response.

app/partials/app.cjsx
eatyourgreens: Wait, I'm an idiot. The API client config will have the environment already, so we can just read it from there instead of getting it from `window.location` again.

eatyourgreens: No, I'm wrong, the API client doesn't export `env`, just the various API hosts: https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js#L59-L66

But I should probably fix this so it doesn't bomb out if there is no `env` query parameter in the window location.

eatyourgreens:
Updated now so that the check for the `env` query parameter works.

app/partials/app.cjsx
wgranger: Right. This seemed to be the best course for now since we are restricting logging to Gravity Spy. We will likely need to change this in the future as we expand the projects that contain logging from geordi.

alexbfree: OK sounds good

alexbfree:
OK this LGTM

PR #2667, opened 2016-06-06
Fixes #1780

### LINKED ISSUE #1780: Allow admins to edit all projects
It'd be great if there was a UI for admins to edit any arbitrary projects. I was thinking specifically so that I can add collaborators to Zooniverse-owned projects without having to log in as zooniverse, but it'd probably make sense to just let admins use the whole project builder UI for every project.

alexbfree:
Allow admins access to project builder on all projects.

Deployed to https://allow-admins-into-project-builder-on-all-projects.pfe-preview.zooniverse.org/ for testing. Add the production environment parameter to test on production projects.

To test, go into admin mode, then add a project ID to the end of 
`https://allow-admins-into-project-builder-on-all-projects.pfe-preview.zooniverse.org/lab/`

Usecases and background in https://github.com/zooniverse/rfc/issues/4

### DIFF
--- a/app/pages/lab/project.cjsx
+++ b/app/pages/lab/project.cjsx
@@ -10,7 +10,7 @@ Router = require 'react-router'
 ModalFormDialog = require 'modal-form/dialog'
 WorkflowCreateForm = require './workflow-create-form'
 workflowActions = require './actions/workflow'
-isAdmin = require '../lib/is-admin'
+isAdmin = require '../../lib/is-admin'
 
 DEFAULT_SUBJECT_SET_NAME = 'Untitled subject set'
 DELETE_CONFIRMATION_PHRASE = 'I AM DELETING THIS PROJECT'
--- a/app/pages/lab/project.cjsx
+++ b/app/pages/lab/project.cjsx
@@ -10,6 +10,7 @@ Router = require 'react-router'
 ModalFormDialog = require 'modal-form/dialog'
 WorkflowCreateForm = require './workflow-create-form'
 workflowActions = require './actions/workflow'
+isAdmin = require '../lib/is-admin'
 
 DEFAULT_SUBJECT_SET_NAME = 'Untitled subject set'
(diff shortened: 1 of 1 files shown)


app/pages/lab/project.cjsx
```diff
@@ -252,7 +252,7 @@ module.exports = React.createClass
           <p className="form-help">Loading project</p>
         </div>
       } then={([project, owners]) =>
-        if @props.user in owners
+        if @props.user in owners or @props.user.admin
```
brian-c: Use the function at `lib/is-admin` to check if the user is properly in admin mode.

alexbfree: OK. What is the difference between `!!apiClient.params.admin` and `@props.user.admin` ?

brian-c: `user.admin` just indicates that a user can _access_ admin mode.

`apiClient.params.admin` indicates if admin mode is actually switched on.

app/pages/lab/project.cjsx
```diff
@@ -10,6 +10,7 @@ Router = require 'react-router'
 ModalFormDialog = require 'modal-form/dialog'
 WorkflowCreateForm = require './workflow-create-form'
 workflowActions = require './actions/workflow'
+isAdmin = require '../lib/is-admin'
```
brian-c: Needs one more `../`.

alexbfree: yep, sorry, prematurely committed before i finished testing!

app/pages/lab/project.cjsx
alexbfree: I see. Thanks for the explanation. And presumably `apiClient.params.admin` also tells you they are an admin AND have it turned on (since they couldn't have turned it on otherwise)

alexbfree:
Redeployed now at https://allow-admins-into-project-builder-on-all-projects.pfe-preview.zooniverse.org/

Seems to be working.

PR #2668, opened 2016-06-06
Don't link to inactive workflows

### LINKED ISSUE #2658: can make an inactive workflow show up if you click classify if you have  check box ticked in workflow editor for default workflow
While starting the process to turn on the HSC Comet Hunters workflow I came across this. I checked the box to make it the default workflow, but the workflow is still inactive and if you're not logged in and hit classify it took me this non-active workflow. 

Naively I thought that because the workflow was inactive the website would automatically switch to making another active workflow default when you hit classify. I think project owners will forget to unclick this checkmark if they have multiple workflows and switch the default workflow to inactive in the visablity page

brian-c:
Inactive workflows that were stored as the project's default or in the user's preferences were still be linked from the project-level navigation.

This checks for active-ness before setting the query on that link.

Closes #2658.

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -43,10 +43,12 @@ ProjectPage = React.createClass
     background: null
     avatar: null
     pages: []
+    selectedWorkflow: null
 
   componentDidMount: ->
     document.documentElement.classList.add 'on-project-page'
     @fetchInfo @props.project
+    @getSelectedWorkflow @props.project, @props.preferences
     @updateSugarSubscription @props.project
 
   componentWillUnmount: ->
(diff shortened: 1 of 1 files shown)


app/pages/project/index.cjsx
```diff
@@ -82,6 +87,20 @@ ProjectPage = React.createClass
       .then (pages) =>
         @setState {pages}
 
+  getSelectedWorkflow: (project, preferences) ->
+    @setState selectedWorkflow: 'PENDING'
+
+    preferredWorkflowID = preferences?.preferences.selected_workflow ? project.configuration?.default_workflow
+    if preferredWorkflowID?
+      apiClient.type('workflows').get preferredWorkflowID
```
camallen: should be able to filter on active here as well as the workflow id if you want instead.

camallen:
LGTM - not blocking on the api call `active=true` query param.

app/pages/project/index.cjsx
brian-c: Yeah, it felt kinda weird to query since we've already got the ID. I think the `if workflow.active` check is more explicit. Does this affect the back end performance-wise at all?

camallen: you can combine the two calls id + active=true, shouldn't affect performance in the API (if you notice it is then we're doing something wrong).

camallen: sorry - just re-read..i really don't mind. I agree the code is explicit here without more knowledge of the workflow active filters.

Issue #2212, opened 2016-02-15
Common Navigation Needed for User Profiles to Solve Dead-Ends


alexbfree:
I noticed that it is very awkward to navigate between your profile, collection, favourites, and stats, as many of the menu options disappear

When you first click "Profile", you get

![screenshot 2016-02-15 16 53 04](https://cloud.githubusercontent.com/assets/1473244/13055159/a615fb78-d405-11e5-9e60-bd84cc88d1fc.png)

If you then click Collections or Favourites you get:

![screenshot 2016-02-15 16 53 09](https://cloud.githubusercontent.com/assets/1473244/13055175/c1abf7b6-d405-11e5-80fe-0330fbbf1c56.png)

Notice the "Recent Comments" option has disappeared and is no longer accessible, nor is "Your stats"

If you then click into Collections, or Favourites, you get
![screenshot 2016-02-15 16 54 58](https://cloud.githubusercontent.com/assets/1473244/13055196/db751740-d405-11e5-8b23-a13807af69ba.png)
![screenshot 2016-02-15 16 53 26](https://cloud.githubusercontent.com/assets/1473244/13055195/db74738a-d405-11e5-91d7-d3d4588b3380.png)

Notice that there is no way to navigate to Recent Comments, or Your Stats, and no way to navigate to the opposite one of Collections/Favourites.

We should have a common navigation across all "Profile" pages. Right now you have to click back on your Profile and start all over again if you are in anything other than Recent Comments.

alexbfree:
Further to this, the navigation when viewing another user's profile is also highly confusing.
For example when visiting 
https://www.zooniverse.org/users/gaga7

You are subject to all the problems above, but also, when you click Collections, you get to 
https://www.zooniverse.org/collections/gaga7 

which contains the same "submenu"
![screenshot 2016-02-16 14 59 21](https://cloud.githubusercontent.com/assets/1473244/13079807/e9d2d224-d4bd-11e5-818a-40176ee64aa8.png)

however none of these options relate to the user you are viewing, nor the current view. All includes all collections from everyone, and the other two links take you into your own collections and favourites.

alexbfree:
Another related issue. If you click View collection from the All Collections view, not only do you lose the context of which user's collection it is (only user avatar remains) but you have no nav menu at all
e.g. https://www.zooniverse.org/collections/gardenmaeve/curious
[...]

PR #2670, opened 2016-06-07
[ready for merge if approved] Ensure preservation of project context

### LINKED ISSUE #2253: Ensuring Preservation of Project Context

alexbfree:
Fixes #2253.

NB This supersedes the previous approach which was overcomplicated and confusing.

## The problem(s) being solved:

- No way to view the collections/favorites specific to a project 
- Very easy to lose the fact you are "in" a project and drop back to Zoo-wide context, simply by clicking on a collection, user profile etc. This most often happens in Project Talk.

This PR can be tested on staging at https://maintain-project-context.pfe-preview.zooniverse.org/ (add `?env=production` to use on live projects)

# New Features 

This PR introduces:

## 1. A new "Collect" menu within Projects

screenshot 2016-06-07 14 58 15 2

This new "Collect" menu (analogous to the Zoo-wide one) allows the user to browser all collections for a project, or just the collections/favourites within that project for a specific user.

## 2. Ability to view a user within the context of a project

screenshot 2016-06-07 14 56 40 2

This ensures that when you click on a username in Project Talk, you remain in the context of that project.
Comments are filtered to only show those for this project, and 

## 3. Adapt all collection, favorites & user menus according to what currently makes sense

For example:
screenshot 2016-06-07 15 15 53 2

The options available will carefully adapt according to:

- **Base Type** - are you looking at collections, favorites, a user profile or recent comments
- **Filter** - are you looking at the data for a user, a project, everything, or a user+project 
- **Context** - are you currently within the page/area that corresponds to a user, a project, everything, or a user+project 
- **Perspective** - are you logged out, logged in as the same user being looked at, or logged in as a different user than the one being looked at.

Key factors influencing the menu design were:

- It should always be clear what context you are in/what you are looking at
- As you click within a single menu, the options should not jump around confusingly 
- The line of menu links should never get too long - this involved shortening text and ensuring there are not too many options at any given time.
- You should be able to change focus between the whole project or a specific user without confusion.

## 4. A link that lets you explicitly step from a project to the wider Zooniverse

screenshot 2016-06-07 14 57 57 2

This link exists on all the collection, favourite and user profile menus, and essentially means "show me everything for this user, not just on this project but on all projects".
Several different texts were considered - "Exit project", "Leave project", "View on Zooniverse.org" but @chrislintott came up with the slightly playful "To the Zooniverse!". This is a good option because it injects a bit more fun into the site, while still being clear about what it is doing. 

## 5. Menu navigation when viewing a collection

screenshot 2016-06-07 15 23 55

As described in #2212, it's easy to hit a navigational dead end when viewing a collection. By adding a couple of contextual menu options, we are now able to ensure there is always a way back to that user's collections, or to that user's profile - keeping the project context if there is one.

## 6. Ensure that no links can accidentally take you out of a project

Now, to leave a project, you need to explicitly click "To The Zooniverse!" or go to the main Talk link in the main Zooniverse navigation. Clicking a user profile or collection within Project Talk, will never take you out of the project.

----

Note: all screenshots above are from *within* the project context, but the equivalent non-project menus for collections, favourites and user profiles have also been updated. For example:

screenshot 2016-06-07 15 26 44

----

Please try out this feature on staging, for example at 
https://maintain-project-context.pfe-preview.zooniverse.org/projects/alexbfree/computer-vision-serengeti/collections/?env=production

### DIFF
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -48,7 +48,6 @@ module.exports = React.createClass
       "#{@props.translationObjectName}.title.generic"
 
   getOwnerForTitle: ->
-    console.log 'filter is ',@props.filter
     if @props.filter? and "owner" of @props.filter
       return @props.nonBreakableOwnerName
 
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -27,6 +27,8 @@ module.exports = React.createClass
     document.documentElement.classList.remove 'on-secondary-page'
 
   getMessageKeyToUseForTitle: ->
+    console.log @props.filter
+    console.log @props.filter?, "project_ids" of @props.filter, "owner" of @props.filter, @viewingOwnCollections
     if @props.filter?
       if "project_ids" of @props.filter
         if "owner" of @props.filter
@@ -48,6 +50,7 @@ module.exports = React.createClass
       "#{@props.translationObjectName}.title.generic"
 
   getOwnerForTitle: ->
+    console.log "nbo",@props.nonBreakableOwnerName
     if @props.filter? and "owner" of @props.filter
       return @props.nonBreakableOwnerName
 
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -36,13 +36,19 @@ module.exports = React.createClass
     {location} = @props
 
     <div className="secondary-page all-resources-page">
-      <section className={"hero #{@props.heroClass}"}>
-        <div className="hero-container">
+      {if @props.project?
+        <div className="hero-with-context">
           <Translate component="h1" user={@userForTitle()} content={"#{@props.translationObjectName}.title"} />
-          {if @props.heroNav?
-            @props.heroNav}
         </div>
-      </section>
+      else
+        <section className={"hero #{@props.heroClass}"}>
+          <div className="hero-container">
+            <Translate component="h1" user={@userForTitle()} content={"#{@props.translationObjectName}.title"} />
+            {if @props.heroNav?
+              @props.heroNav}
+          </div>
+        </section>}
+
       <section className="resources-container">
         <PromiseRenderer promise={@props.listPromise}>{(ownedResources) =>
           if ownedResources?.length > 0
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -26,11 +26,35 @@ module.exports = React.createClass
   componentWillUnmount: ->
     document.documentElement.classList.remove 'on-secondary-page'
 
-  userForTitle: ->
-    if @props.ownerName
-      "#{@props.ownerName}'s"
+  getMessageKeyToUseForTitle: ->
+    if @props.filter?
+      if "project_ids" of @props.filter
+        if "owner" of @props.filter
+          if @viewingOwnCollections
+            "#{@props.translationObjectName}.title.project.ownedBySelf"
+          else
+            "#{@props.translationObjectName}.title.project.ownedByOther"
+        else
+          "#{@props.translationObjectName}.title.project.allOwners"
+      else
+        if "owner" of @props.filter
+          if @viewingOwnCollections
+            "#{@props.translationObjectName}.title.allProjects.ownedBySelf"
+          else
+            "#{@props.translationObjectName}.title.allProjects.ownedByOther"
+        else
+          "#{@props.translationObjectName}.title.allProjects.allOwners"
     else
-      'All'
+      "#{@props.translationObjectName}.title.generic"
+
+  getOwnerForTitle: ->
+    console.log 'filter is ',@props.filter
+    if @props.filter? and "owner" of @props.filter
+      return @props.nonBreakableOwnerName
+
+  getProjectForTitle: ->
+    if @props.filter? and "project_ids" of @props.filter
+      return @props.nonBreakableProjectName
 
   getPageClasses: ->
     classes = 'secondary-page all-resources-page'
@@ -44,7 +68,7 @@ module.exports = React.createClass
     <div className={@getPageClasses()}>
       <section className={"hero #{@props.heroClass}"}>
         <div className="hero-container">
-          <Translate component="h1" user={@userForTitle()} content={"#{@props.translationObjectName}.title"} />
+          <Translate component="h1" owner={@getOwnerForTitle()} project={@getProjectForTitle()} content={"#{@getMessageKeyToUseForTitle()}"} />
           {if @props.heroNav?
             @props.heroNav}
         </div>
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -32,23 +32,23 @@ module.exports = React.createClass
     else
       'All'
 
+  getPageClasses: ->
+    classes = 'secondary-page all-resources-page'
+    if @props.project?
+      classes += ' has-project-context'
+    classes
+
   render: ->
     {location} = @props
 
-    <div className="secondary-page all-resources-page">
-      {if @props.project?
-        <div className="hero-with-context">
+    <div className={@getPageClasses()}>
+      <section className={"hero #{@props.heroClass}"}>
+        <div className="hero-container">
           <Translate component="h1" user={@userForTitle()} content={"#{@props.translationObjectName}.title"} />
+          {if @props.heroNav?
+            @props.heroNav}
         </div>
-      else
-        <section className={"hero #{@props.heroClass}"}>
-          <div className="hero-container">
-            <Translate component="h1" user={@userForTitle()} content={"#{@props.translationObjectName}.title"} />
-            {if @props.heroNav?
-              @props.heroNav}
-          </div>
-        </section>}
-
+      </section>
       <section className="resources-container">
         <PromiseRenderer promise={@props.listPromise}>{(ownedResources) =>
           if ownedResources?.length > 0
--- a/app/components/owned-card-list.cjsx
+++ b/app/components/owned-card-list.cjsx
@@ -5,6 +5,7 @@ Translate = require 'react-translate-component'
 apiClient = require 'panoptes-client/lib/api-client'
 PromiseRenderer = require '../components/promise-renderer'
 OwnedCard = require '../partials/owned-card'
+ContextualLinks = require '../lib/contextual-links'
 {Link} = require 'react-router'
 
 module.exports = React.createClass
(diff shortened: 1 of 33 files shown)


alexbfree:
OK I have now made all the updates following feedback, staging is updated. If people approve then it's good to merge.

brian-c:
Will review further tomorrow, but at a quick glance, all these `!important`s need to go. Fix selector specificity and if you're up for it, start using BEM.

app/components/owned-card-list.cjsx
```diff
@@ -26,34 +27,45 @@ module.exports = React.createClass
...
-  userForTitle: ->
-    if @props.ownerName
-      "#{@props.ownerName}'s"
-    else
-      'All'
+  getPageClasses: ->
+    classes = 'secondary-page all-resources-page'
+    if @props.project?
+      classes += ' has-project-context'
+    classes
```
srallen: You could use [classnames](https://www.npmjs.com/package/classnames) for conditionally adding classes to components. It's already in `package.json` and a fairly clean and readable.

app/lib/contextual-links.coffee
```diff
@@ -0,0 +1,435 @@
...
+    if props.favorite and props.params?.favorites_owner?
+      return props.params.favorites_owner
+    else if !props.favorite and props.params?.collection_owner?
+      return props.params.collection_owner
+    else if !props.project? and props.params?.owner?
+      return props.params.owner
+    else if props.user?
+      return props.user.login
+    else
+      return props.params.owner # which may be null
```
srallen: Coffeescript has implicit returns, so these aren't necessary here.

app/lib/contextual-links.coffee
```diff
@@ -0,0 +1,435 @@
...
+    if firstPart == "projects" and pathParts.length < 6 and lastPart != "all"
+      filters["project_ids"] = props.project.id
+    if firstPart == "collections" and pathParts.length == 2 and pathParts[1] != "" and pathParts[1] != "all"
+      filters["owner"] = pathParts[1]
+    if firstPart == "favorites" and pathParts.length == 2 and pathParts[1] != "" and pathParts[1] != ""
+      filters["owner"] = pathParts[1]
+    if pathParts.length>4 and pathParts[3] == "collections" and pathParts[4] != "" and pathParts[4] != "all"
+      filters["owner"] = pathParts[4]
+    if pathParts.length>4 and pathParts[3] == "favorites" and pathParts[4] != "" and pathParts[4] != "all"
+      filters["owner"] = pathParts[4]
```
srallen: If this is going to be in coffeescript, could you be consistent with using its syntax, such as using `is` instead of `==` and `isnt` instead of `!=`. In javascript, as an aside, you should use `===` or `!==`.

parrish:
I'll probably dig into reviewing the code on Monday, but here's some high level feedback:

- On /projects/owner/name/users/username
  - Request to /comments?user_id=user_id&sort=-created_at&project_id=project_id should use section instead of project_id (`"project-#{ project.id }"`)

  - "View username's Zooniverse.org Profile"
    - project owner name instead of user (see image), links to /users/
    - inconsistent naming (should be "Profile")
    - username header should probably also link to profile

  - /projects/owner/name/collections/username should probably be /projects/owner/name/username/collections
  - /projects/owner/name/favorites/username should probably be /projects/owner/name/username/favorites

  - On /projects/owner/name/collections/username
    - "All" link unexpectedly links to all project collections rather than all user collections (Same is true on /collections/username)
    - "My collections" is also unexpected in the scope of this navigation
    - Clearly delineating between user and project in the navigation might help this

- Inconsistent pagination throughout

- Pagingation color needs visibility improvements

owner_link

pagination1
pagination2

brian-c:
- The `/project/:owner/:name/collections/all` route is broken and should not exist anyway (it should refer to a collection owner named "all").

- There are several more `/all` routes. These should probably be handled by an IndexRoute, as is done in the rest of the app.

- In general, don't hard code non-breaking spaces into output, use `white-space: nowrap`

- ContextualLinks module is really, really ... a lot. These are basically just relative links. If this kind of string-building module is needed to work with some other components that aren't flexible enough (OwnedCardList, perhaps?), those components should be rewritten, not built on top of. This is an absolutely gargantuan code smell. Probably from OwnedCards, which should be replace whenever they're encountered.

Generally I feel like this isn't making any complex changes, it's just adding some routes that present a scoped view of existing information. If we can't do this in a simple way because of some existing code, we need to fix that code, not add another level of complexity on top to work around it.

srallen:
I agree about the `ContextualLinks` module. When starting to review this yesterday, my initial impression of it was that it seemed to be an overly complex solution. Is there a reason that this approach was taken?

parrish:
Okay, I've worked through the code about as much as I'm able to.  In general, I agree with the previous comments, but here's some specific feedback.

Talk-Api uses sections (e.g. `section: 'project-123'`), not project id's
  - No request should specify project_id: the fact that it works currently is a bug that will be fixed.

`ContextualLinks`: I don't have hard numbers in mind, but 400+ lines of code is certainly too big.  Some of it seems unnecessary as well.  For instance
   - [`getContextUserLogin`](https://github.com/zooniverse/Panoptes-Front-End/blob/03f47a81f0f6b7dfcc92910bd00418809a28c96d/app/lib/contextual-links.coffee#L51): should be inlined rather than generalized
   - [`makeTextUnbreakable`](https://github.com/zooniverse/Panoptes-Front-End/blob/03f47a81f0f6b7dfcc92910bd00418809a28c96d/app/lib/contextual-links.coffee#L64): should not be used
   - [`prefixLinkIfNeeded`](https://github.com/zooniverse/Panoptes-Front-End/blob/03f47a81f0f6b7dfcc92910bd00418809a28c96d/app/lib/contextual-links.coffee#L68): should be inlined or passed as a prop
   - [`getFiltersFromPath`](https://github.com/zooniverse/Panoptes-Front-End/blob/03f47a81f0f6b7dfcc92910bd00418809a28c96d/app/lib/contextual-links.coffee#L76) and [`safelyGetPath`](https://github.com/zooniverse/Panoptes-Front-End/blob/03f47a81f0f6b7dfcc92910bd00418809a28c96d/app/lib/contextual-links.coffee#L102): should use react-router rather than parsing the url as a string

I tried to come up with some constructive refactoring, but I think the best course of action is to just rethink the necessity of this module.

I think I've figured out the primary source of complexity here though.

##### It's better to have multiple components or [composed components](http://www.zhubert.com/blog/2016/02/05/react-composability-patterns/) that avoid the logical complexity of using a single component in many different contexts.

CKrawczyk:
I have not taken a close look at this code, but based on its goal could this be a good use case for React's built in `context` https://facebook.github.io/react/docs/context.html ?  It works in a similar way to the "Decorator Pattern" in the link @parrish provided.

alexbfree:
@brian-c I also didn't like putting the `!important`'s there but I am not sure how to fix this in Stylus. The order seems to be predetermined by the order within in each stylus file and the order in which the stylus files are loaded. Do you have any suggestions how I can work around this?

alexbfree:
@srallen I guess the reason for this approach appearing complex is a couple of things - firstly that the problem is actually more complex than you would think - there are a number of different factors affecting how the menu will appear and behave (filter, context, perspective etc) and secondly that it's the first thing I've built in PFE and React and so I don't necessarily yet know the most optimal way to do everything. BTW I am not particularly attached to this being in coffeescript - I just happened to start this before I knew about the plan to move everything to ES6.

alexbfree:
@parrish 
> Talk-Api uses sections (e.g. section: 'project-123')
Can i reliably construct a section ID by appending project ID to "project-" ?

> getContextUserLogin should be inlined rather than generalized
Could you clarify what you mean by that?

Thank you for the links on composed components. I will read up on that. If you have any immediate thoughts on what the composed components should represent, conceptually, that would be great. Just to put me on the right track. Thanks for react router suggestion too, will look at that.

alexbfree:
@parrish 

>"All" link unexpectedly links to all project collections rather than all user collections (Same is true on /collections/username)
>"My collections" is also unexpected in the scope of this navigation

This was actually a deliberate UX/design choice, that everything in the menu is limited by project scope implicitly rather than explicitly and that when you want to view across projects, you are sent to zooniverse home, outside the project.

I realised there isn't a need for users to be able to view collections from other projects within the context of a specific context, and that having options for all, project, user+project within the same navigation is unmanageable and leads to too-long link names and too many options per menu.
I'd be happy to discuss the design thinking with you and discuss improvements but this particular design is the result of several weeks of iteration through less effective UX options.

alexbfree:
Thank you everyone for your feedback. It's my first attempt to build a feature of any size for PFE, and clearly there is a lot I still have to learn about the codebase and the most effective ways to do things (I did go on a React course on Friday so that will help). 

I'm really not sure how best to proceed at this point - Will go away and think about it.

alexbfree:
@parrish 
>/projects/owner/name/collections/username should probably be /projects/owner/name/username/collections
/projects/owner/name/favorites/username should probably be /projects/owner/name/username/favorites

the disadvantage is that you can't then switch between project / non project context by simply adding the project prefix... at the moment all the project specific URLs for collections, favorites, etc mirror the non-project equivalent.

e.g. 
`/projects/owner/name/collections/username` 
follows the pattern of
`/collections/username`

and 
`/projects/owner/name/collections/`
follows the pattern of
`/collections/`

If you do it the other way round you now have to cope with

`/projects/owner/name/collections/` when you want all in project
`/projects/owner/name/username/collections/` for a specific user in a projec and
`/collections/username` for a specific user across all projects

brian-c:
@alexbfree Would you be interested in pairing on this over Screenhero sometime this week?

alexbfree:
@brian-c Thank you for the offer, that sounds like a good way forward. I'll PM you to arrange.

alexbfree:
After working on this with @brian-c this afternoon, we've come up with a new simpler approach. Closing this PR, will make a new one.

PR #2671, opened 2016-06-07
Dynamically increase text task textarea height

mcbouslog:
update text task textarea element, row attribute, based on content typed (specifically by comparing scrollHeight to offsetHeight).

staging site here:  https://text-task-auto-height.pfe-preview.zooniverse.org.

in Firefox the textarea will only increase up to 5 rows, while other browsers (Chrome, Safari, in process of testing on Edge) will increase indefinitely.

please let me know what you think, thanks!

### DIFF
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -64,6 +64,7 @@ module.exports = React.createClass
 
   componentDidMount: ->
     @setState initOffsetHeight: @refs.textInput.offsetHeight
+    @updateHeight()
 
   setTagSelection: (e) ->
     textTag = e.target.value
@@ -84,10 +85,6 @@ module.exports = React.createClass
     newAnnotation = Object.assign @props.annotation, {value}
     @props.onChange newAnnotation
 
-  handleClear: (e) ->
-    if e.which is 8 or e.which is 46
-      @setState textareaHeight: @state.initOffsetHeight
-
   updateHeight: ->
     @setState textareaHeight: @refs.textInput.scrollHeight + 2
 
@@ -102,7 +99,6 @@ module.exports = React.createClass
           onChange={@handleChange}
           rows="1"
           style={height: @state.textareaHeight}
-          onKeyDown={@handleClear}
         />
       </label>
       {if @props.task.text_tags
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -61,27 +61,6 @@ module.exports = React.createClass
   getInitialState: ->
     rows: 1
 
-  render: ->
-    <GenericTask question={@props.task.instruction} help={@props.task.help} required={@props.task.required}>
-      <label className="answer">
-        <textarea autoFocus={@props.autoFocus} className="standard-input full" rows={@state.rows} ref="textInput" value={@props.annotation.value} onChange={@handleChange} />
-      </label>
-      {if @props.task.text_tags
-          <div className="transcription-metadata-tags">
-            {for tag, i in @props.task.text_tags
-              <input type="button" className="standard-button text-tag" key={i} value={tag} onClick={@setTagSelection} />}
-          </div>}
-    </GenericTask>
-
-  handleChange: ->
-    if @refs.textInput.scrollHeight > @refs.textInput.offsetHeight
-      rows = @state.rows + 1
-      @setState rows: rows
-
-    value = @refs.textInput.value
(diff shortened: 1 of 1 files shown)


simensta:
Hey @mcbouslog, I've briefly looked at the staging branch. I think the textarea height should also dynamically decrease when a user deletes large amounts of text.

mcbouslog:
@simensta - great suggestion, updated accordingly!

app/classifier/tasks/text/index.cjsx
```diff
@@ -93,3 +79,50 @@ module.exports = React.createClass
       value = textBefore + startTag + textInBetween + endTag + textAfter
     newAnnotation = Object.assign @props.annotation, {value}
     @props.onChange newAnnotation
+
+  handleClear: (e) ->
+    # keyCode 8 is backspace, 46 is delete
+    if e.which is 8 or e.which is 46
+      @setState rows: 1
+
+  handleResize: ->
```
mcbouslog: handleResize could be simplified to:
```
    heightChange = @refs.textInput.scrollHeight - @refs.textInput.offsetHeight + 2
    return unless heightChange > 0
    rows = @state.rows
    switch heightChange
      when 44, 30
        rows = rows + 2
      when 66, 52
        rows = rows + 3
      when 88, 75
        rows = rows + 4
      else
        rows = rows + 1
    @setState rows: rows
```
but, then the (Chrome, Safari) edge case where there are 6+ rows of text and one character is deleted the textarea would be resized to 2 rows, instead of staying at 6+ rows (which is/will-be the case with Firefox). simply typing any additional character will properly resize the textarea back to 6+ rows, but immediately after deleting a single character, before typing another character, the textarea will be smaller sized than content.

i don't think the extra 3? lines of code are too bad, though it does make things less readable, but open to suggestions.

mcbouslog:
`rows` set to 1 because it defaults to what looks like 2. a style argument could be made to let it be the default height (what looks like 2 rows). though i thought if we're setting the height dynamically might as well make it as small as possible initially, but open to suggestion (initially 1 row vs 2 rows).

tried handling the resizing through just `onKeyDown` or `onKeyUp` with one function, but some use case wouldn't work. the combination of `handleClear` and `updateHeight` through `onKeyDown` and `onChange` seemed to work the best.

use cases that work:
- paste multiple lines, resizes
- type multiple lines, resizes, delete character by character, resizes appropriately
- type multiple lines, resizes, delete the entire content, resizes
- type multiple lines, resizes, delete half the content, resizes

use case that don't work until you press (including delete) one more key:
- type multiple lines, resizes, highlight any/all content and type a key thereby deleting lines, should resize, but doesn't

app/classifier/tasks/text/index.cjsx
```diff
@@ -58,22 +58,12 @@ module.exports = React.createClass
...
-      </label>
-      {if @props.task.text_tags 
-          <div className="transcription-metadata-tags">
-            {for tag, i in @props.task.text_tags
-              <input type="button" className="standard-button text-tag" key={i} value={tag} onClick={@setTagSelection} />}
-          </div>}
-    </GenericTask>
+  getInitialState: ->
+    textareaHeight: undefined
+    initOffsetHeight: undefined
```
brian-c: What's the relevance of the textarea's initial height? Its height should be set automatically on mount.

app/classifier/tasks/text/index.cjsx
```diff
@@ -93,3 +83,43 @@ module.exports = React.createClass
       value = textBefore + startTag + textInBetween + endTag + textAfter
     newAnnotation = Object.assign @props.annotation, {value}
     @props.onChange newAnnotation
+
+  handleClear: (e) ->
```
brian-c: I don't think this works as you intended. This'll fire on every keydown, but it's immediately followed up by the `onChange` handler, so the results are never applied.

app/classifier/tasks/text/index.cjsx
```diff
@@ -58,22 +58,12 @@ module.exports = React.createClass
...
+  getInitialState: ->
+    textareaHeight: undefined
+    initOffsetHeight: undefined
 
-  handleChange: ->
-    value = @refs.textInput.value
-    newAnnotation = Object.assign @props.annotation, {value}
-    @props.onChange newAnnotation
+  componentDidMount: ->
+    @setState initOffsetHeight: @refs.textInput.offsetHeight
```
brian-c: Call `updateHeight` here, or else going back to a text task keeps its height at one row.

brian-c:
Solid, thanks!

PR #2673, opened 2016-06-08
Add nib

### LINKED ISSUE #2672: Formatting problems with iPad (Landscape)
As raised by @mrniaboc :
img_20160608_115136

The nav bar (overflowing/overlapping), the projects slider(missing/only one chevron showing), and the projects list (only showing 1 per row?) all need attention

eatyourgreens:
Should add CSS prefixes back in and close #2672

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "file-loader": "~0.8.5",
     "html-webpack-plugin": "~2.16.0",
     "json-loader": "~0.5.4",
+    "nib": "~1.1.0",
     "publisssh": "~1.1.0",
     "react-transform-hmr": "~1.0.4",
     "rimraf": "~2.5.2",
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Looks like `nib` still needs to be imported somewhere else that I'm missing. Pinging @amyrebecca @brian-c @srallen in case you have any ideas how to get this to work.

brian-c:
Been banging my head against this all morning. May be of help to someone else: I think it has something to do with the dev server. A simple example works fine with just adding `@require './node_modules/nib';` to the top of main.styl as expected.

brian-c:
I think it was a load-order issue with common.styl. Not entirely 100% confident I understand why though. Opened #2680 to replace this PR.

PR #2675, opened 2016-06-08
Ensure that collection owner name is displayed when we have a mix of owners

### LINKED ISSUE #2674: Regression bug: Collection owner names are no longer appearing
I've noticed that when you go to e.g. http://zooniverse.org/collections or http://zooniverse.org/favorites, the code that was written to display the username on that collection is no longer triggering.

Note that it should *not* appear when you go to http://zooniverse.org/collections/username or http://zooniverse.org/favorites/username since those collections/favorites all share the same owner.

This showing of owners is important when you are showing a mix of collections from different users - and can especially cause confusion since the *project* owner's name often appears in favorite collection names, rather than the *collection* owner.

alexbfree:
Fixes #2674.

### DIFF
--- a/app/pages/collections.cjsx
+++ b/app/pages/collections.cjsx
@@ -72,7 +72,7 @@ List = React.createClass
       heroNav={<CollectionsNav user={@props.user} />}
       heroClass="collections-hero"
       ownerName={@props.params?.owner}
-      skipOwner={!@props.params?.owner}
+      skipOwner={@props.params?.owner?}
       imagePromise={@imagePromise}
       cardLink={@cardLink} />
 
--- a/app/partials/owned-card.cjsx
+++ b/app/partials/owned-card.cjsx
@@ -59,9 +59,7 @@ module.exports = React.createClass
         <div className="details">
           <div className="name"><span>{@props.resource.display_name}</span></div>
           {if !@props.skipOwner
-            <PromiseRenderer promise={@props.resource.get('owner')}>{ (owner) ->
-              <div className="owner">{owner?.display_name ? 'LOADING'}</div>
-            }</PromiseRenderer>}
+            <div className="owner">{@props.resource.links.owner.display_name}</div>}
           {<div className="description">{@props.resource.description}</div> if @props.resource.description?}
           {<div className="private"><i className="fa fa-lock"></i> Private</div> if @props.resource.private}
           <button type="button" tabIndex="-1" className="standard-button card-button"><Translate content={"#{@props.translationObjectName}.button"} /></button>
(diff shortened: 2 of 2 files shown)


alexbfree:
Deployed to https://fix-missing-owner-names.pfe-preview.zooniverse.org/ for testing:

https://fix-missing-owner-names.pfe-preview.zooniverse.org/collections
https://fix-missing-owner-names.pfe-preview.zooniverse.org/favorites

srallen:
I believe the API call to fetch the owner for each collection was taken out on purpose for performance reasons in #2029. However, this was back when projects and collections shared `owned-card`. My question is should this be added back in?

alexbfree:
pasting here the reply i wrote in Slack (sorry hadn't seen you'd replied here too)

right but that was from when owned cards were used for projects too, right?
owned cards are solely used for collections/favorites now - I think removing the name made sense for projects, but not for collections
the line that says “if /collections then show the owner” is still there but not working
@props.skipOwner makes sense - but isn’t working as the variable name implies
owner is now ​*always*​ skipped
until this fix

alexbfree:
bottom line, YES it does make sense to add this back in. It's partly there anyway.

This changes:
screenshot 2016-06-08 20 07 30

To:
screenshot 2016-06-08 20 07 16

which I hope everyone would agree is an unequivocally good thing.

alexbfree:
One further thought. I think @amyrebecca added some user stuff into context for the Geordi work. It may actually be possible to get the display name without an explicit network request now. Also I think we do this successfully in Talk pages so we could look at what we are doing there.

srallen:
Here's the rest of the optimizations: #2083

I'm not saying that owner names shouldn't be there content-wise. To clarify, I think it should be checked if there's a more optimized way to get that information rather than making an API call each time, which is why it seems it was removed to begin with.

alexbfree:
OK fair enough. I actually think we have a bit of a general problem in PFE with user names not being always available - there are often times when you have the user login only and not the display name.. perhaps we need a general solution (not yet sure what that would be). I guess it does seem a bit crazy to make 20 odd separate name requests for each page

Another idea is panoptes API could offer a single endpoint which you give a list of user IDs and it returns all those users' details. 

Or maybe better still, the collections list that's returned could include the display names

alexbfree:
I've logged a suggested enhancement to the Panoptes API: https://github.com/zooniverse/Panoptes/issues/1824

alexbfree:
have made some attempt to fix in Panoptes, but need help from a backend person on getting the tests updated/PR complete. See https://github.com/zooniverse/Panoptes/pull/1825

alexbfree:
OK display_name is now available via the API (see https://github.com/zooniverse/Panoptes/pull/1825 https://github.com/zooniverse/Panoptes/pull/1824 )

I've updated this PR to simply display the display_name from the response.

Should be uncontroversial to merge now... @srallen over to you?

alexbfree:
I updated https://fix-missing-owner-names.pfe-preview.zooniverse.org/ for ease of testing.

alexbfree:
Thanks @srallen !

PR #2677, opened 2016-06-08
Option for flex on subject viewer

simensta:
This PR uses a workflow configuration to optionally add a `flex` property to the subject viewer. The `flex` property will only be added to the subject-viewer if the `workflow.configuration.image_layout` contains `'flex-option'`. Primarily motivated by my transcription projects that are requesting more pan/zoom-able space. 

Staged [here](https://option-for-flex-on-subject-image.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?workflow=2359) with an example workflow.

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -66,7 +66,7 @@ module.exports = React.createClass
       "subject-viewer--layout-#{@props.workflow?.configuration?.multi_image_layout}": @props.workflow?.configuration?.multi_image_layout
     })
     # Feature detect IE11 and apply flex prop. Revisit or remove when IE11 is no longer supported.
-    rootStyle = flex: "1 1 auto" if "ActiveXObject" in window or window.ActiveXObject isnt undefined or @props.workflow.configuration.image_layout and 'no-max-height' in @props.workflow.configuration.image_layout
+    rootStyle = flex: "1 1 auto" if "ActiveXObject" in window or window.ActiveXObject isnt undefined or @props.workflow.configuration.image_layout and 'flex-option' in @props.workflow.configuration.image_layout
     mainDisplay = ''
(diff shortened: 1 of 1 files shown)


brian-c:
- Mind making a couple variables and breaking that line up?

- I don't really understand what "flex-option" refers to, and calling it "flex" seems pretty implementation-specific. Can we come up with a name to better suit what the feature does?

- Also I don't entirely understand the goal here. I think it's to remove the size limitation of subject images, but they're still bound by `max-height: 90vh` rule so I don't think this changes much except on huge screens. Is the intention that images can scroll beyond the size of the viewport?

simensta:
- **Variables:** Good call on breaking up the line with variables. Will do.
- **Goal & Naming:** The naming can definitely be improved. I am trying to create an option to expand subject image space, ideally by absorbing unused real estate. Toggling the `max-height` property was one idea for creating a larger subject image. However, toggling that property doesn't affect any change (i.e., remove that property and the image is still height contained). I am having a hard time figuring out exactly what is containing the image height. Any ideas about what I am missing?

brian-c:
I think I see the issue here. Looks like some subject-sizing functionality got screwed up a bit at some point. I'm going to put a PR up to take care of that, and then this should become a lot more straightforward.

simensta:
Sounds good. Thanks!

brian-c:
See #2689. Once that's in you should be able to remove the `max-height` from the size-finding `` tag to get it to take up the right amount of space.

@trouille also brought up the idea of making the task area a bit smaller in this mode. That's set by the `.task area`'s `flex-basis`.

I think I'd do this with a class name on the Classifier so the necessary style overrides can be done in the CSS.

simensta:
Cool, I'll try removing max-height and adjusting the task area now that #2689 is merged. Closing this PR because I want to start fresh. Thanks!

PR #2679, opened 2016-06-09
add holding page while stats server is being fixed

CKrawczyk:
Adds the message "Project statistics are currently unavalible and we are working to fix the issue" as the stats page while the stats server is being fixed.


camallen:
Seems the stats index has changed and the UI is now throwing errors. we'll go with this holding page until we get this back online.

PR #2680, opened 2016-06-09
Add nib

brian-c:
Adds nib and fixes order in main.styl.

Not 100% sure why this is the right order.

Closes #2673.

### DIFF
--- a/css/main.styl
+++ b/css/main.styl
@@ -1,7 +1,8 @@
+@require "./common"
+@require "../node_modules/nib"
+
 @require "./zoo-font"
 @require "../node_modules/drag-reorderable/style.css"
-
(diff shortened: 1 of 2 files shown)


eatyourgreens:
:+1:

Out of interest, why don't we need to add
```
stylus: {
  use: [nib()]
    }
```
to the webpack config?

brian-c:
Most of Nib is implemented as stylus mixins, which we get with `@require "../node_modules/nib"`. It also does some image generation for gradients in old browsers, which it needs Node for, and that's what it gets from the 'use(nib())` call. Gonna try one more thing to get that working, might have another PR in a few minutes.

PR #2689, opened 2016-06-14
More appropriate subject image flex rules

brian-c:
This replaces the IE-specific `flex: 1 1 auto` on the SubjectViewer with `flex: 0 1 auto` for all browsers. That means the element will try to be as wide as it needs to display its content (`flex-basis: auto`, shrink if there's not enough room (`flex-shrink: 1`), and (this is the change) not grow any larger than it needs (`flex-grow: 0`), which means the black bars from #2329 won't be a problem. I also removed the potential for black bars in grid view.

I've put some a few various image sizes and proportions in the dev classifier mock data to demo what each of them looks like.

Also includes a button style fix that's been buggin' me.


simensta:
Looks good, thanks @brian-c!

PR #2694, opened 2016-06-15
workflow config for large image layout

simensta:
This PR creates a classifier layout that supports larger (i.e., not-height-contained) subject images and a smaller task area. This version of the classifier layout is engaged when `workflow.configuration.image_layout` contains `'no-max-height'`.

In addition, I made some adjustments to the scroll behavior for the classifier component. Instead of scrolling to the middle of the subject image, the page scrolls to just above the classifier component.

Check out the staged version with an example workflow [here](https://wrkflw-config-for-large-image-format.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?reload=0&workflow=2359).

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -94,7 +94,8 @@ Classifier = React.createClass
       @addAnnotationForTask classification, @props.workflow.first_task
 
   render: ->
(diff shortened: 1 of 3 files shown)


brian-c:
Looks great, thanks!

PR #2695, opened 2016-06-15
fix text task autoheight reset

mcbouslog:
fixes issue with back to back text tasks (including within combo) where if first text task resized, second text task inherits resized height instead of reseting to default height.

 - [staged branch](https://text-task-autoheight-fix.pfe-preview.zooniverse.org/)

 - [project on staging for testing](https://pfe-preview.zooniverse.org/projects/markb-panoptes/nfn-testing) - Text Workflow to help test

### DIFF
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -66,6 +66,11 @@ module.exports = React.createClass
     @setState initOffsetHeight: @refs.textInput.offsetHeight
     @updateHeight()
 
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.task.instruction isnt @props.task.instruction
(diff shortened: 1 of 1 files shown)


app/classifier/tasks/text/index.cjsx
```diff
@@ -66,6 +66,11 @@ module.exports = React.createClass
     @setState initOffsetHeight: @refs.textInput.offsetHeight
     @updateHeight()
 
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.task.instruction isnt @props.task.instruction
```
brian-c: I think this works, I'd just check that `nextProps.task isnt @props.task` since two tasks' instructions aren't guaranteed to be unique.

mcbouslog:
good call! updated accordingly. thanks!

brian-c:
Solid.

Issue #946, opened 2015-07-02
Would like to open links in a new window


mkosmala:
From what I've read (http://stackoverflow.com/questions/4425198/markdown-target-blank), Markdown doesn't directly support the opening of hyperlinks in a new window. However, the Markdown in Panoptes also doesn't support HTML, which would allow me to open in a new window. Can we have a solution that will allow for links to open in new windows, please?

vrooje:
I'd like this to be the default behavior (for external links) as well.

aliburchard:
+1 for this (especially as default behavior).

mschwamb:
related to #620

AdrianEvans:
I think this default behaviour idea should be implemented fairly quickly and is a much easier fix than changing the markdown scheme. Currently every-time somebody clicks a link they are taken away from the website. It's putting me off putting link onto pages to provide a wider research context or additional help. I don't want these things to be at the detriment to keeping people engaged with classification.

aliburchard:
We get this request from  a lot of project builders and volunteers. *We* know that you can open in a new tab by right clicking the link, but the vast majority of people don't know that, think that hard about it, or want to have to make that extra step. I think doing this by default would be really helpful.

chrislintott:
@edpaget @aweiksnar What say you?

edpaget:
I don't think there's a good way to make it defualt behaviour given the markdown library we use. We can add an extension to our version of markdown so that something like `!![aasdfasdf](https://google.com)` adds the `target="_blank"`.

aweiksnar:
@edpaget I'll look into making an extension to markdown-it, similiar to this https://github.com/leff/markdown-it-named-headers, it'd just add the target="_blank" to `a` tags

edpaget:
Didn't realize that would be possible. Do we actually want to do that? I'd rather it be an optional thing than have every single link in every markdown element do that.

aweiksnar:
Toggle-able on a per-link basis would be ideal. It'd have to be a modified syntax to give that level of control though. Maybe something like `[My Link](+tab+http://www.mylink.com)` could work

edpaget:
Yeah I was suggesting the double bang at the front. But am fine with whichever.

aweiksnar:
Ok cool, I'll see if I can put this together and we can figure out syntax later

aweiksnar:
[...]

PR #2697, opened 2016-06-16
Show an icon for external links

alexbfree:
We don't want to define "open in new tab" behaviour as it can and should depend on the user's preference or browser settings. We do want the behaviour of a link to be predictable. Therefore, a sensible way forward would be to ensure that any link that takes you off site is marked as an external link. Like this:

screenshot 2016-06-16 15 13 09

or this:
screenshot 2016-06-16 15 12 54

or this:
screenshot 2016-06-16 15 22 01

This is especially useful when a menu contains a list of external and in-page links, and also when you are in a middle of a task that would be destroyed/interrupted by clicking off page (as in the second example screenshot).

The code in this PR will automatically suffix any external link with an external link icon (which is already present in our font awesome pack).

External links are identified simply as those beginning `http` (best practice would suggest any link within the site should be relatively specified and not start with `http`).

For those occasions where we do not want this to appear on specific links, we can add the `suppress-external-link-icon` class to the `<a>`'s attributes and this will ensure no icon appears.

Deployed to https://external-links.pfe-preview.zooniverse.org/?env=production for testing.

I have opted for a CSS-only solution because I think this is the best way not to miss any. Especially as project builders or Talk commenters can add their own external links.

### DIFF
--- a/app/pages/about/team-page.cjsx
+++ b/app/pages/about/team-page.cjsx
@@ -417,7 +417,7 @@ module.exports = React.createClass
             <div key={teamMember} className="team-member">
               <img src={imageSrc} alt="#{details.name}" />
               <div className="team-member-details">
-                <h3>{details.name}, {details.title} {if details.twitter then <a href="http://twitter.com/#{details.twitter}" target="_blank"><i className="fa fa-twitter"></i></a> }</h3>
+                <h3>{details.name}, {details.title} {if details.twitter then <a href="http://twitter.com/#{details.twitter}" className="suppress-external-link-icon" target="_blank"><i className="fa fa-twitter"></i></a> }</h3>
                 <p>{details.bio}</p>
               </div>
             </div>
--- a/app/pages/profile/comment-link.cjsx
+++ b/app/pages/profile/comment-link.cjsx
@@ -57,7 +57,7 @@ module?.exports = React.createClass
         {if @state.board?.id and @state.discussion?.id
           <span>
             {' '}in{' '}
-            <a href={@state.href}>
+            <a className="suppress-external-link-icon" href={@state.href}>
               {if @state.project? and @state.owner
                 <span>
                   <strong className="comment-project" title="#{@state.owner.display_name}/#{@state.project.display_name}">{@state.project.display_name}</strong>
--- a/app/partials/main-footer.cjsx
+++ b/app/partials/main-footer.cjsx
@@ -112,9 +112,9 @@ module.exports = React.createClass
             <a href="http://blog.zooniverse.org/" target="_blank"><Translate content="footer.talk.blog" /></a>
           </div>
           <div className="site-map-section social-media">
-            <a href="https://www.facebook.com/therealzooniverse" target="_blank"><i className="fa fa-facebook"></i></a>
-            <a href="https://twitter.com/the_zooniverse" target="_blank"><i className="fa fa-twitter"></i></a>
-            <a href="https://plus.google.com/+ZooniverseOrgReal" target="_blank"><i className="fa fa-google-plus"></i></a>
+            <a href="https://www.facebook.com/therealzooniverse" className="suppress-external-link-icon" target="_blank"><i className="fa fa-facebook"></i></a>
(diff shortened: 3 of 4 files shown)


alexbfree:
copying @jwbmartin as we were just discussing this idea the other day

jwbmartin:
Hmmmmm..... I think that icon would be useful if the link actually opened a new window or tab to let the users know what to expect. But our links don't function in that way. They function like any link does and people know what to expect and know how to navigated the back button of a browser. I think this is solving a non-existent problem and adding unnecessary visual clutter.

srallen:
1) We are defining the behavior in some cases such as the links going out to twitter and facebook in the footer with `target="_blank"`. It's my understanding that most internet users understand that many sites open external links in a new tab where appropriate and can determine when this happens by context. 

2) With markdown, the default behavior is to open in the same tab. We can't change that, however, some markdown syntax was added to allow volunteers manually specify when they want the link to open in a new tab. See #946

3) As far as UX goes, I've seen it debated either way. Our volunteers, as far as I know, haven't communicated any confusion or frustration about what an external link is or where they go. So...

I'm inclined to agree that this is a non-issue and will defer to whatever @jwbmartin wants here unless there's an accessibility concern I'm not aware of that @eatyourgreens does know about.

eatyourgreens:
That icon usually indicates that a link opens in a new window or tab so I agree with @jwbmartin, I'm not sure it's useful here. Re. accessibility, I think VoiceOver will announce external links as such from the URL, and won't read the icon glyph anyway.

Fair point about leaving the site when classifying. That could be solved by preserving classifier state across sessions, like Shakespeare's World or Operation War Diary.

alexbfree:
So... if we could come up with a solution whereby wherever this icon appears, it also applies target=_blank to open in a new tab, then folks would be happier with that?

I think the second case (where you actually *can't* go back) is most critical. I guess we could solve that another way, like Jim says, but that's a much bigger change - this could be a good interim step.

>Our volunteers, as far as I know, haven't communicated any confusion or frustration about what an external link is or where they go. So...

Pretty much every person I've asked for feedback about my proposed maintain project context feature / new menu navigation has said it is confusing because some of the links take you away, and some don't. People need a way to differentiate the "leave project context" link from the "stay in project" links and I thought it would be better to establish a convention across the whole site and follow it, rather than do something special just for that one case.

I've also had feedback from a couple of people that the main project menu in Computer Vision: Serengeti is confusing because Snapshot Serengeti and MICO links take you to another page, while About / Classify etc do not.

Perhaps we should get more user feedback.

camallen:
Please don't make links open in a new window / tab..

alexbfree:
I can see this makes sense as a general principle, but surely in the second case you would agree opening in a new tab is beneficial @camallen ? Or at least marking it?

If you are in the middle of classifying an image on Computer Vision: Serengeti, and need to review the original Snapshot Serengeti subject page to help make your classification (this does happen a lot when an animal moves out of frame for example), you need to click that link. If there's no indication it will destroy your current tab AND no automatically opening it in a new tab, there is a very very high chance the user will never get to classify that image. They get part way through classifying and doing the thinking, maybe even drew some boxes... but they click that link and then - boom - the work is lost. We need to do something about this in line with "not wasting users' time".

And I think the same argument could be applied to any link that is on screen during classification, or message-writing, or talk-post-writing, or any other interruptible task. If the link will take you off-screen, and by default will not open in new tab but re-use the current one, how is the user supposed to know their partially-completed task is at risk?

camallen:
yeah - i really don't like sites / links that behave that way. It's my browser, please don't decide how I like links to open. Sure signal that it'll lose state or just fix that state loss (back button) issue but do not open in a new tab for me. 

pinging @marten for his thoughts.

srallen:
It sounds like what really the concern is is maintaining app state when a volunteer browses away and the confusing hierarchy of the navigation on our site when working on a project, which @jwbmartin is already working on a new design for. As far as the icon goes, I'm still of the opinion that it's @jwbmartin's call, too.

srallen:
The new navbar design has been merged and the point brought up about maintaining app state from a partial classification if a volunteer browses away is a separate issue. This could be solved by a listener on `beforeunload` using an alert or storing the classification in local storage.

PR #2702, opened 2016-06-16
Restore fix for IE11

### LINKED ISSUE #2701: IE not showing subject images anymore
Was there a change to how images are displayed? I've got two people now reporting the same thing , that in IE they don't see any subject images on Comet Hunter in the classification interface. The rest of the side loads correctly, but no subject image displays 

Talk discussion [here](https://www.zooniverse.org/projects/mschwamb/comet-hunters/talk/210/60615?comment=119430 )

brian-c:
Fixes #2701 again (sorry!).

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -66,6 +66,10 @@ module.exports = React.createClass
       "subject-viewer--layout-#{@props.workflow?.configuration?.multi_image_layout}": @props.workflow?.configuration?.multi_image_layout
     })
 
+    isIE = 'ActiveXObject' of window
+    if isIE
+      rootStyle = flex: '1 1 auto'
+
     mainDisplay = ''
     {type, format, src} = getSubjectLocation @props.subject, @state.frame
     if @state.inFlipbookMode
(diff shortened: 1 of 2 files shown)


srallen:
This is almost there, but the black borders have shown up again: 
screen shot 2016-06-16 at 1 36 50 pm

srallen:
Just to be clear, I'm seeing the black borders in Win7 IE 11.

srallen:
Trying to remember, I think that's what I added this line for: https://github.com/zooniverse/Panoptes-Front-End/pull/2330/files#diff-c0e272b86a2a14a6be7e2de7ea5e142aR38

But I'm totally open to a better way to deal with this.

brian-c:
I'm on Win10 IE11 and I get this:

screen shot 2016-06-16 at 3 43 16 pm

The top and bottom border are actually part of the image. The side borders you're seeing (maybe a screen size issue, though I can't reproduce by resizing) are likely from the added-to-IE `flex-grow: 1`, and they don't really affect anything, they're just not pretty. I don't think removing them is worth the trouble until we can rewrite this display.

Issue #2698, opened 2016-06-16
Have tutorial reset scroll bar when you move on to the next slide in the tutorial


mschwamb:
I'm noticing this on the new Notes from Nature - The images  and text are very long that I have to scroll on my 13-inch screen but when I advance to the next stage of the tutorial I'm still at the bottom of tutorial scroll bar and need to scroll all the way up start reading and scroll down. Because my OS/browser hides the scroll bar for the first time it wasn't clear I wasn't at the top of the tutorial window

This was in latest version of mac firefox.  screen capture below:

![tutorial_scroll](https://cloud.githubusercontent.com/assets/753325/16128350/fe216914-3432-11e6-9cf0-01143ecd4790.gif)

srallen:
Closed by #2703

PR #2703, opened 2016-06-17
Tutorial scroll fix 2698

simensta:
Addresses #2698. Also adds an overflow scroll fix for IE.

### DIFF
--- a/app/components/step-through.cjsx
+++ b/app/components/step-through.cjsx
@@ -1,5 +1,8 @@
 React = require 'react'
+ReactDOM = require 'react-dom'
 ReactSwipe = require 'react-swipe'
+animatedScrollTo = require 'animated-scrollto'
+
 
 module.exports = React.createClass
   displayName: 'StepThrough'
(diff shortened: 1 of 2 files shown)


srallen:
Could we consider adding the scroll to top behavior to the tutorial component itself instead of to `StepThrough`? Conceivably `StepThrough` could be used by a component elsewhere, even though it's not at the moment (I almost used it with the mini-course, for instance). I think the `handleNextClick` found on [L117-118](https://github.com/zooniverse/Panoptes-Front-End/blob/fab2dafe95a91b1e64dcb2a7dbfe2f373fc8b737/app/lib/tutorial.cjsx#L117-L118) would be a good spot to add this behavior.

Alternatively, if it is in `StepThrough`, the scroll to top should be agnostic of the component it is used in and not rely on a specific class name.

simensta:
When the scroll logic is handled by `handleNextClick` in the `Tutorial`, the scroll logic is only executed when Continue is pressed and not when the dots and arrows are pressed. In order to have consistent behavior across the `Tutorial`'s navigation elements, I kept the scroll logic in the `StepThrough` component. However, the scroll logic no longer relies on a specific class name. Thoughts?

srallen:
Could you rebase this? A few big changes have come in since, like react-router upgrade, etc.

simensta:
Good call -- will do.

PR #2705, opened 2016-06-20
React Router 2

brian-c:
- Updates react-router to version 2.4.x

- Replaces `PassHistoryContext` with `PassRouterContext` (I also spent an unreasonable amount of time trying the update the ModalForm lib to render popups in the same context by injecting them at the end of a Redux-style `ModalFormProvider` app-level wrapper component. It sorta works, but they don't unmount properly. I think I'm on the right track, so I might have another go at this in the future. See the render-in-context branch of the modal-form repo.)

- Removes `<base />` tag from HTML and defines `output.publicPath` in webpack build config (it was already in the dev config, not sure why that's different).

- Cleans up main.cjsx a bit

**Does not** replace instances of the History mixin (deprecated). I'll trudge through that job once this is merged.

Assigning a bunch of y'all, sorry, just trying to avoid missing anything.


parrish:
This seems to be working well for me.

I staged a deploy at https://router-upgrade.pfe-preview.zooniverse.org if anybody wants to check it out.

srallen:
This works for me too; checked FF, Chrome, Safari, and IE. Great job 💯

parrish:
Happy to hit merge on this unless anybody objects.

brian-c:
Just rebased, Mr. @parrish.

parrish:
👍

PR #2706, opened 2016-06-20
Home page redesign

parrish:
This adds a redesigned not-logged-in home page.

The current home-page component is rendered when the user is logged in.

https://home-page-redesign.pfe-preview.zooniverse.org/

TO-DO:
  - [x] Deploy social aggregator service
  - [x] Get a final seal of approval from @jwbmartin 
  - [x] ~~ Choose some actual featured collections ~~ (removed)
  - [x] Choose featured projects

Feedback welcome


srallen:
I don't think screen readers will know what to do with the SVG in the middle of Zooniverse. I think it'll read it as "zo" and "niverse." This what the [zooniverse-logotype](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/zooniverse-logotype.cjsx) SVG was made for and has a screen-readable aria label on it.

parrish:
Hmm, that didn't exist when I added that.

The only problem is that the logotype font doesn't match and makes line-height / font-size matching awkward (any suggestions there?).  The zooniverse-logo partial also has the aria-label of 'Zooniverse Logo'.

Also, and I don't know how to approach this delicately, but how concerned are we about ARIA?  I suppose it's possible to have non-visual projects, but we currently don't (that I know of).  Would changing it to `<h1 aria-label="The Zooniverse">THE ZO<ZooniverseLogo />NIVERSE</h1>` override the contained labels?  I don't think anybody on the team has access to a screen reader to even test it though.

srallen:
I hear you about the logotype SVG. I made it with the old design in mind, so it sounds like it won't work with the new one well. 

I'm not sure if adding `aria-label` to the `h1` would override the label in the logo SVG, however, I do know that OS X comes with VoiceOver which is helpful for basic testing. Cmd+F5 enables it.

mschwamb:
I think most people that come to the site will click on the first button. Learn more takes you to the 'About page'. I think you might want to have 'Learn more' be 'Get Involved Now' like it is in the current home page that takes you to the projects page.

Also the background video is pretty astronomy focused in case that matters.

app/pages/home-not-logged-in.cjsx
```diff
@@ -0,0 +1,66 @@
...
+        }
+
+        <h1>THE ZO<ZooniverseLogo />NIVERSE</h1>
+        <h1 className="lighter">is people powered research</h1>
+        <h1>
+          <Link to="/about" className="standard-button">Learn More</Link>
+        </h1>
+      </section>
+
+      <div className="home-link-down" onClick={@scrollDown}>
```
eatyourgreens: `div` tags can't receive keyboard focus. This should be a `button`.

app/pages/home-not-logged-in.cjsx
```diff
@@ -0,0 +1,66 @@
...
+        else
+          <video className="home-video" autoPlay loop>
+            <source type="video/ogg" src="./assets/home-video.ogv"></source>
+            <source type="video/webm" src="./assets/home-video.webm"></source>
+            <source type="video/mp4" src="./assets/home-video.mp4"></source>
+          </video>
+        }
+
+        <h1>THE ZO<ZooniverseLogo />NIVERSE</h1>
+        <h1 className="lighter">is people powered research</h1>
```
eatyourgreens: Should be `p` not `h1`.

app/pages/home-not-logged-in.cjsx
```diff
@@ -0,0 +1,66 @@
...
+        <h1>
```
eatyourgreens: Should be `p` not `h1`.

app/pages/home-not-logged-in.cjsx
```diff
@@ -0,0 +1,66 @@
...
+          <Link to="/about" className="standard-button">Learn More</Link>
+        </h1>
+      </section>
+
+      <div className="home-link-down" onClick={@scrollDown}>
+        <i className="fa fa-angle-down" onClick={@scrollDown} />
+      </div>
+
+      <section className="home-discover" ref="discover">
+        <h1>Discover, teach, and learn</h1>
```
eatyourgreens: `h2` not `h1`.

app/pages/home-not-logged-in.cjsx
```diff
@@ -0,0 +1,66 @@
...
+
+        <p>
+          The Zooniverse enables everyone to take part in real cutting edge
+          research in many fields across the sciences, humanities, and more.
+          The Zooniverse creates opportunities for you to unlock answers and
+          contribute to real discoveries.
+        </p>
+
+        <h1>
```
eatyourgreens: `p` not `h1`

eatyourgreens:
A page should have only one level-1 heading, but this has 5 `h1` tags.

Clickable elements should be buttons or links, not `div`. Otherwise, you have to go through all the extra work that Léonie Watson lists in this talk, to make it behave like a button or link: https://vimeo.com/167255877

User-agent sniffing is a pretty out-dated technique, though I can see why you're doing it here. Is there a better way of detecting whether the video should load eg. sniffing the network connection speed?

srallen:
I used a media queries with Chimp & See to show a background image up to a max-width of 900px and then the video any larger than that: https://github.com/zooniverse/chimpandsee/blob/cdfa405f8a072ac1b5b073805bd363efe48a9f3a/css/navigation.styl#L122-L146

I imagine that wouldn't cover 100% of mobile devices or tablets, but probably a lot of them. Not sure if this is the best approach, but it works.

parrish:
> A page should have only one level-1 heading, but this has 5 h1 tags.

It's worth reviewing [the spec](https://www.w3.org/TR/html5/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements) with respect to sections, but yeah, I need to cleanup the button `h1`s.

> Clickable elements should be buttons or links, not div.

In this case, the element is just form rather than function, so I don't care either way.

> User-agent sniffing is a pretty out-dated technique, though I can see why you're doing it here. Is there a better way of detecting whether the video should load eg. sniffing the network connection speed?

The only way to measure connection speed (that I can figure) is to download a large-ish file and time it with `onprogress` / `onload`.  I suppose you could replace the video with the image if it's loading too slowly, but it would take a few seconds to make that determination.  

> I used a media queries with Chimp & See to show a background image up to a max-width of 900px and then the video any larger than that

That's still going to pickup a lot of mobile devices, particularly in landscape.

Honestly the videos are pretty small(mp4: 1.9MB, webm: 5.2MB, ogv: 7.2MB) and play while loading, so I'm not sure how concerned to be about it.  Looks like google analytics [dropped support for connection speeds](https://analytics.googleblog.com/2011/03/fond-farewell-to-connection-speed.html), so who knows.

alexbfree:
Looking good! 

A couple of small observations:

- "people powered" should be "people-powered"
- "cutting edge" should be "cutting-edge"
- Zooniverse O logo is smaller/misaligned compared to rest of text in "THE ZOONIVERSE"
- Social feeds section looks a bit bland. Add social media icons/links to our feeds - and maybe some thumbnails from posts?
- Is there supposed to be a white band between vertical sections? Looks a bit odd to me.

srallen:
> That's still going to pickup a lot of mobile devices, particularly in landscape.

> Honestly the videos are pretty small(mp4: 1.9MB, webm: 5.2MB, ogv: 7.2MB) and play while loading, so I'm not sure how concerned to be about it. Looks like google analytics dropped support for connection speeds, so who knows.

Yeah, it's hard to target with media query since resolution and screen sizes change. More can be covered with more [precise media queries](https://css-tricks.com/snippets/css/media-queries-for-standard-devices/), but not sure if it's worth the effort, imo. FWIW, iOS and Android don't allow autoplay, so even if it does show up, it won't immediately start eating their bandwidth and they have to choose to play it.

parrish:
Updated for feedback.

> Social feeds section looks a bit bland. Add social media icons/links to our feeds - and maybe some thumbnails from posts?

Up to @jwbmartin.  I think she's taking a pass at that today?

> Is there supposed to be a white band between vertical sections? Looks a bit odd to me.

It's as designed.  I agree though.

eatyourgreens:
> > A page should have only one level-1 heading, but this has 5 h1 tags.

>It's worth reviewing [the spec](https://www.w3.org/TR/html5/sections.html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements) with respect to sections, but yeah, I need to cleanup the button `h1`s.

[Current advice to HTML authors](https://www.w3.org/WAI/tutorials/page-structure/headings/) is to ignore the section heading spec since [none of the browsers have implemented the outlining algorithm](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines).

The current home page structure is still kind of confusing, if you try to use headings to navigate the page. I'd suggest at least heading the subsections under The Zooniverse with `h2` rather than `h1`. Maybe markup bylines and calls to action as paragraphs, rather than headings, too.

- h1 THE ZO Zooniverse Logo NIVERSE
  - h2 is people powered research
  - h2 Learn More
- h1 Discover, teach, and learn
  - h2 Get started
- h1 Snapshot Supernova
  - h2 Join our team
- h1 test collect
  - h2 Check it out
- h1 Social Feed
  - h2 Twitter
  - h2 Facebook
  - h2 Blog
    - h3 (Missing heading)
      - h4 (Missing heading)
        - h5 (Missing heading)
          - h6 Projects
          - h6 About
          - h6 Talk

parrish:
> Current advice to HTML authors is to ignore the section heading spec since none of the browsers have implemented the outlining algorithm.

Yeah, there are a lot of contentious suggestions here.  Voiceover also doesn't seem to behave any differently when using alternate heading levels.  Frankly, I'm pretty nonplussed about this where it's at right now -- I mean, the page we're currently discussing this on uses multiple non-sectioned `h1`s.

If the team feels strongly about

- section
  - h1
  - h2
  - h2

vs

- section
  - h2
  - p
  - p

then I'm happy to just change it as I really don't care enough to argue it.

parrish:
Updated and redeployed at https://home-page-redesign.pfe-preview.zooniverse.org/.

Remaining to-do's:
  - Get @jwbmartin's seal of approval
  - Settle on the featured projects

jwbmartin:
Ok I just went back to this and I have a few nit-picky things and then I think it's good

- [ ] Move the down arrow that @parrish hates up a few pixels. It's a little close to the bottom margin
- [ ] Bump up the size of the type for "the Zooniverse" parts (sections 1 and 3) they are call to action and the look a little small
- [ ] The transition of the images of section 3 still looks odd. Can they just do a gentle ease in/out?
- [ ] Speed up the transition speed a little bit

Last question that I'm opening up to all: should the projects we choose for section 3 be hard-coded, pre-chosen images?

srallen:
Looking over this again and one minor CSS quibble: the font sizes should be em to be consistent within its own stylesheet and also because it's better for responsiveness.

parrish:
Rebased, tweaked, and tidied up.  Deployed at https://home-page-redesign.pfe-preview.zooniverse.org

> Last question that I'm opening up to all: should the projects we choose for section 3 be hard-coded, pre-chosen images?

I'm okay with this at least for now.  No reason we couldn't make it dynamic later on.

srallen:
I'm ok with it for now too. That can come later. I think this is ready to merge.

mcbouslog:
It's currently pulling the background images for section 3, featured projects, dynamically from [featured-collections.coffee](https://github.com/parrish/Panoptes-Front-End/blob/14c7102c087ce60eeb067685874d772627396985/app/lib/featured-collections.coffee).

I'm not sure what the easiest way is to test how those production projects' images look in section 3. Only thing I can think of is to pull the branch down locally, change the staging project ids to the production ids, then add `?env=production` to the url?

mcbouslog:
Or I can just hardwire the current (staging) images as suggested.

srallen:
@mcbouslog The conditional in `featured-collections` could be modified to also check the env in a query param in the url in addition to the `NODE_ENV`. I think ultimately what we want is to have a flag on projects that can be set by an admin to make it a home page featured project, but for now they're hardcoded ids we're fetching. I think that is ok for now.

parrish:
https://home-page-redesign.pfe-preview.zooniverse.org/?env=production gives an unexpected result.  Checking now.

parrish:
Ah.  I need to adjust the promoted project crossfading for when the back arrow is clicked.

parrish:
Okay, this should be fixed now.

mcbouslog:
The hardcoded project ids for production, while for the correct projects (Shakespeare's World, Planet Hunters, etc.) will feature different images than the images used in the preview projects.

When we are discussing hardcoding, are we discussing simply hardcoding the project ids, or uploading the images and hardcoding the actual HTML?

parrish:
> When we are discussing hardcoding, are we discussing simply hardcoding the project ids, or uploading the images and hardcoding the actual HTML?

Hardcoding, as in specifying project id and tagline in the config file.

Later on, we could set an attribute on the project record and pull the featured projects from the API.

mcbouslog:
Ok that sounds good Michael, and makes total sense, but my concern is if we want the specific images associated with the preview projects. Because the production projects' `project.image` are different than what we've been looking at on preview, and @jwbmartin is asking if we should hardcode the images from the preview projects.

parrish:
Ah.  I understand.  I'll add it.

mcbouslog:
Here are the images used for the preview projects featured (though Planet Hunters' shouldn't be used on production - very low quality):

shakespeare
planethunters
chimpandsee
galaxyzoo

mschwamb:
@jwbmartin I noticed the Planet Four snippet in the video is of the tutorial. Do you want a screen grab video of marking a real image rather than the tutorial. I've definitely got one if you want it I use in talks. Happy to send it your way if you want it.

mschwamb:
@parrish are you adding in the Daily Zooniverse blog to the social feed?  I noticed that the Zooniverse blog is updated very rarely (last update 2 months ago)  and that the Daily Zooniverse is updated more frequently (July 20th) .  Though maybe through the twitter and facebook feeds you get the daily zooniverse  links.

parrish:
Updated to use the specified images in [`featured-projects.coffee`](https://github.com/zooniverse/Panoptes-Front-End/pull/2706/commits/8751632fd0f685894fb9f797e1faf0a6ffeda9d3#diff-31bd98cedfd52365dfbbd52d67d9cf0cR2)

parrish:
@mschwamb The aggregator is over here: https://github.com/zooniverse/social -- ticked here zooniverse/social#1

I probably could mix the results from daily.zooniverse.org and blog.zooniverse.org.

mschwamb:
@parrish I think that would be great. It would make it look like the Zooniverse is more active at reaching out which it is.

astopy:
This looks great. Two minor suggestions:

- Some of the text is hard to read with images that have white in them. Maybe add a translucent background or something?
   
   screen shot 2016-07-26 at 10 46 44

- Do we need the "Social Feed" heading? The three "Twitter", "Facebook", and "Blog" headings are probably enough by themselves.

parrish:
@astopy There is a translucent layer on top, but it's pretty subtle right now.  I can tweak it a bit.

parrish:
Okay, the opacity of the promoted project overlay has been increased for contrast

I also fixed the behavior of auto-advancing after manually changing the promoted project -- the timer resets after switching manually now.

jwbmartin:
You guys, I think this looks good. Although this has nothing to do with the home page but can we please, please, please have the "Build a Project" link be styled the same way as the rest of the navigation links?

parrish:
Cool.  I'm pretty happy where this is at right now.  Any more concerns or are we ready for the big green button?

srallen:
I think it's ready. 💯

jwbmartin:
🙆🏻

parrish:
Minor issue: the promoted project animation is wonky on iOS and some swipe navigation would be nice.

astopy:
Hmm, so the animation isn't exactly smooth in Safari:

home-page-animation

parrish:
Barf, yeah.  Oddly enough, it only happens (for me) when deployed, but not locally.

I'm going to take a minute to retool that transition -- it's feeling pretty fragile.

srallen:
Noting here #2829 is an issue for the new redesign in IE11.

parrish:
Okay.  I think this is good to go again if anybody wants to check it out again.

srallen:
Just checked through IE11, Edge, Safari, FF, Chrome, and iOS Safari on iPhone 5. The animation on Safari still is abrupt, but I'm not sure if that's fixable. Looks good everywhere else. I think it's ready that if there are any more issues found, it can be iterated in subsequent PRs. 👍

astopy:
👍

rogerhutchings:
I got a weird flash of video before the regular home page the first time I loaded it as a logged in user, but not subsequently - it's probably too transient to lose sleep over.

Also, this is totally a content issue, but Facebook and blog post entries are basically identical. Is it even worth including FB stuff, given the relative infrequency of FB-only posts?

- [ ]

marten:
@rogerhutchings Nah, I get that too, even on refreshes. It also means logged in people who don't get to see the video still download the 2.5mb file, which isn't great.

jwbmartin:
So, we are ready? Not quite?

PR #2710, opened 2016-06-21
Scrolling up on the mouse wheel should zoom in

marten:
This makes it consistent with Google Maps and Photoshop (and probably most of the world).


simensta:
Looks good. Thank you @marten!

PR #2714, opened 2016-06-23
[do not merge] Fix collection pagination (geordi logging issue)

### LINKED ISSUE #2713: Collections next button doesn't work anymore (can't peruse or advance through a collection)
Getting the same issue in Firefox and Chrome if I try to hit the page 2 for a collection it doesn't advance to the new page. If I use the drop down menu I can move to page 2 and then the advance buttons work. 

Example: https://www.zooniverse.org/collections/mschwamb/requested-2016-hirise-targets
If you try and hit Next it doesn't work. 

@parrish is this something you can look at? This was fixed awhile back, not sure how it cropped up again

alexbfree:
Fixes #2713

Geordi call was breaking pagination as logClick was not in scope as willReceiveProps had not yet fired.
Changed to direct geordi.logEvent calls.
Also updated pagination logging to actually record the page number (without which the logs are fairly meaningless)
Also added logging of collection slug, without which the log has no meaningful context.

### DIFF
--- a/app/collections/show-list.cjsx
+++ b/app/collections/show-list.cjsx
@@ -108,7 +108,7 @@ module?.exports = React.createClass
             <div className="collection-subjects-list">{subjects.map(subjectNode)}</div>
 
             <Paginator
-              collection={@props.collection}
+              data={collection:@props.collection.slug}
               page={meta.page}
               onPageChange={@onPageChange}
               pageCount={meta.page_count} />
--- a/app/collections/show-list.cjsx
+++ b/app/collections/show-list.cjsx
@@ -108,6 +108,7 @@ module?.exports = React.createClass
             <div className="collection-subjects-list">{subjects.map(subjectNode)}</div>
 
             <Paginator
+              collection={@props.collection}
               page={meta.page}
               onPageChange={@onPageChange}
               pageCount={meta.page_count} />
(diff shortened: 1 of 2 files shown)


alexbfree:
can be tested at https://fix-collection-pagination-geordi.pfe-preview.zooniverse.org/

app/talk/lib/paginator.cjsx
```diff
@@ -32,15 +32,23 @@ module?.exports = React.createClass
...
 
-  componentWillReceiveProps: (nextProps, nextContext)->
-    @logClick = nextContext?.geordi?.makeHandler? 'change-page'
+  addAvailableData: (data) ->
+    if @props.collection?
+      data.collection = @props.collection.slug
+    data
+
+  componentDidMount: ->
+    @logClick = @context.geordi.makeHandler 'change-page'
```
eatyourgreens: I can't see this handler being called anywhere.

app/collections/show-list.cjsx
```diff
@@ -108,6 +108,7 @@ module?.exports = React.createClass
             <div className="collection-subjects-list">{subjects.map(subjectNode)}</div>
 
             <Paginator
+              collection={@props.collection}
```
eatyourgreens: I think this should be added to the `propTypes` and default props in the Paginator component too.

alexbfree:
marking Do Not Merge while we make some related changes.

eatyourgreens:
The original issue was fixed by #2716 so I'm closing this.

PR #2728, opened 2016-06-24
Fix the URL for the default avatar

### LINKED ISSUE #2727: Default avatar pointing to invalid asset path
image

From [here](https://github.com/zooniverse/Panoptes-Front-End/blob/cd4618544f7f470b7f84f834ed7c4f7ed6fea534/app/partials/avatar.cjsx#L3) looks like the lookup path is adding talk in there somwhere...

https://static.zooniverse.org/www.zooniverse.org/assets/simple-avatar.jpg
vs 
https://static.zooniverse.org/www.zooniverse.org/talk/assets/simple-avatar.jpg

eatyourgreens:
Fixes #2727

### DIFF
--- a/app/partials/avatar.cjsx
+++ b/app/partials/avatar.cjsx
@@ -1,6 +1,6 @@
 React = require 'react'
 
-DEFAULT_AVATAR = './assets/simple-avatar.jpg'
+DEFAULT_AVATAR = '/assets/simple-avatar.jpg'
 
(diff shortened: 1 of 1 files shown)


camallen:
LGTM if this fixes it.

eatyourgreens:
Checked on staging and it fixed the avatars so I'm going to merge.

PR #2730, opened 2016-06-24
Project pages demote site nav

brian-c:
This is a great big rewrite of the main app layout components (the header, main content area, and footer) to allow for @jwbmartin's new project navigation design where the site nav is tucked away behind the project.

ui_animation trim

Also added a super cool construction border when admin mode is active to discourage people from hanging out in admin mode all the time.

Also added eslint, and stuck to it for the most part (although I have a feeling it wasn't running on a few files).

https://project-demotes-site-nav.pfe-preview.zooniverse.org/

PR #2736, opened 2016-06-27
Restore new nav and fix modals

brian-c:
Reverts #2735 (un-reverts #2730) and fixes the modals.

https://restore-new-nav-and-fix-modals.pfe-preview.zooniverse.org/


parrish:
Looks good.

PR #2740, opened 2016-06-27
Remove updateUser function from app context

brian-c:
Can't find any place where this function is used, and it seems super dangerous to expose a way to get the UI's user out of sync with the session's user. Was this left over from something else @amyrebecca?

### DIFF
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -10,12 +10,10 @@ PanoptesApp = React.createClass
 
   childContextTypes:
     user: React.PropTypes.object
-    updateUser: React.PropTypes.func
     geordi: React.PropTypes.object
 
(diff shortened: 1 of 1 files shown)


brian-c:
Also removed App's `env` state, which was unused.

eatyourgreens:
@brian-c looks like those were being used by the Geordi logging
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/geordi-logger.coffee#L17-L21
but there must be better ways to store that data for Geordi.

brian-c:
Good find @eatyourgreens, thanks, I've updated the PR. That is indeed an unexpected use of component state.

srallen:
Superceded by #2759

PR #2732, opened 2016-06-24
Convert SubjectViewer dialogs to modal-form/dialog

brian-c:
This is pretty gross because `alert` from modal-form/dialog doesn't provide a way to close things (since it expects a form, which was a pretty fundamental design mistake). Gonna have a think about a better way to alert things.

Closes #2722.

PR #2741, opened 2016-06-27
Convert GenericTask help dialog to modal-form/dialog

simensta:
Converts the GenericTask's _Need some help dialog_ from '../../lib/alert' to 'modal-form/dialog'.

### DIFF
--- a/app/classifier/tasks/generic.cjsx
+++ b/app/classifier/tasks/generic.cjsx
@@ -1,5 +1,5 @@
 React = require 'react'
-alert = require '../../lib/alert'
+{alert} = require 'modal-form/dialog'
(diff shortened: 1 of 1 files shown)


brian-c:
Check out the help for the first task of http://localhost:3735/projects/jukoch/pattern-perception/classify?workflow=1580&env=production

`.modal-form img { max-width: 100%; }` helps, but I still get some weird scrollbars on the layout and it's not centered quite right. Might need some modification to the modal-form lib itself.

simensta:
Good catch. Since this is not as straightforward as I thought, I don't think this PR should hold up #2732. I'll look into a better fix in the modal-form lib.

PR #2743, opened 2016-06-27
Render while initial-loading

brian-c:
- Render while the initial load is happening. Show a spinner in place of the login/account area.

- Update layout-level components to use the app context's `user`.

### DIFF
--- a/app/components/pass-context.cjsx
+++ b/app/components/pass-context.cjsx
@@ -4,10 +4,12 @@ React = require 'react'
 module.exports = React.createClass
   contextTypes:
     router: routerShape
+    user: React.PropTypes.object
     geordi: React.PropTypes.object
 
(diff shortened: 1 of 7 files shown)


parrish:
When logged in, clicking on the username in the account bar:

> Uncaught Invariant Violation: Constructor.getChildContext(): key "user" is not defined in childContextTypes.
> [Uncaught TypeError: Cannot read property 'querySelector' of undefined](https://github.com/zooniverse/Panoptes-Front-End/blob/3401dc4dbe45b69e53a5414b1dc42e28fc77f57e/app/layout/account-bar.jsx#L72)

brian-c:
Not sure how I missed that, thanks.

parrish:
👍

PR #2745, opened 2016-06-27
Add project notifications to project nav

parrish:
This makes sense to me, but I'm definitely open to feedback.


mschwamb:
@parrish If the link is like the Zooniverse project menu link, since you can't tell if you have updates I'd only check it if I'm logged in and  on Talk see that it has an update since there it tells you on the sidebar if there is new stuff to read.

parrish:
Rebased and updated to show unread counts everywhere it's used.

parrish:
demo at https://project-notifications-link.pfe-preview.zooniverse.org

mschwamb:
@parrish  will this update with a number of unread notifications in the project bar? That's the only time I look at notifications. So I found it extremely useful to go through Talk and check only when it said there were was something to look at it. Is the link on the project navigation bar going to have that feature? 

Also if it will update with the count, then can you also remove the separate link on the Talk navigation bar or are you happy with the redundant link? It's currently still there. 

If I was new navigating to notifications for the first time checking out the site,  I would have no idea how  and when you get notifications. Now that it is separated from being just a Talk link should there be one line that says these are notifications from Talk for the discussions you've followed and discussions you're involved in/mentioned? Just like at the bottom of Talk there's a line about what Talk is.

mschwamb:
@parrish completely misread your comment - I'm blaming the jetlag. Sorry for not reading carefully enough before replying. Thanks for updating the count everywhere. So my questions/feedback are really:

Also if it will update with the count, then can you also remove the separate link on the Talk navigation bar or are you happy with the redundant link? It's currently still there.

If I was new navigating to notifications for the first time checking out the site, I would have no idea how and when you get notifications. Being a link from Talk it made more sense it was associated with Talk. Now that it is separated from being just a Talk link should there be one line that says these are notifications from Talk for the discussions you've followed and discussions you're involved in/mentioned? Just like at the bottom of Talk there's a line about what Talk is.

parrish:
> Can you also remove the separate link on the Talk navigation bar or are you happy with the redundant link? It's currently still there.

I could go either way.  I'll pull it for now though.

> If I was new navigating to notifications for the first time checking out the site, I would have no idea how and when you get notifications

Fair point.  I'll add a one-liner.

parrish:
Updated at https://project-notifications-link.pfe-preview.zooniverse.org

srallen:
I asked @jwbmartin to do a quick look over and she's would prefer the notifications link to use an icon like [this](https://thenounproject.com/search/?q=notification&i=492155), otherwise this LGTM.

mschwamb:
I don't think anyone is going to know what that is. The Zooniverse isn't an app so I wouldn't expect and know what notifications are and know looking at a  bell icon what it was. It's the same issue currently with logging out. My science team collaborations don't know to hit the Zooniverse icon. 

I'm not sure I'd click on the icon, and it seems odd to add one icon in the middle of all text links as all the external links a project owner adds currently come after all the default ones the project builder creates.

srallen:
I disagree that no one will know what the icon is. Notifications are a part of web apps including the desktop versions that you see on Facebook or Twitter. However, how it fits in here and what that looks like could use a design/ux iteration and that would include considerations for discoverability. The icon in the middle of the project nav is a good point. I won't hold up merging this for the ux piece, but we should come back to it.

jwbmartin:
I was thinking it would be located next to the mail icon and function in the same way. Having Notifications spelled out is excessive and it definitely shouldn't be in our already bloated top bar navigation.

mschwamb:
@srallen valid point about the notification icon. 

@jwbmartin Notifications happen frequently if the volunteer is active in a project talk and are at a project level  (there's a separate set of notifications on the zooniverse  level for zooniverse talk).   You get a notification for each project Talk thread that you're following, conversations you're mentioned in and in that project page your notifications link will be updated with the number of new notifications (and currently talk but this PR is removing the link from project talks to just be on the project navigation bar) . If you hide the notifications button away at the top Zooniverse  then someone has to always actively click the to get the zooniverse navigation bar to see if they have project notifications as most people log in to classify and then later move to talk so it means an extra click to two clicks to see if one has notifications. It's been super convenient and helpful to be able to classify and then see if I had notifications and then just click on the link if I did.  

Honestly having a notification icon, mailbox and login/logout button in the top right of the project level would be extremely helpful, and save the now multiple clicks that need to be done to go from logging in to classifying. see #2787 I'm still not sure why the account buttons had to be hidden in the zooniverse level and can't be like what github does and have them in the top right with the notifications.

PR #2747, opened 2016-06-27
Customizable filtering of subject-viewer metadata.  Closes #1281.

### LINKED ISSUE #1281: showing specific metadata on Talk but not in the classification interface?
Not sure if this is the right repo for this question. Currently there are two choices for your metadata, you can hide metadata from being shown in the interface with a symbol in front of a column, but does that mean then that data doesn't show on Talk. Volunteers for Planet Four: Terrains would like to know what full frame CTX image the data came from. We could add the link in the metadata but we wouldn't want that show in the interface. Only in Talk . Does that functionality currently exist or could that be added in the future?

parrish:
Currently, subject metadata keys prefixed with `#` are hidden from the user.

This expands that to a customizable list and adds `!` as a prefix that is hidden unless the subject viewer is displayed in talk.

If we come up with more conditions later on we can just expand the list

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -43,6 +43,7 @@ module.exports = React.createClass
     frameWrapper: null
     allowFlipbook: true
     allowSeparateFrames: false
+    metadataFilters: ['#', '!']
 
   getInitialState: ->
     loading: true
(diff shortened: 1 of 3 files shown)


srallen:
Sorry this sat open for a while, so this probably needs a rebase, but otherwise LGTM.

parrish:
Updated

Issue #2733, opened 2016-06-27
Completely not obvious how to sign in anymore to any zooniverse panoptes projects because the sign in/register and sign out buttons are hidden and extra clicks to navigate after logging in


mschwamb:
Now the entire top navigation bar is a zoonvierse symbol. It's not obvious to new people how to sign up or for current volunteers how to sign in. it look me a second to know to click the zooniverse icon on the right which then zooms out to the view with the top bar. 

I don't think signin/sign out register and the project builder link shouldn't be hidden from view. we want to encourage people to sign in and access their profile. it wasn't obvious what to do. The Zooniverse icon doesn't look like a menu, it looks like it would take you to Zooniverse home. 

<img width="1152" alt="screen shot 2016-06-27 at 5 57 06 pm" src="https://cloud.githubusercontent.com/assets/753325/16375967/abf1c5b0-3c90-11e6-9c0d-03ab2cc6bac4.png">

then if you click the zooniverse symbol 

<img width="1151" alt="screen shot 2016-06-27 at 6 00 08 pm" src="https://cloud.githubusercontent.com/assets/753325/16376050/1539fbbe-3c91-11e6-9d86-e3f52c84b59f.png">

Also this means for people who don't have email notifications set for getting private messages they only see on login/logout  whether or not they have a new message. 

Also if you do log in, you have to click twice, to go anywhere on the project page, because the first click gets rid of the blue outline with the navigation bar which is annoying and not expected. Same thing on logout. You get logged out and the blue outline remains. 

cc @chrislintott @mrniaboc

alexbfree:
i noticed that a "The Zooniverse" label was proposed, which might help a little bit. https://github.com/zooniverse/Panoptes-Front-End/pull/2730#issuecomment-228387736

The comment that sign in needs to be more visible makes sense though.

marten:
Seconding this issue. I brought it up in Slack when this was first posted as GIFs there. We should keep the "sign in / register" or "User Name (NOTIFICATIONS ICON)" section in the bar next to the zoo logo at all times.

marten:
Also, nobody will expect the rest of the menu to be behind that logo, logos pretty much always take you to the homepage. The logo should probably be a hamburger instead.

mschwamb:
That would be a great solution @marten  and would address my concerns for not getting as many logged in classifications which is what I really care about.  

[...]

PR #2748, opened 2016-06-27
Add account/sign-in to demoted nav floaty thing

brian-c:
@jwbmartin ain't gonna like it, but this is a stopgap for #2733.

screen shot 2016-06-27 at 5 32 38 pm

https://always-show-sign-in.pfe-preview.zooniverse.org/


jwbmartin:
NO. I DON'T LIKE IT.

If I come up with a better solution can we do that instead, please?

Issue #2742, opened 2016-06-27
Geordi logging issues


camallen:
https://www.zooniverse.org/projects/birgus2/western-shield-camera-watch/talk/372/62590?comment=122563

2 fold issue (see above link for more details):
+ different envs should be contacting the correct servers
+ down geordi server seems to be raising an error (this should be caught)

camallen:
@alexbfree looks like geordi client uses 2 params to configure the Geordi API url, first is server initialize param [here](https://github.com/zooniverse/geordi-client/blob/ef8c7a0ce4b53db9fdc0b5249981ea17f9ed9ddd/index.js#L48) that sets the env and then that is used to look up the the URL [here](https://github.com/zooniverse/geordi-client/blob/ef8c7a0ce4b53db9fdc0b5249981ea17f9ed9ddd/index.js#L88). 

Any reason not to just pass in the env param to setup the server URL instead of providing a URL that may fail a lookup if it doesn't have the correct key?

alexbfree:
the linked PR #2749 fixes the issue with the wrong geordi server being used.

Error handling for Geordi down coming shortly.

alexbfree:
The PR #2750 will ensure that Geordi fails gracefully. It is the responsibility of the user of the Geordi client's returned promise to decide what to do when Geordi throws an error, so have put handling into PFE. We have something similar in Serengeti's code that calls Geordi.

eatyourgreens:
Hi, Geordi client already defines an `onerror` handler for the XHR request, and rejected promises don't interrupt JavaScript, so I'm a little confused here. 

The issue in the linked Talk thread is being caused by the broken links mentioned there, not Geordi.

camallen:
The issue in talk is 2 fold:
1.  URL's were broken, that is fixed.
2. Geordi was reporting errors when the server was unavailable (recreate it by modifying the geordi host in /etc/hosts to something unreachable)
  + `vendor-2e5aed4….js:288 Uncaught (in promise) Error: Network Error(…)`

Seems geordi is raising on network error in the client [here](https://github.com/zooniverse/geordi-client/blob/ef8c7a0ce4b53db9fdc0b5249981ea17f9ed9ddd/index.js#L99) on these events...Basically I don't think PFE should be raising geordi errors when the server is down..

eatyourgreens:
OK. For reference, graceful failure on errors is already built into the Geordi client, added here
[...]

PR #2749, opened 2016-06-28
Correctly get shell environment

alexbfree:
Fixes part of #2742 . `@state.env` wasn't being set correctly in PFE when `NODE_ENV` is set at the shell.

### DIFF
--- a/app/lib/geordi-logger.coffee
+++ b/app/lib/geordi-logger.coffee
@@ -13,9 +13,15 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
   instance: =>
     @geordi = @geordi || @makeGeordi @keys?.projectToken
 
+  getEnv: ->
+    shell_env = process.env.NODE_ENV if process.env.NODE_ENV=="production"
+    reg = /\W?env=(\w+)/
+    browser_env = window?.location?.search?.match(reg)?[1]
+    shell_env || browser_env || 'staging'
+    
   makeGeordi: (projectSlug) ->
     new GeordiClient
-      env: @state?.env
+      env: @getEnv()
       projectToken: projectSlug || @keys?.projectToken
       zooUserIDGetter: () => @state.user?.id
       subjectGetter: () => @keys?.subjectID
--- a/app/lib/geordi-logger.coffee
+++ b/app/lib/geordi-logger.coffee
@@ -1,7 +1,7 @@
 GeordiClient = require 'zooniverse-geordi-client'
 
 class GeordiLogger # Make calls to the Geordi API to log user activity
-  constructor: (@state, @geordi) ->
+  constructor: (@user, @geordi) ->
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
 
@@ -23,7 +23,7 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
     new GeordiClient
       env: @getEnv()
       projectToken: projectSlug || @keys?.projectToken
-      zooUserIDGetter: () => @state.user?.id
+      zooUserIDGetter: () => @user?.id
       subjectGetter: () => @keys?.subjectID
 
   makeHandler: (defType) -> # Once defined, efficiently logs different data to same event type
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -36,7 +36,8 @@ PanoptesApp = React.createClass
     auth.stopListening 'change', @handleAuthChange
 
   componentWillUpdate: (nextProps, nextState) ->
-    @geordiLogger = @geordiLogger || new GeordiLogger nextState
+    if !(@geordiLogger? && @geordiLogger.user == nextState.user)
+      @geordiLogger = new GeordiLogger nextState.user
 
   handleAuthChange: ->
     auth.checkCurrent().then (user) =>
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -20,15 +20,9 @@ PanoptesApp = React.createClass
     updateUser: @updateUser
     geordi: @geordiLogger
 
-  getEnv: ->
-    reg = /\W?env=(\w+)/
-    browser_env = window?.location?.search?.match(reg)
-    @state?.env || browser_env?[1] || 'staging'
-
   getInitialState: ->
     initialLoadComplete: false
     user: null
-    env: @getEnv()
 
   updateUser: (user) ->
     @setState user: user


alexbfree:
confirmed via testing locally with NODE_ENV=production that this fixes the issue with Geordi logging to wrong server.

alexbfree:
updated the branch to also ensure @state.env is set correctly so you get the right value even if you don't use `getEnv`

app/partials/app.cjsx
```diff
@@ -19,9 +19,12 @@ PanoptesApp = React.createClass
     geordi: @geordiLogger
 
   getEnv: ->
+    shell_env = process.env.NODE_ENV
     reg = /\W?env=(\w+)/
     browser_env = window?.location?.search?.match(reg)
-    @state?.env || browser_env?[1] || 'staging'
+    env = shell_env || browser_env?[1] || 'staging'
+    @setState env: env
```
camallen: why is this using setState? the result of @getEnv() in `getInitialState` will set this anyway.

app/partials/app.cjsx
```diff
@@ -19,9 +19,12 @@ PanoptesApp = React.createClass
     geordi: @geordiLogger
 
   getEnv: ->
+    shell_env = process.env.NODE_ENV
     reg = /\W?env=(\w+)/
     browser_env = window?.location?.search?.match(reg)
-    @state?.env || browser_env?[1] || 'staging'
+    env = shell_env || browser_env?[1] || 'staging'
```
camallen: minor but I think this could just be one line, `env = process.env.NODE_ENV || browser_env?[1] || 'staging'`

I think the intention is clear enough without the need for the extra variable but i'm about a 3/10 on this.

app/partials/app.cjsx
marten: Apart from that, I wouldn't expect a function called `getSomething` to change state.

app/partials/app.cjsx
marten: I do kind of like the `shell_env` but then I'd suggest moving the `?[1]` away from this line, so that you just get `shell_env || browser_env || 'staging'`. Kind of weird if `shell_env` contains a string, but `browser_env` is actually a regex match result.

marten: I mean, if we're going to nitpick... I don't really care too strongly either.

app/partials/app.cjsx
alexbfree: funnily enough, what you describe is what I had originally, but then I was worried it wasn't robust enough. I guess since shell_env cannot change after launch, and as you say getInitialState will call this method anyway, then we will be fine. have updated.

app/partials/app.cjsx
alexbfree: have gone with @marten's suggestion, i think it looks tidier now

brian-c:
Can I recommend figuring out the environment in the file where it's used? It's not a state of the app component. The component state object itself probably shouldn't be passed outside the component that owns it.

alexbfree:
@brian-c to be clear I have not changed anything about where the `@state.env` variable lives. I simply made sure it contains the correct value (as it is currently, the first of the three OR-ed terms is always null and so the code doesn't function as intended).

brian-c:
I'm saying as long as we're messing around with this chunk, let's move it to where it belongs. I almost deleted the env state yesterday because I didn't this it was being used because I wasn't expecting the entire state object to be passed into another module.

alexbfree:
OK @brian-c . I haven't really looked at where it came from or who else is using it. Where do you think it should live? On the face of it, having application-wide variables (set via shell or browser param) accessible from the state of the top level component doesn't seem like a crazy idea.... where would be another good place to put it? seems like anywhere that would qualify as "somewhere central" would be fine, but you know the code better than me.

brian-c:
Looks like it's only used [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/geordi-logger.coffee#L18), so that's where I'd define it.

[Totally unrelated, but now that I'm looking at this more closely: does the logger's user ever update? It looks like the logger is created once, on the App's `componentDidUpdate` (why there?) with the first `nextState` it gets.]

alexbfree:
@brian-c  "Your unrelated comment" sounds like a concerning bug. Please can you raise a ticket and assign Will and myself to it. Thanks

alexbfree:
@brian-c Knowing whether you are on production or development is not anything specific to Geordi, conceptually... It seems to me that there might be many cases in future where a piece of code needs to know whether the app is running on production or not, even if there are no such usages right at this moment.... and so we should put it somewhere central in anticipation of that..... Object oriented design/separation of concerns would suggest that the environment is a property of the system, not of the geordi client integration....

eatyourgreens:
@alexbfree `state.env` is a bit of a hack that was added as part of the Geordi integration here
https://github.com/zooniverse/Panoptes-Front-End/pull/2590/files#diff-111bc95aed61206568877df047f1c6e8
I agree with @brian-c , this should be handled by the Geordi client or the logger class. Everything else in PFE abstracts the environment away into the API Client config.

alexbfree:
I really disagree with the idea of putting this system-level information in just one tiny part of the code, but i guess it is a philosophical design choice with no objective answer. In the interests of progressing this I will do as you suggest.

alexbfree:
OK @brian-c or @eatyourgreens it should be good to merge now

app/lib/geordi-logger.coffee
```diff
@@ -13,9 +13,15 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
   instance: =>
     @geordi = @geordi || @makeGeordi @keys?.projectToken
 
+  getEnv: ->
+    shell_env = process.env.NODE_ENV if process.env.NODE_ENV=="production"
+    reg = /\W?env=(\w+)/
+    browser_env = window?.location?.search?.match(reg)?[1]
+    shell_env || browser_env || 'staging'
```
eatyourgreens: You can change this to `browser_env || shell_env || 'staging'` if you want URL query params to always override NODE_ENV.

eatyourgreens:
Looks like the App component's state is still being passed to the logger, just to set the user ID here https://github.com/alexbfree/Panoptes-Front-End/blob/2d02f241d7e75e359ba1bc2d934f756ebcc97d0f/app/lib/geordi-logger.coffee#L26

Is there a way to pass just the current user ID to the Geordi client?

alexbfree:
Surely @state.user is out of scope for this PR? I did notice that but we are just talking about `env` here aren't we?

alexbfree:
(I say that because this ticket was supposed to quickly fix a problem with geordi logging to the wrong server, and all sorts of other stuff is getting dragged into it. I would like to get this PR merged and then handle other improvements separately. I agree there is more to improve with Geordi integration).

alexbfree:
never mind, it'll only delay this further if we discuss it. I'll move out `@state.user`

alexbfree:
i'll address Brian's issue about user not getting updated too.

app/lib/geordi-logger.coffee
alexbfree: I know. But I don't. I want shell to take priority if shell env=production, otherwise browser params take priority. If shell env not set to production, and browser env not set, use staging.

alexbfree:
OK @brian-c and @eatyourgreens - How about now? LGTM?

app/lib/geordi-logger.coffee
brian-c: panoptes-client uses browser setting before shell setting, since browser setting can be changed on the fly.

eatyourgreens:
Looks ok to me. There are situations where the Geordi `env` will be different from the apiClient `env`:
https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/config.js#L49-L52
I'm happy to merge if that isn't a big deal. Also happy to leave updating the Geordi user ID to a separate issue.

@brian-c with this merged, you can finish cleaning up the App state in #2740 right?

alexbfree:
@eatyourgreens Thanks. Yes I am ok with that different between geordi and PJC - so if you could go ahead and merge that'd be great.

alexbfree:
Thanks @eatyourgreens !

PR #2750, opened 2016-06-28
Ensure Geordi network errors do not get thrown in PFE

alexbfree:
Fixes part of #2742 .

Catch any errors from the geordi logging promise, and warn to console (but not more frequently than every 5 mins)

### DIFF
--- a/app/lib/geordi-logger.coffee
+++ b/app/lib/geordi-logger.coffee
@@ -4,11 +4,20 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
   constructor: (@state, @geordi) ->
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
+  @GEORDI_WARNINGS_TIMEOUT_INTERVAL = 5 * 60 * 1000 # only warn about Geordi down once every 5 mins, to avoid console spam
 
   keys:
     projectToken: 'zooHome'
 
   geordi: null
+
+  suppressWarnings: false
+
+  suppressWarningsForAWhile: ->
+    @suppressWarnings = true
+    setTimeout =>
+      @suppressWarnings = false
+    , GeordiLogger.GEORDI_WARNINGS_TIMEOUT_INTERVAL
 
   instance: =>
     @geordi = @geordi || @makeGeordi @keys?.projectToken
@@ -31,7 +40,10 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
     newEntry = Object.assign {}, logEntry, @keys
     if GeordiLogger.tokens.indexOf(newEntry.projectToken) > -1
       @instance().logEvent newEntry
-      console.warn 'No Geordi logger available for event ', JSON.stringify(logEntry) unless @instance().logEvent
+      .catch (err) =>
+        if !@suppressWarnings
+          console.warn "Warning: Error was encountered logging to Geordi server [#{@state?.env} - #{@instance().GEORDI_SERVER_URL[@state?.env]}]:",err
+          @suppressWarningsForAWhile()
 
   remember: (eventData) ->
     reset = eventData?.projectToken? && (eventData?.projectToken != @instance().projectToken)


alexbfree:
To test this, you can go into your `node_modules/zooniverse-geordi-client/index.js` and modify the staging URL to something broken.

app/lib/geordi-logger.coffee
```diff
@@ -4,12 +4,21 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
   constructor: (@state, @geordi) ->
 
   @tokens = ['zooHome', 'zooTalk', 'zooniverse/gravity-spy']
+  @GEORDI_WARNINGS_TIMEOUT_INTERVAL = 5 * 60 * 1000 # only warn about Geordi down once every 5 mins, to avoid console spam
```
camallen: no need for comments here the param is named well enough. Not sure about the caps though, is that consistent with the rest of the repo? Looks like @tokens doesn't use this convention.

app/lib/geordi-logger.coffee
```diff
@@ -31,7 +40,10 @@ class GeordiLogger # Make calls to the Geordi API to log user activity
     newEntry = Object.assign {}, logEntry, @keys
     if GeordiLogger.tokens.indexOf(newEntry.projectToken) > -1
       @instance().logEvent newEntry
-      console.warn 'No Geordi logger available for event ', JSON.stringify(logEntry) unless @instance().logEvent
+      .catch (err) =>
+        if !@suppressWarnings
+          console.warn "Warning: Error was encountered logging to Geordi server [#{@state?.env} - #{@instance().GEORDI_SERVER_URL[@state?.env]}]:",err
```
marten: Where does `@state` come from? This isn't a React component afaict. Forgot to change after extracting to here maybe?

marten: I suggest something like `[#{@instance().env} - #{@instance().url}]` there.

marten: Nvm, I found the state. Not used to those skinny constructors in Coffeescript. I'd still like it if this didn't do as many grabs into other objects.

marten:
LGTM

app/lib/geordi-logger.coffee
alexbfree: Re the caps: I am using the convention of capitals for non-changing, constant values (see http://www.w3schools.com/js/js_conventions.asp) and mixed case when there is no guarantee the data won't change. (`tokens` could legitimately be changed at runtime).

camallen: Yeah - i figured it was signalling something constant. Just seems strange that there is no difference between the declarations on lines 6-7 but the naming convention changes. I don't really mind the caps but it should be consistent with the rest of the repo.

app/lib/geordi-logger.coffee
alexbfree: Yeah it is passed into the constructor. I know what you mean, but technically it's not reaching into another object though, only the logger itself. Anyway that's not really part of this PR, @state was already there before this PR, I just use it.

app/lib/geordi-logger.coffee
alexbfree: i think the comment is helpful and harmless to leave it in.

brian-c: Caps are fine here. Normally constants are defined as variables, not as class properties.

eatyourgreens:
I'm not entirely sure what this PR does, except make Geordi errors harder to debug by silently swallowing the error messages.

alexbfree:
@eatyourgreens try modifying the address of geordi to a non-existent web address in your node_modules/zooniverse-geordi-client/index.js , then try using the app, you should see the problem this aims to fix. See #2742 reported issue.

eatyourgreens:
Ah right, the main problem I can see is that you are silencing all errors from the Geordi client, rather than just silencing network errors when the server is down. That might make life difficult if you need to debug `422 - Unprocessable entity` or `500 - Internal server error` responses from the Geordi server due to a mistake in the logging code.

Those errors would still show in the network panel, but a developer might not know to look there if there aren't any errors showing in the console.

alexbfree:
OK. So I guess we have two options 

1) Abandon this PR, allow Errors to show in console if geordi has a network error (which means we would say regarding the original reported issue "working as designed/will not fix")
2) Come up with a mechanism so that only network errors are silenced, not other errors, or so that errors are only silenced on production, not dev & staging.

Sounds like you prefer option 1. I am OK with that. @camallen are you OK with that (since you reported the issue)?

camallen:
yep go for it...my console aesthetics shall endure.

alexbfree:
ok. closing.

PR #2755, opened 2016-06-29
Admin page: check that a user exists

brian-c:
The admin page was expecting to have a user before one was determined.

### DIFF
--- a/app/pages/admin.cjsx
+++ b/app/pages/admin.cjsx
@@ -47,17 +47,17 @@ module.exports = React.createClass
 
   render: ->
     <div>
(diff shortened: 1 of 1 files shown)


camallen:
LGTM. that was much easier than i was expecting...out of interest how are those props getting defined and passed into the component? Via react-router?

brian-c:
That specific prop (user) is passed into the page (which is determined by the router) here:

https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/app.cjsx#L57

camallen:
that makes a lot of sense now. thanks for pointing me in the right direction.

PR #2756, opened 2016-06-29
update dropdown match only at start when filtering

mcbouslog:
update dropdown task to match only at start of option label when filtering, not any character in option label.

typing “N” in a list of states will filter to Nevada, Nebraska, New Hampshire..., instead of the current setup where it would search the entire label and filter to Arizo**n**a, Arka**n**sas, Califor**n**ia...

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -159,6 +159,7 @@ module?.exports = React.createClass
               allowCreate={selects[i].allowCreate}
               noResultsText={if not options?.length then null}
               addLabelText="Press enter for {label}..."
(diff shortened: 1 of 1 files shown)


simensta:
Hey @mcbouslog, can you rebase this branch with master?

simensta:
Possible downside: When searching for an item that can be referred to by a number or a word, such as a month, you can end up with no result depending on what you type first:
screenshot of can't find may

mcbouslog:
Yep, known downside. I mentioned that to the NfN team and they were fine with it, month was the only common example we could think of. I think it's an important downside though, so if there's negative feedback we'll certainly consider adding whether to search at start or anywhere as toggle-able option.

PR #2759, opened 2016-06-30
Ensure geordiLogger is set before mount

srallen:
Possible bug fix for reported issue here: https://www.zooniverse.org/projects/adrianevans/fossil-finder/talk/49/63272?page=1

This instantiates the Geordi client in `componentWillMount` and updates it on auth change.

Also removed the extraneous `updateUser`

### DIFF
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -24,9 +24,9 @@ PanoptesApp = React.createClass
     initialLoadComplete: false
     user: null
 
-  updateUser: (user) ->
-    @setState user: user
-
+  componentWillMount: ->
+    @setupGeordi(@state.user)
+  
   componentDidMount: ->
     auth.listen 'change', @handleAuthChange
     generateSessionID()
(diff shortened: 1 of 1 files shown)


app/partials/app.cjsx
```diff
@@ -24,9 +24,9 @@ PanoptesApp = React.createClass
     initialLoadComplete: false
     user: null
 
-  updateUser: (user) ->
-    @setState user: user
-
```
brian-c: `updateUser` is referenced on lines 14 and 20, too, mind removing those? Then we can close #2740 with this.

srallen: Removed

brian-c:
You think it'd be worth removing the `?` from `context.geordi?` (and removing `!!this.context.geordi && ...` checks) in the app so there's no more confusion as to whether or not geordi exists? I'm happy to tack a commit onto this PR if that's alright.

srallen:
Sounds good. Thanks.

PR #2761, opened 2016-06-30
update dropdown search match only at start

mcbouslog:
update dropdown task to match only at start of option label when filtering, not any character in option label.

typing “N” in a list of states will filter to Nevada, Nebraska, New Hampshire..., instead of the current setup where it would search the entire label and filter to Arizona, Arkansas, California...

replaces pull request #2756


simensta:
Looks good -- thanks @mcbouslog!

PR #2764, opened 2016-07-04
Update Geordi user on login/logout

eatyourgreens:
Update the stored user ID for Geordi logging without overwriting any existing stored data, such as the project token or subject ID.


alexbfree:
That looks much safer, looks good.

PR #2766, opened 2016-07-05
Fail if dependencies aren't installed

brian-c:
This'll just prevent us from serving/deploying with bad dependencies, like it does with bad node or npm versions.


eatyourgreens:
Looks good. I installed an outdated `react-router` and `npm start` gives me
```
react-router: installed: 1.0.3, expected: ~2.4.1
Invoke npm install to install missing packages
```
👍

PR #2770, opened 2016-07-07
New Grid Tool

wgranger:
Staged branch (with test project) at:
https://weather-cells-tool.pfe-preview.zooniverse.org/projects/wgranger-test/old-weather-grid-testing/classify

This tool is used to expedite the tedious process of marking cells in a table. Users begin by marking cells, then rows, and finally save a grid to use for future/similar annotations.

### DIFF
--- a/app/classifier/drawing-tools/grid.cjsx
+++ b/app/classifier/drawing-tools/grid.cjsx
@@ -13,6 +13,9 @@ module.exports = React.createClass
   displayName: 'GridTool'
   saveState: true
 
+  componentWillMount: ->
+    @findSchema()
+
   statics:
     initCoords: null
 
@@ -54,7 +57,6 @@ module.exports = React.createClass
   initCoords: null
 
   render: ->
-    @findSchema()
     {x, y, width, height} = @props.mark
 
     points = [
@@ -66,8 +68,16 @@ module.exports = React.createClass
     ].join '\n'
 
     <DrawingToolRoot tool={this}>
-      <polyline points={points} onClick={@destroyTool.bind null, this} />
+      {if @state?.template
+        @renderCells(points)
+      else
+        <polyline points={points} onClick={@destroyTool.bind null, this} />}
     </DrawingToolRoot>
+
+  renderCells: (points) ->
+    for cell in @state.cells
+      
+      <polyline points={points} onClick={@destroyTool.bind null, this} />
 
   destroyTool: ->
     if window.confirm 'Do you want to delete this cell?'
@@ -75,4 +85,5 @@ module.exports = React.createClass
 
   findSchema: ->
     @props.user.get('project_preferences', {project_id: @props.workflow.links.project}).then ([pref]) =>
-      if pref.preferences.grid.inProgress
+      if pref.preferences.cells.length
+        @setState template: pref.preferences.cells
--- a/app/classifier/drawing-tools/grid.cjsx
+++ b/app/classifier/drawing-tools/grid.cjsx
@@ -4,9 +4,8 @@ DragHandle = require './drag-handle'
 Draggable = require '../../lib/draggable'
 deleteIfOutOfBounds = require './delete-if-out-of-bounds'
 DeleteButton = require './delete-button'
-Rectangle = require './rectangle'
 
-MINIMUM_SIZE = 5
+MINIMUM_SIZE = 10
 DELETE_BUTTON_DISTANCE = 9 / 10
 
 module.exports = React.createClass
@@ -23,7 +22,6 @@ module.exports = React.createClass
       y: y
       width: 0
       height: 0
-      type: 'cell'
 
     initStart: ({x, y}, mark) ->
       @initCoords = {x, y}
@@ -49,15 +47,13 @@ module.exports = React.createClass
     initRelease: (cursor, mark, e) ->
       _inProgress: false
 
-    # This must be rewritten. It is likely an annotation will be destroyed when it's a row as the cursor doesn't move much to the left or right
-    # initValid: (mark) ->
-    #   mark.width > MINIMUM_SIZE and mark.height > MINIMUM_SIZE
+    initValid: (mark) ->
+      mark.height > MINIMUM_SIZE
 
     saveState: (mark, template) ->
       for cell in template
         cell.y = mark.y
         cell.height = mark.height
-        cell.type = 'row'
       template
 
   initCoords: null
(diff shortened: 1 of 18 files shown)


mcbouslog:
nice work Will, this is quite the drawing tool, lots going on!

it's consistent across Chrome, Firefox, Safari, IE11 and MSEdge, but here are a few general notes:  
1. console warning on all, `vendor-d34e0f7….js:7 Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `ChangeListener`. See https://fb.me/react-warning-keys for more information.`
2. can add template with name that already exists, creating duplicate template names
3. after drawing headers and rows, but then deciding headers were messy/bad and click clear button for  headers, should rows also clear? or to put differently, will ever be scenario where wouldn't also want rows to clear if headers cleared?
4. this one's minor, but after applying template, want to add few rows, `create new grid` button isn't obviously how I'd edit/add currently applied template? could just be me as idiot though.
5. and we talked about this on Slack Will, but there's some strangeness if have applied a template, or created headers and rows and clicked done, and then attempt (though not sure a volunteer ever would/should) to add a new template with headers and rows, but I don't think this is a priority to address by any means.

I hope to dive into code more tomorrow, and will keep posted.

wgranger:
1. I've noticed this warning appears on other branches too. I don't think it's a cause of these changes because the error points to the `ChangeListener` component. Look out for it on other branches you work on, with a drawing tool, and let me know.
2. Hmm, yeah, I thought about this initially while working through the PR. Maybe I can add this change. 
3. This is likely a bug and an easy fix. I think the clear button should only clear what it is associated with.
4. I'll chew on this for a bit. Might be a workflow/help change.
5. This might be similar to the second point in showing an error to a user when they shouldn't be allowed to do something. That way, they aren't confused when it looks like a grid is about to be placed but isn't.

brian-c:
Code is lookin' good, I'd say it's probably good to merge when @mcbouslog's points are addressed. I think y'all talked about putting this through some kinda UI design process next week (pinging @jwbmartin).

I've prototyped (very very roughly, and also I'm not really familiar with any of the requirements, ha) kinda how I'd expect a grid tool to behave: https://grid-tool-poc.pfe-preview.zooniverse.org/dev/classifier So maybe that's useful, or maybe not at all

wgranger:
2. This seems to be working with the most recent commit.
3. Currently, the clear button only deletes items that are cells or rows, respectively. This seems logical.
4. For now, I will make this change in the help section of the workflow.
5. I talked to Amy about this a bit, and we decided it seems to be fine as it currently works. A second grid will simply not be placed, similarly to how a rectangle will not be placed if it doesn't have a great enough width or height.

@brian-c Wow, that tool you made looks pretty interesting and I'd be curious to see the code for it. This grid tool was created based off of the angular code that was previously responsible for marking cells and rows for the old weather project.

brian-c:
> @brian-c Wow, that tool you made looks pretty interesting and I'd be curious to see the code for it.

https://github.com/zooniverse/Panoptes-Front-End/blob/grid-tool-poc/app/classifier/drawing-tools/grid-poc.jsx

The prototype is just a rectangle with some junk stacked on top. If we go this direction, it should probably be its own standalone thing.

> This grid tool was created based off of the angular code that was previously responsible for marking cells and rows for the old weather project.

Ah, I wasn't aware of that! So my thing might actually be way off what's actually needed. Let's set some time aside next week to figure this thing out. Map Room maybe?

wgranger:
@brian-c I'd be fine to meet at the map room one day next week to chat about this. We can get in touch on Slack about it. Do you see anything outstanding in this PR that would prevent a merge? It would be nice to have the tool available for the newest Oldweather project that will be launched soon.

brian-c:
@wgranger Mind adding this tool to the `"draw"` task of the dev-classifier/mock-data workflow (you might need to disable the save feature if it can't find project prefs [or go ahead and mock some project prefs if you're up for it]) and adding an icon in classifer/tasks/drawing/icons?

app/classifier/drawing-tools/grid.cjsx
```diff
@@ -0,0 +1,150 @@
...
+          <g>
+            <DeleteButton tool={this} x={@props.mark.x} y={@props.mark.y} />
+          </g>}
+    </DrawingToolRoot>
+
+  renderTemplate: ->
+    for cell in @state.template
+      points = @cellPoints cell if @state.activeTemplate is 'grid'
+      points = @rowPoints cell if @state.activeTemplate is 'row'
+      <Draggable key={Math.random()} onDrag={@handleTemplateDrag} onEnd={deleteIfOutOfBounds.bind null, this} disabled={@props.disabled}>
```
brian-c: Setting `key={Math.random()}` means the element gets destroyed and rebuilt from scratch every render. Maybe set `cell._key ?= Math.random()` and use that.

app/classifier/tasks/drawing/grid-buttons.cjsx
```diff
@@ -0,0 +1,188 @@
...
+
+  removeMarks: (type) ->
+    marks = @props.annotation.value.filter((mark) =>
+      mark if mark._type != type
+    )
+    @props.annotation.value = marks
+    @props.preferences.update 'preferences'
+
+  deleteGrid: ->
+    @props.user.get('project_preferences', {project_id: @props.workflow.links.project}).then ([pref]) =>
```
wgranger: When using the dev classifier, nothing is returned from `[pref]` here, perhaps because the classifier doesn't have a valid `project_id`?

brian-c: Isn't `@props.preferences` available here?

wgranger: They are, but I think I decided to make an API call here because `@props.preferences` wasn't being updated correctly when I tried to alter `@props.preferences` directly. I can try to switch it back and see how it goes. I remember this was a frustrating part with dealing with the preferences and for some reason the grid wasn't behaving.

wgranger: @brian-c Ok, in revisiting this API call I noticed that the grid template wouldn't be persisted across classifications unless I made an API call. Updating preferences through `@props` would look fine unless the page was reloaded and all the props were reconstructed from scratch. Making an API call here seemed to persist both ways. However, perhaps I'm missing something here.

brian-c: That's very odd. Is `pref is @props.preferences` true? It should be, but it sounds like it's not.

Also, possibly related: I just noticed you're accessing `@props.workflow.links` here, too. In general, if you can avoid accessing a resource's `links` directly you should. That's an implementation detail of our JSON-API serializer and there's no guarantee it won't change in the future. `project_id: @props.project.id` should work, right?

wgranger: Looks like `@props.project` isn't currently passed to drawing tools, so I'll just have to add that in. Oddly enough, `pref is @props.preferences` returns true.

wgranger:
Think I got everything covered with this most recent commit. Let me know if something still looks wonky.

brian-c:
Looks good to me, thanks.

PR #2771, opened 2016-07-07
Mine the comment box with honeypots

parrish:
We got our first spambot on Talk!

This sets a few honeypots for bots to get tripped up on.  This form is used for commenting, new discussions, and messages, so we should be pretty well covered.

Closes zooniverse/Talk-Api#92

### DIFF
--- a/app/talk/comment-box.cjsx
+++ b/app/talk/comment-box.cjsx
@@ -53,6 +53,8 @@ module?.exports = React.createClass
       @setState {reply: nextProps.reply}
 
   onSubmitComment: (e) ->
+    return @onFakeSubmit(e) if @isRobot()
+
     if @props.logSubmit is true
       @logPageClick 'add-comment', @props.submit
     e.preventDefault()
@@ -67,6 +69,15 @@ module?.exports = React.createClass
         @setFeedback @props.submitFeedback
       .catch (e) =>
         @setState(error: e.message, loading: false)
+
+  isRobot: ->
+    @refs?.honeyCheck?.checked or @refs?.honeyText?.value?.length > 0
+
+  onFakeSubmit: (e) ->
+    e.preventDefault()
+    @hideChildren()
+    @setState subject: null, content: '', error: '', loading: false
+    @setFeedback @props.submitFeedback
 
   onInputChange: (e) ->
     @setState content: e.target.value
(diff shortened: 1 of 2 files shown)


app/talk/comment-box.cjsx
```diff
@@ -118,7 +129,12 @@ module?.exports = React.createClass
               {if @state.showing is 'image-selector' then <span>&nbsp;<i className="fa fa-close" /></span>}
             </button>
 
-          <SingleSubmitButton type="submit" onClick={@onSubmitComment} className='talk-comment-submit-button'>{@props.submit}</SingleSubmitButton>
+            <input type="checkbox" className="honey-input" tabindex="-1" name="agree-to-terms" value="agree-to-terms" ref="honeyCheck"/>
```
brian-c: All these `tabindex`s should be `tabIndex` with a capital I.

app/talk/comment-box.cjsx
```diff
@@ -118,7 +129,12 @@ module?.exports = React.createClass
...
+            <input type="text" className="honey-input" tabindex="-1" name="better-title" ref="honeyText" />
+            <button type="submit" className="honey-input" tabindex="-1" onClick={@onFakeSubmit}>Submit</button>
```
brian-c: Multiple `type="submit"` buttons are ambiguous, and currently a fake one swallows the enter key.

Does the submit button they clicked actually reveal anything about the humanness of the user anyway?

parrish: Enter just gives you a newline in the comment box rather than submitting the form.

Spambots usually have a rule to click either the first or last visible submit button.  This just buries it in the middle.

brian-c: I don't think we should depend on the cursor always being in a textarea.

parrish: Maybe I'm being dense, but other than the submit button, what could have focus where hitting enter could submit the form?

brian-c: Currently nothing, but I can imagine a link or something could be added in the future, at which point users'll lose comments and we won't know why. I understand the intention here, but it has the potential to silently break the form and I don't think the risk is worth it.

parrish: Not sure why we'd add a link to the form, but even if that had focus, it'd still work -- it's not going to submit the form anyway.

Would adding a message like "Posting was not successful" on fake submit be a good compromise?

parrish:
I'm closing this with the intent to revisit it if it continues to be an issue -- currently it looks like a one-off.

PR #2773, opened 2016-07-08
Move pan/zoom into its own component

eatyourgreens:
Moves the pan/zoom controls into their own component, rewritten in ES6.

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -31,6 +31,7 @@ module.exports = React.createClass
     frame = @props.frame
     {type, format, src} = getSubjectLocation @props.subject, @props.frame
     FrameWrapper = @props.frameWrapper
+    zoomEnabled = @props.project? && 'pan and zoom' in @props.project?.experimental_tools
     frameDisplay = switch type
       when 'image'
         <div className="subject-image-frame" >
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Still needs work, I think, because of the close coupling between the pan/zoom controls and the SVG image in the `FrameAnnotator` component.

brian-c:
Not sure if it's helpful, but I hacked out a little zoom-anything component a while back: https://github.com/brian-c/react-zoomable/blob/master/index.js

It's not SVG-specific, but I assume it'll work fine with SVGs. There's an example.html file in the repo.

I wouldn't recommend using it directly, but maybe as some inspiration for the decoupling?

eatyourgreens:
@brian-c thanks! I'll take a look.

I'm happy now that I think I've got all the pan zoom code moved into its own component, so this is ready for review when someone has time to look at it.

eatyourgreens:
Staged at https://pan-zoom-component.pfe-preview.zooniverse.org/

simensta:
This is great -- thank you @eatyourgreens! 

One can no longer zoom by pressing enter after focusing the "+"/"-" interface button. Is that intentional? One can still zoom by pressing the keyboard "+"/"-" button.

eatyourgreens:
@simensta should be fixed now. A misplaced `)` meant the callback wasn't being invoked after a `keyDown` event.

simensta:
Looks good!

PR #2774, opened 2016-07-08
Initially reveal the site header on the project page when not signed in

### LINKED ISSUE #1792: get a white screen if you go to a project url that is private and you're not logged in
Had this happen a few times to collaborators who are new to using the Zooniverse platform. So they are typically not logged in and then go to check a change we made to the project we're building and get a bit confused what to do. Perhaps a message saying this is a private project and you must log in to view it. 
screen shot 2015-10-19 at 7 18 00 am

### LINKED ISSUE #2733: Completely not obvious how to sign in anymore to any zooniverse panoptes projects because the sign in/register and sign out buttons are hidden and extra clicks to navigate after logging in

brian-c:
If there's no user when the project page mounts, the site header is revealed.

👍'd by @jwbmartin. Closes #2748, closes #2733.

I've also tweaked the project-not-found message, so this closes #1792.

### DIFF
--- a/app/layout/index.jsx
+++ b/app/layout/index.jsx
@@ -18,6 +18,7 @@ const AppLayout = React.createClass({
 
   childContextTypes: {
     setAppHeaderVariant: React.PropTypes.func,
+    revealSiteHeader: React.PropTypes.func,
   },
 
   getInitialState() {
@@ -36,6 +37,13 @@ const AppLayout = React.createClass({
           siteHeaderDemoted: variant === 'demoted',
         });
       },
+
+      revealSiteHeader: () => {
+        this.matchWindowScale();
+        this.setState({
+          siteHeaderRevealed: true,
+        });
+      }
     };
   },
 
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -43,6 +43,7 @@ ProjectPage = React.createClass
     project: null
     owner: null
     preferences: null
+    loading: false
 
   getInitialState: ->
     background: null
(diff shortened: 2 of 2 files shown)


parrish:
Logging in causes the header to hide and reveal again.

Otherwise looks good.

mschwamb:
Thanks Brian. 

Could you still add something next to the zoo icon that shows you that's how you get back to that top menu bar. Even if the bar is exposed on log-in how to get back to it's not very clear say if someone wants to log out. In Galaxy Zoo the zooniverse icon is a tab so it has some hint that it pulls that login menu down.

brian-c:
@parrish Fixed the site header jumpiness. I'm thinking we want the header to automatically close after sign-in, right?

@jwbmartin please see @mschwamb's comment. @mschwamb we'll probably comtinue to iterate on this.

parrish:
Looks good.  Happy to continue iterating, but I think closing the header also loses the sign-in feedback.

PR #2778, opened 2016-07-11
Stack video play controls above other elements

eatyourgreens:
Creates a new stacking context for video player controls with `position: relative`.
Should ensure that the slider control for scrubbing a video can be clicked with a mouse.


eatyourgreens:
Fixes a niggling little bug where the video scrubber control can't be dragged with a mouse.

PR #2783, opened 2016-07-12
Allow resizing of the point drawing tool

### LINKED ISSUE #2157: Point marking task: smaller mark size for fine marking
Solar Stormwatch would like to use a marking task with a smaller mark size, for marking the fine lines on their jmap images. Could be implemented by adding the point size as an option to the drawing task eg. select from small, medium or large points. Default to large for backward compatibility with existing projects.

eatyourgreens:
Adds an option to the point tool to allow the size of the points to be specified via a dropdown menu in the project builder.

Closes #2157

### DIFF
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -64,7 +64,7 @@ module.exports = React.createClass
                 mark: mark
                 details: toolDescription.details
                 color: toolDescription.color
-                size: toolDescription.size ? 'large'
+                size: toolDescription.size
 
               toolMethods =
                 onChange: @handleChange.bind this, i
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -64,6 +64,7 @@ module.exports = React.createClass
                 mark: mark
                 details: toolDescription.details
                 color: toolDescription.color
+                size: toolDescription.size ? 'large'
 
               toolMethods =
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Staged at https://point-tool-resize.pfe-preview.zooniverse.org/

srallen:
We may want to consider moving the delete button to a different position when the point is small. The delete button takes a significant area of the point itself. However, functionally, this is working cross-browser. 👍

brian-c:
I agree about moving the delete button. Otherwise, looks good.

app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -64,6 +64,7 @@ module.exports = React.createClass
                 mark: mark
                 details: toolDescription.details
                 color: toolDescription.color
+                size: toolDescription.size ? 'large'
```
brian-c: Can we move the default into the tool's `getDefaultProps`?

eatyourgreens: Good point. I've moved it into the point tool.

eatyourgreens:
Feedback from Chris Scott:
>Right. The points are a good size now but the ‘x’ icon of the selected point makes it tricky to adjust once placed (although easy to delete!).

Definitely need to move the delete button.

eatyourgreens:
OK, I've moved the delete button out to the same radius as large points, and bumped the selected size a little too. Changes should be deployed to https://point-tool-resize.pfe-preview.zooniverse.org/

srallen:
LGTM

PR #2788, opened 2016-07-13
Allow the Zoo Logo title to be overridden

eatyourgreens:
Set the title to 'Show navigation' for the nav toggle button.

### DIFF
--- a/app/layout/site-nav.jsx
+++ b/app/layout/site-nav.jsx
@@ -209,7 +209,7 @@ const SiteNav = React.createClass({
   },
 
   render() {
-    const logo = <ZooniverseLogo width="1.8em" height="1.8em" style={{ verticalAlign: '-0.5em' }} />;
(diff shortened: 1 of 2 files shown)


eatyourgreens:
That's right. When you use an image as the content for a control or a link, the image alt text is applied as the control's label. So the advice for images in links and buttons is to use alt text which describes the purpose of the link or the button, rather than the image itself.

brian-c:
Huh, I did not know that! Still, I'd make that change only when the logo toggles the site header.

srallen:
Yeah, the logo const is being used both in the link and button, but I agree the title should only apply to the button case.

eatyourgreens:
My bad, I'll fix the link.

At the moment, the home link text is picked up on zooniverse.org as "Zooniverse logo". "Zooniverse home" might be clearer labelling but, yeah, it shouldn't be "Show navigation."

eatyourgreens:
Should be fixed now. I get this as the list of form controls on a project page
image
and this as the list of links on the home page
image

PR #2793, opened 2016-07-14
removeEventListener on unmount

srallen:
Noticed this while looking over #2788. This should probably be `removeEventListener` on component unmount.


brian-c:
Whoops, right-on. Thanks.

PR #2795, opened 2016-07-15
add stats text-content styling

mcbouslog:
add stats page `stats-text-content` class/styling similar to previous stats page styling.

my apologies for not catching in recent related pull request review, but in recent commits the `project-text-content` class was reduced. I've added a `stats-text-content` class that extends the current `project-text-content` class and adds the same css as before.

sorry for the white space removal from my text editor, adding `?w=1` (or `&w=1` if already a param) to the url makes it easier to review on GitHub.


CKrawczyk:
LGTM

PR #2803, opened 2016-07-18
Add Invert Image Tool

wgranger:
The invert tool allows users to invert the colors of an image/subject in order to more easily notice objects. This branch was tested and should be working on all browsers (IE, Safari, Firefox, Chrome). I tried to abstract the invert into a modification property to allow the `ModifiedImage` component to be more flexible with other filters in the future.

Staged branch at https://invert-tool.pfe-preview.zooniverse.org/projects/wgranger-test/invert-testing/classify

### DIFF
--- a/app/components/modified-image.cjsx
+++ b/app/components/modified-image.cjsx
@@ -1,0 +1,31 @@
+React = require 'react'
+
+STYLE = {
+    filter: "url('#svg-invert-filter')"
+}
+
+MARKUP =
+  '<svg style=\"position: fixed; right: 100%; top: 100%; visibility: hidden;\">
+    <defs>
+      <filter id=\"svg-invert-filter\">
+        <feColorMatrix in=\"SourceGraphic\" type=\"matrix\" values=\"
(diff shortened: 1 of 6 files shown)


app/components/modified-image.cjsx
```diff
@@ -0,0 +1,34 @@
...
+        </feComponentTransfer>
+      </filter>
+    </defs>
+  </svg>'
+
+module.exports = React.createClass
+  displayName: 'ModifiedImage'
+
+  componentWillMount: ->
+    @setElements()
```
brian-c: You should do this once, outside the component lifecycle. As written, it'll add a new `#svg-invert-filter` every time one of these mounts.

wgranger:
@brian-c Outside the component _lifecycle_ makes sense, but also keep that HTML code in the `ModifiedImage` component? Seems like it would be more concise that way.

wgranger:
Staged branch has been updated.

brian-c:
Looks good to me! Thanks.

PR #2804, opened 2016-07-18
Support Collections, Favorites & User Profiles Within Project Context

### LINKED ISSUE #2253: Ensuring Preservation of Project Context

alexbfree:
Fixes #2253 and supersedes #2670.

This is a new improved approach which aims to keep things as simple as possible while still allowing the different types of content to be seen/navigated in context.

## The problem(s) being solved:

- No way to view the collections/favorites specific to a project 
- Very easy to lose the fact you are "in" a project and drop back to Zoo-wide context, simply by clicking on a collection, user profile etc. This most often happens in Project Talk.

This PR can be tested on staging at https://maintain-project-context-v2.pfe-preview.zooniverse.org/ (add `?env=production` to use on live projects)
For example if you want to jump straight in now and try it, go to: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/alexbfree/computer-vision-serengeti/collections?env=production
or 
https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/alexbfree/computer-vision-serengeti/users/Whoandwhatitis

# New Features 

This PR introduces:

## 1. A new "Collect" menu within Projects

screenshot 2016-07-18 17 38 07

This new "Collect" menu (analogous to the Zoo-wide one) allows the user to browser all collections for a project, or all favourites for a project. They can also check out the collections and favourites they have made themselves within this project.

## 2. Ability to view a user within the context of a project

screenshot 2016-07-18 17 40 27

This ensures that when you click on a username in Project Talk, you remain in the context of that project.
Comments are filtered to only show those for this project, and now you can view a user's collections/favourites for this project too, without having to go to another page. As with the main user profile page, you can view your stats, or if viewing another user, message them.

The design intent is that if you are interested in a user - you'll see his collections here on his profile - if you are interested in a project you'll see all collections or your own via the Collect page.

## 3. Ability to view favorites collections without selecting a user first

screenshot 2016-07-18 17 51 25
screenshot 2016-07-18 17 51 30

This is available both in a project context, and Zoo-wide

## 4. Links that let you explicitly step from a project to the wider Zooniverse

screenshot 2016-07-18 17 42 08
screenshot 2016-07-18 17 43 48
screenshot 2016-07-18 17 43 43

The general approach here is:

- If you click "Show more" (only available when focussed on an individual's collections/favourites), it will show all collections/favourites _within this project_ 
- If you click "Show all username's Zooniverse collections" (only available when focussed on an individual's collections/favourites) it will take you to the general Zooniverse-wide context, but keep you focussed on the same user.
- If you click "Show all Zooniverse collections" it will take you to the general Zooniverse-wide context.

## 5. Menu navigation when viewing a collection

screenshot 2016-06-07 15 23 55

As described in #2212, it's easy to hit a navigational dead end when viewing a collection. By adding a couple of contextual menu options, we are now able to ensure there is always a way back to that user's collections, or to that user's profile - keeping the project context if there is one.

## 6. Ensure that no links can accidentally take you out of a project

Now, to leave a project, you need to explicitly click "Show other Zooniverse collections" or "Show user's Zooniverse collections" or go to the main Talk link in the main Zooniverse navigation. Clicking a user profile or collection within Project Talk, will never take you out of the project.

----

Note: all screenshots above are from *within* the project context, but the equivalent non-project menus for collections, favourites and user profiles have also been updated. For example:

screenshot 2016-07-18 17 47 20

screenshot 2016-07-18 17 47 33

----

Please try out this feature on staging, for example at 
https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/alexbfree/computer-vision-serengeti/collections/?env=production


alexbfree:
Just realized I missed something; standby on review. Will update shortly.

alexbfree:
OK. Fixed the missing parts, rebased and resquashed. Ready for review/merge and also ready to test on https://maintain-project-context-v2.pfe-preview.zooniverse.org/

app/collections/preview.cjsx
```diff
@@ -28,17 +28,18 @@ module.exports = React.createClass
...
-    [owner, name] = @props.collection.slug.split('/')
-
+    baseLink = "/"
+    if @props.project?
+      baseLink += "projects/#{@props.project.slug}/"
     <div className="collection-preview">
       <div className="collection">
         <p className="title">
-          <Link to="/collections/#{owner}/#{name}">
+          <Link to="#{baseLink}collections/#{collection.slug}">
```
parrish: Currently a 404.  Should be `"#{baseLink}/`

app/collections/preview.cjsx
```diff
@@ -28,17 +28,18 @@ module.exports = React.createClass
...
             {@props.collection.display_name}
           </Link>
           {' '}by{' '}
           {if @state.owner
-            <Link className="user-profile-link" to="/users/#{@state.owner.login}">
+            <Link className="user-profile-link" to="#{baseLink}users/#{@state.owner.login}">
```
parrish: Also a 404

app/pages/profile/index.cjsx
```diff
@@ -55,45 +64,69 @@ UserProfilePage = React.createClass
...
-                <Link to="/users/#{@props.profileUser.login}/stats" activeClassName="active" onClick={logClick?.bind(this, 'stats')}>
-                  <Translate content="profile.nav.stats" />
-                </Link>
-              else
-                <Link to="/users/#{@props.profileUser.login}/message" onClick={@logClick.bind null, this} activeClassName="active">
-                  <Translate content="profile.nav.message" />
-                </Link>}
-            </span>
-          </nav>
+          <h1><Avatar user={@props.profileUser} />{@props.profileUser.display_name}<span class="login" className="login-name">({@props.profileUser.login})</span></h1>
```
parrish: Should be `<span className="login"`

app/collections/show.cjsx
```diff
@@ -34,27 +38,61 @@ CollectionPage = React.createClass
...
+      baseLink += "projects/#{@props.project.slug}"
+    baseLink
+
+  getOwnerLogin: ->
+    [owner_login, collection_name] = @props.collection.slug.split('/')
+    owner_login
+
+  getCollectionsLink: ->
+    baseLink = @getBaseLink()
+    [owner_login, collection_name] = @props.collection.slug.split('/')
```
srallen: This looks unnecessary.

app/collections/show.cjsx
```diff
@@ -8,11 +8,15 @@ Avatar = require '../partials/avatar'
...
 TitleMixin = require '../lib/title-mixin'
+classNames = require 'classNames'
 
 counterpart.registerTranslations 'en',
   collectionPage:
     settings: 'Settings'
     collaborators: 'Collaborators'
+    collectionsLink: '%(user)s\'s\u00a0Collections'
+    favoritesLink: '%(user)s\'s\u00a0Favorites'
+    userLink: '%(user)s\'s\u00a0Profile'
```
srallen: Does someone's username need to be translated?

app/collections/show.cjsx
```diff
@@ -34,27 +38,61 @@ CollectionPage = React.createClass
...
+  getCollectionsLinkMessageKey: ->
+    if @props.collection.favorite
+      "collectionPage.favoritesLink"
+    else
+      "collectionPage.collectionsLink"
+
   render: ->
     <PromiseRenderer promise={@props.collection.get('owner')}>{(owner) =>
-      baseLink = "/collections/#{@props.collection.slug}"
+      baseCollectionsLink = "#{@getBaseLink()}collections/#{@props.collection.slug}/"
```
parrish: Sorry, this is where the missing slash is coming from.  Should be `baseCollectionsLink = "#{@getBaseLink()}/collections/#{@props.collection.slug}/"`

app/pages/collections/nav.cjsx
```diff
@@ -0,0 +1,45 @@
+React = require 'react'
+{Link, IndexLink} = require 'react-router'
+Translate = require 'react-translate-component'
+classNames = require 'classNames'
```
parrish: should be `require 'classnames'`  -- case matters

app/pages/profile/feed.cjsx
```diff
@@ -28,7 +28,10 @@ module.exports = React.createClass
       comments: null
       error: null
     }, =>
-      talkClient.type('comments').get({user_id: user.id, page: page, sort: '-created_at'})
+      criteria = {user_id: user.id, page: page, sort: '-created_at'}
+      if @props.project?
+        criteria.project_id = @props.project.id
```
parrish: Comments aren't filtered by project.  As previously mentioned, you can't filter by `project_id`, only **`section`**.  e.g. `"project-#{@props.project.id}"`

app/pages/profile/index.cjsx
```diff
@@ -4,10 +4,13 @@ PromiseRenderer = require '../../components/promise-renderer'
 apiClient = require 'panoptes-client/lib/api-client'
 Translate = require 'react-translate-component'
 {Link, IndexLink} = require 'react-router'
+talkClient = require 'panoptes-client/lib/talk-client'
+Avatar = require '../../partials/avatar'
+classNames = require 'classNames'
```
parrish: should also be `require 'classnames'`

app/collections/show.cjsx
```diff
@@ -8,11 +8,15 @@ Avatar = require '../partials/avatar'
 Loading = require '../components/loading-indicator'
 HandlePropChanges = require '../lib/handle-prop-changes'
 TitleMixin = require '../lib/title-mixin'
+classNames = require 'classNames'
```
parrish: should also be `require 'classnames'`

app/pages/project/index.cjsx
```diff
@@ -7,6 +7,7 @@ PotentialFieldGuide = require './potential-field-guide'
 TitleMixin = require '../../lib/title-mixin'
 apiClient = require 'panoptes-client/lib/api-client'
 {sugarClient} = require 'panoptes-client/lib/sugar'
+classNames = require 'classNames'
```
parrish: should also be `require 'classnames'`

app/collections/show.cjsx
```diff
@@ -34,27 +38,61 @@ CollectionPage = React.createClass
...
+
+  getOwnerLogin: ->
+    [owner_login, collection_name] = @props.collection.slug.split('/')
+    owner_login
+
+  getCollectionsLink: ->
+    baseLink = @getBaseLink()
+    [owner_login, collection_name] = @props.collection.slug.split('/')
+    if @props.collection.favorite
+      "#{baseLink}favorites/#{@getOwnerLogin()}"
```
parrish: Should be `"#{baseLink}/..."`

app/collections/show.cjsx
```diff
@@ -34,27 +38,61 @@ CollectionPage = React.createClass
...
+    else
+      "#{baseLink}collections/#{@getOwnerLogin()}"
```
parrish: Should be `"#{baseLink}/..."`

app/collections/show.cjsx
```diff
@@ -34,27 +38,61 @@ CollectionPage = React.createClass
...
+
+  getProfileLink: ->
+    baseLink = @getBaseLink()
+    "#{baseLink}users/#{@getOwnerLogin()}"
```
parrish: Should be `"#{baseLink}/..."`

css/owned-list-pages.styl
```diff
@@ -19,6 +24,16 @@
         margin: 0
         padding: 2.33em 0
 
+  .in-project-context
+
+    .collection-results-counter
+
+        p:first-child
+          padding-top: 0px !important
```
srallen: Don't think `!important` is necessary here

css/owned-list-pages.styl
```diff
@@ -19,6 +24,16 @@
...
+
+     nav.pagination
+        color: #404040
```
srallen: Could you use the defined color variable in `common.styl`?

app/pages/collections/list.cjsx
```diff
@@ -24,33 +26,72 @@ List = React.createClass
...
+    classes = classNames {
+      "resources-container": true
+      "collections-container": true
+      "in-project-context": @props.project?
+    }
+
+    baseLink = if @props.project? then "/projects/#{@props.project.slug}/" else "/"
+    baseType = if @props.favorite then "favorites" else "collections"
+    shouldShowMoreLink = @props.params.profile_name? || @props.params.collection_owner?
+    showMoreLink = "#{baseLink}#{baseType}/"
```
parrish: This is confusing when it's labeled as "Show more" -- I really didn't expect this to go to all project collections.

app/pages/collections/list.cjsx
```diff
@@ -24,33 +26,72 @@ List = React.createClass
...
 
   cardLink: (collection) ->
-    [collection_owner, collection_name] = collection.slug.split('/')
-    "/collections/#{collection_owner}/#{collection_name}"
+    baseLink = "/"
+    if @props.project?
+      baseLink += "projects/#{@props.project.slug}/"
+    "#{baseLink}collections/#{collection.slug}"
 
   listCollections: ->
```
parrish: When viewing a [user's collections](http://localhost:3735/projects/alexbfree/computer-vision-serengeti/users/Whoandwhatitis/collections?env=production), this doesn't filter correctly.

app/pages/collections/index.cjsx
```diff
@@ -1,54 +1,57 @@
...
 Translate = require 'react-translate-component'
-{Link, IndexLink} = require 'react-router'
+CollectionsNav = require './nav'
+classNames = require 'classnames'
 
 counterpart.registerTranslations 'en',
+  projectcollectionsPage:
+    title: '%(user)s %(project)s Collections'
+    countMessage: 'Showing %(count)s favorites collections.'
+    button: 'View Favorites Collection'
```
parrish: Should be 'View Collection'

css/secondary-page.styl
```diff
@@ -40,6 +40,16 @@
...
+          img.avatar
+            padding-right: 10px !important
+
+          span.login-name
+            padding-left: 7px !important
+
+            span
+              font-size: 0.65em !important
+              font-weight: 200 !important
+
```
srallen: Could these be scoped within css where they're being used and use clear class names rather than tag? `!important` should be avoided and they don't look necessary here anyway. Also, you can apply the `font-size` and `font-weight` to `login-name` itself.

app/pages/collections/favorites-list.cjsx
```diff
@@ -8,7 +8,14 @@ FavoritesList = React.createClass
   title: 'Favorites'
 
   render: ->
-    props = Object.assign({}, @props, {favorite: true, translationObjectName:"favoritesPage"})
+    translationObjectName = "favoritesPage"
```
parrish: app/pages/collections/collections-list.cjsx and app/pages/collections/favorites-list.cjsx  should be refactored to cut down on duplicated code.  A small module to return `{favorite, translationObjectName}` would probably do it.

app/pages/notifications/moderation.cjsx
```diff
@@ -39,9 +39,12 @@ module.exports = React.createClass
         @setState {reports}
 
   render: ->
+    baseLink = "/"
+    if @props.project?
+      baseLink += "projects/#{@props.project.slug}/"
     notification = @props.notification
     [owner, name] = notification.project_slug.split('/') if notification.project_slug
-    path = if notification.project_id then "/projects/#{owner}/#{name}/talk/moderations" else '/talk/moderations'
+    path = if notification.project_id then "/projects/#{notification.project_slug}/talk/moderations" else '/talk/moderations'
```
parrish: If you want to make this change, then you can remove line 46: `[owner, name] = notification.project_slug.split('/') if notification.project_slug` as it won't be needed.

app/pages/profile/comment-link.cjsx
```diff
@@ -59,11 +59,14 @@ module.exports = React.createClass
             <a href={@state.href}>
               {if @state.project? and @state.owner
                 <span>
-                  <strong className="comment-project" title="#{@state.owner.display_name}/#{@state.project.display_name}">{@state.project.display_name}</strong>
-                  ➞
+                  {if !@state.project?
```
parrish: I'm confused.  This is

``` coffeescript
if @state.project?
  if !@state.project?
    # impossible
```

css/secondary-page.styl
```diff
@@ -105,6 +115,33 @@
...
+    .hero-container
+      margin: 0 !important
+      max-width: initial !important
+
+      h1
+        text-align: center
+    .overlay
+      background-image: none !important
+      background-color: transparent !important
```
srallen: If you instead scope this under `on-seconday-page`, all of the `!important`s can be removed and the styles apply correctly.

css/secondary-page.styl
```diff
@@ -105,6 +115,33 @@
...
+
+  .content-container
+    margin: 1em auto
+
+  nav.hero-nav
+    a
+      span
+        text-transform:none !important
```
srallen: Ditto comments on using a class name, scoping, and removing the `!important`

app/collections/show.cjsx
alexbfree: No, but since word order varies between language, the best way to globalize a message involving the username is to pass it in to the full message so it can be correctly positioned.

Consider French for example: "Les collections de Alex"

The alternative would be to only translate part of a sentence and encode the word order algorithmically - which I am sure you can see would be a bad idea.

app/pages/profile/index.cjsx
alexbfree: good spot, thanks

app/pages/profile/feed.cjsx
alexbfree: Ah yes, sorry, this got missed between my old branch and the new.

app/pages/profile/index.cjsx
alexbfree: 👍

app/collections/show.cjsx
alexbfree: happy to change these but out of interest why does it matter if the slash comes from the baseLink function/variable or from the string itself?

css/owned-list-pages.styl
alexbfree: Jordan is gradually introducing new colours, I didn't want to change it everywhere yet, but maybe I will. I'll add a constant anyway. Thanks.

app/pages/collections/list.cjsx
alexbfree: Show more doesn't take you to all project collections.

Show more takes you too all collections within this project (having been focussed on a user's collections).
The link to all project collections explicitly says "Show all Zooniverse collections".

app/pages/collections/index.cjsx
alexbfree: because why? You don't think we should distinguish between an ordinary collection and a favorites collection? I thought it could be useful to distinguish.....

css/secondary-page.styl
alexbfree: I know that `!important` should be avoided, but I am not sure how to reorder things with Stylus such that these aren't needed, which is what I would do in normal CSS... I am assuming we don't want any repetition of classes, nor to change the order in which the `.styl` files are loaded... so am open to suggestions on how I can move these to a different point in the CSS order .... how to do this? add a new `.styl` file?

app/pages/collections/favorites-list.cjsx
alexbfree: ok

app/pages/profile/comment-link.cjsx
alexbfree: yes that doesn't look right does it, something must have got messed up in my merge from the other branch. will review.

css/secondary-page.styl
alexbfree: ah great - thanks! I tried removing them but perhaps not in that exact scoping.

app/pages/collections/list.cjsx
alexbfree: well spotted, thank you.

app/pages/collections/index.cjsx
alexbfree: oh never mind I see what you mean! copy and paste error!

app/pages/collections/list.cjsx
alexbfree: this is now fixed

css/secondary-page.styl
alexbfree: Got it working now, no need of help :-)

css/owned-list-pages.styl
alexbfree: ignore that, what I said refers to the dark blue colour not the `FOREGROUND` colour.

alexbfree:
OK @srallen @parrish Thanks for the feedback. I think I've addressed all your comments now. 
https://maintain-project-context-v2.pfe-preview.zooniverse.org/
Have redeployed to staging. Good to merge?

srallen:
A couple of style issues I noticed: 

Messaging when you don't have any collection items is hard to read:
screen shot 2016-07-19 at 11 43 45 am

Button is pushed to the bottom of the card area. Is the button itself really necessary?
screen shot 2016-07-20 at 9 00 52 am

css/owned-list-pages.styl
```diff
@@ -5,6 +5,11 @@
   .collections-hero
     background: url(/assets/all-projects-hero.png) center center no-repeat
 
+  a.show-more
+    font-weight: normal
+    padding-left: 10px
```
srallen: Could you use `em`?

css/project-list-pages.styl
```diff
@@ -1,75 +1,74 @@
...
-      position: relative
-      text-align: center
-      overflow: visible
-      padding: 0
-      max-width: 1200px
-      margin: 0 auto
-      .pagination
-        padding: 0.5em 0
+  p
+    font-size: 14px
```
srallen: Should use `em`

css/project-list-pages.styl
```diff
@@ -1,75 +1,74 @@
...
+    margin: 0
+    padding: 2.33em 0
 
-    p
-      font-size: 14px
-      margin: 0
-      padding: 2.33em 0
+  p.showing-with-link-para
+    padding-top: 20px
+    padding-bottom: 20px
```
srallen: Use `em`

app/collections/show.cjsx
srallen: Got it. Still don't think `\u00a0` is necessary. This could be achieved by `whitespace: nowrap` in the css.

app/collections/show.cjsx
```diff
@@ -36,25 +40,38 @@ CollectionPage = React.createClass
...
-            <Link to="#{baseLink}/settings" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'settings-collection')}>
+            <Link to="#{baseCollectionLink}/settings" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'settings-collection')}>
               <Translate content="collectionPage.settings" />
             </Link>}
 
           {if isOwner
-            <Link to="#{baseLink}/collaborators" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'collab-collection')}>
+            <Link to="#{baseCollectionLink}/collaborators" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'collab-collection')}>
               <Translate content="collectionPage.collaborators" />
             </Link>}
```
srallen: This could be dried up by putting both links under a single check for `isOwner`

app/pages/collections/nav.cjsx
```diff
@@ -0,0 +1,45 @@
...
+
+      {if @props.user?
+        <Link to="#{baseLink}/collections/#{@props.user.login}" className={classes} activeClassName="active" onClick={@logClick?.bind(this, "collections.my")}>
+          <Translate content="collectionsPage.my" />
+        </Link>}
+
+      {if @props.user?
+        <Link to="#{baseLink}/favorites/#{@props.user.login}" className={classes} activeClassName="active" onClick={@logClick?.bind(this, "favorites.my")}>
+          <Translate content="favoritesPage.my" />
+        </Link>}
```
srallen: Put both links under a single `@props.user` if statement.

app/pages/profile/comment-link.cjsx
```diff
@@ -57,13 +57,13 @@ module.exports = React.createClass
           <span>
             {' '}in{' '}
             <a href={@state.href}>
-              {if @state.project? and @state.owner
+              {if @state.project? and @state.owner? and !@props.project?
```
srallen: This is confusing. Why is a project both in state and in props?

parrish: I was just trying to figure that out too.  I suppose this would be true if you were viewing a user profile outside of a project.  e.g. `/users/parrish`.  That would have `@props.project` (the project context) undefined and `@state.project` (the project of the comment) defined.

Definitely confusing though.  Maybe some renaming would help?

app/pages/collections/list.cjsx
```diff
@@ -24,33 +38,73 @@ List = React.createClass
...
-                count = meta.count
-                <Translate pageStart={pageStart} pageEnd={pageEnd} count={count} content="#{@props.translationObjectName}.countMessage" component="p" />}
+              <p>
+                {if meta
+                  pageStart = meta.page * meta.page_size - meta.page_size + 1
+                  pageEnd = Math.min(meta.page * meta.page_size, meta.count)
+                  count = meta.count
+                  <Translate pageStart={pageStart} pageEnd={pageEnd} count={count} content="#{@props.translationObjectName}.countMessage" />}
+                {if shouldShowMoreLink
+                  <Link className="show-more" to="#{showMoreLink}">
```
srallen: `showMoreLink` doesn't need to be string interpolated. It's already a string. Just set it as `to={showMoreLink}`

app/pages/collections/list.cjsx
```diff
@@ -24,33 +38,73 @@ List = React.createClass
...
+                    <Translate content="#{@props.translationObjectName}.more" />
+                  </Link>}
+                {if shouldShowZooWideLink
+                  <Link className="show-more" to="#{showZooWideLink}">
```
srallen: String interpolation unnecessary for `to`.

app/pages/collections/list.cjsx
```diff
@@ -24,33 +38,73 @@ List = React.createClass
...
+                    <Translate content="#{@props.translationObjectName}.zooWide" />
+                  </Link>}
+                {if shouldShowUserWideLink
+                  <Link className="show-more" to="#{showUserWideLink}">
```
srallen: String interpolation unnecessary for `to`.

app/pages/collections/list.cjsx
```diff
@@ -24,33 +38,73 @@ List = React.createClass
...
+                    <Translate content="#{@props.translationObjectName}.userWide" user="#{userForUserWideLink}"/>
```
srallen: String interpolation unnecessary for `user`.

app/pages/profile/comment-link.cjsx
parrish: I also just realized how kludgey and request heavy this component is too, but I'm happy to refactor in a separate PR later.

app/collections/show.cjsx
```diff
@@ -36,25 +40,38 @@ CollectionPage = React.createClass
...
+      baseCollectionLink = "#{baseLink}/#{baseType}/#{@props.collection.slug}"
+      baseCollectionsLink = "#{baseLink}/#{baseType}/#{owner.login}"
       isOwner = @props.user?.id is owner.id
+      profileLink = "#{baseLink}/users/#{owner.login}"
+      collectionsLinkMessageKey = "collectionPage.#{baseType}Link"
 
       <div className="collections-page">
         <nav className="collection-nav tabbed-content-tabs">
-          <IndexLink to="#{baseLink}" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'view-collection')}>
+          <IndexLink to="#{baseCollectionLink}/#" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'view-collection')}>
```
parrish: Should be `to={baseCollectionLink}` (also dropping the `/#`)

app/pages/profile/index.cjsx
```diff
@@ -35,6 +43,7 @@ UserProfilePage = React.createClass
     @getProfileHeader(@props.profileUser)
 
   componentWillReceiveProps: (nextProps) ->
+    document.documentElement.classList.add 'on-secondary-page'
```
srallen: This shoudn't be necessary. It's set in `componentDidMount`

app/collections/show.cjsx
```diff
@@ -36,25 +40,38 @@ CollectionPage = React.createClass
...
+            <Link to="#{baseCollectionLink}/settings" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'settings-collection')}>
               <Translate content="collectionPage.settings" />
             </Link>}
 
           {if isOwner
-            <Link to="#{baseLink}/collaborators" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'collab-collection')}>
+            <Link to="#{baseCollectionLink}/collaborators" activeClassName="active" className="tabbed-content-tab" onClick={@logClick?.bind(this, 'collab-collection')}>
               <Translate content="collectionPage.collaborators" />
             </Link>}
+          <Link to="#{baseCollectionsLink}/" className="tabbed-content-tab">
```
parrish: Should be `to={baseCollectionsLink}`

app/pages/project/index.cjsx
```diff
@@ -134,6 +136,11 @@ ProjectPage = React.createClass
 
     logClick = @context?.geordi?.makeHandler? 'project-menu'
 
+    collectClasses = classNames {
+      "tabbed-content-tab": true
+      "active": @props.project? and (@props.routes[2].path=="collections" or @props.routes[2].path=="favorites")
```
srallen: Use `is` or `===` instead of `==`.

app/collections/show.cjsx
```diff
@@ -36,25 +40,38 @@ CollectionPage = React.createClass
...
+            <Translate content="#{collectionsLinkMessageKey}" user="#{@props.collection.links.owner.display_name}" />
+          </Link>
+          <Link to="#{profileLink}/" activeClassName="active" className="tabbed-content-tab">
```
parrish: Should be `to={profileLink}`

parrish:
Awesome.  Lots of good changes.  Mostly just minor tweaks left.

app/pages/collections/list.cjsx
parrish: Yeah, that's what I mean by project collections -- all collections for the project.

I expect "Show more" to show more collections for the currently viewed user.

I think "Show all #{project}'s collections" would fit better alongside "Show all #{user}'s Zooniverse collections"

parrish: Either that or drop it entirely, since there's a nav button just above it.

css/buttons.styl
```diff
@@ -44,6 +44,8 @@
 
 .pill-button
   @extends $reset-button
+  color: FOREGROUND !important
+  border-color: FOREGROUND !important
```
srallen: The `pill-button` class is used around the whole site especially in the project builder. This is modifying this class across the board. Please don't add these here with `!important`. Instead write a new class scoped where it will be used and use `@extends` if you need to.

css/project-page.styl
```diff
@@ -183,9 +183,32 @@
...
+    color: #0072ff
+
+&.in-project-context
+  nav.hero-nav
+    text-align: center
+  a.about-tabs
+    color: PROJECT_SUBMENU_BLUE
+  a.about-tabs.active,
+  a.about-tabs:hover
+    border: 2px solid PROJECT_SUBMENU_BLUE !important
```
srallen: `!important` is not necessary here.

srallen:
Checking in with Jordan on the collections cards: Corners should be rounded like the project cards and the `View Collection` button can go away. The card itself can link like it does on the projects page.

app/pages/collections/nav.cjsx
```diff
@@ -0,0 +1,45 @@
...
+
+  componentWillReceiveProps: (nextProps, nextContext)->
+    @logClick = nextContext?.geordi?.makeHandler? 'collect-menu'
+
+  render: ->
+    classes = classNames {
+      "about-tabs": @props.project?
+    }
+    baseLink = ""
+    messageKeyPrefixToUse=""
```
srallen: Could you add spaces around `=` in the assignment here? My text editor thinks everything after it is a string.

app/pages/collections/list.cjsx
srallen: I have to agree with Michael. I expected it to show more collections for the currently viewed page. Dropping it, imo, is a good idea.

app/pages/collections/list.cjsx
```diff
@@ -4,10 +4,24 @@ PromiseRenderer = require '../../components/promise-renderer'
...
 
 List = React.createClass
   displayName: 'List'
 
+  statics: {
+    getPropsForList: (props,favorite)->
+      translationObjectName = "collectionsPage"
+      if props.routes[1]=="collections" or props.routes[1]=="favorites"
+        translationObjectName = "#{props.routes[1]}Page"
+      else if props.routes[2]=="collections" or props.routes[2]=="favorites"
```
srallen: Use `is` or `===` instead of `==`

app/partials/collection-card.cjsx
```diff
@@ -64,6 +64,7 @@ module.exports = React.createClass
         <svg className="card-space-maker" viewBox="0 0 2 1" width="100%"></svg>
         <div className="details">
           <div className="name"><span>{@props.collection.display_name}</span></div>
+          {<div className="public-spaceholder">&nbsp;</div> if !@props.collection.private}
```
srallen: I think this could be achieved with css alone and/or just have the private messaging display inline with the username.

alexbfree: ok. it was just a quick hack to fix an alignment issue i noticed. Not really relevant to this PR so will remove it for now.

app/pages/collections/nav.cjsx
alexbfree: ok

app/pages/profile/index.cjsx
alexbfree: Shouldn't be, and yet, I'm sometimes seeing the recent comments / stats pages load up without styling, so i added this as a catch-all. not sure why that is happening. i'll take another look

app/pages/collections/list.cjsx
alexbfree: Is this the same point as the "does user need to be translated" query earlier? I explained in answer to that why passing the user to the translate function is the correct way to globalize a string containing a username - otherwise you have to algorithmically encode word order, which is language-specific. Or am i missing something?

alexbfree: Oh ok, i see what you mean {userForUserWideLink} instead of "#{userForUserWideLink}"

app/pages/profile/comment-link.cjsx
alexbfree: @state.project is the project to which the comment belongs.
@props.project is the project whose context you are currently viewing things in. 

They are not necessarily the same.

alexbfree: yes, would prefer not refactor as part of this - don't want to add anything else into this PR.

srallen: One of those should be renamed so it's clear.

app/pages/collections/nav.cjsx
alexbfree: 👍

css/project-list-pages.styl
alexbfree: 👍

app/partials/collection-card.cjsx
alexbfree: or actually if i'm removing the button i might just reposition the "Private" label

app/pages/collections/list.cjsx
alexbfree: >I think "Show all #{project}'s collections" would fit better alongside "Show all #{user}'s Zooniverse collections"

Was trying to keep things short. Will consider dropping it.

This whole thing is a tradeoff between surfacing content that might not be evident, vs not having too many menus/options.

app/collections/show.cjsx
alexbfree: Oh, sure, i had no idea your comment was anything to do with `\u00a0`. 
`whitespace: nowrap` is a good idea, will use that.

srallen:
One more style thing, the header is hard to read on the profile if the user has a background set. Maybe opacity?

screen shot 2016-07-20 at 10 09 33 am

css/owned-list-pages.styl
alexbfree: 👍

css/project-list-pages.styl
alexbfree: 👍

app/collections/show.cjsx
alexbfree: 👍

app/collections/show.cjsx
alexbfree: 👍

app/collections/show.cjsx
alexbfree: 👍 

Was getting carried away with the whole "only one surrounding component" thing but I guess that only applies when returning from a `render` function.

app/pages/collections/list.cjsx
alexbfree: 👍

app/pages/collections/list.cjsx
alexbfree: Re:
> Actually, it really makes the most sense to move the /users/name/collections link into the nav bar.

This takes us back to the "To the Zooniverse!" idea that was rejected last time. I'll put it back (I actually prefer it up there) but think we are going round in circles on this one!!

srallen:
Sorry, one last one. The header and nav needs a bit of work on mobile sized viewports. This is using Chrome's dev tools with the iPhone 5 option: 
screen shot 2016-07-20 at 11 32 24 am

app/pages/collections/list.cjsx
srallen: IMO, it's really Jordan's call on this one.

alexbfree: yes, one i have reworked it, will run it by her.

app/collections/show.cjsx
alexbfree: or in fact, it does apply - but as discussed will use a surrounding `<span>` to get around this.

alexbfree:
@srallen re the user profile header thing; I tried darkening the background, but it looks weird with two different backgrounds in place, very jarring no matter what the images are I think. 

e.g.:
screenshot 2016-07-21 15 07 01

I think the better way to go is to have it so that the user's profile background only applies to the Zoo-wide user profile - so I am going to go that way. User profile background will be suppressed and project background takes priority when viewing the user's profile within a project context. Hope you are OK with that. Also that puts project creators in control of how project pages look, rather than users who cannot possibly anticipate how their profile image will look against all different project backgrounds...

alexbfree:
OK I have fixed the spacing on the cards, added corners, removed the buttons, so it now looks like this. Not the most beautiful thing ever, but better than it was and good enough for now I think:
screenshot 2016-07-21 15 50 22

alexbfree:
line spacing when wrapping on multiple lines is now fixed (regardless of screen width)

alexbfree:
colouring of "not found" text is also fixed.

alexbfree:
OK @srallen @parrish All of the second round of review comments have now been addressed, ready for final review. 

Redeployed to https://maintain-project-context-v2.pfe-preview.zooniverse.org/

app/collections/preview.cjsx
parrish: `collection` isn't defined.  Did you mean `@props.collection`?

app/pages/collections/list.cjsx
```diff
@@ -64,24 +132,40 @@ List = React.createClass
...
                     </Link>}
                 </nav>}
             </nav>
           </div>
         else if collections?.length is 0
-          <Translate content="#{@props.translationObjectName}.notFoundMessage" component="div" />
+          <div>
+            <div className="resource-results-counter collection-results-counter">
+              <p>
+                <Translate content="#{@props.translationObjectName}.notFoundMessage" component="div" />
```
parrish: `<div>` can't be a descendant of a `<p>`

app/pages/collections/list.cjsx
```diff
@@ -64,24 +132,40 @@ List = React.createClass
...
+              </p>
+            </div>
+          </div>
         else
-          <Translate content="#{@props.translationObjectName}.loadMessage" component="div" />
+          <div>
+            <div className="resource-results-counter collection-results-counter">
+              <p>
+                <Translate content="#{@props.translationObjectName}.loadMessage" component="div" />
```
parrish: `<div>` can't be a descendant of a `<p>`

app/pages/profile/index.cjsx
```diff
@@ -55,45 +63,69 @@ UserProfilePage = React.createClass
...
+      </Link>
+      {' '}
+
+      <span>
+        {if @props.user is @props.profileUser
+          <Link to="#{baseLink}users/#{@props.profileUser.login}/stats" className={classes} activeClassName="active" onClick={logClick?.bind(this, 'stats')}>
+            <Translate content="profile.nav.stats" />
+          </Link>
+        else
+          <Link to="#{baseLink}users/#{@props.profileUser.login}/message" className={classes} activeClassName="active"onClick={@logClick.bind null, this}>
```
parrish: Missing space between `activeClassName="active"onClick={@logClick.bind null, this}`

app/pages/collections/list.cjsx
```diff
@@ -64,24 +132,40 @@ List = React.createClass
...
                 <nav className="pagination">
                   {for page in [1..meta.page_count]
                     active = (page is +location.query.page) or (page is 1 and not location.search)
                     <Link
-                      key={page}
-                      to={"#{@props.location.pathname}?page=#{page}"}
-                      activeClassName="active"
-                      className="pill-button"
-                      style={border: "2px solid" if active}>
+                      key = {page}
```
parrish: These shouldn't be spaced.  e.g. `key={page}` rather than `key = {page}`

parrish:
Other than these few changes, I think this is ready to go.

app/pages/collections/list.cjsx
alexbfree: I guess that contradicts this advice: https://github.com/zooniverse/Panoptes-Front-End/pull/2804#discussion_r71548262

When do we have spaces and when don't we? spaces for assignments but not for attributes, is that it?

alexbfree:
@jwbmartin Please could you review these new pages & navigation options?

Some test URLs for you:

*In Project Context*:

 - **Project Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/favorites?env=production
 - **Project Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/collections?env=production
 - **User's Profile & Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/users/mschwamb/collections?env=production
 - **User's Profile & Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/favorites/ScienceyGirl?env=production
 - **Collection**: hhttps://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/pszmhk/understanding-animal-faces/collections/zyais/cute-animals?env=production
- **Your Own Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/PROJECTOWNER/PROJECTNAME/users/YOURUSERNAMEHERE?env=production
- **Your Own Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/PROJECTOWNER/PROJECTNAME/users/YOURUSERNAMEHERE?env=production

*Out of Project Context*:

 - **Project Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/favorites?env=production
 - **Project Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/collections?env=production
 - **User's Profile & Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/users/mschwamb/collections?env=production
 - **User's Profile & Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/favorites/ScienceyGirl?env=production
 - **Collection**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/collections/rc135tx/deer?env=production
- **Your Own Collections**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/users/YOURUSERNAMEHERE?env=production
- **Your Own Favorites**: https://maintain-project-context-v2.pfe-preview.zooniverse.org/users/YOURUSERNAMEHERE?env=production

Please note, the collection cards are intended to be "good enough". I'm fully expecting that you and I (or anyone else) will need to connect up and improve the look of them, so don't be too critical - they are supposed to be an improvement on current, but not necessarily final - we would improve them in another PR. Thanks Jordan!

app/pages/collections/list.cjsx
parrish: Yeah, I suppose it does, but everywhere else in the app and general HTML conventions are unspaced.

I suspect that this line is the culprit: `style = {border: "2px solid" if active}>`

The trailing `>` should be removed.

`style={ {border: "2px solid"} if active}` may also fix the syntax parsing.

@srallen, what's your editor?  Do either of those suggestions work for you?

alexbfree:
ok @parrish have addressed your third round comments. Assuming you are happy, we now just need the OK from @srallen and @jwbmartin .

css/owned-card.styl
```diff
@@ -64,6 +64,14 @@
   .name
     margin-bottom: 1em
     height: 2.9em
+    overflow: hidden
+    text-overflow: ellipsis
+    display: -webkit-box
+    -webkit-box-orient: vertical
```
parrish: [nib](https://tj.github.io/nib/) will handle browser prefixing, so `box-orient` is what you want to use here.  `line-clamp` is pretty unsupported.   [A cross-browser solution](http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/) is pretty involved.

css/owned-card.styl
```diff
@@ -58,8 +58,28 @@
...
+  border-radius: 1em
+
   .name
     margin-bottom: 1em
+    height: 2.9em
+    overflow: hidden
+    text-overflow: ellipsis
+    display: -webkit-box
+    -webkit-box-orient: vertical
+    -webkit-line-clamp: 2
```
parrish: [nib](https://tj.github.io/nib/) will handle browser prefixing, so `box-orient` is what you want to use here.  `line-clamp` is pretty unsupported.   [A cross-browser solution](http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/) is pretty involved.  (reposted, because GitHub outdated it immediately)

app/pages/collections/list.cjsx
```diff
@@ -64,24 +132,36 @@ List = React.createClass
...
+                }
                 <nav className="pagination">
                   {for page in [1..meta.page_count]
                     active = (page is +location.query.page) or (page is 1 and not location.search)
                     <Link
                       key={page}
                       to={"#{@props.location.pathname}?page=#{page}"}
-                      activeClassName="active"
-                      className="pill-button"
+                      activeClassName= "active"
```
parrish: This space just reappeared

app/pages/collections/list.cjsx
```diff
@@ -64,24 +132,36 @@ List = React.createClass
...
+                      className={buttonClasses}
                       style={border: "2px solid" if active}>
```
parrish: That trailing `>` should be removed

css/owned-card.styl
alexbfree: fair enough. I grabbed this solution [here](http://stackoverflow.com/questions/3922739/limit-text-length-to-n-lines-using-css)

So what do you think I should do here? Trying to force a uniform height on the cards one way or another.... I could just set max height to 3 lines, but then there might be too much white space.... 

(are you saying don't bother trying to make it work cross-browser? or use the involved solution from the link)? or don't do this at all?)

srallen: Isn't `display: -webkit-box` and the `box` syntax deprecated anyway? [caniuse.com](http://caniuse.com/) is a great reference for this, also noted as deprecated on [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/box-flex) too.

app/pages/collections/list.cjsx
alexbfree: I think that is not correct. It's the closing of the `<Link` component opener, which the attributes have to be within. If I remove it, the template breaks.

alexbfree: unless i missed some other way of doing it, of course...

app/pages/collections/list.cjsx
alexbfree: ok fixed

app/pages/collections/list.cjsx
parrish: Barf.  Yeah, I was misreading `{page}` as an attribute.  That `>` is necessary there.

css/owned-card.styl
brian-c: Nib will read in modern flexbox properties and output something as cross-browser compatible as possible. In general, manually included vendor prefixes should be a rare occurrence.

I've got a little brute-force string truncator component coming with the new home page, which might be useful here. https://github.com/zooniverse/Panoptes-Front-End/blob/new-signed-in-home-page/app/pages/home-for-user/string-truncator.jsx

alexbfree: For now I've switched it to `box-orient`. Seems to work okay on FF, Safari, Chrome.

srallen: `box-orient` is still a part of the old deprecated spec, FYI. I'd avoid using it. IMO, truncation could come in a separate PR.

alexbfree: ok will back out the truncation for now then

parrish: Honestly, this works fine without
``` sass
display: -webkit-box
-webkit-box-orient: vertical
-webkit-line-clamp: 2
```

srallen: `text-overflow: ellipsis` is supported by FF now. Stackoverflow answers can be out of date.

alexbfree: just saw this after committing the back out. have added the rest back.

alexbfree:
Squashed those last few minor commits together.

alexbfree:
Subject to Jordan's approval, are we good now as far as you are both concerned @parrish @srallen ? Thanks so much for the detailed review!!!

app/pages/profile/feed.cjsx
```diff
@@ -44,6 +44,7 @@ module.exports = React.createClass
           @setState({comments})
 
   render: ->
+    console.log @props.params
```
parrish: Snuck in a log

css/owned-card.styl
```diff
@@ -58,8 +58,25 @@
       background: lighten(COBOLT_BLUE, 20%)
 
 .collection-card
+  -webkit-border-radius: 1em
```
srallen: As mentioned, nib will handle this for you, unless there's a specific reason to include this... if so, leave a comment? I usually leave a comment when I do and often it's for `-ms` prefixes for IE solutions.

app/collections/preview.cjsx
```diff
@@ -28,17 +28,18 @@ module.exports = React.createClass
...
-    [owner, name] = @props.collection.slug.split('/')
-
+    baseLink = ""
+    if @props.project?
+      baseLink += "projects/#{@props.project.slug}"
     <div className="collection-preview">
       <div className="collection">
         <p className="title">
-          <Link to="/collections/#{owner}/#{name}">
+          <Link to="#{baseLink}/collections/#{@props.collection.slug}">
```
parrish: Unfortunately, this link can't be relative.

For instance on [subject pages](https://maintain-project-context-v2.pfe-preview.zooniverse.org/projects/brian-testing/ghosts/talk/subjects/3154) the preview is in the "Collections" section and builds a URL like `/projects/brian-testing/ghosts/talk/subjects/projects/brian-testing/ghosts/collections/parrish/test2`.  Wasn't this `/` prefixed a while ago?

app/pages/profile/feed.cjsx
srallen: Just an FYI, the react dev tools for chrome are useful for checking a component's current state and props.

app/collections/preview.cjsx
parrish: Yeah. `baseLink = "/"` was removed [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2804/commits/8eb38c85c436e1a884fafdeb6b46b5bfd89d73ff#diff-28402a6801a176382c8b9c1d36662a6fR31)

app/collections/preview.cjsx
```diff
@@ -28,17 +28,18 @@ module.exports = React.createClass
       @setState {subjects}
 
   render: ->
-    [owner, name] = @props.collection.slug.split('/')
-
+    baseLink = ""
+    if @props.project?
```
parrish: Actually, this is a required prop, isn't it?  In other words, this is only rendered in the context of a project I think

parrish: Yeah, this component will always have a project.

app/pages/profile/feed.cjsx
alexbfree: 👍  Thanks!
I know about that, just stuck in old habits :-)

css/owned-card.styl
alexbfree: 👍

app/collections/preview.cjsx
alexbfree: Sorry, yeah, I switched all the `baseLink`s from `"/"` to `""` per an earlier code review comment, looks like I missed the `/` from the project version of `baseLink` in this case. Well spotted.

app/collections/preview.cjsx
alexbfree: Ah good point - used by Subjects, and Zoo-wide Talk doesn't have Subjects. Will fix.

alexbfree:
OK. Hopefully we are nearly there. How about now @srallen @parrish ?

parrish:
I think this is good to go.  @srallen: care to take another look?

jwbmartin:
👍

css/buttons.styl
```diff
@@ -44,6 +44,8 @@
 
 .pill-button
   @extends $reset-button
+  color: FOREGROUND
+  border-color: FOREGROUND
```
srallen: Looks like colors are still being defined on the `pill-button` class here. I don't really want to go through the whole site where this class is being used to make sure there aren't unintended issues with this. Could you add the colors on the class where it's being used?

srallen:
One small comment on the CSS, otherwise looks good.

alexbfree:
@srallen OK am back from vacation and have addressed that CSS thing now - also have rebased. Please go ahead and merge if you are happy.

alexbfree:
PS thanks @jwbmartin for the approval :-)

css/user-profile.styl
```diff
@@ -18,6 +18,16 @@
     .hero-container
       position: relative
       z-index: 1
+      margin: 0
+      max-width: initial
+
+      img.avatar
+        padding-right: 10px
```
srallen: Apologies this wasn't caught earlier. This padding is causing the avatar to have an odd shape: 
screen shot 2016-08-02 at 1 44 46 pm

srallen:
Generally, I think the css could use a lot more refactoring, but if you can address that css fix with the avatar, then I won't hold this up any longer. Further iteration can come in subsequent PRs.

css/user-profile.styl
alexbfree: Well spotted, sorry i missed that too. fixed now.

alexbfree:
@srallen I think you are right about the CSS. In the absence of any design docs about what is the scope of each file or which classes are supposed to be used where, it's not always obvious where to put things. If you want to raise any issues with specifics that'd be good.

alexbfree:
@srallen the avatar shape is fixed now, we should be good now? thanks.

PR #2805, opened 2016-07-18
Detect autofill in login dialog

### LINKED ISSUE #630: Can't log in with saved password from Apple keychain
Safari 7.1 on Mavericks. If I use my saved username/password from the keychain, the sign-in button doesn't activate, and the enter key won't submit the login form.
panoptes-signin-bug

parrish:
Probably fixes #630

### DIFF
--- a/app/partials/login-dialog.cjsx
+++ b/app/partials/login-dialog.cjsx
@@ -1,5 +1,6 @@
 counterpart = require 'counterpart'
 React = require 'react'
+autofill = require 'react-autofill'
 Translate = require 'react-translate-component'
 SignInForm = require './sign-in-form'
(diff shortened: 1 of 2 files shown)


parrish:
Verified fixed in safari: https://fix-autofill-event.pfe-preview.zooniverse.org/

parrish:
Hmm... That library is using an [interval to monitor changes](https://github.com/Pephers/react-autofill/blob/master/src/index.js#L27-L39).  I'll look to see if there are alternatives.

parrish:
And there really aren't.  I actually don't have this issue in Safari -- the sign in button works on autofill without this.  Maybe it doesn't in older Safari versions, but I think we shouldn't merge this.

PR #2806, opened 2016-07-20
Adjust World Wide Telescope Integration

### LINKED ISSUE #2676: World Wide Telescope Adjustment
- **Adjust true center of cropped image**: Cropped chart is displayed slightly off center on WWT because of discrepancy between coordinate center and chart center.

- **Issue with axis values**: Classification will crash if axis value is placed inside chart rather than outside due to how the `filterBounds` function determines axes. Instructions should be given to the user/classifier to place annotation outside the chart. A try/catch should be added to handle errors. 

- **Account for Axes with Inconsistent Values**: Some axes list RA in hours, minutes, and seconds for one value, but not for all.

wgranger:
Staged at: https://adjust-wwt.pfe-preview.zooniverse.org/projects/wgranger-test/old-astro-test/classify

**Note**: In order to work property once merged, the workflow for Astronomy Rewind (in production) will have to include RA1950 and DEC1950 options in the order specified with this PR.

Fixes #2676

### DIFF
--- a/app/classifier/world_wide_telescope.cjsx
+++ b/app/classifier/world_wide_telescope.cjsx
@@ -108,11 +108,12 @@ class Plate
 
     makeStarCoord = if @starChart.coordinateSystem() == StarChart.EQUATORIAL then StarCoord.fromRaDec else StarCoord.fromGlatGlon
     xAxisDec = if @starChart.xAxis.unit == Axis.DEC || @starChart.xAxis.unit == Axis.DEC1950 || @starChart.xAxis.unit == Axis.GLAT then true else false
+    epoch1950 = if @starChart.xAxis.unit == Axis.DEC1950 || @starChart.xAxis.unit == Axis.RA1950 then true else false
     @fullValues xRange
     @fullValues yRange
     @coordCorners = [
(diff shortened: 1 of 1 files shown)


app/classifier/world_wide_telescope.cjsx
```diff
@@ -104,11 +104,26 @@ class Plate
 
     makeStarCoord = if @starChart.coordinateSystem() == StarChart.EQUATORIAL then StarCoord.fromRaDec else StarCoord.fromGlatGlon
     xAxisDec = if @starChart.xAxis.unit == Axis.DEC || @starChart.xAxis.unit == Axis.GLAT then true else false
+    @fullValues xRange
+    @fullValues yRange
     @coordCorners = [
       makeStarCoord(xRange[0].value, yRange[0].value, xAxisDec), makeStarCoord(xRange[1].value, yRange[0].value, xAxisDec),
       makeStarCoord(xRange[1].value, yRange[1].value, xAxisDec), makeStarCoord(xRange[0].value, yRange[1].value, xAxisDec)
     ]
```
amyrebecca: it's not really clear to me, except from the commit message, what this function is doing. it might need either a more descriptive name or a comment

Issue #2810, opened 2016-07-21
Stats page should only retrieve active workflows


camallen:
When loading project stats the worfklows api the request should have the active=true filter e.g. [here](https://github.com/zooniverse/Panoptes-Front-End/blob/1f94f898f3480f1659aa47604d3af2196eb26a9b/app/pages/project/home.cjsx#L24) to avoid loading testing workflows that aren't used.

PR #2811, opened 2016-07-21
get only active workflows for stats page

CKrawczyk:
This address #2810


camallen:
thanks for the quick fix @CKrawczyk

PR #2813, opened 2016-07-21
[WIP] New signed-in home page

brian-c:
# Do not merge.

I think the UI for this is basically ready to go. Still needs some data though.

- Loading up the ribbon takes a long time and makes a ton of requests. I'm thinking the best solution is to have a pseudo-resource like `Contributions` so we can do `user.get('contributions')` and end up with everything we need to render the ribbon (`{id, slug, name, owner, color, classifications}`, see get-user-ribbon-data.js). @marten might have thoughts.

- Clicking a project from the ribbon should take you to a user-specific projects stats page (see @jwbmartin's mockups), but I'm pretty sure we don't have any of that data available. For now they link directly to the project. Maybe this could reuse some of the project's stats-page data?

- The social media section will come from @parrish's home page PR. This should also be incorporated this into the router in the right place.

- Not super sure what'll go in the News pullout. Maybe a job for @mrniaboc?

Staged here: https://new-signed-in-home-page.pfe-preview.zooniverse.org/home-for-user

(Also, Just noticed the user's avatar and background image, fix up in a  moment.)


app/lib/get-all-project-preferences.js
```diff
@@ -0,0 +1,20 @@
+function getAllProjectPreferences(user, _page = 1, _accumulator = []) {
+  return user.get('project_preferences', {
```
marten: This call should specify `?include=project` which will fix the N-request issue on https://github.com/zooniverse/Panoptes-Front-End/pull/2813/files#diff-7ba34496843ed4dbf5a57e4c5dc0dd3aR6

app/lib/get-user-ribbon-data.js
```diff
@@ -0,0 +1,43 @@
...
+        return null;
+      });
+    }));
+
+    const awaitOwners = awaitProjects.then((projects) => {
+      return Promise.all(projects.map((project) => {
+        if (project === null) {
+          return null;
+        } else {
+          return project.get('owner').catch(() => {
```
marten: All you need the owner for later on is the `display_name`. This is already given to you in a project's `links` section.

marten:
So, turns out you don't need a new resource, just make sure to sideload the correct resources in the initial request. I've added comments on how to get rid of the additional requests.

One question is how this'll look for people who have contributed to tons of projects. Having more than 20 on that ribbon will probably look overcrowded? One option is to do a cutoff to the top-20 contributed by number of classification, but @wgranger and I prefer just showing the most recent 20 projects you've worked on.

srallen:
Gonna close this and open up a new WIP PR with the info of what's left that needs to be done.

Issue #2607, opened 2016-05-18
Usability problems with removing drawn boxes in a drawing task [User reported from Computer Vision: Serengeti]


inaspin:
I am noticing that the "X" function to remove the box placed around the animal is getting in the way - mostly with the ostriches, and particularly those that are further away from the camera. Any chance of getting it moved so it appears nearer the centre of the top line, or at the bottom RH corner where feet are likely to be obscured anyway by the ground cover?

Hope I did this right!    This refers originally to Subject 1983499

alexbfree:
Originally reported on Talk https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/talk/115/54036?comment=107388&page=1

Thanks for reporting this @inaspin !

srallen:
@wgranger this is related to the open PR for moving the drawing mark delete button.

wgranger:
Closed with #2816

Issue #296, opened 2015-05-07
the x to close the ellipse marker can be very close to/ on top of the handle to adjust its size


mschwamb:
I noticed a few times depending on how you draw it the ellipse tool x marker to delete it gets very close to or exactly on top one of the  handles to resize. I can see why you might not want the x to be in the center of the ellipse (the way it is on Planet Four) What about fixing the position of the x so it's always on the opposite side of the ellipse away from the marker and doesn't move?

![screen shot 2015-05-08 at 4 38 26 am](https://cloud.githubusercontent.com/assets/753325/7525454/2cbe76b8-f53c-11e4-89e0-a650799d690a.png)

![screen shot 2015-05-08 at 4 32 25 am](https://cloud.githubusercontent.com/assets/753325/7525461/3d0eb690-f53c-11e4-8369-eb1ec0f1c7da.png)

![screen shot 2015-05-08 at 4 33 39 am](https://cloud.githubusercontent.com/assets/753325/7525462/3d106080-f53c-11e4-927f-3b078e7594cf.png)

vrooje:
Related: for one of the projects we're hoping to launch with it's sometimes hard to draw the marks (lines) as required because the lines need to overlap and the close button is sometimes exactly where I need to draw the second line; or it's big enough, and the line small enough, that I can't move the line.

![screen shot 2015-05-22 at 12 22 08](https://cloud.githubusercontent.com/assets/2357126/7769682/f1b91aa8-007d-11e5-9551-39d23ffce906.jpg)

And here is one that was as close as I was able to get; the smaller line should intersect the large one at the center of each, and I would have liked the smaller one to be even shorter...

![screen shot 2015-05-22 at 12 17 03](https://cloud.githubusercontent.com/assets/2357126/7769702/2cbc4dfa-007e-11e5-933e-3626fecf8355.jpg)

This is also related to #204 - @brian-c I called this an enhancement before but I'm wondering if this is going to seriously affect my classifications for this project.

khufkens:
FYI: I have similar issues when drawing small lines in my "Jungle Rhythms" project (https://www.zooniverse.org/#/projects/khufkens/jungle-rhythms/classify). The X will obscure the feature needed to be marked. Furthermore, it does not serve any purpose at the time.

mschwamb:
[...]

PR #2816, opened 2016-07-25
Adjust delete button

### LINKED ISSUE #649: Hard to delete small drawing objects
This is feedback from our Season Spotter beta (see #443). When small shapes are drawn (e.g. ellipse tool), the "X" to delete them gets hidden behind one of the handles and is very difficult to click. This problem seems to be exacerbated on small screens like mobile devices.

wgranger:
Fixes #649 

This fix keeps delete buttons from being placed over drag handles for shapes. The buffer size (44 pixels) seemed appropriate and adhered to some guidelines for finger size. The buffer is a bit larger in the polygon because there is an additional outer circle along one handle.

Staged at: https://adjust-delete-button.pfe-preview.zooniverse.org/

### DIFF
--- a/app/classifier/drawing-tools/circle.cjsx
+++ b/app/classifier/drawing-tools/circle.cjsx
@@ -9,7 +9,7 @@ MINIMUM_RADIUS = 5
 GUIDE_WIDTH = 1
 GUIDE_DASH = [4, 4]
 DELETE_BUTTON_ANGLE = 45
-BUFFER = 12
+BUFFER = 18
 
 module.exports = React.createClass
   displayName: 'EllipseTool'
@@ -48,14 +48,8 @@ module.exports = React.createClass
 
   getDeletePosition: ->
     theta = (DELETE_BUTTON_ANGLE - @props.mark.angle) * (Math.PI / 180)
-    x = @props.mark.r * Math.cos theta
-    y = -1 * @props.mark.r * Math.sin theta
-    x += (BUFFER / @props.scale.horizontal) if @calculateDistance(x, @props.mark.r, y, 0) < BUFFER / @props.scale.horizontal
-    x: x
-    y: y
-
-  calculateDistance: (x1, x2, y1, y2) ->
-    Math.sqrt(Math.pow(x1 - x2, 2) + Math.pow(y1 - y2, 2))
+    x: (@props.mark.r + (BUFFER / @props.scale.horizontal)) * Math.cos theta
+    y: -1 * (@props.mark.r + (BUFFER / @props.scale.vertical)) * Math.sin theta
 
   render: ->
     positionAndRotate = "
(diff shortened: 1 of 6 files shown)


srallen:
Is there a reason why we are calculating a position for the delete button and not just positioning just outside of the mark in a consistent spot? Depending on the tool and how you draw the mark, you can end up with some awkward positions for the button:

screen shot 2016-07-25 at 2 22 29 pm
screen shot 2016-07-25 at 2 27 37 pm
screen shot 2016-07-25 at 2 26 54 pm

Both the small version of the point tool and the column rectangle just position the button right outside of the mark in the same spot.

srallen:
The small version of the point tool needs to be added to the dev classifier. You can see it here: https://adjust-delete-button.pfe-preview.zooniverse.org/projects/srallen086/srallen-test/classify

wgranger:
We could probably place the delete button outside a consistent spot for the rectangle, circle, and ellipse tool, since their shapes and handles are more predictable. If we place the button right outside a rectangle, for example, and somebody draws a rectangle over the entire screen, isn't there a chance the delete button will be outside the image itself?

The polygon, line, and bezier tools have multiple handles so placing the delete button outside one handler doesn't necessarily ensure it won't be covered up by another handler. However, I don't mind going back over some of this code to try to make it more succinct, if you think it could be done. This was a fun problem to work on.

srallen:
Right, I don't think the same position should be applied to all and would depend on the type of tool. I think really what was catching me was the bouncing around that I saw happening. But yes, there's a chance the delete button will go off screen and this chance exists now if someone happens to draw a mark that way or drag it partially off the area. Perhaps we can chat more tomorrow when I'm in the office?

mschwamb:
Also will close #296

wgranger:
@srallen Alright, these commits should make the code more succinct. I also updated the staged branch.

srallen:
This is much better generally for the UI, imo. The line is still a bit awkward, but not sure what a better solution would be. I don't think I'll have time to do a thorough review in all of the browsers and on mobile before I go out of town on Thursday. @eatyourgreens would you have time to give this a review too? Would it be a good idea to bring in a couple of project owners that use drawing tools to check if this change works?

eatyourgreens:
@srallen I think it would be a good idea to get some people to have a look at their drawing projects on https://adjust-delete-button.pfe-preview.zooniverse.org/?env=production and see if this works for them.

I can check this Edge but not IE at the moment.

wgranger:
fixes #2607

srallen:
I think @mcbouslog has an IE vm setup. Otherwise I can take a look again on Tuesday.

simensta:
I am happy with this modification of the delete button position. This branch works for me on an IE11 vm. 

Although I think the delete button position for the line tool in this PR works, it does feel a bit awkward. It might help to always position the button opposite to the side of the beginning point from which the line extends. That said, I think experimenting with a different delete button position for the line tool can be handled in a separate PR.

wgranger:
@simensta I adjusted the line tool based on your suggestion, and the staged branch has been updated.

srallen:
I think that works better for the line tool.

simensta:
@wgranger, the delete button position on the line tool looks good -- thanks!

Issue #2819, opened 2016-07-26
all workflows not populating in project builder


mcbouslog:
When a project's workflows are [collected](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/get-workflows-in-order.coffee) (in the [Project Builder](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/project.cjsx#L94), on the project's [home](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/home.cjsx#L24) or on the [stats page](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/stats/index.cjsx#L15)) the API response is limited to 20 items. 

NfN now has over 20 active workflows, so in the Project Builder (or home or stats page) a new workflow will not show up as an option, because the API response is limited to 20 items.

A temporary solution is to increase the `page_size` in the API call to 50.

PR #2820, opened 2016-07-26
increase page_size in workflows request to API

mcbouslog:
Temporary solution to issue #2819.

API workflow response limited to 20 items, NfN has over 20 active workflows, so new workflows not showing up in Project Builder.

Temporary solution is to increase page_size param from 20 (default) to 50.

### DIFF
--- a/app/lib/get-workflows-in-order.coffee
+++ b/app/lib/get-workflows-in-order.coffee
@@ -1,5 +1,7 @@
-getWorkflowsInOrder = (project, query) ->
+getWorkflowsInOrder = (project, query = {}) ->
   order = project.configuration?.workflow_order ? []
+  unless query?.page_size?
+    query['page_size'] = 50
 
   project.get('workflows', query).then (workflows) ->
     workflowsByID = {}
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -620,7 +620,7 @@ module.exports = React.createClass
       workflowID: ''
 
   render: ->
-    <PromiseRenderer promise={apiClient.type('workflows').get @props.params.workflowID}>{(workflow) =>
+    <PromiseRenderer promise={apiClient.type('workflows').get @props.params.workflowID, {}}>{(workflow) =>
       <ChangeListener target={workflow}>{=>
         <EditWorkflowPage {...@props} workflow={workflow} />
       }</ChangeListener>
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -11,6 +11,7 @@ FinishedBanner = require './finished-banner'
 Classifier = require '../../classifier'
 seenThisSession = require '../../lib/seen-this-session'
 MiniCourse = require '../../lib/mini-course'
+getWorkflowsInOrder = require '../../lib/get-workflows-in-order'
 
 FAILED_CLASSIFICATION_QUEUE_NAME = 'failed-classifications'
 
@@ -111,7 +112,7 @@ module.exports = React.createClass
       currentWorkflowForProject[props.project.id]
 
   getRandomWorkflow: (project) ->
-    project.get('workflows', active: true).then (workflows) =>
+    getWorkflowsInOrder(project, active: true).then (workflows) =>
       if workflows.length is 0
         throw new Error "No workflows for project #{project.id}"
         project.uncacheLink 'workflows'
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -112,7 +112,7 @@ module.exports = React.createClass
       currentWorkflowForProject[props.project.id]
 
   getRandomWorkflow: (project) ->
-    getWorkflowsInOrder(project, active: true).then (workflows) =>
+    getWorkflowsInOrder(project, {active: true, fields: 'finished_at'}).then (workflows) =>
       if workflows.length is 0
         throw new Error "No workflows for project #{project.id}"
         project.uncacheLink 'workflows'
@@ -121,7 +121,7 @@ module.exports = React.createClass
         @setState {projectIsComplete}
         randomIndex = Math.floor Math.random() * workflows.length
         # console.log 'Chose random workflow', workflows[randomIndex].id
-        workflows[randomIndex]
+        @getWorkflow project, workflows[randomIndex].id
 
   createNewClassification: (project, workflow) ->
     @setState {workflow}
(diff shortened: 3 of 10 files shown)


app/lib/get-workflows-in-order.coffee
```diff
@@ -1,5 +1,7 @@
-getWorkflowsInOrder = (project, query) ->
+getWorkflowsInOrder = (project, query = {}) ->
   order = project.configuration?.workflow_order ? []
+  unless query?.page_size?
```
camallen: Not a blocker for me but can't you just set this query param where you need to call it ala [here](https://github.com/zooniverse/Panoptes-Front-End/blob/1f94f898f3480f1659aa47604d3af2196eb26a9b/app/pages/project/home.cjsx#L24)

app/pages/project/finished-banner.cjsx
```diff
@@ -33,7 +34,7 @@ module.exports = React.createClass
     if project.redirect
       Promise.resolve false
     else
-      project.get('workflows').then (allWorkflows) ->
+      getWorkflowsInOrder(project).then (allWorkflows) ->
```
camallen: 20 workflows here will be a really slow call....are they really all active and unfinished?

mcbouslog:
@camallen - great point, and I'm totally fine changing accordingly, but it seems like all the workflows requests (except ones requesting a single workflow by including workflow id) would want all workflows? In which case my thinking was instead of adding the `page_size: 50` query to all the requests, add to the component, but I'm admittedly ignorant to any negative consequences this may have to the API, please let me know!

Just for reference, I think the following are all the workflows requests:
1. project home page
2. project classify page (for selecting random workflow)
3. project builder - workflows list (to select workflow to edit)
4. project builder - visibility section (for selecting which workflows are active)
5. stats page
6. admin project status page
7. finished banner component

mcbouslog:
And all workflows presumes there are less than 50.

A long term solution would be to load the first 20, then have a "Load more..." button or something to load additional workflows as needed.

app/pages/project/finished-banner.cjsx
mcbouslog: yes i should add a `active: true` query at minimum?

camallen:
right - we can modify the default page size for workflows on the api if needed. 

Remember the really slow workflows load times for the NFN landing page? So wherever we are using this we will suffer from the slow req (how many active workflows are they expecting?). If we can we should identify which calls actually need a full workflow representation and just use the fields param (see nfn landing page call) to create fast api responses as they mostly just need ids for some of those routes you've got (maybe more).

mcbouslog:
Ah yes the fields param! Great idea, I'll update accordingly.

mcbouslog:
ok i think the various workflows API calls have been properly updated with the `fields` param, except for the [stats index page](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/stats/index.cjsx#L15), where it should be:
```
getWorkflowsInOrder @props.project, {active: true, fields: 'classifications_count,completeness,display_name,retired_set_member_subjects_count,retirement,subjects_count'}
```
but i can't get `subjects_count` back from the API, and i'm not sure why (all the others come back ok). the stats page workflows response is limited by `active: true` though, i think good enough for now? but let me know if you can tell why no luck with `subjects_count`.

i'm excited to see how much this will help with the workflows load-time on NfN! let me know what you guys think. thanks,

camallen:
@mcbouslog can you get this deployed for testing?

mcbouslog:
no prob - [here's a staged branch](https://inc-workflow-response-page-size.pfe-preview.zooniverse.org/).

i have a project on preview i could add you both as collaborators on if you'd like, just let me know your preview usernames.

camallen:
the stats page is failing due to the api not respecting a virtual field on the workfow (`subjects_count`) [here](https://github.com/zooniverse/Panoptes/blob/master/app/controllers/api/v1/workflows_controller.rb#L58), raise issue here https://github.com/zooniverse/Panoptes/issues/1881

camallen:
@mcbouslog i've got a PR up for the virtual attribute via fields param. Once that is merged we should test again. I've played around locally and this is working for me but it needs more eyes / testing as it could break things pretty quick. 

Also i still don't like the page param being added in the get workflows in order component but I agree it could get a bit messy the other way.

mcbouslog:
sounds good @camallen, i'll update the stats page workflows call once the related Panoptes PR is merged.

agreed regarding additional review, i'll see if i can pull one more person in. an error with how workflows are loaded would be very bad.

also agree regarding the `page_size` param, and open to suggestion, but hopefully it's ok for now, until i can implement a paging solution.

mcbouslog:
stats page workflow request updated, and staged branch also updated accordingly.

i've roped @srallen in as well.

let me know what you think. thanks!

camallen:
the stats pages are working just fine now, https://inc-workflow-response-page-size.pfe-preview.zooniverse.org/projects/zooniverse/notes-from-nature/stats?env=production

app/pages/lab/workflow.cjsx
```diff
@@ -620,7 +620,7 @@ module.exports = React.createClass
       workflowID: ''
 
   render: ->
-    <PromiseRenderer promise={apiClient.type('workflows').get @props.params.workflowID}>{(workflow) =>
+    <PromiseRenderer promise={apiClient.type('workflows').get @props.params.workflowID, {}}>{(workflow) =>
```
srallen: Is this change intentional?

app/pages/project/classify.cjsx
```diff
@@ -120,7 +121,7 @@ module.exports = React.createClass
         @setState {projectIsComplete}
         randomIndex = Math.floor Math.random() * workflows.length
         # console.log 'Chose random workflow', workflows[randomIndex].id
-        workflows[randomIndex]
+        @getWorkflow project, workflows[randomIndex].id
```
srallen: This change is not necessary. All the workflows were already fetched and this just needs to return a random one out of the response array. It doesn't need to be fetched again.

srallen: Ah I see the addition of `fields` in the query above, so you'd have to fetch the one you want to get back the whole resource. Ignore my comment.

app/lib/get-workflows-in-order.coffee
srallen: Could a comment be left to take the `page_size` query param out of this once pagination is properly implemented?

srallen:
FYI, none of my comments are blockers, so @mcbouslog when you get back, I'll be out of town and if it all looks good to the others, this can go ahead and merge, imo.

PR #2822, opened 2016-07-26
Remove extra styling on nav "Build a project" link

parrish:
As per @jwbmartin in #2706 [here](https://github.com/zooniverse/Panoptes-Front-End/pull/2706#issuecomment-235339595)


jwbmartin:
🙏

srallen:
The admin navbar link was using that class too, so the span and class name needs to removed there too: https://github.com/parrish/Panoptes-Front-End/blob/a5523bbad5a3181e8c17189e2cba1059443bc01d/app/layout/site-nav.jsx#L146-L148

parrish:
Good catch!

Issue #2825, opened 2016-07-26
UX for notifications in nav


srallen:
Work with @jwbmartin to integrate her design and UX for notifications in the navbar. Possibly use an icon [like this](https://thenounproject.com/search/?q=notification&i=492155) next to the mail icon.

mschwamb:
Please don't hide the notifications icon away from the project level. It's something that is used frequency by active talk volunteers ,the moderators, and the research teams to see if there is something on the project level talk that they have been mentioned in or updates to a thread they are following it gets updated with the notification count and currently tells the user there are new notifications.  

 If you hide the notifications button away at the top Zooniverse then someone has to always actively click the zooniverse navigation bar to see if they have project notifications and then no one will use it. 

Honestly having a notification icon, mailbox and login/logout button in the top right of the project level would be extremely helpful, and save the now multiple clicks that need to be done to go from logging in to classifying. see #2787 I'm still not sure why the account buttons had to be hidden in the zooniverse level and can't be like what github does and have them in the top right with the notifications.

camallen:
Seconding @mschwamb concerns about the hidden navigation. As a site user i find it really strange that my personal area is hidden behind the project view. I'm totally in favour of just having the project view embedded all the time and have the top nav always visible, e.g.

![image](https://cloud.githubusercontent.com/assets/295329/17172285/327a76dc-53ec-11e6-920a-99f77536ce33.png)

mschwamb:
agreeing with @camallen 100% on this. It's very clear the navigation of the Zooniverse level is different with the blue bar but is still visible in Cam's suggestion. Science team members have asked me how to log out as it wasn't obvious to hit the Zooniverse icon. Also now when you log in, it takes you a click after logging in to get back into then project mode, which is confusing and extra work for the user.

srallen:
If anyone has a bit of time for this, feel free to also self-assign the issue.

alexbfree:
related: https://github.com/zooniverse/Panoptes-Front-End/issues/2858

mschwamb:
one note - there are two sets of notifications zoonvierse wide and project level so perhaps there 's one notifications icons (one on the Zooniverse bar) and one on the project bar?

Issue #2800, opened 2016-07-18
Not clear how to get the Zooniverse level navigation bar and Account buttons back 


mschwamb:
It's not apparent that the Zooniverse logo does anything to bring back the Zooniverse level navigation bar. So once you double click or login, I think it's not clear how to log out or get to the mailbox. I've been asked by other science team members, how to navigate to access those things.  See additional comments on this subject  in #2733

srallen:
@jwbmartin One of things we did on Chimp and See with buttons that only use icons is add a tiny label below them, specifically the field guide and favorite button on the sides of the classifier. The favorite button also has a small pop up on hover when someone is signed out and the button is disabled to show a small "sign in or sign up to use this" type messaging. Could something like that be a solution here?

PR #2831, opened 2016-07-27
Remove the main layout overlay

### LINKED ISSUE #2787: A way of removing the extra click for non-logged in users on the landing page
The new Zooniverse navigation header layout now reveals the zoom out with the Zooniverse navigation bar if you're not logged in the first time.  I really appreciate that panoptes front end now shows that  when you first get to the first project page to help get people to login  but  because of the  zoomout you have to click once to zoom in and hide the bar and the blue sides and then once to click Get Started!  I find that a bit weird and I think new people will find it weird when they try to click on the Get Started! button that it doesn't just go. As the goal is get people in fast who are checking out the site, is there something that can be done to make it one click?

What about removing the blue sides so it's just the top blue navigation bar appearing and disappearing, then it wouldn't be a zoomout, zoomin which might save the click?

eatyourgreens:
Fixes #2787

### DIFF
--- a/app/layout/index.jsx
+++ b/app/layout/index.jsx
@@ -126,15 +126,6 @@ const AppLayout = React.createClass({
             } : null}
           >
             {this.props.children}
-            <div
-              className={classnames('app-layout__main-overlay', {
-                'app-layout__main-overlay--active': demotedHeaderIsRevealed,
-              })}
-              style={{
-                display: demotedHeaderIsRevealed ? 'block' : 'none',
-              }}
-              onClick={this.togglePrimaryNav}
-            ></div>
           </div>
 
           <footer className="app-layout__footer">
--- a/app/layout/index.jsx
+++ b/app/layout/index.jsx
@@ -47,10 +47,12 @@ const AppLayout = React.createClass({
     };
   },
 
-  handleNavigation() {
-    this.setState({
-      siteHeaderRevealed: false,
-    });
+  hidePrimaryNav() {
+    if (this.state.siteHeaderRevealed) {
+      this.setState({
+        siteHeaderRevealed: false,
+      });
+    }
   },
 
   matchWindowScale() {
@@ -116,6 +118,7 @@ const AppLayout = React.createClass({
                 scale(${this.state.mainContentScale})
               `,
             } : null}
+            onClick={this.hidePrimaryNav}
           >
             {this.props.children}
           </div>
--- a/app/layout/index.jsx
+++ b/app/layout/index.jsx
@@ -47,14 +47,6 @@ const AppLayout = React.createClass({
     };
   },
 
-  componentDidMount() {
-    addEventListener('locationchange', this.handleNavigation);
-  },
-
-  componentWillUnmount() {
-    removeEventListener('locationchange', this.handleNavigation);
-  },
-
   handleNavigation() {
     this.setState({
       siteHeaderRevealed: false,
--- a/app/layout/index.styl
+++ b/app/layout/index.styl
@@ -59,14 +59,6 @@
     &--set-aside
       opacity: 0.9
 
-  &__main-overlay
-    bottom: 0
-    left: 0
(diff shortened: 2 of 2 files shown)


eatyourgreens:
Staged at https://delete-layout-overlay.pfe-preview.zooniverse.org/

chrislintott:
Do not merge without talking to me. #

eatyourgreens:
Looks like clicking on a project page zooms it out to full size, except on the Classify page. I'm not entirely sure why that's happening, but the behaviour should be consistent.

srallen:
Seems that there's an issue with drawing tools or at least the point tool. When zoomed out, the marks are off. Try it out here: https://delete-layout-overlay.pfe-preview.zooniverse.org/projects/srallen086/srallen-test/classify?workflow=1817

Navigating between pages zooms back in. If someone chooses to stay zoomed out, it should be consistent between pages.

mschwamb:
Also if possible can the subject size not change with the zoom out. The main goal of these sites is to review and classify the image/data presented so from a science team perspective if there's a way to keep the image size regardless of whether you have the blue border visible to see the account info or whether you hide it, I would greatly appreciate that.  


Subject image when the border on the left and right aren't showing

screen shot 2016-07-27 at 5 39 28 am

same subject image when the border showing (it's slightly smaller)
screen shot 2016-07-27 at 5 39 21 am

eatyourgreens:
@srallen I've removed the listener that zoomed in when you navigated between views. Is that any better? Worse?

Looks like the CSS transform that's used to rescale the page messes up the calculation of subject image scale for the drawing tools. Fixing the image size might be tricky with flex display, since we essentially let the browser determine the image size, based on the user's display size.

I'm marking this 'Do not merge' because of the drawing tool bug.

eatyourgreens:
@srallen I've confirmed that drawing tool bug on the live site too: #2836

jwbmartin:
Since Brooke has complained that people can't figure out that the logo is the menu we could just add the word menu beneath it 

Example:
screen shot 2016-07-29 at 10 22 12 am

jwbmartin:
The screen was supposed to zoom out a bit because I really wasn't imagining people having the navigation open and trying to classify at the same time since the whole goal was to reduce the clutter.

mschwamb:
@jwbmartin  adding in 'menu' or something that indicates that you can click the zooniverse icon to logout helps with #2800 but if you're going to add in the notifications as an icon and then hide it on the blue bar isn't helpful for the reasons I've argued in  #2825  

The reason I'm harping on this is because the way the original design was, it was not clear or apparent that you can login with /register for a Zooniverse account to classify.  It promoted classifying not-logged in and research teams get way more information about a person if they login (we can track them better for weighting schemes and such and we get their email for newsletters).  This is why I opened #2733 Right now I'm generating gold standard classification data for Planet Four: Terrains and with  I have to be logged in so I can pull out the classifications later. With the current setup, I have to sign in and then always double  check again I'm logged in and my avatar is showingbecause there's now nothing on the project page that indicates I'm logged in other the addition of the Notifications link.  

I get why you might want to hide the Zooniverse navigation, but I don't get why the account profile/logout/login and mailbox, the proposed notifications icon  buttons can't be moved  next to the Zooniverse icon at  the project level.  That would be the simplest solution if you want to keep the Zooniverse navigation bar hidden. Honestly with the blue bar you've setup, it now for me clearly differentiates the two navigation menus, but if you want to still hide the Zooniverse menu that's fine by me as long as there's an obvious way to register, login/logout and access your notifications.

eatyourgreens:
Rebased to remove merge conflicts, and taking off the "do not merge" now that #2836 is fixed.

srallen:
PR LGTM for resolving #2787. Will open separate issue for adding 'menu' to the project nav zooniverse logo.

jwbmartin:
I'm still trying to work out so sort of solution for these problems so sit tight

PR #2835, opened 2016-07-29
add in help line for results page

mschwamb:
add in help line for results page


app/pages/lab/about/results.cjsx
```diff
@@ -1,4 +1,4 @@
-React = require 'react'
+eact = require 'react'
```
srallen: This should be `React`.

mschwamb:
Thanks @srallen. I missed that accidental deletion when looking everything over. It should be fixed now.

srallen:
Thanks!

PR #2837, opened 2016-07-29
Get subject image scale dynamically

### LINKED ISSUE #2836: New layout has broken the drawing tools
The SVG drawing surface calculates image scale on component mount, before any rescaling that happens as a result of the new project layout. This can result in SVG marks being offset significantly from the position of the mouse cursor.

To reproduce: 

1. open any project classify with a drawing task while zoomed out (eg. while not logged-in to Zooniverse.)
2. Click to zoom in.
3. Click on the subject image to draw.

drawing scale bug

eatyourgreens:
Get the rendered subject image dimensions from the DOM when needed, rather than relying on stored data in component state, which may be out of synch with the rendered DOM.
Fixes #2836

### DIFF
--- a/app/classifier/frame-annotator.cjsx
+++ b/app/classifier/frame-annotator.cjsx
@@ -22,23 +22,11 @@ module.exports = React.createClass
     onChange: Function.prototype
 
   getInitialState: ->
-    sizeRect: null
     alreadySeen: false
     showWarning: false
 
   componentDidMount: ->
-    addEventListener 'resize', @updateSize
-    @updateSize()
     @setState alreadySeen: @props.subject.already_seen or seenThisSession.check @props.workflow, @props.subject
-
-  componentDidUpdate: (prevProps, prevState)->
-    # If size of the frame image or viewBoxDimensions has changed, update our sizing information (used for scaling/translating annotations)
-    if @props.naturalWidth isnt prevProps.naturalWidth or @props.naturalHeight isnt prevProps.naturalHeight or @props.viewBoxDimensions isnt prevProps.viewBoxDimensions
-      setTimeout =>
-        @updateSize()
-
-  componentWillUnmount: ->
-    removeEventListener 'resize', @updateSize
 
   componentWillReceiveProps: (nextProps) ->
     if nextProps.annotation isnt @props.annotation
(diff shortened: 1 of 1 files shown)


eatyourgreens:
This should fix the drawing tool problem in #2836 but it's quite a big change to how drawing annotations are calculated, so needs a few eyes on it to check.

eatyourgreens:
Staged at https://fix-2836.pfe-preview.zooniverse.org/

srallen:
Still trying to figure this one out, but I'm seeing the following errors in console:
`Uncaught (in promise) TypeError: Cannot read property 'left' of undefined`
`Uncaught (in promise) TypeError: Cannot read property '_currentElement' of null`
But only with the deployed staging branch. I'm not seeing it when running locally. The first is in the `FrameAnnotator`, the second is I think the `geordiLogger` not being set for the app, which may be what's breaking the `FrameAnnotator`?

eatyourgreens:
My fault. This change catches that error.
https://github.com/zooniverse/Panoptes-Front-End/pull/2837/files#diff-a8671c6b3d2d9b11be38f376f0947342R44
I hadn't deployed the latest version of this branch. Should be fixed now.

srallen:
Ah ok, that seems to have fixed it.

simensta:
I am seeing the following new warning: `Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of 'ChangeListener'.` Still tracking down why that would be...

eatyourgreens:
Here maybe, but I'm not sure?
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L109

srallen:
I've seen that warning before, so I don't think it's coming in with this PR. I'm not actually sure where it's happening either, but that's a good guess.

simensta:
Okay, so that key-prop error shouldn't hold up this PR.

srallen:
Yeah, it should be fixed, but I don't think it needs to come in with this PR. The `ChangeListener` and `PromiseRenderer` should be replaced if possible. See the [readme](https://github.com/zooniverse/Panoptes-Front-End#async-helper-components) about why

srallen:
This LGTM, but want another ok before merging.

parrish:
Looks good!

PR #2839, opened 2016-08-01
More team page updates

srallen:
More updates for the team page. Fixing broken image in production. Making sure the case on the file extension is consistent seems to fix it.


camallen:
LGTM

Issue #2834, opened 2016-07-28
Space Warps: Instant feedback messages for training images


aprajita:
Hi @chrislintott @marten  

just putting the PFE feedback messaging in SpaceWarps into an issue.

Pop-up feedback is an essential and necessary part of SpaceWarps that improved the users experience & training, and therefore skill.

I've updated our workflow
https://www.zooniverse.org/projects/aprajita/swdemo/classify?reload=0&workflow=960
with a dataset that includes "training" images. 

I've provided feedback messages both for success and failure as part of the metadata (i.e. in the manifest)

The manifest has they keywords
subject_id  image_name_1    image_name_2    image_name_3        #Type   #X  #Y  #Tol    #F_Success  #F_Fail Provenance  Col Size-in-arcsec  #X2 #Y2 #Tol2

If "#Type" is SIM or DUD then these are training images

In the case of SIM training images any click within #x,#y +/- #Tol (or #x2,#y2 +/- #Tol2) would be a success, and no click in this region would be a fail.

For the DUD no click would be a success, and a click would be a fail

The feedback messages for each are in #F_Success and #F_Fail

I've split how we could handle the instant feedback as essential, desirable and optimal. If at least Essential could be implemented that would allow us to start.

ESSENTIAL
If there is a click anywhere in a subject that is a SIM or no click on a DUD triggers a success message. No click in a SIM and a click in a DUD triggers a fail message. 
The message appears anywhere in the interface.

DESIRABLE
If there is a click within #x,#y+/-#Tol in a subject that is a SIM or no click on a DUD triggers a success message. No click or a click ouside #x,#y+/-#Tol in a SIM and a click in a DUD triggers a fail message. 
The message appears anywhere in the interface.

OPTIMAL
If there is a click within #x,#y+/-#Tol in a subject that is a SIM or no click on a DUD triggers a success message. No click or a click outside #x,#y+/-#Tol in a SIM and a click in a DUD triggers a fail message. 
The pop-up message appears pointing towards #x,#y

OPTIMAL is most like SpaceWarpsI&II but ESSENTIAL or DESIRABLE would be fine for our initial planned searches that will be centred on a lensed target.

[...]

PR #2840, opened 2016-08-01
Added functionality to support responses to "dud" and "sim" subjects

hughdickinson:
…specified using metadata in the manifest. Also provided functionality to validate the spatial click coordinates of a "sim" classification based on X, Y and tolerance values, also provided as metadata in a manifest.
# Review Checklist
- [ ] Does it work on mobile?
## Optional


camallen:
staged at https://hughdickinson-metadata-based-feedback.pfe-preview.zooniverse.org/

camallen:
@hughdickinson what's your testing project url?

hughdickinson:
A project that demonstrates the new functionality can be found [here](https://hughdickinson-metadata-based-feedback.pfe-preview.zooniverse.org/projects/hughdickinson/testing-metadata-based-feedback).

srallen:
Looks like maybe a bad rebase happened. Some grid tool functionality looks like it was removed when it shouldn't have been. L20 and L409-413 need to be added back in.

Are we ok with using `#` as a part of the metadata header to id the messages? We're telling project builders in the UI of the subject uploader that any header using `#` stays hidden from volunteers.

Also, I think this may a good opportunity to rework the summary part of the classifier and this makes sense to me as being in its own component rather than adding more to the classifier index.cjsx file. If you're up for it, ES6 too! But I don't know what the timeline for this may be, so if the deadline is strict, cjsx is ok for now too.

app/classifier/index.cjsx
```diff
@@ -268,10 +267,46 @@ Classifier = React.createClass
...
+        subjectYPos = parseFloat(@props.subject.metadata['#Y1'])
+        subjectTol = parseFloat(@props.subject.metadata['#Tol1'])
+        userXPos = @props.classification.annotations[0]?.value[0]?.x
+        userYPos = @props.classification.annotations[0]?.value[0]?.y
+        isWithinTolerance = @withinTolerance(userXPos, userYPos, subjectXPos, subjectYPos, subjectTol)
+
+        console.log(subjectClass)
+        console.log(userMadeAnnotation)
+        console.log(@props.classification.annotations?.length)
+        console.log(@props.classification.annotations[0])
```
srallen: The `console.log`s should be removed.

srallen:
Are these messages specific to a single project? If so, we should consider what we're doing on [L310](https://github.com/hughdickinson/Panoptes-Front-End/blob/f5ad7c591306a988832a528a07a565a73594c198/app/classifier/index.cjsx#L310) to make sure this doesn't break another project's summary that doesn't have that kind of subject metadata. Though, I'm not convinced this kind of messaging should live in the subject metadata to start.

camallen:
see #2834 for more details on what this is trying to achieve. 

Also the subject metadata is the ideal place to hold the ref for the summary information. It's subject specific information that can be re-used across projects. However hard coding language specific strings in there won't work. We would probably have to add workflow content references in there or something of the like for translations on this content to be supported (not that we currently are doing translations properly).

app/classifier/index.cjsx
```diff
@@ -17,7 +17,7 @@ workflowAllowsFlipbook = require '../lib/workflow-allows-flipbook'
 workflowAllowsSeparateFrames = require '../lib/workflow-allows-separate-frames'
 WorldWideTelescope = require './world_wide_telescope'
 MiniCourseButton = require './mini-course-button'
-GridTool = require './drawing-tools/grid'
```
eatyourgreens: Is this line supposed to be deleted?

app/classifier/index.cjsx
```diff
@@ -406,11 +425,6 @@ Classifier = React.createClass
     @props.classification.update 'annotations'
 
   completeClassification: ->
-    currentAnnotation = @props.classification.annotations[@props.classification.annotations.length - 1]
-    currentTask = @props.workflow.tasks[currentAnnotation?.task]
-    currentTask?.tools?.map (tool) =>
-      if tool.type is 'grid'
-        GridTool.mapCells @props.classification.annotations
```
eatyourgreens: Are these lines supposed to be deleted too?

app/classifier/index.cjsx
```diff
@@ -268,10 +268,29 @@ Classifier = React.createClass
         </p>}
     </div>
 
+  withinTolerance: (userX, userY, metaX, metaY, tolerance) ->
+    distance = Math.sqrt((userY - metaY)**2 + (userX - metaX)**2)
+    isWithinTolerance = distance < tolerance
+    return isWithinTolerance
+
```
eatyourgreens: Looks like this has moved into the feedback component, so it isn't needed here any more?

hughdickinson:
@eatyourgreens You are correct about the last. Not sure how the others got changed, but I didn't do it deliberately.

app/classifier/tasks/metadata-based-feedback.cjsx
```diff
@@ -0,0 +1,69 @@
+React = require 'react'
+
+module.exports = React.createClass
+  displayName: 'MetadataBasedFeedback'
+
+  withinTolerance: (userX, userY, metaX, metaY, tolerance) ->
+    distance = Math.sqrt((userY - metaY)**2 + (userX - metaX)**2)
+    isWithinTolerance = distance < tolerance
+    return isWithinTolerance
```
srallen: This is just an fyi about coffeescript: returns are implicit, so this line can just be `isWithinTolerance`

app/classifier/tasks/metadata-based-feedback.cjsx
```diff
@@ -0,0 +1,69 @@
...
+
+  withinAnyTolerance: (userX, userY) ->
+    for key, value of @props.subject?.metadata
+      if key.indexOf(@props.metaSimCoordXPattern) == 0
```
srallen: Generally in javascript, `===` should be used. In coffeescript, `is` can also be used. [Here's a good summary about why](http://stackoverflow.com/questions/359494/which-equals-operator-vs-should-be-used-in-javascript-comparisons)

srallen:
We need confirmation that the same success and failure messages for sims/duds will be used for Space Warps in #2834 to avoid redundancy in the subject metadata like we discussed.

hughdickinson:
Thanks for all the code review. I plan to implement the suggestions when I have time. This may take a couple of months though.

hughdickinson:
Specific coding suggestions have been implemented and tested. Ready for another review depending upon whether the whole approach has been deemed inappropriate in the interim.

srallen:
@hughdickinson Could you rebase to resolve the merge conflicts? @eatyourgreens do you have time to review this?

app/classifier/index.cjsx
eatyourgreens: I'm still wondering why these lines have been deleted. Would a rebase fix this?

srallen:
@hughdickinson Could you set this PR to allow changes to be made by the maintainers? I believe you should see this option in the right sidebar of this PR. This will allow us to finish up the conflict resolution and rebase before merging.

srallen:
Superceded by #3202.

PR #2844, opened 2016-08-02
Prevent undefined ref from disrupting navigation.

### LINKED ISSUE #2830: Zooniverse toggle button rendering twice in Chrome
I don't know how I managed to trigger this, but the toggle button, top right, is rendering twice in the DOM in Chrome.

image

parrish:
Closes #2830

Oddly enough, it was an undefined component ref that was causing #2830. 

To reproduce the original issue

double-icon

And the fixed version:

fixed-icon


parrish:
Deployed at https://double-zooniverse-icon.pfe-preview.zooniverse.org/

PR #2845, opened 2016-08-02
Fix logged-in homepage flex box in IE

### LINKED ISSUE #2829: Home page is pushed up in IE11
This is due to a bug in the `.hero` class. `Flex` doesn't work well with the `min-height` property. Setting `height: 800px` is a quick fix, but there will likely be overflow in mobile or other viewports.

screen shot 2016-07-26 at 4 08 18 pm

parrish:
Closes #2829 

This also
  - removes the not-logged-in conditions from the logged-in page.
  - moves `.flex-container` to `layout.styl` (I imagine it's more generally useful)


parrish:
Deployed at https://fix-ie-flexbox.pfe-preview.zooniverse.org/

Issue #2817, opened 2016-07-26
Arrow Drawing Task


hughdickinson:
It would be really useful to be have an arrow drawing tool to enable directionality of lines to be explicitly defined. Is this possible?

camallen:
can't this already be determined by the order of the points in the classification as we record the order of points already? E.g. `[ point_1, point_2 ]` is a direction from 1 to 2.

hughdickinson:
It can, but for the purposes of our project, the directionality is sufficiently important that it would be good to have a visual representation so that classifiers do not inadvertently draw the line in the wrong direction.

jelliotartz:
Closing as this has not received comments or action for over a year.

PR #2851, opened 2016-08-03
Added new components for "arrow", "mars-arrow", and "compass" drawing tools

hughdickinson:
The _arrow_ tool is just like a line with an arrowhead to explicitly indicate a direction to the user.

The _mars-arrow_ tool includes a circle at the midpoint of the arrow line and can be used to imply an approximate location as well as a direction.

The _compass_ tool adds a perpendicular to the arrow line at the line midpoint to help with more precise positioning of the arrow.

Tool icons have also been provided.
# Review Checklist
- [ ] Does it work on mobile?
## Optional


eatyourgreens:
If the only difference is in how the arrow renders, we should consolidate these into a single arrow tool component with an options menu to choose which arrow is rendered. Something along the lines of the size options that are used with the point tool to choose whether it renders as a small or large circle.

hughdickinson:
An example test project that uses the arrow tools is [here](https://new-arrow-tools.pfe-preview.zooniverse.org/projects/hughdickinson/testing-new-arrows).

CKrawczyk:
You might want to adjust the position of the delete button so it is fixed during drawing to keep it in line with https://github.com/zooniverse/Panoptes-Front-End/pull/2816.

rogerhutchings:
ref #2817

hughdickinson:
Many thanks for the code review. I plan to implement the suggested fixes within the next two months.

hughdickinson:
Specific coding suggestions have been implemented and tested. Ready for another review.

hughdickinson:
Will do. Will the #3081 merge happen soon?

srallen:
@hughdickinson it was merged yesterday, so if you rebase, you should get the fixed dev classifier.

hughdickinson:
@srallen Sorry, I noticed that was the case only after commenting! Anyway, changes implemented. One potential issue - It looks like the default `instruction` for the drawing tools only really applies to points. Is this an issue and is there a way to override it for a particular tool?

srallen:
I think what you're looking to edit is here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/generic-editor.cjsx#L198-L210

hughdickinson:
@srallen Changes made. Initially I had protected the new statements with `@canUse`-based conditional directives, but this seemed to imply more pervasive changes, so I removed them.

eatyourgreens:
I think all of the common code in these three components can be extracted into a single Arrow tool, with a dropdown menu in the project builder to select how the component is rendered. A switch on the value of the dropdown would then select the rendering to use. Rendering could be done either by implementing 3 different render methods in the base component, or handing off rendering to a separate, child component of the base Arrow.

I'm happy to talk this through on Slack if you need some pointers.

srallen:
@eatyourgreens's refactor suggestion sounds like the right way to go to me too. I can be available for any questions too.

srallen:
I'm gonna close this for now since it's far behind master. It can be reopened or a new PR opened when it's ready.

PR #2853, opened 2016-08-04
Limit username length to 255 characters

rogerhutchings:
via zooniverse/Zooniverse#51 - the API errors out if usernames are too long


simoneduca:
@rogerhutchings LGTM, happy for me to merge?

eatyourgreens:
Should be `maxLength` instead of `maxlength`, I think: https://facebook.github.io/react/docs/tags-and-attributes.html

simoneduca:
Ha! Good catch, @eatyourgreens.

srallen:
LGTM

PR #2856, opened 2016-08-04
props.user is always undefined on first mount

### LINKED ISSUE #2767: Can't directly load project pages in lab
Going directly to one of my projects (e.g. https://www.zooniverse.org/lab/2731) now loads /lab/ instead.

eatyourgreens:
Remove the first user check since we should wait until the auth request has happened.
Fixes #2767


astopy:
LGTM when it comes to fixing #2767. Are there any potential side effects from removing this check? What else should I test?

eatyourgreens:
I'm not sure, but the method I deleted was added in response to #1125

PR #2863, opened 2016-08-07
Change name -> profile_name in PM form

astopy:
Untested at the moment. Staging now...


alexbfree:
Thanks for fixing this @astopy @eatyourgreens . I think when I tested this I tried sending a message, and saw that it didn't break the UI, but I guess i didn't go as far as checking the recipient inbox, I probably assumed "well the back-end hasn't changed so that must still work since the UI didn't break". Oops! Definitely think we need some better regression tests we can run to protect against oversights like this!

PR #2864, opened 2016-08-07
Change props.name to props.profile_name

eatyourgreens:
Looks like this got changed in `router.cjsx` recently


eatyourgreens:
Superseded by #2863

PR #2876, opened 2016-08-10
Only display recent projects that have user classifications

### LINKED ISSUE #2865: Logged in home page should only show projects a user has contributed to
https://www.zooniverse.org/talk/17/65759?comment=138573 

We need to only show projects from project prefs with active counts (not just projects that have been viewed) [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/home-logged-in.cjsx#L90
). It's the same as the stats page from before #2779.

eatyourgreens:
Filter a user's recent projects for those with activity count > 0.
Also removes a couple of PromiseRenderers from the logged-in home page.
Closes #2865

### DIFF
--- a/app/pages/home-logged-in.cjsx
+++ b/app/pages/home-logged-in.cjsx
@@ -72,9 +72,10 @@ RecentProjects = React.createClass
   
   componentWillMount: ->
     @props.user
-      .get("project_preferences", page_size: 4, sort: '-updated_at')
+      .get("project_preferences", page_size: 10, sort: '-updated_at')
       .then (projectPreferences) =>
-        userProjects = projectPreferences.map (projectPreference) -> projectPreference.links.project
+        userProjects = projectPreferences.map (projectPreference) -> projectPreference.links.project if projectPreference.activity_count > 0
+        userProjects = userProjects.filter(Boolean).slice 0, 4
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Staged at https://fix-2865.pfe-preview.zooniverse.org/

app/pages/home-logged-in.cjsx
```diff
@@ -60,6 +60,56 @@ FeaturedProjects = React.createClass
...
+  getInitialState: ->
+    projects: []
+    activityCounts: {}
+    title: ""
+    href: ''
+    action: ""
+  
+  componentWillMount: ->
+    @props.user
+      .get("project_preferences", page_size: 10, sort: '-updated_at')
```
camallen: seems this end point should be able to filter on activity counts instead of filtering later...thoughts?

camallen:
LGTM

app/pages/home-logged-in.cjsx
eatyourgreens: That would be useful here too: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/classifications-ribbon.cjsx#L100
rather than filtering that array later.

PR #2879, opened 2016-08-10
Bugfix: actually display your recent projects when logged in

eatyourgreens:
Fixes the `if` block so that it renders the project cards.


camallen:
tested on my dev server and it's working.

Issue #1868, opened 2015-10-26
Some way of indicating a project when sending a PM


DarrenMcRoy:
Something requested for Talk: in V2 Talk, PMs would indicate which project they were sent from. This probably isn't possible in V3 Talk as so, since you don't have to (actually can't) be on a specific project when you send a PM.

However, the cross-project moderators say that they found that indicator of what project a message was sent from to be extremely helpful, as they now often get messages that assume they know what project the person is talking about.

Would it be possible to resolve this somehow? Any ideas? Perhaps some sort of drop-down to select the project you're talking about... or would that become too long very quickly? Any ideas?

yshish:
As for the 'too long drop-down to select', I've got another idea: Could it be possible to make it as an autocompleting the name of the project after someone starts typing it (but it would probably need some example/Help how to use it). Not sure whether it could work even if you start with a word in the middle of the name (I'm afraid it won't be very useful if it wasn't possible). So for Wildcam Gorongosa, if you started typing *Goro...* it would offer you *Wildcam Gorongosa* 

Could this be possible?

DarrenMcRoy:
To take the other side, this might be considered a really unnecessary feature when we could instead train people to just write the exact same thing, the name of the project they're talking about, in the "subject' line.

Maybe this is solvable by adding a note that says "If you are messaging about a specific project, please mention the project?"

aweiksnar:
> To take the other side, this might be considered a really unnecessary feature when we could instead train people to just write the exact same thing, the name of the project they're talking about, in the "subject' line.

Yep I'd definitely agree with this. Adding a UI to populate the subject line or even linking private messages to projects in the api feels like it'd be extraneous here. I'd rather put the dev time elsewhere if it's ok with everyone

eatyourgreens:
Re-opening this for @alexbfree now that we have the ability to send PMs from specific projects.

alexbfree:
@eatyourgreens and I discussed this a bit (see https://zooniverse.slack.com/archives/general/p1465311829000085 and https://zooniverse.slack.com/archives/dev/p1470647316000554), I think there is something we can do here so am reopening.

We should tag messages by project, like Ouroboros does:
![screenshot 2016-08-08 10 20 15](https://cloud.githubusercontent.com/assets/1473244/17475517/50c245dc-5d53-11e6-9b8e-678c1d92da06.png)

Not sure if messages have a place we can add this metadata but we should look into it. 

[...]

PR #2880, opened 2016-08-10
Add project prefix to user profile URLs when in project

### LINKED ISSUE #2866: User profile links are inconsistent within a project's Talk pages
If you click on a username from this page, https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/talk/115 it will take you to a user profile page for that project https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/users/StarwatcherHB

On other Talk pages, the username links back to the top-level Zooniverse profile
https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/talk/115/74131
https://www.zooniverse.org/projects/alexbfree/computer-vision-serengeti/talk/recents

alexbfree:
Fixes a few places where I'd missed adding the project prefix - it was missing and wrongly sending people back to Zoo-wide context.

Thanks @eatyourgreens for spotting this.

Fixes #2866

### DIFF
--- a/app/talk/comment.cjsx
+++ b/app/talk/comment.cjsx
@@ -132,7 +132,7 @@ module.exports = React.createClass
       baseLink += "projects/#{@props.project.slug}/"
     <div key={comment.id} className="comment-reply-line" ref="comment-reply-#{comment.id}">
       <p>
-        <Link to="/users/#{comment.user_login}">{comment.user_display_name}</Link>
+        <Link to="#{baseLink}users/#{comment.user_login}">{comment.user_display_name}</Link>
         {if comment.reply_id
           <span>
             {' '}in reply to <Link to="#{baseLink}users/#{comment.reply_user_login}">{comment.reply_user_display_name}</Link>'s{' '}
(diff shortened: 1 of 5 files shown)


alexbfree:
Deployed to https://fix-bad-user-profile-links.pfe-preview.zooniverse.org/ in case you want to test.

eatyourgreens:
Is the inbox used anywhere other than zooniverse.org/inbox? I'm not sure about modifying that, because the list of messages there will have come from people on a mixture of projects, so I'm not sure if there is a specific project context for the inbox.

Other than that, looks ok to me but I'm not very familiar with the Talk components.

alexbfree:
good point about inbox. It would jump to the context of the project you are in, rather than the context of the message, which isn't what we want. i'll remove that one.

eatyourgreens:
Yeah, private messages don't have a project context, as per #1868

alexbfree:
Actually on second thoughts it wouldn't jump to a project context... it's impossible for @props.project to be set while viewing the inbox. So the links will remain without a project context. It'll make no difference to the rendered links (just tested this too)

I'll remove it anyway, in case we ever use that page from a project context.

alexbfree:
ok @eatyourgreens that should be good to merge now. Thanks in advance.

PR #2882, opened 2016-08-10
[Work in progress]: Towards speeding up Recent Projects

eatyourgreens:
Uses a faster API endpoint for Recent Projects, at the expense of losing the owner name from each card.

### DIFF
--- a/app/components/project-icon.cjsx
+++ b/app/components/project-icon.cjsx
@@ -11,32 +11,14 @@ module.exports = React.createClass
     onClick: null
     linkTo: true
 
-  getInitialState: ->
-    href: ''
-    avatar: null
-
-  componentDidMount: ->
-    @getDetails @props.project
-
-  componentWillReceiveProps: (nextProps) ->
-    unless nextProps.project is @props.project
-      @getDetails nextProps.project
-
-  getDetails: (project) ->
-    project.get 'owner'
-      .then (owner) =>
-        @setState {owner}
-    project.get 'avatar'
-      .catch =>
-        null
-      .then (avatar) =>
-        @setState {avatar}
-
   render: ->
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Staged at https://recent-projects.pfe-preview.zooniverse.org/

alexbfree:
I think that this shouldn't be necessary after https://github.com/zooniverse/Panoptes/pull/1825 - Panoptes should be returning the user details everywhere it's returned as a linked resource. This allowed me to speed up the collections page to show owners without separate calls to get user details - I think it should work for projects owners too (haven't checked tho)

eatyourgreens:
Projects need to load in linked avatars too, just to make this more interesting.

alexbfree:
True. I guess there would be a case for that info to be returned as a linked resource by Panoptes too.

eatyourgreens:
`apiClient.type('projects').include('owner, avatar').get()` will include them with the request too.

alexbfree:
Ah of course - seems obvious now you say it!

eatyourgreens:
Hat-tip to @marten for pointing out side loading via `.include()`. It's documented in the [Panoptes API](http://docs.panoptes.apiary.io/#reference/projects/project-collection/list-all-projects) but isn't mentioned in the [API client readme](https://github.com/zooniverse/json-api-client) (does work though.)

eatyourgreens:
@camallen could you raise an issue on zooniverse/json-api-client about caching? I think the API client caches resources by type and ID. Maybe `projectCard` needs to be a distinct resource type from `projects`, otherwise the API client is going to expect the resources returned by that call to match the resources returned by any other `projects` request.

eatyourgreens:
Superseded by #2928 Probably still affected by #2921 though.

PR #2885, opened 2016-08-11
Photo in correct place

BNewing:
Moved the photo across to the right place now! @camallen @alexbfree


camallen:
LGTM - @alexbfree good to merge?

Issue #824, opened 2015-06-26
Not being able to switch or add subjects sets on a live project without making it not live


mschwamb:
Not for launch, but I just noticed that I can't switch subject sets once a project is live. That seems problematic that you have to take a project down temporarily to add a subject set? Thought I'd mention it in case it's any issue for another of the other live projects

chrissnyder:
Yea, @aliburchard brought up a similar issue on adding subject sets to/from workflows. Definite thing to fix.

vrooje:
Linking this to zooniverse/Panoptes#1021

chrissnyder:
Labeling this as Sprint 1, but nothing really to do on it on the FE. As soon as zooniverse/Panoptes#1021 is closed, we can close this.

chrissnyder:
Moving this to Future Plans, pending zooniverse/Panoptes#1021

vrooje:
Apparently zooniverse/Panoptes#1021 is done, so now all that's left is front-end mods?

srallen:
Project owners on live projects can now manage a workflow's activeness and can make a workflow inactive to add new subject sets.

PR #2889, opened 2016-08-12
help text addition to visability page 

mschwamb:
It still wasn’t clear to me that I didn’t need to bring the site back to development to swap subject sets. I think an explicit line to remind project owners/collaborators especially when they haven’t been to the project builder in a while might be helpful. Feel free to reject if you think the text on the visibility page is already sufficient


app/pages/lab/visibility.cjsx
```diff
@@ -211,4 +211,5 @@ module.exports = React.createClass
           </div>
       }</PromiseRenderer>
       <p className="form-help">In a live project active workflows are available to volunteers and cannot be edited. Inactive workflows can be edited if a project is live or in development.</p>
+      <p className="form-help">On a live project, if you want to switch which subjects sets are associated with an active workflow: set the workflow to inactive, next change which subject sets are linked in the Workflow Section within the Project Builder, then return the workflow to active.</p>
```
camallen: I'd be in favour of just allowing the linking components to be active instead of this faff. The api will allow these changes to a live workflow.

mschwamb: @camallen if that's doable that would be easier and more straight forward I think on the user side of the project builder. I suggested the text since currently this is the recommended way of switching subject sets, and its not intuitive when you get to the visibility page. See #824, but I think what you suggest is more elegant and less clicks

srallen:
I agree that we should make the subject sets directly editable in the active workflow. This is a good intermediate explanation on how to do this since this question comes up, so I'll merge. @mschwamb could you open an issue for making the subject sets editable for active workflows? This is probably something we can get to after a few of our upcoming project deadlines.

PR #2890, opened 2016-08-15
Allow WWT to support new workflow

### LINKED ISSUE #2861: World Wide Telescope bugs
Testing out Astronomy Rewind and came across a few issues/questions. 

1) 

I thought we had found a fix for if a volunteer marks the axes numbers inside the box they've drawn around the figure itself. Is that not the case? 

In this test (within the old https://www.zooniverse.org/lab/1856/workflow/1700 Workflow 2 for images with sky axes labeled), the WWT button didn't come up as an option when I marked the image as shown. 

marksinside-nowwt

wgranger:
Closes #2861 

This PR mainly fixes two issues:

1.  Selects only drawing tasks from annotations and uses them in the WorldWide Telescope (WWT) component, thus making WWT more flexible to different workflows. Also, all annotations are run through a try, so we shouldn't see any more freezes.

2. The X and Y axes are now found using a slope formula. This allows users to mark inside a chart, if necessary, and the WWT link should still accurately appear.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -278,6 +278,7 @@ Classifier = React.createClass
             annotations={@props.classification.annotations}
             subject={@props.subject}
             user_name={@props.user.display_name}
+            workflow={@props.workflow}
           />
         </strong>
         }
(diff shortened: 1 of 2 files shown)


amyrebecca:
i thought i'd already seen a PR that did this so this was some weird deja vu :) that said, all of this looks fine so i'm going to go ahead and merge it.

Issue #2852, opened 2016-08-03
Generalized pagination solution


srallen:
We should make sure all paginated resources that are returned from the API are properly paginated in the UI. This is related to #2819 and #2752. There's a [paginator component](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/lib/paginator.cjsx) in the Talk UI code that is used in parts of the site. There are other instances of pagination being written as needed like in [the project builder on the subject sets page](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/subject-set.cjsx#L93-L101).

A start of a [general reusable component](https://github.com/zooniverse-ui/pagerz) began in the old zooniverse-ui, but wasn't finished. We could finish this, consider using [one that is already built](https://github.com/AdeleD/react-paginate), or make sure we write pagination consistently as needed.

alexbfree:
+1 to this. Pagination is something we shouldn't be reinventing!

alexbfree:
Per #2868 , we should ensure that the generalized pagination solution correctly auto-scrolls to top of page (which currently happens for Talk pages, and project Talk pages, but not for collections, project collections, and lists of projects)

srallen:
Adding scroll behavior middleware is a generalized solution needed for the router, not for pagination. See #2762. Scroll to top fixed in #2874

srallen:
Paginator was generalized and moved into zooniverse/Zooniverse-React-Components

Issue #2752, opened 2016-06-28
Can't create new subject sets through project builder.


dwright04:
I have 28 subject sets.  I can only see 20 (issue [#1136](https://github.com/zooniverse/Panoptes-Front-End/issues/1136)) of these in the project builder sidebar.  I have a subject set called "Untitled subject set", but I cannot select it from the sidebar to upload subjects, change the name or delete it.  As new subject sets are created with the name "Untitled subject set" ([#1923](https://github.com/zooniverse/Panoptes-Front-End/issues/1923), [#1924](https://github.com/zooniverse/Panoptes-Front-End/issues/1924)) I get the error "Validation failed: Display name has already been taken" and can no longer add new subject sets through the project builder.

![screen shot 2016-06-28 at 15 00 34](https://cloud.githubusercontent.com/assets/7902164/16418308/204fbbf0-3d41-11e6-97e9-97bde64a9de2.png)

aliburchard:
Is this maybe because we don't elegantly handle lots of subject sets? (This is a problem in a number of projects -- and if we start telling teams with hundreds of thousands of images to limit their subject set size to 10s of thousands of images, then this is probably going to happen more and more...)

Issue #1136, opened 2015-07-20
No more than 20 subject sets appear in project builder


mkosmala:
I made 21 subject sets. But only 20 show up. Thinking I forgot to make one, I tried re-making it, but got an error (pink box). I think that only the first 20 subject sets retrieved are displayed. There is no way to see more than 20 of them (though they probably exist).

mschwamb:
fixed with the new subjects page in the project builder

PR #2900, opened 2016-08-18
Edit lab side-bar subject sets list API call

mcbouslog:
Temporary solution to #1136, #2752.

Better solution still Generalized Pagination #2852, which if I have time I'll try and start asap, but if not done before ZTM, something to keep in mind for then.

@camallen, @srallen - similar to recent workflow API call change I know this isn't ideal, but helps until pagination implemented. This change is for only within the Project Builder, subject sets left side bar list; smaller scope than workflow call change.


srallen:
LGTM, but we really need to prioritize pagination soon.

PR #2901, opened 2016-08-18
Custom sign in prompt

srallen:
This adds a custom sign in prompt for Gravity Spy (under the experimental tool flag for workflow assignments - open to adding a separate flag just for this). This is similar to the Planet Hunters style prompt. It will appear initially and then every 3 classifications. There are several features dependent on signed in users including seeing the mini-course and workflow assignment, so we really want to encourage volunteers to sign in/sign up. 

This is WIP since the message content hasn't been finalized nor how often to prompt.

Staged at: https://custom-sign-in-prompt.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing/classify?workflow=2333
Must be signed out to test.


mschwamb:
@srallen you might want to remove the default log in giggling button on the bottom right of the subject when this component is active on a workflow.  Looks great. I'd love to have this for my projects when this feature is finished.

srallen:
I've updated the prompt to be every 5 classifications. I think I'll leave the animated prompt for now. We'll be beta testing this custom prompt with Gravity Spy, so we'll find out if users find it annoying to the point of leaving. We want the sign in suggestions to be a bit annoying though to get them to do it.

@trouille do we have a finalized message to use?

trouille:
Thanks @srallen I've pinged the research team to have confirmation of wording by early this afternoon. Will post here. 

If a project has an announcement banner, do you know how this prompt and the announcement banner will sit together?

srallen:
The announcement and/or finished banner will appear above it. I'd like to refactor these into a more generic growl style notification component, but that'll have to be later.

srallen:
Copy has been finalized as is. Removing WIP label.

PR #2902, opened 2016-08-19
added shouldUpdateScroll to router

### LINKED ISSUE #2898: Impossible to set the range of classification stats as needed. The page reloads and scrolls up before I finish the setting.
When I click to change the range of Classification Stats graph the page reloads and scrolls up before I finish setting the range. It happens too soon, there is actually no time for setting the range as the page reloads and scrolls up immediately as the cursor starts moving even before it gets to the needed place and even before I release the pressed mouse button. It means that I don't see the cursor while setting the range of the graph and I don't know how much to move it to get the range I need.

I would expect it to let me finish the setting, and not start reloading before it is set (in case the reloading is necessary, is it?) And after it reloads it also should take me back to the graph I was setting up (as there are more of them on the stats page).

By *finish the setting* I mean after I release the pressed mouse button. Currently it happens while the button is still pressed.

Using Chromium 48.0.2564.82 on Ubuntu

CKrawczyk:
This fixes #2898

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -12,7 +12,15 @@ if location?.hash.charAt(1) is '/'
 browserHistory.listen ->
   dispatchEvent new CustomEvent 'locationchange'
 
-ReactDOM.render <Router history={browserHistory} render={applyRouterMiddleware(useScroll())}>{routes}</Router>,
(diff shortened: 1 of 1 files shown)


srallen:
LGTM

PR #2903, opened 2016-08-19
Show buttons for workflows on the home page based on level

srallen:
For Gravity Spy: This adds a feature that filters the workflow buttons on the project home page based on the user's level. If logged out, then the standard 'Get Started' button will show and the default workflow will load.

I also refactored the home page to not use `PromiseToSetState` or `HandlePropChanges`. These abstracted away the lifecycle of the component making it difficult to reason about what was happening and difficult to debug since errors were set in the state instead of logging to the console.

Staged at: https://workflow-level-buttons.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing

When you're ready to test this out, let me know because I will need to modify your user account's project preference's settings via the python client so you can see what the new functionality does.

### DIFF
--- a/app/pages/project/home.cjsx
+++ b/app/pages/project/home.cjsx
@@ -52,7 +52,7 @@ module.exports = React.createClass
             <small>at {@props.project.redirect}</small>
           </a>
         # For Gravity Spy
-        else if @props.project.experimental_tools.indexOf 'workflow assignment' isnt -1
+        else if @props.project.experimental_tools.indexOf('workflow assignment') isnt -1
           if @props.user?
             currentWorkflowAtLevel = @state.workflows?.filter (workflow) =>
               workflow if workflow.id is @props.preferences?.settings.workflow_id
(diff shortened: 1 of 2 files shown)


app/pages/project/home.cjsx
```diff
@@ -49,6 +51,27 @@ module.exports = React.createClass
             <strong>Visit the project</strong><br />
             <small>at {@props.project.redirect}</small>
           </a>
+        # For Gravity Spy
+        else if @props.project.experimental_tools.indexOf 'workflow assignment' isnt -1
```
mcbouslog: I think `indexOf 'workflow assignment'` should be `indexOf('workflow assignment')` otherwise I get stuck in the Gravity Spy condition?

srallen:
Discovered zooniverse/Panoptes#1920. Relabeling as WIP since this can't be tested until this is resolved.

app/pages/project/home.cjsx
srallen: Thanks, forgot this was a problem in coffeescript.

srallen:
This is gonna need a refactor at some point, but it's functional for the beta.

PR #2904, opened 2016-08-20
Add content to how-to document on what happens when adding '!' to the fieldname in the manifest file

mschwamb:
Add to the how to document a line or two to explain how to add a ‘!’ to the fieldname in the manifest file to make the column accessible only on Talk


parrish:
Thanks Meg.  It looks like some weird rebase history wound up in this though.

GitHub *REALLY* needs to change how they show diffs on pull requests -- you can end up merging changes without realizing it.  If you look at the patch this PR generates, there are a bunch of changes hiding in this [patch](https://patch-diff.githubusercontent.com/raw/zooniverse/Panoptes-Front-End/pull/2904.patch)

For simple text changes like this, sometimes it's easiest to just use the [web-editor](https://github.com/zooniverse/Panoptes-Front-End/edit/master/app/pages/lab/how-to-page.cjsx).  Just check the "Create a new branch for this commit and start a pull request" option

mschwamb:
@parrish Okay. Currently I was taking my cloned repo  on my githubt account and using the button update from Zooniverse master and then adding in my change and pushing to the repo and the issuing a pull request. I'll close this and add in the change as you requested.

PR #2910, opened 2016-08-23
Hide subject metadata prefixes when displayed

### LINKED ISSUE #2907: hide the '!' for Talk metadata if the column head in the metadata file has a '!'
With the new feature to have metadata that only shows in Talk it looks a bit weird when you view it to have the '!' If that could be removed from the string when displayed it would look nicer.  cc @parrish 

screen shot 2016-08-22 at 11 21 34 pm

parrish:
Closes #2907


simensta:
LGTM. Thanks @parrish!

PR #2911, opened 2016-08-23
Workflow assignment dialog

srallen:
This adds the dialog UI for workflow assignment/promotion as well as a bug fix for the Tutorial and removed a redundant API call. Also, updated the eslint file and changed it to `.eslintrc` so it would work with Sublime.

Still need to:
- Finalize copy for the dialog
- Finalize any styling for the dialog
- Do some additional testing when I can successfully get Project Preferences of another user and promote them (via python client)

This leaves out the potential opt out feature. I think that's gonna take more thought and time if we do it at all. FYI @trouille 

Staged at: https://workflow-assignment-dialog.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -100,7 +100,7 @@ module.exports = React.createClass
   shouldWorkflowAssignmentPrompt: (preferences) ->
     # Only for Gravity Spy which is assigning workflows to logged in users
     assignedWorkflowID = preferences?.settings?.workflow_id
-    if @props.project.experimental_tools.indexOf 'workflow assignment' > -1 and @props.user?
+    if @props.project.experimental_tools.indexOf('workflow assignment') > -1 and @props.user?
       if assignedWorkflowID? and assignedWorkflowID isnt @props.location.query.workflow
         @setState promptWorkflowAssignmentDialog: true if @state.promptWorkflowAssignmentDialog is false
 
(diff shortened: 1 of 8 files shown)


srallen:
Content and styling are done.

srallen:
Rebased to include the workflow level buttons feature.

app/pages/project/classify.cjsx
```diff
@@ -92,6 +94,16 @@ module.exports = React.createClass
   componentWillUnmount: () ->
     @context.geordi?.forget ['workflowID']
 
+  componentWillReceiveProps: (nextProps) ->
+    @shouldWorkflowAssignmentPrompt(nextProps.preferences)
+
+  shouldWorkflowAssignmentPrompt: (preferences) ->
+    # Only for Gravity Spy which is assigning workflows to logged in users
+    assignedWorkflowID = preferences?.settings?.workflow_id
+    if @props.project.experimental_tools.indexOf 'workflow assignment' > -1 and @props.user?
```
mcbouslog: This `indexOf` doesn't have the parentheses around `'workflow assignment'`, but it seems to be working fine, so I'm not sure if the weird Coffeescript and indexOf compiling thing is actually happening. Maybe the `if assignedWorkflowID?` conditional below is stopping it from causing problems?

srallen: Probably would be a good idea to include the parentheses anyway. Thanks for catching that again here.

PR #2914, opened 2016-08-24
Add "nothing here" option as unlinked task

wgranger:
Allows addition of "nothing here" option which will end classification immediately.


eatyourgreens:
I'm not sure I understand what this adds. Doesn't the project builder already allow me to add an answer where the next task is the end of the classification workflow?

wgranger:
Currently, you can only do that for a single choice question, not for a multiple choice question. Multiple choice questions all have the same next task.

eatyourgreens:
Ah, right. So this needs to be a combination of a multiple choice task, where all the answers go to one next task, and a single choice answer where the answer jumps to the end of the classification.

I'm inclined to say that we should either allow individual choices from a multiple choice task to override the next task setting (like they do now if you don't check 'choose multiple') or use the combo task to build a multiple choice task plus a single answer task which branches in two directions.

I'm hesitant about this PR because it hardcodes the button label as 'Nothing Here', hardcodes the behaviour as jump to the end of the workflow and because it doesn't allow for adding options like 'Nothing here' to survey projects.

wgranger:
Hmm. That makes sense. The original solution, before the recent refactored commit, consisted of using a "nothing here component" rather than hardcoding. However, it was only available for question tasks, and the question lingered if this was only for question tasks or for other options. 

Do you know of a project that uses the survey tool that I can look at for reference and see if I can revert back to the other approach, while also opening it up to other tasks, particularly survey.

eatyourgreens:
Here's an example that @mrniaboc and @aliburchard set up today.
https://www.zooniverse.org/projects/mrniaboc/zooniverse-go
You might need to be added to the project in order to see it.

It would be a cool if we could use the combo task to build a step that consisted of a survey task and a question task. I think that would handle this case too.

aliburchard:
@wgranger some PFE projects that use "nothing here" in the species list include:
https://www.zooniverse.org/projects/birgus2/western-shield-camera-watch
https://www.zooniverse.org/projects/panthera-research/camera-catalogue
https://www.zooniverse.org/projects/zooniverse/snapshot-wisconsin

previously, custom projects used the equivalent of a combo task to do the 'nothing here', e.g. https://www.snapshotserengeti.org/#/classify

(as a tangent, I'm really curious how much of a difference it makes if the "nothing here" is in the species list vs. a separate question on the same page...)

wgranger:
@aliburchard Yes, that was a convo I had w/ Jordan about how "Nothing Here" should be displayed, with the ultimate answer being similar to any other question option. However, that might change if it's open to other tasks.

wgranger:
Staged at: http://nothing-here-shortcut.pfe-preview.zooniverse.org/projects/wgranger-test/nothing-here-shortcut/classify

Is available for survey tool editor and generic task editor. Note: This isn't currently listed as an experimental feature.

eatyourgreens:
On the test page, I can select `Yes`, to answer the question *and* click `Nothing here` and submit that as my classification. The classification summary says my answer is 'nothing here' but the submitted classifications has the 'Yes' question index in the annotation value, as well as a `nothingHere` flag set to `true`. I'm not sure how Nero or any aggregation code should handle that classification 🤔

Sorry, but I think adding a flag to the task annotation in addition to the value isn't the best way to implement this. I'd be happier if it behaved like an answer to the single answer task, overwriting the annotation value. The UI should indicate that the answers are mutually exclusive too, clearing any other answer if the checkbox is selected. Alternatively, this checkbox should be disabled if an answer has been selected for the classification task.

I think it's a bad idea, in the long run,  to have the `Nothing Here` text hardwired into the component, with the choice restricted to either have one button or none. This prevents project builders from adding options like the `Fire` checkbox in [Snapshot Serengeti](https://www.snapshotserengeti.org/#/classify). 

Can we allow for more than one checkbox, and allow project builders to edit the button text?

Last, minor niggle: I can add a 'nothing here' answer to a single answer task and also show a 'Nothing Here' button, which is confusing. I think that ties into my comment above, about this being added as a flag on the annotation, rather than an answer in its own right.

wgranger:
Gotcha. Several things to consider here. Yes, I mistakingly took out that functionality to deselect other options in the "multiple" component when "nothing here" is selected. It will need to be improved upon when I see how the survey tool stores annotations. 

Also, I think I'm starting to see the purpose of `fire`. I was a little unclear of what it was supposed to mean, but I assume it works the same way as selecting "nothing here". 

I'm not quite sure what you mean by disabling the checkbox if an answer has been selected for a classification task. Do you mean that once a volunteer selects an answer, that checkbox should be disabled? Seems like a volunteer might potentially change their mind about something they see during the annotation.

Considering the work that needs to be done on this, I'm changing it to WIP.

eatyourgreens:
Re. disabling the checkbox, I was looking at Snapshot Serengeti and noticed that the 'Nothing Here' button actually vanishes if you identify and animal. It comes back if you delete your identification. So I thought that one way to approach this checkbox was that it should only be enabled if you haven't supplied an answer for the current task.

Good point about changing your mind. On Serengeti you have to delete your classification to change your mind. I guess here you'd have to deselect your answers, but it would be quicker just to click 'Nothing Here' and have that clear the form, right, rather than disabling the button as I suggested?

I'm open to either of those but we can't let you submit an ambiguous classification.

eatyourgreens:
One more comment, sorry. Ideally, this component shouldn't be coupled to the other tasks eg. changing the task annotation, it shouldn't care what type of task it's dealing with. Otherwise, including this when developing new tasks could become painful. 

One way to approach that might be to use the Decorator pattern to wrap the task component in a component that adds the nothing here checkbox and interacts with the task through `props.children`, regardless of the exact type of task. http://www.zhubert.com/blog/2016/02/05/react-composability-patterns/

wgranger:
Ok, I've rewritten this to make it more flexible and allow multiple, editable shortcuts. It is staged at: https://nothing-here-shortcut.pfe-preview.zooniverse.org/projects/wgranger-test/test-project-shortcut/classify

My only lingering question concerns what to do with `@props.annotation.value`. This is a pesky flag that has made the logic much more difficult because it is treated differently by different components (either as an array or string). However, concerning this PR, if shortcut/nothingHere is being used, the value will be empty and a `shortcut` flag will be added to the annotation. 

I'm still wondering how an annotation _summary_ should handle an empty value. The tricky part is that each component handles `annotation.value` differently in the summary. Sometimes the summary looks for a simple string, other times it looks for particular keys in an object (as in the survey tool).

eatyourgreens:
Any thoughts from @zwolf or @marten on how this would be handled in data exports or aggregation?

wgranger:
Staged at: https://nothing-here-shortcut.pfe-preview.zooniverse.org/projects/wgranger-test/test-fire/classify

Was also added as an experimental feature with recent commit. Nothing Here is now included as a separate task that is unlinked and invisible (as a "normal" task) in the project builder.

camallen:
Just noting that i'm working on the annotation formatter for csv(ish) exports. If this feature ships without that work then the classification exports will error out, not a blocker though.

wgranger:
Ok. Has it been erroring out as is? I've been able to get classification exports from test projects as is, but maybe you're getting error notifications from Honey Badger?

camallen:
Interesting, great if it works! I assume it has uninformative errors, what does a row for that task annotation in the dump look like?

wgranger:
`[{"task":"T3","value":{"shortcut":"Fire"}}]`

camallen:
sweet, thanks @wgranger - i'll double check it's not just dumping the raw annotations.

wgranger:
Just got note that this task **will** need to record answers from the original task that is being annotated, as opposed to clearing out the value. So, it will look like values from both the NothingHere task and the original/main task will need to be recorded in the output. 

@eatyourgreens I'll make the changes and hope to have it ready by early next week. 
@camallen This will probably affect how the data dump comes from the CSV export.

camallen:
Interesting, is there a use case for both?

wgranger:
An example use case would be a fire on a camera trap. The user would record "fire" (that would trigger the shortcut) as well as any animal (rhino, lion, etc.) that might be in the image.

eatyourgreens:
So I can record a survey classification as "1 lion present" and "nothing here"? I guess we trust volunteers not to submit contradictory data but this does seem like bad interface design to me.

wgranger:
Yes. Because of our use case, the button will not clear out the already recorded value.

wgranger:
What we will plan to do at this point is to get this feature into PFE as an experimental feature to allow the research team to test the feature. Next, we will tackle the option of clearing out the value with the click of nothing here in a separate PR. Perhaps this could be a toggle for each option on the workflow/lab page.

simensta:
Thanks for your work on this Will. My comments are as follows:
1) Naming: Although this feature began as a "Nothing Here" checkbox, the possible answers and use cases have broadened. I think the name of the component in both the code and the project builder should more closely reflect the functionality. Perhaps something like "Shortcut Question"?

2) In the project builder, I think a brief description of the functionality should be included near the checkbox for enabling the feature. 

3) If the shortcut question is on the second task, and the user navigates to the previous question via the back button, the shortcut question value is not reset.

wgranger:
Hey. Thanks a ton for the review. Here's a little summary of the changes I made.

1) Yes, Cam also brought up this point so the CSV exporter knows what to look for in the backend. I changed the title to `shortcut` since, by convention, our other tasks have one word titles, and that seemed the most appropriate.

2) Nice. I had a descriptor when hovering over the workflow title, but I think this makes much more sense.

3) Good catch on this. I failed to notice this when I was playing around with the components. I think I got this with a `componentWillReceiveProps`. I also deleted the `shortcut` tag from the current annotation when it is being toggled between active/inactive in the classifier.

camallen:
@wgranger FYI - i've updated the backend to recognise the `shortcut` task types in the annotation dumps.

wgranger:
Nice. I've also updated the code on the front end.

simensta:
After reviewing the new changes, I have the following questions:
- Should a user be able to select more than one answer for the shortcut question? For example, I can select "Nothing Here" or "Fire", but not both.
- If the shortcut question is enabled for more than one task, changing the shortcut answers for one of the tasks, changes the shortcut answers for all of the tasks. Are the shortcut options suppose to be the same for all of the tasks in a workflow?

wgranger:
Nice catching that bug. It was an error with the taskID in `shorcut-editor`. 

Yes, the shortcut acts as a single choice option, and the backend is set to handle the export of the value as so. A user should only be able to select one option.

PR #2915, opened 2016-08-24
add gravity spy gold standard feedback

mcbouslog:
[staged branch and testing project here](https://gs-gold-standard-notice.pfe-preview.zooniverse.org/projects/markb-panoptes/gravity-spy-gold-standard-notice-test-project).

let me know staging user and I'll add to test project. the test project has a test gold standard subject set and a test regular subject set.

even though there should only be one selection made, if user makes multiple selections, should `match` be strict (`.every`) or forgiving (`.some`, current setup)?

currently, `hide_classification_summaries` is needed to be checked to hide a regular subject's summary, but that's easily change-able if preferred.

possible ToDo - in wrong response, link references to Tutorial and Field Guide. Though there maybe an argument to leave as is so primary action is to click Next and continue classifying, where they can access those resources within the next subject.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -64,7 +64,6 @@ Classifier = React.createClass
     try
       @context.geordi?.forget ['subjectID']
 
-
   loadSubject: (subject) ->
     @setState
       subjectLoading: true
@@ -138,6 +137,8 @@ Classifier = React.createClass
         <div className="task-area">
           {if currentTask?
             @renderTask currentClassification, currentAnnotation, currentTask
+          else if @subjectIsGravitySpyGoldStandard()
+            @renderGravitySpyGoldStandard currentClassification
           else if not @props.workflow.configuration?.hide_classification_summaries # Classification is complete; show summary if enabled
             @renderSummary currentClassification}
         </div>
@@ -149,8 +150,6 @@ Classifier = React.createClass
 
     # Should we disabled the "Back" button?
     onFirstAnnotation = classification.annotations.indexOf(annotation) is 0
-
-
 
     # Should we disable the "Next" or "Done" buttons?
     if TaskComponent.isAnnotationComplete?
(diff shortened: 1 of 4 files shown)


app/classifier/index.cjsx
```diff
@@ -445,6 +477,12 @@ Classifier = React.createClass
   renderBackButtonWarning: ->
     <p className="back-button-warning" >Going back will clear your work for the current task.</p>
 
+  subjectIsGravitySpyGoldStandard: ->
+    if @props.workflow.configuration?.gravity_spy_gold_standard and @props.subject.metadata?['#Type'] is 'Gold'
```
srallen: This line can be just `@props.workflow.configuration?.gravity_spy_gold_standard and @props.subject.metadata?['#Type'] is 'Gold'` and either true or false will be returned.

srallen:
Noting here that @trouille is asking about the `.every` vs `.some` question.

srallen:
About hiding classification summaries: I think showing gold standard summary responses and hiding classification summaries as separate toggles is desirable. Great job with this!

Launching the tutorial and/or field guide have their own buttons that are visibly close to the task area. I don't see any reason to add redundant links to them.

app/pages/lab/workflow.cjsx
```diff
@@ -286,6 +286,24 @@ EditWorkflowPage = React.createClass
...
             </div>}
 
+          {if 'Gravity Spy Gold Standard' in @props.project.experimental_tools
+            <div>
+              <div>
+                <AutoSave resource={@props.workflow}>
+                  <span className="form-label">Gravity Spy Gold Standard</span><br />
+                  <small className="form-help">Notify a user how they've classified a Gold Standard subject.</small>
+                  <br />
+                  <label htmlFor="gravity_spy_gold_standard">
```
srallen: `htmlFor` is not necessary here since the input is the child of the `label` element. If you do use `htmlFor`, it needs to match the id of the form input it is for. In this case, the input is missing the id.

srallen:
I would prefer this to be eventually be pulled out into its own component since it's so project specific, but that can come in a later PR. LGTM.

PR #2919, opened 2016-08-25
New Additions for Signed In Home Page

wgranger:
PR for Brian's `new-signed-in-home-page` branch. Addition of sidebar pullout and social media section. 

Staged at: https://new-signed-in-social-media.pfe-preview.zooniverse.org/home-for-user
Must be signed in to view page.


simensta:
I think we should replicate the hover-experience for focus-events when a user is tabbing through the page.

simensta:
Oops! Accidentally closed this PR. Should be reopened now. Apologies!

astopy:
I like this a lot. I just have a few things to raise:

- Is the plan to actually deploy this at `/home-for-user`? I'd prefer it if it just replaced the regular home page when you're signed in.
- Related to above, it needs to handle users who aren't signed in.
- Some of the text can be hard to read: 
  
  screen shot 2016-08-26 at 09 21 44
- The ribbon takes a long time to load, though that's probably an API issue.

CKrawczyk:
The news tab blends in with sub-category header, making it hard to tell there is a close button until the page is scrolled.  At first I could not figure out how to close the sub-category.

screen shot 2016-08-26 at 11 37 50 am

wgranger:
@CKrawczyk - Good catch. This could probably be fixed with a stroke around the tab button.

@astopy Yes, that's a final thing that needs to be considered - how the home page will handle if a user is already logged in or not. I agree, the `/home-for-user` extension isn't ideal. 

The API call was slowed down a bit once another call was made to retrieve data sets for each project the user was working on. I'm not quite sure another way around that, but perhaps @marten has an idea? Maybe this is another instance of where `contributions` may help?

@simensta - Are you thinking the focus behavior might be best on the "messages", "my recent projects", etc. sections?

simensta:
I think the following areas could benefit from hover/focus styles:
- the "messages", "recent projects", "collections", and "builds" links
- for the focused sections of the project ribbon, it would be nice to see the associated tooltip

Let me know if this feedback belongs on PR #2813, instead of here!

jwbmartin:
I do agree that some of the text in the contribution reel are hard to read but since all of the colors are generated randomly I'm not sure how to solve that problem.

I'd like to see how it looks with a stroke around the tab. Another option would be the tiniest little drop shadow.

srallen:
Not a part of the diff here, but the account I'm testing with doesn't have a profile header set and is throwing `Uncaught (in promise) TypeError: Cannot read property 'src' of undefined`. This should be handled more gracefully.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
+import React from 'react';
+import apiClient from 'panoptes-client/lib/api-client';
+import { Link } from 'react-router';
```
srallen: `Link` is unused in this component.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+const NewsSection = React.createClass({
+
+  getDefaultProps() {
+    return {
+      newDatasets: [],
+    };
+  },
+
+  componentDidMount() {
```
srallen: Put this after `getInitialState`.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+  getInitialState() {
+    return {
+      avatars: {},
+      projects: [],
+      publications: [],
+    };
+  },
+
+  recentPublications() {
+    const articles = []
```
srallen: Missing semi-colon.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+    for (var category in Publications.articles()) {
+      Publications.articles()[category].map((project) =>{
+        articles.push(...project.publications)
+      });
+    };
+    const newestPublications = articles.sort((a, b) => {
+      return new Date(b.date) - new Date(a.date)
```
srallen: Missing semicolon.

app/pages/about/publications-page.cjsx
```diff
@@ -418,6 +418,10 @@ publicationCategories =
 module.exports = React.createClass
   displayName: 'PublicationsPage'
 
+  statics:
```
srallen: I think instead of adding a static method here, we should extract the publication data into its own js file to live in the `lib` folder since this data is now being used in two different components.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+    return {
+      avatars: {},
+      projects: [],
+      publications: [],
+    };
+  },
+
+  recentPublications() {
+    const articles = []
+    for (var category in Publications.articles()) {
```
srallen: I believe what you want to do here is iterate over the keys of the object, so I think a better way to write this would be (and once the data is pulled out into its own file):
```
Publications.keys().forEach((category) => { // do stuff } );
```

srallen: My bad, the syntax should be `Object.keys(Publications).forEach()`

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+      Publications.articles()[category].map((project) =>{
```
srallen: Space after `=>`

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        articles.push(...project.publications)
```
srallen: missing semicolon

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        return project.get('avatar')
+        .catch(() => {
+          return null;
+        })
+        .then((avatar) => {
+          this.state.avatars[project.id] = avatar;
+          this.forceUpdate();
+        });
+      }));
+    })
```
srallen: missing semicolon

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+          {article.citation}
+          <i className="fa fa-long-arrow-right"></i>
+        </a>
+      </p>
+    </div>
+  },
+
+  renderNewDatasets(data) {
+    const link = window.location.origin + "/projects/" + data.href
+    return <div>
```
srallen: Wrap the return in parentheses.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        .then((avatar) => {
+          this.state.avatars[project.id] = avatar;
+          this.forceUpdate();
+        });
+      }));
+    })
+  },
+
+  renderPublication(article) {
+    return <div>
```
srallen: Wrap this `<div>` in this return in parentheses.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        <a href={article.href} target="_blank">
+          {article.citation}
+          <i className="fa fa-long-arrow-right"></i>
+        </a>
+      </p>
+    </div>
+  },
+
+  renderNewDatasets(data) {
+    const link = window.location.origin + "/projects/" + data.href
```
srallen: missing semicolon

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+          <div className='news-pullout-section'>
+            <h4> Recent Publications </h4>
+            {this.state.publications.map((article) => {
+              return this.renderPublication(article);
+            })}
+          </div>
+
+          <div className='news-pullout-section'>
+          <h4> Newest Project </h4>
```
srallen: this needs an indent.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+          {this.state.projects.map((project) => {
+            const avatarSrc = !!this.state.avatars[project.id] ? this.state.avatars[project.id].src : null;
+            return <ProjectCard key={project.id} project={project} imageSrc={avatarSrc} />;
+          })}
+          </div>
+
+          <div className='news-pullout-section'>
+            <h4> New Datasets </h4>
+            {this.props.newDatasets.map((data) => {
```
srallen: Could you add a props validation for `newDatasets`?

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        })
+        .then((avatar) => {
+          this.state.avatars[project.id] = avatar;
+          this.forceUpdate();
+        });
+      }));
+    })
+  },
+
+  renderPublication(article) {
```
srallen: The `<div>` this is returning needs a `key` property since it is in the map function in the render.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+      <p className="news-section-content">
+        <a href={article.href} target="_blank">
+          {article.citation}
+          <i className="fa fa-long-arrow-right"></i>
+        </a>
+      </p>
+    </div>
+  },
+
+  renderNewDatasets(data) {
```
srallen: The `<div>` this is returning needs a `key` property since it is in the map function in the render.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+        <h5 className="news-section-title">{data.project} </h5>
+        <h5> has been updated! </h5>
+        <p className="pullout-timestamp">{data.timestamp}</p>
+      </a>
+    </div>
+  },
+
+  render() {
+    return (
+      <div className='news-pullout-main'>
```
srallen: Let's use double quote on the classes here in the render to be consistent in usage.

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  renderNewDatasets(data) {
+    const link = window.location.origin + "/projects/" + data.href
+    return <div>
+      <a href={link} target="_blank">
+        <h5 className="news-section-title">{data.project} </h5>
+        <h5> has been updated! </h5>
+        <p className="pullout-timestamp">{data.timestamp}</p>
+      </a>
+    </div>
```
srallen: missing semicolon

app/pages/home-for-user/index.jsx
```diff
@@ -9,6 +9,10 @@ import RecentMessagesSection from './recent-messages';
 import MyBuildsSection from './my-builds';
 import ProjectStats from './project-stats';
 import qs from 'qs';
+import HomePageSocial from '../home-not-logged-in/social';
+import NewsSection from './news-section';
+import apiClient from 'panoptes-client/lib/api-client';
```
srallen: Doesn't look like `apiClient` is being used in this component.

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
+      newData: this.fetchRecentSubjects(ribbonData),
+    });
+  },
+
   findProjectLink(project) {
     return `/projects/${project.slug}`;
   },
 
+  fetchRecentSubjects(ribbonData) {
+    const newestSubjects = []
```
srallen: missing semicolon

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
+    ribbonData.map((project) => {
+      if (project['recentSubjectSet'] !== undefined){
```
srallen: Any reason this isn't using dot notation? How about `if (!!project.recentSubjectSet) {`?

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
+        newestSubjects.push(project)
```
srallen: missing semicolon

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
+    });
+    const recentSets = newestSubjects.sort((a, b) => {
+      return new Date(b['recentSubjectSet'].updated_at) - new Date(a['recentSubjectSet'].updated_at)
+    }).slice(0, 3);
+
+    return recentSets.map((project) => {
+      return {
+        project: project.name,
+        href: project.slug,
+        timestamp: moment(new Date(project['recentSubjectSet'].updated_at)).fromNow(),
```
srallen: Maybe use dot notation here?

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
 
+  fetchRecentSubjects(ribbonData) {
+    const newestSubjects = []
+    ribbonData.map((project) => {
+      if (project['recentSubjectSet'] !== undefined){
+        newestSubjects.push(project)
+      }
+    });
+    const recentSets = newestSubjects.sort((a, b) => {
+      return new Date(b['recentSubjectSet'].updated_at) - new Date(a['recentSubjectSet'].updated_at)
```
srallen: Maybe use dot notation? Also missing the semicolon.

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
+    }).slice(0, 3);
+
+    return recentSets.map((project) => {
+      return {
+        project: project.name,
+        href: project.slug,
+        timestamp: moment(new Date(project['recentSubjectSet'].updated_at)).fromNow(),
+      }
```
srallen: missing semicolon

app/lib/get-user-ribbon-data.js
```diff
@@ -1,36 +1,37 @@
 import getAllProjectPreferences from './get-all-project-preferences';
 import getColorFromString from './get-color-from-string';
+import apiClient from 'panoptes-client/lib/api-client'
```
srallen: missing semicolon

app/lib/get-user-ribbon-data.js
```diff
@@ -1,36 +1,37 @@
...
-          return project.get('owner').catch(() => {
-            return null;
-          });
-        }
+        });
       }));
     });
 
-    return Promise.all([awaitProjects, awaitOwners]).then(([projects, owners]) => {
+    return Promise.all([awaitProjects, awaitSubjectSets]).then(([projects, subject_sets]) => {
```
srallen: Should use camel case for `subject_sets`

app/lib/get-user-ribbon-data.js
```diff
@@ -1,36 +1,37 @@
...
       return prefsResources.map((prefsResource, i) => {
-        if (projects[i] !== null && owners[i] !== null) {
+        const sortedSubjects = subject_sets[i].sort((a, b) => {
+          return new Date(b.updated_at) - new Date(a.updated_at)
```
srallen: missing semicolon

app/lib/get-user-ribbon-data.js
```diff
@@ -1,36 +1,37 @@
 import getAllProjectPreferences from './get-all-project-preferences';
 import getColorFromString from './get-color-from-string';
+import apiClient from 'panoptes-client/lib/api-client'
 
 function getUserRibbonData(user) {
   return getAllProjectPreferences(user).then((prefsResources) => {
     const awaitProjects = Promise.all(prefsResources.map((prefsResource) => {
-      return prefsResource.get('project').catch(() => {
+      return apiClient.type('projects').get(prefsResource.links.project).catch(() => {
```
srallen: Why was this change made?

app/lib/get-user-ribbon-data.js
```diff
@@ -1,36 +1,37 @@
...
-        }
+        });
       }));
     });
 
-    return Promise.all([awaitProjects, awaitOwners]).then(([projects, owners]) => {
+    return Promise.all([awaitProjects, awaitSubjectSets]).then(([projects, subject_sets]) => {
       return prefsResources.map((prefsResource, i) => {
-        if (projects[i] !== null && owners[i] !== null) {
+        const sortedSubjects = subject_sets[i].sort((a, b) => {
```
srallen: Is the goal to only sort the first page of subject sets? There will be some projects that will have 20+ subject sets (Notes from Nature I think? Gravity Spy will too eventually). I would suggest passing in the sort param into the API call instead of doing this.

app/pages/home-for-user/circle-ribbon.jsx
```diff
@@ -110,6 +112,14 @@ const CircleRibbon = React.createClass({
     this.props.onClick(clickedProject.id);
   },
 
+  calcLargeArc(classifications) {
+    if (classifications / this.state.totalClassifications >= .5) {
```
srallen: Could you change this to `0.5`?

app/pages/home-for-user/circle-ribbon.jsx
```diff
@@ -170,15 +181,17 @@ const CircleRibbon = React.createClass({
...
-            <image
-              xlinkHref={this.props.image}
-              x={this.props.weight + this.props.gap}
-              y={this.props.weight + this.props.gap}
-              width={imageSize}
-              height={imageSize}
-              clipPath={`url('#circle-ribbon-clip-${this.id}')`}
-              className={`url('#circle-ribbon-shadow-${this.id}')`}
-            />
+            <Link to={`/users/${this.props.user.login}/stats`}>
```
srallen: Since this is in a SVG, then I think you'll want to be using the `SVGLink` component here. Also it needs an aria label.

app/lib/get-user-ribbon-data.js
srallen: Oh I see, is it because of the added `include` param in the project preferences API call?

app/pages/home-for-user/index.jsx
```diff
@@ -120,10 +125,36 @@ const HomePageForUser = React.createClass({
...
   },
 
+  fetchRecentSubjects(ribbonData) {
+    const newestSubjects = []
+    ribbonData.map((project) => {
+      if (project['recentSubjectSet'] !== undefined){
+        newestSubjects.push(project)
+      }
+    });
+    const recentSets = newestSubjects.sort((a, b) => {
```
srallen: This sort can be done with the API, so doing this manually here isn't necessary. Also, it'll only be sorting the first page of subject sets returned.

srallen: Also, the naming here is a bit confusing... we're sorting subject sets, not subjects, correct?

srallen:
I agree the routes should be updated. I think `/home-for-user` was only intended to use for development.

srallen:
Just an aside, ESLint should have caught a lot of the missing semicolons, etc. I would suggest using that in your text editor or using on the command line as a check in the future.

app/lib/get-user-ribbon-data.js
wgranger: Yes. Needed to use `apiClient` to retrieve the included items.

app/lib/get-user-ribbon-data.js
wgranger: I could not for the life of me get subject sets to sort in the query. I tried `sort: '-updated_at'`, but to no avail. Am I missing something?

srallen: Ok, I double checked on Panoptes and `updated_at` needs to be added as a sortable field: https://github.com/zooniverse/Panoptes/blob/95671795f2790da789c16a35b4bb8fa1f677f637/app/serializers/subject_set_serializer.rb#L8

Could you open a issue ticket there for it?

wgranger: done

jwbmartin:
This is a polling of all those involved in this thread but do you think the sections that you open should extend down to the social media section or stay as is? I'm a little on the fence about which I prefer.
screen shot 2016-08-29 at 3 05 16 pm

css/home-not-logged-in.styl
```diff
@@ -19,7 +19,7 @@ $flex-section
   justify-content: center
   text-align: center
 
-.on-home-page-not-logged-in
+.on-home-page-not-logged-in, .on-home-page-logged-in
```
srallen: It's confusing to add this to `css/home-not-logged-in.styl`. I wouldn't expect to look for a style for a logged in state in a file named for not logged in. If it's styles that apply to both, then perhaps we should leverage the existing `home-page.styl` file? There's lots there that I imagine isn't being used anymore since both pages have now been rewritten.

css/buttons.styl
```diff
@@ -52,7 +52,7 @@
   padding: 0 0.8em 0 0.5em
   vertical-align: middle
 
-.on-home-page, .on-home-page-not-logged-in, .on-landing-page
+.on-home-page, .on-home-page-not-logged-in, .on-landing-page, .on-home-page-logged-in
```
srallen: Can we just use the `on-home-page` class and just make sure the home page root uses that class?

css/typography.styl
```diff
@@ -41,7 +41,7 @@ h6
   font-weight: bold
   margin: 0
 
-.on-home-page, .on-landing-page, .on-home-page-not-logged-in
+.on-home-page, .on-landing-page, .on-home-page-not-logged-in, .on-home-page-logged-in
```
srallen: Same comment here, can we just use `on-home-page` and make sure the root has that class applied?

css/home-not-logged-in.styl
```diff
@@ -354,8 +354,11 @@ $flex-section
 
             h3, p
               a
-                text-decoration: none
                 color: inherit
+                display: block
```
srallen: Are these styles only for the logged in state? If so, then these shouldn't be here since they'll also be applied when `home-page-not-logged-in` class is used in a parent element. I'd prefer these to be in the stylus file for the component they're used in. Maybe the `index.styl` in the `home-page-for-user` folder? Use BEM style naming there to be consistent, too, please.

srallen:
The bottom edge of the project card in the news pull out is hard to see:
screen shot 2016-08-30 at 4 34 21 pm

app/pages/home-for-user/news-section.jsx
```diff
@@ -0,0 +1,127 @@
...
+          <div className='news-pullout-section'>
+          <h4> Newest Project </h4>
+          {this.state.projects.map((project) => {
+            const avatarSrc = !!this.state.avatars[project.id] ? this.state.avatars[project.id].src : null;
+            return <ProjectCard key={project.id} project={project} imageSrc={avatarSrc} />;
+          })}
+          </div>
+
+          <div className='news-pullout-section'>
+            <h4> New Datasets </h4>
```
srallen: Is `Datasets` the best to use here? @mrniaboc What wording do you use to communicate to volunteers that there is new data? I don't think `Datasets` should be used since the API resource is called `Subject Sets` and especially if we don't use it anywhere else. IMO, this should be `New Data` or `New Subjects`.

srallen:
Maybe some margin below the `see all` button? Depending on the size of the window, it can be right up against the project cards: 
screen shot 2016-08-30 at 4 45 01 pm

app/pages/home-for-user/news-section.jsx
wgranger: This is something that could get ironed out a bit. We could call it Updated Projects and just display the most recently updated projects without having to make an extra API call for `subject_sets`. @mrniaboc  and I originally discussed having the most recent datasets appear, which would be tied to subject sets. Not sure what the terminology should be if we continue doing this.

srallen:
Project cards don't seem to be inheriting the bottom border radii. I think the details div will need the border radius styles defined. 

screen shot 2016-08-30 at 4 50 48 pm

jwbmartin:
@srallen good catch on the project card v. background. I had added a gray stroke around the card but now maybe adding some color to the background might help. Standby let me mess around with a few thoughts

css/home-not-logged-in.styl
wgranger: This is actually used for both components, so I can probably have the styling added to `on-home-page` instead, since I've now added it to the root.

wgranger:
This has been restaged at: https://new-signed-in-social-media.pfe-preview.zooniverse.org/

Also, the `/home-for-user` path has been removed, so just make sure you're signed in and you should see changes on the homepage.

srallen:
`publications.cjsx` should be just a js file, not cjsx, since there's no jsx in it. It's most of the way there with the syntax to make it vanilla js. Also, it should be moved into the `lib` folder, since it's not a component for a view or anything.

app/pages/home-for-user/news-pullout.jsx
```diff
@@ -0,0 +1,130 @@
...
+        projects,
+      });
+
+      return Promise.all(projects.map((project) => {
+        return project.get('avatar')
+        .catch(() => {
+          return null;
+        })
+        .then((avatar) => {
+          this.state.avatars[project.id] = avatar;
```
srallen: State should not be mutated directly like this: (see notes section about `setState`) https://facebook.github.io/react/docs/component-api.html

You can set the state of projects and avatars at the end of all of your promises.

app/pages/home-for-user/recent-projects.jsx
```diff
@@ -58,14 +63,17 @@ const RecentProjectsSection = React.createClass({
...
-          this.forceUpdate();
-        });
+        if (project.links.avatar.id) {
+          return apiClient.type('avatars')
+          .get(project.links.avatar.id)
+          .catch(() => {
+            return null;
+          })
+          .then((avatar) => {
+            this.state.avatars[project.id] = avatar;
```
srallen: Same comment as in the news component. State should not be directly mutated.

app/pages/home-for-user/generic-section.jsx
```diff
@@ -21,7 +22,7 @@ const HomePageSection = React.createClass({
             }}
           ></i>
           <span className="home-page-section__header-label">{this.props.title}</span>
-          <a href="#" className="secret-button" title="Close this section" aria-label="Close this section">×</a>
+          <Link to="#" className="secret-button" title="Close this section" aria-label="Close this section">x</Link>
```
srallen: I don't think making this a react-router link is necessary since functionally all this is doing is taking you back to the top of the page.

srallen:
I'm seeing in errors in IE11, but since it's being discussed that we may not support IE11, I'm going to leave that for now and it can be handled in a separate PR if we decide to continue to support. Just noting it here.

app/pages/home-for-user/generic-section.jsx
wgranger: This was changed from a simple anchor tag because the section would not close once navigating back to the homepage from another link. The `Link` component resolved this issue.

srallen:
Loading animation for the circle ribbon doesn't work in Edge, but I don't consider this as a blocking issue.

app/pages/home-for-user/generic-section.jsx
srallen: Huh, ok, that's odd, but if it works, then cool.

app/pages/home-for-user/news-pullout.styl
```diff
@@ -0,0 +1,46 @@
...
+    overflow-y: scroll
+    width: 100%
+
+  &__section
+    display: flex
+    flex-direction: column
+    margin-bottom: 1.5em
+    max-width: 220px
+
+    &--timestamp
```
srallen: The `--` in BEM is intended to indicate modifiers, so something like `.button--red` or `.button--blue`. In this case, these are elements of the section block. How I would approach this is to have section be its own block and timestamp, content, and title be its elements: `news-section__timestamp`. This is separate from the parent block, so the resulting div would have these classes: `home-page-news-pullout news-section__timestamp`

app/pages/home-for-user/recent-projects.jsx
```diff
@@ -39,13 +40,17 @@ const RecentProjectsSection = React.createClass({
     });
 
     user.get('project_preferences', {
-      page_size: 8,
+      page_size: 20,
```
srallen: The `page_size` default is 20, so this can probably be removed.

app/pages/home.cjsx
```diff
@@ -1,5 +1,5 @@
 React = require 'react'
-HomePageLoggedIn = require './home-logged-in'
+HomePageLoggedIn = require('./home-for-user').default
```
srallen: Try
```
`import HomePageLoggedIn from './home-for-user'`
```

app/pages/home-for-user/recent-projects.jsx
wgranger: @eatyourgreens Thoughts?

eatyourgreens: I have no idea what's going on here. @srallen is right, state should be changed via `this.setState`.

wgranger:
This has been restaged. The most recent commit `Fix Bug with Newest Project Link` is because items in the pullout don't have access to `browserHistory`. More info can be found here: https://github.com/reactjs/react-router/issues/3225. Open to another solution if one is found or if this solution might cause an issue down the road.

srallen:
Going to go ahead and merge this with Brian's existing PR and then rebasing that so @eatyourgreens and I can continue working on the remaining issues: mostly IE/Edge related as well as possibly improving the ribbon API calls.

PR #2920, opened 2016-08-26
Remove field guide default item

### LINKED ISSUE #2704: remove automatically generated field guide option
When you enable field guide on a project, it generates an entry (and thus displays by default on every project page). Can we either 1) turn off the default entry or 2) enable users to "hide" the field guide until they want it displayed? (This second option is especially important for existing projects that are now adding field guides and don't want the field guides "live" until they are finished.)

srallen:
Closes #2704


wgranger:
Looks good.

PR #2923, opened 2016-08-29
Removed gold standard importer and pulsar hunters feedback

### LINKED ISSUE #2847: Strip out gold standard UI
The gold standard UI in the project builder and the pulsar hunters feedback should be removed.

srallen:
Closes #2847. This removes gold standard importer out of the project builder. Project builders should instead use the export mode in the classification interface to create gold standard data or contact us about special cases (like Gravity Spy). Also removed is the custom Pulsar Hunters feedback.


camallen:
@srallen can this remove the classifier interface calls (and wiring) to get gold standard classification data too please? https://github.com/zooniverse/Panoptes-Front-End/blob/ee1ec4f21c6140a264e41b8f2800ef944f9b6213/app/classifier/index.cjsx#L85

srallen:
@camallen I started to go through and remove the rest of the gold standard classification UI that you pointed out. There are a few lines I'm not sure about:

- I'm not sure what this is for: https://github.com/zooniverse/Panoptes-Front-End/blob/6eba6cf2fcea4fd3dba738e5d75a7e63e2edb6f2/app/classifier/index.cjsx#L41
I don't see it used anywhere or changed anywhere.
- There was logic written for testing classification quality: https://github.com/zooniverse/Panoptes-Front-End/blob/6eba6cf2fcea4fd3dba738e5d75a7e63e2edb6f2/app/classifier/index.cjsx#L41
This calls a test in the lib folder, which then calls some logic dependent on the type of task, for example: https://github.com/zooniverse/Panoptes-Front-End/blob/6eba6cf2fcea4fd3dba738e5d75a7e63e2edb6f2/app/classifier/tasks/multiple.cjsx#L74-L88

Should all of this be stripped out too?

srallen:
Sorry, here's the link to the code about testing classification quality: https://github.com/zooniverse/Panoptes-Front-End/blob/6eba6cf2fcea4fd3dba738e5d75a7e63e2edb6f2/app/classifier/index.cjsx#L382-L386

camallen:
Re points

1. me either, if you can't see it being used then it's probably vestigal.
2. I have no idea but i assume it's from #1934

> When the classification is submitted, it's given a "quality" score (which is how closely it matches the gold standard). All task modules have a new testAnnotationQuality comparison function. These can probably use some iteration.

srallen:
I don't see an immediate usefulness, but wanted to double check since this has grown to touch a lot of files. I suppose if anyone is interested, they can go back and look through commit history.

srallen:
Relabeling as WIP until I finish it up.

srallen:
Staged at: https://fix-2847.pfe-preview.zooniverse.org/

@camallen I think I got all the cruft, but if you could look over, I'd appreciate it.

@mcbouslog Could you look over to make sure I didn't break anything with either regular or gold standard classification? We're leaving in the "expert mode" that a project owner can toggle on to classify. I tried it, and both types of classifications look ok to me, but would be good to have another person check.

mcbouslog:
this is great! so much cleaner =). code LGTM, tested workflows/classifying regularly, as owner in Expert Mode, and as expert (using a non-Admin user I have on preview) in Expert Mode. generated classification export CSV and Expert Mode appears to be properly saving Expert classifications, waiting for another classification export CSV to confirm additional testing, should receive shortly, and will then update this comment.

camallen:
@srallen looks ok to me, some of the task known vs submitted values looks like an interesting (task specific way) of comparing annotations for feeback...might want to remember the notion of these commits for the future as more projects require feeback on known subject answers.

srallen:
Agreed, but if we're going to be leveraging subject metadata instead of expert classification in the future, they all would need some tweaking and it's probably best for now to removing it and reference back if and when we decide to add in quality tests.

srallen:
@mcbouslog I've removed the slug. Good catch.

camallen:
@srallen agreed about removing as it's not used and we've got the commit history

Issue #2921, opened 2016-08-28
Last visited project pictures don't load on https://www.zooniverse.org/ 


yshish:
When I load https://www.zooniverse.org/ the four last visited project pictures don't show up. The browser console shows this error:

>Failed to load resource: the server responded with a status of 416 (Requested Range Not Satisfiable)

I tried load it signed in and then again after I signed out and signed in back again. The result was the same. 

I used incognito window in Chromium 48, Ubuntu.

yshish:
Oops, that error message has probably nothing to do with that as it mentions this video link: https://www.zooniverse.org/assets/home-video.ogv 

But it shows also these two errors:

Failed to load resource: the server responded with a status of 401 (Unauthorized) https://www.zooniverse.org/oauth/token 

Failed to get bearer token  
vendor-beb57659d335131a9500.js:12

camallen:
I think this has todo with projects that have been turned to private after you've beta tested them. We can do more filtering on this before we attempt to show the project cards.

yshish:
Not sure how you mean that, the project pictures which didn't load for me are for Penguin Watch, Chimp&See, Science Gossip, Snapshot Serengeti. When I display my personal stats (the ribbon) all project pictures are there, even these four.

I've just tried to make a classification to Plankton portal and reload the Zooniverse home page. The PP was the only one which displayed there, but the three other project pictures didn't.

camallen:
The last set of error messages are expected if you are signed out. 

How do you know which projects should be displayed if you can't see them....? Can you provide a screenshot perhaps?

eatyourgreens:
Sounds like an error fetching the project avatars. Recent projects makes separate API calls for those. 

I did a couple of classifications for Science Gossip and the picture shows up on my phone.

eatyourgreens:
![image](https://cloud.githubusercontent.com/assets/59547/18033514/4a754bda-6d1f-11e6-9efd-6d3f803c454f.png)

yshish:
This is what I see - it displays the project names but not the pictures:

https://drive.google.com/file/d/0B3yzHuLZuyBnSEU5UDFvMWdDZ0U/view?usp=sharing

eatyourgreens:
No project owner names there either, so I think `getDetails()` is failing in the Project Icon component. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/project-icon.cjsx

camallen:
[...]

PR #2928, opened 2016-08-30
Optimise recent projects

eatyourgreens:
Include avatars and owners in the initial API calls.
Update ProjectIcon to use cached resources if possible.

Supersedes #2882

### DIFF
--- a/app/components/project-icon.cjsx
+++ b/app/components/project-icon.cjsx
@@ -45,7 +45,6 @@ module.exports = React.createClass
       if !!@props.project.redirect
         <a href={@props.project.redirect} className="stats-project-icon">{content}</a>
       else
-        [owner, name] = @props.project.slug.split '/'
-        <Link to="/projects/#{owner}/#{name}" className="stats-project-icon">{content}</Link>
+        <Link to="/projects/#{@props.project.slug}" className="stats-project-icon">{content}</Link>
     else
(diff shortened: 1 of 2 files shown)


eatyourgreens:
This is an alternative to #2882, which removes the network requests from the `ProjectIcon` component but doesn't use the project cards API route.

Staged at https://optimise-recent-projects.pfe-preview.zooniverse.org

camallen:
i like this - i still think #2921 won't be fixed by this but it's a neat solution using resource links.  LGTM

eatyourgreens:
I'm not 100% sure, but I think adding the `include` param might actually force the client to make the request for the full project resource, rather than grabbing the partial resource that was cached by the cards request. Needs checking, and still a bit of a workaround.

wgranger:
Looks good. I think this is a clever way to make the API call more efficient. Just out of curiosity, what was the reason for adding `alt=''` in the most recent commit?

eatyourgreens:
@wgranger Yeah, `alt` is a required attribute. If it's missing, screenreaders will generally readout the image URL, which isn't massively helpful here. You can set it to an empty string if you judge an image to be purely presentational. Here, the project name is already present inside the link, so additional text seems unnecessary in order to understand where this link goes, so I set the `alt` to nothing.

PR #2931, opened 2016-08-30
Use default workflows

srallen:
Fixes some issue discussed in the front end channel.

This fixes the following scenarios:
- A logged out user clicks 'Get Started' and the project has a default workflow set, the default workflow id is now passed as a param. Before a random workflow was selected.
- A user navigates directly to a project's classify page. If logged in and there is a selected workflow passed down as a prop, then load that workflow.
- Do not show the workflow assignment dialog if the workflow query is undefined

This does not fix when a user browses directly to the classify page and is not logged in, then random workflows are still being loaded. I'm working on another branch to do some refactoring since I'm starting to dislike some of the dependency on the workflow query in the URL and there are redundant API calls for the workflow. This is a good stop gap until I'm done with that.

Staged: https://use-default-workflows.pfe-preview.zooniverse.org/

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -70,6 +70,7 @@ module.exports = React.createClass
   getDefaultProps: ->
     query: null
     project: null
+    selectedWorkflow: null
     simulateSaveFailure: (location?.search ? '').indexOf('simulate-classification-save-failure') isnt -1
 
   getInitialState: ->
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Sorry, missed this before. Looks good.

camallen:
@srallen it would be amazing to make the classifier component smarter about how it uses workflows and tutorials right now as it fetches (both) after each classification even if we're on the same workflow and those resources haven't changed (especially when lots of people visit the site).

srallen:
@camallen I'm working on a branch locally to eliminate redundant API calls for the workflow selection and to eliminate the remaining bugs with random workflows being selected when a project is configured to use a default. I'm refactoring to move the API request for the workflow into the project page controller component and passing it down as props since the navbar as well as the home page needs to know what the currently selected workflow is.

Tutorial is another to look at as well, but probably in another PR.

camallen:
brilliant!

PR #2936, opened 2016-08-31
Add content to how-to document on what happens when adding '!' to the…

mschwamb:
Add to the how to document a line or two to explain how to add a ‘!’ to the fieldname in the manifest file to make the column accessible only on Talk


parrish:
Looks good!

PR #2967, opened 2016-09-05
Add workflow configuration for cloning markers on multi-image subjects.

### LINKED ISSUE #2888: Cloning markers when marking in multi-image per subjects
Hi there

In our SpaceWarps demo site, several early testers have found it confusing that if they place a marker in one image it doesn't appear in all three images shown (either in multiple image view or flip-book mode). 

In our case all images per subject contain exactly the same area, so in this case placing a marker somewhere in one of the images applies to all images and the testers would have found it useful and less confusing if this marker once made was shown in all images.

Is there scope to automatically show a marker placed in one image in all images for that subject?

I realise this won't work for projects where the individual images for a subject are not of the same area.

Thanks!
Aprajita.

simoneduca:
Fixes #2888 

Removes `when parseInt(mark.frame) is parseInt(@props.frame)`, if the clone markers option is selected.

# Review Checklist

- [x] Does it work on mobile? Tested on iOS
- [x] Did you deploy a staging branch? https://fix-2888.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -35,50 +35,89 @@ module.exports = React.createClass
 
   render: ->
     skippedMarks = 0
-
     <g>
       {for annotation in @props.classification?.annotations ? []
+        
         annotation._key ?= Math.random()
         isPriorAnnotation = annotation isnt @props.annotation
         taskDescription = @props.workflow.tasks[annotation.task]
         if taskDescription.type is 'drawing'
-          <g key={annotation._key} className="marks-for-annotation" data-disabled={isPriorAnnotation || null}>
-            {for mark, i in annotation.value when parseInt(mark.frame) is parseInt(@props.frame)
-              mark._key ?= Math.random()
+          if @props.workflow.configuration.multi_image_tool
+            <g key={annotation._key} className="marks-for-annotation" data-disabled={isPriorAnnotation || null}>
+                {for mark, i in annotation.value
+                  mark._key ?= Math.random()
 
-              if skippedMarks < @props.classification._hideMarksBefore and not @props.classification.completed
-                skippedMarks += 1
-                continue
+                  if skippedMarks < @props.classification._hideMarksBefore and not @props.classification.completed
+                    skippedMarks += 1
+                    continue
 
-              if mark._copy?
-                continue
+                  if mark._copy?
+                    continue
 
-              toolDescription = taskDescription.tools[mark.tool]
+                  toolDescription = taskDescription.tools[mark.tool]
 
-              toolEnv =
-                containerRect: @props.containerRect
-                scale: @props.scale
-                disabled: isPriorAnnotation
-                selected: mark is @state.selection and not isPriorAnnotation
-                getEventOffset: @props.getEventOffset
-                preferences: @props.preferences
+                  toolEnv =
+                    containerRect: @props.containerRect
+                    scale: @props.scale
+                    disabled: isPriorAnnotation
+                    selected: mark is @state.selection and not isPriorAnnotation
+                    getEventOffset: @props.getEventOffset
+                    preferences: @props.preferences
 
-              toolProps =
-                classification: @props.classification
-                mark: mark
-                details: toolDescription.details
-                color: toolDescription.color
-                size: toolDescription.size
+                  toolProps =
+                    classification: @props.classification
+                    mark: mark
+                    details: toolDescription.details
+                    color: toolDescription.color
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -35,50 +35,89 @@ module.exports = React.createClass
...
+                    onSelect: @handleSelect.bind this, annotation, mark
+                    onDeselect: @handleDeselect
+                    onDestroy: @handleDestroy.bind this, annotation, mark
+
+                  ToolComponent = drawingTools[toolDescription.type]
+                  <ToolComponent key={mark._key} {...toolProps} {...toolEnv} {...toolMethods} />}
+            </g>
+          else   
+            <g key={annotation._key} className="marks-for-annotation" data-disabled={isPriorAnnotation || null}>
+                {for mark, i in annotation.value when parseInt(mark.frame) is parseInt(@props.frame)
```
eatyourgreens: If you change this to `when @props.workflow.configuration.multi_image_tool or parseInt(mark.frame) is parseInt(@props.frame)` then I think you can get rid of the outer `if` block and remove a lot of duplicated code here.

app/components/multi-image-subject-options-editor.cjsx
```diff
@@ -48,9 +48,17 @@ module.exports = React.createClass
             <input type="radio" id="multi_image_grid3" name="multi_image_layout" value="grid3" onChange={@handleSelectLayout} />
             <label htmlFor="multi_image_grid3">Grid (3col)</label>
           </div>}
+        <div>
+          <input type="checkbox" id="multi_image_clone_markers" name="multi_image_tool" onChange={@handleSelectTool} />
```
eatyourgreens: `name` should be the same as `id` here.

app/components/multi-image-subject-options-editor.cjsx
```diff
@@ -48,9 +48,17 @@ module.exports = React.createClass
...
+          <label htmlFor="multi_image_clone_markers">Clone markers in all frames</label>
+        </div>  
       </div>
     }</ChangeListener>
-
+    
+  handleSelectTool: (e) ->
```
eatyourgreens: This toggles whether marks are cloned or not, so maybe change the method name to `toggleCloneMarks`.

app/components/multi-image-subject-options-editor.cjsx
```diff
@@ -48,9 +48,17 @@ module.exports = React.createClass
...
+    @props.workflow.update
+      'configuration.multi_image_tool': e.target.checked
```
eatyourgreens: `multi_image_tool` doesn't really tell you what this configuration setting does. I'd change it to `multi_image_clone_markers`, to match the `id` of the checkbox.

app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -35,15 +35,14 @@ module.exports = React.createClass
...
     <g>
-      {for annotation in @props.classification?.annotations ? []
+      {for annotation in @props.classification?.annotations ? []      
         annotation._key ?= Math.random()
         isPriorAnnotation = annotation isnt @props.annotation
         taskDescription = @props.workflow.tasks[annotation.task]
         if taskDescription.type is 'drawing'
           <g key={annotation._key} className="marks-for-annotation" data-disabled={isPriorAnnotation || null}>
-            {for mark, i in annotation.value when parseInt(mark.frame) is parseInt(@props.frame)
+            {for mark, i in annotation.value when @props.workflow.configuration.multi_image_clone_markers or parseInt(mark.frame) is parseInt(@props.frame)
```
eatyourgreens: Sorry, one other change. The default for `props.workflow` is `null`, so this should be `@props.workflow?.configuration.multi_image_clone_markers` to avoid breaking errors when `workflow.configuration` doesn't exist.

simoneduca: Makes sense, thanks. It's worth noting that, by the same reasoning, we should probably change these lines as well: https://github.com/zooniverse/Panoptes-Front-End/blob/3f0142adc2a5cb4501da8164463ad888a56ae068/app/components/multi-image-subject-options-editor.cjsx#L27 and https://github.com/zooniverse/Panoptes-Front-End/blob/3f0142adc2a5cb4501da8164463ad888a56ae068/app/components/multi-image-subject-options-editor.cjsx#L41. 
The `?` should be after `workflow`, not `configuration`.

eatyourgreens:
Looks good, and works for me on Solar Stormwatch II.

PR #2974, opened 2016-09-06
Remove broken Gravity Spy gold standard check

### LINKED ISSUE #2963: 'done & Talk' button now showing for all projects even though the summary page is active and there is not supposed to be 'done & talk'
Now all my projects have 'Done & Talk' and 'Talk'.   I think that less people enter Talk with 'Done & Talk' projects  because volunteers don't know what it means so they hit Done because they want to move on quickly. My understanding was that 'Done & Talk' and 'Done' was only going to be an opt-in feature for projects that wanted quick review where Talk wasn't improved and that a project owner could have the summary page and the option to have 'done' and then the summary page where it would be 'talk' and 'next' 

 By having the summary and seeing 'next' or 'talk' got more people in Planet Four: Terrains Talk then we ever got into Planet Four which effectively has the same thing as the 'done & Talk' and 'done'  By having the summary page I think I get way more people into Planet Four: Terrains talk. 

Is this a bug? or were all summary pages taken away?

eatyourgreens:
Fixes #2963 .

Describe your changes.
This check returns true for almost all subjects, showing the 'Done & Talk' button on every project.

# Review Checklist

- [x] Does it work on mobile?

## Optional


simoneduca:
LGMT. Application shows done&talk button when classification summary is off.

PR #2981, opened 2016-09-06
Sugar-based Experiment Framework, and Volcrowe/Comet Hunters Experiment

alexbfree:
This adds a notification framework such that the Experiment Server can post interventions and experiment progress to Sugar, PFE can pick that data up, and act accordingly.

This adds an experiment for Volcrowe using the following projects at different stages:

1. Initial testing on staging - Planet Four Terrains (`mschwamb/planet-four-terrains`) => try it at http://sugar-notify-to-user.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/classify

2. Testing on production - Comet Hunters HSC Prototype (`mschwamb/hsc-comet-hunters-prototype`) => will happen automatically with the merge of this PR.

3. Going live on production - Comet Hunters (`mschwamb/comet-hunters`) => will be a couple of small tweaks to this code as a 2nd PR.

The experiment design is further detailed at https://docs.google.com/document/d/19-QDC5QQNW43JJYA4goV0NBTI1quvD6ayZ29JJOcVIo/edit

Latest version is deployed at http://sugar-notify-to-user.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-terrains/classify

Usernames:
CONTROL: username = alexzootest1 password = password
QUESTIONS: username = alexzootest3 password = password
STATEMENTS: username = alexzootest14 password = password

You can also try it on your own Zooniverse username and see what cohort you get assigned to.

Interventions are triggered periodically - you'll see one after 8 classifications or less, for questions and statements cohorts.

# Current status:

- [X] Update Sugar with new endpoints for Experiments (done by @parrish - Thanks!)
- [X] Update Experiment Server with new endpoints for Sugar Experiments
- [X] Add Experiment config to Experiment Server
- [X] Add some automated tests for experiment server / experiment flow
- [X] Add intervention monitor to PFE to listen for notifications from Sugar
- [X] Show an intervention in place of start of a new classification when prompted
- [X] PFE updates experiment server when end of classification or end of intervention
- [X] Ensure experiments ignored for other projects
- [x] Post data to Geordi when experiment progresses
- [x] Post question response data to Geordi
- [x] Deal with end of experiment state 
- [x] Insert calculated values to some of the messages
- [x] Add an opt-out mechanism (backend)
- [x] More testing - esp multi-session and end-to-end
- [x] Add styling (question box size, possibly containing box for whole intervention, etc)
- [x] Address feedback from Joe and Meg
- [x] Support opt out in front end

# Review Checklist

- [x] Does it work on mobile?

## Optional

No, mostly React. Little bit of CoffeeScript. I don't know ES6 and I haven't time to learn it before I leave Zooniverse next Friday - sorry!

N/A


app/classifier/index.cjsx
```diff
@@ -17,12 +17,17 @@ workflowAllowsSeparateFrames = require '../lib/workflow-allows-separate-frames'
 WorldWideTelescope = require './world_wide_telescope'
 MiniCourseButton = require './mini-course-button'
 GridTool = require './drawing-tools/grid'
+Intervention = require '../lib/intervention'
+experimentsClient = new (require '../lib/experiments-client')
+
+PULSAR_HUNTERS_SLUG = 'zooniverse/pulsar-hunters'
```
srallen: Might need to do a rebase with master. This line should be removed with this merged PR: https://github.com/zooniverse/Panoptes-Front-End/pull/2923

app/classifier/index.cjsx
```diff
@@ -156,80 +189,85 @@ Classifier = React.createClass
       onChange: -> classification.update()
 
     <div className="task-container" style={disabledStyle if @state.subjectLoading}>
-      {persistentHooksBeforeTask.map (HookComponent) =>
-        <HookComponent {...taskHookProps} />}
-
-      <TaskComponent autoFocus={true} taskTypes={tasks} workflow={@props.workflow} task={task} preferences={@props.preferences} annotation={annotation} onChange={@handleAnnotationChange.bind this, classification} />
+      {if @state.renderIntervention==true
```
srallen: This can just be `if @state.renderIntervention`

app/lib/experiments-client.coffee
```diff
@@ -0,0 +1,206 @@
...
+  getProjectSlugs: ->
+    @ensureEnvironmentSet()
+    switch @EXPERIMENT_SERVER_URL_TO_USE
+      when @EXPERIMENT_SERVER_PRODUCTION_URL
+        config.PROJECT_SLUGS_PRODUCTION
+      else
+        config.PROJECT_SLUGS_STAGING
+
+  lookupProjectSlug: (enabled_experiments, experiment_name) ->
+    this_project_slug for this_project_slug, experiments_for_this_slug of enabled_experiments when experiment_name in experiments_for_this_slug
```
srallen: camelCase is the convention in javascript. Could you use that here?

app/lib/experiments-client.coffee
```diff
@@ -0,0 +1,206 @@
...
+  postDataToExperimentServer: (interventionMonitor, geordi, experiment_name, user_id, session_id, type, id) ->
+    @ensureEnvironmentSet()
+
+    if not session_id?
+      session_id = getSessionID()
+
+    post_path = "/experiment/#{experiment_name}/user/#{user_id}/session/#{session_id}/#{type}/#{id}"
+    post_url = "#{@EXPERIMENT_SERVER_URL_TO_USE}#{post_path}"
+
+    request = new XMLHttpRequest()
```
srallen: We have the polyfill setup for fetch. It's the newer spec and it uses Promises: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API

srallen:
Just a few quick comments after a brief look over:
- I'm seeing a lot of snake case variable names. camelCase is the javascript convention.
- Generally speaking in javascript you should use `===` or `!==` in if statements. In coffeescript this is `is` or `isnt`
- It doesn't really make sense to me to use context for this. Can we just pull in the intervention client into the components where it's being used as a default prop?

package.json
```diff
@@ -9,6 +9,8 @@
     "data-uri-to-blob": "0.0.4",
     "debounce": "~1.0.0",
     "drag-reorderable": "~0.1.0",
+    "events": "~1.1.1",
+    "fast-levenshtein": "~1.0.7",
```
eatyourgreens: I don't think `fast-levenshtein` is meant to be added here.

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,70 @@
...
+                                                 type,
+                                                 intervention_id
+
+  render: ->
+    if @context.interventionMonitor?.latestFromSugar
+      experiment_name = @context.interventionMonitor?.latestFromSugar["experiment_name"]
+      user_id = @props.user.id
+      session_id = getSessionID()
+      intervention_id = @context.interventionMonitor?.latestFromSugar["next_event"]
+      intervention_details = experimentsClient.constructInterventionFromSugarData @context.interventionMonitor?.latestFromSugar
```
srallen: I don't think any of these need to be in the render. Instead I would suggest using a combination of props/state and use them directly in your handlers.

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,70 @@
+React = require 'react'
+InterventionMonitor = require './intervention-monitor'
+http = require 'http'
```
srallen: Doesn't look like this is being used in this component.

app/classifier/index.cjsx
```diff
@@ -51,8 +78,14 @@ Classifier = React.createClass
       @prepareToClassify nextProps.classification
 
     @context.geordi.remember subjectID: @props.subject?.id
+    @reCheckIfInterventionNeeded()
+
+  componentWillMount: () ->
+    @context.interventionMonitor.setProjectSlug @props.project.slug
 
   componentWillUnmount: () ->
+    @context.interventionMonitor.removeListener 'interventionRequested', => {}
```
eatyourgreens: The second argument here should exactly match the listener function that you want to remove. This code leaves your original event listener function in place even after the component has unmounted.

app/classifier/index.cjsx
```diff
@@ -51,8 +78,14 @@ Classifier = React.createClass
...
+    @context.interventionMonitor.removeListener 'classificationTaskRequested', => {}
```
eatyourgreens: Ditto.

app/classifier/index.cjsx
alexbfree: good spot, I will fix that. That would have been causing the listener to not deregister. 
I'm not sure it would enable me to remove the isMounted check though - as the intervention monitor can still call setState at a time when the component isn't mounted.

app/classifier/index.cjsx
alexbfree: yeah i had some confusion over that line and some other parts while rebasing, i know it's not quite right yet. will clean up before moving out of WIP state.

app/classifier/index.cjsx
alexbfree: True (!)

Sometimes I don't trust CoffeeScript so opt for being explicit ;-)

app/lib/experiments-client.coffee
alexbfree: Good point. I've been switching between Python, Ruby and CoffeeScript so much for the different components of this and sometimes forget where I am!

app/lib/experiments-client.coffee
alexbfree: Not familiar with polyfills or fetch - is there an example somewhere?

app/lib/intervention.cjsx
alexbfree: well spotted

app/lib/intervention.cjsx
alexbfree: 👍

package.json
alexbfree: Yeah i had a pain with that line in rebase, will fix before moving out of WiP state

app/lib/experiments-client.coffee
alexbfree: Never mind, Jim talked me through it and think I know what to do now. Thanks

alexbfree:
Hi @srallen  - good feedback on is, isnt, camelCase etc, thanks. Just a side effect of switching languages too many times per day - ruby conventions slipping through - will fix!!

Regarding context. I do think it makes sense for the intervention monitor to be always present. My rationale is that I want the intervention monitor to be running at all times. It's important that we do not miss the response that the experiment server posts to Sugar, even if you click around between different parts of the UI. In the future this would also make it easier for the interventions to be serviced at different times, not just each time you classify (for example, an intervention when you first view your collections, or when you go to the about page of the project, or post a message). Like Geordi logging, I see intervention monitoring as a cross-cutting concern. That's the way I see it anyway..

eatyourgreens:
I think @srallen was pointing out that the Intervention Monitor should be a singleton that you require in to components that use it, like the `apiClient`.

srallen:
@eatyourgreens Yes, it should just be required in the components that interventions are setup for. Context in React should be very rarely used and I don't want us to get in a pattern of putting data into it because it is outside of component state/props. It's also still considered an experimental feature which means it may change significantly or be removed in later versions of React.

https://facebook.github.io/react/docs/context.html#when-not-to-use-context

`defaultProps` in specific components really seem like a better use case here.

app/classifier/index.cjsx
```diff
@@ -156,80 +192,85 @@ Classifier = React.createClass
       onChange: -> classification.update()
 
     <div className="task-container" style={disabledStyle if @state.subjectLoading}>
-      {persistentHooksBeforeTask.map (HookComponent) =>
-        <HookComponent {...taskHookProps} />}
-
-      <TaskComponent autoFocus={true} taskTypes={tasks} workflow={@props.workflow} task={task} preferences={@props.preferences} annotation={annotation} onChange={@handleAnnotationChange.bind this, classification} />
+      {if @state.renderIntervention
```
srallen: If a project is setup for an intervention, is there a chance volunteers won't see a summary for a subject? If so, they miss a chance to go to the subject's talk page too. Why take the classification summary space for the intervention instead of a dialog for instance?

srallen:
Why was the experimental client moved into context? This seems unnecessary and a misuse of React context as it's only being used in the project classify page and the classifier index. It should be setup as a default prop on the project classify page and passed down to the child classifier component instead. 

Also, the latest commit doesn't address a question I had about the conditional for rendering the Intervention, so I'll move it here so it can be seen:

If a project is setup for an intervention, is there a chance volunteers won't see a summary for a subject? If so, they miss a chance to go to the subject's talk page too. Why take the classification summary space for the intervention instead of a dialog for instance?

app/classifier/index.cjsx
```diff
@@ -34,13 +38,35 @@ Classifier = React.createClass
...
+    @setState renderIntervention: @context.interventionMonitor?.shouldShowIntervention()
+    @context.interventionMonitor.on 'interventionRequested', @enableIntervention
+    @context.interventionMonitor.on 'classificationTaskRequested', @disableIntervention
     @loadSubject @props.subject
     @prepareToClassify @props.classification
     {workflow, project, preferences, user} = @props
     Tutorial.startIfNecessary {workflow, user, preferences}
 
+  reCheckIfInterventionNeeded: ->
+    # Sometimes, the intervention request arrives at a time when this component is not able to receive that
```
srallen: I don't think this comments apply anymore as you aren't checking if the component is mounted.

app/classifier/index.cjsx
```diff
@@ -49,10 +75,15 @@ Classifier = React.createClass
       @loadSubject subject
     if nextProps.classification isnt @props.classification
       @prepareToClassify nextProps.classification
+    @context.geordi?.remember subjectID: @props.subject?.id
+    @reCheckIfInterventionNeeded()
```
srallen: Did you intend to have this check every time props is received? If the client is going to be in the app context, you can check if there is a change in context using `nextContext`. However, I don't think context should be used.

app/classifier/index.cjsx
```diff
@@ -49,10 +75,15 @@ Classifier = React.createClass
       @loadSubject subject
     if nextProps.classification isnt @props.classification
       @prepareToClassify nextProps.classification
+    @context.geordi?.remember subjectID: @props.subject?.id
```
srallen: This will remember the current props subject id, not the nextProps... since this is in `componentWillReceiveProps` and there is no check happening, this is running every time new props is received and remembering current props. Not sure what the goal is here, but you might want `nextProps.subject?.id` instead after a conditional that checks `if @props.subject isnt nextProps.subject`

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,97 @@
...
+  componentDidMount: ->
+    interventionBody = document.getElementsByClassName("intervention-body")[0]
+    linksInInterventions = interventionBody.getElementsByTagName("a")
+    for link in linksInInterventions
+      link.addEventListener("click", @logLinkClick)
+
+  render: ->
+    <div className="intervention">
+      <h3 className="intervention-title">{@props.interventionDetails.title}:</h3>
+      <p dangerouslySetInnerHTML={@getBodyMarkup(@props.interventionDetails.body)} className="intervention-body"/>
```
srallen: Why use `dangerouslySetInnerHTML`? I don't see any HTML being used in the config file.

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,97 @@
...
+    @context.experimentsClient.logExperimentData @context.geordi, 'interventionLinkClicked', logData
+
+  componentDidMount: ->
+    interventionBody = document.getElementsByClassName("intervention-body")[0]
+    linksInInterventions = interventionBody.getElementsByTagName("a")
+    for link in linksInInterventions
+      link.addEventListener("click", @logLinkClick)
+
+  render: ->
+    <div className="intervention">
```
srallen: For semantic HTML, this should be a `<form>` tag.

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,97 @@
...
+        <textarea placeholder="Enter your answer here" className="intervention-question"/>}
+      <hr/>
+      <nav className="task-nav">
+        {if @props.interventionDetails.type is config.INTERVENTION_TYPES.STATEMENT
+          <button type="button" onClick={@endStatement} className="intervention-ok continue major-button">
+            <span>Continue</span>
+          </button>}
+        {if @props.interventionDetails.type is config.INTERVENTION_TYPES.QUESTION
+          <span>
+            <button type="button" onClick={@skipIntervention} className="intervention-cancel back minor-button">
```
srallen: Should be button type `submit`

app/lib/intervention.cjsx
```diff
@@ -0,0 +1,97 @@
...
+              <span>Skip this question</span>
+            </button>
+            <button type="button" onClick={@answerQuestion} className="intervention-submit continue major-button">
```
srallen: Should be button type `submit`

app/lib/intervention.cjsx
srallen: Ah I see a couple do have anchor tags in them.

app/classifier/index.cjsx
alexbfree: I think it's still necessary, isn't it the case that if no-one is listening to Sugar when the notification arrives, we still need to do this initial check to see if the intervention monitor was updated while this component wasn't loaded?

app/classifier/index.cjsx
alexbfree: Yeah I saw your other comment, will see about moving it out of context. To be honest I've been having some problems keeping a single, constant singleton class in existence without having it in context. I will try the defaultProps approach you suggested.

app/classifier/index.cjsx
alexbfree: good spot, that does sound wrong. will fix.

app/classifier/index.cjsx
eatyourgreens: Here's how the API client creates an instance that's shared with any component that imports it, if that helps. https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/api-client.js#L4

app/lib/intervention.cjsx
alexbfree: Yes, some of the interventions have links, I haven't yet thought of a better way to insert them since it's variable. I understand Jim's points that `<Link>` would be preferable but not sure how to achieve this with variable data. I guess i could build a react function that produces the bodies wrapped in a `<p>` or something....

app/lib/intervention.cjsx
alexbfree: 👍

app/lib/intervention.cjsx
alexbfree: 👍

app/lib/intervention.cjsx
alexbfree: 👍

app/classifier/index.cjsx
eatyourgreens: Be careful about using `geordi.remember` with a corresponding `geordi.forget`. Also, I think this is overwriting the subjectID that was stored when the subject loaded, and I'm not sure if that's your intention. https://github.com/zooniverse/Panoptes-Front-End/blob/909f9b824f276febb57b0d6a24c012185432ef77/app/classifier/index.cjsx#L53

app/lib/intervention.cjsx
eatyourgreens: Yeah, I'd guess that a JSX function along the lines of `() -> <span> My intervention text <Link to=url>some link text</Link></span>` would be needed here as the value of your config setting, so the config would probably have to be JSX.

alexbfree:
I know there are still some unanswered questions above which I will come back to.

On the question of moving experimentsClient interventionMonitor out of context, @eatyourgreens I had a look at what `apiClient` does... I can't exactly follow that model because unlike `apiClient`, the experiments client and intervention monitor both have state, so I need to ensure the same instance is used everywhere. I guess I just have to instantiate them in `/partials/app.cjsx` and then pass that instance down through all components that need it?

alexbfree:
Well, I learned something today. I didn't know you could use `require` in that way - I thought that exporting a `new` instance in a library file would result in a different instance for each place you `require` it. I guess the `require` gets worked out at runtime, it doesn't just import the code but rather the object, and passes the same one subsequently if it already exists. Awesome. Thanks @eatyourgreens

app/classifier/index.cjsx
alexbfree: Have improved the comment wording to be more accurate.

app/classifier/index.cjsx
alexbfree: The intent is that it ensures that subject-related clicks have the subject ID stored with them (prior to adding this line, they didn't). 

The data seems OK in Geordi.

I think we are covered here for the `forget`: https://github.com/zooniverse/Panoptes-Front-End/blob/909f9b824f276febb57b0d6a24c012185432ef77/app/classifier/index.cjsx#L57

app/classifier/index.cjsx
alexbfree: > Did you intend to have this check every time props is received? 

No, you're correct. I've moved it to the initial pre-render only.

alexbfree:
> If a project is setup for an intervention, is there a chance volunteers won't see a summary for a subject? If so, they miss a chance to go to the subject's talk page too. Why take the classification summary space for the intervention instead of a dialog for instance?

This is a really good point. We don't want to do that. Actually I've remembered, in my original design I intended it come in between classifications, so I suppose it should come before the first question, not before the summary.

I will look at an alternative design. I did consider `Dialog`s but didn't like the way it flowed. I will play around with it and try something else. I'm onto the final stages of styling and cosmetics now anyway so now is the best time for me to look at this.

srallen:
Maybe the before or after task hooks are something to consider? They're documented in the readme: https://github.com/zooniverse/Panoptes-Front-End#classifier-tasks

alexbfree:
OK I've changed the UI to use Dialogs. Which should address Sarah's concern as well as the feedback I've had from Joe and Meg about it being not distinct enough from normal workflow.

It now looks like this and doesn't disrupt the UX flow as it appears floating over the top:
screen shot 2016-09-14 at 16 47 08
screen shot 2016-09-14 at 16 49 14

alexbfree:
Almost there. I think I've addressed most of the code review comments. The main thing I have outstanding is to wire up the opt out link to the backend.

srallen:
I wonder if we're duplicating functionality with the "Interesting fact" version of the intervention with mini-courses. Interesting facts about the project seem like a good use case for a mini-course, which also has an opt-out feature. If you want to track it, we could look into adding Geordi to it.

alexbfree:
> I wonder if we're duplicating functionality with the "Interesting fact" version of the intervention with mini-courses. 

You may will be right, we could raise an issue for this as a future improvement as I don't have any bandwidth to investigate that option.

alexbfree:
OK @eatyourgreens @srallen I think this is now ready to merge. Thanks for all your valuable feedback and comments so far. Hopefully I've addressed everything to a "good enough" level. I've done lots of testing and everything seems to work.

Please bear in mind I have got no time left to do improvements, only critical issues that would prevent us from launching this. I would like to merge this a.s.a.p. (which will link with HSC Prototype on production). Once confirmed all good on production, we can switch it to the real Comet Hunters and send out launch emails etc.

Please do a final review and merge whenever you can. I'm redeploying the staging branch, you can test it with the credentials above. If you test opt out (you don't need to, I've tested it) let me know and I can reset the record in the Experiment Server DB afterwards so that user can be re-used for testing.

Thanks in advance!

alexbfree:
To make this easier to manage after I've left, I've created a new branch with all this code squashed into a single commit. I'll close this PR and open a new one momentarily.

alexbfree:
The new PR replacing this one with a single commit is #3042

PR #2984, opened 2016-09-07
Upgrade to React 15

eatyourgreens:
Updates other react-\* modules to latest versions too.
Depends on zooniverse/markdownz#47

# Review Checklist
- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://react-15.pfe-preview.zooniverse.org/
## Optional


shaunanoordin:
Seems OK on Win7+IE11

Tested OK:
* Logging in and logging out
* Viewing Home page, logged-in home page
* Browsing projects on Projects page
* Classifying in projects (basic radio button selection, circle marker, rectangle marker, basic transcribe, cropping)

eatyourgreens:
There are warnings on the Talk and inbox pages because MarkdownEditor uses `defaultValue` and `value` together on a textarea. React 15 will says you can have one, but not both, so we have to choose.

rogerhutchings:
Okay, looks like React 15 is pretty noisy when it comes to validating props and such, so there are a bunch of warnings elsewhere too - the site nav component. But I haven't found anything that doesn't work yet, so I reckon this is all good :)

rogerhutchings:
That said, let's not merge and deploy today

eatyourgreens:
Yeah, the `Link`, `Translate` and `PromiseRenderer` components pass their props through to their children, so will throw [unknown props warnings](https://facebook.github.io/react/warnings/unknown-prop.html) on HTML elements.

I've been going through cleaning up `...@props` when I've found it in the coffeescript. A lot of the time, it seems to be used to pass CSS classes and styles through to elements.

Speaking of which, should we move inline styles into the CSS files anyway? I don't like them because they're impossible to override.

eatyourgreens:
@srallen looks like `react-select 1.0` introduces breaking changes. I'm going to try rolling back to `0.9.1`

eatyourgreens:
#3233 updates `react-select`, which will fix the invalid peer warnings, when it's finished.

PR #2985, opened 2016-09-07
Remove React.findDOMNode

eatyourgreens:
`React.findDOMNode` was deprecated in React 14, and removed completely in 15.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


simoneduca:
LGTM.

PR #2979, opened 2016-09-06
New signed in home

srallen:
This picks up from Brian's previous WIP new signed in home page PR. 

Things left to do:
- Fix IE/Edge issues: ribbon loading animation doesn't work and the section pop ups don't work in IE.
- Any more ribbon loading optimization
- Final pass on code review
- Final ok from @jwbmartin on design implementation 

Staged at: https://new-signed-in-home.pfe-preview.zooniverse.org/
# Review Checklist
- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch?
## Optional
- [x] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?

PR #2998, opened 2016-09-08
Fix stats avatars

eatyourgreens:
Describe your changes.
Should fix project avatars not being shown on the user profile stats page.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://stats-avatars.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/classifications-ribbon.cjsx
+++ b/app/components/classifications-ribbon.cjsx
@@ -1,4 +1,5 @@
 React = require 'react'
+apiClient = require 'panoptes-client/lib/api-client'
 
 ClassificationsRibbon = React.createClass
   displayName: 'ClassificationsRibbon'
@@ -86,7 +87,7 @@ module.exports = React.createClass
     @setState loading: true
     @getAllProjectPreferences(user).then (preferences) =>
       projects = for preference in preferences
-        preference.get('project').catch =>
+        apiClient.type('projects').get(preference.links.project).catch =>
           null
       Promise.all(projects).then (projects) =>
         counts = for i in  [0...preferences.length] when projects[i]?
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Not the world's greatest fix: GET `avatars/avatar_id`, which succeeds if the avatar is cached, errors out otherwise. On error, fallback to `projects/project_id/avatar`.

This seems to be a workaround for what are probably bugs in the API client's implementation of resource caching.

srallen:
Just a thought, maybe we can absorb the `ClassificationRibbon` changes from #2979 into this to reduce the changes come in with that PR. @camallen suggest sounds like a good idea too.

wgranger:
I'm in the process of moving the `ClassificationRibbon` changes from the new home page over to this PR to make that merge easier. One note about sideloads I'm curious about. Yes, pulling the entire project into a call does slow things down a bit. We've been doing this with the new home page because it requires the `updated_at` tag from each project the user subscribes to in order to sort those projects, and this tag isn't available with project cards. 

Is there any change we can include the `updated_at` tag with the `cards: true` call? That way we could avoid pulling the entire project resource with a sideload. 

@camallen ideas?

eatyourgreens:
Maybe add that as a sort parameter to the cards API route? `sort=-updated_at`. I don't know if it supports that at the moment.

wgranger:
Wow, that worked, but we're also using the `updated_at` tag to show how recently each project was updated on the pull out tab, so it might be that we need the specific time/date each project was updated.

eatyourgreens:
After chatting to @camallen we're going to merge this, so that the volunteers aren't seeing broken avatars on their stats pages. I'm going to keep this branch open so that we can continue to work on optimising that API query, and merge those changes later.

@wgranger feel free to push to this branch and reopen when it's ready to merge again.

wgranger:
Gotcha. Yes, at this point I have all the changes moved over from the new signed in home page, just waiting to see how we want to approach the sideload.

camallen:
Feel free to merge this and we'll discuss the sideload in the new signed in PR

PR #3000, opened 2016-09-09
Don't scroll for links to document fragments

eatyourgreens:
Should fix markdown footnote links, discussed here on Talk: https://www.zooniverse.org/talk/18/87732?comment=164609&page=1
and tested here: https://www.zooniverse.org/talk/13/18624?comment=164558&page=2

Describe your changes.
Turns off scroll-to-top if the hash changed in the window location.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fragment-links.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -15,7 +15,9 @@ browserHistory.listen ->
 # make sure project stats page does not scroll back to the top when the URL changes
 shouldUpdateScroll = (prevRouterProps, routerProps) ->
   pathname = routerProps.location.pathname.split('/')
-  if ('stats' in pathname) and ('projects' in pathname)
+  isStats = ('stats' in pathname) and ('projects' in pathname)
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Here's a link for testing: https://fragment-links.pfe-preview.zooniverse.org/talk/13/18624?page=2&scrollToLastComment=true&env=production#fn1

srallen:
Seems to work well except in the case of having a hash in the location (I clicked on a footnote) and then I click to go to the next page of the discussion. I'd expect to scroll to top, but it doesn't. Other places on the site are working as expected: the pagination on the projects page, the stats page, and talk pagination without a hash in the location.

eatyourgreens:
Should be fixed now. I think it just needs to check if the new location hash isn't empty.

srallen:
Great 👍

PR #3009, opened 2016-09-12
Delete JSONEditor

rogerhutchings:
Deletes the JSONEditor component, which isn't being used by anything and has no references to it.


simensta:
Looks good -- thanks for cleaning this up @rogerhutchings!

PR #3018, opened 2016-09-14
Swap out History mixin in collections

rogerhutchings:
As the History mixin is deprecated as of `react-router` 2, I'm replacing it with `context.router` throughout. To keep the PRs small and easily testable, I'm creating them by area of function; this is for collections.

Related: #2705


simoneduca:
It works fine for me on Chrome, FF and Safari. I've staged it to https://remove-history-mixin-from-collections.pfe-preview.zooniverse.org/ in case anyone else wants to try it.

Issue #3014, opened 2016-09-13
Small bugs in Project Builder markings tasks


mkosmala:
## Expected behavior
~~1. Checkbox for "Clone markers in all frames" should stay checked once checked.~~
2. Min and Max boxes for tool types should have minimum of 0 and a way to select 'infinity' for Max. 

## Current behavior
~~1. On refresh, "Clone markers in all frames" becomes unchecked in the Project Builder interface (but task behavior remains).~~
2. Min and Max boxes can take on negative values. Once the Max box is changed to a non-infinity number, there's no way to change it back to 'infinity'. And having played with these boxes a bit, they have become stuck, and I can no longer change them. Buggy.

## Steps to replicate
~~1. Check "Clone markers in all frames". Refresh browser page.~~
2. Click the up and down buttons next to Min and Max a bunch. Click and change them by keyboard, too. At some point they'll stop behaving properly.

simoneduca:
Hi Margaret, thanks for catching this. I've fixed the checkbox bug, but the min-max values are still broken. That's a known issue and not trivial to fix. Also, in future, please submit one issue per bug; it makes it easier for us to keep track, cheers.

eatyourgreens:
@mkosmala #3031 should fix the problem with the min/max values. Do you need this urgently for one of your projects?

PR #3019, opened 2016-09-14
Remember clone marks checkbox value or default it to false

simoneduca:
Fixes partially #3014 .
Keep track of clone markers checkbox value or default it to false.


eatyourgreens:
👍

PR #3020, opened 2016-09-14
Swap out History mixin in Talk

rogerhutchings:
Swap out the History mixin for `context.router` in Talk.

Related to #3018

### DIFF
--- a/app/talk/latest-comment-link.cjsx
+++ b/app/talk/latest-comment-link.cjsx
@@ -21,6 +21,9 @@ module.exports = React.createClass
     discussion: React.PropTypes.object
     title: React.PropTypes.bool
     preview: React.PropTypes.bool
+
+  contextTypes:
+    geordi: React.PropTypes.object
     router: React.PropTypes.object.isRequired
 
   getDefaultProps: ->
@@ -31,9 +34,6 @@ module.exports = React.createClass
     commentUser: null
     latestCommentText: ''
     roles: []
-
-  contextTypes:
-    geordi: React.PropTypes.object
 
   logProfileClick: (profileItem) ->
     @context.geordi?.logEvent
--- a/app/talk/latest-comment-link.cjsx
+++ b/app/talk/latest-comment-link.cjsx
@@ -4,7 +4,7 @@ talkClient = require 'panoptes-client/lib/talk-client'
 {timeAgo} = require './lib/time'
 DisplayRoles = require './lib/display-roles'
 Avatar = require '../partials/avatar'
-{Link, History} = require 'react-router'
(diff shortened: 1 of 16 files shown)


app/talk/quick-subject-comment-form.cjsx
```diff
@@ -19,12 +18,14 @@ PAGE_SIZE = talkConfig.discussionPageSize
 
 module.exports = React.createClass
   displayName: 'TalkQuickSubjectCommentForm'
-  mixins: [State, History]
```
rogerhutchings: I couldn't see any of the State mixin methods used anywhere in this component...

eatyourgreens:
Getting this error when I try to run this branch.
`Failed propType: Required prop router was not specified in TalkLatestCommentComment. Check the render method of TalkBoardDisplay.`

app/talk/latest-comment-link.cjsx
```diff
@@ -16,13 +16,13 @@ truncate = (string = '', ending = '', length = 80) ->
...
 module.exports = React.createClass
   displayName: 'TalkLatestCommentComment'
-  mixins: [History]
 
   propTypes:
     project: React.PropTypes.object
     discussion: React.PropTypes.object
     title: React.PropTypes.bool
     preview: React.PropTypes.bool
+    router: React.PropTypes.object.isRequired
```
eatyourgreens: This should be in context, rather than props?

app/talk/latest-comment-link.cjsx
```diff
@@ -16,13 +16,13 @@ truncate = (string = '', ending = '', length = 80) ->
 
 module.exports = React.createClass
   displayName: 'TalkLatestCommentComment'
```
eatyourgreens: Should this be `TalkLatestCommentLink`?

app/talk/latest-comment-link.cjsx
rogerhutchings: it absolutely should

app/talk/latest-comment-link.cjsx
rogerhutchings: you'd think :troll:

eatyourgreens:
Cheers. There are tons of warnings about `createHref`. Should we update those calls too?

`Warning: the query argument to createHref is deprecated; use a location descriptor instead.`

rogerhutchings:
I think we probably should - leave that with me.

rogerhutchings:
@eatyourgreens On second thoughts, I'll fix the `createHref` calls in a separate PR - if you're happy with this, let's merge it now

PR #3022, opened 2016-09-14
Removed note about testAnnotationQuality

srallen:
I missed this in the PR for removing the gold standard classifications importer. The `testAnnotationQuality` function depended on gold standard classifications being loaded and as written won't work with that feature removed. At a later date, it may be of interest to look through the commit history and do a rewrite of these using gold standard subject metadata compared to a volunteer's classification instead. In the meantime, the readme shouldn't have a reference to it.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


srallen:
Gonna close since this may be added back in for specific projects.

PR #3029, opened 2016-09-14
add allow workflow query admin option

mcbouslog:
- add "allow workflow query" option as experimental tool in project's admin page
- add conditional for setting workflow that if workflow query in url and experimental feature enabled, then set workflow to query

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? [Staged Branch](https://allow-workflow-query.pfe-preview.zooniverse.org/).

## Optional

### DIFF
--- a/app/pages/admin/project-status.cjsx
+++ b/app/pages/admin/project-status.cjsx
@@ -26,6 +26,7 @@ EXPERIMENTAL_FEATURES = [
   'invert'
   'workflow assignment'
   'Gravity Spy Gold Standard'
+  'allow workflow query'
 ]
 
 ProjectToggle = React.createClass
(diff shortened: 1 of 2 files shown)


mcbouslog:
Not sure the best way for you to test, I edited my local NfN landing to point to my local PFE.

srallen:
I think you'll also want to use `onChangePreferences` to store the selection in the user's project preferences or in the temp guest preferences so the selection stays the same during the session. This is so if they navigate between talk and the classify, the selection they chose will still load. The selected workflow is stored in `preferences.selected_workflow` in the project preferences resource.

mcbouslog:
great catch! updated code and staged branch accordingly.

PR #3031, opened 2016-09-15
Fix min/max for drawing tools

### LINKED ISSUE #3014: Small bugs in Project Builder markings tasks

eatyourgreens:
Fixes #3014 .

Describe your changes.
Deletes max/min keys from a drawing tool when their value is an empty string.
Sets some sensible minimum values for min and max number of points.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-min-max.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/drawing/min-max-editor.jsx
+++ b/app/classifier/tasks/drawing/min-max-editor.jsx
@@ -10,35 +10,49 @@ const MinMaxEditor = React.createClass({
 
   getInitialState() {
     return {
-      min: null,
-      max: null,
+      tool: null,
     };
   },
 
   componentWillMount() {
-    this.setState({ min: this.props.choice.min, max: this.props.choice.max });
+    this.setState({
+      tool: this.props.choice,
+    });
   },
 
   componentWillReceiveProps(newProps) {
-    this.setState({ min: newProps.choice.min, max: newProps.choice.max });
+    this.setState({
+      tool: newProps.choice,
+    });
   },
 
   onChangeMin(e) {
-    this.setState({ min: e.target.value });
-    this.updateWorkflow(e.target.name, e.target.value);
+    const tool = this.state.tool;
+    if (e.target.value) {
+      tool.min = e.target.value;
+    } else {
+      delete tool.min;
+    }
+    this.updateWorkflow(tool);
   },
 
   onChangeMax(e) {
-    const newMax = e.target.value && e.target.value < this.state.min ?
-      this.state.min :
+    const tool = this.state.tool;
+    const newMax = e.target.value && e.target.value < this.state.tool.min ?
+      this.state.tool.min :
       e.target.value;
-    this.setState({ max: newMax });
-    this.updateWorkflow(e.target.name, newMax);
+    if (newMax) {
+      tool.max = newMax;
+    } else {
+      delete tool.max;
+    }
+    this.updateWorkflow(tool);
   },
 
-  updateWorkflow(name, value) {
+  updateWorkflow(tool) {
     const changes = {};
-    changes[name] = value;
+    changes[this.props.name] = tool;
+    this.setState({ tool });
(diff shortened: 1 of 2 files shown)


eatyourgreens:
I'd be much happier if this removed the `min` and `max` keys from a tool when the value is an empty string.

Also, it looks like you can set `max < min` and save those values, or set either value to a negative number and save.

simoneduca:
Is this still WIP or is it ready for review?

eatyourgreens:
Feel free to leave comments, but I wouldn't be happy merging this at the moment as I'm not sure what side effects would be produced by the changes to `handle-input-change`.

simoneduca:
I see. An idea would be to not using the `handle-input-change` at all by having the min-max settings in a little component with its own logic and event handling.

eatyourgreens:
I like that idea a lot better than changing the change handler for all project builder input fields. Maybe have our own number input component that validates its value before calling the change handler.

srallen:
I've come across the `handle-input-change` handler before and decided not to use it with a new feature I added. IMO, it's an abstraction that isn't helpful in all cases. It's inconsistently used in the code base and I'm of the opinion that it'd be better to have separate handlers/components for more specific use cases even if it means a bit of code repetition.

eatyourgreens:
@srallen Agreed. There's a component now with its own change handler. I still need to add some more validation of the fields, as I think it's still possible to save null or empty values, which might mess up the front-end display of the tool. However, it should stop you from entering a max value which is smaller than the min value.

eatyourgreens:
Updated to remove the min/max keys from the drawing tool if their values are empty strings. This should be ready now.

shaunanoordin:
Code looks fine to me, so here're the user tests.

Test Description
* We're testing two linked features:
* (1) Users can set a drawing tool's Min/Max values on admin, and these values can be blank.
  * We're testing four permutations of Min/Max values: blank/blank, 1/blank, blank/3, 1/3.
* (2) Once set, Min/Max values for the drawing tools are stated clearly on Classify interface
  * e.g. with text "X of 1 required, 3 maximum drawn" for Min/Max values of 1/3.
  * e.g. with text "X of 1 required" for Min/Max values of 1/blank.
* Known quirk: Users can manually type in negative numbers into Min or Max fields.
* Known quirk: Users can manually set Min value to be larger than Max value.

screen shot 2016-09-19 at 17 46 57
_Setting Min and Max values in the admin_

screen shot 2016-09-19 at 17 47 38
_The Min and Max values reflecting correctly on the Classify interface._

Test Results:
* OSX+Chrome53: OK
* Win7+IE11: OK, with a quirk...
  * Min/Max `<input>` boxes on the admin appear to be normal `type=text` boxes instead of `type=range`
* OSX+FF48: OK
* OSX+Safari9: OK

Status: LGTM! Will merge in 3 hours unless someone wants to add their comments.

eatyourgreens:
Thanks @shaunanoordin Updating the min value to larger than the max value should now update both.

shaunanoordin:
Status: Nice one! Update looks good to me.
Actions: Merging

PR #3033, opened 2016-09-15
Enable multi-images default view

### LINKED ISSUE #2708: Multiple images: default view
In the space warps demo https://www.zooniverse.org/lab/1434/workflow/960, we have multiple images and we let the user choose between flip book or separate image views.
When a new subject is loaded, we would like the default view to be the separate images rather than the flipbook - is it possible to add an option for builders to set which will be the default view?
If not, would someone be able to change this for SW?
Many thanks
Aprajita.

simoneduca:
Doesn't quite fix #2708 .

Describe your changes.
The issue is addressed but it results in a strange bug (see pic), I could do with some help with.

screen shot 2016-09-15 at 15 44 38

That's caused by the `renderFrame` method in the `subject-viewier` thinking there are 4 frames instead of 3. If you check the logs I put in you see a fourth frame called 'values', which shouldn't be there and I have not idea what it is.


eatyourgreens:
#3034 reports an extra frame, called `values`, showing up for some projects too.

eatyourgreens:
Fixed by #3045

simoneduca:
Nice one, thanks! This should be ready to go the, @eatyourgreens

app/components/frame-viewer.cjsx
```diff
@@ -29,6 +29,7 @@ module.exports = React.createClass
   render: () ->
     subject = @props.subject
     frame = @props.frame
+    console.log 'PROPS FRAME FROM SUBJECT-VIEWER', frame
```
eatyourgreens: console log

app/components/multi-image-subject-options-editor.cjsx
```diff
@@ -30,16 +30,20 @@ module.exports = React.createClass
...
           </select>
         </div>
-        {if @props.workflow.configuration?.multi_image_mode is 'separate' or @props.workflow.configuration?.multi_image_mode is 'flipbook_and_separate'
+        <div>
+          <input type="checkbox" id="enable_switching_flipbook_and_separate" name="enable_switching_flipbook_and_separate" checked={enableSwitchingChecked} onChange={@toggleEnableSwitching} />
+          <label htmlFor="enable_switching_flipbook_and_separate">Allow users to choose flipbook or separate frames</label>
+        </div>
+        {if @props.workflow.configuration?.multi_image_mode is 'separate'
           <div>
             <label>Show separate frames as</label><br/>
```
eatyourgreens: `label` here isn't associated with an input element. Wrap it around the input that it's labelling `<label><input/></label>` (preferable) or link it up using `for` and an ID.

app/components/subject-viewer.cjsx
```diff
@@ -79,20 +79,20 @@ module.exports = React.createClass
     {type, format, src} = getSubjectLocation @props.subject, @state.frame
     if @state.inFlipbookMode
       mainDisplay = @renderFrame @state.frame
+      console.log ' @state.frame',  @state.frame
```
eatyourgreens: console log

app/components/subject-viewer.cjsx
```diff
@@ -79,20 +79,20 @@ module.exports = React.createClass
...
         if not @state.inFlipbookMode or @props.subject?.locations.length < 2 or subjectHasMixedLocationTypes @props.subject
-          if @props.allowFlipbook and @props.allowSeparateFrames
+          if @props.workflow?.configuration.enable_switching_flipbook_and_separate
             <button className="secret-button" aria-label="Toggle flipbook mode" title="Toggle flipbook mode" onClick={@toggleInFlipbookMode}>
               <i className={"fa fa-fw " + if @state.inFlipbookMode then "fa-th-large" else "fa-film"}></i>
             </button>
         else
           <span className="tools">
-            {if @props.allowFlipbook and @props.allowSeparateFrames
+            {if @props.workflow?.configuration.enable_switching_flipbook_and_separate
```
eatyourgreens: Out of interest, why are we checking this twice (and rendering the button twice)?

app/lib/workflow-allows-flipbook.cjsx
```diff
@@ -1,5 +1,5 @@
 module.exports = (workflow) ->
```
eatyourgreens: This isn't a component, so the file should be `.coffee` instead of `.cjsx`.

app/lib/workflow-allows-separate-frames.cjsx
```diff
@@ -1,5 +1,5 @@
 module.exports = (workflow) ->
```
eatyourgreens: Not a component either.

app/components/subject-viewer.cjsx
simoneduca: If I understand the maze of ifs correctly here, the first check happens when flipbook is the default and the second when separate is the default.

eatyourgreens: Might be something to do with whether the play controls get rendered, I'm not sure. Anyway, it's outside the scope of this PR but looks like a bit of code that could be tidied up to avoid duplication.

PR #3036, opened 2016-09-15
Adds a notification counter to application context

### LINKED ISSUE #2951: notifications button doesn't unfll after you've read all your notifications
When you read your unread notifications the notifications icon stays filled which was confusing. If the page reloads or you navigate to zooniverse home and back then the icon unfills. If you stay within the project level since the page never reloads the icon stays filled which is confusing. Can you make the icon unfill when you hit the 'read' button.  Maybe for @parrish 

I tried this on Firefox on mac. 

I read my notifications and hit the button:

screen shot 2016-09-02 at 6 51 01 am

then I go to classify

screen shot 2016-09-02 at 6 50 52 am

icon is still filled so it looks like I have notifications when In don't 

 Also a comment on the new suggested form for filling out a github issue for this repo. Maybe for @srallen ?  It's geared towards bug fixes, but I thought this was also a place for feature requests. I'm not sure how I would fill out the suggested form to ask for new features not available because there is no expected behavior. Where do feature requests go?

parrish:
Fixes #2951.

This moves notification counting into a separate class and adds that to the app context.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://add-notifications-context.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/lib/notifications-counter.js
+++ b/app/lib/notifications-counter.js
@@ -28,7 +28,7 @@ class NotificationsCounter {
 
   setSection() {
     if(this.owner && this.name) {
-      let slug = `${this.owner}/${this.name}`;
+      const slug = `${this.owner}/${this.name}`;
 
       if(this.slug !== slug) {
         this.slug = slug;
--- a/app/lib/notifications-counter.js
+++ b/app/lib/notifications-counter.js
@@ -1,0 +1,83 @@
+import apiClient from 'panoptes-client/lib/api-client';
+import talkClient from 'panoptes-client/lib/talk-client';
+
+class NotificationsCounter {
+  constructor() {
+    this.callbacks = [];
+  }
+
+  update(user, owner, name) {
+    if(!user) {
+      return Promise.resolve(0);
+    }
+
+    if(this.loading) {
+      return;
+    }
+
+    this.loading = true;
+
+    this.owner = owner;
+    this.name = name;
+    this.setSection().then((section) => {
+      this.countUnread().then(() => {
+        this.loading = false;
+      });
+    });
+  }
+
+  setSection() {
+    if(this.owner && this.name) {
+      let slug = `${this.owner}/${this.name}`;
+
+      if(this.slug !== slug) {
+        this.slug = slug;
+        return this.getProject(slug).then((projects) => {
+          return this.section = `project-${projects[0].id}`;
+        });
+      } else {
+        return Promise.resolve(this.section);
+      }
+    } else {
+      this.owner = this.name = this.slug = this.section = null;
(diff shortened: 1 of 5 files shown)


parrish:
Yeah, receiving multiple types of notifications (e.g. "somebody replied to a thread you follow" and "somebody mentioned you") is a known issue.  Never did come up with a good resolution to that.

app/partials/app.cjsx
```diff
@@ -12,35 +13,53 @@ PanoptesApp = React.createClass
...
+
     auth.listen 'change', @handleAuthChange
     generateSessionID()
     @handleAuthChange()
 
   componentWillUnmount: ->
     auth.stopListening 'change', @handleAuthChange
 
+  componentWillReceiveProps: (nextProps) ->
+    @updateNotificationsCount params: nextProps.params
```
parrish: Insofar as I know, props only changes when the route changes in this component.  Unless I'm wrong, then that's what the goal is.

app/partials/app.cjsx
```diff
@@ -12,35 +13,53 @@ PanoptesApp = React.createClass
...
     initialLoadComplete: @state.initialLoadComplete
     user: @state.user
     geordi: @geordiLogger
+    notificationsCounter: @state.notificationsCounter
+    unreadNotificationsCount: @state.unreadNotificationsCount
 
   getInitialState: ->
     initialLoadComplete: false
     user: null
+    notificationsCounter: new NotificationsCounter()
```
parrish: Happy to move it to props.

app/lib/notifications-counter.js
```diff
@@ -0,0 +1,83 @@
...
+    this.setSection().then((section) => {
+      this.countUnread().then(() => {
+        this.loading = false;
+      });
+    });
+  }
+
+  setSection() {
+    if(this.owner && this.name) {
+      let slug = `${this.owner}/${this.name}`;
```
parrish: Yep.  Changing

parrish:
Updated.  Thanks!

app/partials/app.cjsx
srallen: Right, that's the only props, but I guess I'm just thinking about if in the future there are any other props, we might want an explicit check, but that doesn't need to happen now.

app/partials/app.cjsx
srallen: I think I'd prefer it as props unless anyone has a good reason it shouldn't be.

parrish: Yep, the counter changed to a prop in a70a3aba.  The unread count is still a state, but that shouldn't need to change.

PR #3041, opened 2016-09-16
Expert comparison summary

srallen:
Related to #2923. This adds back the expert classification comparison summary part of the previous gold standard classification import feature, but now behind a feature flag. I've turned on the feature for MWP and Microscopy Masters. It can be tested with the Pickers Practice workflow on Microscopy Masters:

https://expert-comparison-summary.pfe-preview.zooniverse.org/projects/jbrugg/microscopy-masters/?env=production

Note: This does not add back in the classification quality calculations which would involve touching a lot more of the code base to add back in.

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -289,14 +289,6 @@ Classifier = React.createClass
             <button type="button" onClick={@toggleExpertClassification.bind this, false}>Hide</button>
           else
             <button type="button" onClick={@toggleExpertClassification.bind this, true}>Show</button>}
-
-          {unless true or isNaN @state.classificationQuality
-            qualityString = (@state.classificationQuality * 100).toString().split('.')[0] + '%'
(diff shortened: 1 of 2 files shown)


app/classifier/index.cjsx
```diff
@@ -64,14 +67,38 @@ Classifier = React.createClass
...
+        workflow_id: workflow.id,
+        subject_ids: [subject.id]
+      })
+        .catch ->
+          []
+        .then ([expertClassification]) ->
+          expertClassification
+
+    awaitExpertClassification.then (expertClassification) =>
+      expertClassification ?= subject.expert_classification_data?[workflow.id]
```
mcbouslog: this looks good, just more a question for my general understanding - we're trying to get the expert classification from the API above (line 87), but if the response is null or undefined, we try to get it from the subject itself, specifically an `expert_classification_data` property?

PR #3052, opened 2016-09-19
edit workflows request if query

### LINKED ISSUE #3043: Test Workflow button on inactive workflows issue
## Expected behavior
Test Workflow button clicked on inactive workflow allows preview of workflow.

## Current behavior
Doesn't load, only getting active workflows from API.

mcbouslog:
Fixes #3043.

Also fixes long load time from NfN expedition card to classifying interface. Long load time was caused by loading of all active workflows, which in NfN's case (and I think currently unique to NfN) includes many large workflows.

- added conditional to workflow load where if workflow query and either the project accepts queries (like NfN) or user is admin, owner or collaborator (clicked Test Workflow button) then only load that specific workflow

To load one workflow it might be simpler to use `apiClient` and bypass `getAllWorkflows` and `getSelectedWorkflow`, but I thought it would be preferred to keep the refactored flow? I have a different branch that does utilize `apiClient`, if that would be preferred.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? [https://workflow-query.pfe-preview.zooniverse.org](https://workflow-query.pfe-preview.zooniverse.org)

## Optional

mentioning @srallen, @eatyourgreens

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -9,6 +9,7 @@ apiClient = require 'panoptes-client/lib/api-client'
 {sugarClient} = require 'panoptes-client/lib/sugar'
 classNames = require 'classnames'
 getWorkflowsInOrder = require '../../lib/get-workflows-in-order'
+isAdmin = require '../../lib/is-admin'
 
 counterpart.registerTranslations 'en',
   project:
@@ -87,8 +88,12 @@ ProjectPage = React.createClass
       @context.geordi?.remember projectToken: nextProps.project?.slug
 
     # Only call to get workflow if we know if there is a user or not and the project is finished loading
-    if nextContext.initialLoadComplete and not nextProps.loading and nextProps.preferences and @state.activeWorkflows.length is 0
-      @getAllWorkflows(nextProps.project)
+    if nextContext.initialLoadComplete and not nextProps.loading and nextProps.preferences and @state.activeWorkflows.length is 0 and not @state.loadingSelectedWorkflow
+
+      if nextProps.location.query?.workflow? and ('allow workflow query' in nextProps.project.experimental_tools or @checkUserRoles(nextProps.project, nextProps.user))
+        @getAllWorkflows(nextProps.project, { id: nextProps.location.query.workflow })
+      else
+        @getAllWorkflows(nextProps.project)
 
     if nextProps.preferences?.preferences? and @state.selectedWorkflow?
       if nextProps.preferences?.preferences.selected_workflow isnt @state.selectedWorkflow.id
@@ -118,10 +123,9 @@ ProjectPage = React.createClass
       .then (pages) =>
         @setState {pages}
 
-  getAllWorkflows: (project) ->
+  getAllWorkflows: (project, query = { active: true }) ->
     @setState { loadingSelectedWorkflow: true }
-
-    getWorkflowsInOrder(project, { active: true })
+    getWorkflowsInOrder(project, query)
       .then (activeWorkflows) =>
         @setState { activeWorkflows }
       .then =>
@@ -130,9 +134,9 @@ ProjectPage = React.createClass
         @getSelectedWorkflow(project, @props.preferences)
 
   getSelectedWorkflow: (project, preferences) ->
-    # preference user selected workflow, then project owner set workflow, then default workflow
+    # preference workflow query, then user selected workflow, then project owner set workflow, then default workflow
     # if none of those are set, select random workflow
-    if @props.location.query?.workflow? and 'allow workflow query' in @props.project.experimental_tools
+    if @props.location.query?.workflow? and ('allow workflow query' in @props.project.experimental_tools or @checkUserRoles(project, @props.user))
       preferredWorkflowID = @props.location.query.workflow
       unless preferences?.preferences.selected_workflow is preferredWorkflowID
         @props.onChangePreferences 'preferences.selected_workflow', preferredWorkflowID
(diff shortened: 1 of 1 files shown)


app/pages/project/index.cjsx
```diff
@@ -87,8 +88,12 @@ ProjectPage = React.createClass
       @context.geordi?.remember projectToken: nextProps.project?.slug
 
     # Only call to get workflow if we know if there is a user or not and the project is finished loading
-    if nextContext.initialLoadComplete and not nextProps.loading and nextProps.preferences and @state.activeWorkflows.length is 0
-      @getAllWorkflows(nextProps.project)
+    if nextContext.initialLoadComplete and not nextProps.loading and nextProps.preferences and @state.activeWorkflows.length is 0 and not @state.loadingSelectedWorkflow
```
camallen: Minor, I'd move this to a well named function so one doesn't have to interpret all the boolean state to understand this conditional

app/pages/project/index.cjsx
```diff
@@ -130,9 +134,9 @@ ProjectPage = React.createClass
         @getSelectedWorkflow(project, @props.preferences)
 
   getSelectedWorkflow: (project, preferences) ->
-    # preference user selected workflow, then project owner set workflow, then default workflow
+    # preference workflow query, then user selected workflow, then project owner set workflow, then default workflow
```
camallen: I'm not a huge fan of commenting like this as I should be able to read the code to understand this. That said if it's common practice in PFE then....

app/pages/project/index.cjsx
eatyourgreens: Ditto what Cam said. The logic here is hard to understand. Is it possible to at least separate out the loading code somehow so that we don't have to worry here about whether or not the component is loading data?

srallen: Yep, this is the logic that I mentioned needed further refactor in my PR. These checks are to prevent fetching the workflow multiple times while we're waiting for the initial user fetch and the project fetch since it's in `componentWillReceiveProps` part of the lifecycle. What I'd like to do is move the workflow fetch into the set of promises in `fetchProjectData` here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/index.cjsx#L322

I don't think I would hold up this PR for that though. Extracting it out into its own function is a good step toward that.

app/pages/project/index.cjsx
srallen: Yeah there are comments all over the classifier, so I left a few myself. I could go either way on it.

eatyourgreens:
@camallen should we consider offloading random selection to the API?

Or does PFE switch workflows randomly per classification? (Personally, I think this is a really bad idea as it makes classifying harder)

camallen:
it should select the workflow and store it till the page reloads / the user chooses another one. Don't think we should offload this state selection to the API as it's per session and has to respect the user choice too.

eatyourgreens:
Ok. I was thinking that requesting all workflows just to then pick one out of the array at random might be a pain point for the API if traffic is high on a project. Depends on the number of workflows we base our selection on, I guess. Like, I wouldn't want to do this for an NfN project.

camallen:
Ah right, we could do this for sure. I guess it depends on how the front end is wired up (does it show them all or just a get started button). I'd be happy to implement the api to a random choice from an array if we need it.

srallen:
What is shown to the user depends on the project configuration, if they allow the user to choose workflows and if they have a default workflow set. If both of those are false, then the user sees "Get Started" and the workflow is randomly selected. If a user is allowed to choose but goes directly to the classify page or clicks on the classify link in the nav bar and there isn't a default workflow, then the workflow is randomly chosen. 

The classifier used to fetch a new random workflow when the subject queue was empty, but it does not anymore. I also agree that's a bad idea.

It'd be great if both project completeness and random workflow selection is api side, so then the only reason to fetch all workflows would be if the project had enabled the allow user to choose option for the home page. Then, that request could be scoped only to the home page component and that config setting with some kind of pagination if applicable rather than fetching at the project index.

eatyourgreens:
Am I right in thinking that random selection = PFE requests all workflows, then selects one from that array and throws the rest away?

srallen:
@eatyourgreens yes, a random index is chosen and then that workflow id is passed along to fetch that specific workflow. The array of workflows is stored in the state since the project completeness function uses them too, so they aren't thrown away, but I think project completeness is something the api could do also.

camallen:
@srallen for reference we have a project completeness metric [here](https://github.com/zooniverse/Panoptes/blob/622b4f20f4824f2082c3c5577e6b10134b3d0dd3/app/workers/calculate_project_completeness_worker.rb#L15) on the project that only considers the active workflows, however it depends on the linked subjects / retired counts being aligned on the launch dates (think pre-launch retirement), maybe we just have to make this more robust instead of summiting the active workflows?

srallen:
I'll have to take a look at that. I'd prefer we leverage project states that the API is giving us. The front end is just checking the `finished_at` prop on active workflows linked to the project to set a local component state for completeness. But that refactor makes sense to come in a separate PR. I'll finish reviewing this as it is since I think NfN need to be able to test inactive workflows.

camallen:
Understood - none of this is blocking in my mind but we could be better with our resource usage.

eatyourgreens:
No objections from me either.

mcbouslog:
Interesting comments! Thanks all, I'm learning just from following the conversation. And thank you Sarah for reviewing and merging!

PR #3065, opened 2016-09-22
adding in line break needed in markdown help

mschwamb:
forgot to add the line break so the two examples show up on the same line for superscript and subscript. This PR fixes that.


eatyourgreens:
Thanks Meg!

PR #3070, opened 2016-09-26
Remove 100% height for small screens

### LINKED ISSUE #3068: Zooniverse lab page looks garbled on iphone
Not sure if this  on ipad but on my iphone (OS 10.0.2) I see that the Zooniverse lab page is all garbled.

image1

eatyourgreens:
Fixes #3068 .

Describe your changes.
Removes the `height: 100%` that was applied to `.landing-page`, which breaks the page layout when the content is longer than the viewport height.

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://fix-3068.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/css/landing-page.styl
+++ b/css/landing-page.styl
@@ -5,12 +5,11 @@
   color: #f9f9f9
   display: flex
   flex-direction: column
-  height: 100vh
+  min-height: 100vh
   justify-content: center
   padding: 2.5em 3vw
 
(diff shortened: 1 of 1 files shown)


eatyourgreens:
@srallen Rebased, and I've replaced `height` with `min-height`, which should fix the overlap with the footer.

PR #3074, opened 2016-09-27
Update user details for latest comment links

### LINKED ISSUE #3067: Usernames and avatar don't change on Talk front page when you switch from one Talk to another through the Talk side bar
cc @parrish 

Here's P4: Terrains
screen shot 2016-09-25 at 10 33 41 am

and then I  went to Microscopy Masters through the Talk side bar  and it has the P4: T usernames and avatars from its front page listed in the Microscopy

screen shot 2016-09-25 at 10 33 30 am
 Masters. I haven't ever posted on their talk.

eatyourgreens:
Fixes #3067 .

Describe your changes.
Updates the latest comment link state if the comment props change.
Also rips out a PromiseRenderer that was requesting user roles from the API on every re-render.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3067.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/talk/latest-comment-link.cjsx
+++ b/app/talk/latest-comment-link.cjsx
@@ -4,7 +4,6 @@ talkClient = require 'panoptes-client/lib/talk-client'
 {timeAgo} = require './lib/time'
 DisplayRoles = require './lib/display-roles'
 Avatar = require '../partials/avatar'
-PromiseRenderer = require '../components/promise-renderer'
 {Link, History} = require 'react-router'
 {Markdown} = (require 'markdownz').default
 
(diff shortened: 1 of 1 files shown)


rogerhutchings:
It looks like it's solved the filed bug, so that's cool. But on a related issue, I've found I can keep whole boards around:

1. Go to https://fix-3067.pfe-preview.zooniverse.org/projects/alex-panoptes/test-proj/talk
1. There are two boards
1. Navigate to https://fix-3067.pfe-preview.zooniverse.org/projects/drrogg/annotate/talk via the link in the bottom right bar
1. Those two boards are still there
1. Refresh, and they go

This may be just because the AnnoTate board is empty, so if you reckon this is a separate issue I can create one.

eatyourgreens:
@rogerhutchings Good catch. `componentWillReceiveProps` wasn't checking if the comment existed. Should be fixed now

PR #3078, opened 2016-09-28
Only Display Workflows with Subject Sets

### LINKED ISSUE #1933: Don't let users pick a workflow if it does not have an associated subject set
It might be useful if the front end checks if a workflow has an associated subject set before it is added to the list of choices volunteers can pick from when "Volunteers can choose which workflow they work on" is checked.  That way users will only be able to pick from workflows that would not instantly give an error message if the project builder forgot to link a subject set.

Also related to [this talk post](https://www.zooniverse.org/talk/18/18584/?scrollToLastComment=true&page=1) as a way to "hide" a workflow.

wgranger:
Fixes #1933  .

Describe your changes.
From a project home page, users can only select a workflow if an associated subject set is available.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/home.cjsx
+++ b/app/pages/project/home.cjsx
@@ -57,14 +57,15 @@ module.exports = React.createClass
             </Link>
         else if @props.project.configuration?.user_chooses_workflow
           @props.activeWorkflows.map (workflow) =>
-            <Link
-              to={"/projects/#{@props.project.slug}/classify"}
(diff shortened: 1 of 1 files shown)


srallen:
Gonna close this based on the above comments and since it will take some more work. Probably a good task for a week we're working on PFE.

PR #3079, opened 2016-09-28
Remove history mixin.

simoneduca:
Also remove annoying warning `browser.js?030c:49 Warning: [react-router] props.history and context.history are deprecated. Please use context.router.`

### DIFF
--- a/app/pages/projects.cjsx
+++ b/app/pages/projects.cjsx
@@ -17,9 +17,6 @@ ProjectsPage = React.createClass
 
   mixins: [TitleMixin]
 
-  contextTypes:
-    location: React.PropTypes.object
-
   emptyPromise:
     then: ->
     catch: ->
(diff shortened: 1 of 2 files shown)


app/pages/projects.cjsx
```diff
@@ -38,7 +35,7 @@ ProjectsPage = React.createClass
         delete query[key]
     newLocation = Object.assign {}, @props.location, {query}
     newLocation.search = ""
-    @props.history.replace newLocation
```
eatyourgreens: Should `history` be removed wherever it's being passed in as a prop too?

rogerhutchings: I think it should. It's passed in only for route-level components, but it's still deprecated.

rogerhutchings: The only other thing on this - we're replacing `replace` with `push`, but I prefer it that way anyway.

simoneduca: What Rog said :)

PR #3080, opened 2016-09-28
Remove PromiseRenderer from Talk components

eatyourgreens:
Describe your changes.
This makes a start at removing the PromiseRenderer component from Talk.
# Review Checklist
- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://talk-promise-renderer.pfe-preview.zooniverse.org
## Optional

### DIFF
--- a/app/subjects/mention-list.cjsx
+++ b/app/subjects/mention-list.cjsx
@@ -36,7 +36,12 @@ module.exports = React.createClass
         <h2>Mentions:</h2>
         <div>
           {for mention, i in @state.mentions
-            <Comment key={"mention-#{mention.comment.id}-#{i}"} data={mention.comment} locked={true} linked={true} />}
+            <Comment
+              key={"mention-#{mention.comment.id}-#{i}"}
+              data={mention.comment}
+              locked={true}
+              linked={true}
+              params={@props.params} />}
         </div>
       </div>
     else
--- a/app/talk/data-export-button.cjsx
+++ b/app/talk/data-export-button.cjsx
@@ -1,20 +1,28 @@
 React = require 'react'
-PromiseRenderer = require '../components/promise-renderer'
 talkClient = require 'panoptes-client/lib/talk-client'
 moment = require 'moment'
 
 module.exports = React.createClass
   displayName: 'DataExportButton'
+  
+  componentWillMount: ->
+    @exportGet()
+  
+  componentWillReceiveProps: (newProps)->
+    @exportGet() if newProps.exportType isnt @props.exportType
 
   getInitialState: ->
     isAvailable: null
     exportRequested: false
     exportError: null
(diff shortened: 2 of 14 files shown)


parrish:
Looks good 👍 .  I made a few small changes to fix some regressions that were introduced

- Comments weren't consistently being passed `props.params` (used in `commentSubjectTitle`)
- The data-export-button didn't update state after creating the default subject board

eatyourgreens:
Thanks for catching those bugs. I've updated the comment mentions too so that the component keys aren't dependent on the array order.

app/subjects/mention-list.cjsx
```diff
@@ -35,13 +35,14 @@ module.exports = React.createClass
       <div>
         <h2>Mentions:</h2>
         <div>
-          {for mention, i in @state.mentions
+          {@state.mentions.map (mention) ->
             <Comment
-              key={"mention-#{mention.comment.id}-#{i}"}
+              key={"mention-#{mention.comment.id}"}
```
parrish: It needed the index.  When there are multiple mentions per comment they will have the same key.

eatyourgreens: Ah right. I've deleted that commit but I'm uneasy about the dependency on array order here. Using the array index as an identifier for each mention seems brittle to me. I guess it's okay since we only fetch the mentions array once from the API?

parrish: Also, the order won't change unless the commenting user changes the mentions.

parrish:
I'm good for a merge on this.  It'd probably be good to get another set of eyes on it first though.  @srallen, @rogerhutchings would you have a chance to check this out?

app/talk/data-export-button.cjsx
```diff
@@ -10,11 +9,14 @@ module.exports = React.createClass
     isAvailable: null
     exportRequested: false
     exportError: null
+    mostRecent: null
 
   exportGet: ->
     @_exportsGet or= talkClient.type('data_requests').get(section: @section(), kind: @props.exportType).then (requests) =>
-      @setState(isAvailable: true) if requests.length > 0
-      requests
+      if requests.length > 0
```
srallen: I don't think this conditional is working as intended. When I click the data export button for comments, I see in the network tab that the POST request is going out and there is a response, but the UI isn't changing from showing "no recent requests"

eatyourgreens: @srallen the data exports button should be fixed now.

srallen: Awesome. Looks good. Thanks.

srallen:
@rogerhutchings Did you have time to look this over? Or should we go ahead and merge?

rogerhutchings:
LGTM, merge away

PR #3083, opened 2016-09-28
Fix unique key warning in classifier

srallen:
This fixes the annoying unique key React warning that would show up with the Classifier. I've added a key prop to the `HookComponent` in the `coverable-task-container` div.

# Review Checklist

- [x] Does it work on mobile?

## Optional

- [ ] If changes are made to the classifier, does the dev classifier still work? Need to merge #3081 before this can be checked. Can rebase once merged.

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -223,13 +223,15 @@ Classifier = React.createClass
             disableInterventionFunction={@disableIntervention}
           />}
         <div className="coverable-task-container">
-          {persistentHooksBeforeTask.map (HookComponent) =>
-            <HookComponent {...taskHookProps} />}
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Will the random number change a component's key on every render? If not then this looks good to me.

srallen:
As we discussed in the hangout, since the `HookComponent` is an abstraction and if there are hooks and what they are changes depending on the tool, I won't be able to precisely give it a unique key id. I think in this case, this may be the best we can do to at least get rid of the warning that is being thrown. I could leave a comment in the code for a possible refactor later?

eatyourgreens:
Maybe keys in PFE are something to talk about next week? I don't think random keys will break the classifier here, so happy to merge and kill that warning.

PR #3109, opened 2016-10-05
More tests

simensta:
Refactors SVGImage into ES6 as well as adds a test. Also, adds enzyme.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


simensta:
Good catch, @srallen. `Failed propType` warning should be gone now.

PR #3116, opened 2016-10-10
Persist annotations utility

### LINKED ISSUE #2956: Create an experimental tool/feature that allows task answers to persist if go back to previous task(s)
## Expected behavior
Clicking the Back button will not clear work for current task.

## Current behavior
Clicking the Back button will clear work for current task.

## Steps to replicate
any workflow with more than one task, or [this demo workflow](https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?reload=1&workflow=2426)

srallen:
Fixes #2956. This adds the annotation persistence if the back button is clicked by storing it in a temporary cached classification resource separate from the classification in the classifier and then checking for and loading a cached annotation when the next button is clicked if the workflow is configured to use annotation persistence and `persist annotations` is added to a project's experimental tool's array.

This should be the functionality needed for NfN but also keep it from breaking if branching logic is added later. Annotations in the classifier are popped off as expected when back is clicked, re-added if it exists in the cache. This can be seen with the branching logic of the dev classifier.

Should be tested with the NfN project. I've added `persist annotations` to its experimental tools array already. @mcbouslog you'll need to enable the option for the workflows that need this. I'm not sure which or if all will.

One of their workflows: https://persist-annotations-utility.pfe-preview.zooniverse.org/projects/zooniverse/notes-from-nature/classify?reload=1&workflow=2343&env=production

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -59,7 +59,6 @@ Classifier = React.createClass
     @setState renderIntervention: true
 
   componentDidMount: ->
-    console.log('cacheClassification', @props.cacheClassification)
     experimentsClient.startOrResumeExperiment interventionMonitor, @context.geordi
     @setState renderIntervention: interventionMonitor?.shouldShowIntervention()
     interventionMonitor.on 'interventionRequested', @enableIntervention
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -437,7 +437,6 @@ Classifier = React.createClass
 
     if @props.workflow.configuration.persist_annotations
       cachedAnnotation = @props.cacheClassification.isAnnotationCached(taskKey)
-      console.log('cachedAnnotation', cachedAnnotation)
       if cachedAnnotation?
         annotation = cachedAnnotation
 
@@ -447,6 +446,7 @@ Classifier = React.createClass
   # Back up:
   destroyCurrentAnnotation: ->
     lastAnnotation = @props.classification.annotations[@props.classification.annotations.length - 1]
+
     @props.classification.annotations.pop()
     @props.classification.update 'annotations'
 
@@ -456,7 +456,8 @@ Classifier = React.createClass
   completeClassification: ->
     if @props.workflow.configuration.persist_annotations
       @props.cacheClassification.delete()
-
+    
+    currentAnnotation = @props.classification.annotations[@props.classification.annotations.length - 1]
     currentTask = @props.workflow.tasks[currentAnnotation?.task]
     currentTask?.tools?.map (tool) =>
       if tool.type is 'grid'
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -22,6 +22,9 @@ experimentsClient = require '../lib/experiments-client'
 interventionMonitor = require '../lib/intervention-monitor'
 `import CacheClassification from '../components/cache-classification'`
 
+# For easy debugging
+window.cachedClassification = CacheClassification
+
 Classifier = React.createClass
   displayName: 'Classifier'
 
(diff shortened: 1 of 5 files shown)


app/classifier/index.cjsx
```diff
@@ -431,20 +433,35 @@ Classifier = React.createClass
...
+        annotation = cachedAnnotation
+
     classification.annotations.push annotation
     classification.update 'annotations'
 
   # Back up:
   destroyCurrentAnnotation: ->
+    lastAnnotation = @props.classification.annotations[@props.classification.annotations.length - 1]
+
     @props.classification.annotations.pop()
```
eatyourgreens: `pop` returns the last element of the array, so I think you can shorten this down to `lastAnnotation = @props.classification.annotations.pop()`.

eatyourgreens:
This looks really good to me. 👍

app/classifier/index.cjsx
```diff
@@ -40,6 +41,7 @@ Classifier = React.createClass
     subject: null
     classification: null
     onLoad: Function.prototype
+    cacheClassification: new CacheClassification
```
eatyourgreens: Should this move into the `CachedClassification` component, so that there's only ever one instance of `CachedClassification` in the interface?

app/classifier/index.cjsx
srallen: You're right, but doing this instead seems to be breaking returning to the first task in a branching workflow. Not sure why yet.

mcbouslog:
- [x] I tested on NfN-like workflows, and a NfN-like branching workflow - looked good.
tomorrow morning I'll...
- [ ] test on an official NfN workflow (one linked above will work, 2343)
- [ ] finish code review

app/classifier/index.cjsx
eatyourgreens: `export const cachedClassification = new CacheClassification` should do the trick, I think. No big deal if it doesn't work, though.  https://developer.mozilla.org/en/docs/web/javascript/reference/statements/export

srallen: Thanks for the tip. It works if I set it to a variable first and then export it. I was getting an undefined error with the import into the classifier otherwise.

app/components/cache-classification.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+    }
+    cachedClassification.annotations.push(newAnnotation);
+
+    cachedClassification.update('annotations');
+  }
+
+  delete() {
+    this.state = { cachedClassification: null };
+  }
```
mcbouslog: we think the apiClient is keeping a cache, which causes annotations given for a previous subject to sometimes pre-populate for a different subject. though since it's not saving the resource, `.delete()` isn't working. however, I think adding...
```
this.state.cachedClassification.update('annotations');
```
after the state is updated to null, will fix the issue?

app/components/cache-classification.js
```diff
@@ -0,0 +1,67 @@
+import apiClient from 'panoptes-client/lib/api-client';
+
+class CacheClassification {
+  constructor() {
+    this.state = {
+      cachedClassification: null,
+    };
+
+    window.cachedClassification = this.state.cachedClassification;
```
mcbouslog: per conversation with Sarah, we're thinking about moving this to classifier/index.cjsx

app/components/cache-classification.js
srallen: Yes, it's being cached in `_typeCache`. I'm thinking about pulling the `apiClient` from the class since all we're really interested in is the annotations array. I thought it might be useful to have the classification resource structure, but didn't think about the caching in the client itself and issues that it may cause.

mcbouslog: thanks Jim! there's an `.unchache` function associated with a apiClient resource, so would...
```
this.state.cachedClassification.uncache();
```
within the delete method be preferred?

srallen:
@mcbouslog Great catch. Staging branch is updated for final testing.

PR #3124, opened 2016-10-12
Speed up stats API call

wgranger:
Refers to #2998 

Describe your changes.
Changes made in this PR speed up the API call to fetch project classifications for the stats page. This also reflects changes that will be made with the new signed in home page. Some refactoring was also made to handle redundant code, and the `ProjectIcon` component was made to handle a project resource card.

# Review Checklist

- [X] Does it work on mobile?

https://stats-avatars-move.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/classifications-ribbon.cjsx
+++ b/app/components/classifications-ribbon.cjsx
@@ -1,7 +1,7 @@
 React = require 'react'
 apiClient = require 'panoptes-client/lib/api-client'
-getColorFromString = require('../lib/get-color-from-string').default
-getUserClassificationCounts = require('../lib/get-user-classification-counts').default
+`import getColorFromString from '../lib/get-color-from-string';`
+`import getUserClassificationCounts from '../lib/get-user-classification-counts';`
 
 ClassificationsRibbon = React.createClass
   displayName: 'ClassificationsRibbon'
--- a/app/components/classifications-ribbon.cjsx
+++ b/app/components/classifications-ribbon.cjsx
@@ -79,21 +79,15 @@ module.exports = React.createClass
   getClassificationCounts: (user) ->
     @setState loading: true
 
-    getUserClassificationCounts(@props.user).then (classificationCounts) =>
-      console.log('Got counts', classificationCounts)
-      awaitProjects = Promise.all Object.keys(classificationCounts).map (projectID) =>
-        return apiClient.type('projects').get projectID
-
-      awaitProjects.then (projects) =>
-        console.log('Got projects', projects)
-        pairs = []
-        for i in [0...projects.length]
-          pairs.push
-            project: projects[i].display_name
-            classifications: classificationCounts[projects[i].id]
-        @setState
-          loading: false
-          projects: pairs
+    getUserClassificationCounts(@props.user).then (projects) =>
+      pairs = []
+      for i in [0...projects.length]
+        pairs.push
+          project: projects[i].display_name
+          classifications: projects[i].activity_count
+      @setState
+        loading: false
+        projects: pairs
 
   render: ->
     if @state.loading
--- a/app/components/classifications-ribbon.cjsx
+++ b/app/components/classifications-ribbon.cjsx
@@ -1,5 +1,7 @@
 React = require 'react'
 apiClient = require 'panoptes-client/lib/api-client'
+getColorFromString = require('../lib/get-color-from-string').default
+getUserClassificationCounts = require('../lib/get-user-classification-counts').default
 
 ClassificationsRibbon = React.createClass
   displayName: 'ClassificationsRibbon'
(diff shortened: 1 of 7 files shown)


app/components/classifications-ribbon.cjsx
```diff
@@ -1,5 +1,7 @@
 React = require 'react'
 apiClient = require 'panoptes-client/lib/api-client'
+getColorFromString = require('../lib/get-color-from-string').default
```
eatyourgreens: I think you can use backticked javascript to load the js files eg. `import getColorFromString from '../lib/get-color-from-string';`

app/components/classifications-ribbon.cjsx
```diff
@@ -1,5 +1,7 @@
...
+getUserClassificationCounts = require('../lib/get-user-classification-counts').default
```
eatyourgreens: same here

app/components/classifications-ribbon.cjsx
```diff
@@ -81,33 +74,20 @@ module.exports = React.createClass
 
   componentWillReceiveProps: (nextProps) ->
     unless nextProps.user is @props.user
-      @getClassificationCounts @props.user
+      @getClassificationCounts nextProps.user
```
eatyourgreens: good catch!

app/lib/get-user-classification-counts.js
```diff
@@ -0,0 +1,26 @@
+import getAllProjectPreferences from './get-all-project-preferences';
+import apiClient from 'panoptes-client/lib/api-client';
+
+function getUserClassificationCounts(user) {
+  return getAllProjectPreferences(user).then((projectPreferences) => {
+    const projectIDs = projectPreferences.map((projectPreference) => {
+      return projectPreference.links.project;
+    });
+    return apiClient.type('projects').get({ id: projectIDs, cards: true, page_size: projectPreferences.length }).catch(() => {
```
eatyourgreens: How does this scale for the volunteers that have participated in many projects? Should we limit the number of project IDs passed to the API call? Pinging @camallen here.

app/pages/profile/stats.cjsx
```diff
@@ -3,12 +3,13 @@ apiClient = require 'panoptes-client/lib/api-client'
 ClassificationsRibbon = require '../../components/classifications-ribbon'
 PromiseRenderer = require '../../components/promise-renderer'
 ProjectIcon = require '../../components/project-icon'
+getUserClassificationCounts = require('../../lib/get-user-classification-counts').default
```
eatyourgreens: You can use backticked javascript here too.

app/components/project-icon.cjsx
```diff
@@ -24,20 +24,23 @@ module.exports = React.createClass
...
+      apiClient.type 'avatars'
+        .get project.links?.avatar?.id
+        .then (avatar) =>
+          @setState avatar: avatar.src
+        .catch =>
+          project.get 'avatar'
+            .catch =>
+              null
+            .then (avatar) =>
+              @setState avatar: avatar.src
```
eatyourgreens: this block was a horrible hack on my part to get avatars working again, after the stats page broke. It would be better if we could retrieve avatars without making two API calls per project. I think the avatar should already be present in the resource if we're using the cards route, so maybe we don't need the `else` block?

app/lib/get-user-classification-counts.js
camallen: I assume you'll still need paging on this call too, should be easy to try out.

wgranger: Alright. I took out any projectPreference with an activity count of one or less.

app/components/project-icon.cjsx
wgranger: Took out the else block on this one. Now, this should always revert to the default/generic avatar if one isn't found in `avatar_src`. However, I also had to alter the `RecentProjects` component so it would ask for cards on the API call. That seemed to be the only other place where `ProjectIcon` was being used without cards.

app/pages/home-logged-in.cjsx
```diff
@@ -84,7 +84,7 @@ RecentProjects = React.createClass
             title: "home.recentProjects.title"
             href: "/users/#{@props.user.login}/stats"
             action: "home.recentProjects.button"
-          projectsPromise = apiClient.type('projects').get userProjects, include: ['avatar', 'owners']
+          projectsPromise = apiClient.type('projects').get(id: userProjects, cards: true, include: ['avatar', 'owners'])
         else
```
eatyourgreens: I think the cards endpoint includes the avatars and owners, so you can get rid of the `include` parameter here.

wgranger: Good catch

eatyourgreens:
Looks good, but I just checked it out locally and my stats page is showing "Users can only view their own stats." above the ribbon and avatars.

eatyourgreens:
Oh wait, looks like the live page does that too. I thought that message only displayed if you tried to look at someone else's stats. Anyway, feel free to ignore my previous comment.

PR #3127, opened 2016-10-14
Clarification of How to Become Beta Tester

BNewing:
I added in a slight wording change to the beta project email tick box in the Settings, to try to make it clearer that that's how you can become a beta tester - or rather get the emails that enable you to do the beta testing and therefore become one.

The change is inspired by talking to my uncle who is a big Zooniverse fan and volunteer, and hadn't realised that that's what that tick box does.


camallen:
LGTM

PR #3128, opened 2016-10-14
Changing Credited Name to Real Name on the Settings Page

BNewing:
At the moment, when you sign up, you're asked for your 'Real Name', but on the settings page this is referred to as 'Credited Name'. This suggested change would alter the reference on the Settings page to 'Real Name' so that they match up.

It doesn't change anything deeper than what is displayed in the web page though, and it looks like in the code in other places it might still be being referred to as credited name. At the moment, I've no idea how to change any of that.

This suggestion is a result of the glossary that Meg and I are writing - it was in defining the term in the glossary that we noticed that they're called different things in the two places.

### DIFF
--- a/app/pages/settings/account.cjsx
+++ b/app/pages/settings/account.cjsx
@@ -105,7 +105,7 @@ module.exports = React.createClass
             <br />
 
             <AutoSave resource={@props.user}>
-              <span className="form-label">Credited name</span>
+              <span className="form-label">Real name</span>
               <br />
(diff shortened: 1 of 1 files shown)


simoneduca:
Adding alumni to the mention-bot blacklist didn't seem to work then https://github.com/zooniverse/Panoptes-Front-End/commit/da3457c0773778eb6b00789e78ed873915fd133b

srallen:
@simoneduca Looks like mention bot is ignoring the config for others too: https://github.com/facebook/mention-bot/issues/178

camallen:
LGTM @BNewing

PR #3134, opened 2016-10-18
An inactive workflow shouldn't be a project's default

srallen:
This fixes a bug encountered with the Understanding Animal Faces project described on slack. The default workflow was made inactive and new visitors to the project's classify page would try to load the default workflow which produced an error as well as a loop to try to assign a workflow it couldn't.

Instead of tasking the classifier with clearing out an inactive default workflow, I think it makes more sense to have this action happen when the workflow is made inactive in the project builder. I also added a note for project builders so they know that the default setting for the project will be removed if the default workflow is made inactive. 

One issue is that the asterisk indicating the default won't disappear until the page is refreshed or they navigate to cause a reload of the sidebar in the project builder. I'm not sure how to fix this without refactoring to remove the `ChangeHandler` and `PromiseRenderer` handling the workflow listing in the sidebar, but this is not within scope of fixing this bug.

Staged branch: https://fix-workflow-selection-for-defaults.pfe-preview.zooniverse.org/

# Review Checklist

- [x] Does it work on mobile?

## Optional

- [x] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state? Removed `PromiseRenderer` on the visibility page.

### DIFF
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -10,13 +10,15 @@ module.exports = React.createClass
   getDefaultProps: ->
     project: null
 
-  getInitialState: ->
-    error: null
-    setting:
-      private: false
-      beta_requested: false
-      launch_requested: false
-    workflows: null
+  getInitialState: -> {
+    error: null,
+    setting: {
+      private: false,
+      beta_requested: false,
+      launch_requested: false,
+    },
+    workflows: null 
+  }
 
   mixins: [SetToggle]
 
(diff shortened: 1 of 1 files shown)


srallen:
Assigned @camallen and @simoneduca since you were looking at this bug earlier today. Feel free to reassign.

app/pages/lab/visibility.cjsx
```diff
@@ -17,11 +16,18 @@ module.exports = React.createClass
       private: false
       beta_requested: false
       launch_requested: false
+    workflows: null
```
camallen: spacing

srallen: I refactored `workflows` to not use the `SetToggle` mixin since I added the custom handling to remove the project default config if necessary. I added some curly braces to make it clear what is what here.

PR #3137, opened 2016-10-19
Hide mark details unless frame numbers match

### LINKED ISSUE #3132: Cloning markers also clones Sub-task windows that need to be closed individually
## Expected behavior

In https://github.com/zooniverse/Panoptes-Front-End/issues/2888 cloned markers in multiple images was enabled - thanks! I've just noticed that if you have a drawing task with clone markers set- any sub-tasks also get cloned. I would expect that if the marker is cloned, the sub-task only appears once. 
## Current behavior

At the moment each cloned marker has a sub-task window that needs to be selected for each cloned marker
E.g. Here's a screen shot for Space Warps with cloned markers and a sub-task, each of the 3 sub-task menu's need to be clicked & as you can place more than one mark you can get as many sub-tasks menus as images open. While the category selected in one is replicated in the others - each menu needs closing individually with the 'OK' button.

image

I think one 'OK' press should close all the sub-task menus open when markers are cloned.

In the case of multiple makers being made you get a set of new sub-task menus each time that replace the previous set. I think this may be OK for us in this current version of the project but may impact other projects.

I'm not sure what the best option to declutter/simplify this is but perhaps having only one sub-task menu open for cloned markers would work.
## Steps to replicate

Try the workflow on
https://www.zooniverse.org/projects/aprajita/swdemo/classify?reload=1&workflow=2629

if you select yes on the first question and then proceed to the marking the image you can replicate this.
## Additional information
- **Operating system:**
- **Browser:**

eatyourgreens:
Fixes #3132 .

Describe your changes.
Only shows subtask popups on the original marker.

Note that this might make it difficult to go back and edit subtasks, as you have to remember to click on the original mark that you placed, not one of its copies.
# Review Checklist
- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://fix-3132.pfe-preview.zooniverse.org/
## Optional

### DIFF
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -53,6 +53,11 @@ module.exports = React.createClass
                 continue
 
               toolDescription = taskDescription.tools[mark.tool]
+              
+              if parseInt(mark.frame) is parseInt(@props.frame)
+                {details} = toolDescription
+              else
+                details = null
 
               toolEnv =
                 containerRect: @props.containerRect
@@ -65,7 +70,7 @@ module.exports = React.createClass
               toolProps =
                 classification: @props.classification
                 mark: mark
-                details: toolDescription.details
+                details: details
                 color: toolDescription.color
                 size: toolDescription.size
 


eatyourgreens:
Rebasing this to latest master. @aprajita Is it still needed for Space Warps?

aprajita:
hi @eatyourgreens, it's not needed for spacewarps public project, but i was drawing up an inspection sub-project for our collaborators. It was more just to flag up the potential issue just in case it affects other projects.   thanks

PR #3146, opened 2016-10-21
Fix 2307

### LINKED ISSUE #2307: Ability to re-order Links in Product Builder
Something I just noticed while adding a new link to the project builder... noting it here until I or someone else gets to it.

Right now there is no way to re-order the links to external pages, for example if you want to add a new one at the start.

The workarounds are either to copy and paste the text "downwards" from cell to cell, freeing up space at  the top, or to delete all links and add new ones.

The former has the disadvantage that bad links will temporarily be saved (due to autosave).
The latter has the disadvantage that the URLs are lost and you have to temporarily paste them elsewhere.

We should add a little drag icon to each row so you can re-order.

Not high priority in the grand scheme of things, but important for the usability of that feature.

CKrawczyk:
Fixes #2307 

Adds the ability to drag and drop the items of the `external links` table.  I have also extracted the external links editor into its own file and converted it to es6.

external-link-editor

The only eslint errors are about the use of `.bind()` and `._key`.  Both of these are in keeping with the current style of the code base and I think can be ignored.

This branched is staged at: https://fix-2307.pfe-preview.zooniverse.org/

~~Currently works in Chrome, but in Firefox the mouse can't be used to select the cursor position in the text box (not sure why this is).~~

**Edit**:
This now works in all browsers.  The `drag-reorderable` code does not support touch events, but everything else works on mobile.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/external-links-editor.jsx
+++ b/app/pages/lab/external-links-editor.jsx
@@ -1,0 +1,100 @@
+import React from 'react';
+import AutoSave from '../../components/auto-save.coffee';
+import handleInputChange from '../../lib/handle-input-change.coffee';
+
+class ExternalLinksEditorRow extends React.Component {
+  render() {
+    return (
+      <AutoSave tag="tr" resource={this.props.project}>
+        <td>
+          <input type="text" name={`urls.${this.props.idx}.label`} value={this.props.link.label} onChange={handleInputChange.bind(this.props.project)} />
+        </td>
+        <td>
+          <input type="text" name={`urls.${this.props.idx}.url`} value={this.props.link.url} onChange={handleInputChange.bind(this.props.project)} />
+        </td>
+        <td>
+          <button type="button" onClick={this.props.handleRemoveLink}>
+            <i className="fa fa-remove"></i>
+          </button>
+        </td>
+      </AutoSave>
+    );
+  }
+}
+
+ExternalLinksEditorRow.propTypes = {
+  project: React.PropTypes.object,
+  link: React.PropTypes.object,
+  idx: React.PropTypes.number,
+  handleRemoveLink: React.PropTypes.func,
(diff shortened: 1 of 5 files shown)


CKrawczyk:
After a bit of testing I have identified the issue with the input boxes on Firefox are due to the drag reorder: https://github.com/zooniverse-ui/drag-reorderable

It looks like the mouse clicks are not being passed down into the input box.  This also means that click-and-drag no longer selects text in the input boxes (in any browser) since that triggers the reordering of the rows.  I don't know how we could work around this...

CKrawczyk:
@srallen or @eatyourgreens do either of you have time to review this?

srallen:
@CKrawczyk I might be able to get to this later today.

app/pages/lab/external-links-editor.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+        <AutoSave resource={this.props.project}>
+          <button type="button" onClick={this.handleAddLink}>Add a link</button>
+        </AutoSave>
+      </div>
+    );
+  }
+}
+
+ExternalLinksEditor.DefaultProps = {
```
srallen: should be `defaultProps`

app/pages/lab/external-links-editor.jsx
```diff
@@ -0,0 +1,127 @@
...
+  project: {},
+};
+
+ExternalLinksEditor.propTypes = {
+  project: React.PropTypes.object,
```
srallen: I think this could be set to be `isRequired`

CKrawczyk:
I added some tests for this component and I updated the `npm test` command to search the entire `app` folder for `*.spec.js` files.

PR #3150, opened 2016-10-24
Don't Allow Single Task to Skip to End

### LINKED ISSUE #3120: Single Choice Tasks Default to "Done"
## Expected behavior

If a next task is defined on the workflow, the "next" button should be available instead of "done" on task load.
## Current behavior

Upon first loading a project (with a next task) the green "next" button defaults to "done" when an answer hasn't been selected and a single choice question task is being used.
screen shot 2016-10-11 at 7 53 56 pm
## Steps to replicate

Create  workflow.
Choose a single choice task with a next task defined (not "end of classification").
When that single choice task loads, on the classifier, "done" is the default.
## Additional information
- **Operating system:**
- **Browser:**
  Chrome

wgranger:
Fixes #3120  .

Describe your changes.
Changes made will disable the "next" button when an answer/value isn't recorded for the single  choice task. This, in turn, makes each single choice task "required". However, it seems there would not be an instance when a volunteer should skip a single choice question without an answer.
# Review Checklist
- [ ] Does it work on mobile?
  https://fix-3120.pfe-preview.zooniverse.org/projects/wgranger-test/single-task-bug/classify
## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -457,7 +457,7 @@ Classifier = React.createClass
   completeClassification: ->
     if @props.workflow.configuration.persist_annotations
       CacheClassification.delete()
-    
+
     currentAnnotation = @props.classification.annotations[@props.classification.annotations.length - 1]
     currentTask = @props.workflow.tasks[currentAnnotation?.task]
     currentTask?.tools?.map (tool) =>
--- a/app/classifier/tasks/single.cjsx
+++ b/app/classifier/tasks/single.cjsx
@@ -68,7 +68,7 @@ module.exports = React.createClass
       value: null
 
     isAnnotationComplete: (task, annotation) ->
-      annotation.value? or not task.required
+      annotation.value?
 
   getDefaultProps: ->
     task: null


CKrawczyk:
I can think of a few use cases where you would not want the single task to be required (e.g. for a filtering workflow having "no selection" = "this question is not applicable").  In any case, if this change is made current project owners should be warned first, just to make sure it does not silently break existing projects.

marten:
Non-required questions as the singular question seem like a bad idea, how do you distinguish between accidental clicks on "next" and intentional "n/a". Why not just have an explicit "N/A" option?

marten:
That said, changing this on the fly like this seems like asking for a lot of trouble. Has anyone looked how many single-choice questions aren't required right now? What will we break?

wgranger:
I don't think this is necessarily going to break anything. All it does is makes the next button disabled unless an answer is selected for the question.

marten:
Well, it's making previously non-required questions now be required. Which is a breaking change compared to previous behaviour. Before we do that, we should check which workflows have non-required questions like this, and check if it seems like that was intentional (e.g. in the manner that @CKrawczyk described). I definitely don't think we should push this out without any notice to project builders.

CKrawczyk:
Also, if this will make all single task question required, it should disable that checkbox in the project builder, otherwise we would present users with an option that does not do anything.

wgranger:
I'm closing this for now. May reopen it later after we discuss notifying project builders.

PR #3153, opened 2016-10-25
Add arrow hotkeys to Tutorial

### LINKED ISSUE #2857: Tutorial should allow left / right key navigation
I felt the left right UI arrows on the tutorial should allow me to nav through the tutorial. 

Apparently listeners should be added to this component: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/step-through.cjsx

rogerhutchings:
Fixes #2857.

Rewrite `StepThrough` component in ES6, add left / right arrow shortcuts.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


rogerhutchings:
Staging branch: https://tutorial-hotkeys.pfe-preview.zooniverse.org

app/components/step-through.jsx
```diff
@@ -0,0 +1,140 @@
...
+            title="Previous" 
+            disabled={this.state.step === 0} 
+            onClick={this.goPrevious}
+          >
+            ◀
+          </button>
+          
+          <span className="step-through-pips">
+            {allSteps.map(thisStep =>
+              <label key={thisStep} className="step-through-pip" title="Step #{i + 1}">
```
srallen: Title here is still using coffeescript style string interpolation.

rogerhutchings:
Is that when the pips have focus? `keydown` should only be capturing left / right keystrokes, so that might be default behaviour

rogerhutchings:
I've just tried out on the Gravity Spy tutorial, that _is_ the existing behaviour. Up / down works the same as left / right when the pips have focus.

srallen:
Ah, ok, you're right. I am wondering if we should have up/down scroll rather than do the same thing, but if it's default then perhaps we shouldn't change it unless it's better for accessibility. That can come in a separate PR though.

PR #3167, opened 2016-10-28
Disable pusher in production while we debug issues

marten:
Since deploying this, we've been going over our daily message cap. We're also seeing a lot of connects/disconnects. I wonder if it's counting connects and disconnects as "messages". Disabling this on production will stop going over the cap (if it's PFE that is causing that), and restore functions for other parties like SOCIAM.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


rogerhutchings:
LGTM. Just a reminder that there's still a `pusher` property in the context, but it's left as `undefined` in production.

PR #3174, opened 2016-11-01
Adds status to query params on projects page.

### LINKED ISSUE #3169: Rewrite project status filters as query params
Project status filters - active, paused (out of data) and finished - were added in #3085. However, they were implemented as different routes (e.g. `/projects/finished`), rather than query params like the other filters. After asking around in #3152, the consensus is that they should be query params instead, so they will need to be reimplemented (and tests added!)

simoneduca:
Fixes #3169 .

# Review Checklist

- [x] Does it work on mobile?
- [x] Staging branch https://add-status-to-query-params.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/layout/site-nav.jsx
+++ b/app/layout/site-nav.jsx
@@ -49,13 +49,20 @@ const SiteNav = React.createClass({
 
   getInitialState() {
     return {
-      isMobile: true
+      isMobile: true,
     };
   },
 
   componentDidMount() {
     this.handleResize();
     addEventListener('resize', this.handleResize);
+  },
+
+  componentWillReceiveProps(nextProps, nextContext) {
+    this.logClick = !!nextContext &&
+      !!nextContext.geordi &&
+      !!nextContext.geordi.makeHandler &&
+      nextContext.geordi.makeHandler('top-menu');
   },
 
   componentWillUnmount() {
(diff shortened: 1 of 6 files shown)


rogerhutchings:
- Changing the status filter wipes out any other query parameters; we want to keep them, really. You can do that by using the `onClick` method to trigger `updateQuery()`
- I'd suggest moving that `componentWillMount` method to its own method, just for clarity.
- Remove that `|| null` check

simoneduca:
Meh. Since we're using buttons instead of links now, if one clicks 'Projects' on the main nav, active state is lost on 'Active'.

simoneduca:
@rogerhutchings this should be good to go, except for the little bug in the screenshot above. Any suggestions welcome!

rogerhutchings:
You could check the location as it's passed in as a prop and see if it has the default query params?

simoneduca:
On `StatusLink` component? But the problem is clicking on "projects" in the main nav component `site-nav`

rogerhutchings:
No, in the projects component. It won't help, because the location is just a prop that's being updated - the projects component isn't remounted or anything like that. So if you check the props when they change, and the location doesn't have a query, you can add it in

simoneduca:
Thanks, that should be it.

camallen:
@simoneduca have you tested that this only shows the non-retired / paused projects on the main live section now?

simoneduca:
I did, restaged here https://add-status-to-query-params.pfe-preview.zooniverse.org

simoneduca:
And I don't it's right still. I can still see a paused project under active tab.

simoneduca:
Waiting on this https://github.com/zooniverse/Panoptes/pull/2028

simoneduca:
Backend should be sorted now. @rogerhutchings I think this is good to go.

rogerhutchings:
Looks good. The only outstanding thing I can see from the staged branch is that the top links use a different font to the regular Zoo one

simoneduca:
Well spotted, thanks. This is sorted now.

rogerhutchings:
okay, lgtm

PR #3175, opened 2016-11-01
Fix Back Button Bug with New Shortcut

### LINKED ISSUE #3171: Phantom back button w/ experimental 'shortcut' tool
## Expected behavior
If there's only one task, there shouldn't be a grayed out 'back' button on the classification interface, next to done. 

## Current behavior
Currently there's a grayed out 'back' button next to 'done' in the classification interface for a 'single' task workflow that includes the shortcut option. 

See https://www.zooniverse.org/lab/2416/workflow/1687

If instead I make a single task workflow that just has the marking tool, there's no grayed out 'back' button. 

## Additional information

- **Operating system:**
mac osx 
- **Browser:**
chrome

screen shot 2016-10-31 at 8 18 45 pm

wgranger:
Fixes #3171  .

Describe your changes.
Doesn't display the back button if there is only one task and the experimental shortcut. I also noticed when testing this change that the drawing task summary needed to handle when a value wasn't present in the annotation (if, for example, the shortcut or "Nothing Here," is used. (See screenshot, with "Test" being the example drawing tool)

screen shot 2016-11-01 at 12 58 25 pm

# Review Checklist

- [ ] Does it work on mobile?

https://fix-3171.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/drawing/summary.cjsx
+++ b/app/classifier/tasks/drawing/summary.cjsx
@@ -19,16 +19,14 @@ module.exports = React.createClass
     expanded: @props.expanded
 
   render: ->
-    marks = (mark for mark in @props.annotation.value)
-
     <div>
       <div className="question">
         <Markdown>
           {@props.task.instruction}
         </Markdown>
-        {if @state.expanded and marks.length isnt 0
+        {if @state.expanded and @props.annotation.value.length isnt 0
           <button type="button" className="toggle-more" onClick={@setState.bind this, expanded: false, null}>Less</button>
-        else if marks.length isnt 0
+        else if @props.annotation.value.length isnt 0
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/drawing/summary.cjsx
```diff
@@ -19,14 +19,16 @@ module.exports = React.createClass
     expanded: @props.expanded
 
   render: ->
+    marks = (mark for mark in @props.annotation.value)
```
mcbouslog: sorry I'm prob missing something here, but what does `marks` give us that `@props.annotation.value` doesn't?

wgranger:
Gotcha. Yes, I think I originally had those written out longer to help name the variables something sensible, but I'm totally down for condensing those variables to cut down on lines of code. It makes sense that way.

mcbouslog:
Closes #3171

PR #3180, opened 2016-11-02
Fix props defined on the wrong component

### LINKED ISSUE #3176: Subject Count Badge (zero case)
For empty collections it looks a bit out of place.
image

Alternatives that seem a bit better:
- keep badge, plus additional zero.
- ditch badge on an empty collection.
- (_other ?_)

@wgranger 
(+_thanks for adding the badges_)

eatyourgreens:
Fixes #3176 .

Describe your changes.
Defines some propTypes that were set on `FlexibleLink` rather than `CollectionCard` and sets some default values for them.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://collection-card-props.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/partials/collection-card.cjsx
+++ b/app/partials/collection-card.cjsx
@@ -8,9 +8,7 @@ FlexibleLink = React.createClass
   displayName: 'FlexibleLink'
 
   propTypes:
-    subjectCount: React.PropTypes.number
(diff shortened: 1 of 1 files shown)


wgranger:
Thanks, Jim. LGTM.

PR #3185, opened 2016-11-03
Call for Projects page

BNewing:
Fixes # .

Creation of a Call for Projects page for @hrspiers  

Staged at https://callforprojects.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional


simoneduca:
Menu's a bit wonky.
screen shot 2016-11-03 at 16 38 26

BNewing:
Nooo... @simoneduca  Is there a way around that other than moving it so it's further in the menu at a point where it's unlikely to get divided up like that?

astopy:
Wouldn't this be better off as a blog post?

hrspiers:
@astopy I think a page (that we will remove once the call has finished) is nicer - it looks more professional. Also, it provides somewhere we can put the Call for Projects alongside one another - vvh will be putting up a call soon on the same page.

astopy:
@hrspiers I'm assuming we'll want to blog it as well in that case to publicise it? Removing the page completely after it's published is a problem because we'd be creating broken links from the blog, Twitter, Facebook, etc. that might confuse/frustrate people who click them in the future. If it's going to be a (semi-)permanent thing that gets updated with multiple calls for projects in the future then that's not so bad. We can just keep it up permanently and update the content accordingly when there aren't any current calls.

If we want to put it on zooniverse.org rather than on the blog, then I'm still not sure the About section is the right place. Maybe a new page under News instead?

astopy:
@BNewing If you want to make sure it doesn't wrap onto multiple lines I think you can just use `&nbsp;` rather than a regular space.

srallen:
I think this is a good use case of using `white-space: no wrap` CSS to keep it from wrapping.

simoneduca:
I'm with Adam in finding the about page a bit odd for a call for projects.

BNewing:
Regarding the white space bit - I had a play around in the Chrome page tools but couldn't find where to add in the whitespace CSS without it upsetting the page formatting.  I then tried to add in the `&nbsp;` but that doesn't get translated into HTML so just stays in the title. Am I doing something wrong?

camallen:
@BNewing you want to pair on this?

hrspiers:
@astopy yes that makes sense to keep the page to save breaking links (but to update it instead). Good point!

Let's keep it under the About tab - I think this is a more obvious/appropriate place than News.

BNewing:
@camallen yes please on pairing!

BNewing:
How do people feel about the nav being on two rows? I've restaged it: https://callforprojects.pfe-preview.zooniverse.org/

I felt like it makes more sense to have it all visible rather than disappearing off screen with an arrow indicating there is more content but then that is just me and I'm also not experienced at design stuff at all!

srallen:
@jwbmartin will probably have an opinion on the nav wrapping and what to do about it.

jwbmartin:
This shouldn't be in the About section I think news would be a better place for it.

chrislintott:
@jwbmartin I don't agree - this is goign to be around for a year or so, so it doesn't really fit in news. I'd prefer it in About.

jwbmartin:
The About page should be a relatively static representation of who we are as an organization. If anything this should just be an announcement added to the Projects slides on the Homepage or a banner message on the Projects Page with a link to more information. From a usability standpoint, it doesn't even make sense categorically. If I go to a company's site to look for a job (and this has the same tone as saying "we are hiring") I don't go to the About I look in Contact or maybe Blog

mschwamb:
you might want to be explicit in stating the years associated the dates since it starts with the opening in Dec  2016 and with review in 2017.   @chrislintott  The way it reads it sounds like this is only around until January 31st, 2017. Perhaps there needs to be clarified that this is the first call for proposals?

hrspiers:
Would we be able to resolve this with creating a new tab for the header called "Get Involved" - this could be a home for this "Call for Projects" page. This may also provide a more natural place for the "Volunteering", "Jobs" and "Education" pages. 

It would also be somewhere we could put other similar content in the future e.g. if we wanted to create an "Events" page where we listed any upcoming events where the Zooniverse has a presence. We could also consider having somewhere where people can enter an email address to subscribe to our mailing lists (I don't think we provide this anywhere do we?). Plus once we have more of a museum presence, we may want a page to describe that - which would fit nicely under this tab. 

What do people think? I'd really like to get this page up this week if possible.

BNewing:
I like that idea :) that particular nav bar is looking pretty busy and there is an Acknowledgements page which currently is slotted in to go on the same nav bar. It sounds like it'd make stuff a bit easier to find as well - I had to search a bit to find the jobs bit when I was applying to the internship.

simoneduca:
That's a good plan.

simoneduca:
Also I wonder if we should move Collect from such high visibility place. It always confuses me.

chrislintott:
@jwbmartin What do you think?

astopy:
I like that idea.

jwbmartin:
I agree with Simone about moving collect out of top nav and I think adding in a "Get Involved" area is a good compromise to this issue. If we remove Collect and there is a huge outpouring from users about it then it will be something we will have to rethink.

hrspiers:
Great!

I've just asked @BNewing to take a look at creating the "Get Involved" section tomorrow, to be positioned in top nav after "About"/ before "Talk", and to remove "Collect" from top nav.

I've suggested the "Get Involved" page have the same design/structure as the "About" section, with the following pages in this order - "Volunteering", "Education", "Collect" and "Call for Projects".

BNewing:
Things left to do: 
[ ] Get the Favorites and Collect page working
[ ] Remove Collect from top nav bar
[ ] Make sure all the links in the pages still work
[ ] Check out what the branch conflict is

BNewing:
Does this look okay? I think I've done everything that needs doing. Seeing as you've both already commented on this thread would either @simoneduca or @astopy be up for having a look over it :)?

Staged here: https://callforprojects.pfe-preview.zooniverse.org/about/team

simoneduca:
LGTM, nice one Beverley.

jwbmartin:
The links colors are still that weird blue (is it just because it is staging?) but other than that it is good!

PR #3186, opened 2016-11-03
add option to change completeness calculation

CKrawczyk:
Address the concerns brought up in the talk thread: https://www.zooniverse.org/talk/14/118004

This allows project builders to select between "classification count" or "retirement count" for calculating the completeness for each workflow that is presented on the project stats page.  It defaults to "retirement count" since that is the method returned by the backend.  It also adds help text describing how each method is calculated:

screen shot 2016-11-03 at 4 47 44 pm
 
Note: after this change goes live I think we should notify the project builders for live projects to let them know this is an option available to them.

@camallen feel free to re-assign this if you don't have time to look at it.

# Review Checklist

- [x] Does it work on mobile?
https://completion-stats-toggle.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -70,6 +70,8 @@ EditWorkflowPage = React.createClass
 
     noTasksDefined = Object.keys(@props.workflow.tasks).length is 0
 
+    stats_completeness_type = @props.workflow.configuration.stats_completeness_type ? 'retirement'
+
     disabledStyle =
       opacity: 0.4
       pointerEvents: 'none'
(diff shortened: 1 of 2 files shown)


app/pages/lab/workflow.cjsx
```diff
@@ -70,6 +70,8 @@ EditWorkflowPage = React.createClass
 
     noTasksDefined = Object.keys(@props.workflow.tasks).length is 0
 
+    stats_completeness_type = @props.workflow.configuration.stats_completeness_type ? 'retirement'
```
camallen: do we want to use default props for this instead?

app/pages/lab/workflow.cjsx
```diff
@@ -331,6 +333,57 @@ EditWorkflowPage = React.createClass
           </p>
 
           <hr />
+          
+          <div>
+            <span className="form-label">How should the project completeness be calculated?</span>
+            <br />
+            <p className="form-help">
+              <small>
+                Use this option to change how your project's completeness is caclucated on the public statistics page.
```
camallen: spelling caclucated

app/pages/lab/workflow.cjsx
```diff
@@ -331,6 +333,57 @@ EditWorkflowPage = React.createClass
...
+              When using "Classification Count" the completeness will increase after each classification. The
+              {' '}total number of classifications needed to complete the workflow is estimated assuming a constant retirement limit.
+              {' '}If the retirement limit is changed and/or subjects sets are unlinked from an <b>active</b> workflow this estimate will be inaccurate.
+              {' '}To avoid these issues completed subject sets should not be removed, instead completed workflows should be deactivated
+              {' '}and new ones created for new subject sets (note: workflows can be copied using the <i className="fa fa-copy"/> button at the top of the page).
+              </small>
+            </p>
+            <p className="form-help">
+              <small>
+              When using "Retirement Count" the completeness will increase after each image retires.  Since the images are
```
camallen: true but we only calculate the completeness metric every hour...the UI could check it's accurate and recalculate if it's not.

app/pages/project/stats/stats.jsx
```diff
@@ -273,6 +274,12 @@ export class WorkflowProgress extends React.Component {
         );
       }
       classificationsString += ` / ${total.toLocaleString()}`;
+      // not entirely sure why this first check is needed, but the code crashes without it
+      if (this.props.workflow.configuration) {
+        if (this.props.workflow.configuration.stats_completeness_type === 'classification') {
```
camallen: can you do `this.project.workflow.configuration?.stats_completeness_type`

srallen: The existential operator doesn't exist in ES6. It's been proposed for ES7, but probably not coming any time soon.

camallen: sad times....that check looks fine to me then

app/pages/lab/workflow.cjsx
CKrawczyk: As far as I can tell react does not let you define defaults for nested props.  As long as `workflow` is passed in as a prop anything assigned to `workflow` in `getDefaultProps` is ignored.

app/pages/lab/workflow.cjsx
CKrawczyk: ✔️

app/pages/lab/workflow.cjsx
CKrawczyk: I will just add in a note saying that it is only re-calculated once and hour.

PR #3188, opened 2016-11-04
use the projects cards api end points for signed in home page

camallen:
Use the project cards route for the signed in home page for more efficient api resource usage that should speed up the render times (less network time).

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/home-for-user/my-builds.jsx
+++ b/app/pages/home-for-user/my-builds.jsx
@@ -48,10 +48,26 @@ const MyBuildsSection = React.createClass({
     apiClient.type('projects').get({
       current_user_roles: ['owner'],
       sort: '-updated_at',
+      include: ['avatar']
     })
     .then((projects) => {
       this.setState({
         projects,
+      });
+      projects.map((project) => {
+        if (project.links.avatar.id) {
+          apiClient.type('avatars').get(project.links.avatar.id)
+          .catch(() => {
+            return null;
+          })
+          .then((avatar) => {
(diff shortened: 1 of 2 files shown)


app/pages/home-for-user/my-builds.jsx
```diff
@@ -101,8 +88,7 @@ const MyBuildsSection = React.createClass({
 
         <div className="project-card-list">
           {this.state.projects.map((project) => {
-            const avatarSrc = !!this.state.avatars[project.id] ? this.state.avatars[project.id].src : null;
-            return <ProjectCard key={project.id} project={project} imageSrc={avatarSrc} href={`/lab/${project.id}`} />;
+            return <ProjectCard key={project.id} project={project} href={`/lab/${project.id}`} />;
```
camallen: @srallen passing the project prop in here has a cut down resource response, e.g. no project.tags attributes. so when following the link the lab page falls over as it crashes trying to split the tags

camallen:
I'm having troubles getting this sorted, the `resource.refresh(true)` function works but doesn't seem to emit the changes correctly to the resource (it prints fine but the attribute methods don't return the refreshed state). If someone is available to pair on this next week, i'll have a crack on it then.

camallen:
i ended up just using the cards params for the recent projects end points and left my builds as full resource requests with included avatars. It should speed up the 'my recent projects' ui component and not do much on the 'my builds'

wgranger:
This looks to be more efficient. Good stuff.

Issue #1156, opened 2015-07-21
Request: Two-step confirmation for deleting workflows


frankvitaleiv:
Currently if you hit "delete workflow" there is no pop-up to say "no" and cancel the action.  Could this be implemented, similar to the confirmation window for deleting a subject set?

aliburchard:
YES please!

rogerhutchings:
fixed via bab6d4af66f693acfe5076c3d48299d1da896b48

PR #3206, opened 2016-11-09
Add missing .push method when deleting workflows

rogerhutchings:
Adds missing `push` method when confirming a workflow delete. Previously, the delete was successful but threw a `_this.context is not a function` error without redirecting.

Ref #1156


simoneduca:
LGTM.

PR #3209, opened 2016-11-09
New Notifications Page

wgranger:
Describe your changes.
This is the lovely new notifications page. Notifications are now arranged by project sections. Note: the notifications page can be found by clicking the bell on the top right of the nav bar and are available at the global level as well as the project level. 

Staged at: https://new-notifications-page.pfe-preview.zooniverse.org/

# Review Checklist

- [X] Does it work on mobile?

## Optional


app/pages/notifications/notification-section.jsx
```diff
@@ -0,0 +1,253 @@
...
+
+  componentWillMount() {
+    if (this.props.section === 'zooniverse') {
+      this.setState({
+        name: 'Zooniverse',
+      });
+    } else {
+      apiClient.type('projects').get({ id: this.props.projectID, cards: true })
+      .catch(() => {
+        return null;
```
rogerhutchings: Is this to just swallow errors silently? Not sure that's a great idea 😨

wgranger:
@rogerhutchings Good call on the errors. The most recent commit should make those errors visible.

rogerhutchings:
Cool, there were some other things that sprung to mind on a second look:

- It would be great to have tests for this
- It'd also be great to have this wired up for translations with counterpart
- The notifications page component might as well be rewritten in ES6 at this point
- The close section icon should probably be an up arrow, as an X implies deleting the section
- The section should wait for data to load before opening, as it currently opens a bit before that happens
- The notification section component manages the state for its UI _and_ the data it contains. It might be worth splitting it up, so we have the data and a reusable toggle state component

rogerhutchings:
Btw @wgranger, I don't think your local github setup is right, I can't see your username / avatar against your commits. Running `git config --global user.email "YOUR EMAIL ADDRESS"` should fix it

app/lib/get-notification-data.js
```diff
@@ -0,0 +1,70 @@
...
+      });
+    });
+  });
+}
+
+function moderationData(notification) {
+  return talkClient.type('moderations').get(notification.source_id).then((moderation) => {
+    const comment = moderation.target || moderation.destroyed_target;
+    const promises = [];
+    moderation.reports.map((report) => {
```
rogerhutchings: `.map()` returns an array, so you can define `promises` as the result of that (just return each promise instead of using array.push)

app/lib/get-notification-data.js
```diff
@@ -0,0 +1,70 @@
...
+  return talkClient.type('discussions').get(notification.source_id).then((discussion) => {
+    return talkClient.type('comments').get({ discussion_id: discussion.id, sort: 'created_at', page_size: 1 }).then(([comment]) => {
+      return { notification, data: { discussion, comment } };
+    });
+  });
+}
+
+function getNotificationData(notifications) {
+  const notificationData = notifications.map((notification) => {
+    if (notification.source_type === 'Comment') {
```
rogerhutchings: I'm not a fan of long if/elses - a neater option might be to create an object and look things up against it, e.g.:

```
const lookup = {
  'Comment': commentData,
  ...
};

...

// inside the map()
return lookup[notification.source_type].call(undefined, notification)
```

Long if/else might be easier to grok, see what you think

app/pages/notifications/notification-section.jsx
```diff
@@ -0,0 +1,261 @@
...
+      this.setState({
+        name: 'Zooniverse',
+      });
+    } else {
+      apiClient.type('projects').get({ id: this.props.projectID, cards: true })
+      .catch((error) => {
+        this.setState({ error });
+      })
+      .then(([project]) => {
+        return this.setState({
```
rogerhutchings: doesn't need to be returned

wgranger:
I addressed some of the linting issues. There were a couple that seemed they could be safely disabled (after reading the lint docs for those particular rules). The remaining linting issues don't appear to be breaking, but I was unable to find a simple solution to them.

rogerhutchings:
This needs rebasing, but should be good to go otherwise.

wgranger:
All rebased and should be good to go.

wgranger:
Ok. There's a weird notification error I need to resolve after the rebase.

PR #3222, opened 2016-11-13
Removing 'dot' part from register-name warning-text.

### LINKED ISSUE #3221: Retracted
When Registering a Zooniverse account. The base-name field shows a 'invalid character in text' warning. Which conflicts on dot usage.
image

MvGulik:
Fixes #3221.
>When Registering a Zooniverse account. The base-name field shows a 'invalid character in text' warning. Which conflicts on dot usage.

Describe your changes.
Removing the dot part from the warning text.
Note: Presuming here that dot's are not allowed in the users base-name, and the warning is wrong. (which could be a wrong assumption of mine!)

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/partials/register-form.cjsx
+++ b/app/partials/register-form.cjsx
@@ -184,7 +184,7 @@ module.exports = React.createClass
     name = @refs.name.value
 
     exists = name.length isnt 0
-    badChars = (char for char in name.split('') when char.match(/[\w\-\']/) is null)
+    badChars = (char for char in name.split('') when char.match(/[\w\-\'\.]/) is null)
 
     @setState
       badNameChars: badChars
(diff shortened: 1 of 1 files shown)


camallen:
'.' chars are allowed in usernames at the api level so we should just modify the regex to not report this character.

MvGulik:
>we should just modify the regex.

Should "Just" not be "also" ?

MvGulik:
Bump (*done, I think*)

MvGulik:
Right, it took a wile for the bad coin to dislodge.

Closing this patch.

Holding of on re-pushing the Regex patch alone.
(*Can't really test/debug/etc ..., which feels really weird to me.*)

PR #3226, opened 2016-11-15
Navbar for the Project Builder Help Pages

BNewing:
Fixes # .

Creating a nav bar to link up all of the Project Builder Documentation together

Staged: https://help-navbar.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional


rogerhutchings:
Hmmm. The sub navigation implies that these pages are all children of the same parent, i.e. Help. But clicking on Best Practices takes you to a totally different section, which then has its own children which replaces the previous sub navigation, and I'm not sure that's right.

What does potential reviewer (:trollface:) @mrniaboc think?

BNewing:
I was a bit unsure about that bit! I wasn't sure how else to do it without merging that nav bar in though D: 

Thank you for looking over this Roger!

camallen:
I agree with roger on this. How about a link from the |how to build a project" page to the best practices page? Its really all the top nav link is doing anyway. 

Could you rename the "How to" nav heading to "how to build a project" as well?

camallen:
Maybe remove the "how to page" from the help and just link to the Best Practices content as it seems like this replacing the original how to page.

BNewing:
I like the first two suggestions! I'm not sure I understand the second though. I've just redrafted a new 'how to' page which was going to replace the old how-to page and I was thinking of putting that where hte current how to page is, if that makes sense

camallen:
Yeah, it does. It kind of seems like your new best practices maybe replaces the how to build a project page. but if you've replaced that as well then i think just link to it from the how to and see how it goes.

BNewing:
I was going to put the new how-to-page in a different place to best practices because I interpreted them as  separate things. The how-to page I'm writing is more about how to use the project builder to build a project where as best practices is for strategy for after its built - that's how I've understood it. In which case linking to the best-practices from the how-to page would be be a good idea. I'll do a commit for that and push it up

BNewing:
I've made the changes including adding the link in to that page. I'll also add in a section linking to the Best Practices pages on the new page that will replace that one.

app/pages/lab/help/glossary.cjsx
```diff
@@ -19,13 +19,13 @@ counterpart.registerTranslations 'en',
...
+      + **Collection** - A collection is similar to favoriting an image, but allows volunteers to additionally link together groups of subjects. These collections can be set to either public or private, and you can see all of the public collections on the [Collections](/collections) page. Volunteers can only add to their own collections, and not those of others.
 
-      + **Favorites** - If a volunteer finds an image on a project that they like and wants to be able to see again in the future, they can mark it as a ‘Favorite’ by clicking the heart icon underneath it. These images then show on the volunteer’s [Favorites](collections/favorites-list) page.
+      + **Favorites** - If a volunteer finds an image on a project that they like and wants to be able to see again in the future, they can mark it as a ‘Favorite’ by clicking the heart icon underneath it. These images then show on the volunteer’s [Favorites](/collections) page.
```
camallen: This should be `/favorites` i think

PR #3231, opened 2016-11-16
Fix project search external links

### LINKED ISSUE #3228: Search Bar Bug
## Expected behavior
Should be able to search and go to project

## Current behavior
It goes back to home screen. I have a video of the behavior if you want to check it out

## Steps to replicate
Go to Project Page type something into the search bar and click on the result

- **Operating system:**
iOS 10.10.5

- **Browser:**
Chrome

camallen:
Fixes #3228 - Allow external links to navigate correctly and avoid double lookups via the api.

# Review Checklist

- [ ] Does it work on mobile?
- [X] Did you deploy a staging branch? https://project_search_external_links.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/projects/projects-search-selector.jsx
+++ b/app/pages/projects/projects-search-selector.jsx
@@ -11,14 +11,12 @@ class SearchSelector extends Component {
     this.searchByName = this.searchByName.bind(this);
   }
 
-  navigateToProject(projectId) {
-    apiClient.type('projects').get(projectId)
-      .then(project => {
-        if (project.redirect != null && project.redirect.length !== 0) {
-          return browserHistory.push(project.redirect);
-        }
-        return browserHistory.push(['/projects', project.slug].join('/'));
-      });
+  navigateToProject(projectUrl) {
+    if (projectUrl.match(/^http.*/)) {
+      window.location.assign(projectUrl);
+    } else {
+      browserHistory.push(['/projects', projectUrl].join('/'));
+    }
   }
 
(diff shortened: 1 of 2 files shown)


app/pages/projects/projects-search-selector.jsx
```diff
@@ -11,14 +11,12 @@ class SearchSelector extends Component {
...
-    apiClient.type('projects').get(projectId)
-      .then(project => {
-        if (project.redirect != null && project.redirect.length !== 0) {
-          return browserHistory.push(project.redirect);
-        }
-        return browserHistory.push(['/projects', project.slug].join('/'));
-      });
+  navigateToProject(projectUrl) {
+    if (projectUrl.match(/^http.*/)) {
+      window.location.href = projectUrl;
```
eatyourgreens: I think this should be `window.location = projectUrl;` or `window.location.assign(projectUrl);` as per the first example at https://developer.mozilla.org/en-US/docs/Web/API/Window/location

simoneduca: How come we don't need the api call anymore? Where is projectId passed to navigateToProject?

camallen: Because all you needed was the project slug / redirect that i pass in from the previous api calls as the value on the select (they are all uniq URL's / slugs).

camallen:
@eatyourgreens updated - thanks for the doc links.

eatyourgreens:
:+1:

PR #3233, opened 2016-11-16
update react-select

eatyourgreens:
Updates react-select to 1.0, as part of the upgrade to React 15.

Updates:

- [x] user search
- [x] tag create/search
- [x] project sort/search
- [x] collection search
- [x] dropdown task
- [x] grid tool

Staged at https://react-15-select.pfe-preview.zooniverse.org/

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -83,6 +83,11 @@ module.exports = React.createClass
 
   getInitialState: ->
     optionsKeys: {}
+    selectedOptions: []
+
+  componentWillMount: ->
+    @menus = []
+    @syncAnnotations()
 
   componentDidMount: ->
     annotationValues = @props.annotation.value
@@ -96,11 +101,33 @@ module.exports = React.createClass
       @handleFocus()
 
   componentDidUpdate: (prevProps) ->
+    if prevProps.task isnt @props.task
+      @syncAnnotations()
     if prevProps.task isnt @props.task and @props.autoFocus is true
       @handleFocus()
 
+  syncAnnotations: () ->
+    # this can only be called after props have been updated
+    # as it relies on props.task and props.annotation.
+    # clear selections, then check each annotation.value 
+    # and set the corresponding selected option for each select
+    selectedOptions = []
+    optionsKeys = []
+    @props.annotation.value.map (annotation, i) =>
+      console.log 'ANNOTATION', i, annotation
+      if annotation.option
+        optionsKey = if optionsKeys.length then optionsKeys.join(';') else '*'
+        console.log 'MENU', @getOptions(i, optionsKey)
+        [selected] = @getOptions(i, optionsKey).filter (option) -> option.value is annotation.value
+        console.log 'OPTION', selected
+        optionsKeys.push selected.value if selected?.value
+        selectedOptions[i] = selected
+      else
+        selectedOptions[i] = {label: annotation.value, value: annotation.value}
+    @setState {selectedOptions}
+
   handleFocus: ->
-    @refs['select-0'].getInputNode().focus()
+    @menus[0].focus()
 
   handleOptionsKeys: (i, value) ->
     {id, condition} = @props.task.selects[i]
@@ -114,10 +141,11 @@ module.exports = React.createClass
       optionsKeys[id] = "#{optionsKeys[condition]};#{value}"
     @setState optionsKeys: optionsKeys
 
-  getOptions: (i) ->
+  getOptions: (i, optionsKey) ->
     select = @props.task.selects[i]
-    optionsKey = if not select.condition? then '*' else @state.optionsKeys[select.condition]
-    select.options[optionsKey]
+    optionsKey ?= if not select.condition? then '*' else @state.optionsKeys[select.condition]
+    options = select.options[optionsKey]
+    if options then options else []
 
   getDisabledAttribute: (i) ->
     {selects} = @props.task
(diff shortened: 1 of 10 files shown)


app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -152,15 +157,18 @@ module.exports = React.createClass
...
-            <Select
+            {if selects[i].allowCreate
+              SelectComponent = Select.Creatable
+            else
+              SelectComponent = Select}
+            <SelectComponent
               options={options}
               onChange={@onChangeSelect.bind(@, i)}
-              value={@props.annotation.value[i]?.value}
+              value={@state.value[i]?.value}
```
mcbouslog: so this is driving me crazy - if there's a conditional dropdown, with `allowCreate`, that already has a few options, but a user enters a new option, it adds that new option to the options array in the Select component (child of CreatableSelect), and everything appears to work fine, including specifically the new option is displayed in the input. However, if there aren't already a few options, and a user enters a new option, it gets saved to state and props, but it doesn't show up in the input, and it's not being added to the options prop in the Select component.

In this workflow (`projects/markb-panoptes/focus-testing-project/classify?reload=1&workflow=2737`), the second dropdown in the first task has options, so you should be able to add an option fine. In the second task, the second dropdown does not have any options, but you can add an option and it saves it in props/state/annotation, though it doesn't display in the input.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -100,7 +105,7 @@ module.exports = React.createClass
       @handleFocus()
 
   handleFocus: ->
-    @refs['select-0'].getInputNode().focus()
+    @refs['select-0'].focus()
```
mcbouslog: With back-to-back single dropdowns I get the following error: `TypeError: this.refs.select-0.focus is not a function(…)`. Here's a demo workflow: `projects/markb-panoptes/focus-testing-project/classify?reload=0&workflow=2738`. I've tried just using the Select.Creatable component only, confirmed `ref="select-0"` is indeed in the Select component, not sure what's going on. Maybe it's trying to focus to that component ref before the component has rendered?

eatyourgreens: I'm trying to reproduce this but can't. The dropdown control is being focussed for both questions.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -100,7 +105,7 @@ module.exports = React.createClass
       @handleFocus()
```
eatyourgreens: Maybe set the dropdown value on `componentWillReceiveProps` as well as on mount? There's something weird going on where the dropdown gets its value from the annotation property, so property changes need to override component state. I think I've got my head round it now.

app/components/tag-search.cjsx
```diff
@@ -1,13 +1,33 @@
...
+    tags: []
+  
+  componentWillMount: ->
+    @setState tags: @props.value
+  
+  componentWillReceiveProps: (newProps) ->
+    @setState tags: newProps.value unless newProps.value is @state.tags
+
+  onChange: (options) ->
+    console.log 'CHANGE', options
```
eatyourgreens: Yep. Should be gone now.

app/classifier/tasks/dropdown/index.cjsx
eatyourgreens: `react-select` used to expand simple string and number values out to Option objects `{label, value}` but 1.0 removed that. I suspect this bug is due to values still being passed around as simple strings in some places.

eatyourgreens:
@mcbouslog 7143382 should have fixed a few things that you mentioned. I've also renamed a couple of things to try and make the code a bit clearer.

srallen:
I'm going to defer to @mcbouslog for the final review of this PR since changes were needed for the dropdown task. Everything else LGTM.

eatyourgreens:
I've made another pass through the dropdown task, updated the commit message to be more informative and rebased.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -95,15 +102,33 @@ module.exports = React.createClass
...
+    {selects} = @state
+    annotations.map (annotation, i) =>
+      if annotation.option
+        [selected] = @getOptions(i).filter (option) -> option.value is annotation.value
+        selects[i].selected = selected
+      else
+        selects[i].selected = {label: annotation.value, value: annotation.value}
+    @setState {selects}
+
   handleFocus: ->
```
mcbouslog: once adding a custom answer to a conditional dropdown that doesn't have any options provided, the focus gets stuck.

i.e. in preview Focus Testing Project, Dropdown Workflow, second task, second dropdown (Smaller Numbers, after selecting anything except 100 for Bigger Number) - type any custom answer.
`projects/markb-panoptes/focus-testing-project/classify?reload=1&workflow=2737`

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -133,56 +158,69 @@ module.exports = React.createClass
...
             />
           </div>
         }
       </div>
     </GenericTask>
 
-  onChangeSelect: (i, newValue) ->
+  onChangeSelect: (i, option) ->
+    {selects} = @state
+    newValue = option.value if option?
```
eatyourgreens: I think that's due to `@onChangeSelect(key, null)` in `clearRelated`. Should be `@onChangeSelect(key, {label: null, value: null})` now that options are required to have labels and values.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -133,56 +158,69 @@ module.exports = React.createClass
...
     value = @props.annotation.value
 
-    optionsValues = @getOptions(i)?.map (option) -> option.value
-    newIndex = optionsValues?.indexOf(newValue)
+    options = @getOptions(i)
```
eatyourgreens: Changing the code back gives the same result for me: `newIndex` is set to 0 because the new option has been added to the beginning of the options array before the change handler is called.

app/classifier/tasks/dropdown/index.cjsx
eatyourgreens: e7e0146 should take care of this.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -115,12 +140,12 @@ module.exports = React.createClass
     @setState optionsKeys: optionsKeys
 
   getOptions: (i) ->
-    select = @props.task.selects[i]
+    select = @state.selects[i]
```
eatyourgreens: I'm finding that this change gives me `option: false` for all options, created or not.

eatyourgreens:
@mcbouslog I've made another pass through this with changes that we've discussed on Slack, and rebased.

Issue #2944, opened 2016-09-01
Create classification data export filtered by workflow (front-end piece)


mcbouslog:
see [Panoptes Issue #1729](https://github.com/zooniverse/Panoptes/issues/1729) for back-end piece
## Expected behavior

In Project Builder, Data Exports, can optionally filter classifications export .CSV by workflow. Possibly with dropdown, that sends workflow ID as param to back-end classification CSV creation end point. 
## Current behavior

N/A
## Additional information

-Notes from Nature request

PR #3235, opened 2016-11-16
Workflow export ui

amyrebecca:
Adds a UI (a button on the data dumps page and a modal dialog) to export classifications for an individual workflow only, instead of an entire project at once.

# Review Checklist

- [x] Does it work on mobile?

## Optional


camallen:
LGTM - tested and the UI works mostly. When i request a new workflow classification dump, i don't get an updated most recent link details for the existing / requested media (dump) resource.

ui-w-export

app/pages/lab/data-dumps.cjsx
```diff
@@ -2,6 +2,8 @@ React = require 'react'
 counterpart = require 'counterpart'
 DataExportButton = require '../../partials/data-export-button'
 TalkDataExportButton = require '../../talk/data-export-button'
+ExportWorkflowsDialog = require('../../partials/export-workflows-dialog').default
```
srallen: Prefer ````import ExportWorkflowDialog from '../.../partials/export-workflows-dialog```` syntax with backticks, but not a blocking change.

app/partials/export-workflows-dialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+    apiClient.type("workflows").
+      get(project.links.workflows).
+      then((workflows) => {
+        this.setState({workflows: workflows})
+      })
+  }
+
+  componentDidMount(){
+    this.updateWorkflowsFromProject(this.props.project)
```
srallen: Could we add `defaultProps` and props validations for `this.props.project`?

app/partials/export-workflows-dialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+    this.updateWorkflowsFromProject(nextProps.project)
+  }
+
+  toggleExport(){
+    if(this.refs.workflowList.selectedIndex >= 0)
+      this.setState({workflow_selected: true})
+  }
+
+  requestDataExport(){
+    var workflow_id = this.refs.workflowList.value
```
srallen: Should be `const`

app/partials/export-workflows-dialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+    apiClient.post(`/workflows/${workflow_id}/classifications_export`, {media: {content_type: 'text/csv'}})
+  }
+
+  render(){
+    return (
+        <div>
+          <label className="form-label">Select a Workflow:</label>
+          <select size="5" ref="workflowList" className="multiline-select standard-input" style={{padding: "0.3vh 0.3vw"}} onChange={this.toggleExport.bind(this)}>
```
srallen: Binding to `this` should be done in the constructor. 

Also fyi, string literals in refs are being deprecated, so we should avoid this and find an alternative if possible.

app/partials/export-workflows-dialog.jsx
```diff
@@ -0,0 +1,62 @@
...
+            {this.state.workflows.map((result) => (
+              <option key={result.id} value={result.id}>{result.display_name}</option>
+            ))}
+          </select>
+          <div style={{textAlign: "right"}}>
+            <button className="minor-button" style={{marginLeft: "1em"}}>Cancel</button>
+            <button className="standard-button" style={{marginLeft: "1em"}} disabled={!this.state.workflow_selected} onClick={this.requestDataExport.bind(this)}>Export</button>
```
srallen: Same comment as above about binding.

srallen:
Just noting that this closes #2944 when it merges.

amyrebecca:
I know I said I was out sick today but I wanted to get this in before the weekend. Anyways I'm not sure why my linter wasn't picking up the .eslintrc but since @srallen mentioned the linter errors I was able to get it fixed.

amyrebecca:
in re: @camallen's note, this is actually correct behavior--the link is for a full dump, whereas this generates only a partial dump

amyrebecca:
This should now be ready for merge. I should really know better than to test things with the Admin box checked, by the way...

camallen:
Not a blocker (can come in another PR) but the backend should create a dump media resource for the parital (per workflow) classification export and we can use this media resource for the links / timestamps for most recent as per [here](@zwolf )

@zwolf can you liaise with @amyrebecca to get these links working in the admin interface? This way people don't have to rely email links to get the data.

zwolf:
Related: https://github.com/zooniverse/Panoptes/pull/2063

amyrebecca:
Just discussed this with @zwolf and @trouille and the features that @camallen mentioned will be enabled by some pull requests against panoptes that aren't yet deployed. Laura was fine with merging this as-is and then circling back around to get the links on the next pass.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -27,10 +27,21 @@ class ExportWorkflowsDialog extends React.Component {
...
+    const order = (project.configuration && project.configuration.workflow_order) || [];
+
+    // TODO: this API call duplicates information fetched to draw the lab sidebar.
+    // when we do the lab refactor, this should be cached somewhere
+    project.get('workflows', { fields: 'display_name' })
       .then((workflows) => {
-        this.setState({ workflows });
+        let result = workflows;
+
+        if (order) {
```
srallen: `project.configuration.workflow_order` is an array and an empty array checked this way will be truthy. When `order` is an empty array, the `result` variable will set as an empty array and the dialog will not have workflows to select from.

amyrebecca: This is actually quite tricky to accomplish since the export is started by clicking a button in a modal dialog. I'd like to merge this PR without that behavior unless you feel strongly about it.

srallen: Yeah, there will need to be some refactoring to make it work. I'm ok with this being in a later iteration and PR.

amyrebecca:
I'm not sure where the comment went, but in re: the workflow export message, I'm not actually sure how to accomplish this because the export is performed from inside a modal dialog instead of by interacting with the main page. Maybe passing in an event handler to be fired?

app/pages/lab/export-workflows-dialog.jsx
amyrebecca: Yeah absolutely--I have committed a fix for that.

srallen:
Looks like something maybe went wrong with a rebase. There are duplicate commits now from PRs that have already been merged into master, like the seven-ten stuff, help content, etc.

amyrebecca:
UGH yes, I see them. I'm not sure how they got in there, I'll have to try to remove them with a rebase if I can. I'm not sure why it shows as "okay to merge" under these circumstances

amyrebecca:
This is really confusing, the extraneous commits don't show up in my local branch. I may delete this pull request and recreate it?
screen shot 2016-11-29 at 1 41 06 pm

amyrebecca:
!!!! now all those other commits are gone? but i did see them

srallen:
Hmm, maybe the issue was with github. Looks fine now. Weird. I'll do one last review locally just to be sure what I pull down looks good.

srallen:
I pulled down the latest and requested workflow classification exports for a couple of different test projects I have on staging yesterday afternoon. I still haven't received any emails for it, though if I request the normal classification export, I get that almost instantly since there are so few classifications. 

I know I said I wouldn't hold up for the export request state UI, but right now it just seems broken, so some kind of messaging seems necessary. I checked the network tab and it looks like the `POST` is being made successfully, but without a message or an email, I'm left without any information that it's working.

zwolf:
Emails on workflow exports are coming in a new Panoptes PR soon.

srallen:
@zwolf Thanks for the update. This PR should wait for that.

amyrebecca:
I definitely got emails containing CSV exports in the past, so I'm surprised this is broken now, especially given that the post means I didn't break it on our end, and the endpoint was working fine previously. I agree we should definitely wait until the PR is in and it's fine if we have to have the messaging, it'll just take one more commit.

zwolf:
For workflow classification exports? Staging or production? When?

amyrebecca:
i was getting them on staging pretty consistently during the time i was finishing up this PR, which was admittedly a couple of weeks ago now. the endpoint worked perfectly for me from the first time i called it. i suppose it could have been an artifact of which workflows i was exporting?

zwolf:
There appear to be two different paths by which classification export emails might be sent. 

The first is: `CreateClassificationsExport` (operation) --> `ClassificationsDumpWorker` --> `DumpMailerWorker`. This seems like the one that should be working for both.

Then there's this:`ClassificationDataMailerWorker` --> `ClassificationDataMailer`. Hm. I can't find anywhere that the former is actually called, so it must be some bit of magic. 

The second path is def what's sending the project version of this email and I'm not sure why the first is ignoring the `DumpMailerWorker` callback, cause it's clearly getting as far as the classifications dump worker. Nor why it seems to be intermittent, though it hasn't worked for me lately. Thoughts, @camallen / @marten ?

camallen:
@zwolf both paths use the same data mailer worker, the "magic" is [here](https://github.com/zooniverse/Panoptes/blob/7c0106054b20b4e13dc8e5f345784d8679b60e58/app/workers/concerns/dump_mailer_worker.rb#L14). Finally, seems the dump workers are stuck in retry loops on staging, https://panoptes-staging.zooniverse.org/sidekiq/retries

amyrebecca:
Added--as @srallen suggested--messaging to the user to indicate that their export data has been requested and that they'd receive it in their email (like the other export buttons do). Also added error reporting etc.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  toggleExport() {
+    if (this.workflowList.selectedIndex >= 0) {
+      this.setState({ workflowSelected: true });
+    }
+  }
+
+  requestDataExport() {
+    try {
```
srallen: It doesn't make sense to me to use a try/catch block instead of calling a `then()` and `catch()` on the api client promise. Is there a reason to use try/catch here. I think we would want to wait for the `POST` to complete and then catch the error or then perform a function.

This is what I'm describing to do instead: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/data-export-button.cjsx#L23-L28

srallen:
I enabled the experimental tool flag for a test project of mine in staging and am seeing lots of errors. Not sure where it's coming from since the stack trace isn't super helpful:
screen shot 2016-12-06 at 3 09 11 pm

Requesting for a workflow classification export also shows that error: 
screen shot 2016-12-06 at 3 11 11 pm

The UI for error messaging is looking good at least :)

amyrebecca:
I was getting that error intermittently too--I wonder if it's from the backend? I didn't change anything that would have affected the other promises that page fires (to get the links for previous exports)

zwolf:
@amyrebecca nope, it's not the backend: 
screen shot 2016-12-06 at 3 33 34 pm

amyrebecca:
You're right, I just ran the export on my machine and it worked fine.  This is strange.

amyrebecca:
It doesn't seem like any of these methods should throw 500s though?

screen shot 2016-12-06 at 3 37 40 pm

zwolf:
If it's actually calling 
`          https://panoptes-                    staging.zooniverse.org/api/projects/1618/workflow_contents_export?admin=true:1staging.zooniverse.org/api/projects/1618`/workflow_contents_export?admin=true` 

then that'd def 404.

amyrebecca:
Oh uh, yeah, I would expect that not to work for sure, but let me look and see if that's the actual request being made or if the debug trace is just printing messy.

amyrebecca:
It's just bad pretty-printing, that request was actually:
screen shot 2016-12-06 at 3 47 02 pm

amyrebecca:
And 404s are fine, they actually return some JSON data and the client handles it, the real problem is when we get 500s instead
screen shot 2016-12-06 at 3 47 44 pm

zwolf:
Yeah, it looks like something weird is going on. I'm looking into it.

PR #3238, opened 2016-11-17
Decode blog titles

### LINKED ISSUE #3237: Signed-in page: Decode html in blog posts title
My attempt to fix the problem doesn't seem to work locally. This is what I've done for reference https://github.com/zooniverse/Social/pull/3 and then updated talk-api https://github.com/zooniverse/Talk-Api/pull/96 to grab Social from git. This is because Parrish is the only one who has access to the Social gem. While I look for an alternative fix, I suggest we clean the html entities on the front-end by using, e.g. https://www.npmjs.com/package/he. I've tested it and it works. Doesn't clear htlm tags though.

simoneduca:
Fixes #3237 

Adds package to decode html entities in blog titles. Doesn't decode hmtl tags.
# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://decode-blog-titles.pfe-preview.zooniverse.org/


simoneduca:
Cool, thanks! I was wondering if this is actually been fixed on Social. In which case this is useless. I haven't seen garbled text anymore, but who knows. I'll check with @parrish before merging

parrish:
Just chiming here too. `he` is a much more comprehensive fix.  👍

simoneduca:
Good point Sarah, thank you. Cheers Michael.

PR #3242, opened 2016-11-17
Refactor home buttons

### LINKED ISSUE #3148: Gravity Spy: level visibility on landing page
## Expected behavior

Ability to see the other levels on landing page, but not be able to access them (will need to change wording)
## Current behavior

## Additional information
- **Operating system:**
- **Browser:**

srallen:
This is a refactor of the project home page and the 'Get started!' or workflow buttons depending on project state and/or experimental features. This is in anticipation of setting up a experiment split for workflow assignment for Gravity Spy. Additionally fixes #3148.

Staging link: https://refactor-home-buttons.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy?env=production

@amyrebecca and @wgranger could you review and prioritize for merge on Monday?

@eatyourgreens I added an `ariaDisabled` attribute to the `<Link />` components that are disabled, but VoiceOver wasn't communicating that they were disabled. Am I using it wrong or should I use something else?

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/pages/project/home.jsx
```diff
@@ -0,0 +1,106 @@
...
+  onChangePreferences: () => {},
+  preferences: {},
+  project: {},
+};
+
+ProjectHomePage.propTypes = {
+  activeWorkflows: React.PropTypes.arrayOf(React.PropTypes.object).isRequired,
+  onChangePreferences: React.PropTypes.func.isRequired,
+  preferences: React.PropTypes.object,
+  project: React.PropTypes.shape({
```
wgranger: Nice. I assume this deals with vague propTypes.

eatyourgreens:
@srallen `disabled` usually applies to form controls, or things that are emulating form controls. If you want to disable a link, why not render it as a `<span>` with the same CSS classes instead?

app/pages/project/home-workflow-buttons.jsx
```diff
@@ -0,0 +1,98 @@
...
+    return (
+      <Link to={`/projects/${this.props.project.slug}/classify`} className="call-to-action standard-button">
+        Get started!
+      </Link>
+    );
+  }
+}
+
+ProjectHomeWorkflowButtons.contextTypes = {
+  user: React.PropTypes.object,
```
wgranger: Why is `user` now kept in context rather than props?

wgranger:
I made just a couple line comments. Everything looks pretty clean and clear. Also, are we getting rid of Geordi? I noticed it was removed from the project home.

app/pages/project/home.jsx
srallen: `shape` allows us to define nested prop types of an object.

app/pages/project/home-workflow-buttons.jsx
srallen: `user` was added to the context with the layout refactor Brian did. It's one of the few things that makes sense to go into context as it's data that the whole app likely should be aware of. Context is essentially like global variables. Early on, context in the React API was experimental and unstable, so it wasn't really used until recently.

However, rewriting all of the components that use user to use context instead of props hasn't happened yet. I'd like to see that refactor unless someone can make a good case why user should stay in props and passed down explicitly via parent to child.

srallen:
Re: geordi. I didn't see any explicit tracking, so I don't think it needs to be included here, however, that does remind me geordi will be used with the experiment, so @amyrebecca can answer if it really needs to be included via context.

srallen:
@eatyourgreens good idea with rendering a span instead. I'm gonna make that change.

app/pages/project/home.jsx
amyrebecca: yeah shape is really really cool.

amyrebecca:
I would be surprised if we were getting rid of Geordi since I'm integrating it to consume split information from seven-ten so I'm a little surprised it was removed from the home page. Is it possible it was just left out during the rewrite?

app/pages/project/home-workflow-buttons.jsx
amyrebecca: the only advantage to leaving it in props is that you don't have to remember to explicitly look at the context when re-rendering a component (componentWillUpdate receives nextProps, nextState, and nextContext separately) but as long as people are aware of that it's not a compelling reason

srallen:
Yeah I'll add it back. I forgot about the geordi integration with the experiment when I was rewriting and as is without the experiment, it wasn't doing anything.

app/pages/project/home.jsx
```diff
@@ -0,0 +1,106 @@
+import React from 'react';
+import FinishedBanner from './finished-banner';
+import ProjectMetadata from './metadata';
+import ProjectHomeWorkflowButtons from './home-workflow-buttons';
+import markdownz from 'markdownz';
+const Markdown = markdownz.Markdown;
```
srallen: Yeah, I had it that way before, but it was throwing some kind of import error or compiling error; I don't recall exactly which and I didn't spend any time on debugging it since getting the refactor done was more important. This seemed to be a working solution for it.

eatyourgreens: I've run into this before: markdownz doesn't export Markdown and MarkdownEditor, so you ed up having to do this https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/wrapped-markdown.jsx#L4

I think @rogerhutchings had an idea why this was happenng, but I can't remember what it was.

rogerhutchings: In https://github.com/zooniverse/markdownz/blob/master/src/index.js:

```export default {Markdown, MarkdownEditor};```

!==

```export {Markdown, MarkdownEditor};```

amyrebecca: thanks for explaining, i learned something today :)

rogerhutchings: https://github.com/zooniverse/markdownz/pull/48

PR #3243, opened 2016-11-18
I've added two more publications to Snapshot Serengeti

hrspiers:
Hi,

Two more publications have been added to the list.

Thanks
Helen


rogerhutchings:
legit, but in future it'd be easier if you could make a pull request against the main repo, rather than a fork (i keep having to look up how to locally check out a forked branch)

PR #3244, opened 2016-11-18
Rewrite project About pages to minimize API calls

rogerhutchings:
As pointed out by @eatyourgreens, the front end makes way too many API calls. On the project content pages (about, research etc), a call is made firstly when the about page loads, then each and every time a tab is changed. For the team page, a further call is made to get the project roles, and then again for each individual user listed.

I've rewritten these components in ES6, ditched the `PromiseRenderers`, used the [container / component model](https://medium.com/@dan_abramov/smart-and-dumb-components-7ca2f9a7c7d0#.zcj4c6qpv) to restructure them so requests only happen in the route index, and batched calls together where I could.

I also added in counterpart to translate missing content messages, not that they should really come up.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://about.pfe-preview.zooniverse.org/

## Optional


rogerhutchings:
Things to do:

- When switching tabs, two more API calls get made every time, one to `/conversations` and the other to `/notifications`. They aren't coming from these components, so they should probably be tracked down and ~~killed~~ reviewed.
- I'm still using the `Avatar` component in the team page, which makes another call for each user. I wonder if we could batch together the avatar requests (not sure how they're linked to the user resource atm, other than using the `user.get('avatar')` resource method), and then pass them a URL string instead of a user object?

app/pages/project/about/index.jsx
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  render() {
+    return (this.state.loaded)
+      ? this.renderAbout()
+      : null;
+  }
+
+  renderAbout() {
+    const { state: { pages, team }, props: { children, project } } = this;
```
rogerhutchings: destructuring like a baws

eatyourgreens:
Looks good to me, but needs a rebase to fix merge conflicts.

rogerhutchings:
Thanks @eatyourgreens, rebase is done so I'll merge

PR #3247, opened 2016-11-18
update text on policies page so it matches what's on the acknowledgements page

mschwamb:
update the line of acknowledgements text in the policies page so it matches what's on the acknowledgements page. We decided that it would best to make sure the actual text is still in policies but use the acknowledgements page as a place for people searching for it can easily find this info

### DIFF
--- a/app/pages/lab/help/lab-policies.cjsx
+++ b/app/pages/lab/help/lab-policies.cjsx
@@ -48,7 +48,7 @@ counterpart.registerTranslations 'en',
       - Communicate results to their communities, via open access publication, a blog or elsewhere.
       - Acknowledge Zooniverse in any publications. Please use the following text:
 
-      _This publication uses data generated via the Zooniverse.org platform, development of which was supported by a Global Impact Award from Google, and by the Alfred P. Sloan Foundation._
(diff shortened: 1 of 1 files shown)


mschwamb:
@camallen when you have time, might this be something you can check?

app/pages/lab/help/lab-policies.cjsx
```diff
@@ -48,7 +48,7 @@ counterpart.registerTranslations 'en',
       - Communicate results to their communities, via open access publication, a blog or elsewhere.
       - Acknowledge Zooniverse in any publications. Please use the following text:
 
-      _This publication uses data generated via the Zooniverse.org platform, development of which was supported by a Global Impact Award from Google, and by the Alfred P. Sloan Foundation._
+      _This publication uses data generated via the [Zooniverse.org](https://www.zooniverse.org/) platform, development of is which funded by generous support, including a Global Impact Award from Google, and by a grant from the Alfred P. Sloan Foundation._
```
eatyourgreens: 'development of is which funded' should be 'development of which is funded'

mschwamb:
missed that... thanks @eatyourgreens - this should be updated now

mschwamb:
@eatyourgreens I fixed the typo I missed the first time. Is this good to go now?

PR #3093, opened 2016-10-04
Add Seven-Test split testing

parrish:
Adds A/B split testing for
- Landing page text
- Workflow advancement text
- Mini-course visibility

Check it out at https://splits.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing

# Review Checklist
- [x] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [x] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch?
## Optional
- [x] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?

PR #3254, opened 2016-11-21
[WIP] Gravity Spy split setup

srallen:
This adds the UI piece for the first a/b split test for Gravity Spy. The experiment is a 50/50 split between _new_ users to the project. One half will see the standard buttons and will start with the beginner workflow. One half will see the 'Get Started!' button and on click will have their project preferences updated to load the level 4 workflow aka apprentice workflow.

This PR depends on #3242, the experiment setup in Seven-Ten production, and Geordi integration, so it's labeled WIP. Once #3242 is merged, I'll rebase and clean up the commit history. 

https://gs-split-setup.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -14,6 +14,7 @@ experimentsClient = require '../../lib/experiments-client'
 {Split} = require('seven-ten')
 {VisibilitySplit} = require('seven-ten')
 
+
 FAILED_CLASSIFICATION_QUEUE_NAME = 'failed-classifications'
 
 PROMPT_MINI_COURSE_EVERY = 5
--- a/app/pages/project/home-workflow-buttons.jsx
+++ b/app/pages/project/home-workflow-buttons.jsx
@@ -7,6 +7,7 @@ export default class ProjectHomeWorkflowButtons extends React.Component {
   constructor(props) {
     super(props);
 
+    this.handleSplitWorkflowAssignment = this.handleSplitWorkflowAssignment.bind(this);
     this.shouldWorkflowBeDisabled = this.shouldWorkflowBeDisabled.bind(this);
     this.renderRedirectLink = this.renderRedirectLink.bind(this);
     this.renderWorkflowButtons = this.renderWorkflowButtons.bind(this);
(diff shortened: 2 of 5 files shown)


parrish:
#3093 Needs to get merged before this - it's doing the split config/setup that this needs

parrish:
This should be implemented as a new split type in the seven-ten api.

Rather than hardcoding the ids, it's better to set it on the variant.

It's also worth moving the component setup out of PFE and into the client.  e.g. [text-split](https://github.com/zooniverse/Seven-Ten-Client/blob/master/src/text-split.jsx)

parrish:
I added the split type here: https://github.com/zooniverse/Seven-Ten/commit/3933b09908a2d930625087151ddcc9ada3abad7e

srallen:
I merged #3093 and will resolve the new conflicts that came up. 

Not sure I'll have time to extract out into the client this new kind of split before I leave for the holiday. Hardcoded isn't ideal, but it is working and I think the experiment will only be running for a short time. @trouille would know. If extracting out is blocking the PR, then we'll have to see who will have time to get that done before next Tuesday. I agree long term that we should add a generic component based on this kind of split for future use.

parrish:
Nah, it's not a blocker.  If it's something re-usable, then it'd be good to get it pulled out.  I was having some trouble trying to unpick those test failures.

trouille:
To answer the above -- the researchers would like for the experiment to run until 250 new users see the experiment and 250 new users the control/default, so ~500 total.

srallen:
#3242 is now merged and I've rebased, but it looks like I have a failing test now. I'll take a look at that.

srallen:
Ok, the test is fixed. @amyrebecca and @parrish this is ready for review and is just waiting on the Geordi config and the experiment in production.

I'm available for the next ~2 hours or so if anything else needs work, but after that I'll be out and won't have regularly internet access until I get back to Chicago.

amyrebecca:
I'm looking through the code now, not seeing a lot to complain about. Any thoughts on whether this would be better to merge now vs. waiting for you to get back in case there are problems?

parrish:
This looks alright for a merge. It all shouldn't really impact other projects too.

PR #3287, opened 2016-11-30
Fix Message Overflow for Small Screens

wgranger:
**Describe your changes.**
On the landing page, longer messages are overflowing out of their parent DIV for smaller screens. This was first noticed on a Android with a larger screen (hence the branch name). CSS and screen size cutoff was changed.

https://android-view.pfe-preview.zooniverse.org/

screen shot 2016-11-30 at 3 30 16 pm

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/home-for-user/recent-messages.jsx
+++ b/app/pages/home-for-user/recent-messages.jsx
@@ -174,19 +174,17 @@ const RecentCollectionsSection = React.createClass({
         </span>
         <img role="presentation" src={avatarSrc} className="recent-conversation-link__partner-avatar" />
         <span className="recent-conversation-link__partner">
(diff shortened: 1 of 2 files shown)


rogerhutchings:
I can't read the message line on smaller screens, as my user name pushes the whole thing too far off to the right. What about dropping it down to a new line and then hiding the overflow?

wgranger:
Yes, the message line is hidden if the screen width is 400 or less. Are you suggesting the message line is included for smaller screens, but placed underneath the sender? Or are you checking this branch with a slightly wider screen (> 400) and seeing the tiny message block that could be pushed underneath?

wgranger:
Alright. The message is moved to a new line for smaller screens. Ellipsis is also added to the overflow for longer content.

PR #3289, opened 2016-11-30
Add user's total classifications to logged-in home

### LINKED ISSUE #1048: Add total classification count to stats page
The ribbon on the stats page is great, but I would also be nice to have a "total" classification count (across all projects) someplace on the page.

simoneduca:
Fixes #1048 

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-1048.pfe-preview.zooniverse.org/
- [x] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)? No.


simoneduca:
I think this is ready to go.

simoneduca:
One related thing that might go into another PR: there seem to be a lot of api calls on this page, but I can't figure out more efficient way to get the same data. Thoughts?
These are the calls:
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/get-all-project-preferences.js#L2, which is called by https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/get-user-classification-counts.js#L5, which in turn is called on `componentDidMount` and `componentWillReceiveProps` of the new container.

simoneduca:
Restaged at https://fix-1048.pfe-preview.zooniverse.org/

eatyourgreens:
My stats page shows up differently on this branch from the live site. The projects are all there, but the order is mixed up and the grey bar on the right is missing.

simoneduca:
Thanks Jim. I'm a bit stuck on this, I think I've got some coffeescript wrong. Could probably do another pair of eyes.

PR #3294, opened 2016-11-30
Decode blog excerpts.

simoneduca:
@srallen, Michael's comment on Slack made me realise that it makes sense to sanitise blog excerpts too.


parrish:
👍

PR #3299, opened 2016-12-01
Add workflow stats visibility setting

srallen:
This adds a workflow configuration setting to hide a workflow from the project's statistics page. Use case is if a project is using an active workflow for a specific a/b split and it doesn't make sense to make the stats for that workflow visible to volunteers. Gravity Spy currently has a split going on using an active workflow that will be made inactive once the experiment is done. Conceivably there may be other use cases for project builders to hide an active workflow from the stats page.

workflow id `3063` should not now be visible on Gravity Spy's stats page with this update:
https://hide-stats.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy/stats?env=production

# Review Checklist

- [x] Does it work on mobile?

## Optional


CKrawczyk:
LGTM

PR #3310, opened 2016-12-07
Upgrade to React 15.4

eatyourgreens:
Depends on zooniverse/markdownz#52 and zooniverse/Seven-Ten-Client#2


eatyourgreens:
Looks like this breaks Talk. Navigation stops working after posting a comment.

Also, there are a bunch of errors because markdownz is on React 15.3.

eatyourgreens:
Getting [Refs must have an owner](https://facebook.github.io/react/warnings/refs-must-have-owner.html) errors from components that use `MarkdownEditor`. Might be caused by multiple copies of React but I'm not sure.

simoneduca:
Is this still breaking Talk?
Also it looks like in preparation for major v16, the new v15.5.0 has introduced some breaking changes https://facebook.github.io/react/blog/2017/04/07/react-v15.5.0.html.
I wonder if it's worth jumping to 15.5.0 and skipping 15.4.0?

eatyourgreens:
@simoneduca the Talk problem is caused by trying to run multiple different versions of React and use named refs. Simple fix is to update React in the dependencies via the two PRs above, then try to keep everything in synch in future.

15.5 sounds like it breaks most of PFE, particularly the old coffeeScript components. Have we got time to go through and fix those old files to work with the `createClass` change?

simoneduca:
Got it. But I don't understand if your question about having time is rhetorical. It'd needs doing at some point, unless we want to pin PFE to 15.4, right?

eatyourgreens:
My question was serious. It looks like a lot of work to me, so worth discussing with the whole team.

srallen:
I think we should just upgrade to 15.4 for now and prioritize the rest of the coffeescript to ES6 conversion, classifier improvements, and fixes to the panoptes/json api clients.

eatyourgreens:
rebased and updated to add the updated markdownz.

eatyourgreens:
Staged at https://react-15.pfe-preview.zooniverse.org/

Talk should work now, with the markdownz update. seven-ten might be broken.

simoneduca:
Talk seems to be working fine and the site as a whole as well.
Regarding seven-ten I can see warnings like`GET https://seven-ten.zooniverse.org/split_user_variants?filter[projects.slug]=michiganzoomin/michigan-zoomin 404 (Not Found)` but  I'm not sure if it's problematic.

camallen:
404's are expected if there is nothing setup in seven-ten. we need to test this on a project that's using it.

simoneduca:
I can confirm seven-ten calls are all 200 and the data seems ok for Gravity Spy, which is using currently using the service.
So, LGTM.

PR #3311, opened 2016-12-07
Updated call for projects

hrspiers:
Hi,

I've updated the call for projects to include the Text Transcription call for projects.

Thanks
Helen


parrish:
👍

PR #3318, opened 2016-12-09
Adding in markdown toolbar to Announcements textbox

### LINKED ISSUE #2318: Adding the editing tool bar to the annoucements text box
Likely for announcements you might want to link to a blog post since it's for site news. It would be great especially for those who don't know markdown as well to include the tool bar that's on the 

screen shot 2016-03-17 at 8 35 26 am

mschwamb:
fixes #2318
took a stab at adding the markdown editing toolbar to the Announcements textbox


mschwamb:
@srallen I took a stab at this since it looked like a straight up copy from the syntax used for text boxes on the page.

mschwamb:
@srallen  thanks. I've made that change.

PR #3321, opened 2016-12-09
Replacing old profile stats page.

### LINKED ISSUE #1048: Add total classification count to stats page

simoneduca:
Add total classifications to user's home.
Add individual classifications to recent projects.
Removes profile stats page.
Fix #1048 

Suggestion for next PR: Recent projects now shows _any_ project with more than 0 classifications up to 20 projects. So the name "recent projects" doesn't seem appropriate anymore.
Add some sorting or reinstate the "see all" button and show all projects with paging.

Staged at https://add-classifications-recent-projects.pfe-preview.zooniverse.org/


astopy:
Looks good, but the "Your Stats" link on the profile page needs to be updated or removed.

eatyourgreens:
I can't find my most recent projects any more, to quickly go back to them. I think it should be sorted by `-updated_at`, like the current API query. Maybe also show just a few with a button to show the full set?

simoneduca:
Good point Adam, done.
Jim, on it :)

simoneduca:
I made a start on the slide for seeing all projects, but it's not finished. Feel free to jump on it and edit it.

eatyourgreens:
@simone I think d238c7c works correctly, but looks pretty ugly.

eatyourgreens:
I've added a little bit of animation and tidied up the layout a bit. By the way, when you're testing, I found it helps to use an account that has classified on a lot of projects. Otherwise, we might break project stats for the volunteers who use it the most.

eatyourgreens:
Just a note that this is still pretty slow. I'm waiting ~5-6s for my recent projects to appear (the page has to make 5 or 6 requests to get all the project preferences.)

We should maybe update it so it starts rendering before all of the network requests have completed, or maybe not load all of the project stats until they are specifically requested by the recent projects component.

camallen:
@eatyourgreens agree it should render and re-render as new data comes in. I've got to add some caching api side that should speed this request up. The data doesn't change much and it's on the list for this week.

eatyourgreens:
@simoneduca @camallen I've written a new user ribbon data function that updates state as each set of project preferences comes back from the API. Staged at https://add-classifications-recent-projects.pfe-preview.zooniverse.org/?env=production It feels much faster, even though the total time to gather all of your project data is exactly the same.

eatyourgreens:
Just a note for myself that projects are still sorted in the wrong order. The preferences request should use `sort: '-updated_at'` and projects should be rendered in the order that the user was last active on them, not when the project was last updated.

eatyourgreens:
5f69e39 should sort project preferences by `updated_at` descending, giving me projects back in the order that I was last active on them.

However, I'm still getting different Recent Projects on https://add-classifications-recent-projects.pfe-preview.zooniverse.org/?env=production#focus=projects compared with https://www.zooniverse.org/#focus=projects

Can anyone see what I'm doing wrong here?

eatyourgreens:
Cheers @camallen I think you are exactly right. I need to preserve the order that comes back from that initial request for preferences.

eatyourgreens:
Project preference sort order should be preserved now in the projects array. I think this is ready to go @camallen

eatyourgreens:
I've gone through and fixed a bunch of linter errors that were easy to fix.

simoneduca:
Woot, thanks for picking this up Jim!

PR #3327, opened 2016-12-12
edit survey task keys to include vowels

### LINKED ISSUE #3290: Survey values are weirdly processed in data exports
@mkosmala commented on [Wed Nov 30 2016](https://github.com/zooniverse/Panoptes/issues/2066)

Treeversity is a project that uses the survey tool. When I look at the data from the survey in the data exports, I see that the chosen values are all-caps versions of the original text with all spaces and vowels(!) removed. So, for example, the option "Leaf" becomes "LF" and the option "Autumn leaf color" becomes "TMNLFCLR". This is really unintuitive for project groups. I'm okay with dropping spaces, but please keep the vowels! It would also help to keep the original capitalization.

---

@camallen commented on [Wed Nov 30 2016](https://github.com/zooniverse/Panoptes/issues/2066#issuecomment-263919384)

@mkosmala what do the task survey key => values look like in the workflow? I think this is a vestige of the survey csv uploader setting those keys.

---

@mkosmala commented on [Wed Nov 30 2016](https://github.com/zooniverse/Panoptes/issues/2066#issuecomment-263921316)

@camallen The CSV that we upload doesn't have a key column. We have a name and a description. I'm sure something is converting names to (weird) keys, but I don't Panoptes well enough to know where. The Project Builder doesn't have anything other than names showing. I'm pulling a workflow export right now...
CC: @donnytree

---

@eatyourgreens commented on [Wed Nov 30 2016](https://github.com/zooniverse/Panoptes/issues/2066#issuecomment-263929123)

Looks like this is something that the survey task editor is doing in the front end here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/survey/editor.cjsx#L370

mcbouslog:
Fixes #3290 (front-end issue, back-end issue [2066](https://github.com/zooniverse/Panoptes/issues/2066) pending).

Only applies to subsequent survey tasks, or subsequently re-imported survey tasks.

Historically with the survey task editor, to create the object keys for choices, characteristics and questions - vowels and whitespace were removed from the related human-readable label/string, and the string was capitalized.

This PR edits the regex to keep the vowels from the human-readable label/string, and leaves the removal of whitespace and capitalization as is.

This makes the keys more human-readable (as requested per #3290), and also makes the keys more unique (i.e. previously "Bear" and "Boar"'s keys would both be `br`, which would cause Boar's values to replace Bear's).

# Review Checklist

- [ ] Does it work on mobile?
- [X] Did you deploy a [staging branch](https://edit-survey-task-key.pfe-preview.zooniverse.org/)?

## Optional

### DIFF
--- a/app/classifier/tasks/survey/editor.cjsx
+++ b/app/classifier/tasks/survey/editor.cjsx
@@ -367,7 +367,7 @@ module.exports = React.createClass
     value?.charAt(0).toUpperCase() in ['T', 'X', 'Y', '1']
 
   makeID: (string) ->
-    string.replace(/[aeiouy\W]/gi, '').toUpperCase()
+    string.replace(/\W/g, '').toUpperCase()
 
   createChoice: (name) ->
     choiceID = @makeID name


mcbouslog:
as expected and intended, the classifications data export CSV's annotations include a lot more characters now (classifications from workflow 2434 are with historic no-vowel keys, from workflow 2771 are new keys with vowels) - [survey-task-test-project-classifications.xlsx](https://github.com/zooniverse/Panoptes-Front-End/files/647083/survey-task-test-project-classifications.xlsx)

mentioning @camallen

PR #3328, opened 2016-12-12
ESLint rule update

srallen:
Adds a rule for multiline ternaries to warn. Are there any other rules we want to add or update?

# Review Checklist

- [ ] Does it work on mobile?

## Optional


parrish:
Just a few preferences:

  - [`"object-curly-spacing"`](http://eslint.org/docs/rules/object-curly-spacing): `["error", "always", { "objectsInObjects": false }]`
    - Probably a holdover from Ruby convention, but it seems more readable to me
  - [`"array-callback-return"`](http://eslint.org/docs/rules/array-callback-return): `"off"`
    - A lot of times you're not worried about returning the result
  - [`"comma-dangle"`](http://eslint.org/docs/rules/comma-dangle): `["error", "never"]`
    - Such a weird convention to encourage something that'd break syntax in any other language.

srallen:
Also noting that based out of slack discussion: set ["jsx-no-bind"](https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-bind.md) to warn instead.

srallen:
I made the `"comma-dangle"` rule warn instead of error since there are a few refactors already using that rule and it might be annoying to have those error now.

Issue #2397, opened 2016-04-07
Field guide should not show on the landing page of a project 


mschwamb:
At least in Firefox the field guide shows up on the landing page of a project with an active field guide. I think that's a bit confusing as you're not sure what you're supposed to click or go to. 

<img width="1146" alt="screen shot 2016-04-07 at 7 03 09 am" src="https://cloud.githubusercontent.com/assets/753325/14353706/2317a75a-fc8f-11e5-87e3-88fb0f451b38.png">

brian-c:
This is as intended, the field guide is available anywhere in the project.

mschwamb:
I get the having the field guide showing up everywhere else on a project website , but I find it odd to have it show on the landing page. Other previous custom Zooniverse projects with site guides have a clean landing page. My thinking is that if you're new and checking out the project, you have no idea the task but you would have the ability to review the field guide without even knowing what the task of the project is, and that the landing page should as minimal as possible to get people straight in to classify and into then task.

chrislintott:
I agree with Meg, though I think it’s a low priority fix.

PR #3329, opened 2016-12-12
Remove Field Guide from Project Homepage

wgranger:
Fixes #[2397](https://github.com/zooniverse/Panoptes-Front-End/issues/2397)

Describe your changes.
This removes the Field Guide tab from the right side of the project landing page. The tab will still be available on other project sections: Talk, About, Classify, Collect.

# Review Checklist

- [X] Does it work on mobile?

https://field-guide-landing.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -218,6 +218,7 @@ ProjectPage = React.createClass
 
   render: ->
     projectPath = "/projects/#{@props.project.slug}"
+    onHomePage = projectPath is @props.location.pathname
 
     pages = [{}, @state.pages...].reduce (map, page) =>
       map[page.url_key] = page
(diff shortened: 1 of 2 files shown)


app/pages/project/index.cjsx
```diff
@@ -311,7 +312,7 @@ ProjectPage = React.createClass
       {unless @props.project.launch_approved or @props.project.beta_approved
         <Translate component="p" className="project-disclaimer" content="project.disclaimer" />}
 
-      <PotentialFieldGuide project={@props.project} />
+      <PotentialFieldGuide project={@props.project} onHomePage={onHomePage} />
```
srallen: Instead of passing down a prop, you can just directly check here and not render. In coffeescript, something like this should work:
```
unless @props.location.pathname is "/projects/#{@props.project.slug}"
  <PotentialFieldGuide />
```

wgranger:
Well, that cleaned things up a bit.

Issue #3216, opened 2016-11-11
Components that still use PromiseRenderer


rogerhutchings:
~~This _doesn't_ include Talk...~~ it totally does

- [x] `app/collections/favorites-button.cjsx`
- [x] `app/collections/collaborators.cjsx`
- [x] `app/collections/settings.cjsx`
- [x] `app/collections/show-list.cjsx`
- [x] `app/collections/show.cjsx`
- [x] `app/components/media-area.cjsx`
- [x] `app/components/promise-renderer.cjsx` (#6243. We'll just delete this once the other components stop using it)
- [x] `app/pages/admin/project-status-list.cjsx`
- [x] `app/pages/admin/project-status.cjsx` (in #3331)
- [x] `app/pages/admin/user-settings.cjsx` (in #3331)
- [x] `app/pages/lab/collaborators.cjsx` (#6233)
- [x] `app/pages/lab/project-details.cjsx` (in #3322)
- [x] `app/pages/lab/project.cjsx` (#6219)
- [x] `app/pages/lab/subject-set.cjsx`
- [x] `app/pages/lab/workflow-viewer/index.cjsx` (James, see (see https://github.com/zooniverse/Panoptes-Front-End/issues/3912))
- [x] `app/pages/lab/workflow.cjsx` (#6228)
- [x] `app/pages/profile/index.cjsx` (in #3342)
- [x] `app/pages/profile/stats.cjsx`
- [x] `app/pages/project/about/education.cjsx`
- [x] `app/pages/project/about/faq.cjsx`
- [x] `app/pages/project/about/index.cjsx`
- [x] `app/pages/project/about/research.cjsx`
- [x] `app/pages/project/about/results.cjsx`
- [x] `app/pages/project/about/team.cjsx`
- [x] `app/pages/settings/email.cjsx`(in #4182)
- [x] `app/pages/settings/profile.cjsx`(in #4142)
- [x] `app/partials/data-export-button.cjsx`
- [x] `app/partials/display-name-slug-editor.cjsx`

simoneduca:
Currently removing it from `app/pages/project/about/team.cjsx`

rogerhutchings:
cc @eatyourgreens @simoneduca 

ref #3215

eatyourgreens:
See #3080 for examples of removing these from the Talk components.

eatyourgreens:
I've made a start on collections this afternoon.

eatyourgreens:
[...]

PR #3330, opened 2016-12-13
Refactor media area to ES6. Remove promise renderer

srallen:
For #3216, refactors the `media-area.cjsx` into ES6 and removes the `PromiseRenderer`. Since this file was technically a set of components, I broke them out, created a "controller" component to handle the data and a separate view component as its child. I also moved it into the lab folder since it is used by the `EditMediaPage` component and the survey task editor. `Thumbnail` remains in the components folder since it is used by talk.

This could use some tests, but I'd like to do that separately since this is already touching so many files. If you'd prefer tests coming in with this though, I can do that.

Staged at https://refactor-media-area.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional


srallen:
Just noticed I broke some drag and drop target styles. Gonna look at this morning.

srallen:
Also, this [talk comment](https://www.zooniverse.org/talk/18/144354) seems to be about `504` and `412` errors being thrown on the `PUT` request on the workflow when there's a large batch of images dropped onto the target for upload. Gonna take a look at my refactor to see if that's still an issue with it.

srallen:
I've fixed the drag and drop styling. For the talk comment, I moved the error div to be at the top of the media area to make sure it is seen. There were probably errors in the UI when the issue described in the Talk thread happened, but they couldn't be seen previously since there was enough media causing an overflow and scroll bar and the error div was originally at the bottom of all of that.

app/components/thumbnail.jsx
```diff
@@ -0,0 +1,68 @@
...
+        <div>
+          <video width="300" controls>
+            <source src={this.props.src} type="video/mp4" />
+           </video>
+        </div>
+      );
+    }
+
+    return (
+      <img alt="" {...this.props} src={src} {...dimensions} style={style} onError={this.handleError} />
```
mcbouslog: more for my understanding than a review note, what does destructuring `...this.props` give the `img` component we're not already setting individually? or more likely I'm misreading.

app/components/thumbnail.jsx
```diff
@@ -0,0 +1,68 @@
...
+    };
+  }
+
+  getThumbnailSrc({ origin, width, height, src }) {
+    const srcPath = src.split('//').pop();
+    return (`${origin}/${width}x${height}/${srcPath}`);
+  }
+
+  handleError() {
+    if (!this.state.failed) {
```
mcbouslog: when attempting to upload the FilterIcons (.svg's) from Ali's sample survey task via Project Builder, Media or Survey Task editor, console had 415 errors and the following error message:
```
thumbnail.jsx?bf6a:20 Uncaught TypeError: Cannot read property 'state' of null(…)
```

Uploading ReferencePhotos (.jpg's) worked great.

parrish: `this` is unbound here.  Adding `this.handleError = this.handleError.bind(this);` to the constructor fixes it

app/components/thumbnail.jsx
```diff
@@ -0,0 +1,68 @@
...
+      maxWidth: this.props.width,
+      maxHeight: this.props.height,
+    };
+
+    if (this.props.format === 'mp4') {
+      return (
+        <div>
+          <video width="300" controls>
+            <source src={this.props.src} type="video/mp4" />
+           </video>
```
parrish: Alignment

app/components/thumbnail.jsx
parrish: That's from what I mentioned here: https://github.com/zooniverse/Panoptes-Front-End/pull/3330#discussion_r92707937

app/components/thumbnail.jsx
parrish: It adds a bit of flexibility to the component (e.g. using arbitrary image attributes)

srallen:
Thanks for catching those. @parrish, I've restaged for another review.

PR #3331, opened 2016-12-13
Admin updates

parrish:
Towards #3216 

https://admin-updates.pfe-preview.zooniverse.org/

Rewrites admin/project-status and admin/user-settings in ES6 and removes PromiseRenderer and ChangeListener.

# Review Checklist

- [x] Does it work on mobile?


app/pages/admin/project-status.jsx
```diff
@@ -0,0 +1,149 @@
...
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
+  getProject() {
+    const slug = `${this.props.params.owner}/${this.props.params.name}`;
+    return apiClient.type('projects').get({ slug }).then((projects) => {
+      const project = projects[0];
+      // TODO: We ought to improve this ChangeListener replacement
+      project.listen('change', this.forceUpdate.bind(this));
```
parrish: I guess I need to stopListening at some point too...  One sec...

parrish:
Cool.  This should be good to go now.

app/pages/admin/project-status/toggle.jsx
```diff
@@ -0,0 +1,71 @@
...
+  constructor(props) {
+    super(props);
+    this.set = this.set.bind(this);
+    this.state = {
+      error: null,
+      setting: { }
+    };
+  }
+
+  set(property, value) {
```
srallen: Directly modifying the state is an anti-pattern. Can this be reworked to use `setState`?

app/pages/admin/user-settings.jsx
```diff
@@ -0,0 +1,78 @@
...
+
+  componentDidMount() {
+    if (this.props.editUser != null) {
+      this.props.editUser.listen('change', this.boundForceUpdate);
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (this.props.editUser != null) {
+      this.props.editUser.stopListening('change', this.boundForceUpdate);
```
srallen: This should also stop listening on `componentWillUnmount`

app/pages/admin/user-settings/limit-toggle.jsx
```diff
@@ -0,0 +1,33 @@
+import React, { Component } from 'react';
+
+import AutoSave from '../../../components/auto-save';
+import handleInputChange from '../../../lib/handle-input-change';
+
+class LimitToggle extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      error: null
```
srallen: Maybe I'm just missing it, but does this get set by anything?

app/pages/admin/user-settings.jsx
```diff
@@ -0,0 +1,78 @@
...
+    const handleChange = handleInputChange.bind(this.props.editUser);
+
+    return (
+      <div className="project-status">
+        <h4>Settings for {this.props.editUser.login}</h4>
+        <ul>
+          <li>
+            <AutoSave resource={this.props.editUser}>
+              <input type="checkbox" name="admin" checked={this.props.editUser.admin} disabled onChange={handleChange} />{' '}
+              Admin
```
srallen: I don't think the admin toggle actually works. I think the API returns an error and it can only be set directly on the db. This should probably be instead only a read only field that shows us who's an admin. I'm ok with this change being a separate PR, though, if you'd rather not add that for this PR.

app/pages/admin/project-status.jsx
```diff
@@ -0,0 +1,160 @@
...
+    this.setState({ error: null });
+    let selected = event.target.value;
+    selected = selected === 'none' ? undefined : selected;
+    return workflow.update({ 'configuration.level': selected }).save()
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
+  rerender() {
+    if (this.isMounted()) {
```
srallen: The latest version of React is unhappy with us using this: 
```
Warning: isMounted(...) is deprecated in plain JavaScript React classes. Instead, make sure to clean up subscriptions and pending requests in componentWillUnmount to prevent memory leaks.
Uncaught TypeError: this.isMounted is not a function(…)
```

srallen: Just to add a note: this error threw when I clicked on visibility setting for a project.

app/pages/admin/project-status/toggle.jsx
parrish: This was a pretty literal translation of the mixin code.  Should be able to improve it

app/pages/admin/user-settings.jsx
parrish: `admin`, `login_prompt`, and `private_profile` were all disabled before this PR -- I just re-implemented what was there.  Happy to remove the `AutoSave` and `onChange`.  You're call

app/pages/admin/user-settings/limit-toggle.jsx
parrish: Hmm.. Nope, also vestigial from the original code

parrish:
@srallen Should be fixed up.  Let me know if I missed anything

parrish:
Nah, I'm happy for a merge unless you want me to add those `propType`s in first

srallen:
Oh yeah, just for consistency, let's add those `propType`s in. But otherwise this is good to merge.

parrish:
Updated with `propType`s

PR #3339, opened 2016-12-15
Play video on click

wgranger:
Describe your changes.
This allows the ability for a video to play when clicking on the still image. A boolean prop is also added to give the component the ability to hide controls, if needed. Bat detective has video subjects to test functionality.

# Review Checklist

- [x] Does it work on mobile?

https://video-click-play.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/video-player.jsx
+++ b/app/components/video-player.jsx
@@ -11,19 +11,19 @@ const VideoPlayer = React.createClass({
     onLoad: React.PropTypes.func,
     showControls: React.PropTypes.bool,
     src: React.PropTypes.string,
-    type: React.PropTypes.string,
+    type: React.PropTypes.string
   },
 
   getDefaultProps() {
     return {
-      showControls: true,
+      showControls: true
     };
   },
 
   getInitialState() {
     return {
       playing: false,
-      playbackRate: 1,
+      playbackRate: 1
     };
   },
 
@@ -99,11 +99,6 @@ const VideoPlayer = React.createClass({
     const rates = [0.25, 0.5, 1];
     return (
       <div className="subject-video-frame">
-        <button
-          className="subject-video-frame__hidden-button"
-          onClick={this.playVideo.bind(this, !this.state.playing)}
-        />
-
         <video
           className="subject"
           ref="videoPlayer"
@@ -111,6 +106,7 @@ const VideoPlayer = React.createClass({
           type={`${this.props.type}/${this.props.format}`}
           preload="auto"
           onCanPlay={this.props.onLoad}
+          onClick={this.playVideo.bind(this, !this.state.playing)}
           onEnded={this.endVideo}
           onTimeUpdate={this.updateScrubber}
         >
@@ -155,7 +151,7 @@ const VideoPlayer = React.createClass({
         {this.props.children}
       </div>
     );
-  },
+  }
 
 });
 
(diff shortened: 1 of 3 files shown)


wgranger:
The linter complained about adding a click listener to a `<video>` tag, so an invisible button overlay was added. I'm open to other suggestions or reverting back to the video click listener.

eatyourgreens:
Clicks on the video element are being blocked by the SVG image layer. Adding another transparent button on top isn't the way to go to solve this, and will look like a clickjacking technique to any security plugins.

I'd suggest just remove the SVG layer so that the video element can be clicked directly.

app/components/video-player.jsx
```diff
@@ -84,14 +91,19 @@ const VideoPlayer = React.createClass({
...
     return (
       <div className="subject-video-frame">
+        <button
+          className="subject-video-frame__hidden-button"
+          onClick={this.playVideo.bind(this, !this.state.playing)}
+        />
+
         <video
           className="subject"
           ref="videoPlayer"
```
eatyourgreens: I think you've missed out `controls={!this.props.showControls}` on the `video` element here. I'm assuming that the browser's native controls should show if the custom controls are turned off.

wgranger: The idea was to hide all controls for the video element, including native controls. This was to adhere to styling for the new project landing page.

wgranger:
Ok, the `onClick` was added to the video element. This onClick actually worked before, with or without the svg layer, but the linter gave warnings about attaching an onClick to that element.

wgranger:
screen shot 2016-12-19 at 1 57 36 pm

This is an example of the styling. However, if these subjects were videos, they could be played with a click.

eatyourgreens:
I'm a bit confused, because you can already click the video element to play videos outside the classifier e.g.  https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/230/146363

wgranger:
I think that link allows you to play videos on click on mobile, but not desktop.

eatyourgreens:
How about subjects that use the `Thumbnail` component instead of the `SubjectViewer`? https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/tags/insect

eatyourgreens:
Was the linter warning about adding `onClick` to an element that can't receive keyboard focus? That's pretty easy to fix too.

Otherwise, this looks good to me.

eatyourgreens:
I've just done a quick check using the keyboard in Chrome for Arizona Batwatch and the native video controls at https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/tags/insect. You can tab to individual videos and the video element will receive focus. You can then tab to the play button, within the video element (it's hidden in the shadow DOM for that element), and toggle it with the spacebar when it has focus.

I didn't check any other browsers, but I think that `jsx-a11y` warning about adding click to elements without corresponding keyboard interaction is wrong here. Keyboard support might vary between browsers, though.

wgranger:
No, the particular linter gave the warning, `visible, non-interactive elements should not have mouse or keyboard event listeners`

eatyourgreens:
Yeah, that's an accessibility warning from `jsx-a11y`. Fine to ignore it here, if the video can be controlled with the keyboard.

PR #3342, opened 2016-12-16
Rewrite the user profile in ES6

parrish:
Towards #3216 

Staged at https://es6-profile.pfe-preview.zooniverse.org

# Review Checklist

- [x] Does it work on mobile?

## Optional


parrish:
> The user prop in index.jsx needs a default prop.

Yeah?  We shouldn't need a default for the current user, right?

parrish:
I updated the centering though

srallen:
I think I was getting `profileUser` and `user` mixed up when I was reading the code. Looking it over again, In this case, it doesn't really matter if the user prop is undefined or set as null as a default while waiting on the data from the parent. `profileUser` has a default as needed in the child component.

PR #3365, opened 2017-01-03
Mark all read

wgranger:
Describe your changes.
This adds a "Mark All as Read" button to the notifications page. This was a suggestion brought up by a volunteer on general talk. This PR also adjusts how the unread count appears, as numbers over 99 were appearing off center.

# Review Checklist

- [x] Does it work on mobile?

https://mark-all-read.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/notifications/notification-section.jsx
+++ b/app/pages/notifications/notification-section.jsx
@@ -200,7 +200,7 @@ export default class NotificationSection extends Component {
           </div>
 
           <div className="notification-section__item">
-            <button className="notification-section__expand" title="Toggle Section">
+            <button title="Toggle Section">
               <i className={buttonType} />
             </button>
           </div>
--- a/app/pages/notifications/notification-section.jsx
+++ b/app/pages/notifications/notification-section.jsx
@@ -119,15 +119,15 @@ export default class NotificationSection extends Component {
   }
 
   markAllRead() {
-    talkClient.put('/notifications/read', { section: this.props.section });
     this.state.notificationData.forEach((data) => {
       data.notification.update({ delivered: true });
     });
     this.setState({ unread: 0 });
-    return talkClient.type('notifications').get({ page_size: 1, delivered: false })
-    .then(([notification]) => {
-      const count = notification.getMeta().count;
-      if (count === 0) this.context.notificationsCounter.setUnread(0);
+    return talkClient.put('/notifications/read', { section: this.props.section }).then(() => {
+      return talkClient.type('notifications').get({ page_size: 1, delivered: false }).then(([notification]) => {
+        const count = notification ? notification.getMeta().count : 0;
+        if (count === 0) this.context.notificationsCounter.setUnread(0);
+      });
     });
   }
 
@@ -223,6 +223,8 @@ export default class NotificationSection extends Component {
 
   render() {
     const l = this.state.currentMeta;
+    const firstNotification = (l.page * l.page_size) - (l.page_size - 1) || 0;
+    const lastNotification = Math.min(l.page_size * l.page, l.count) || 0;
 
     return (
       <div className="notification-section">
@@ -274,7 +276,7 @@ export default class NotificationSection extends Component {
               pageCount={this.state.lastMeta.page_count}
               pageSelector={false}
               previousLabel={<span><i className="fa fa-chevron-left" /> previous</span>}
-              totalItems={<span className="notification-section__item-count">{(l.page * l.page_size) - (l.page_size - 1)} - {Math.min(l.page_size * l.page, l.count)} of {l.count}</span>}
+              totalItems={<span className="notification-section__item-count">{firstNotification} - {lastNotification} of {l.count}</span>}
             />
           </div>
         )}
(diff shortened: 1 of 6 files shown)


app/pages/notifications/notification-section.jsx
```diff
@@ -117,6 +118,15 @@ export default class NotificationSection extends Component {
     });
   }
 
+  markAllRead() {
+    talkClient.put('/notifications/read', { section: this.props.section });
+    this.state.notificationData.forEach((data) => {
+      data.notification.update({ delivered: true }).save();
```
parrish: You don't want to call `.save()` on these -- the update is already done in the put request

app/pages/notifications/notification-section.jsx
```diff
@@ -220,6 +237,13 @@ export default class NotificationSection extends Component {
           </span>
         )}
 
+        {(this.props.expanded && this.state.unread > 0) && (
+          <label htmlFor="markRead">
+            <input type="checkbox" onChange={this.markAllRead} />
```
parrish: This probably ought to be a button rather than a checkbox

app/pages/notifications/notification-section.jsx
```diff
@@ -117,6 +118,15 @@ export default class NotificationSection extends Component {
     });
   }
 
+  markAllRead() {
+    talkClient.put('/notifications/read', { section: this.props.section });
+    this.state.notificationData.forEach((data) => {
+      data.notification.update({ delivered: true }).save();
+    });
+    this.setState({ unread: 0 });
+    this.context.notificationsCounter.setUnread(0);
```
parrish: You can't really use this to clear the counter since you're only marking one section as read -- e.g. you don't know if the other sections contain unread or not.

parrish: An alternative is to `talkClient.type('notifications').get({ page_size: 1, delivered: false })`, get the count from the meta, and update if zero.

wgranger:
Good suggestions. Restaged with changes.

app/pages/notifications/notification-section.jsx
```diff
@@ -117,6 +118,19 @@ export default class NotificationSection extends Component {
     });
   }
 
+  markAllRead() {
+    talkClient.put('/notifications/read', { section: this.props.section });
+    this.state.notificationData.forEach((data) => {
+      data.notification.update({ delivered: true });
+    });
+    this.setState({ unread: 0 });
```
parrish: Looks like this is a bit nondeterministic in the unread check.  Sometimes the read request finishes before the unread check, sometimes vice versa.

app/pages/notifications/notification-section.jsx
```diff
@@ -117,6 +118,19 @@ export default class NotificationSection extends Component {
...
+    return talkClient.type('notifications').get({ page_size: 1, delivered: false })
+    .then(([notification]) => {
+      const count = notification.getMeta().count;
```
parrish: `notification` can be null here

parrish: Just playing around with this.  Maybe something like this?

```
markAllRead() {
    this.state.notificationData.forEach((data) => {
      data.notification.update({ delivered: true });
    });
    this.setState({ unread: 0 });
    return talkClient.put('/notifications/read', { section: this.props.section }).then(() => {
      return talkClient.type('notifications').get({ page_size: 1, delivered: false }).then(([notification]) => {
        const count = notification ? notification.getMeta().count : 0;
        if (count === 0) this.context.notificationsCounter.setUnread(0);
      });
    });
  }
```

parrish:
I'm seeing some weird warnings too.


```
Warning: Component's children should not be mutated.
in span (created by NotificationSection)
```

```
Warning: Unknown props `notification`, `user`, `startedDiscussion` on <a> tag. Remove these props from the element. For details, see https://fb.me/react-unknown-prop.
in a (created by Link)
in Link (created by CommentNotification)
```

Uncaught (in promise) TypeError: Cannot read property 'getMeta' of undefined


Not sure exactly when/where these are coming from.

wgranger:
Those errors were coming from a couple places. The mutations one was coming from the Paginator child of the NotificationSection component. The `Unknown props` error occurred because unknown props were fed into the Link component due to the splat operator being used.

app/pages/notifications/notification-section.jsx
```diff
@@ -274,7 +276,7 @@ export default class NotificationSection extends Component {
               pageCount={this.state.lastMeta.page_count}
               pageSelector={false}
               previousLabel={<span><i className="fa fa-chevron-left" /> previous</span>}
-              totalItems={<span className="notification-section__item-count">{(l.page * l.page_size) - (l.page_size - 1)} - {Math.min(l.page_size * l.page, l.count)} of {l.count}</span>}
+              totalItems={<span className="notification-section__item-count">{firstNotification} - {lastNotification} of {l.count}</span>}
```
parrish: Oh.  This was the cause of the `Component's children should not be mutated` warning?

wgranger: Yes, passing in calculations threw a warning, but getting the value beforehand worked alright.

parrish: Pfff..  that's bizarre.  Good catch though

PR #3371, opened 2017-01-05
Allow Video Subjects to Show with Discussion Preview

### LINKED ISSUE #3352: Video Subject Previews are Broken
## Expected behavior
Some sort of preview should be shown for video subjects, or the preview should be omitted.

## Current behavior
Video subjects appear broken, likely because the preview is expecting an image.

## Steps to replicate
https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/230
Take a look at the "Notes" board on Arizona Batwatch talk. There is a broken image icon near the subject title.

## Additional information

- **Operating system:**
El Capitan
- **Browser:**
Chrome

wgranger:
Fixes #3352 .

Describe your changes.
Video subjects were not showing on the talk discussion preview. This is apparent when looking at Bat Detective, for example.

# Review Checklist

- [X] Does it work on mobile?

https://fix-3352.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/thumbnail.jsx
+++ b/app/components/thumbnail.jsx
@@ -47,9 +47,10 @@ export default class Thumbnail extends React.Component {
     };
 
     if (this.props.format === 'mp4') {
+      const width = this.props.width < 999 ? this.props.width : '300';
       return (
         <div>
-          <video width="300" controls onClick={this.playVideo}>
+          <video width={width} controls={this.props.controls} onClick={this.playVideo}>
             <source src={this.props.src} type="video/mp4" />
           </video>
         </div>
@@ -63,17 +64,19 @@ export default class Thumbnail extends React.Component {
 }
 
 Thumbnail.defaultProps = {
+  controls: true,
   format: 'image',
   height: MAX_THUMBNAIL_DIMENSION,
   origin: 'https://thumbnails.zooniverse.org',
   src: '',
-  width: MAX_THUMBNAIL_DIMENSION,
+  width: MAX_THUMBNAIL_DIMENSION
 };
 
 Thumbnail.propTypes = {
+  controls: React.PropTypes.bool,
   format: React.PropTypes.string,
   height: React.PropTypes.number,
   origin: React.PropTypes.string,
   src: React.PropTypes.string,
-  width: React.PropTypes.number,
+  width: React.PropTypes.number
 };
--- a/app/components/thumbnail.jsx
+++ b/app/components/thumbnail.jsx
@@ -47,10 +47,9 @@ export default class Thumbnail extends React.Component {
     };
 
     if (this.props.format === 'mp4') {
-      const width = this.props.width < 999 ? this.props.width : '300';
       return (
         <div>
-          <video width={width} controls={this.props.controls} onClick={this.playVideo}>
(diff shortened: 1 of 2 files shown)


app/components/thumbnail.jsx
```diff
@@ -47,9 +47,10 @@ export default class Thumbnail extends React.Component {
     };
 
     if (this.props.format === 'mp4') {
+      const width = this.props.width < 999 ? this.props.width : '300';
       return (
```
eatyourgreens: I don't think you should be overriding the default width here. If the default value isn't sensible, then change that instead.

app/components/thumbnail.jsx
```diff
@@ -47,9 +47,10 @@ export default class Thumbnail extends React.Component {
...
         <div>
-          <video width="300" controls onClick={this.playVideo}>
+          <video width={width} controls={this.props.controls} onClick={this.playVideo}>
```
eatyourgreens: Maybe pass `this.props.height` to the video element as well as width, and set the same styles for max width and height as are set for images.

app/talk/discussion-preview.cjsx
```diff
@@ -62,8 +62,9 @@ module.exports = React.createClass
       <div className="preview-content">
 
         {if @state.subject?
+          subject = getSubjectLocation(@state.subject)
           <div className="subject-preview">
-            <Thumbnail src={getSubjectLocation(@state.subject).src} width={100} />
+            <Thumbnail src={subject.src} format={subject.format} width={100} controls={false} />
```
eatyourgreens: Why are video controls hidden here? Won't this prevent the videos from being controlled via the keyboard?

wgranger: The controls cover the video too much, and since this is a small Thumbnail, I figure it's important to show as much of the subject as possible. The Thumbnail can still be clicked to play.
screen shot 2017-01-06 at 9 40 06 am

eatyourgreens: That's fine for people who use a mouse, or similar pointing device, to interact with the video, but renders it inaccessible for people who primarily interact with the page via the keyboard. I'm not sure how big of a deal that is for the small previews, but it's something to bear in mind when adding click interactions to DOM elements that aren't buttons or links.

Do you think it would be better if these small previews were wrapped in links, so that clicking on them went through to the subject page? There might actually be a very old PFE issue asking for this change for preview images.

wgranger:
Ok, subject previews are now links. I had to rewrite the `discussionLink` function a bit in the TalkDiscussionPreview component, but I think it should work.

eatyourgreens:
Looks good to me but I'm getting `Cannot read property 'removeEventListener' of undefined` errors for the Talk comment box with this branch. eg. go to https://fix-3352.pfe-preview.zooniverse.org/projects/panthera-research/camera-catalogue/talk/445/150224?env=production and click on any link at the top to go to another page.

I can't reproduce this error on master, but I can't immediately see what changes on this branch might have caused this.

eatyourgreens:
My bad, I think that error was caused by me forgetting that I was testing React 15.4 locally. This works fine with a fresh `npm install`.

PR #3374, opened 2017-01-09
Freehand line

rogerhutchings:
Creates a freehand line and shape tool, for use in etch-a-cell and floating forests reboot.

I'm not sure how to set this as an experimental tool - I had a look, and couldn't find anything that looked like it defined a list of regular / experimental tools. Any help gratefully received.

cc @trouille @hrspiers 

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://freehand-line.pfe-preview.zooniverse.org/

## Optional


wgranger:
If I remember correctly, you'll have to add this as an experimental feature in `app/pages/admin/project-status/experimental-features.jsx` and then toggle it on for the particular project in the "set project status" portion of the admin page. The item will then be saved as an active experimental feature under the project resource (`project.experimental_tools`). 

These tools will also need to be added to the dev classifier in `mock-data.coffee`.

Seems like it might be good to be able to drag these items, but I'm not sure what the use case is.

app/classifier/drawing-tools/freehand-shape.cjsx
```diff
@@ -0,0 +1,69 @@
+React = require 'react'
+ReactDOM = require 'react-dom'
```
wgranger: Any reason `ReactDOM` is included here? I noticed it isn't in the other component.

rogerhutchings:
- I've added [svg-path-properties](https://github.com/rveciana/svg-path-properties), which provides a JS implementation of `getTotalPath()`. 
- I thought about dragging when I wrote the tool, but I'm not really sure how relevant it is; if you moved the path to correct a duff bit, you'd move the rest of it out of alignment ¯\_(ツ)_/¯
- I've added the tools to the experimental toolset (thanks @wgranger), and staged a new version at https://freehand-line.pfe-preview.zooniverse.org/

shaunanoordin:
## PR Review
* PR properly adds Freehand Line and Freehand Shape tools to the classifier interface, as seen on https://freehand-line.pfe-preview.zooniverse.org/dev/classifier
* Tool allows Lines and Shapes to be added to the Classification; selection/deletion also work correctly. Lines and Shapes stay stable throughout rotation, translation and zoom.
* There is however one possible instance of **unexpected behaviour**, detailed in the next section.
* Tested on OSX+Chrome55/FF50/Safari9, Win7+IE11 (yup, consistent across all browsers including indernerd exploraaaa) on staging branch.

## Unexpected Behaviour
Issue: Freehand Line invisibly blocks actions in its "shape area".

Analysis:
* For example, let's say we create a Freehand Line around this rhino here:
screen shot 2017-01-10 at 17 50 25
* If we try to add any further markers on that rhino (whether another Freehand Line, or a Point, etc), our attempt will fail. (Try adding a Point on the rhino's torso, for example.)
* This is because, SVG-element-wise, that Freehand Line is functionally equivalent to this Freehand Shape:
screen shot 2017-01-10 at 17 58 53
* The two 'ends' of the Freehand Line invisibly join to create a shape that blocks any further drawing input.
* This is technically 'correct' - in the sense that a Line in this case is just a polygon with no fill, and the blocking behaviour (click to select) is consistent with all our other polygonal markers - but it'll be super confusing to users.

## Status
I have no recommendations for the Freehand Line at the moment, but I'll look into the SVG structure to see if there any good ways to make this work for the user.

wgranger:
I love your appreciation for keeping things alphabetical. This might be a bit pedantic, but it would also be nice to have an icon to display in the dev classifier for both freehand tools. See `app/classifier/tasks/drawing/icons.cjsx`.

rogerhutchings:
Talking to @shaunanoordin last night, I had an idea for making the freehand line UI a bit easier. Currently, you need a pixel-perfect click on the line to select it (if you want to delete it, for example) - putting another line underneath with a wider transparent stroke should make it a bit more forgiving.

rogerhutchings:
The clicks being caught outside of the line are caused by a bit of unexpected SVG behaviour: paths are _implicitly filled_, making them act like a closed shape. This can be fixed by adding `fill="none"` to the path.

I've also added the invisible stroke to make paths easier to select.

shaunanoordin:
## PR Review (Update)
* Confirmed: Adding `fill="none"` to the Freehand Lines removes the 'blocked clicks' unexpected behaviour.
* Tested on OSX+Chrome55/FF50/Safari9, Win7+IE11

## Status
LGTM! Functionality-wise, we're good. Looks like the only thing needed for a merge are those icons.

rogerhutchings:
added the icons

PR #3377, opened 2017-01-09
Change Background Button

### LINKED ISSUE #3187: Add Button on New Homepage to Change Background
## Expected behavior
A user should be able to change the background of her or his homepage _from the landing page_ with the click of a button. This is similar to how some social media sites allow you to change the cover photo.

## Current behavior
The default background is a solid green color, and the user might not be aware that (s)he can change the wallpaper.

## Additional information
A similar component currently exists in the edit profile section.

wgranger:
Closes #3187 

Describe your changes.
This PR adds a button on a user's homepage to change a profile background. Most of the line changes on this PR is rewriting the user home page into our standard ES6 and abstracting media actions into their own file to make them flexible. There's also some work done to make the button more accessible with focus, which was being difficult with simple `&:focus` styling.

# Review Checklist

- [ ] Does it work on mobile?

https://edit-background.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/home-for-user/index.jsx
+++ b/app/pages/home-for-user/index.jsx
@@ -22,26 +22,11 @@ const SECTIONS = {
   builds: MyBuildsSection
 };
 
-const HomePageForUser = React.createClass({
-  propTypes: {
-    user: React.PropTypes.object.isRequired,
-    location: React.PropTypes.object
-  },
+export default class HomePageForUser extends React.Component {
+  constructor(props) {
+    super(props);
 
-  contextTypes: {
-    setAppHeaderVariant: React.PropTypes.func
-  },
-
-  getDefaultProps() {
-    return {
-      actions: mediaActions,
-      metadata: {},
-      user: {}
-    };
-  },
-
-  getInitialState() {
-    return {
+    this.state = {
       backgroundSrc: '',
       avatarSrc: '',
       showNews: false,
@@ -52,28 +37,32 @@ const HomePageForUser = React.createClass({
       selectedProjectID: null,
       openSection: null
     };
-  },
+
+    this.getRibbonData = this.getRibbonData.bind(this);
+    this.toggleNews = this.toggleNews.bind(this);
+    this.updateBackground = this.updateBackground.bind(this);
+  }
 
   componentDidMount() {
     this.context.setAppHeaderVariant('detached');
     addEventListener('hashChange', this.handleHashChange);
     this.fetchRibbonData(this.props.user);
-  },
+  }
 
   componentWillReceiveProps(nextProps) {
     if (nextProps.user !== this.props.user) {
       this.fetchRibbonData(nextProps.user);
     }
-  },
+  }
 
   componentWillUnmount() {
     this.context.setAppHeaderVariant(null);
     removeEventListener('hashChange', this.handleHashChange);
-  },
+  }
 
   handleHashChange() {
     this.forceUpdate();
-  },
+  }
 
   fetchRibbonData(user) {
     if (!user) {
@@ -120,7 +109,7 @@ const HomePageForUser = React.createClass({
         loading: false
       });
     });
-  },
+  }
 
   getRibbonData(user, _page = 1) {
     const getRibbonData = this.getRibbonData;
@@ -173,7 +162,7 @@ const HomePageForUser = React.createClass({
         }
       }
     });
-  },
+  }
 
   getProjectsForPreferences(preferences) {
     const projectIDs = preferences.map((projectPreference) => {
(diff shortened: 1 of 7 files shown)


app/pages/home-for-user/index.jsx
```diff
@@ -1,75 +1,76 @@
...
+    this.toggleFocus = this.toggleFocus.bind(this);
+    this.updateBackground = this.updateBackground.bind(this);
+    this.createLinkedResource = this.props.actions.createLinkedResource.bind(this);
+    this.uploadMedia = this.props.actions.uploadMedia.bind(this);
+  }
 
   componentDidMount() {
+    const backgroundBtn = document.getElementById('change-background');
+    backgroundBtn.addEventListener('focus', this.toggleFocus, true);
+    backgroundBtn.addEventListener('blur', this.toggleFocus, true);
```
srallen: Is there a reason why these event listeners are being added this way instead of directly on the element? If so, then I'd prefer using a ref to be consistent with using the virtual DOM instead of directly manipulating the DOM.

wgranger: Gotcha. I can try using a ref. I'm not too familiar with refs, so this will be a good opportunity to work with them and learn something.

srallen: Or just add `onFocus` and `onBlur` directly to the element. React supports both.

css/home-page-for-user.styl
```diff
@@ -55,6 +55,31 @@ GRAY = #EFF2F5
     padding: 1em 2em
     text-align: left
 
+  &__change-background
+    border: 1px solid white
+    border-radius: 0.8em
+    cursor: pointer
+    font-weight: bold
+    padding: 0.8em
+    position: relative
```
srallen: This should be `position: absolute` and `top` and `left` instead. Right now, the button will move around depending on the screen size and can potentially overlap with other elements in the nav bar. Setting the margin is unnecessary with absolute positioning.

app/pages/home-for-user/index.jsx
```diff
@@ -202,7 +212,16 @@ const HomePageForUser = React.createClass({
             <div>{this.state.error.toString()}</div>
           )}
 
-          {!!hashQuery.project ? (
+          {screen.width > 700 && (
```
srallen: Is the intention to hide the button if the window is less than 700px? If so, then use `window.innerWidth` instead of screen. Right now, if I drag my browser to be smaller than 700px, the button is still visible.

app/pages/home-for-user/index.jsx
srallen: I just noticed that the only thing happening is a style is being set by these handlers. This can be done entirely with CSS instead, actually. `:focus` is a css pseudo-class you can target. Then, adding an event listener to `blur` would be unnecessary as well as tracking it in the component state.

wgranger: I wasn't able to get the `:focus` pseudo-class to work with this. I'm not sure if it's because of the input? However, I also noticed that other media drop buttons on PFE, with the `standard-button` class were not receiving focus although they were defined to have such in CSS.

srallen: The styles work. Adding `&:hover, &:focus` to line 70 of the style file, then checking the label element using dev tools by right clicking and selecting `:focus`. What isn't working is the tab focus using keyboard. Adding `tabIndex="0"` to the element will make it work, but I'm not sure if that's the best approach. @eatyourgreens is there a better way?

wgranger: Ahh. Good call. Just pushed up a commit using `tabIndex`. Restaged as well.

eatyourgreens: `tabIndex` isn't needed on an input and shouldn't be used on a label (clicking on a label should focus the associated input). What you're trying to do here is style the adjacent text when the input has focus. One way to do this with just CSS is by wrapping the label text in a span and using sibling selectors `input:focus + span`.

wgranger: That makes sense, but putting a `span` adjacent to the `input` tag will reveal the default "Choose file" of the `input` tag. I haven't found another way to hide this wording yet.

wgranger:
I'm still open to styling suggestions on that `input`. For now, I reverted back to the `onBlur` and `onFocus` events. Also rebased, which was a bit tricky.

srallen:
@wgranger with @eatyourgreens suggestion, you can perhaps use `visibility: hidden;` on the input's css? Just hide it?

eatyourgreens:
Hiding an input with `visibility: hidden` or `display: none` removes it from the UI completely, so you can't tab to it with the keyboard. Hiding it offscreen with absolute positioning might work. OWD has some examples of hidden radio buttons, using a similar technique, for the categories down the lefthand side of the classifier at https://www.operationwardiary.org/#/classify. That might be helpful, but file inputs are a pain to style compared to radio buttons.

eatyourgreens:
@wgranger the `onFocus` and `onBlur` listeners should be attached to the file input, and the label element should not be in the documents tab order. At the moment, you can tab to the label, but you really want to tab to the file input, which will then open a file dialog when you press enter/space.

PR #3389, opened 2017-01-12
Make sure freehand tools only appear if enabled

rogerhutchings:
Ugh, my bad - thanks for catching that @wgranger 

Freehand tools should now only appear when enabled.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-freehand-tools.pfe-preview.zooniverse.org/

## Optional


eatyourgreens:
Eww, that is an ugly hack. My suprised you missed this.

eatyourgreens:
Not surprised, iOS, not surprised.

PR #3398, opened 2017-01-16
Infinite loop

wgranger:
Describe your changes.
This PR adds an infinite loop to the subject flipbook so a user doesn't have to keep pressing "play" after `playIterations` has run its course. This was deceptively easy to add, so I'm curious if I'm missing something with the flipbook functionality. Seems to work well though.

I also enlarged the play/pause icon size based on feedback.

# Review Checklist

- [x] Does it work on mobile?

https://infinite-loop.pfe-preview.zooniverse.org/projects/wgranger-test/single-task-bug
* The "Infinite Loop" workflow on this project has a flipbook. 

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -161,6 +161,7 @@ Classifier = React.createClass
           allowFlipbook={workflowAllowsFlipbook @props.workflow}
           allowSeparateFrames={workflowAllowsSeparateFrames @props.workflow}
           onChange={@handleAnnotationChange.bind this, currentClassification}
+          playIterations={Infinity if @props.workflow?.configuration.infiniteLoop}
         />
 
         <div className="task-area">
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Would it be worth making `playIterations` a workflow setting too, with a default of 3 for backward compatibility? I think that would also be a small change but feel free to ignore it if it turns out to be more complicated.

wgranger:
Are you thinking of making `playIterations` a number somebody sets in the project builder, with an option for infinite? That makes sense.

eatyourgreens:
Yeah, maybe a number you can input, then set it to `Infinite` if the input is blank. Default setting would have to be 3, I guess, since that's the current hardcoded value.

wgranger:
Are you thinking of keeping this on the `configuration` tag, or making it a separate/new tag on the workflow resource?

eatyourgreens:
`workflow.configuration` sounds good. Won't the Panoptes API throw 422 errors if you try to add arbitrary new properties to a Workflow resource?

wgranger:
It will. Wasn't sure if we were going to add another tag on the backend, but I was worried that might bloat the resource. I think keeping it on `configuration` makes sense. I'll push up a commit with those changes above soon.

eatyourgreens:
Looks good to me. Maybe set a placeholder to indicate that blank=Infinity, like the min/max controls do for the drawing tools. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/min-max-editor.jsx#L87

PR #3408, opened 2017-01-17
Fix Validation Errors

### LINKED ISSUE #1954: Not entering subject in 'send a message' hangs the interface (minor bug)
I went to the 'send a message' screen and wrote a message. I selected a recipient, but forgot to type anything in the subject box. When I clicked 'Send Message', I got an error message saying that the subject couldn't be blank. However, the buttons ('Linked Image' and 'Send Message') remained non-active and the "working" circle icon continued to be present even after I entered a subject. So I had to reload the form and enter everything from scratch (remembering to add a subject this time).

wgranger:
Fixes #1954

Describe your changes.
Validations weren't working correctly with the inbox form. Sometimes you'd get an error from the API, other times from the corresponding validation files. Often times you'd hit an error that would freeze the input form.

Also, the API required a subject field to be at least 3 characters, which is why that validation was added on the frontend.

# Review Checklist

- [x] Does it work on mobile?

https://empty-message-bug.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/talk/inbox-form.cjsx
+++ b/app/talk/inbox-form.cjsx
@@ -20,9 +20,11 @@ module.exports = React.createClass
   getInitialState: ->
     validationErrors: []
 
-  validations: (message, subject, recipients) ->
-    userErrors = if (not recipients.length) then ['Messages must have a recipient'] else []
-    subjectErrors = getErrors(subject, subjectValidations)
+  validations: (message) ->
+    recipient_ids = ReactDOM.findDOMNode(@).querySelector('[name="userids"]').value
+
+    userErrors = if (not recipient_ids.length) then ['Messages must have a recipient'] else []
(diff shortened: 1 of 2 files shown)


app/talk/inbox-form.cjsx
```diff
@@ -20,9 +20,11 @@ module.exports = React.createClass
   getInitialState: ->
     validationErrors: []
 
-  validations: (message, subject, recipients) ->
-    userErrors = if (not recipients.length) then ['Messages must have a recipient'] else []
-    subjectErrors = getErrors(subject, subjectValidations)
+  validations: (message) ->
+    recipient_ids = ReactDOM.findDOMNode(@).querySelector('[name="userids"]').value
```
eatyourgreens: See #3409 I'm updating `UserSearch` to get rid of these DOM queries. They aren't guaranteed to work with future versions of `react-select`.

PR #3409, opened 2017-01-18
Remove DOM queries from user search

eatyourgreens:
Adds methods to get the selected user(s) from `UserSearch` and clear the select control.

Removes the DOM queries that got the values directly from hidden form inputs generated by `react-select`.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://user-search.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -51,7 +51,7 @@ RoleCreator = React.createClass
         <p className="form-help error">{errorMessage}</p>}
       <form style={style}>
         <div>
-          <UserSearch />
+          <UserSearch ref={(component) => @userSearch = component} />
         </div>
 
         <table className="standard-table">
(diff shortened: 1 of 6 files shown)


eatyourgreens:
Another way to approach this might be to remove the `setState` calls from `UserSearch` and instead treat it as a [controlled form component](https://facebook.github.io/react/docs/forms.html#controlled-components), with `value` passed in as a prop.

eatyourgreens:
Good question. I searched PFE for components that used `UserSearch`, but I didn't check to see if they're in use themselves.

srallen:
Looking a little closer at it, it appears to be using a couple of helpers in the talk/lib folder, one of which checks to see if you're admin. That's probably not working right or never worked to begin with if none of us noticed it was broken. I'll open a separate issue for it, but we might not even need it at all.

PR #3412, opened 2017-01-18
Track mini-course and tutorial completion

srallen:
This adds geordi to the mini-course and tutorial to track its completion timestamp in geordi for future analysis. It's a bit more invasive than I'd like, but the modal-form doesn't have access to the app context.

@amyrebecca could you look over to let me know if it's setup correctly and if there's any way to check if it's actually logging? Anything to look for in the network tab in dev tools?

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/lib/mini-course.cjsx
+++ b/app/lib/mini-course.cjsx
@@ -109,8 +109,6 @@ module.exports = React.createClass
     optOut: false
 
   componentDidMount: ->
-    @props.geordi.makeHandler 'tutorial-completion'
-
     # If user navigates away, record the next slide to load in prefs
     window.addEventListener 'beforeunload', @handleProjectPreferencesOnUnmount
 
@@ -168,8 +166,9 @@ module.exports = React.createClass
   logToGeordi: (datetime) ->
     @props.geordi.logEvent {
       type: 'mini-course-completion'
-      user: @props.user.id
-      minicourse: @props.minicourse.id
-      completed_at: datetime
+      data: {
+        minicourse: @props.minicourse.id
+        minicourseCompletedAt: datetime
+      }
     }
 
--- a/app/lib/mini-course.cjsx
+++ b/app/lib/mini-course.cjsx
@@ -23,7 +23,7 @@ module.exports = React.createClass
       else
         Promise.resolve()
 
-    start: (minicourse, projectPreferences, user) ->
+    start: (minicourse, projectPreferences, user, geordi) ->
       MiniCourseComponent = this
       if minicourse.steps.length isnt 0
         awaitMiniCourseMedia = minicourse.get 'attached_images'
@@ -37,7 +37,7 @@ module.exports = React.createClass
             mediaByID
 
         awaitMiniCourseMedia.then (mediaByID) =>
-          Dialog.alert(<MiniCourseComponent projectPreferences={projectPreferences} user={user} minicourse={minicourse} media={mediaByID} />, {
+          Dialog.alert(<MiniCourseComponent projectPreferences={projectPreferences} user={user} minicourse={minicourse} media={mediaByID} geordi={geordi} />, {
             className: 'mini-course-dialog',
             required: true,
             closeButton: true,
@@ -46,7 +46,7 @@ module.exports = React.createClass
             .catch =>
               null # We don't really care if the user canceled or completed the tutorial.
 
-    restart: (minicourse, projectPreferences, project, user) ->
+    restart: (minicourse, projectPreferences, project, user, geordi) ->
       resetPreferences = {
         "preferences.minicourses.opt_out.id_#{minicourse.id}": false,
         "preferences.minicourses.slide_to_start.id_#{minicourse.id}": 0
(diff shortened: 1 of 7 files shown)


srallen:
Staged at https://add-tutorial-timestamp-geordi.pfe-preview.zooniverse.org/

srallen:
@wgranger do you know what I can look for to check if the geordi logging is working? Thanks.

wgranger:
You could check the geordi SQL database. I can email you a rundown of how to get connected.

eatyourgreens:
@srallen look for POST requests to `https://geordi.zooniverse.org/api/events/` but I'm not sure what the payload of those requests is.

srallen:
I'm not seeing a POST to geordi, so I don't think I have it setup right. Also, staging will POST to production geordi?

eatyourgreens:
Good point. I'm not sure how the client is set up for the different environments.

eatyourgreens:
Looks like it should post to a staging server from dev and staging environments
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/geordi-logger.coffee#L17-L21

wgranger:
Ok. Took me a minute to remind myself how Geordi was working. So, Geordi is still currently only tracking four tokens `['zooHome', 'zooTalk', 'zooniverse/gravity-spy', 'mschwamb/comet-hunters']`, so you aren't going to see anything pop up on the networks tab unless you're in Zoo global, Talk, Gravity Spy, or Comet Hunters. I can see a POST going to Geordi on the staged branch (when I'm logged in, looking at Gravity Spy, and closing the tutorial. Give it a shot, look for `events/` in the network tab, and when you click on that you should see the payload at the bottom of the Headers tab to the right.

app/lib/tutorial.cjsx
```diff
@@ -111,3 +105,29 @@ module.exports = React.createClass
...
+        links:
+          project: project.id
+        preferences: {}
+      # Build this manually. Having an index (even as a strings) keys creates an array.
+      projectPreferences.preferences ?= {}
+      projectPreferences.preferences.tutorials_completed_at ?= {}
+      projectPreferences.update "preferences.tutorials_completed_at.#{@props.tutorial.id}": now
+      projectPreferences.save()
+
+      @logToGeordi now
```
wgranger: Currently, this will only hit `logToGeordi` if the user is logged in, due to the conditional above. Is this intentional that we only want to log if a user is present?

app/lib/mini-course.cjsx
```diff
@@ -153,8 +157,19 @@ module.exports = React.createClass
 
       @props.projectPreferences.update "preferences.minicourses.completed_at.id_#{@props.minicourse.id}": now
       @props.projectPreferences.save()
+
+      @logToGeordi now
```
wgranger: Having a bit of a tougher time testing this on the staged branch because I'm always getting the `else` of this conditional, but it appears to be set up identically to the tutorial instance and should work.

app/lib/tutorial.cjsx
srallen: Yeah, for purposes of behavior study, it wouldn't be useful to know whenever a logged out user views the tutorial. We don't share IP info for privacy reasons, it can't really be linked to classifications for instance. Unless geordi tracks de-identified browser sessions?

app/lib/mini-course.cjsx
srallen: The GS mini-course is fairly long. I'll add my test project to the list of projects setup to track to make sure this works.

srallen:
Thanks for pointing out the filter on projects. That would explain it.

app/lib/mini-course.cjsx
```diff
@@ -107,6 +109,8 @@ module.exports = React.createClass
     optOut: false
 
   componentDidMount: ->
+    @props.geordi.makeHandler 'tutorial-completion'
```
wgranger: I'm a bit hesitant to say this, but I _don't_ think this is doing anything here. If you look at other instances of PFE where we use `geordi.makeHandler`, it is usually attached to a component and later called/bound with a click handler. Perhaps @eatyourgreens can confirm this.

app/lib/tutorial.cjsx
```diff
@@ -53,55 +53,49 @@ module.exports = React.createClass
...
-    steps: []
+    geordi: {}
     media: {}
+    preferences: null
+    tutorial: {}
+    user: null
 
-  render: ->
+  componentDidMount: ->
+    @props.geordi.makeHandler 'tutorial-completion'
```
wgranger: Same note as above. I'm not positive this is doing anything unless it's used as a click handler. Since you're passing geordi straight in as a prop and doing your logging later, I think this can be removed. I'd double check though.

app/lib/tutorial.cjsx
wgranger: Makes total sense from a privacy standpoint.

app/lib/mini-course.cjsx
eatyourgreens: From memory, `geordi.makeHandler` returns a function so you probably do need to assign it to an event listener somewhere, in order for it to be called.

app/lib/tutorial.cjsx
eatyourgreens: Yep, this returns a function that logs `tutorial-completion` to Geordi when it's called, but I don't think this code ever calls that function.

You can probably delete this from the tutorial and the mini-course.

srallen: Ok good to know. I didn't totally understand what it was doing and adding just in case it was needed for the POST.

srallen:
Ok, thanks for the pointers. I've removed the unnecessary handlers and fixed the data format for the log event.

app/lib/mini-course.cjsx
srallen: I checked with my test project and mini-course is logging correctly.

PR #3411, opened 2017-01-18
style changes for text subjects

simensta:
**Describe your changes.**
- Increase the color contrast between the text and the background of text subjects. 
- Break line at newline characters in text subjects

Staged [here](https://text-subject-styling.pfe-preview.zooniverse.org/).

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] Did you deploy a staging branch?
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #3416, opened 2017-01-19
consistent background color for text subjects

simensta:
Describe your changes.
As a follow up to #3411, this PR adds a consistent background color to text subjects. Staged [here](https://text-subject-styling.pfe-preview.zooniverse.org/).

# Review Checklist
- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/components/text-viewer.jsx
+++ b/app/components/text-viewer.jsx
@@ -37,6 +37,9 @@ class TextViewer extends Component {
 
   render() {
     let { content } = this.state;
+    if (content === "") {
+      return null;
+    }
     if (SUPPORTED_TYPES.indexOf(this.props.type) === -1) {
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Looks good to me. I'm not 100% sure about the rounded corners on subjects. Here's a test which should have one jpeg and one plain text frame: https://text-subject-styling.pfe-preview.zooniverse.org/projects/eatyourgreens/diagnosis-london/classify?env=production
What do you think?

simensta:
Good point on the border-radius property. It affects Diagnosis London subjects differently than the DCW subjects. To keep text subject styling as consistent as possible, I think the border-radius property should be removed. I'll update the PR. Thanks!

PR #3423, opened 2017-01-24
Fix Image Appearance on Tutorials

### LINKED ISSUE #3419: If the tutorial image is smaller than the width of the tutorial box, center the image
screen shot 2017-01-21 at 1 36 59 pm

wgranger:
Fixes #3419.

Describe your changes.
Two Issues here: Smaller images were not being centered on tutorials. Also, larger images were showing outside of their container Div, hence the `max-width` change.

# Review Checklist

- [x] Does it work on mobile?

https://fix-3419.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/css/tutorial.styl
+++ b/css/tutorial.styl
@@ -29,6 +29,7 @@
     border-top-left-radius: 8px
     border-top-right-radius: 8px
     display: block
+    margin: auto
     max-width: 100%
 
 .step-through-direction
--- a/css/tutorial.styl
+++ b/css/tutorial.styl
@@ -25,8 +25,8 @@
     > .media-card-content
       @extends .content-container
 
-    img
-      max-width: 100%
+      img
+        max-width: 100%
 
   .media-card-media
     border-top-left-radius: 8px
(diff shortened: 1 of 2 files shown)


css/tutorial.styl
```diff
@@ -25,10 +25,14 @@
     > .media-card-content
       @extends .content-container
 
+    img
+      max-width: 100%
```
srallen: This isn't necessary. The image is already getting `max-width: 100%` from the `media-card-media` class that is applied to the `` tag.

wgranger: I noticed that `media-card-media` already had that attribute, but I was noticing the error on some tutorials, particularly the one for Amazon Aerobotany. This CSS addition appeared to fix the issue.

srallen: Ok, so this is targeting images in the markdown instead of uploaded image media linked to the tutorial step (the `media-card-media`). The tutorial for Amazon Aerobotany is using images its markdown content, which you're right, should have a max-width applied too. The `img` styling here should be indented to be a child of `media-card-content` so that it's clear that the style is to target images in the markdown.

wgranger: Gotcha. Thanks for the explanation.

PR #3427, opened 2017-01-25
Move discussion reply box (and convert to ES6)

### LINKED ISSUE #2973: Resizing the notes and message board thread Talk  comment box so that it fits right below the previous comments
A version of this ticket was closed previously  with the response that a Talk design pass was being made and would address this issue. That was a while ago, so I've created a new ticket. 

example: hit reply in a [thread](https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/29/86792) like this

Full page capture when you hit reply and there are not enough comments to fill all the way down to the bottom edge of the left side bar:

fireshot capture 25 - planet four_ t__ - https___www zooniverse org_project

What you see in your visible browser when you hit reply to the comment:

screen shot 2016-09-05 at 4 14 53 pm

same thing for: https://www.zooniverse.org/projects/mschwamb/planet-four-terrains/talk/10/85980?page=1&scrollToLastComment=true

screen shot 2016-09-05 at 4 25 28 pm

If there is any developer time in the future to resize the comment box so it fits in the space below the already published comments, I think it would be very welcomed by the Talk community. Replying on Talk thread is a commonly used feature, so it would be nice if one could see the full discussion or subject image rather than having this huge gap

This behavior happens in all browsers I tested (Chrome, fixfox, and safari for mac) 

cc @parrish

eatyourgreens:
Fixes #2973 .

Describe your changes.
Rewrites the discussion thread comment box in ES6 and moves it from the very bottom of the page to the bottom of the list of comments.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-2973.pfe-preview.zooniverse.org/?env=production

## Optional

### DIFF
--- a/app/talk/discussion-comment.jsx
+++ b/app/talk/discussion-comment.jsx
@@ -1,0 +1,103 @@
+import React from 'react';
+import { Link } from 'react-router';
+import talkClient from 'panoptes-client/lib/talk-client';
+import Avatar from '../partials/avatar';
+import DisplayRoles from './lib/display-roles';
+import CommentBox from './comment-box';
+import SignInPrompt from '../partials/sign-in-prompt';
+import commentValidations from './lib/comment-validations';
+import alert from '../lib/alert';
+import { getErrors } from './lib/validations';
+
+class DiscussionComment extends React.Component {
+  constructor(props) {
+    super(props);
+    this.onSubmitComment = this.onSubmitComment.bind(this);
+    this.commentValidations = this.commentValidations.bind(this);
+    this.state = {
+      commentValidationErrors: []
+    };
+  }
+
+  onSubmitComment(e, textContent, subject, reply) {
+    let comment = {
+      user_id: this.props.user.id,
+      discussion_id: +this.props.params.discussion,
+      body: textContent
+    };
+    if (!!subject && subject.id) {
+      comment = Object.assign(comment, {
+        focus_id: +subject.id,
+        focus_type: 'Subject'
+      });
+    }
+    if (reply) {
+      comment = Object.assign(comment, {
+        reply_id: reply.comment.id
+      });
+    }
+
+    return talkClient.type('comments')
+    .create(comment)
+    .save()
+    .then(this.props.onSubmitComment);
+  }
+
+  commentValidations(commentBody) {
+    // TODO: return true if any additional validations fail
+    const commentValidationErrors = getErrors(commentBody, commentValidations);
+    this.setState({ commentValidationErrors });
+    return !!commentValidationErrors.length;
+  }
+
+  promptToSignIn() {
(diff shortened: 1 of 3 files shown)


app/talk/discussion-comment.jsx
```diff
@@ -0,0 +1,99 @@
...
+            reply={this.props.reply}
+            logSubmit={true}
+            onClickClearReply={this.props.onClearReply}
+            header={null}
+          />
+        </section>
+      );
+    } else {
+      return (
+        <p>Please <button className="link-style" type="button" onClick={this.promptToSignIn}>sign in</button> to contribute to the discussion</p>
```
eatyourgreens: `this.promptToSignIn` is undefined here.

app/talk/discussion-comment.jsx
```diff
@@ -0,0 +1,99 @@
...
+  commentValidations(commentBody) {
+    // TODO: return true if any additional validations fail
+    const commentValidationErrors = getErrors(commentBody, commentValidations);
+    this.setState({ commentValidationErrors });
+    return !!commentValidationErrors.length;
+  }
+
+  render() {
+    if (!!this.props.discussion && this.props.discussion.locked) {
+      return this.lockedMessage();
```
eatyourgreens: `this.lockedMessage` is undefined here.

This if block should probably be moved up into the parent component.

eatyourgreens:
Doesn't work for locked threads or anonymous users yet.

app/talk/discussion.cjsx
```diff
@@ -49,6 +46,18 @@ module.exports = React.createClass
     if @props.params.discussion isnt nextProps.params.discussion
       @setDiscussion(nextProps.params.discussion)
         .then => @setComments(nextProps.location.query.page ? 1)
+        .then =>
+          if nextProps.user isnt @props.user
+            talkClient
```
srallen: I'm not really familiar how the talkClient handles errors, should there be a catch on this?

app/talk/discussion.cjsx
```diff
@@ -69,18 +78,18 @@ module.exports = React.createClass
             pathname: @props.location.pathname
             query: @props.location.query
 
+        if @props.user?
+          talkClient
```
srallen: Same question about error catching

app/talk/discussion.cjsx
eatyourgreens: I'm not sure. I copied this straight from the coffeescript without really thinking about it.

eatyourgreens: Probably should be catching errors, though. If the roles call fails, the only side effect I can think of is that you don't get the banner under your username beside the comment box.

PR #3432, opened 2017-01-25
Add Back Project Specific Notifications Page

wgranger:
Describe your changes.
This adds back the route to a project specific notifications page based on user feedback. The project specific page does two things: 

- Double checks that the project notifications are present, if available.

- Automatically opens the section associated with that project.

# Review Checklist

- [x] Does it work on mobile?

https://open-section.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/notifications.jsx
+++ b/app/pages/notifications.jsx
@@ -100,7 +100,7 @@ export default class NotificationsPage extends React.Component {
       );
     } else if (this.state.projNotifications.length === 0) {
       notificationView = (
-        <div className="centering talk-module">
+        <div className="centering talk-module notifications-title">
           <Translate content="notifications.noNotifications" />
           <Translate content="notifications.participation" />
         </div>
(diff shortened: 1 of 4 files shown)


wgranger:
@eatyourgreens Do you have any suggestions on this? I'd like to test functionality when `getProjectNotifications` is called with a project prop, but I'm having a trouble mocking the response from `talkClient`. I've looked into spies and stubs, but I can't get the function to set state on line 46 without the API call hitting the `then`.

eatyourgreens:
@wgranger depends what you're trying to test, but if your test is `when a project is selected, the corresponding section should be open` then something like
```
wrapper.setState({expanded: 'project-1234'});
wrapper.update();
```
then test that there are 3 sections listed, that project 1234 is open and that the others are closed.

wgranger:
Ok, yeah I was trying to test the component would set the state correctly in `getProjectNotifications`, but perhaps that is a bit overkill. I'll just look to set the states manually in the tests.

PR #3319, opened 2016-12-09
[WIP] Classifier ES6 rewrite

CKrawczyk:
This PR rewrites the classifier interface into ES6 and add tests for each component.  Here is a list of what has been done and what still needs to be done:
- [x] Convert `tutorial-button`
- [x] Convert `mini-course-button`
    - [x] extract common elements of `tutorial-button` and `mini-course-button` to higher order component (`restart-button`)
    - [x] write tests for `tutorial-button` and `mini-course-button`
- [x] Convert `warning-banner`
    - [x] write tests for `warning-banner`
- [x] Convert `classification-summary`
    - [x] write tests for `classification-summary`
    - [x] figure out how to get tests that require `.cjsx` to run (I think we will need `coffee:coffee-react/register`, but that still did not work)
- [x] Convert `world-wide-telescope` ([PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3463) is up)
    - [x] write tests for `world-wide-telescope`
- [x] Convert `frame-annotator`
    - [x] write tests for `frame-annotator`
- [x] Convert `metadata-based-feedback` (Converted and tested, but unmerged since functionality changes are needed for Space Warps use)
    - [x] write tests for `metadata-based-feedback`
- [ ] Convert `index`
    - [x] Split out independent components to different files (`render-summary`, `render-expert-options`, `classifier-helpers`, `index`, and `classifier`).
    - [ ] write tests for all these components

For the most part the classifier works, but I have yet to test any project using the gravity-spy or world-wide-telescope flags or subjects with expert feedback.

One bug that still needs to be addressed is the `gold standard` checkbox of the `render-expert-options` popover is not causing a re-render of the component.  This only happens when it is clicked from the `classification-summary` page, the same exact component works just fine on the `classification` page.

As there are many components that are displayed only if particular flags are set, I have written a `VisibilityWrapper` that takes the prop `visible` and displays its children only if it is truthy.  This avoids the need to place a bunch of `if` statements in the `jsx`.

Other things that would be nice, but might be better for a later PR:
- [ ] Convert and write tests for the `task` folder
- [ ] Convert and write tests for the `drawing-tools` folder

I will be on vacation for the next month, if anyone wants to help with this feel free to commit directly to the branch.

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
[...]

PR #3434, opened 2017-01-25
Add first-to-classify subject split testing

parrish:
Adds a 7-10 message for the first user to classify a subject.

Staged at https://first-to-classify.pfe-preview.zooniverse.org/projects/parrish/a-project/classify

We should also take care that this gets picked up in #3319

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -101,11 +101,23 @@ Classifier = React.createClass
     if @props.project.experimental_tools?.indexOf('expert comparison summary') > -1
       @getExpertClassification @props.workflow, @props.subject
 
+    @loadClassificationsCount subject
+
     preloadSubject subject
       .then =>
         if @props.subject is subject # The subject could have changed while we were loading.
           @setState subjectLoading: false
           @props.onLoad?()
+
+  loadClassificationsCount: (subject) ->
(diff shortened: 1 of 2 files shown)


app/classifier/index.cjsx
```diff
@@ -101,12 +101,24 @@ Classifier = React.createClass
...
+    @loadClassificationsCount subject
+
     preloadSubject subject
       .then =>
         if @props.subject is subject # The subject could have changed while we were loading.
           @setState subjectLoading: false
           @props.onLoad?()
 
+  loadClassificationsCount: (subject) ->
+    if @props.splits?['subject.first-to-classify']
```
parrish: Worth noting that this API call never fires unless we're actually testing this

app/classifier/index.cjsx
```diff
@@ -101,12 +101,24 @@ Classifier = React.createClass
...
+      query =
+        workflow_id: @props.workflow.id
+        subject_id: subject.id
+
+      apiClient.type('subject_workflow_statuses').get(query).then ([sws]) =>
+        classificationCount = sws?.classifications_count or 0
```
parrish: A SubjectWorkflowStatus _shouldn't_ exist without classifications, but just being cautious.

PR #3437, opened 2017-01-26
Add segmented freehand tools, refactor existing freehand tools to return points array

rogerhutchings:
Adds multi-step freehand drawing tools for larger freehand shapes, and refactors the existing ones to return an array of coordinates instead of a path string to make them consistent with the existing drawing tools.

The tools can be tested out at https://freehand-segment.pfe-preview.zooniverse.org/projects/rogerhutchings/etch-a-cell

cc @hrspiers 

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://freehand-segment.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/drawing-tools/freehand-helpers.coffee
+++ b/app/classifier/drawing-tools/freehand-helpers.coffee
@@ -1,0 +1,8 @@
+createPathFromCoords = (coordsArray) ->
+  [firstCoord, otherCoords...] = coordsArray
+  path = "M #{firstCoord.x},#{firstCoord.y} "
+  path += "L #{x},#{y} " for {x, y} in otherCoords
+  path
+
+module.exports = 
+  createPathFromCoords: createPathFromCoords
--- a/app/classifier/drawing-tools/freehand-line.cjsx
+++ b/app/classifier/drawing-tools/freehand-line.cjsx
@@ -3,66 +3,71 @@ DrawingToolRoot = require './root'
 deleteIfOutOfBounds = require './delete-if-out-of-bounds'
 DeleteButton = require './delete-button'
 {svgPathProperties} = require 'svg-path-properties'
+{createPathFromCoords} = require './freehand-helpers'
 
-MINIMUM_LENGTH = 20
-GRAB_STROKE_WIDTH = 6
 BUFFER = 16
 DELETE_BUTTON_WIDTH = 8
-
-getPathNode = (pathString) ->
-  document.querySelectorAll("path[d=\"#{pathString}\"]")[0]
+GRAB_STROKE_WIDTH = 6
+MINIMUM_LENGTH = 20
+SELECTED_STROKE_WIDTH = 6
+STROKE_WIDTH = 
 
 module.exports = React.createClass
   displayName: 'FreehandLineTool'
 
   statics:
-    initStart: ({x, y}, mark) ->
-      _inProgress: true
-      path: "M #{x},#{y}"
-
-    initMove: ({x, y}, mark) ->
-      path: mark.path + " L #{x},#{y}"
-
-    initRelease: (coords, mark) ->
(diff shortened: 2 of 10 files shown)


app/classifier/drawing-tools/freehand-helpers.coffee
```diff
@@ -0,0 +1,8 @@
+createPathFromCoords = (coordsArray) ->
+  [firstCoord, otherCoords...] = coordsArray
+  path = "M #{firstCoord.x},#{firstCoord.y} "
+  path += "L #{x},#{y} " for {x, y} in otherCoords
```
shaunanoordin: This looks legit, but I'm wondering if the data ever needs to differentiate between an open path or a closed path. i.e. M 0 0 L 1 0 L 0 1 Z instead of M 0 0 L 1 0 L 0 1

(Edit: I have no idea how to work these GitHub reviews properly; let's go old school comments for a while.)

shaunanoordin:
## Quick review
* Auto-segmentation corrector: No idea how to test this?
* Freehand Shape: OK 👍 
* Freehand Line: The 'underlying fat path' is missing `fill=none`, causing the shape to invisibly block clicks within its shape.
* Freehand Segment Line -> "Freehand Segment Line": OK 👌 
* Freehand Segment Line -> "Tool Name": presumably Freehand Line Shape? If so, works OK, although suggestion: either give it a coloured fill to indicate you can't click within the shape, or add fill=none to allow clicking within the shape. (Not sure which functionality you'd want, but either or.)
* Freehand Segment Shape: OK ✋ 
* Testing on localhost with OSX+Chrome55

shaunanoordin:
## Quick review (update)
* Minor UI: When tracing marks using Freehand Shape, Freehand Line, Freehand Segment Line or Freehand Segment Shape, tracing slowly causes the mouse cursor to rapidly switch between the crosshair and pointer (pointing hand) shapes.
  * This is because the pointer is rapidly switching between the SVG "canvas" (which has CSS `cursor: crosshair`) and the drawn line/shape (CSS `cursor: pointer`). In one microsecond the pointer is on canvas, then a line-point is created directly beneath the pointer, then in the next microsecond the pointer has moved one pixel away back onto the canvas, etc.
* Minor UX: Clicking "Hide previous marks" does not account for marks made _after_ it's activated. Need to confirm this is intended behaviour.
  * For example, I create Shape A and Shape B, then click Hide Previous Marks. A and B both disappear and Hide Previous Marks indicates the counter "2", indicating two hidden marks.
  * Then, with Hide Previous Marks still active, I create Shape C and Shape D.
  * Now both Shapes C and D are visible, but Previous Marks now updates its counter to "4".
  * It appears the counter counts the _total_ number of marks, not the number of marks _hidden_ or the number of marks _visible._
* Same results across OSX+FF51/Safari9, Win7+IE11

Status:
* Looks mostly good to go.
* Most of these comments are minor and non-blocking, except for the Freehand Line `fill:none` issue.

app/classifier/drawing-tools/freehand-helpers.coffee
rogerhutchings: When a path is closed by the drawing tools, it adds the first point back to the end of the coords array. Reason being, while we use SVG for paths on PFE, researchers might not, and this is probably the clearest way of showing it's a closed shape.

shaunanoordin:
## PR Review
* As a user, I can use the Freehand Shape/Freehand Line/Freehand Segment Line/Freehand Segment Shape to create shapes & paths on the Classifier.
* Tested on http://localhost:3735/projects/rogerhutchings/etch-a-cell/
* Tested on OSX+Chrome55/FF51/Safari9, Win7+IE11

## Status
All major issues addressed, this is now LGTM.

PR #3439, opened 2017-01-26
Frame Annotator Rewrite into ES6

wgranger:
Describe your changes.
This rewrites the `FrameAnnotator` component into ES6. There were a few bugs that arose while writing this code that are also addressed in this PR. Those include:

- **Crop Tool**: Making the tool visible and scale correctly for drag functionality

- There was a bug in the classifier concerning drawing tools. If you make a mark on a subject with a drawing tool, navigate away from the classifier, and return to the same workflow and subject, the classifier will hang because the `getSizeRect()` is returning null, and the mark you made previously doesn't know the width, height, x, y, etc. of null.

- **Triangle Tool**: The dev classifier used `class` instead of `className` for this icon.

# Review Checklist

- [X] Does it work on mobile?

https://frame-rewrite.pfe-preview.zooniverse.org/

## Optional

A couple warnings persist, but they appear minor.

### DIFF
--- a/app/classifier/frame-annotator.jsx
+++ b/app/classifier/frame-annotator.jsx
@@ -42,7 +42,7 @@ export default class FrameAnnotator extends React.Component {
       bottom += pageYOffset;
       return { left, right, top, bottom, width, height };
     }
-    return null;
+    return { left: 0, right: 0, top: 0, bottom: 0, width: 0, height: 0 };
   }
 
   getScale() {
--- a/app/classifier/frame-annotator.jsx
+++ b/app/classifier/frame-annotator.jsx
@@ -13,6 +13,11 @@ export default class FrameAnnotator extends React.Component {
       alreadySeen: false,
       showWarning: false
     };
+    this.getScreenCurrentTransformationMatrix = this.getScreenCurrentTransformationMatrix.bind(this);
+    this.getEventOffset = this.getEventOffset.bind(this);
+    this.normalizeDifference = this.normalizeDifference.bind(this);
+    this.eventCoordsToSVGCoords = this.eventCoordsToSVGCoords.bind(this);
+    this.handleAnnotationChange = this.handleAnnotationChange.bind(this);
   }
 
   componentWillMount() {
@@ -57,7 +62,7 @@ export default class FrameAnnotator extends React.Component {
   // find the original matrix for the SVG coordinate system
   getMatrixForWindowCoordsToSVGUserSpaceCoords() {
     const transformationContainer = this.refs.transformationContainer;
-    transformationContainer.getScreenCTM().inverse();
+    return transformationContainer.getScreenCTM().inverse();
   }
 
   // get the offset of event coordiantes in terms of the SVG coordinate system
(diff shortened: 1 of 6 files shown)


app/classifier/frame-annotator.jsx
```diff
@@ -0,0 +1,282 @@
...
+        svgStyle.pointerEvents = 'all';
+      }
+    }
+
+    const svgProps = {};
+
+    if (TaskComponent) {
+      BeforeSubject = TaskComponent.BeforeSubject;
+      InsideSubject = TaskComponent.InsideSubject;
+      AfterSubject = TaskComponent.AfterSubject;
```
eatyourgreens: `{BeforeSubject, InsideSubject, AfterSubject} = TaskComponent;`

app/classifier/frame-annotator.jsx
```diff
@@ -0,0 +1,282 @@
...
+
+    Object.keys(tasks).map((task) => {
+      const Component = tasks[task];
+      if (Component.getSVGProps) {
+        const props = Component.getSVGProps(hookProps);
+        if (props) {
+          Object.keys(props).map((key) => {
+            svgProps[key] = props[key];
+          });
+        }
```
eatyourgreens: I'm not 100% sure what this is doing, but I think `Object.assign(svgProps, props)` would work, or maybe even `Object.assign(svgProps, Component.getSVGProps(hookProps));`

app/classifier/frame-annotator.jsx
```diff
@@ -0,0 +1,282 @@
...
+      }
+    });
+
+    return (
+      <div className="frame-annotator">
+        <div className="subject-area">
+
+          {BeforeSubject && (
+            <BeforeSubject {...hookProps} />)}
```
eatyourgreens: `!!BeforeSubject &&`

app/classifier/frame-annotator.jsx
```diff
@@ -0,0 +1,282 @@
...
+          <svg ref="svgSubjectArea" className="subject" style={svgStyle} viewBox={createdViewBox} {...svgProps}>
+            <g ref="transformationContainer" transform={this.props.transform}>
+              <rect ref={(rect) => { this.sizeRect = rect; }} width={this.props.naturalWidth} height={this.props.naturalHeight} fill="rgba(0, 0, 0, 0.01)" fillOpacity="0.01" stroke="none" />
+              {type === 'image' && (
+                <Draggable onDrag={this.props.panByDrag} disabled={this.props.disabled}>
+                  <SVGImage className={this.props.panEnabled ? 'pan-active' : ''} src={src} width={this.props.naturalWidth} height={this.props.naturalHeight} modification={this.props.modification} />
+                </Draggable>
+              )}
+
+              {(InsideSubject && !this.props.panEnabled) && (
```
eatyourgreens: `!!InsideSubject &&`

app/classifier/frame-annotator.jsx
```diff
@@ -0,0 +1,282 @@
...
+            </g>
+          </svg>
+
+          {this.props.children}
+
+          {warningBanner && (
+            warningBanner
+          )}
+
+          {AfterSubject && (
```
eatyourgreens: `!!AfterSubject &&`

app/classifier/frame-annotator.jsx
wgranger: Webpack complains about this syntax and won't rebuild. I've tried `const {BeforeSubject, InsideSubject, AfterSubject} = TaskComponent;`, but this fails because it needs to check that `TaskComponent` has been defined. It doesn't seem to like using destructuring assignment in a conditional.

app/classifier/frame-annotator.jsx
wgranger: Yep, that really helps clean up those few lines.

app/classifier/frame-annotator.jsx
eatyourgreens: Weird. Maybe it has problems with the curly brackets.

eatyourgreens: Aha, the answer is on MDN.

```
The ( .. ) around the assignment statement is required syntax when using object literal destructuring assignment without a declaration.

{a, b} = {a:1, b:2} is not valid stand-alone syntax, as the {a, b} on the left-hand side is considered a block and not an object literal.

However, ({a, b} = {a:1, b:2}) is valid, as is var {a, b} = {a:1, b:2}
```
https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment

wgranger: Oh, interesting. Thanks for finding this.

PR #3446, opened 2017-02-01
Fix another 'cannot read property split of undefined' bug

eatyourgreens:
Fixes search on top-level Zoo Talk, where project slugs are undefined.

### DIFF
--- a/app/talk/discussion-preview.cjsx
+++ b/app/talk/discussion-preview.cjsx
@@ -5,7 +5,8 @@ LatestCommentLink = require './latest-comment-link'
 apiClient = require 'panoptes-client/lib/api-client'
 getSubjectLocation = require '../lib/get-subject-location'
 
-`import Thumbnail from '../components/thumbnail';`
+# `import Thumbnail from '../components/thumbnail';`
+Thumbnail = require('../components/thumbnail').default
 
 module.exports = React.createClass
   displayName: 'TalkDiscussionPreview'
--- a/app/talk/discussion-preview.cjsx
+++ b/app/talk/discussion-preview.cjsx
@@ -16,6 +16,9 @@ module.exports = React.createClass
 
   contextTypes:
     geordi: React.PropTypes.object
+  
+  getDefaultProps: ->
+    project: {}
 
   getInitialState: ->
     subject: null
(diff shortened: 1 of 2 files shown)


eatyourgreens:
https://www.zooniverse.org/talk/search/?query=dailyzoo Is broken at the moment. I think this will fix it.

wgranger:
This change _does_ work for the search query because in that instance the parent component, `search-result`, sends in a default "project" prop of an empty object through the spread operator. However, this will break when the parent is something like `TalkBoard` because there is no default project prop available for `TalkDiscussionPreview` to receive, and it will choke when it looks for `.slug` on the project prop that doesn't exist. 

Setting an empty object as the default "project" prop in `TalkDiscussionPreview` should fix this.

app/talk/discussion-preview.cjsx
```diff
@@ -43,7 +43,7 @@ module.exports = React.createClass
       {owner, name} = @props.params
       "/projects/#{owner}/#{name}/talk/#{discussion.board_id}/#{discussion.id}"
 
-    else if @props.project # otherwise fetch from project
+    else if @props.project.slug # otherwise fetch from project
```
wgranger: This component needs a default project prop since one will not always be passed down from the parent component.

eatyourgreens:
Weird. Any idea why it's only broken for search results, but not discussion boards, on the live site? At the moment, hashtag links don't work in Zoo Talk so it would be good to get this fixed.

wgranger:
Yeah, I think it's because on the live site, search results always hits that `@props.project` conditional no matter if it's in a project talk or global talk. If it's global talk, it won't be able to find the `.slug` to `.split` because the project prop will be an empty object. It works on a project talk because there will be a `.slug` to `.split` on that project prop.

On the other hand, the discussion boards (`TalkBoard`) always work because, as it currently is, that `else if @props.project` line is only reached if it's in a project specific talk because a complete project resource/prop is passed down to the Discussion Preview. In global talk, no project prop is passed down, so that `else if @props.project` line is ignored.

I pulled this branch down and ran it locally in the production environment, and it looks like the hashtag links are also working when a default project prop is given to `TalkDiscussionPreview`, but I could be missing something.

eatyourgreens:
Thanks. That makes sense. I've added tests to catch those cases, and updated `DiscussionPreview` so it passes them.

eatyourgreens:
Staged at https://zoo-talk-search.pfe-preview.zooniverse.org?env=production

wgranger:
Nice. This is looking good.

PR #3452, opened 2017-02-07
Convert classification summary

CKrawczyk:
The ES6 conversion and tests for `classification-summary` (taken from #3319) updated to include the first-to-classify split testing from #3434.  

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://convert-classification-summary.pfe-preview.zooniverse.org/

## Optional

There are `no-underscore-dangle` Errors, but this is caused by following the `_key` naming convention used throughout the app.  Should we add this into our `.eslintrc` to avoid an error?
There is also an `no-unused-vars` Error for the `apiClient` that was added in #3434, I assume this package is needed for the split testing so I kept it in.


eatyourgreens:
I've added tests for `TextSplit`, which uncovered a bug in the if statement: `props.classificationCount` will be false if the count is 0, so the split message wouldn't be shown.

app/classifier/classification-summary.jsx
```diff
@@ -0,0 +1,55 @@
...
+        splits={props.splits}
+        default={''}
+        elementType={"p"}
+      />
+    );
+  }
+  let body = 'No annotations';
+  if ((props.classification) && (props.classification.annotations.length > 0)) {
+    body = [];
+    for (const annotation of props.classification.annotations) {
```
srallen: Instead of using a for loop, you could use map which returns an array and assign body to it, `body = props.classification.annotations.map((annotation) => { return (annotation.thing) });`

app/classifier/tasks/survey/editor.cjsx
```diff
@@ -7,7 +7,8 @@ surveyEditorHelp = require './editor-help'
 AutoSave = require '../../../components/auto-save'
 handleInputChange = require '../../../lib/handle-input-change'
 NextTaskSelector = require '../next-task-selector'
-`import MediaArea from '../../../pages/lab/media-area/';`
+# `import MediaArea from '../../../pages/lab/media-area/';`
+MediaArea = require('../../../pages/lab/media-area/').default
```
srallen: Should this change be coming in with this PR?

eatyourgreens: I think the tests will crash without this change, but I haven't tested that to make sure.

eatyourgreens: Just tried this out locally, and mocha crashes on the `ClassificationSummary` tests without this line.

PR #3454, opened 2017-02-07
Restore the updated projects list to news section

eatyourgreens:
This section is blank at the moment, because `state.updatedProjects` was removed in #3321

### DIFF
--- a/app/pages/home-for-user/news-pullout.jsx
+++ b/app/pages/home-for-user/news-pullout.jsx
@@ -91,7 +91,7 @@ const NewsSection = React.createClass({
     const timestamp = moment(new Date(project.updated_at)).fromNow();
     return (<div key={project.id} className="home-page-news-pullout news-section__link">
       <a href={link}>
-        <h5 className="home-page-news-pullout news-section__title">{project.name} </h5>
+        <h5 className="home-page-news-pullout news-section__title">{project.display_name} </h5>
         <h5> has been updated! </h5>
         <p className="home-page-news-pullout news-section__timestamp">{timestamp}</p>
       </a>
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Thinking about it, home page projects are now sorted by when the user preferences were last updated, so this is probably wrong.

app/pages/home-for-user/news-pullout.jsx
```diff
@@ -101,6 +101,10 @@ const NewsSection = React.createClass({
   render() {
     const projLink = `${window.location.origin}/projects/${this.state.newestProject.slug}`;
     const avatarSrc = !!this.state.newestAvatar ? this.state.newestAvatar.src : null;
+    const recentProjects = 
+      this.props.updatedProjects
+      .sort((a, b) => { return new Date(b.updated_at) - new Date(a.updated_at); })
```
wgranger: Is this necessary? When the initial API call is made to retrieve project preferences, it is sorted by `-updated_at`. I assume that would do the sorting for you, but I could be missing something here.

wgranger:
Good catch. I hadn't noticed this was missing.

app/pages/home-for-user/news-pullout.jsx
wgranger: Oh wait a minute, but then I notice the code makes a call to projects later with `getProjectsForPreferences`, so if the sorting is needed here to sort the projects you eventually get, that makes sense.

eatyourgreens: You want the projects that were recently updated by their owners here, not the projects that you have most recently classified on, so the list has to be sorted by `project.updated_at` rather than `preference.updated_at`.

wgranger: Gotcha.

PR #3462, opened 2017-02-08
[WIP] Frame annotator tweaks

srallen:
Moves the `<WarningBanner />` component out of the frame annotator and into the subject viewer where it better fits to make testing a bit more scoped to what the frame annotator actually does. Also fixes a few linter issues.

My test project only has a couple of subjects, so you can get an 'Already Seen' warning quickly: https://frame-annotator-tweaks.pfe-preview.zooniverse.org/projects/srallen086/srallen-testing/classify

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/frame-annotator.jsx
+++ b/app/classifier/frame-annotator.jsx
@@ -10,8 +10,7 @@ export default class FrameAnnotator extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      alreadySeen: false,
-      showWarning: false
+      alreadySeen: false
     };
     this.getScreenCurrentTransformationMatrix = this.getScreenCurrentTransformationMatrix.bind(this);
     this.getEventOffset = this.getEventOffset.bind(this);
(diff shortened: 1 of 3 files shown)


eatyourgreens:
img_6840
This seems to add the buttons in Talk too, where they don't make sense. 

I couldn't leave a comment about a subject after pressing Done And Talk either.

srallen:
Alright, I'll take a look at those two cases. Thanks.

eatyourgreens:
I couldn't reproduce that Done & Talk bug today, so it might only happen when you comment on a subject after viewing it for the first time.

The `SubjectViewer` was originally designed to be independent of workflow, so I'm not sure if it's the best place to be checking for a subject's classification status.

srallen:
That makes sense, but `<FrameAnnotator />` doesn't really seem like the right fit either since it handles the svg layer. I'm gonna think about this a little more because my changes broke the subject viewer in the collections view too.

srallen:
ETA: I might end up closing this and adding tests for the frame annotator as is.

eatyourgreens:
I think the `FrameAnnotator` has ended up being the dumping ground for a bunch of classification code that doesn't live in a generic subject viewer. Just looking at the huge number of props for that component makes me think it's not a good piece of code. It breaks the OO guideline that a class should do just one thing, and do it well.

I think we should split out the SVG stuff into a dedicated `DrawingSurface` component (or similar.) What do you think?

eatyourgreens:
Also, refactoring the SVG drawing code is perhaps a job for whoever takes on #3465

srallen:
Yes! I think wrapper component that handles the warning banner and the before, inside, and after hooks would be a good idea. I'll open up an issue in the rfc repo so we can discuss architectural changes we want to make to the classifier after we're done with the rewrite and tests.

srallen:
Gonna close and add tests for frame annotator as is. This was not the small change I thought it would be.

eatyourgreens:
I have some ideas about the hooks too, but they can wait for a discussion of refactoring the classifier.

PR #3463, opened 2017-02-08
WorldWide Telescope ES6 Rewrite

wgranger:
Describe your changes.
ES6 doesn't allow class variables, so I had to be a bit creative with this rewrite. I've been testing this against the education subject set of Astronomy Rewind, which can also be found on a test project in staging.

# Review Checklist

- [ ] Does it work on mobile?

https://wwt-es6.pfe-preview.zooniverse.org/projects/wgranger-test/old-astro-test

## Optional


eatyourgreens:
I might be misunderstanding your comment, but class variables are part of JavaScript so exist in ES6 too eg. `MyClass.x = 1` would set a static property on `MyClass`. Similarly, you can add class methods by defining functions on the class, instead of an instance `MyClass.doStuff = () => { return "Hello!";}`.

wgranger:
Yes, that's what I wound up doing. However, I was unable to define a static variable from within a class (without using the constructor).

eatyourgreens:
You could use `this.constructor.staticProperty = 1;` to add class properties and methods inside the class but I find that syntax harder to read.

eatyourgreens:
Looks good to me, but could maybe do with a look from someone more familiar with the WWT component. I'm happy to merge this if it isn't breaking anything.

wgranger:
I believe @amyrebecca might be the only other person with WWT knowledge. However, she may be on vacation.

eatyourgreens:
Good point. I'll merge this.

PR #3471, opened 2017-02-10
convert single task to ES6

CKrawczyk:
Converts the `single` task to ES6

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://convert-single-task.pfe-preview.zooniverse.org/

## Optional


app/classifier/tasks/single.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+Summary.propTypes = {
+  task: React.PropTypes.object,
+  annotation: React.PropTypes.object,
+  expanded: React.PropTypes.bool
+};
+
+Summary.defaultProps = {
+  task: null,
+  annotation: null,
```
eatyourgreens: This default value will throw an error when checking properties eg. `if annotation.value`. A better default might be an empty annotation object, if that doesn't break the code.

CKrawczyk: Isn't that taken care of by the `getDefaultAnnotation` method?  Although I don't mind changing the default prop to `{ value: null }`.

CKrawczyk:
The staging link has been updated.

app/classifier/tasks/single.jsx
eatyourgreens: Oh yeah, I missed that. So tasks define a default value, then a method that completely ignores that value in favour of something else?

eatyourgreens: Oh wait, I was referring to the default on the `Summary` component, rather than the task.

CKrawczyk: Oh, good call.  As for the task components, the `getDefaultAnnotation` method is called via the `preparetoclassify` method on mount (https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L76, https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L143, https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L499), so the `defaultProp` will never be used.  By extension the `Summary` component should always have a non-null annotation object.

It might be in order to remove the `Summary`'s default annotation and just make the prop `isRequired`.

eatyourgreens: Good idea. I like that, if it doesn't break the code ie. we can rely on the classifier always passing annotations.

eatyourgreens: By the way, you can also use [`React.PropTypes.shape`](https://facebook.github.io/react/docs/typechecking-with-proptypes.html) to enforce that the annotation should have a value.

CKrawczyk:
I found and fixed a bug introduced in the `isAnnotationComplete` function (returned false when value was `0`) and added tests catch this bug.  The staged link has been updated.

PR #3490, opened 2017-02-15
Collection Contributions

### LINKED ISSUE #3263: Collaborative Collection
## Expected behavior
When you create a collaborative collection the collection should appear in all of the collaborators "My Collections" list. Perhaps it shows up with a small icon like http://fontawesome.io/icon/users/ similar to how private collections appear with a small lock icon to indicate private collections 

## Current behavior
It only appears in the creator of the collection list

## Steps to replicate
Make a collection. Add a buddy as a collaborator. Ask buddy to look at his/her collections. Not see it. Be confused.

zwolf:
Fixes #3263  .

Collections now have a `contributor` role that allows a user to add new subjects to a collection and nothing more. This is accessible from the collection's collaboration settings by the owner. 

* Logged in users looking at their own collection lists now see all of the collections that they are either an owner, collaborator, or contributor to. 
* If the collection is not owned by the logged in user, there's a little icon near the owner's name when viewing their own list.
* All owner names are now visible.

Staged at https://collab-collections.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -9,6 +9,7 @@ ID_PREFIX = 'COLLECTION_COLLABORATORS_PAGE_'
 POSSIBLE_ROLES = [
   "owner",
   "collaborator",
+  "contributor",
   "viewer"
 ]
 
@@ -19,6 +20,9 @@ ROLES_INFO =
   collaborator:
     label: "Collaborator"
     description: "Collaborators have full access to add and remove subjects from the collection."
+  contributor:
+    label: "Contributor"
+    description: "Contributors can add subjects to the collection, but not edit or delete from them."
   viewer:
     label: "Viewer"
     description: "Viewers can see this collection even if it's private."
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -22,7 +22,7 @@ ROLES_INFO =
     description: "Collaborators have full access to add and remove subjects from the collection."
   contributor:
     label: "Contributor"
-    description: "Contributors can add subjects to the collection, but not edit or delete from them."
+    description: "Contributors can only add subjects to the collection."
   viewer:
     label: "Viewer"
     description: "Viewers can see this collection even if it's private."
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -32,3 +32,6 @@ require('./lib/log-deployed-commit')()
 window?.zooAPI = require 'panoptes-client/lib/api-client'
 window?.talkAPI = require 'panoptes-client/lib/talk-client'
 require('./lib/split-config')
+
+# zooAPI.root = 'http://localhost:3000/api'
+# talkAPI.root = 'http://localhost:3001'
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -32,6 +32,3 @@ require('./lib/log-deployed-commit')()
 window?.zooAPI = require 'panoptes-client/lib/api-client'
 window?.talkAPI = require 'panoptes-client/lib/talk-client'
 require('./lib/split-config')
-
-# zooAPI.root = 'http://localhost:3000/api'
-# talkAPI.root = 'http://localhost:3001'
--- a/app/pages/collections/list.cjsx
+++ b/app/pages/collections/list.cjsx
@@ -72,7 +72,7 @@ List = React.createClass
         @setState {collections}
 
   shared: (collection) ->
-    if (@props.params.collection_owner is @props.user.login) or (@props.params.profile_name is @props.user.login)
+    if (@props.params.collection_owner is @props.user?.login) or (@props.params.profile_name is @props.user?.login)
       @props.user?.id isnt collection.links.owner.id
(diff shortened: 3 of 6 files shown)


app/collections/collaborators.cjsx
```diff
@@ -19,6 +20,9 @@ ROLES_INFO =
   collaborator:
     label: "Collaborator"
     description: "Collaborators have full access to add and remove subjects from the collection."
+  contributor:
+    label: "Contributor"
+    description: "Contributors can add subjects to the collection, but not edit or delete from them."
```
camallen: Maybe just `Contributors can only add subjects to the collection`

app/main.cjsx
```diff
@@ -32,3 +32,6 @@ require('./lib/log-deployed-commit')()
 window?.zooAPI = require 'panoptes-client/lib/api-client'
 window?.talkAPI = require 'panoptes-client/lib/talk-client'
 require('./lib/split-config')
+
+# zooAPI.root = 'http://localhost:3000/api'
```
camallen: remove these left overs

app/pages/collections/list.cjsx
```diff
@@ -65,6 +65,9 @@ List = React.createClass
       .then (collections) =>
         @setState {collections}
 
+  shared: (collection) ->
```
camallen: should you just compare the logged in user id vs the collection link owner id?

srallen: I don't think it should be compared to the logged in user since the `collections/:collection_owner` route with any public collections can be viewed while logged out. Logged out users should be able to see if the collection is owned by the `:collection_owner` user or is a collab collection and who the owner is.

app/pages/collections/list.cjsx
```diff
@@ -50,7 +50,7 @@ List = React.createClass
   listCollections: (props) ->
     query = {}
     if props.params?.collection_owner?
-      query.owner = props.params.collection_owner
+      query.current_user_roles = "owner,contributor,collaborator"
```
srallen: This is still a bit wonky. Logged in as my staging user, srallen086, and viewing `/collections/zwolfenbarger` I see my own collections. `zwolfenbarger` in the route is the `prop.params.collection_owner`. I think the conditional here will need to be reworked since the collection owner won't always be the logged in user.

screen shot 2017-02-16 at 11 09 25 am

The `query.owner = props.params.collection_owner` will need to remain. There needs to be a way to be able to fetch the roles for the `:collection_owner` user param.

app/pages/collections/list.cjsx
```diff
@@ -50,7 +50,7 @@ List = React.createClass
...
     else if props.params?.profile_name?
       query.owner = props.params.profile_name
```
srallen: This is for the `users/:profile_name/collections` route and similarly would need a way to fetch the collections for each role that is linked the `:profile_name` user . Unless we decide that if a user is viewing the collections via the profile that they should only strictly see the collections that that the `:profile-owner` owners. To be honest, I'm not too keen on this route existing at all since it loads up the same collection list component as `/collections/:collection_owner`

app/pages/collections/list.cjsx
zwolf: This was when we thought that the difference between `params.collection_owner` and `params.profile_owner` was the route, with the former being our own page and the latter being someone else's. We have since discovered that is not the case. Using owner will def not work because we explicitly want to see collections that we do not own. 

I need a way to, on that page, determine whether I should use `current_user_roles` and view my own owned, collaborated, and contributed collections, OR use `owner` because I'm looking at someone else's and should only see their public, owned collections.

app/collections/collaborators.cjsx
zwolf: ✔️

app/main.cjsx
zwolf: ✔️

app/pages/collections/list.cjsx
zwolf: `if props.params.collection_owner is props.user.display_name` sure seems like it should work, but there's apparently a race condition where that `props.user` object only exists after the user is grabbed but this is rendered before.

app/pages/collections/list.cjsx
srallen: Spoke with @zwolf and he clarified the functionality here for me, so I understand the suggestion now. Comparing to user id is what should be done.

app/pages/collections/list.cjsx
```diff
@@ -65,6 +71,10 @@ List = React.createClass
       .then (collections) =>
         @setState {collections}
 
+  shared: (collection) ->
+    if (@props.params.collection_owner is @props.user.login) or (@props.params.profile_name is @props.user.login)
```
srallen: It's not breaking the app, but there is a TypeError being thrown if you load the `collections/:collection_name` route directly while it's waiting for the user authentication. Adding the existential operator to `@props.user` in the conditional should fix it.

PR #3496, opened 2017-02-16
fix dropdown task with multiple selects sharing condition

### LINKED ISSUE #3488: Dropdown task breaks if two menus depend on the same parent menu
## Expected behavior
Create a task with three menus, where the selected option for menu 1 determines the available options for menus 2 and 3.

Selecting an option for 1 should enable options to be selected for 2 and 3.

## Current behavior
Selecting an option for menu 2 clears menu 3, since `optionsKeys` simply concatenates all selected options for the task.

mcbouslog:
Fixes #3488.

Just keeps track of `optionsKeys` in state, making `annotation.values` source of options answered and selected, via `getOptionFromAnnotation` to convert annotation to option as react-select want's it.

Notes from related (merged) PR:
- [ ] is use of `assert.deepEqual` ok?
- [ ] just use multiSelects, eliminate singleSelect?
- [ ] couldn't get `.find({key: 'countyID'})` to work instead of `.find('#countyID')`
- [ ] rename `county` to `region` for clarity?

# Review Checklist

- [ ] Does it work in all major browsers: ~~Firefox~~, ~~Chrome~~, Edge, ~~Safari~~?
- [ ] ~~Does it work on mobile?~~
- [x] Did you deploy a staging branch? - https://dropdown-fixes.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -82,23 +82,21 @@ module.exports = React.createClass
         not select.required or (annotation.value[i]?.value? and annotation.value[i]?.value isnt "")
 
   getInitialState: ->
-    selectedOptions: []
+    optionsKeys: {}
 
   componentWillMount: ->
     @menus = []
     @syncAnnotations()
 
-  componentDidMount: ->
-    annotationValues = @props.annotation.value
-    unless annotationValues.length
-      @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.task isnt @props.task
+      @syncAnnotations()
 
+  componentDidMount: ->
     if @props.autoFocus is true
       @handleFocus()
 
   componentDidUpdate: (prevProps) ->
-    if prevProps.task isnt @props.task
-      @syncAnnotations()
     if prevProps.task isnt @props.task and @props.autoFocus is true
       @handleFocus()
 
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -82,23 +82,21 @@ module.exports = React.createClass
...
     @menus = []
     @syncAnnotations()
 
-  componentDidMount: ->
-    annotationValues = @props.annotation.value
-    unless annotationValues.length
-      @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
+  componentWillReceiveProps: (nextProps) ->
+    if nextProps.task isnt @props.task
+      @syncAnnotations()
```
eatyourgreens: This will call `syncAnnotations` before `props.task` has updated. Just a heads up because there's a comment on that method that warns not to do this.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -82,23 +82,21 @@ module.exports = React.createClass
...
 
+  componentDidMount: ->
     if @props.autoFocus is true
       @handleFocus()
 
   componentDidUpdate: (prevProps) ->
-    if prevProps.task isnt @props.task
-      @syncAnnotations()
```
eatyourgreens: See previous comment. Shouldn't we be waiting for props to update before setting the annotation value.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -107,38 +105,48 @@ module.exports = React.createClass
...
-    optionsKeys = []
-    @props.annotation.value.map (annotation, i) =>
-      if annotation.option
-        [selected] = @getOptions(i, optionsKeys).filter (option) -> option.value is annotation.value
-        optionsKeys.push selected.value if selected?.value
-        selectedOptions[i] = selected
-      else
-        selectedOptions[i] = {label: annotation.value, value: annotation.value}
-    @setState {selectedOptions}
+    if @props.annotation.value.length?
```
eatyourgreens: Looks like this comment doesn't describe what the method does any more?

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -107,38 +105,48 @@ module.exports = React.createClass
...
+      for answer, i in @props.annotation.value
+        @handleOptionsKeys(i, answer.value)
 
   handleFocus: ->
     @menus[0].focus()
 
-  getOptions: (i, optionsKeys) ->
-    optionsKey = if optionsKeys.length then optionsKeys.join(';') else '*'
+  handleOptionsKeys: (i, value) ->
```
eatyourgreens: What does `handleOptionsKeys` do? Can the method name be changed to describe what it does?

PR #3497, opened 2017-02-17
Dropdown option keys

### LINKED ISSUE #3488: Dropdown task breaks if two menus depend on the same parent menu

eatyourgreens:
Fixes #3488 .

Describe your changes.
Uses a hash to store option keys for the dropdown task, instead of an array. Individual menus can then look up the selected options for the menus that control them.
# Review Checklist

- [ ] Does it work on mobile?

## Optional


eatyourgreens:
@mcbouslog this is the direction I was thinking of going in for the dropdown task. Not sure if it's any better than the approach in #3496

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -92,6 +92,7 @@ module.exports = React.createClass
     annotationValues = @props.annotation.value
     unless annotationValues.length
       @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
+      @syncAnnotations()
```
eatyourgreens: This magically fixes something but I don't know what.

mcbouslog:
made a PR for possible edits; agree this structure preferred.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -121,24 +120,30 @@ module.exports = React.createClass
   handleFocus: ->
     @menus[0].focus()
 
-  getOptions: (i, optionsKeys) ->
-    optionsKey = if optionsKeys.length then optionsKeys.join(';') else '*'
+  getOptionsKey: (i, optionsKey = '') ->
+    {selects} = @props.task
+    select = selects[i]
+    [condition] = selects.filter (filterSelect) => filterSelect.id is select.condition
```
eatyourgreens: This can probably be speeded up so that it doesn't have to loop through the `selects` array every time it's called.

Maybe we should store a lookup table of menu ID => selected value.

eatyourgreens: Or maybe when a dropdown menu changes value, it should look ahead for all menus that it controls, then update those too?

mcbouslog: I think `clearRelated` kind of looks ahead and clears other menus that the menu that has changed controls, is that similar to what you're referring to?

I like the idea of the lookup table, would we store it in state? If so, that might be similar to the `optionsKeys` object Brian and I tried in the [old version](https://github.com/zooniverse/Panoptes-Front-End/commit/32c68b104b2f6d60c0327094696b23f8daafd2fb).

eatyourgreens: Hmmm, I got rid of that `optionKeys` variable because it didn't always stay in synch with `state.selectedOptions`. Anyone working on this code would have to somehow know that updating one of those variables should also update the other. A lookup table that stored the selected options keyed by `select.id` would be ideal here.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -121,24 +120,30 @@ module.exports = React.createClass
...
+    conditionIndex = selects.indexOf(condition)
+    optionsKey = if optionsKey.length is 0 then @props.annotation.value[conditionIndex]?.value else "#{@props.annotation.value[conditionIndex]?.value};#{optionsKey}"
```
eatyourgreens: An alternative to this would be store an array of option keys, then `optionKeys.join(';')` at the end.

eatyourgreens: Any reason we use the annotation prop rather than the values stored in state here?

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -121,24 +120,30 @@ module.exports = React.createClass
...
+    if condition.condition? then @getOptionsKey(conditionIndex, optionsKey) else optionsKey
```
eatyourgreens: `condition.condition` is a bad variable name here.

app/classifier/tasks/dropdown/index.cjsx
mcbouslog: There are two tests that fail when state is used:
1. `DropdownTask with multiple selects and all annotations provided, no custom answers, should have the supplied annotations selected`
2. `DropdownTask with multiple selects and all annotations provided, including custom answers, should have the supplied annotations selected`
I think due to if `annovation.values` has answers already (clicking back, or persist annotations option enabled), `syncAnnotations` runs before mount, or the state has a chance to update.

eatyourgreens: How weird. `syncAnnotations` runs before mount so that selected options are properly set when the menus render. Keeping state in sync with props is a problem, though.

app/classifier/tasks/dropdown/index.cjsx
eatyourgreens: I'm an idiot. The tests won't wait for state to change, I guess, which is why it would be dangerous to use it here.

eatyourgreens:
Staged at https://dropdown-option-keys.pfe-preview.zooniverse.org/ in case that latest commit has completely broken everything.

app/classifier/tasks/dropdown/index.cjsx
mcbouslog: maybe `selectCondition` instead? not sure condition is the best word for something that is required before the select in focus. just did a quick Google thesaurus search and a few contender words popped up - antecedent or precursor?

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -102,43 +98,48 @@ module.exports = React.createClass
     if prevProps.task isnt @props.task and @props.autoFocus is true
```
eatyourgreens: Good catch!

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -102,43 +98,48 @@ module.exports = React.createClass
...
       @handleFocus()
 
-  syncAnnotations: () ->
+  selectedOptions: () ->
     # this can only be called after props have been updated
```
eatyourgreens: Nope, not now that we aren't storing anything in component state.

app/classifier/tasks/dropdown/index.cjsx
eatyourgreens: I've gone with `parentMenu` 😄

eatyourgreens:
Updated and staged with those small changes. `getOptionKeys` could maybe be improved, particular for tasks that have complex connected menus, but that can be done later.

mcbouslog:
This looks good to me! Tested out in dev classifier and a few testing workflows as well.

PR #3498, opened 2017-02-17
Allow a default frame to be set for flipbooks

### LINKED ISSUE #3464: Make default frame number configurable for multiple frame subjects
## Expected behavior
Allow the default frame for the subject flipbook to be configured as a workflow preference eg. default to the 3rd frame out of 5.

## Current behavior
Subjects with multiple frames default to the first frame.

## Additional information
Should this setting be ignored when the flipbook frames are viewed side-by-side?

rogerhutchings:
Fixes #3464.

Respects a `default_frame` property if set in a multi-frame subject's metadata so the flipbook starts on that frame.

Also fixes a bug where flipbook loops would finish on the next frame, not the current frame.

Can be tested at https://default_frame.pfe-preview.zooniverse.org/projects/rogerhutchings/default-frame-test/classify, default frame should be ~~4~~ 3, now the property is 1-indexed

cc @hrspiers 

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://default_frame.pfe-preview.zooniverse.org/projects/rogerhutchings/default-frame-test/classify

## Optional

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -50,7 +50,7 @@ module.exports = React.createClass
   getInitialState: ->
     loading: true
     playing: false
-    frame: @props.frame ? 0
+    frame: @props.frame ? parseInt(@props.subject.metadata.default_frame, 10) ? 0
     frameDimensions: {}
     inFlipbookMode: @props.allowFlipbook
     promptingToSignIn: false
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -48,18 +48,22 @@ module.exports = React.createClass
     workflow: null
 
   getInitialState: ->
-    initialFrame = if @props.frame
-        @props.frame 
-      else if @props.allowFlipbook and @props.subject.metadata.default_frame? and @props.subject.metadata.default_frame <= @props.subject.locations.length
-          parseInt(@props.subject.metadata.default_frame, 10) - 1 
-      else 0
-
     loading: true
     playing: false
-    frame: initialFrame
+    frame: @getInitialFrame()
     frameDimensions: {}
     inFlipbookMode: @props.allowFlipbook
     promptingToSignIn: false
+
+  getInitialFrame: ->
+    {frame, allowFlipbook, subject} = @props
+    default_frame = parseInt(subject.metadata.default_frame, 10)
+    initialFrame = 0
+    if frame?
+      initialFrame = frame
+    else if allowFlipbook and typeof default_frame is 'number' and !isNaN(default_frame) and default_frame > 0 and default_frame <= subject.locations.length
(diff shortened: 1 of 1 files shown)


shaunanoordin:
I suspect the `parseInt()` may be the cause of the problem. Given...

`frame: @props.frame ? parseInt(@props.subject.metadata.default_frame, 10) - 1 ? 0`

...if `@props.subject.metadata.default_frame` is an empty string or null, `parseInt()` will return `NaN` instead of `null`. I think CoffeeScript may consider `NaN` a valid answer, so the final `... ? 0` default value doesn't trigger.

rogerhutchings:
Good catch, I'll get this fixed

rogerhutchings:
Okay, just pushed a change and restaging the branch

shaunanoordin:
## PR Review (Update)
* As a user, when I access a Classification with multiple images (frames), the first image/frame I see is the specified "default frame"
* I can ALSO access a Classification with a single image (no multi-frames) normally.
* Tested on https://default_frame.pfe-preview.zooniverse.org/projects/rogerhutchings/default-frame-test/classify (multiframe) and https://default_frame.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats (single image)
* Tested with OSX+FF51/Chrome56/Safari9 and Win7+IE11

### Minor observations
* Tangential: With IE11 and Safari, when we click "play" to auto-flip through images, the auto-flip will end at one frame before the starting frame, instead of at the starting frame. e.g. with frames A, B, C, D, if we click "play" at frame C, the auto-flip will end at frame B. In FF and Chrome, the auto-flip would correctly end at C. This is neither a major issue nor directly caused by this PR, but might be worth noting because... uh, it doesn't make sense?
* Edge case: The `default_frame` PR tests perfectly OK with existing projects with _legit_ configs/setups. However, I'm mildly curious to know what happens when someone sets `props.subject.metadata.default_frame` to 0 or a negative number.

## Status
LGTM! Happy to see this merged, unless @eatyourgreens has additional comments.

rogerhutchings:
- Hah; I changed the auto-flip because in Chrome / FF it was finishing one frame _too late_. I imagine it's to do with how `timeout`s are implemented, one might be rounding down and the other up? We could track progress by the number of flips, rather than time.
- I imagine if the default_frame was 0 / a negative number it would break, but what'd be the point? Just in case, however, I'll add in a check for that.

shaunanoordin:
## PR Review (Update)
* Updated code seems logical and catches the weird possibility of `default_frame` being an invalid number. 
* All previous tests (multiframe & single image) check out OK
* Tested on OSX+Chrome on http://localhost:3735/projects/rogerhutchings/default-frame-test/classify and http://localhost:3735/projects/brooke/i-fancy-cats/classify

## Status
LGTM!

PR #3500, opened 2017-02-17
edit options and optionsKey

mcbouslog:
Related to #3497.

- Not sure this helps, but pulls out `optionsKey` craziness to it's own function.
- Removes magical `@syncAnnotations` that didn't belong, related to... 
- `getDisabledAttribute` refactored. `getDisabledAttribute` shows the input as enabled or disabled, which for select's with `allowCreate = true` we want always enabled, but for select's with `allowCreate = false` we only want enabled if condition answered with a provided answer (`option = true`).
- edits test to reflect initial `annotation.value` array of null/false answers after mount

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://dropdown-option-keys-mb.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -92,7 +92,6 @@ module.exports = React.createClass
     annotationValues = @props.annotation.value
     unless annotationValues.length
       @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
-      @syncAnnotations()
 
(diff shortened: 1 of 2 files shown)


eatyourgreens:
@mcbouslog Thanks! Look like I completely misunderstood how `condition` works. So this looks recursively back up the list of menus, selecting matching IDs?

PR #3512, opened 2017-02-22
Fetch subjects to classify using new API

marten:
Panoptes is deployed with this API.

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -20,11 +20,6 @@ FAILED_CLASSIFICATION_QUEUE_NAME = 'failed-classifications'
 
 PROMPT_MINI_COURSE_EVERY = 5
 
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks good to me. Is passing a path to `get` an undocumented feature of the API client? I don't think it's mentioned in zooniverse/json-api-client or zooniverse/panoptes-javascript-client

marten:
I didn't actually look at the documentation, but looked up how PFE was handling other special paths called on a resource collection, like https://github.com/zooniverse/Panoptes-Front-End/blob/aae8d7484d30997db0a0ca16812703bb79d9a9ee/app/classifier/index.cjsx#L125

I can add some documentation on this on the json-api-client repo to make sure we don't forget to keep it working.

eatyourgreens:
@marten documentation sounds great. I'm not sure if this is a feature of the json API client, or a bug in the way it handles resource types 😳

PR #3516, opened 2017-02-22
Add fields filter to project index workflows request

mcbouslog:
Improves workflows response size on project index.

- edits `getAllWorkflows` to add `fields` filter to get only workflow attributes needed for index, home, home-workflow-buttons and home-workflow-button
- edits `getWorkflow` to get selected, uncached, workflow

Not sure this is the best solution, and as an integral part of the volunteer/classification process this will need thorough review.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://add-fields-workflow.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -137,7 +137,7 @@ ProjectPage = React.createClass
       .then (pages) =>
         @setState {pages}
 
-  getAllWorkflows: (project, query = { active: true, fields: "active,completeness,configuration,display_name,finished_at,retirement,created_at" }) ->
+  getAllWorkflows: (project, query = { active: true, fields: "active,configuration,display_name" }) ->
     @setState { loadingSelectedWorkflow: true }
     getWorkflowsInOrder(project, query)
       .then (activeWorkflows) =>
@@ -179,10 +179,17 @@ ProjectPage = React.createClass
       @state.activeWorkflows[randomIndex].id
 
   getWorkflow: (selectedWorkflowIndex) ->
-    @setState {
-      selectedWorkflow: @state.activeWorkflows[selectedWorkflowIndex],
-      loadingSelectedWorkflow: false
-    }
+    apiClient.type('workflows').get({ id: "#{@state.activeWorkflows[selectedWorkflowIndex].id}" })
+      .catch (error) =>
+        console.error error
+        @setState {
+          loadingSelectedWorkflow: false
+        }
+      .then ([workflow]) =>
+        @setState {
+          selectedWorkflow: workflow,
+          loadingSelectedWorkflow: false
+        }
 
   isWorkflowInactive: (project, selectedWorkflowID) ->
     selectedWorkflowIndex = @state.activeWorkflows.findIndex (workflow, index) ->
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks ok to me. So `state.activeWorkflows` stores a limited description of the full workflows list, for use in menus etc.? And the full workflow resource is only requested when I decide to classify with that workflow?

I wouldn't mind a second opinion from @srallen, as she's more familiar with the workflow selection code.

app/pages/project/index.cjsx
```diff
@@ -179,10 +179,17 @@ ProjectPage = React.createClass
...
-      loadingSelectedWorkflow: false
-    }
+    apiClient.type('workflows').get({ id: "#{@state.activeWorkflows[selectedWorkflowIndex].id}" })
+      .catch (error) =>
+        console.error error
+        @setState {
+          loadingSelectedWorkflow: false
+        }
+      .then ([workflow]) =>
+        @setState {
```
eatyourgreens: Should this perhaps also store the full workflow in `state.activeWorkflows[selectedIndex]`? We'd need to check whether we have the full resource already, before the API request, so that we don't request it again this session.

srallen:
I'll take a look at this today.

app/pages/project/index.cjsx
```diff
@@ -137,7 +137,7 @@ ProjectPage = React.createClass
       .then (pages) =>
         @setState {pages}
 
-  getAllWorkflows: (project, query = { active: true }) ->
+  getAllWorkflows: (project, query = { active: true, fields: "active,configuration,display_name" }) ->
```
srallen: `completeness` is needed for the project metadata

app/pages/project/index.cjsx
mcbouslog: would the following be in the right direction?
```
  getWorkflow: (selectedWorkflowIndex) ->
    if @state.activeWorkflows[selectedWorkflowIndex].tasks?
      @setState {
        selectedWorkflow: @state.activeWorkflows[selectedWorkflowIndex],
        loadingSelectedWorkflow: false
      }
    else
      apiClient.type('workflows').get({ id: "#{@state.activeWorkflows[selectedWorkflowIndex].id}" })
        .catch (error) =>
          console.error error
          @setState {
            loadingSelectedWorkflow: false
          }
        .then ([workflow]) =>
          activeWorkflows = @state.activeWorkflows
          activeWorkflows[selectedWorkflowIndex] = workflow
          @setState {
            selectedWorkflow: workflow,
            activeWorkflows: activeWorkflows,
            loadingSelectedWorkflow: false
          }
```

srallen: I can see checking for `@state.selectedWorkflow` to prevent unnecessary API calls, but I'm not sure if I understand the utility of storing the full resource in `state.activeWorkflows` too. It's used for two reasons: to check if the selected workflow is active (for now, but the project resource now only returns active workflows links! zooniverse/Panoptes#2073) and to use with rendering the buttons on the home page. Neither use needs the full resource, imo.

srallen: Oh and `state.activeWorkflows` is used in selecting a random workflow too, but is just used for doing the math of selecting a random index.

srallen:
I'm gonna go ahead and merge this since this is good for now and I'm working on a more thorough refactor this week.

Issue #3240, opened 2016-11-17
Update lab workflow edit page UI


srallen:
## Expected behavior
Allow project owners to update certain active workflow options and fields.

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Current the UI locks down the entire edit page if the workflow is active.

Not sure if there was already an open issue for this, but I couldn't find one.

Issue #2896, opened 2016-08-16
Add ability to switch subject sets on an active workflow without having to to make the workflow inactive


mschwamb:
Following discussions in https://github.com/zooniverse/Panoptes-Front-End/pull/2889
It would simpler for project owners and collaborators if they could check and uncheck subject sets to make them linked to an active workflow without having to leave the workflow page. @camallen says the api already supports this. cc @srallen

mschwamb:
fixed by #3517

Issue #2355, opened 2016-03-30
Making the 'visualize this workflow' button in the workflow part of the project builder not locked when you have a live workflow


mschwamb:
Perhaps one for @CKrawczyk Having the 'visualize this workflow' button be available when a workflow is locked and live would be helpful for data analysis and research team discussions. We were talking about data reduction for Comet Hunters this week and the PI was doing a classification to remind himself about how we ordered the questions. I think the ability to pull up the visualization for discussions, talks, etc would be really useful if it's possible

alexbfree:
Good point! I think this is part of a more general problem with the way that the workflow is disabled when live. Really everything should be available in a read-only manner, it's really only editing that we need to block. My suggestion would be that we need to update the workflow page to isolate and block its edit functionality, but not its read functionality.

aliburchard:
It's actually available when the project is locked/live. It appears gray, but you can still click on it.

mschwamb:
The grey signals not accessible, so I don't think that's  intuitive to know you can click the button with the workflow locked and that my project would be okay if I did. Maybe the coloring of the button can be changed so that it's very apparent you can can click on it even if the workflow is locked.

CKrawczyk:
I looked into how to make this not be grey when a project is live, but I see no obvious way of doing it.  I am not really sure how the greyed out style is being applied to that page in the first place.  Maybe somebody with more PFE experience can help with this.

aliburchard:
Yeah - that needs a bit of an overhaul. Related to the need for a conversation about what we should allow projects to do while "live" (e.g. we need to enable linking subject sets on live projects).

CKrawczyk:
To give some context, the opacity is being set directly on the element of the "columns-container" not in the css via class name, this makes it impossible to override the opacity of just one element inside of the column (even using an `!important`).  This would need to be externalized into the css to allow changes to be made on an element-by-element basis.

mschwamb:
fixed by #3517

PR #3517, opened 2017-02-23
Only disallow making changes to tasks while live

marten:
The API has always had a small list of things you can't do. Mostly you can't change tasks, but text changes and linking subjects is absolutely fine.


srallen:
This will close #3240 when it merges.

mschwamb:
presume this also fixes #2355

mschwamb:
also fixes #2896

PR #3523, opened 2017-02-24
Validate project before allowing beta application

rogerhutchings:
Now requires the following statements to be true about a project before an owner can apply for beta:

- Must have an active workflow
- Must have all about pages filled in (doesn't check very hard, just to see how many pages are attached to a project; you can circumvent this by entering some content and deleting it again :/ )
- ~~Must have at least 100 subjects in active workflows~~ [see below](https://github.com/zooniverse/Panoptes-Front-End/pull/3523#issuecomment-282685877)

@mrniaboc 

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://validate-project.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -51,38 +51,28 @@ module.exports = React.createClass
 
   validateProject: ->
     @setState validationErrors: []
-    projectData = 
+    validationCriteria = 
       activeWorkflows: @state.workflows?.filter (workflow) -> workflow.active
-      activeSubjects: 0
-      pages: []
+      missingPages: []
 
     @props.project.get 'pages'
-      .then (pages) -> 
-        projectData.pages = pages
-        if projectData.activeWorkflows.length
-          apiClient
-            .type 'subject_sets'
-            .get uniq projectData.activeWorkflows?.map (workflow) -> workflow.links.subject_sets
-      .then (sets) ->
-        console.info sets
-        if sets
-          projectData.activeSubjects = sets.reduce (count, set) -> 
-            count += set.set_member_subjects_count
-            count
-          , 0
+      .then (projectPages) -> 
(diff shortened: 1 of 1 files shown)


mschwamb:
@rogerhutchings I wouldn't require all of the about pages to be filled if I read this correctly. Because on the about section is results that won't necessarily be filled out. None of my panoptes projects did I put content at  launch, since those are supposed to be for news about results from then project.

 Also for example there's an Education tab, we didn't make one for P4: Ridges because there's not really much a class could do with that project that we could think of so we skipped that page. 

 Can you just check the FAQ and the research are pages are filled out only? I think those are two that should be required, but everything else I think depending on the project could be optional.

mschwamb:
or maybe just count for 3 pages:about, faq, and team ( as team is autogenerated)?

CKrawczyk:
I just want to point out that this validation uses `set_member_subjects_count` for subject sets, and this counter is know to be wrong for some projects (https://github.com/zooniverse/Panoptes/issues/2155), so this PR should wait until that bug is fixed.

rogerhutchings:
@mschwamb Fair enough on Results, I'll change it to require two pages called "Research" and "FAQ". "Team" isn't actually autogenerated as a page, only the list of collaborators; there's also a page resource called Team that goes with it.

@CKrawczyk Good catch - I can separate this into two PRs so we can get the other validations in now.

rogerhutchings:
Okay, this now checks explicitly for the Research and FAQ changes, and no longer checks the number of active subjects - that's been moved to #3525. This should now be ready for review.

cc @mrniaboc

rogerhutchings:
Having spoken to @camallen, he reckons not to let the Panoptes issue stop this PR, so I'm closing in favour of #3525

PR #3531, opened 2017-03-01
Add "App Status" banner

### LINKED ISSUE #3530: Add "App Status Banner" to Zooniverse.org
## Feature Request
* Users of zooniverse.org must be able to see any status updates - notably "emergency broadcasts" such as "sorry folks, server is down, brb"
* Zooniverse admins must be able to send out status updates to users.
* Proposed mechanism is as follows:
  * PFE will check a static resource to determine if there's a status message. (REQUIREMENT: this resource must be low-cost and easily scalable, as it'll be accessed every time a user accesses zooniverse.org. An API call to a paid API service is out of the question. A simple .txt file on an S3 bucket is acceptable - better if it's cached.)
  * If the static resource indicates a status message, that message will be displayed on the PFE as a banner.

Part of the (Stargazing Live 2017 project)[https://github.com/orgs/zooniverse/projects/5]. Original text:
```
Let PFE check if status.zooniverse.org is indicating a known issue. If so, display a small banner on every page. - marten
```

shaunanoordin:
## PR Overview
Closes #3530

* New feature: a "status banner" is displayed on zooniverse.org, if PFE detects a status message on a secondary location.
* Mechanism: PFE checks static.zooniverse.org for a static resource (expecting to find a simple .txt or .html file). If the resource exists, the status banner appears and displays the contents of the file to the user. (Final static resource URL not finalised: see Notes)
* Zooniverse admins will need to manually add/edit the static resource. (e.g. via the AWS CLI) We're going for _a simple and primitive, but reliable_ solution.
* The static "app status" resource will be hosted on static.zooniverse.org, as that would make it scalable to the number of PFE users. (Each active user will ping the resource once per loading of Zooniverse.org, whether or not there's a status message.)

Tagging @marten as request originator.

## Notes
* Open dev question: **should we only allow text for the status update message, or are we going to allow HTML and JavaScript to be part of the message?** (Shaun: leaning towards HTML+JS)
* Open dev question: XMLHttpRequest or fetch()? (Shaun: I have both working, but I like XMLHttpRequest since nothing else is using fetch(), and we don't need that extra `isomorphic-fetch` package.json dependency.)
* **Current dev issue:** setting up the static "status message" resource on static.zooniverse.org - along with the necessary CORS settings - is as easy as pulling out teeth. _And not through the mouth._
* Future plans: a more advanced implementation that uses WebSockets and Amazon Lambda to more consistently monitor changes to the status update resource. (Tagging @rogerhutchings for the idea.)

screen shot 2017-03-01 at 15 52 11
_Behold, the status banner. At the bottom of the screen. With a nonsense message._

## Status
WORK IN PROGRESS - please don't merge, but feedback is welcome (2017.03.01)

## Preview
URL: https://status-banner.pfe-preview.zooniverse.org/
What you're expected to see:
* The status banner should appear at the bottom of the screen, with gibberish text.

## Review Checklist
- [x] Does it work on mobile?

### DIFF
--- a/app/partials/app-status.jsx
+++ b/app/partials/app-status.jsx
@@ -6,6 +6,7 @@ import fetch from 'isomorphic-fetch';
 //const APP_STATUS_URL = 'https://www.theguardian.com/uk/rss';
 //const APP_STATUS_URL = 'http://localhost:3735/favicon.ico';
 const APP_STATUS_URL = 'https://static.zooniverse.org/variants.shakespearesworld.org/sake.txt';
+//const APP_STATUS_URL = 'https://static.zooniverse.org/index.html';
 
 export default class AppStatus extends React.Component {
   constructor(props) {
@@ -70,7 +71,7 @@ export default class AppStatus extends React.Component {
     return (
       <div className="app-status">
         <button className="fa fa-close" onClick={this.hide.bind(this)}></button>
-        <div className="message">{this.state.message}</div>
+        <div className="message" dangerouslySetInnerHTML={{ __html: this.state.message }}></div>
       </div>
     );
   }
--- a/app/partials/app-status.jsx
+++ b/app/partials/app-status.jsx
@@ -60,7 +60,6 @@ export default class AppStatus extends React.Component {
           show: true,
           message: text,
         });
-        if (this.button) { this.button.focus(); }  //<button autofocus=true/> won't work, so we'll manually trigger it.
       }
     })
     .catch((err) => {
@@ -74,7 +73,7 @@ export default class AppStatus extends React.Component {
     
     return (
       <div className="app-status">
-        <button ref={b=>this.button=b} className="fa fa-close" onClick={this.hide.bind(this)}></button>
+        <button className="fa fa-close" onClick={this.hide.bind(this)} autoFocus={true}></button>
         <div className="message">{this.state.message}</div>
       </div>
     );
--- a/app/partials/app-status.jsx
+++ b/app/partials/app-status.jsx
@@ -33,6 +33,7 @@ const APP_STATUS_URL = 'https://static.zooniverse.org/zooniverse.org-status.txt'
 export default class AppStatus extends React.Component {
   constructor(props) {
     super(props);
+    this.button = null;
     
     this.state = {
       show: false,
@@ -58,7 +59,8 @@ export default class AppStatus extends React.Component {
         this.setState({
           show: true,
           message: text,
-        });  
+        });
+        if (this.button) { this.button.focus(); }
       }
     })
     .catch((err) => {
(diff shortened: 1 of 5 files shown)


astopy:
> If the resource exists, the status banner appears and displays the contents of the file to the user.

Only superusers can actually delete things from zooniverse-static (and only then when they're explicitly using the superuser role, so that it's not possible to accidentally delete a bunch of stuff), though anyone can overwrite individual files.

From a quick look at the code, I think it's already doing the right thing, but just to make it clear: You'll want to check if the file exists _and_ contains stuff. If it doesn't exist (it'll actually be a 403 rather than a 404 with S3) or if it's empty then don't display the banner.

> setting up the static "status message" resource on static.zooniverse.org - along with the necessary CORS settings

It should just be a case of uploading a file, and CORS should already be set up. Let me know if you're having trouble.

> a more advanced implementation that uses WebSockets and Amazon Lambda to more consistently monitor changes to the status update resource

Nice idea, but probably overkill for something that's rarely (hopefully never) going to be used.

srallen:
Re the two questions:
I think I'd prefer txt. Even though access to the static bucket should be locked down, fetching and displaying whatever is there could open us up to a vulnerability if we allowed html/js.

We already polyfill for fetch (see: https://github.com/zooniverse/Panoptes-Front-End/blob/master/views/index.ejs#L72) and it's the current web spec standard, so that is what should be used.

eatyourgreens:
👍 for `fetch` here's [an example for grabbing text subjects](/zooniverse/Panoptes-Front-End/blob/master/app/components/text-viewer.jsx#L16) You don't need the `import` as @srallen says, it's supported in the browser now.

shaunanoordin:
## PR Update
Thanks for the feedback, everyone!

@astopy 
* You're right, the CORS works perfectly for a newly uploaded file! My previous error waltz must have been because I was testing on older files.
* Static status message URL set at https://static.zooniverse.org/zooniverse.org-status.txt
* Tested Input/Output:
  * static file found, is non-empty => show contents of static file in banner
  * static file found, is empty => don't show status banner
  * static file not found (403, 404, etc) => don't show status banner
* Tested on OSX+FF/Chrome/Safari, Win7+IE11

@srallen @eatyourgreens 
* Thanks for the fetch polyfill! I've removed the `isomorphic-fetch` dependency and phased out XMLHttpRequest in favour of _vanilla_ `fetch()`.
* Status Message will _only be displayed as text._ HTML+JS functionality has been removed for safety reasons.

## Notes
Not sure if tests are needed for this PR since the feature requires an external resource? If you believe I should write a test before this PR gets merged, please point me to a good example and I'll hop to it.

## Status
Ready for review, but please **do NOT merge until zooniverse.org-status.txt has been emptied**.

eatyourgreens:
Code looks good to me. Maybe autofocus the button so I can dismiss the popup with the enter or space keys?

shaunanoordin:
@eatyourgreens autofocus has been added! `<button autofocus />` doesn't seem to work so I've triggered it manually with `button.focus()` when the banner appears.

eatyourgreens:
React might have its own funky spelling of the attribute. `autoFocus` perhaps?

shaunanoordin:
Goddammit, React. Thanks @eatyourgreens - it was `autoFocus`.

shaunanoordin:
## PR Update
The status update file - https://static.zooniverse.org/zooniverse.org-status.txt - has been emptied, so this PR is now merge-safe. 👍

Issue #3527, opened 2017-02-27
Classifier ES6 conversion and tests


srallen:
As a follow up from all the work that was done related to #3319, there are some classifier related components that still need an ES6 conversions and/or tests written:

Classifier:
- [ ] `index.cjsx`
  - [ ] Split out independent components to different files (`render-summary`, `render-expert-options`, `classifier-helpers`, `index`, and `classifier`. PR #3522 related).
  - [ ] Tests
- [ ] `metadata-based-feedback.cjsx` (ES6 rewrite with tests in branch https://github.com/zooniverse/Panoptes-Front-End/tree/space-warps-feedback-refactor but needs some work, see: #3361)
  - [ ] Tests

Viewer (Found in components folder since some viewer components are shared with Talk, should they move to classifier folder?):
- [ ] `flag-subject-button.cjsx`
  - [ ] Tests
- [ ] `frame-viewer.cjsx` (Open [PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3553))
  - [ ] Tests
- [ ] `subject-viewer.cjsx`
  - [ ] Tests
- [ ] Tests for `text-viewer.jsx`
- [ ] Tests for `video-player.jsx`

Tasks:
- [ ] `drawing-task-details-editor.cjsx`
  - [ ] Tests
- [ ] `next-task-selector.cjsx`
  - [ ] Tests
- [ ] `shortcut.cjsx`
  - [ ] Tests
  - [ ] DRY component by receiving `addNewTask` function from parent
- [ ] `generic-editor.cjsx`
  - [ ] Tests
- [ ] Combo: `editor`, `index`, `summary`
  - [ ] Tests
- [ ] Crop: `index`, `initializer`, `summary`
  - [ ] Tests
- [ ] Drawing: `all-marks-display`, `grid-buttons`, `hide-previous-marks-toggle`, `icons`, `index`, `marking-initializer`, markings-renderer`, `summary`
  - [ ] Tests
- [ ] Dropdown: `presets` folder should be ES6 and json, `dropdown-dialog`, `dropdown-list`, `editor-help`, `editor`, `index`
   - [ ] Tests (Done for `index`)
- [ ] Survey: `annotation-view`, `choice`, `chooser`, `editor-help`, `editor`, `index`, `summary`, `utility`
  - [ ] Tests
- [ ] Text: `editor`, `index`
  - [ ] Tests

Drawing Tools:
[...]

PR #3522, opened 2017-02-24
Extract Task component

eatyourgreens:
Builds on #3520 and extracts the Task area to its own component.
# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch? https://classifier-task-es6.pfe-preview.zooniverse.org/
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #3532, opened 2017-03-01
Add Shortcut to the Dev Classifier

wgranger:
Describe your changes.
Adds "shortcut" to the dev classifier and creates tests for the component. Also, some styling was changed to clean up code and use preexisting classes.

# Review Checklist

- [x] Does it work on mobile?

https://shortcut-to-classifier.pfe-preview.zooniverse.org/

## Optional


wgranger:
This project that is under review uses the shortcut: https://shortcut-to-classifier.pfe-preview.zooniverse.org/projects/sweenkl/steller-watch/classify/?env=production

Styling changes were made because the `.unlinked-shortcut` class originally styled the shortcuts to always have the dark background of the classifier buttons, as opposed to the light background of the dev classifier buttons. Removing CSS from `.unlinked-shortcut` and giving shortcuts pre-existing classes solved this.

I can prob write this into ES6 pretty quickly right now.

wgranger:
Oh. Just saw your comment after I pushed. I went ahead and rewrote those two components into ES6.

app/classifier/tasks/shortcut.jsx
```diff
@@ -0,0 +1,148 @@
...
+  task: React.PropTypes.shape(
+    {
+      answers: React.PropTypes.array,
+      question: React.PropTypes.string
+    }
+  )
+};
+
+Summary.defaultProps = {
+  annotation: null,
```
srallen: A default prop of null will throw console warnings if the prop is also required. Either prop can't be required or this needs to be set to a default value.

app/classifier/tasks/shortcut.jsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  render() {
+    const options = this.props.workflow.tasks[this.props.task.unlinkedTask].answers;
+
+    return (
+      <div className="unlinked-shortcut">
+
+        {options.map((answer, i) => {
+          if (answer._key == null) { answer._key = Math.random(); }
```
srallen: Use strict equality whenever possible.

app/components/shortcut-editor.jsx
```diff
@@ -0,0 +1,130 @@
...
+
+        <small className="form-help">
+          Give volunteers the choice to skip to the end
+          of a classification if one of the following options is selected.
+        </small>
+
+        {shortcuts && (
+          <div className="workflow-task-editor-choices">
+            {shortcuts.answers.map((shortcut, index) => {
+              if (shortcut._key == null) { shortcut._key = Math.random(); }
```
srallen: Use strict equality.

app/components/shortcut-editor.jsx
```diff
@@ -0,0 +1,130 @@
...
+  constructor(props) {
+    super(props);
+    this.addAnswer = this.addAnswer.bind(this);
+  }
+
+  toggleShortcut(e) {
+    let nextTaskID;
+
+    if (e.target.checked) {
+      const taskCount = Object.keys(this.props.workflow.tasks).length;
```
srallen: `Uncaught TypeError: Cannot read property 'props' of undefined` when trying to use the checkbox button on a test project.

app/components/shortcut-editor.jsx
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  render() {
+    let shortcuts;
+    if (this.props.task) {
+      shortcuts = this.props.workflow.tasks[this.props.task.unlinkedTask];
+    }
+    const handleChange = handleInputChange.bind(this.props.workflow);
+    const children = React.Children.map(this.props.children, (child) => {
+      React.cloneElement(child);
```
wgranger: I forgot to return here, is why the original task isn't appearing.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+const emptyTask = {
+  question: 'What is it',
+  type: 'single'
+};
+
+describe('ShortcutEditor', function () {
+  describe('shortcutRendering', function () {
+    let wrapper;
+
+    beforeEach(function () {
```
srallen: I think this can just be a `before()` call. The set of tests in this describe block aren't interacting with the component, just checking the expected render.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      wrapper = mount(<ShortcutEditor task={task} workflow={workflow} />);
```
srallen: `shallow` or `render` is probably a better choice to use here since you're only checking the render and don't need the full React lifecycle available. Once we start getting a lot of the code tested, using `mount` when not actually needed will slow down the test run. Good explanation of the differences here: https://github.com/airbnb/enzyme/issues/465#issuecomment-227697726

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+
+const emptyTask = {
+  question: 'What is it',
+  type: 'single'
+};
+
+describe('ShortcutEditor', function () {
+  describe('shortcutRendering', function () {
```
srallen: This can just be `describe('rendering', function() {` or whatever makes sense for the block of tests. The describe statements should be readable like the it statements.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      const textarea = wrapper.find('textarea').first();
+      assert.equal(textarea.text(), 'Nothing Here');
+    });
+
+    it('should display the correct number of shortcuts', function () {
+      assert.equal(wrapper.find('textarea').length, 2);
+    });
+  });
+
+  describe('emptyRendering', function () {
```
srallen: same comment about readable describe statement.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    let wrapper;
+
+    beforeEach(function () {
+      wrapper = mount(<ShortcutEditor task={emptyTask} workflow={emptyWorkflow} />);
```
srallen: Same comment about use of mount.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+
+    it('should display the correct number of shortcuts', function () {
+      assert.equal(wrapper.find('textarea').length, 2);
+    });
+  });
+
+  describe('emptyRendering', function () {
+    let wrapper;
+
+    beforeEach(function () {
```
srallen: Same comment about using `before` here.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+    });
+
+    it('should call toggleShortcut with an input toggle', function () {
+      const toggleStub = sinon.stub(wrapper.instance(), 'toggleShortcut');
+      wrapper.update();
+      wrapper.find('input').simulate('change');
+      sinon.assert.called(toggleStub);
+    });
+
+    it('should remove the correct shortcut when deleted', function () {
```
srallen: This isn't really testing the removal, but rather that the handler is firing. I think it'll be clearer if described like the previous test `it('should call remove handler when clicked')` or something like that.

app/classifier/tasks/shortcut-editor.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      const removeStub = sinon.stub(wrapper.instance(), 'removeChoice');
+      const removeButton = wrapper.find('button.workflow-choice-remove-button').first();
+      wrapper.update();
+      removeButton.simulate('click');
+      sinon.assert.called(removeStub);
+    });
+
+    it('should add a default shortcut when expanded', function () {
```
srallen: Same comment as about about test description.

app/classifier/tasks/shortcut.jsx
```diff
@@ -0,0 +1,148 @@
...
+
+    if (this.props.annotation.value) {
+      answer = (
+        <div className="answer">
+          <i className="fa fa-check-circle-o fa-fw" />
+          <Markdown tag="span" inline={true}>{this.props.task.answers[this.props.annotation.value].label}</Markdown>
+        </div>
+      );
+    } else {
+      answer = <div className="answer">No answer</div>;
```
srallen: In the dev classifier as well as in a test project, when I click the shortcut option and click done, the summary I see is 'No answer'

app/classifier/tasks/shortcut.jsx
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  toggleShortcut(i, shortcut, e) {
+    if (e.target.checked) {
+      this.props.annotation.shortcut = { index: i };
+      this.setState({ index: i });
+    } else {
+      delete (this.props.annotation.shortcut);
+      this.setState({ index: null });
+      this.props.classification.update('annotations');
```
srallen: FYI, @eatyourgreens asked a question about the use of the classification resource rather than the onChange handler that other tasks use: https://github.com/zooniverse/Panoptes-Front-End/pull/3522#discussion_r104884244

CKrawczyk:
FYI there have been some changes to the shortcut task to make it work with the new task viewer rewrite: https://github.com/zooniverse/Panoptes-Front-End/pull/3571

wgranger:
An [issue](https://github.com/zooniverse/Panoptes/issues/2228) has been opened on the backend to handle shortcut exports with multiple answers.

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+  addAnswer() {
+    this.props.workflow.tasks[this.props.task.unlinkedTask].answers.push({ label: 'Nothing Here' });
+    this.props.workflow.update('tasks');
+  }
+
+  removeChoice(index) {
+    if (this.props.task) {
+      const answers = this.props.workflow.tasks[this.props.task.unlinkedTask].answers;
+      answers.splice(index, 1);
+      if (!answers.length) {
```
srallen: I'd prefer an explicit comparison here rather than a truthy value, `if (answers.length === 0)` should work.

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+      return React.cloneElement(child);
+    });
+
+    return (
+      <div>
+
+        {children}
+        <hr />
+
+        <label htmlFor="shortcut" title="Shortcut Options to End Classification">
```
srallen: This `htmlFor` isn't pointing at an `id`. You'll need to add `id="shortcut"` on to the input so the browser knows what this is referencing.

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+          <AutoSave resource={this.props.workflow}>
+            <span className="form-label">Shortcut Option</span>{' '}
+            <input type="checkbox" checked={shortcuts != null} onChange={this.toggleShortcut} />
```
srallen: Use strict equality whenever possible in javascript. Loose equality, `==` and `!=` can have unexpected results because it does type conversion: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Equality_comparisons_and_sameness

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+    }
+    const handleChange = handleInputChange.bind(this.props.workflow);
+    const children = React.Children.map(this.props.children, (child) => {
+      return React.cloneElement(child);
+    });
+
+    return (
+      <div>
+
+        {children}
```
srallen: Maybe I'm missing it, but why does the `ShortcutEditor` need to wrap the `TaskEditorComponent`?

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+    let nextTaskID;
+
+    if (e.target.checked) {
+      const taskCount = Object.keys(this.props.workflow.tasks).length;
+
+      let taskIDNumber = -1;
+
+      while (!((nextTaskID != null) && !(nextTaskID in this.props.workflow.tasks))) {
+        taskIDNumber += 1;
+        nextTaskID = `T${taskCount + taskIDNumber}`;
```
srallen: Lines 17-23 look to be identical to `addNewTask` in the workflow lab page: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/workflow.cjsx#L525-L541

Is there a reason why you're not just passing `addNewTask` down as a prop and using that? You can rewrite your handler to call `addNewTask` as well as the couple of Shortcut specific lines, I think just line 29 and line 31?

app/classifier/tasks/shortcut-editor.jsx
```diff
@@ -0,0 +1,136 @@
...
+
+        <small className="form-help">
+          Give volunteers the choice to skip to the end
+          of a classification if one of the following options is selected.
+        </small>
+
+        {shortcuts && (
+          <div className="workflow-task-editor-choices">
+            {shortcuts.answers.map((shortcut, index) => {
+              if (shortcut._key == null) { shortcut._key = Math.random(); }
```
srallen: Use strict equality. See note online 75 about why.

wgranger:
Oh yes. I'm really glad you brought that up. I noticed the issues related to that Panoptes [PR](https://github.com/zooniverse/Panoptes/issues/2212). Good it is being mentioned here.

app/classifier/tasks/shortcut-editor.jsx
wgranger: We decided to use this [Decorator pattern](https://github.com/zooniverse/Panoptes-Front-End/pull/2914#issuecomment-243503256) when shortcut was reviewed.

app/classifier/tasks/shortcut-editor.jsx
wgranger: So, if `addNewTask` is bound to the parent and passed down as a prop, then the task prop in the shortcut-editor will change when the function is called and line 29 in shortcut-editor (`this.props.task.unlinkedTask = nextTaskID;`) will be set on the shortcut task rather than the original task it needs to be linked to. I tried passing the function itself as a prop (`addNewTask={this.addNewTask}`), but the functionality on the child component was being wonky.

srallen: Ok, could you make a note about DRYing this up in the classifier rewrite issue #3527? Let's tackle it again later.

wgranger: Done

PR #3533, opened 2017-03-02
Modal focus

eatyourgreens:
Extracts modal keyboard behaviour from `Dialog` into a reusable `ModalFocus` component.

Adds ModalFocus to drawing details forms as a test. It should also work with the sign-in/register forms, which use the Dialog component.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://modal-focus.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -1,5 +1,6 @@
 React = require 'react'
 StickyModalForm = require 'modal-form/sticky'
+ModalFocus = require('../../components/modal-focus').default
 
 STROKE_WIDTH = 1.5
 SELECTED_STROKE_WIDTH = 2.5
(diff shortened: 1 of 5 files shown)


simensta:
Is it possible for the first input to receive focus when the modal opens, instead of the "Ok" button? 
drawing tool sub task example

eatyourgreens:
Sure, it should focus whichever element the task assigns autofocus to.

app/classifier/drawing-tools/root.cjsx
```diff
@@ -62,14 +63,16 @@ module.exports = React.createClass
...
-            <button type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
-          </p>
+          <ModalFocus onEscape={@handleDetailsFormClose}>
+            {for detailTask, i in toolProps.details
+              detailTask._key ?= Math.random()
+              TaskComponent = tasks[detailTask.type]
+              <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
+            <hr />
+            <p style={textAlign: 'center'}>
+              <button autoFocus type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
```
simensta: I think the autoFocus on the "OK" button is overriding the autoFocus specified for the textarea in [text subTasks that are not required](https://modal-focus.pfe-preview.zooniverse.org/projects/scribe-2/project-state-testing/classify).

eatyourgreens: I remember now: question subtasks, like the ones in I Fancy Cats, don't autofocus so I added autofocus to the button instead.

eatyourgreens: > Only one form element in a document can have the `autofocus` attribute…
https://developer.mozilla.org/en/docs/Web/HTML/Element/input
which makes sense but makes this a bit more complicated than I thought.

eatyourgreens: I've added a workaround which checks the task type before autofocusing the submit button. It's not a great piece of code but it should work for the current task setup.

app/classifier/tasks/generic.jsx
```diff
@@ -17,6 +17,7 @@ export default class GenericTask extends React.Component {
         <Markdown className="classification-task-help">
           {this.props.help}
         </Markdown>
+        <button autoFocus className="standard-button" onClick={this.reject}>Close</button>
```
simensta: The "Close" button isn't working for me, looks like `this.reject` is missing.

eatyourgreens: Should be fixed now. Hooray for undocumented helper functions!

Issue #3492, opened 2017-02-16
Refactor media fetch for tutorials, field guide, mini-course, etc


srallen:
## Expected behavior
Linked media for any tutorials or field guides should be fetched for once rather than each time they are opened.
 
## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Field guide, tutorial, and mini-course fetches for media each time they're opened: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/tutorial.cjsx#L45-L55
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/mini-course.cjsx#L29-L39
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/potential-field-guide.cjsx#L33-L37

simensta:
I am working on the field guide and @mcbouslog is working on the tutorial and the mini-course.

mcbouslog:
@srallen and @simensta - I think we can close this issue?

srallen:
Yep

PR #3535, opened 2017-03-02
Optimize field guide

simensta:
Addresses the field guide portion of #3492. 

**Describe your changes.**
I added the parameter `include: 'attached_images'` when requesting the field guide resource. This eliminates the need to fetch the resources for the field guide icons (i.e., attached_images) in a later request. 

Also, I added the `displayName` property to the PotentialFieldGuide, FieldGuide, and ProjectPage.

# Review Checklist
- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/components/field-guide.cjsx
+++ b/app/components/field-guide.cjsx
@@ -3,6 +3,8 @@ React = require 'react'
 CroppedImage = require './cropped-image'
 
 module.exports = React.createClass
+  displayName: 'FieldGuide'
+  
   getDefaultProps: ->
     title: null
     content: null
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -42,6 +42,8 @@ SOCIAL_ICONS =
 AVATAR_SIZE = 100
 
 ProjectPage = React.createClass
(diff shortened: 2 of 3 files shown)


simensta:
Good catch. After adding more than ~20 articles to a field guide the icon images fail to load. Also related, deleting an article does not cause the associated icon id to be deleted from `link.attached_images.ids` for the field guide.

srallen:
The Gravity Spy field guide has 20 items, so I would think it would work, but it sounds like we're hitting a page size issue. With this in mind, we might not be able to use the include in the original query if we need to deal with pagination. I'm not sure about the links, but that sounds like a bug or maybe it's done in a sidekiq worker so it's not immediate.

srallen:
I am wondering if the best solution for this is not to make it an included resource request, but request for the field guide in the classifier like the tutorial and chain the get request for the included media with an increased page size, then store it in the classifier state.

simensta:
Good point. In addition, since the field guide now only appears on the '/classify' page, it makes sense for the field guide request and component to move to the classifier. I'll try that approach.

simensta:
Closing in favor of #3574.

PR #3538, opened 2017-03-03
make the drawing task work on a combo task

CKrawczyk:
This PR makes drawing tasks work within a combo task (needed for @tingard's custom project).  Since this touches the `combo` task, `frame-annotator`, and `makings-renderer` this should be reviewed carefully to make sure it does not break existing projects.  So far I have only tested the `dev-classifier`.

With these changes the crop task mostly works in the combo task, but does causes errors when the `next` button is pressed.  Because of this I added the `not fully supported` message to the combo task.  Additionally since both the `crop` and `drawing` tasks have an annotation layer, they can't both be in the same combo task, so I removed `crop` from the `dev-classifier`s combo task.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://drawing-combo-task.pfe-preview.zooniverse.org/

## Optional


app/classifier/tasks/combo/index.cjsx
```diff
@@ -71,13 +71,28 @@ ComboTask = React.createClass
...
       <g className="combo-task-persist-inside-subject-container">
-        {Object.keys(props.taskTypes).map (taskType) ->
+        {Object.keys(props.taskTypes).map (taskType, idx) ->
           unless taskType is 'combo'
             TaskComponent = props.taskTypes[taskType]
             if TaskComponent.PersistInsideSubject?
               fauxClassification =
                 annotations: allComboAnnotations
-              <TaskComponent.PersistInsideSubject key={taskType} {...props} classification={fauxClassification} />}
+                update: () => props.classification.update(arguments)
```
eatyourgreens: Would `update: props.classification.update` work here?

CKrawczyk: I tried that at first and got `Uncaught TypeError: Array.prototype.indexOf called on null or undefined` errors when a drawn point is moved/updated (I think this error has to do with the proper `this` being passed into `update`).

app/classifier/frame-annotator.jsx
```diff
@@ -110,12 +110,18 @@ export default class FrameAnnotator extends React.Component {
...
     const { type, src } = getSubjectLocation(this.props.subject, this.props.frame);
     const createdViewBox = `${this.props.viewBoxDimensions.x} ${this.props.viewBoxDimensions.y} ${this.props.viewBoxDimensions.width} ${this.props.viewBoxDimensions.height}`;
 
     if (this.props.annotation) {
       taskDescription = this.props.workflow.tasks[this.props.annotation.task];
       TaskComponent = tasks[taskDescription.type];
+      // combo tasks contain many task types
+      taskTypeList = [taskDescription.type];
+      if (taskDescription.type === 'combo') {
+        taskTypeList = taskDescription.tasks.map((task) => { return this.props.workflow.tasks[task].type; });
```
rogerhutchings: Can be simplified to:

```js
taskTypeList = taskDescription.tasks.map(task => this.props.workflow.tasks[task].type);
```

CKrawczyk: The `.eslintrc` currently has `"arrow-body-style": ["warn", "always"]` instead of `as-needed` or `never`, so I assumed we were moving away from the simplified syntax (but I can change it if you want).

srallen: I think that rule was set early on before we started to do consensus on what the rules should be. I'd be open to having that one change.

CKrawczyk: I updated the `eslint` rule to `as-needed` and used the simplified version of the callback.

CKrawczyk:
This works just fine in the dev classifier but does not work on a staged project: https://drawing-combo-task.pfe-preview.zooniverse.org/projects/tingard/combo-drawing-test/classify

Not really sure way it works for one but not the other.

eatyourgreens:
NB that test project isn't public on staging, so requires admin to be turned on.

CKrawczyk:
It is public now

CKrawczyk:
I tracked down the bug with the staged workflow, turns out I was using the wrong index value when searching for the annotation to pass down to the `PersistInsideSubject`.  This index just happened to line up correctly on the dev classifier.

eatyourgreens:
That sounds very similar to the bug behind #3581

app/classifier/tasks/combo/index.cjsx
eatyourgreens: Should this be `props.classification.update()`?

CKrawczyk:
Scratch the saving issue, that was due to a "self looping" workflow making the annotations look funny, with a more realistic workflow the classifications save just fine.

CKrawczyk:
The `PrrsistInsideSubject` section is getting bit complicated for it to handle multiple combo tasks each with a drawing task.  Turns out if things are not set up correctly the shapes drawn in previous tasks will disappear and the `isAnnotationComplete` function fail.

CKrawczyk:
Also closing #3628 so combo summary show the summary for each component (this can be split into its own PR if needed).

eatyourgreens:
Could this be handled by adding a renderer specifically for combo tasks to [/app/classifier/annotation-renderer/index.jsx](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/annotation-renderer/index.jsx)?

CKrawczyk:
Yes, I think adding a renderer for a combo task would help clean up this code.  I will start looking into a refactor.

eatyourgreens:
Maybe pass down `combo` as a type [here](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/frame-annotator.jsx#L91). I don't think that will break anything.

CKrawczyk:
Turns out the solution is easier than that, just added `AnnotationRenderer: SVGRenderer` to the combo task.  The `InsideSubject` and `PersistInsideSubject` components already takes care of formatting everything.  The only issue I can see happening is if you try to combo an SVG renderer task with an html renderer task.

eatyourgreens:
Go me! 👍 I was hoping the refactor of the annotation components would make this easier, but didn't realise it would be this easy.

I think there are some combinations of tasks that might not work, but I'm not sure if those are combinations that would be used in a real project. I'd be inclined to ignore this unless someone has an actual case for supporting it.

For both types of rendering to work, the combo task would have to be able to generate either HTML or SVG from its `InsideSubject` hooks, based on the current frame type.

CKrawczyk:
It looks like a few tests are failing, I will see if I can figure out what is causing that.  At the moment this will peg the combo task as `SVGRenderer` only.  It would be nice if the tasks in the combo determined the `AnnotationRenderer` but static class methods can't be set dynamically (as far as I know).

eatyourgreens:
For that to work, I think you would have to write a new renderer, which loops through an array of annotations then invokes the specific renderer for each annotation in turn, based on the current frame type.

The individual renderers use the annotation that's passed in as a prop to check whether to use the SVG or text components, so I don't think you would have to add any new checks.

That should allow you to remove some of the new code from the rendering methods inside the combo task itself, by delegating to existing component methods.

eatyourgreens:
Re. the failing tests, one is failing because `MarkingsRenderer` is being rendered twice.
```
<PersistInsideSubject taskTypes={{...}} workflow={{...}} task={{...}} classification={{...}} annotation={{...}} frame={0} scale={{...}} naturalWidth={0} naturalHeight={0} containerRect={{...}} getEventOffset={[Function]} onChange={[Function]} preferences={[undefined]} normalizeDifference={[Function]} getScreenCurrentTransformationMatrix={[Function]}>
  <g className="combo-task-persist-inside-subject-container">
     <MarkingsRenderer taskTypes={{...}} workflow={{...}} task={{...}} classification={{...}} annotation={{...}} frame={0} scale={{...}} naturalWidth={0} naturalHeight={0} containerRect={{...}} getEventOffset={[Function]} onChange={[Function]} preferences={[undefined]} normalizeDifference={[Function]} getScreenCurrentTransformationMatrix={[Function]}>
             <g />
     </MarkingsRenderer>
   </g>
</PersistInsideSubject>
<MarkingsRenderer taskTypes={{...}} workflow={{...}} task={{...}} classification={{...}} annotation={{...}} frame={0} scale={{...}} naturalWidth={0} naturalHeight={0} containerRect={{...}} getEventOffset={[Function]} onChange={[Function]} preferences={[undefined]} normalizeDifference={[Function]} getScreenCurrentTransformationMatrix={[Function]}>
  <g />
</MarkingsRenderer>
```

eatyourgreens:
The other test is failing because `props.annotation` is `{}` on [line 96 of the combo task](/zooniverse/Panoptes-Front-End/pull/3538/files#diff-bb5b836390cff5322dc3f0b13c12df09R96).

CKrawczyk:
I already tried looping over each annotation in the combo task and making a new `AnnotationRenderer` for each, but that does not work since each `SVGRenderer` re-draws the subject in SVG for each task in the combo.  This ends up drawing the image above the annotations.

Maybe this could work if the `SVGRenderer` was changed to accept the `InsideSubject` and `PersistInsideSubject` as children instead of grabbing them in the component.

eatyourgreens:
My bad in choosing the name I think. The `AnnotationRenderer` is written to be invoked once for a subject then render all the hooks inside the subject. I should have called it `ClassificationRenderer` or `AnnotationsRenderer` or something like that.

Possibly it needs to be invoked with the fake workflow and classifications that you've generated, so as to override the tasks and annotations it would otherwise get from the workflow and classifications.

eatyourgreens:
Looks good to me. The API for task hooks could be better designed, so that we don't have to use the `fauxAnnotation` and `fauxClassification` hack to render the tasks within the combo task, but I think that's well outside the scope of this PR.

I'd appreciate a second opinion from @rogerhutchings or @srallen before merging, since this changes the code for all SVG drawing tasks.

srallen:
I can maybe take a look at this on Friday.

app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -18,9 +18,8 @@ module.exports = React.createClass
     # console.log 'Old marks were', @state.oldSetOfMarks
     newSetOfMarks = []
     # Automatically select new marks.
-    annotations = nextProps.classification?.annotations ? []
-    annotation = annotations[annotations.length - 1]
-    if annotation?
+    annotation = nextProps.annotation
```
eatyourgreens: Could you explain what this change does here? Looks like the previous code took the last annotation from the `classification.annotations` array. Is `nextProps.annotation` guaranteed to be that same annotation?

app/classifier/tasks/drawing/markings-renderer.cjsx
```diff
@@ -36,7 +35,7 @@ module.exports = React.createClass
   render: ->
     skippedMarks = 0
     <g>
-      {for annotation in @props.classification?.annotations ? []      
+      {for annotation in @props.classification?.annotations ? []
```
eatyourgreens: We should probably change this component to take an array of annotations as a prop. Then I don't think you would have to create a fake classification just to pass in annotations to be rendered.

app/classifier/tasks/combo/index.cjsx
```diff
@@ -65,20 +67,48 @@ ComboTask = React.createClass
...
+                  # if the current annotation is for the combo task pass in the `inner` annotations
+                  fauxAnnotation = allComboAnnotations[idx]
+                else
+                  fauxAnnotation = props.annotation
+              else
+                fauxAnnotation = props.annotation
+              <TaskComponent.PersistInsideSubject
+                key={taskType}
+                {...props}
+                onChange={fauxChange}
```
eatyourgreens: Marks can only be changed during the current task (`InsideSubject`) so why does `PersistInsideSubject` have an `onChange` handler?

srallen:
Just FYI, I'm checking to make sure existing use isn't broken. NfN is one that uses combo tasks. Looks ok: https://drawing-combo-task.pfe-preview.zooniverse.org/projects/zooniverse/notes-from-nature/classify?env=production&reload=1&workflow=3978

app/classifier/tasks/drawing/markings-renderer.cjsx
CKrawczyk: This was changed because taking the last `classification.annotations` will give a "combo annotation" (an annotation that has a `value` as an array of `annotations`), but `nextProps.annotation` will give the inner annotation only (i.e. just the drawing annotation).

app/classifier/tasks/combo/index.cjsx
CKrawczyk: That is a good question, but it seems `markings-renderer.cjax` expects it: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/markings-renderer.cjsx#L92

CKrawczyk:
I have updated the editor to allows for adding a drawing task.  I also updated the staged link.

app/classifier/tasks/combo/index.cjsx
CKrawczyk: Note: removing the `onChange` prop from the `markings-renderer.cjsx` breaks the `onDrag` of all drawing tasks.  It looks like the `PersistInsideSubject` handles annotations after they are initialized with `InsideSubject`. (It has quite a misleading name).

app/classifier/tasks/drawing/markings-renderer.cjsx
eatyourgreens: Thanks. I was more wondering if this still works with drawing tasks, which are presumably passing annotations in via the classification. So I was asking if this change is backwards compatible with the existing code.

app/classifier/tasks/combo/index.cjsx
eatyourgreens: Cheers. Yeah, the task hooks API is pretty counterintuitive, in my opinion.

app/classifier/tasks/drawing/markings-renderer.cjsx
CKrawczyk: For every case I have tested it works.  Since the active task always appends to the end of the `classification.annotations` list, there should be no instance where the `annotation` passed on `props` does not match the last item in the classification.

eatyourgreens: I had a search through the code and the annotation is passed down here, so that's fine. https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/classifier.jsx#L216

`MarkingRenderer` shouldn't have been using `classification.annotations` to get something that was already in the props.

eatyourgreens:
I'm happy with this if @srallen is also happy.

I think it has also flagged up that the hooks interface for tasks is confusing, and in need of an overhaul. I think that's already being discussed on the [classifier RFC](/zooniverse/rfc/issues/12).

app/classifier/tasks/combo/index.cjsx
eatyourgreens: Unrelated to this PR, but I am now wondering why marks aren't draggable across all tasks in a workflow, if dragging is handled by `PersistInsideTask`.

rogerhutchings:
Just another vote for redesigning the drawing API - https://github.com/zooniverse/Panoptes-Front-End/issues/3560

Using statics is a bit nuts, but I think it's a "legacy" React issue that can be worked around. @eatyourgreens and I talked about a more OO-approach, using redux to store the canonical data and creating throwaway objects to manipulate classifications / annotations / anything else that doesn't have a set of handy methods on its prototype that results in code duplication

app/classifier/tasks/combo/index.cjsx
CKrawczyk: Marker selection is handled by `InsideSubject` and you can't drag unless a mark is selected. (as far as I can tell, I could be reading the code wrong...)

eatyourgreens: You're right. The drawing hooks code is kind of crazy.

eatyourgreens: Or at least, counterintuitive in that the renderer does more than just render marks.

app/classifier/tasks/combo/summary.jsx
```diff
@@ -0,0 +1,35 @@
...
+    <div>
+      {body}
+    </div>
+  );
+};
+
+ComboTaskSummary.propTypes = {
+  annotation: React.PropTypes.shape(
+    { value: React.PropTypes.array }
+  ).isRequired,
```
srallen: Since this is required, this should have a default prop defined. If something goes wrong with the parent component and annotation is undefined, this component will break without the default prop.

eatyourgreens: I've found that even with a default prop defined for annotations, higher level classifier components can pass down the annotation as `null`, which will override the default value and break things anyway.

This usually happens on first render, at the beginning of a task, when we are waiting for the annotation to initialise.

srallen: Good point and I think recently there was a bug because of this if I'm remembering correctly?

eatyourgreens: That sounds familiar. I remember mentioning to @mcbouslog we should be careful with props because `null` is not the same as `undefined`, so it might have come up in Notes From Nature.

PR #3541, opened 2017-03-03
Refactor tutorial media requests

mcbouslog:
Addresses the tutorial and mini-course portion of #3492.

- adds `include: 'attached_images'` to tutorial/mini-course request
- on start checks if there are `attached_images`, else resolves Promise
- gets `attached_images` based on `include` cache in apiClient, not stringed `.get` on tutorial/mini-course resource like previously

this [test project](https://refactor-tutorial-media-requests.pfe-preview.zooniverse.org/projects/markb-panoptes/nfn-test-mb)'s workflows 1 and 2 have separate tutorials and mini-courses with media, workflow 3 has no media. 

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://refactor-tutorial-media-requests.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/components/mini-course.cjsx
+++ b/app/components/mini-course.cjsx
@@ -17,7 +17,7 @@ module.exports = React.createClass
     find: (workflow) ->
       # Prefer fetching the tutorial for the workflow, so we know which one to fetch if multiple exist.
       if workflow?
-        apiClient.type('tutorials').get workflow_id: workflow.id, kind: "mini-course"
+        apiClient.type('tutorials').get workflow_id: workflow.id, kind: "mini-course", include: 'attached_images'
           .then ([minicourse]) ->
             minicourse
       else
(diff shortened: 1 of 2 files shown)


app/components/mini-course.cjsx
```diff
@@ -17,7 +17,7 @@ module.exports = React.createClass
     find: (workflow) ->
       # Prefer fetching the tutorial for the workflow, so we know which one to fetch if multiple exist.
       if workflow?
-        apiClient.type('tutorials').get workflow_id: workflow.id, kind: "mini-course"
+        apiClient.type('tutorials').get workflow_id: workflow.id, kind: "mini-course", include: 'attached_images'
```
eatyourgreens: Should this be `include: ['attached_images']`?

app/components/tutorial.cjsx
```diff
@@ -16,7 +16,7 @@ module.exports = React.createClass
     find: (workflow) ->
       # Prefer fetching the tutorial for the workflow, if a workflow is given.
       if workflow?
-        apiClient.type('tutorials').get workflow_id: workflow.id
+        apiClient.type('tutorials').get workflow_id: workflow.id, include: 'attached_images'
```
eatyourgreens: Should this be `include: ['attached_images']`?

srallen:
FYI, I made an issue on the json-api-client about the inconsistent syntax for includes: https://github.com/zooniverse/json-api-client/issues/35

app/components/tutorial.cjsx
```diff
@@ -42,15 +42,18 @@ module.exports = React.createClass
...
-        awaitTutorialMedia = tutorial.get 'attached_images'
-          .catch ->
-            # Checking for attached images throws if there are none.
-            []
-          .then (mediaResources) ->
-            mediaByID = {}
-            for mediaResource in mediaResources
-              mediaByID[mediaResource.id] = mediaResource
-            mediaByID
+        if tutorial.links.attached_images.ids?.length
```
mcbouslog: I didn't realize this until just now, but apparently whether a tutorial never has or had all the attached images deleted, if there are no attached images it looks like:
```
links
    attached_images
        href: "/tutorials/106/attached_images"
        type: "attached_images"
```
notably there's no `id` attribute at all, let alone the empty array I was expecting at minimum if the media was deleted.

so if I add the `isnt 0` I think the conditional gets met because `undefined` isnt `0` and it makes the request I don't want it to, but I'm not sure I'm interpreting things correctly.

maybe just `if tutorial.links.attached_images.ids?` would be preferred?

srallen: Ok, how about `if tutorial.links.attached_images.ids? and tutorial.links.attached_images.ids.length isnt 0`? so checks for existence and that the array has length.

mcbouslog: LGTM; I updated accordingly.

zwolf:
The db was missing an index that this should have been using: https://github.com/zooniverse/Panoptes/pull/2226

zwolf:
Timeout issue solved by https://github.com/zooniverse/Panoptes/pull/2226, as you were

mcbouslog:
@zach I watched as it appeared to work great yesterday, but when I point to Gravity Spy from the staged branch using `?env=production` like @srallen's comment above mentions I still get a 504 on the tutorial. are you guys seeing the same? do I need to redeploy this staged PFE branch post backend update?

zwolf:
The change isn't deployed on the production database yet, which you are using by applying that query param. Should go live today.

PR #3544, opened 2017-03-06
Fix classifier mount

### LINKED ISSUE #3543: Classifier mounts on subject change
## Expected behavior
Classifier component should mount, and initialise, when the classify page first loads. Subject changes should be handled by `componentWillReceiveProps` or `componentDidUpdate`.

## Current behavior
`ClassifierPageWrapper` unmounts then remounts `Classifier` between subjects, causing any initialisation code (eg. API calls) in classifier components to run for every subject. I'd expect that initialisation should only happen if the project or workflow change.

## Steps to replicate
Log calls to `Classifier.componentWillMount` in the console while classifying a series of subjects on a given project.

## Additional information
This bug is the underlying causing of user collections being re-fetched from the API for every subject, even though they are stored in component state on fetch.

eatyourgreens:
Fixes #3543  .

Describe your changes.
Mounts the Classifier once, then reacts to subject and classification changes when props change.

There could be components that relied on the old behaviour, so there may still be component code, dotted through the classifier components, that needs to be changed to run on `componentWillReceiveProps` or `componentDidUpdate` as well as on `componentWillMount` or `componentDidMount`.

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://fix-classifier-mount.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/frame-annotator.jsx
+++ b/app/classifier/frame-annotator.jsx
@@ -28,6 +28,9 @@ export default class FrameAnnotator extends React.Component {
     if (nextProps.annotation !== this.props.annotation) {
       this.handleAnnotationChange(this.props.annotation, nextProps.annotation);
     }
+    if (nextProps.subject !== this.props.subject) {
+      this.setState({ alreadySeen: nextProps.subject.already_seen || seenThisSession.check(nextProps.workflow, nextProps.subject) });
+    }
   }
 
   getSizeRect() {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Playing around with [I Fancy Cats](https://fix-classifier-mount.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats), I'm not 100% sure that I've got the subject seen status right for new subjects.

PR #3542, opened 2017-03-03
Move pages, avatar, and background requests to index promise

srallen:
Refactors the requests for the project pages, avatar, and background and moves them into the promise that is fetching other linked resources for the project. Also adds a request for project roles since there are some children components requesting for that too. The intention is to remove duplicate requests in children and just pass these down as props and used where needed. 

https://optimize-project-requests.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch?
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #3545, opened 2017-03-06
Project home reorganization, optimization

srallen:
This builds upon #3542. Project home has been put into its own folder with the index split into a class component handling the API requests and state and a functional view component. The subject requests for the talk subject previews is optimized to use the subject ids array instead of iterating through. This also leverages the new project avatar prop coming in from #3542. While refactoring the `<TalkImages />` component didn't seem useful, so I eliminated it and just brought in what it's rendering into the home view component.

Radio Meteor Zoo is actually using a researcher avatar:
https://optimize-home-requests.pfe-preview.zooniverse.org/projects/zooniverse/radio-meteor-zoo?env=production

Planet Four Ridges is using the project avatar:
https://optimize-home-requests.pfe-preview.zooniverse.org/projects/mschwamb/planet-four-ridges?env=production

This could use some test, but need to move on to rethinking the workflow fetch.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


srallen:
@wgranger just rebased and fixed the project completeness for the home page.

wgranger:
Nice. Thanks. I'll be getting to this shortly.

app/pages/project/home/index.jsx
```diff
@@ -0,0 +1,127 @@
...
+
+  fetchResearcherAvatar(props) {
+    if (props.project.configuration && props.project.configuration.researcherID) {
+      if (props.project.configuration.researcherID === props.project.display_name) {
+        if (props.projectAvatar && props.projectAvatar.src) {
+          this.setState({ researcherAvatar: props.projectAvatar.src });
+        }
+      }
+
+      apiClient.type('users').get(this.props.project.configuration.researcherID)
```
wgranger: Perhaps move this API call into the conditional to only occur if `props.projectAvatar` is not found? It looks like this call is being made even if the `researcherAvatar` state is being set above to a project avatar.

srallen:
@wgranger Ok I've fixed those two items. Should be good to go now.

wgranger:
Cool. Looks good.

PR #3548, opened 2017-03-07
Fix 'seen this session'

### LINKED ISSUE #2158: 404 error after classifying a subject that has multiple locations
Noticed this while classifying on Wildebeest Watch. If you classify a subject that has multiple images associated with it, the classification adds the subject ID multiple times to the links array, then the front-end tries to GET something like `/api/subjects/{subject_id},{subject_id},{subject_id}` where `subject_id` is the subject that you've just classified.

It looks like the number of repeated IDs in the classification links and the URL is set by the number of image frames in the current subject.

eatyourgreens:
Don't fetch the subject when we already have its ID.

Fixes #2158 .

Describe your changes.
Removes the subjects API call after classifying, since we already have the subject ID for the seens array.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-subject-seens.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/lib/seen-this-session.coffee
+++ b/app/lib/seen-this-session.coffee
@@ -6,9 +6,9 @@ auth.listen 'change', ->
   subjectsSeenThisSession.splice 0
 
 module.exports =
-  add: (workflow, subjects) ->
-    for subject in subjects
-      subjectsSeenThisSession.push "#{workflow.id}/#{subject.id}"
+  add: (workflow_id, subject_ids) ->
+    for subject_id in subject_ids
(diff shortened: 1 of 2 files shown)


eatyourgreens:
The ones I've looked at are just IDs. I don't think the API client stores full resources in those arrays, but I'm not 100% sure about that.

eatyourgreens:
Staged now at https://fix-subject-seens.pfe-preview.zooniverse.org/?env=production

PR #3552, opened 2017-03-07
Prevent inactive workflows on live projects from being set as default

srallen:
This disables the checkbox for setting a workflow as default for live projects and inactive workflows. I can't think of a scenario where a project builder would want to do this purposely and it also breaks the workflow selection causing an infinite loop attempt to fetching a workflow that isn't active. 

I also added `classnames` to manage this and the inline styles added in #3517 

https://inactive-default-workflows.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -14,6 +14,7 @@ FileButton = require '../../components/file-button'
 WorkflowCreateForm = require './workflow-create-form'
 workflowActions = require './actions/workflow'
 ShortcutEditor = require '../../components/shortcut-editor'
+classnames = require 'classnames'
 
 DEMO_SUBJECT_SET_ID = if process.env.NODE_ENV is 'production'
   '6' # Cats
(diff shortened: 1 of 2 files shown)


app/pages/lab/workflow.cjsx
```diff
@@ -247,10 +249,11 @@ EditWorkflowPage = React.createClass
...
 
-          <div>
+          <div className={disabledIfWorkflowInactive}>
             <AutoSave resource={@props.project}>
               <span className="form-label">Set as default</span><br />
               <small className="form-help">If you have more than one workflow, you can set which should be default. Only one can be default.</small>
+              {<span><br /><small className="form-help">Inactive workflows on live projects cannot be made default.</small></span> if projectLiveWorkflowInactive}
               <br />
               <label>
                 <input ref="defaultWorkflow" type="checkbox" checked={@props.project.configuration?.default_workflow is @props.workflow.id} onChange={@handleDefaultWorkflowToggle} />
```
eatyourgreens: Should this checkbox be disabled too?

srallen:
@eatyourgreens I've disabled the checkbox if true. Good catch.

PR #3553, opened 2017-03-07
Rewrite FrameViewer into ES6

wgranger:
Tests are also included. 

Describe your changes.
Just a rewrite for the FrameViewer into ES6. I kepy `refs` in because I'm curious if parents/children are expecting them.

https://frame-viewer-es6.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional


app/components/frame-viewer.jsx
```diff
@@ -0,0 +1,149 @@
...
+            <div className="subject-image-frame" >
+              <img
+                ref="subjectImage"
+                className="subject pan-active"
+                src={src}
+                style={SUBJECT_STYLE}
+                onLoad={this.handleLoad}
+                tabIndex={0}
+                onFocus={this.refs.panZoom ? this.refs.panZoom.togglePanOn : NOOP}
+                onBlur={this.refs.panZoom ? this.refs.panZoom.togglePanOff : NOOP}
```
wgranger: The conditionals in the original cjsx were a bit odd, so curious if using the `NOOP` here is the best bet for a default.

app/components/frame-viewer.jsx
```diff
@@ -0,0 +1,149 @@
...
+    if (this.props.onLoad) { this.props.onLoad(e, this.props.frame); }
+  }
+
+  render() {
+    let type;
+    let format;
+    let src;
+    const zoomEnabled = this.props.workflow.configuration.pan_and_zoom;
+    const FrameWrapper = this.props.frameWrapper;
+    ({ type, format, src } = getSubjectLocation(this.props.subject, this.props.frame));
```
srallen: I think these can just be directly set as constant variables, `const { type, format, src } = getSubjectLocation(this.props.subject, this.props.frame)`

app/components/frame-viewer.jsx
srallen: Use `() => {}` instead of `NOOP`

app/components/frame-viewer.jsx
```diff
@@ -0,0 +1,149 @@
...
+            <TextViewer src={src} type={type} format={format} frame={this.props.frame} onLoad={this.handleLoad} />
+          );
+        default:
+          return null;
+      }
+    })(type);
+
+    if (FrameWrapper) {
+      return (
+        <PanZoom ref="panZoom" enabled={zoomEnabled} frameDimensions={this.state.frameDimensions}>
```
srallen: String literal refs are being deprecated, can this instead use the recommended callback assignment? https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-string-refs.md

app/components/frame-viewer.jsx
```diff
@@ -0,0 +1,149 @@
...
+    { configuration: React.PropTypes.object }
+  )
+};
+
+FrameViewer.defaultProps = {
+  classification: {
+    annotations: []
+  },
+  frame: 0,
+  onChange: NOOP,
```
srallen: Use `() => {}` instead of `NOOP`

app/components/frame-viewer.spec.js
```diff
@@ -0,0 +1,55 @@
...
+
+const subject = {
+  locations: [{ image: 'www.testimage.com' }]
+};
+
+const videoSubject = {
+  locations: [{ video: 'www.testvideo.com' }]
+};
+
+const clickSimulate = {
```
srallen: This variable name doesn't really match what it is doing. Neither sinon nor the built in enzyme simulate method is being called here.

app/components/frame-viewer.spec.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    it('should load a wrapper when included', function () {
+      assert.equal(wrapper.find('FrameAnnotator').length, 1);
+    });
+  });
+
+  describe('if loading other subjects', function() {
+    it('should load a video subject correctly', function () {
+      const wrapper = mount(<FrameViewer subject={videoSubject} />);
```
srallen: I don't think `mount` needs to be used here since none of the React lifecycle methods are being tested. `render` or `shallow` would be better.

app/components/frame-viewer.spec.js
```diff
@@ -0,0 +1,55 @@
...
+      assert.equal(wrapper.find('VideoPlayer').length, 1);
+    });
+
+    it('should load nothing without a subject or wrapper present', function () {
+      const wrapper = mount(<FrameViewer />);
```
srallen: Same comment about use of `mount`

app/components/frame-viewer.spec.js
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import assert from 'assert';
+import { mount } from 'enzyme';
+import FrameViewer from './frame-viewer';
+import FrameAnnotator from '../classifier/frame-annotator';
+
+const subject = {
+  locations: [{ image: 'www.testimage.com' }]
```
srallen: Maybe you can leverage the images from the mock data from the dev classifier instead of using a fake url?

app/components/frame-viewer.spec.js
```diff
@@ -0,0 +1,55 @@
...
+};
+
+const videoSubject = {
+  locations: [{ video: 'www.testvideo.com' }]
```
srallen: Similarly, we probably should have a mock data video and text subject for use with the dev classifier.

srallen: I should note, adding those should come in a separate PR.

app/components/frame-viewer.jsx
```diff
@@ -0,0 +1,149 @@
+import React from 'react';
+import LoadingIndicator from '../components/loading-indicator';
+import getSubjectLocation from '../lib/get-subject-location';
+import VideoPlayer from './video-player';
+import PanZoom from './pan-zoom';
+import TextViewer from './text-viewer';
+
+const SUBJECT_STYLE = { display: 'block' };
```
srallen: Since this style isn't changing because of a conditional, can you convert it to a class instead?

wgranger: Define it within the class?

srallen: I mean as a CSS class.

wgranger:
@srallen Do you have any idea why that test might be failing in Travis? It seems to pass fine locally. I tried to push up a commit to fix the test, but the commit didn't change much.

srallen:
@wgranger could you rebase your branch and update the PR?

srallen:
@wgranger could you do one last rebase again? Lots of changes have come in with the classifier and bug fixes since SGL. 

Re the tests, be sure that your branch is up to date with the latest master. The test script for travis changed while you were working on this before.

wgranger:
This branch has been rebased and restaged.

PR #3556, opened 2017-03-07
Optimize all workflows request

srallen:
Builds on #3545. This refactors the workflow fetch to be in the project page controller component. I still need to do some work on it, but wanted some initial thoughts. I wanted to eliminate as much as possible fetching for all active workflows that has an arbitrary page size number. For getting a random workflow, I've decided to just pick out of the `project.links.workflows` array, trying to fetch it with the `active: true` param and if it comes back undefined, try again. But I'm not sure if this is better than current behavior. 

https://optimize-all-workflows-request.pfe-preview.zooniverse.org

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/home/index.jsx
+++ b/app/pages/project/home/index.jsx
@@ -20,7 +20,7 @@ export default class ProjectHomeContainer extends React.Component {
   }
 
   componentDidMount() {
-    this.showWorkflowButtons();
+    this.showWorkflowButtons(this.props, this.context);
     this.fetchResearcherAvatar(this.props);
     this.fetchTalkSubjects(this.props);
   }
@@ -35,7 +35,7 @@ export default class ProjectHomeContainer extends React.Component {
     }
   }
 
-  fetchAllWorkflows(props = this.props, query = { active: true, fields: 'active,completeness,configuration,display_name' }) {
+  fetchAllWorkflows(props, query) {
     if (this.state.activeWorkflows.length === 0) {
       getWorkflowsInOrder(props.project, query)
         .then((activeWorkflows) => {
@@ -78,12 +78,12 @@ export default class ProjectHomeContainer extends React.Component {
     }
   }
 
-  showWorkflowButtons(props = this.props, context = this.context) {
+  showWorkflowButtons(props, context) {
     const workflowAssignment = this.props.project.experimental_tools.includes('workflow assignment');
 
     if ((props.project.configuration && props.project.configuration.user_chooses_workflow && !workflowAssignment) ||
       (workflowAssignment && context.user)) {
-      this.setState({ showWorkflowButtons: true }, this.fetchAllWorkflows);
+      this.setState({ showWorkflowButtons: true }, this.fetchAllWorkflows.bind(this, this.props, { active: true, fields: 'active,completeness,configuration,display_name' }));
     } else {
       this.setState({ showWorkflowButtons: false });
     }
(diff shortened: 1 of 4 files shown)


srallen:
I've rebased and updated the staging link. I think I'll give writing a few useful tests a try because I'm not confident I'm not breaking something somewhere that I didn't think to check.

app/pages/project/home/index.jsx
```diff
@@ -33,6 +35,15 @@ export default class ProjectHomeContainer extends React.Component {
     }
   }
 
+  fetchAllWorkflows(props = this.props, query = { active: true, fields: 'active,completeness,configuration,display_name' }) {
```
eatyourgreens: Are default params ok to use here? MDN says no support in IE or Opera. https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Default_parameters

srallen:
I removed the default params. I was unaware of the lack of support in those browsers. There are several other places in the code we should probably rework that I know are using default params. Thanks for the heads up about that.

srallen:
FYI, I'm working on a bug with the tutorial getting null preferences sometimes.

srallen:
There's more work to do with reworking the request for preferences and the listener for it as well as moving the actual workflow request into the classifier, but those should be separate branches to stay in scope here. 

For the testing, I'm having difficulty with figuring out good tests and API mocking, so I'd like to do that separately too. I don't _think_ I'm breaking anything with the changes in this as is, but we'll find out quickly once merged like with the task rewrite.

eatyourgreens:
Not logged in, I'm able to get a GZ3D workflow showing in Muon Hunters, if I go to GZ3D, pick a workflow, then navigate to Projects > Muon Hunters > Get Started.

srallen:
@eatyourgreens that bug should be fixed now. The preferences handling needs a lot of work and some thought and this is partially related to unexpected behavior from the API client's handling of resource caching.

Basically what's happening is this: `fetchProjectData` is called in `componentWillReceiveProps` when the path is changed. When the project data is fetched, this includes the call to get the user project preferences. For logged out users, the API client is leveraged to create temporary guest preferences that can be written to. What I'm finding is that when you visit a workflow that allows you to choose, like GZ:3D, the selection is written to the guest preferences. Then, when you navigate to a new project, what should be happening is new guest preferences are created with an empty object in the preferences property.

Instead what is happening is that it still has the `selected_workflow` that was stored from GZ:3D, though the `project.id` in its links is correctly updated to the new project. I'm not sure why this would be happening unless that call to create is hitting the client's cache and only partially updating the guest preferences on the create call.

So the preferences still have a `selected_workflow` of the last project you visited, it attempts to load that workflow. I fixed the workflow fetch to include the project id, so at least this will now fail and then call `clearInactiveWorkflow`, clear out the `selected_workflow` id in the guest preferences and attempt the workflow selection again.

app/pages/project/index.cjsx
```diff
@@ -60,11 +60,11 @@ ProjectPageController = React.createClass
 
   componentWillUpdate: (nextProps, nextState) ->
     if nextProps.location.query?.workflow? and @canFetchWorkflowByQuery(nextProps.project, nextProps.user)
```
eatyourgreens: Should `nextProps.project` be `nextState.project` here?

eatyourgreens:
Looks good to me. One small change where I think you've missed a change from props to state.

Personally, I still find all the `if` statements in `getSelectedWorkflow` very hard to follow, and potentially very hard to extend or maintain, and I think workflow selection should be split out into its own class or component, but that is a discussion for zooniverse/rfc#12.

eatyourgreens:
Might need an update to include #3621

app/pages/project/index.cjsx
srallen: Yep. Thanks for catching that.

srallen:
@eatyourgreens this is rebased now to include #3621. I also agree about extracting out workflow selection into its own class. Good idea.

PR #3563, opened 2017-03-08
Improve keyboard control for survey task choices

eatyourgreens:
Towards #2132 

Describe your changes.
Autofocus the survey image flipper when the choice first mounts, allowing images to be changed with the cursor keys.
Autofocus the confusions popup when that opens, so that it can be dismissed from the keyboard if required.
Add a visual style to question answers when those radio buttons have focus.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://survey-choice-keyboard.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -96,7 +96,7 @@ module.exports = React.createClass
                   <ImageFlipper images={@props.task.images[filename] for filename in otherChoice.images} />
                   <Markdown content={choice.confusions[otherChoiceID]} />
                   <div className="survey-task-choice-confusion-buttons" style={textAlign: 'center'}>
-                    <button type="submit" autoFocus={true} className="major-button identfiy">Dismiss</button>
+                    <button type="submit" autoFocus={true and otherChoice.images.length < 2} className="major-button identfiy">Dismiss</button>
                     {' '}
                     <button type="button" className="standard-button cancel" onClick={@props.onSwitch.bind null, otherChoiceID}>I think it’s this</button>
                   </div>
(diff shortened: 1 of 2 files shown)


eatyourgreens:
The Choice component only receives focus if it has an image slideshow, but this seems to be a step in the right direction.

eatyourgreens:
https://survey-choice-keyboard.pfe-preview.zooniverse.org/projects/pat-lorch/focus-on-wildlife-cleveland-metroparks/classify?env=production seems to be a good project to test this out. Open a choice, then navigate through it with the cursor and tab keys.

app/classifier/tasks/survey/choice.cjsx
```diff
@@ -97,7 +97,7 @@ module.exports = React.createClass
                   <ImageFlipper images={@props.task.images[filename] for filename in otherChoice.images} />
                   <Markdown content={choice.confusions[otherChoiceID]} />
                   <div className="survey-task-choice-confusion-buttons" style={textAlign: 'center'}>
-                    <button type="submit" className="major-button identfiy">Dismiss</button>
+                    <button type="submit" autoFocus={true and otherChoice.images.length < 2} className="major-button identfiy">Dismiss</button>
```
mcbouslog: this is just for my understanding - but why the `true and ` before the condition we're interested in?

eatyourgreens: Ha! Nothing clever. This is a mistake on my part, not reading the code properly.

eatyourgreens: I'm going to have to revisit the logic for how autofocus is assigned when the image slideshow isn't present, so I'll leave changing this to a future PR.

PR #3573, opened 2017-03-09
Use subject.favorite to get favourited status

### LINKED ISSUE #3494: Determine if subject is a favorite in subject queue request not per subject get
## Expected behavior
For API optimization, request favorite status for each subject in the subject queue request and set local app state for subject queue. This would be similar to the already seen/seen this session functionality.

## Current behavior
Currently the favorite collection and the check if a subject is a favorite is done in the `componentWillMount` lifecycle method of the favorite button component. This means the collection get request is done at least once per subject.

eatyourgreens:
Fixes #3494  .

Describe your changes.
Uses the `subject.favorite` flag to set favourited status.
Removes API call to check for subject in collection.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://subject-favorite-flag.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -39,18 +39,29 @@ module.exports = React.createClass
       .get({project_ids: @props.project?.id, favorite: true, owner: @props.user.login})
       .then ([favorites]) -> if favorites? then favorites else null
 
+  findSubjectInCollection: (favorites) ->
+    if @props.subject.hasOwnProperty 'favorite'
+      Promise.resolve @props.subject.favorite
+    else if favorites?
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Seems to work in the classifier interface but I'm not sure that it works on Talk, or in collections.

camallen:
@eatyourgreens should we just drop the check and allow users to press the button no matter what. response should be ok / already in collections error that would signify to make the heart solid. Anything else we could report an error on?

eatyourgreens:
@camallen that works for the classifier but on Talk, and in collections, the heart should be solid already if they've previously favourited the subject.

PR #3574, opened 2017-03-09
Move field guide to classifier

simensta:
This PR addresses the field guide portion of #3492 and replaces PR #3535.

**Describe your changes.**
The PotentialFieldGuide component is now required in the `/classifier/index` instead of the `pages/project/index`.

The request for the field_guide has been moved to the ClassifierWrapper (see `loadFieldGuide()`), where the field_guide data and icons are set in state and passed to the Classifier and PotentialFieldGuide as props.

The icons for the field guide are not requested via an `include` on the original request because of the page_size issue identified in #3535. I've set the page_size on the request for the field guide, but I wasn't sure what number to specify. Any input on what the page size should be?

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://move-field-guide-to-classifier.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -271,6 +271,7 @@ Classifier = React.createClass
         else if not @props.workflow.configuration?.hide_classification_summaries # Classification is complete; show summary if enabled
           @renderSummary currentClassification}
       </div>
+      <PotentialFieldGuide guide={@props.guide} guideIcons={@props.guideIcons} />
     </div>
 
   renderSummary: (classification) ->
@@ -462,9 +463,14 @@ module.exports = React.createClass
   getInitialState: ->
     subject: null
     expertClassifier: null
+    guide: null
+    guideIcons: null
     userRoles: []
     tutorial: null
     minicourse: null
+
+  componentWillMount: ->
+    @loadFieldGuide(@props.project.id)
 
   componentDidMount: ->
     @checkExpertClassifier()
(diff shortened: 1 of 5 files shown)


app/classifier/index.cjsx
```diff
@@ -655,6 +672,8 @@ module.exports = React.createClass
       <Classifier {...@props}
         workflow={@props.workflow}
         subject={@state.subject}
+        guide={@state.guide}
```
srallen: I don't think the classifier itself needs the field guide and its linked media as props?

app/classifier/index.cjsx
```diff
@@ -156,6 +157,7 @@ Classifier = React.createClass
       @addAnnotationForTask classification, @props.workflow.first_task
 
   render: ->
+    console.log "@props.project", @props.project
```
simensta: I'll remove this console.log.

app/classifier/index.cjsx
simensta: Agreed. Do you think it is better to request the field guide data in the classifier? Or, is it better to move the request back to the PotentialFieldGuide component?

srallen: Oh, I think the request in the classifier wrapper component in the index file is fine for now, though we may get feedback from volunteers that were using it on Talk after this is merged. I just mean that the guide and images are being passed down as props to the classifier component where those props aren't being used.

srallen: Ok ignore my comment. I confused myself because the index file is the classifier and the classifier wrapper components. We need to split these out.

Issue #3361, opened 2016-12-22
How do I check conventions for and enable feedback on Space Warps Demo?


aprajita:
https://github.com/zooniverse/Panoptes-Front-End/pull/3202 has been merged so am starting a new issue as @srallen advised. This issue is part of the SW feedback requested in https://github.com/zooniverse/Panoptes-Front-End/issues/2834 (now closed).

## Expected behavior
Same numbering is used in current behaviour
1) Expect to see a success message for SIMs, but failure message shown even when marker appears to be in the right place
2) Not sure how to implement feedback in the SWDemo site

## Current behavior
_Please include any error messages from the browser console and/or screenshots_

1) I uploaded a selection of SW subjects including sims & duds onto 
https://space-warps-feedback.pfe-preview.zooniverse.org/projects/hughdickinson/testing-metadata-based-feedback/classify
to see how this works. 

The success and fail messages seem to be OK for duds, but for sims the success message doesn't appear always giving a failure message (see fig. below). I suspect this is something in to do with either the starting point from where the x,y positions should be defined and/or the tolerance.  We include the "radius" for tolerance around the lensed feature's x,y position in the manifest under "Tol".
As I can't see the manifest for @hughdickinson 's test dataset, I can't see what the required format/keywords are to provide this info. Please could you let me know what the conventions for this info should be?

[image](https://cloud.githubusercontent.com/assets/1917970/21443485/fcbd92d4-c8a5-11e6-8120-d1df72bad4fb.png)
this is for image_4.png showing the failure message and marker on the feature. I expected the feature to be marked around (165, 269) and gave a generous tolerance of 80 so a success message should have been recorded.

2) Also if this functionality is now merged, I'm not quite sure how to select this functionality under our SWDemo workflows (https://www.zooniverse.org/lab/1434) - please could you advise?

## Additional information

- **Operating system:**
- **Browser:**

srallen:
@hughdickinson could you share the manifest for the test data for the metadata feedback so we know how the subject metadata needs to be formatted for this feature? Thanks.

aprajita:
@hughdickinson i was wondering if you might be able to share the manifest with me as the feedback didn't seem to work well ( query 1 above)?

Also can you advise me on query 2 above?

Many thanks!

aprajita:
[...]

PR #3590, opened 2017-03-10
Extract classification summary into component

amyrebecca:
Extract the code to show a classification summary (whether it is a metadata summary, a WWT summary, or a default summary) from the classifier index.

Renames former classification-summary.* to default-classification-summary.*
Introduces new classification-summary.jsx

# Review Checklist

- [x] Does it work on mobile?

## Optional


amyrebecca:
I still haven't written the tests for this yet but I wanted to put it out as a WIP. It's been months since the last time I worked with Coffeescript and it shows!

amyrebecca:
Also I've been having a fun time trying to rebase this onto the latest set of merges to master but when I get through that I'll push it up.

eatyourgreens:
Looks good to me. Since `renderSummary` returns a component, I would remove that method completely and replace the call to `renderSummary()` with a component.

app/classifier/classification-summary.jsx
```diff
@@ -1,54 +1,89 @@
...
-  classification: React.PropTypes.object,
-  classificationCount: React.PropTypes.number,
-  onToggle: React.PropTypes.func,
-  splits: React.PropTypes.object
+  project: React.PropTypes.object.isRequired,
+  workflow: React.PropTypes.object.isRequired,
+  subject: React.PropTypes.object.isRequired,
+  classification: React.PropTypes.object.isRequired,
+  expertClassification: React.PropTypes.object,
+  splits: React.PropTypes.array,
```
srallen: I'm not sure why this changed, but this now has a prop type validation error. It should be an object.

app/classifier/index.cjsx
```diff
@@ -342,7 +301,6 @@ Classifier = React.createClass
       </nav>
     </div>
 
-
   renderGravitySpyGoldStandard: (classification) ->
```
srallen: This method is also a kind of summary that could be moved into the `ClassificationSummary` component. There's a method on L378 that checks if this summary should be shown.

srallen:
Also, just a note, I'm not sure if a merge conflict in a rebasing got solved correctly, but this seems to missing the work I did from #3556 which merged last Friday.

amyrebecca:
I'm not sure why the stuff from #3556 would have gone missing. I just re-fetched upstream/master before I did my rebase. I just tried it again to make sure, is it still gone?

srallen:
I just deleted my local copy and refetched the branch and it still doesn't look up to date. The last commit I see before the two from this branch is `8df76ad93f93ac9f3cb0900c569b6d63c00f03d2` which is from Monday, March 13.

amyrebecca:
okay this is really weird...
screen shot 2017-03-21 at 4 11 16 pm

srallen:
Something pushing the updated branch back up? If you switch to the branch view for `extract-summary` on github, it says it's 72 commits behind master.

amyrebecca:
Yeah, I see that too. I'm not sure what's going on here.

```shell
$ git checkout master
$ git pull upstream master
$ git checkout extract-summary
$ git rebase upstream/master
$ git push origin extract-summary --force-with-lease
```

It says everything is up-to-date. I'll keep trying to figure out what's gone wrong here, it's frustrating.

amyrebecca:
I was pushing to `origin` instead of `upstream` but I made this PR from `upstream` instead because I thought someone else might have to pick it up when I switched off for something like a week ago :/ No wonder things weren't acting the way I expected.

amyrebecca:
I'm working on the refactoring that sarah and jim suggested, should be done this am.

eatyourgreens:
Sorry, I missed the comment about the `TaskNav` component. Moving that component out of the `Task` component will probably break the page layout, since it will move beneath the mini course and tutorial buttons.

amyrebecca:
I will give it a test with that project and see what's going on with it.

amyrebecca:
Oh, yeah, yikes--I didn't have that kind of task in the workflow I was testing with. It works much better now (in the sense that "works at all" is much better than "does not do anything but error")

amyrebecca:
I'm not sure I understand the way the page is meant to lay out if the tutorial or field guide buttons are enabled. They are meant to be *below* the task navigation? It should be possible to reorganize the page to make that work with a little patience.

srallen:
@amyrebecca Maybe working on the TaskNav part can be broken up into a separate PR? I'd really like to get the basic summary extraction merged in asap tomorrow so we can work on the metadata feedback for SGL.

amyrebecca:
it wasn't really feasible to remove the renderSummary method as suggested without also moving the TaskNav, unless I were to move the TaskNav inside the ClassificationSummary, which doesn't make much sense. if we have to break it up i'll start trying to figure out how, but it'd probably be faster to just fix the layout issue at this point.

srallen:
The field guide button is a pull out component and shows up z-indexed on the side. It's the tutorial and mini-course buttons that appear below the task. 

Could the classification summary component be written take children props, so in the classifier: `<ClassificationSummary><TaskNav /></ClassificationSummary>` and add `{this.props.children}` where appropriate in the classification summary render? Then keep and reuse TaskNav in the Task component as it was written?

eatyourgreens:
I wrote the `TaskNav` component to render the Back, Next and Done buttons when stepping through workflow tasks. It also contains the logic that creates a new annotation for the first task, caches annotations and manages them as you work through a sequence of tasks. Maybe I should have called it `WorkflowNav`?

Although the Talk links under the summary have the `task-nav` CSS class, I'm not sure it makes sense to try and render those with the same component that's also managing the workflow navigation. They look the same, but I think they're doing something different.

I'd be wary about removing the `return null` when no annotation is present. It forces the component to wait for initialisation to complete, in the classifier, before rendering. Without it, I think I was getting errors in #3522

That said, I'm not seeing any errors for a couple of test projects on staging (one survey task, one questions and drawing workflow), so the code seems ok.

app/classifier/task-nav.jsx
```diff
@@ -105,27 +112,26 @@ class TaskNav extends React.Component {
   }
 
   render() {
-    if (!this.props.annotation) {
-      return null;
-    }
+    const completed = !!this.props.classification.completed;
+
     const task = this.props.task ? this.props.task : this.props.workflow.tasks[this.props.workflow.first_task];
```
eatyourgreens: I'm not 100% sure, but removing this might break the classifier for some types of tasks. It forces the component to wait until the classification, task and annotations have initialised before rendering.

app/classifier/index.cjsx
```diff
@@ -184,23 +182,6 @@ Classifier = React.createClass
...
-            >
-              {if @props.expertClassifier
-                <ExpertOptions
-                  classification={currentClassification}
-                  userRoles={@props.userRoles}
-                  demoMode={@props.demoMode}
-                  onChangeDemoMode={@props.onChangeDemoMode}
-                />}
-            </TaskNav>
             <p>
```
eatyourgreens: Tutorial buttons etc. here should be rendered after the task navigation buttons.

eatyourgreens:
I'm happy to merge this. The Classifier component is looking so much better now that all the hairy conditional rendering logic has been removed (mostly into the TaskNav component.)

app/classifier/task-nav.jsx
eatyourgreens: On second thoughts, looks like the additional checks for the existence of `annotation` have addressed this.

app/classifier/index.cjsx
eatyourgreens: Fixed in 64a59ee

eatyourgreens:
Staged at https://extract-summary.pfe-preview.zooniverse.org/

eatyourgreens:
The experimental tool flags in the classifier don't match [the strings listed in the admin pages](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/admin/project-status/experimental-features.jsx). Looks like this has been broken for a while.

amyrebecca:
There are so many more corner cases for this than I had ever imagined! I'm going to look into the most recent ones and hopefully we can get this put to bed today.

eatyourgreens:
@amyrebecca I think the commits I added this morning should pretty much take care of it. I also think some of this has been broken on the live site but noone has noticed.

amyrebecca:
Yeah! I'm going to re-check the stuff that Sarah mentioned and if there are any remaining problems, we can get them cleared up ASAP

eatyourgreens:
One small thing that I've just noticed. The conditional for WorldWide Telescope has changed between this PR and master. Master is checking workflow configuration. This PR checks project tools.

amyrebecca:
yeah, i'll get that, it looks like a quick fix. in practice WWT will probably only be turned on for projects that use it but it's better to be correct

amyrebecca:
Which project are you testing the metadata based feedback on? I'm happy to mop that up before we merge it, I just don't know where to test it.

srallen:
@amyrebecca https://extract-summary.pfe-preview.zooniverse.org/projects/hughdickinson/testing-metadata-based-feedback/classify

amyrebecca:
Cool, I'll get a fix in asap

amyrebecca:
Actually looking over this, it seems like it's working correctly? It wants to look the message up in the metadata, but doesn't provide any lookup values so it just defaults to "". It's a bad user experience but it would work fine if the subject had the right metadata on it.
screen shot 2017-03-23 at 2 14 12 pm
screen shot 2017-03-23 at 2 14 40 pm
screen shot 2017-03-23 at 2 15 04 pm
screen shot 2017-03-23 at 2 15 20 pm

srallen:
@amyrebecca I don't recall the test project ever having a blank messaging, but nobody should have this feature enabled since some other bugs and details need working out as is being discussed in #3361. I think this is good to merge?

srallen:
As an aside, I might be misremembering because I have a stale branch that I working on fixing the bugs for it that hasn't merged yet.

amyrebecca:
Yeah I agree it'd be better to provide some default values! Since the changes would happen in a file that this PR didn't touch, I'm inclined to agree with you that we could note this and move on. :)

PR #3592, opened 2017-03-13
Tasks generic tests

eatyourgreens:
Add some generic tests to run against all the task components.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -48,8 +48,10 @@ module.exports = React.createClass
       annotation.value isnt '' or not task.required
 
   getDefaultProps: ->
-    task: null
-    annotation: null
+    task:
+      instruction: ''
+    annotation:
+      value: null
     onChange: NOOP
 
   getInitialState: ->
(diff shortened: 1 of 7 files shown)


app/classifier/tasks/text/index.cjsx
```diff
@@ -48,8 +48,10 @@ module.exports = React.createClass
       annotation.value isnt '' or not task.required
 
   getDefaultProps: ->
-    task: null
-    annotation: null
+    task:
+      instruction: ''
+    annotation:
+      value: null
```
wgranger: So I'm not sure if this is an issue, but I'm getting the warning `Warning: Failed prop type: Required prop `annotation.value` was not specified in `TaskNav`.` when this is passed into `TaskNav` because it's looking for that required prop, which is given, but it's `null`. I noticed this same thing with the single choice task. Not sure if it's an issue though, as the warning doesn't cause anything to crash.

wgranger:
LGTM. Just that one comment I made on `null` being passed to `annotation.value`.

app/classifier/tasks/text/index.cjsx
eatyourgreens: Yeah, I made annotations a required property in the task navigation, but it turns out they aren't always defined when a subject first loads. Nothing in `TaskNav` will crash if that value isn't supplied. Validating annotations is a bit of a nightmare.

PR #3595, opened 2017-03-13
move subjectDimensions to state and add to the classification in the TaskNav componenet

### LINKED ISSUE #3591: Sometimes the classifier component disappears  after a subject finishes loading
## Expected behavior
The classifier should stay on the page after a subject loads.

## Current behavior
See https://zooniverse.slack.com/files/coleman.krawczyk/F4HD07KFE/classifier_bug.mov for a video of the bug.  No errors show up in the console.

## Steps to replicate
Go to the `Bar` or `Spiral` workflows for [GZ: 3D](https://www.zooniverse.org/projects/klmasters/galaxy-zoo-3d), it does not happen every time.  From what I can tell it seems to happen when the subject finishes loading after the classifier component is mounted.

## Additional information

- **Operating system:**
OSX
- **Browser:**
Seen if Firefox and Chrome

CKrawczyk:
Fixes #3591.

This sets the `subjectDimensions` as `state` on the classifier so it does not trigger a race condition when updated the `classification` object.  The `subjectDimensions`are passed as a prop down to the `TaskNav` component, and are added to the `classification` object with the rest of the `metadata`.

The original race condition only triggered when a users subject queue was empty and the subject on the classification page started to load before `TaskNav` but finished loading after `TaskNav`.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3591.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -388,7 +388,6 @@ Classifier = React.createClass
     @context.geordi?.remember subjectID: @props.subject?.id
 
     {naturalWidth, naturalHeight, clientWidth, clientHeight} = e.target
-    changes = {}
     subjectDimensions = @state.subjectDimensions.slice()
     subjectDimensions["#{frameIndex}"] = {naturalWidth, naturalHeight, clientWidth, clientHeight}
     @setState({ subjectDimensions })
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -55,6 +55,7 @@ Classifier = React.createClass
     selectedExpertAnnotation: -1
     showingExpertClassification: false
     subjectLoading: false
+    subjectDimensions: []
     annotations: []
 
   componentDidMount: ->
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Also fixed by #3596

eatyourgreens:
I'd much rather the `TaskNav` component didn't need to know about the subject. It calls a callback after Done is pressed, so maybe that callback could handle updating the image dimensions?

eatyourgreens:
Have you verified that this fixes the bug? I can still reproduce #3591 on a local copy of this branch, by artificially delaying the subject image load using `setTimeout`.

eatyourgreens:
I'm also getting two tutorials loading, for two different workflows (the default workflow and the selected workflow), so maybe something weird is going on with workflow selection here?

srallen:
@eatyourgreens that could be happening because method handling selecting the workflow id and the request for the workflow is being fired twice likely because of a prop update (my guess is user project preferences since it has a listener). Maybe check if this is fixed in #3556?

CKrawczyk:
closed in favor of #3596.

eatyourgreens:
I've written up the weird tutorial behaviour in #3598

PR #3600, opened 2017-03-14
Edit Access to Collaborated Collections

### LINKED ISSUE #3435: No edit access to collaborated collections
## Expected behavior
I am a collaborator on someone else's collection, so when I view that collection (via their link, say, `/collections/trouille/test55`), I should see the same edit tabs ("Settings", "Collaborators", etc) that they see.

## Current behavior
They're treated like anyone else's collections and I am unable to, for instance, remove myself as a collaborator.

## Additional information

- **Operating system:**
- **Browser:**

wgranger:
Fixes #3435 

Describe your changes.
Allows collection collaborators to also have access to "Settings" and "Collaborators" tabs of collections.

_Also changes the functionality of other user roles on collections_:
**Viewer**: can view a collection (even if private) but cannot add to the collection.
**Contributor**: can add to/view a collection but cannot edit settings
**Collaborator**: can add to/view/edit a collection
**Owner**: owns the collection and can do everything

# Review Checklist

- [x] Does it work on mobile?

https://edit-access-collections.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -150,11 +150,11 @@ RoleRow = React.createClass
       @props.roleSet.delete()
 
     promise
-      .catch (error) =>
-        @setState { error }
       .then =>
         callback =>
-          @setState saving: false
+        @setState saving: false
+      .catch (error) =>
+        @setState { error }
 
   render: ->
     { owner } = @state
--- a/app/collections/collaborators.cjsx
+++ b/app/collections/collaborators.cjsx
@@ -151,7 +151,7 @@ RoleRow = React.createClass
 
     promise
       .then =>
-        callback =>
+        callback()
         @setState saving: false
       .catch (error) =>
         @setState { error }
--- a/app/collections/show.jsx
+++ b/app/collections/show.jsx
@@ -69,6 +69,17 @@ const CollectionPage = React.createClass({
     document.documentElement.classList.remove('on-collection-page');
   },
 
+  canCollaborate() {
+    if (!this.props.user) return false;
+
+    return this.props.roles.some((role) => {
+      const idMatch = (role.links.owner.id === this.props.user.id);
+      const isOwner = role.roles.includes('owner');
+      const isCollaborator = role.roles.includes('collaborator');
+      return ((isOwner && idMatch) || (isCollaborator && idMatch));
+    });
+  },
+
   render() {
     if (!this.state.owner) {
       return null;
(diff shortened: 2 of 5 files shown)


app/collections/collaborators.cjsx
```diff
@@ -150,11 +150,11 @@ RoleRow = React.createClass
       @props.roleSet.delete()
 
     promise
-      .catch (error) =>
-        @setState { error }
       .then =>
         callback =>
-          @setState saving: false
+        @setState saving: false
```
eatyourgreens: I think the indentation might be wrong here.

app/collections/show.jsx
```diff
@@ -69,6 +69,17 @@ const CollectionPage = React.createClass({
...
   },
 
+  canCollaborate() {
+    if (!this.props.user) return false;
+
+    return this.props.roles.some((role) => {
+      const idMatch = (role.links.owner.id === this.props.user.id);
+      const isOwner = role.roles.includes('owner');
+      const isCollaborator = role.roles.includes('collaborator');
+      return ((isOwner && idMatch) || (isCollaborator && idMatch));
```
eatyourgreens: `(isOwner || isCollaborator) && idMatch`

app/collections/collaborators.cjsx
wgranger: The problem here is that `@setState saving: false` wasn't triggering when a role was checked on (line 140) because the default callback was being used and skipping the state change. Changing the indention solved this, but perhaps there is a better option?

eatyourgreens: Ah right. No, I was asking because I'm not familiar with this bit of code and wasn't sure why it had changed.

app/components/collection-search.cjsx
```diff
@@ -23,7 +23,7 @@ module.exports = React.createClass
     query =
       page_size: 20
       favorite: false
-      current_user_roles: 'owner,collaborator,contributor,viewer'
+      current_user_roles: 'owner,collaborator,contributor'
```
srallen: For some reason this isn't showing up in the diff view on github which is a bit concerning.

wgranger: This makes sense because I added `viewer` in one commit and then removed it in a later commit when I noticed this `collection-search` was responsible for adding subjects to a collection. It should be kept as it currently it on master.

app/collections/collaborators.cjsx
```diff
@@ -150,11 +150,11 @@ RoleRow = React.createClass
       @props.roleSet.delete()
 
     promise
-      .catch (error) =>
-        @setState { error }
       .then =>
         callback =>
```
wgranger: That makes sense.

PR #3618, opened 2017-03-16
Add simple filter for already seens

srallen:
This adds a simple filter to the subjects array to remove the already seens. If the array is empty, then it falls back to the original response. This is a step toward a more intelligent handling of the subject queue. I still want to take some time to think about when the subject queue end point should be requested again.

https://skip-already-seen.pfe-preview.zooniverse.org/

and in production: https://skip-already-seen.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy/classify?env=production

# Review Checklist

- [ ] Does it work on mobile?

## Optional


camallen:
Nice - i was going to suggest some reflection on the other data we serialize with each subject from that end point. Specifically the user_has_finished workflow and i think we should serialize if the workflow has finished in there too so the front end can detect that change without having to reload the workflow itself.

PR #3622, opened 2017-03-17
Remove API call for user avatar in favor of using user.avatar_src

srallen:
As a follow up to #3576, this removes the separate user avatar API call that was on the signed in home page. 

https://remove-avatar-fetch.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/home-for-user/circle-ribbon.jsx
+++ b/app/pages/home-for-user/circle-ribbon.jsx
@@ -13,12 +13,14 @@ const CircleRibbon = React.createClass({
     size: React.PropTypes.string,
     weight: React.PropTypes.number,
     gap: React.PropTypes.number,
-    image: React.PropTypes.string,
     loading: React.PropTypes.bool,
     data: React.PropTypes.array,
     hrefTemplate: React.PropTypes.func,
     onClick: React.PropTypes.func,
-    user: React.PropTypes.object.isRequired,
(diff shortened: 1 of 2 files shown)


app/pages/home-for-user/circle-ribbon.jsx
```diff
@@ -168,6 +170,8 @@ const CircleRibbon = React.createClass({
       tooltipPosition = this.getTooltipPoint(hoveredArc, 50);
     }
 
+    const avatar = this.props.user.avatar_src ? this.props.user.avatar_src : '/assets/simple-avatar.png';
```
wgranger: Can this just be `const avatar = this.props.user.avatar_src` since the default is always set to `'/assets/simple-avatar.png'`?

wgranger:
Cool. I really like the change of adding `avatar_src` to the user resource.

app/pages/home-for-user/circle-ribbon.jsx
srallen: The default will only be used while the user prop is null, basically, while we're waiting on authentication. This will set it to the fallback simple avatar if specifically the `avatar_src` for the current logged in user is null.

wgranger: Ohh, gotcha. That makes sense.

Issue #3564, opened 2017-03-08
Survey task choices: visual layout doesn't reflect source order


eatyourgreens:
## Expected behavior
Tabbing through survey task choices should move sequentially down each column in turn. eg. for Chicago Wildlife Watch, the order should be Beaver, Bird, Cat (domestic), Chipmunk etc.

## Current behavior
The actual tab order of the choices on CWW is Beaver,  Human, Skunk, Bird ie. the choices are actually laid out in rows across the page, even though we read them vertically.

## Steps to replicate
Select choices in a survey task project using the tab key.

eatyourgreens:
#3623 partly addresses this by using the up/down cursor keys to navigate the list in alphabetical order.

PR #3623, opened 2017-03-17
More survey task keyboard improvements

eatyourgreens:
Describe your changes.
If a choice has been chosen, focus that button when the chooser mounts.

Add backspace to remove the currently selected choice.

Add up/down keys to navigate the choice list, when focused.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://survey-choice-keyboard.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/survey/chooser.cjsx
+++ b/app/classifier/tasks/survey/chooser.cjsx
@@ -80,12 +80,12 @@ module.exports = React.createClass
 
     <div className="survey-task-chooser">
       <div className="survey-task-chooser-characteristics">
-        {for characteristicID in @props.task.characteristicsOrder
+        {for characteristicID, i in @props.task.characteristicsOrder
           characteristic = @props.task.characteristics[characteristicID]
           selectedValue = characteristic.values[@props.filters[characteristicID]]
           hasBeenAutoFocused = false
 
-          <TriggeredModalForm key={characteristicID} ref="#{characteristicID}-dropdown" className="survey-task-chooser-characteristic-menu" trigger={
+          <TriggeredModalForm key={characteristicID} ref="#{characteristicID}-dropdown" className="survey-task-chooser-characteristic-menu" triggerProps={autoFocus: i is 0 and not @props.focusedChoice} trigger={
             <span className="survey-task-chooser-characteristic" data-is-active={selectedValue? || null}>
               <span className="survey-task-chooser-characteristic-label">{selectedValue?.label ? characteristic.label}</span>
             </span>
--- a/app/classifier/tasks/survey/chooser.cjsx
+++ b/app/classifier/tasks/survey/chooser.cjsx
@@ -78,7 +78,16 @@ module.exports = React.createClass
                 if autoFocus
                   hasBeenAutoFocused = true
 
-                <button key={valueID} type="submit" title={value.label} className="survey-task-chooser-characteristic-value" disabled={disabled} data-selected={selected} autoFocus={autoFocus} onClick={@handleFilter.bind this, characteristicID, valueID}>
+                <button
+                  key={valueID}
+                  type="submit"
+                  title={value.label}
+                  className="survey-task-chooser-characteristic-value"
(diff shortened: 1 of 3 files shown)


eatyourgreens:
This should allow you to carry on classifying with the tab key when the choice component closes.

That said, it would be more intuitive to choose options with the cursor keys, and also fix #3564

eatyourgreens:
Backspace to remove a selected choice, when that choice has focus, would be good too.

app/classifier/tasks/survey/chooser.cjsx
```diff
@@ -2,6 +2,11 @@ React = require 'react'
 TriggeredModalForm = require 'modal-form/triggered'
 sortIntoColumns = require 'sort-into-columns'
 
+# key codes
+BACKSPACE = 8
```
mcbouslog: Pressing "Delete" on my Mac in Firefox makes the browser go back. Before it goes back it accurately deletes the survey task selection, which I can see when I click the "Get Started" button again, as it'll be where I left off (showing selections I still had, removed selection I pressed "Delete" to remove).

app/classifier/tasks/survey/chooser.cjsx
```diff
@@ -11,6 +16,27 @@ module.exports = React.createClass
     onFilter: Function.prototype
     onChoose: Function.prototype
 
+  componentWillMount: ->
+    # refs for the choices
+    @choiceButtons = []
```
eatyourgreens: Yes, this is a workaround for #3564, since the buttons aren't rendered in the same order that they display visually.

`@choiceButtons` is recalculated after a render, so I kept it out of component state.

app/classifier/tasks/survey/chooser.cjsx
eatyourgreens: I've missed a `preventDefault` for backspace. Easy to fix.

app/classifier/tasks/survey/chooser.cjsx
```diff
@@ -11,6 +16,27 @@ module.exports = React.createClass
...
+    @sortChoiceButtons()
+
+  componentDidUpdate: ->
+   @sortChoiceButtons()
+
+  sortChoiceButtons: ->
+    # overrides default DOM focus order by sorting the buttons according to task.choicesOrder
+    newChoiceButtons = []
+    @choiceButtons
+      .filter Boolean
```
eatyourgreens: Yep. 👍🏻

app/classifier/tasks/survey/index.cjsx
```diff
@@ -63,11 +64,20 @@ module.exports = React.createClass
     newAnnotation = Object.assign {}, @props.annotation, _choiceInProgress: true
     @props.onChange newAnnotation
 
+  handleRemove: (choiceID) ->
+    selectedChoices = (item.choice for item in @props.annotation.value)
+    index = selectedChoices.indexOf choiceID
```
eatyourgreens: I think this is a bug in the survey task. Clicking a selected choice should probably edit the annotation, not create a new one.

app/classifier/tasks/survey/chooser.cjsx
eatyourgreens: I went back and re-read [the docs for refs](https://facebook.github.io/react/docs/refs-and-the-dom.html), and now I think `@choiceButtons` is only calculated on mount.

app/classifier/tasks/survey/chooser.cjsx
eatyourgreens: Turns out [Google removed backspace in Chrome 52](http://venturebeat.com/2016/08/14/restore-backspace-shortcut-chrome/), which is why I missed this.

eatyourgreens:
Now with added autofocus on first mount/render, like the text transcription tasks.

eatyourgreens:
Choice component updated so that the questions can be autofocused if there are no slideshow controls.

PR #3631, opened 2017-03-21
Stop Logging with Notification Tests

wgranger:
Describe your changes.

This isn't a priority, just a quick fix I made because notification tests were logging each time `npm test` was run. This was because when testing, `componentWillMount` was being run without a way to catch an error with the API call, causing the error to log out.

https://fix-notification-tests.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/notifications.spec.js
+++ b/app/pages/notifications.spec.js
@@ -4,6 +4,7 @@ import React from 'react';
 import assert from 'assert';
 import Notifications from './notifications';
 import { mount, shallow } from 'enzyme';
+import sinon from 'sinon';
 
 const testNotifications = [
   { id: '123',
@@ -21,7 +22,8 @@ const testNotifications = [
 ];
 
 describe('Notifications', function() {
-  let wrapper, notifications;
+  let wrapper;
+  let notifications;
 
   describe('it will display according to user', function() {
     it('will ask user to sign in', function() {
@@ -30,18 +32,20 @@ describe('Notifications', function() {
     });
 
     it('will notify when no notifications present', function() {
+      const stub = sinon.stub(Notifications.prototype, 'componentWillMount');
       wrapper = mount(<Notifications user={{ id: 1 }} />);
       assert(wrapper.contains(<span>You have no notifications.</span>));
+      stub.restore();
     });
   });
 
   describe('it correctly display projects', function() {
     beforeEach(function () {
       wrapper = shallow(
-        <Notifications user={{ id: 1 }} />,
+        <Notifications />,
       );
       wrapper.instance().groupNotifications(testNotifications);
-      notifications = shallow(wrapper.instance().renderNotifications())
+      notifications = shallow(wrapper.instance().renderNotifications());
     });
 
     it('will place zooniverse section first', function() {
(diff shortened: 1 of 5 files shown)


app/pages/notifications.spec.js
```diff
@@ -30,18 +32,20 @@ describe('Notifications', function() {
...
       wrapper = mount(<Notifications user={{ id: 1 }} />);
       assert(wrapper.contains(<span>You have no notifications.</span>));
+      stub.restore();
     });
   });
 
   describe('it correctly display projects', function() {
     beforeEach(function () {
       wrapper = shallow(
-        <Notifications user={{ id: 1 }} />,
```
srallen: Shouldn't the following tests fail without a user? I'm confused about why this changed and why the tests are passing.

app/pages/notifications.spec.js
```diff
@@ -56,20 +60,20 @@ describe('Notifications', function() {
   describe('will open sections correctly', function() {
     beforeEach(function () {
       wrapper = shallow(
-        <Notifications user={{ id: 1 }} />,
+        <Notifications />,
```
srallen: Same comment about the lack of user prop here.

app/pages/notifications.spec.js
wgranger: So, I took out the user here so `getProjectNotifications` would not run and log the API catch error to the console when the component is mounted, but I could also provide a stub instead as I did above. Running `groupNotifications(testNotifications)` is mocking an example of what might be returned by an API call in `getProjectNotifications`.

wgranger:
I can easily move everything into the notification folder if that would help organization.

app/pages/notifications.spec.js
srallen: This sounds like to me that when the API calls is made needs to be refactored. Those API calls shouldn't be made if there is no user. L35 in notifications.jsx likely should change to `if (nextProps.user !== null && nextProps.user !== this.props.user) { this.getProjectNotifications(); }`  Forcing the instance to render without a user doesn't seem like it the right way to test the expected rendered output for when the user should be logged in to me.

srallen: One of the options for `shallow` is to disable lifecycle methods too: https://github.com/airbnb/enzyme/blob/master/docs/api/shallow.md

Which I think you might want in this case since you want to pass the mocked response and not have that API call run.

wgranger: Yes. I'll have to change the api call to `componentDidMount`, because `disableLifecycleMethods` doesn't affect `componentWillMount`, but hopefully this won't change anything.

srallen: asynchronous calls to a remote server should go in `componentDidMount`: https://facebook.github.io/react/docs/react-component.html#componentdidmount

wgranger:
Ok, I think I reorganized things in a way that makes more sense.

app/pages/notifications/index.jsx
```diff
@@ -25,12 +25,16 @@ export default class NotificationsPage extends React.Component {
     };
   }
 
-  componentWillMount() { // eslint-disable-line
-    if (this.props.user) this.getProjectNotifications();
+  componentDidMount() { // eslint-disable-line
```
srallen: This inline eslint disable is not necessary.

app/pages/notifications/index.jsx
```diff
@@ -25,12 +25,16 @@ export default class NotificationsPage extends React.Component {
...
+    if (this.props.user) {
+      this.getProjectNotifications();
+    }
   }
 
   componentWillReceiveProps(nextProps) { // eslint-disable-line
```
srallen: eslint disable not necessary.

app/router.cjsx
```diff
@@ -13,7 +13,7 @@ React = require 'react'
 `import ProjectStatus from './pages/admin/project-status';`
 `import EditMediaPage from './pages/lab/media';`
 `import UserProfilePage from './pages/profile/index';`
-`import NotificationsPage from './pages/notifications';`
+`import NotificationsPage from './pages/notifications/index';`
```
srallen: You don't need to specify `index` here. An import of a folder will load any file named index first. (This generally works for web file structure and isn't just for node). Not sure why others did that above in the imports.

wgranger:
This branch has been rebased and restaged.

PR #3637, opened 2017-03-22
Bump panoptes-client to 2.6

eatyourgreens:
Describe your changes.
This makes HTTP status codes available with API errors, as `error.status`.

During SGL, this should allow us to put up a `service unavailable` message for endpoints that are temporarily offline (returning 503 status codes.)

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://api-client-errors.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "markdownz": "~7.0",
     "modal-form": "~2.4",
     "moment": "~2.9.0",
-    "panoptes-client": "~2.5.0",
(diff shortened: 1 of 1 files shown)


eatyourgreens:
This update doesn't change the Panoptes client code that generates the error messages displayed by PFE, so in theory it should be backward compatible with existing API error-handlers.

PR #3641, opened 2017-03-23
Fix App Status component - better colours on Project Page and no more 'empty' text

shaunanoordin:
## PR Overview
* There are two issues, reported by @astopy on Slack at 8am today (GMT+0):
  * The App Status component is unreadable on individual Project pages, as the text becomes white.
  * If the [status text file](https://static.zooniverse.org/zooniverse.org-status.txt) consists of only a newline, the App Status component thinks there's a message.
screen shot 2017-03-23 at 08 13 00
* This PR fixes both issues by...
  * _overriding the colour overrides_ on Project pages.
  * `trim()`-ing any text the App Status compo receives - this eliminates not just stray newlines, but any sort of whitespace that serves no purpose.
screen shot 2017-03-23 at 13 49 30
* Tested with OSX+Chrome - but note that due to the external resource (a text file on `zooniverse-static`), testing had to be done under localhost conditions with an [external test file.](https://static.zooniverse.org/zooniverse.org-status-TEST.txt)

# Review Checklist

- [x] Does it work on mobile?

### DIFF
--- a/app/partials/app-status.jsx
+++ b/app/partials/app-status.jsx
@@ -53,13 +53,13 @@ export default class AppStatus extends React.Component {
     })
     .then((text) => {
       console.log('AppStatus: Received status data from ' + APP_STATUS_URL + '.');
-      const cleanedtext = text.trim();  //If text is just white space or newlines...
-      if (!cleanedtext || cleanedtext === '') {  //...ignore it.
+      const cleanedText = (text) ? text.trim() : '';  //If text is just white space or newlines...
+      if (cleanedText === '') {  //...ignore it.
         console.log('AppStatus: Nothing to report.');
       } else {
         this.setState({
           show: true,
-          message: cleanedtext,
+          message: cleanedText,
         });
       }
     })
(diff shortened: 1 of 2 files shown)


app/partials/app-status.jsx
```diff
@@ -53,12 +53,13 @@ export default class AppStatus extends React.Component {
     })
     .then((text) => {
       console.log('AppStatus: Received status data from ' + APP_STATUS_URL + '.');
-      if (!text || text === '') {
+      const cleanedtext = text.trim();  //If text is just white space or newlines...
```
eatyourgreens: This might break if `text` is `null` or `undefined`.

eatyourgreens:
One small comment: this might fail if `text.trim` doesn't exist.

app/partials/app-status.jsx
shaunanoordin: That's a good point - I overlooked that my `null` check was moved after the trim. Update shortly.

shaunanoordin:
Thanks @eatyourgreens , the code now properly checks for the existence of `text`.

shaunanoordin:
As suggested by @camallen , the URL for the global status text message now identifies if the PFE app is running on staging/development or production.

PR #3647, opened 2017-03-24
Reorganise classifier

eatyourgreens:
Describe your changes.
Splits Classifier and ClassifierWrapper into seperate files.

Rewrites Classifier in ES6 and adds prop type validation.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://reorganise-classifier.pfe-preview.zooniverse.org/

## Optional


app/classifier/classifier.jsx
```diff
@@ -0,0 +1,429 @@
...
+  toggleExpertClassification(value) {
+    this.setState({ showingExpertClassification: value });
+  }
+
+  changeDemoMode(demoMode) {
+    this.props.onChangeDemoMode(demoMode);
+  }
+
+  subjectIsGravitySpyGoldStandard() {
+    this.props.workflow.configuration.gravity_spy_gold_standard && this.props.subject.metadata['#Type'] === 'Gold';
```
srallen: This needs a `return` for the gravity spy feedback to work.

srallen:
Just as a side note, I don't think this branch is breaking it, but I'm still not totally 100% clear on that since it seems ok on the current production site. The expert classifications with the Picker Practice workflow in Microscopy Masters seems to be buggy still on the staged branch. I'm seeing a `this.setState is not a function` being thrown when trying to toggle the expert classification in the summary. There is an expert classification set in the state, so it should work. Not sure what's going on yet.

eatyourgreens:
Thanks. I'd forgotten to bind the `toggleExpertClassification` event handler. That should be fixed now, along with the other comments.

app/classifier/classifier.jsx
```diff
@@ -26,6 +26,7 @@ export default class Classifier extends React.Component {
     this.handleAnnotationChange = this.handleAnnotationChange.bind(this);
     this.handleSubjectImageLoad = this.handleSubjectImageLoad.bind(this);
     this.completeClassification = this.completeClassification.bind(this);
+    this.toggleExpertClassification = this.toggleExpertClassification(this);
```
srallen: Typo here, should be `this.toggleExpertClassification.bind(this)`

eatyourgreens: D'oh! Fixed.

PR #3650, opened 2017-03-24
If subject favorite property is missing, dont try any other API requests

srallen:
This is the only behavior that came up as a question while doing testing with the features that can be flipped on and off with the API. Features that can be flipped: https://github.com/zooniverse/Panoptes/wiki/Feature-Flags

If `skip_subject_selection_context` is flipped on, then the favorites property will be missing from the subject. Currently, if it is missing, then the fallback behavior is to request the subjects endpoint to check if the subject is present in the collection. I've removed this API call because the only time the favorite property will be missing is if this is flipped on and we're under heavy load. We probably don't want to be making any more API requests during that time if it's not absolutely necessary.

https://flipper-testing.pfe-preview.zooniverse.org/
In order to test this, the `skip_subject_selection_context` needs to be flipped on. The only thing odd about this is that if someone views their favorites collection, then none of the heart icons will be lit up. If a subject is already a favorite and someone clicks the button, the API will error 400, but it will appear to work for the volunteer.

An alternative would be to check if the subject id is present the favorites collection's `links.subjects` array? 

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -159,6 +159,7 @@ Classifier = React.createClass
         user={@props.user}
         project={@props.project}
         subject={@props.subject}
+        isFavorite={@props.subject.favorite}
         workflow={@props.workflow}
         preferences={@props.preferences}
         classification={currentClassification}
--- a/app/collections/favorites-button.cjsx
+++ b/app/collections/favorites-button.cjsx
@@ -39,12 +39,9 @@ module.exports = React.createClass
       .get({project_ids: @props.project?.id, favorite: true, owner: @props.user.login})
       .then ([favorites]) -> if favorites? then favorites else null
 
-  findSubjectInCollection: (favorites) ->
+  findSubjectInCollection: () ->
     if @props.subject.hasOwnProperty 'favorite'
       Promise.resolve @props.subject.favorite
-    else if favorites?
-      favorites.get('subjects', id: @props.subject.id)
-        .then ([subject]) -> subject?
     else
       Promise.resolve false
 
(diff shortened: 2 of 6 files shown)


camallen:
If the subject is showing up in the favourites collection (we know it's a favourite by the collection attributes) then the heart could automagically be set to on...right? 

Agree re the extra api call, be nice to not include it but @eatyourgreens had a reason for doing this and I think it was talk related.

eatyourgreens:
I think we should pass the initial state of the button into this component from its parents: `subject.favorite` for the Classifier, `true` for Favorites and `false` for Talk and other collections.

Speaking of Talk,  `componentWillMount` is going to run `findFavoriteCollection` and hit the API on every Talk subject discussion, since they unmount and remount the Subject Viewer.

Those calls aren't actually needed until you click on the Favourite button. I'm going to update this PR today.

eatyourgreens:
I've updated the PR. Ended up just rewriting the API calls so that they should only be invoked when you press the Favourites button.

Components are now expected to pass the initial state into the subject viewer as a boolean. I've left Talk as `false` so the heart won't be highlighted when you look at your faves on the discussion pages. That might be confusing.

eatyourgreens:
https://flipper-testing.pfe-preview.zooniverse.org/ is updated with my changes too.

eatyourgreens:
Oh, if I create an ordinary collection and put some favourites in, the heart won't show up highlighted there either.

srallen:
@eatyourgreens I just made some updates that fixes the favorite button on the subject talk page. I'm not sure what to do about a favorite that is in a normal collection  yet. Subjects requested there don't have the serialized favorites property and the collection isn't a favorite. That's the one case I think requesting for the favorite collection that belongs to the user is necessary.

srallen:
@eatyourgreens Favorites button should be working everywhere now: normal collections, favorite collections, subject talk page, and classifier

eatyourgreens:
Thanks. I think this is ready to merge.

eatyourgreens:
I'll leave the branch up in case any issues arise tomorrow.

PR #3655, opened 2017-03-27
Collection settings ES6 conversion and tests

srallen:
I decided to do some ES6 conversions and test writing for collection settings since we don't have the bandwidth to do some of the collection feature implementation at the moment.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


app/collections/settings.jsx
```diff
@@ -0,0 +1,161 @@
...
+    this.deleteCollection = this.deleteCollection.bind(this);
+    this.handleToggle = this.handleToggle.bind(this);
+    this.redirect = this.redirect.bind(this);
+  }
+
+  componentDidMount() {
+    this.props.collection.listen('delete', this.redirect);
+  }
+
+  componentDidUnMount() {
```
wgranger: Is this now a valid lifecycle method?

app/collections/settings.jsx
```diff
@@ -0,0 +1,161 @@
...
+              <button className="major-button" disabled={true}><i className="fa fa-spinner" /></button>
+              {' '}
+            </div>}
+
+          {!this.state.isDeleting &&
+            <div>
+              <button className="major-button" onClick={handleDelete}>Yes, delete it!</button>
+              {' '}
+              <button className="minor-button" onClick={resolve}>No, don't delete it.</button>
+            </div>}
```
eatyourgreens: One of these buttons should have `autoFocus={true}`.

app/collections/settings.jsx
srallen: Yep it's a typo on my part.

app/collections/settings.jsx
srallen: Added it to the cancel since that made sense to me. But could add it to the confirm button instead.

srallen:
Addressed comments and cleaned up some linter stuff that wasn't configured correctly.

app/collections/settings.jsx
eatyourgreens: Cancel makes sense to me. Confirm runs the risk of accidentally pressing Enter and deleting the collection.

Maybe reverse the order of the buttons, so you can tab from cancel to confirm? Sorry to be a pain about this.

srallen: Not a big deal to reverse. Makes sense.

srallen:
Staged at https://collection-settings.pfe-preview.zooniverse.org/

wgranger:
LGTM

PR #3660, opened 2017-03-28
Update panoptes-client to 2.7

eatyourgreens:
Describe your changes.
Adds `http_cache=true` to all API requests

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://http-caching.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -19,7 +19,7 @@
     "markdownz": "~7.0",
     "modal-form": "~2.4",
     "moment": "~2.9.0",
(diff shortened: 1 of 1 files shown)


camallen:
@srallen @eatyourgreens if you are both happy with this not impacting anything i'd like to have it available. Also if we're doing this let's do it now and test it before luanch.

PR #3667, opened 2017-03-29
Cache active users in a shared store.

eatyourgreens:
Fixes: ActiveUsers component spamming Panoptes to look up users

Describe your changes.
Uses a shared variable to store the user cache.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://active-user-cache.pfe-preview.zooniverse.org/

## Optional


eatyourgreens:
A better, long term solution would be to rewrite Talk to share single instances of the sidebar components. At the moment, each page component creates its own instance of `ActiveUsers`, which initialises with no stored data.

PR #3670, opened 2017-03-29
Fix fetch for iOS webview on App Status

### LINKED ISSUE #3668: Modify AppStatus to Allow Mobile Webview
## Expected behavior
Mobile webview should work properly on a native app.

## Current behavior
The native webview component cannot identify the fetch function [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/app-status.jsx#L47), which causes the native app to continuously load in iOS.

## Steps to replicate
In the mobile app, select a project with a workflow that takes you the webview (as opposed to switching to Safari).

## Additional information
A mobile [issue](https://github.com/zooniverse/mobile/issues/104) has been opened, but the issue has been traced to the `AppStatus` component in PFE.

- **Operating system:**
- **Browser:**

robinschaaf:
Fixes #3668 and https://github.com/zooniverse/mobile/issues/104

  *  Add a test to check if the fetch() function exists, needed to display in the mobile app webview, for iOS < 10.3.  If fetch() doesn't exist it will perform an XHR instead.
  *  Clean up some linter issues (separate commit)

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/partials/app-status.jsx
+++ b/app/partials/app-status.jsx
@@ -3,7 +3,7 @@
 ===================
 
 The AppStatus banner has one job: it displays a static message to the users, if
-it detects a non-empty "status message" file at a specified static URL. 
+it detects a non-empty 'status message' file at a specified static URL. 
 (The static resource is defined by the hardcoded APP_STATUS_URL.)
 
 Intended use: Zooniverse admins can manually change the status file (e.g. via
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Quick thoughts:
* Code looks solid; both `fetch()` and `XMLHttpRequest` sections work correctly on OSX+Chrome for the following cases: status file exists and has text; status file exists and has no text; status file does not exist. (Tested by forcing a false on the switching condition and by altering the URLs on localhost, respectively)
* Unfortunately I have no way to test if the `(typeof fetch === 'function')` condition itself executes properly in the mobile app webview, but the logic looks solid enough to me.
* Testing continues for OSX+FF/Safari and Win7+IE11
* This PR needs to be followed up with a second one that standardises Lab Status with this code, and hopefully a design update for App Status. (This are separate things and not blocking this PR)

wgranger:
After amending the assumptions in the comments I think this will be good to go.

PR #3672, opened 2017-03-29
Type checking for sim metadata

srallen:
The ABC SGL project subject metadata is using a string instead of a boolean for the sim metadata. This adds in a bit of type checking. 

https://sim-feedback-type-bug.pfe-preview.zooniverse.org/projects/ianc2/exoplanet-explorers/classify?env=production

And on staging (flower subject is a sim with a boolean value of true):
https://sim-feedback-type-bug.pfe-preview.zooniverse.org/projects/srallen086/sgl-tester/classify

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/classification-summary.jsx
+++ b/app/classifier/classification-summary.jsx
@@ -24,6 +24,18 @@ class ClassificationSummary extends React.Component {
     this.hasExpert = !!this.props.expertClassification;
   }
 
+  isSubjectASim() {
+    const simMetadataField = this.props.subject.metadata['#sim'];
+
+    if (typeof simMetadataField === 'string') {
+      return simMetadataField.toLowerCase() === 'true';
+    } else if (typeof simMetadataField === 'boolean') {
+      return simMetadataField;
+    }
+
+    return false;
+  }
+
   render() {
     const tools = this.props.project.experimental_tools || [];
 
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Code looks good, although I'm still running a few tests.

Note: https://sim-feedback-type-bug.pfe-preview.zooniverse.org/projects/ianc2/k2-candidate-vetting/classify?env=production doesn't seem to work - I assume the correct URL is https://sim-feedback-type-bug.pfe-preview.zooniverse.org/projects/ianc2/exoplanet-explorers/classify?env=production

srallen:
Yep, I'll update the original post. The project got renamed after I submitted this PR

shaunanoordin:
Also, two additional (but non-blocking) comments on the "This was a simulated planet!" message:

* [UI] Do we want to make the message more prominent? I'm not sure how much we want to alert the users to the fact that if it was a simulated planet, but if it's meant to be important, then the colours might need to be tweaked to add more contrast from the 'normal' white-coloured messages.
* [Text] I'm checking with @mrniaboc whether _only planets_ are ever simulated, or if the science team also simulated _non-planets_ . (i.e. intentional false results for calibration) If the latter, we may need to change the message to a generic "This was a simulated result!"

These are separate PR items, but I wanted to make a note of them here.

srallen:
I've been discussing with @mrniaboc as well about some UI changes: to swap the 'Your classification:' header with the sim message and ideally to remove the 'Your classification' part entirely, which may take a bit more work.

srallen:
> I'm not sure if we can always rely on the fact that .metadata exists in this.props.subject.metadata['#sim']. If we're sure we won't run into null/undefined errors due to the way the component is set up, then I'm happy to give this PR a thumbs up.

This is why I put it behind a workflow configuration flag. In the lab I added a message about the expected convention and require the feature to be explicitly turned on for it to work. This will need some revision for use with other projects, but I decided that it'd be best to wait for @rogerhutchings to consider this need for the generalized metadata feedback he is working on.

srallen:
I went ahead and swapped the header and the sim message to come in with this PR since it's a simple change.

PR #3673, opened 2017-03-29
Update Lab Status to match App Status + improve App Status style

shaunanoordin:
## PR Overview
* Follows up on #3670 
* The code for Lab Status has been updated to match App Status, in that it uses XMLHttpRequest as a fallback. Logically, we shouldn't need to worry about the Lab being used in an iOS native app environment(...?) but hey, we should keep the code consistent.
* Also, the style for the App Status message has been made more prominent.
screen shot 2017-03-29 at 22 52 03

In the case of the alert messages, @beckyrother if you have some spare time, I'm keen to know your thoughts about standardising the UI/UX design for 'alert messages', 'info messages', and similar popups/modulars. (Or whatever those component types are called.)

You can see an example of one alert message here on the homepage - https://update-app-status-lab-status.pfe-preview.zooniverse.org/ - and an alert message on the Build a Project/Lab page - https://update-app-status-lab-status.pfe-preview.zooniverse.org/lab

(If you don't have time, no worries, we'll use the current styles for now and revisit the visual design another day.)

# Review Checklist

- [x] Does it work on mobile?

### DIFF
--- a/app/partials/lab-status.jsx
+++ b/app/partials/lab-status.jsx
@@ -4,7 +4,7 @@
 
 A spin-off of the AppStatus component, for the Lab page.
 
-(shaun.a.noordin 20170327)
+(shaun.a.noordin 20170329)
 ********************************************************************************
  */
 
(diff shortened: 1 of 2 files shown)


beckyrother:
Thanks @shaunanoordin! This is definitely something I'd like to revisit, but for now let's just try to keep them looking consistent across the site. I think these are ok, but why are they different sizes?

shaunanoordin:
Ah, the different sizes! Here's what's happening:

screen shot 2017-03-30 at 16 15 48
**Lab Status** on Projects List: the component takes up 80% of the _screen_ width.

screen shot 2017-03-30 at 16 14 28
**Lab Status** on the Individual Project page: the component is squeezed into the _right-hand panel_ of the Project Builder settings, taking up 80% of the _panel_ width

Unfortunately due to the way that the components are structured, the Lab Status in the Individual Project page couldn't be fit into a more consistent layout, (i.e. outside the right-hand panel) to match Lab Status in the Project List.

screen shot 2017-03-30 at 16 14 52
**App Status** on any page: this is a separate component altogether so keeping a consistent size with Lab Status wasn't factored in.

wgranger:
Probably best to remove the `hide()` function here since it doesn't look like it's being used by LabStatus.

shaunanoordin:
Acknowledged - `hide()` is now gone!

shaunanoordin:
👍 Thanks! I've reset the LabStatus message for staging, so we should all be back to normal now.

shaunanoordin:
Well, after we account for caching of course. _starts hammering Cmd+R_

PR #3675, opened 2017-03-30
Only show the sim feedback if subject is a sim

srallen:
This changes the sim feedback to only show that message and not also the default classification summary. ~~There's a chance that the message itself might change, so I've marked this as WIP.~~ 

https://only-show-sim-feedback.pfe-preview.zooniverse.org/projects/srallen086/sgl-tester/classify

https://only-show-sim-feedback.pfe-preview.zooniverse.org/projects/ianc2/exoplanet-explorers/classify?env=production

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/classification-summary.jsx
+++ b/app/classifier/classification-summary.jsx
@@ -81,6 +81,12 @@ class ClassificationSummary extends React.Component {
       );
     }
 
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Noted on the WIP status. 👍 From a look through the code, the functionaly's LGTM, but let me run it on localhost so I can see if those external `<div>`s were important for styling or etc.

Issue #3680, opened 2017-03-30
Returning to Classify page after Talk presents users with previous classification summary


simensta:
## Expected behavior
If a user presses the "Talk" button after finishing a classification, the user should be able to press "CLASSIFY", in the navigational header, and return to the classify page with a new subject.

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
If a user presses "Talk" after finishing a classification and, then, uses the "CLASSIFY" link to return to the classification interface, the user is presented with the classification summary.

Pressing "Next" fetches a new subject, but any drawn marks remain and the summary remain.

## Steps to replicate
https://www.zooniverse.org/projects/zooniverse/decoding-the-civil-war/talk/465/226700?comment=408223

## Additional information

- **Operating system:** Mac OS
- **Browser:** Chrome

simensta:
This is only an issue when the Hide Classification Summary is NOT enabled. For example, you can replicate the issue in [Decoding the Civil War](https://www.zooniverse.org/projects/zooniverse/decoding-the-civil-war) (classification summary provided), but not in [Mapping Change](https://www.zooniverse.org/projects/zooniverse/mapping-change/classify) (classification summary hidden).

simensta:
I enabled the Hide Classification Summary feature for Decoding the Civil War, so you can no longer replicate the issue on that project.

eatyourgreens:
Looks like [`completeClassification`](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L298) should always call `onCompleteAndLoadAnotherSubject`. Easy to fix, but we should wait until #3647 is merged, and fix it there to avoid rebasing problems.

eatyourgreens:
On second thoughts, looking at that code, changing that might break Gravity Spy gold standards.

eatyourgreens:
[Annotations are persisting on Milky Way Project](https://www.zooniverse.org/talk/17/248605?comment=414431) too, so that may be a separate bug.

eatyourgreens:
Re. the Done & Talk button, there's a bug where `loadAnotherSubject` won't be called if [`saveClassification` isn't resolved](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/classify.cjsx#L232-L233) when the classify page unmounts.

simensta:
Fixed by #3681.

PR #3681, opened 2017-03-31
Load the next subject when clicking Talk link

eatyourgreens:
Fixes a bug where navigating to Talk leaves a completed classification stored in the project page state, which breaks the classify component on mount.

Describe your changes.
Adds an event handler to the Talk link so that it behaves exactly like the Next button.

Also takes a bunch of code that was waiting on `classification.save()` and runs it immediately instead.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-summary-nav.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -247,15 +247,15 @@ module.exports = React.createClass
     else
       classification.save()
 
+    Split.classificationCreated(classification)
+    {workflow, subjects} = classification.links
+    seenThisSession.add workflow, subjects
     savingClassification
       .then (classification) =>
         if @state.demoMode
           console?.log 'Demo mode: Did NOT save classification'
         else
           console?.log 'Saved classification', classification.id
-          Split.classificationCreated(classification)
-          {workflow, subjects} = classification.links
-          seenThisSession.add workflow, subjects
           classification.destroy()
         @saveAllQueuedClassifications()
       .catch (error) =>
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Fixes #3680

app/pages/project/classify.cjsx
```diff
@@ -230,7 +230,7 @@ module.exports = React.createClass
 
   saveClassificationAndLoadAnotherSubject: ->
     @saveClassification()
-      .then @loadAnotherSubject()
+    @loadAnotherSubject()
```
eatyourgreens: This is potentially dangerous, since loading a new subject destroys the current classification.

I think `saveClassification` creates a copy of the current classification, but I'm not sure.

eatyourgreens: This seems to work for classifications on Snapshots at Sea, and will fix the Milky Way Project.

simensta: Would passing `@state.classification` as an argument to `@saveClassification()` be safer? Then, we don't have to worry about `@loadAnotherSubject()` executing before `@saveClassication()` can [make a copy of the classification](https://github.com/zooniverse/Panoptes-Front-End/pull/3681/files#diff-83addd0f5535ee9ffe878b365f61712bL241).

eatyourgreens: [Line 241](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/classify.cjsx#L241) stores the classification in a local variable within the method, which should be okay since that won't be changed by loading a new subject and updating state.

If we add arguments to `saveClassification`, we'll have to find everything that invokes it as a callback, and ensure that those calls pass in a classification too.

PR #3684, opened 2017-03-31
Fix 3676

### LINKED ISSUE #3676: only render pan and zoom tools for image frames
## Expected behavior
The pan and zoom controls should only render for subject frames of the type image. The buttons are not functional on other subject frame types (e.g., video, text).

## Current behavior
The pan and zoom controls are rendered for all frames of a subject that is attached to a workflow with pan and zoom enabled. 

## Steps to replicate
Create workflow with pan and zoom enabled. Attach a subject set with both images and text subject frames. 

## Additional information

- **Operating system:**
- **Browser:**

simensta:
Fixes #3676. 

Describe your changes.
- The PanZoom buttons and eventListeners only render for subject frames of the type 'image'.
- For the TextViewer, `#onLoad()` is now called using a `ref`. In `componentDidMount` and `componentWillMount`, `ReactDOM.findDOMNode(this)` was returning null. The error from [`ReactDOM.findDOMNode(this).addEventListener('load', this.props.onLoad)` on line 31](https://github.com/zooniverse/Panoptes-Front-End/compare/fix-3676?expand=1#diff-f8270d0c4ee56789ba85382262b11d88L31) prevented classification submission for text subject frames.

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://fix-3676.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/frame-viewer.cjsx
+++ b/app/components/frame-viewer.cjsx
@@ -50,7 +50,7 @@ module.exports = React.createClass
         <TextViewer  src={src} type={type} format={format} frame={@props.frame} onLoad={@handleLoad} /> 
 
     if FrameWrapper
-      <PanZoom ref="panZoom" enabled={zoomEnabled} frameDimensions={@state.frameDimensions}>
+      <PanZoom ref="panZoom" enabled={zoomEnabled} frameType={type} frameDimensions={@state.frameDimensions}>
         <FrameWrapper 
           frame={frame} 
           naturalWidth={@state.frameDimensions?.width or 0} 
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -42,19 +42,19 @@ const PanZoom = React.createClass({
   componentDidMount() {
     // these events enable a user to navigate an image using arrows, +, and - keys,
     // while the user is in pan and zoom mode.
-    if (this.props.enabled && this.props.frameType === "image") {
+    if (this.props.enabled && this.props.frameType === 'image') {
       addEventListener('keydown', this.frameKeyPan);
       addEventListener('wheel', this.frameKeyPan);
     }
   },
 
-  componentWillReceiveProps(nextProps){
+  componentWillReceiveProps(nextProps) {
     if (nextProps.enabled) {
-      if (nextProps.frameType === "image"){
+      if (nextProps.frameType === 'image') {
         // when the frame is an image, add the pan-zoom event listeners
         addEventListener('keydown', this.frameKeyPan);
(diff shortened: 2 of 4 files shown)


app/components/frame-viewer.cjsx
```diff
@@ -50,7 +50,7 @@ module.exports = React.createClass
         <TextViewer  src={src} type={type} format={format} frame={@props.frame} onLoad={@handleLoad} /> 
 
     if FrameWrapper
-      <PanZoom ref="panZoom" enabled={zoomEnabled} frameDimensions={@state.frameDimensions}>
+      <PanZoom ref="panZoom" enabled={zoomEnabled} frameType={type} frameDimensions={@state.frameDimensions}>
```
eatyourgreens: Should this render if `type` isn't image?

app/components/pan-zoom.jsx
```diff
@@ -41,12 +42,26 @@ const PanZoom = React.createClass({
   componentDidMount() {
     // these events enable a user to navigate an image using arrows, +, and - keys,
     // while the user is in pan and zoom mode.
-    if (this.props.enabled) {
+    if (this.props.enabled && this.props.frameType === 'image') {
```
eatyourgreens: Maybe drop the `frameType` prop and just set `enabled` to false in the parent?

app/components/pan-zoom.jsx
```diff
@@ -77,7 +92,7 @@ const PanZoom = React.createClass({
     return (
       <div>
         {children}
-        {this.props.enabled ?
+        {(this.props.enabled && this.props.frameType === 'image') ?
```
eatyourgreens: Again, why not just set `enabled` to false in the parent?

eatyourgreens:
I would enable/disable the component in the frame viewer, based on subject type, rather than adding a new prop to check.

Alternatively, we could rewrite the frame viewer so that pan-zoom isn't rendered except for images.

A subject's `onLoad` handler shouldn't be called until after the subject has loaded, but this change invokes the callback when `componentWillMount` runs. I think it's also running the callback even if there's an error loading the subject. The reason for this is to disable the classification interface so that Done can't be pressed unless the subject was fully loaded.

simensta:
Good point. Conditionally rendering either the PanZoom component or the PanZoom controls from the FrameViewer is preferable to passing another prop to PanZoom. 

I didn't fully understand the original use case for `onLoad`, now I see why this solution doesn't work. 

Since the TextViewer's undefined DOM node is being solved in PR #3689 and the approach to rendering PanZoom will be different, I am going to close this PR and open a new one for #3676.

PR #3687, opened 2017-04-03
Add annotation check to exoplanets feedback

eatyourgreens:
Describe your changes.
Adds an annotation check so that SGL exoplanets feedback only shows depending on the answer to the question.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/classification-summary.jsx
+++ b/app/classifier/classification-summary.jsx
@@ -91,7 +91,7 @@ class ClassificationSummary extends React.Component {
 
     if (this.showExoplanetSimFeedback()) {
       return (
-        <p style={{ fontWeight: 'bold' }}>This was a simulated planet. We include these to help calibrate the project. Keep going to discover a real planet!</p>
+        <p style={{ fontWeight: 'bold' }}>Well done! You've found a simulated planet. We include these to help calibrate the project. Keep going to discover a real planet!</p>
       );
     }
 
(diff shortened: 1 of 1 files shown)


app/classifier/classification-summary.jsx
```diff
@@ -36,6 +36,14 @@ class ClassificationSummary extends React.Component {
     return false;
   }
 
+  showExoplanetSimFeedback() {
+    return (this.props.workflow.configuration &&
+      this.props.workflow.configuration.sim_notification &&
+      this.isSubjectASim() &&
+      this.props.classification.annotations[0].value === 0
```
eatyourgreens: That's a very good question. I think the exoplanets project requires an answer before you can classify.

shaunanoordin: In that case, I'm OK to leave this as is. There's no logical way to reach this section of the code unless the project is Exoplanets Explorer, and if EE always enforces an answer, then we won't need to worry about nulls.

shaunanoordin:
I've just checked with Grant and Marten - yeah, apparently we have a flooooood of sims in the Subject sets. That's not related to the front-end at all, so I have no further issues to investigate. This PR is LGTM.

srallen:
I updated the tester project I made on staging so that it works with this now. The reason it didn't work is that I had a multiple choice task type setup and now that the annotation is being checked, the task type needs to match what is expected.

PR #3688, opened 2017-04-03
[Wait to Merge]Add Collection functionality to add-delete multiple items

### LINKED ISSUE #3483: Collection subject view UI
## Expected behavior
Adding to a collection needs better design and discoverability. Should there be an interface on the collection view page separate from the button in the subject viewer?

## Current behavior
Currently the only way to add a subject to a collection is to use the button available in the subject viewer and it opens up a modal dialog to search for a collection/create a collection and then add the subject to it.

mcbouslog:
WAIT TO MERGE - wait for notice to be given to users before merging.

Fixes #3483.

Changes to general design of a collection.
Add functionality to select then add or remove multiple subjects in a collection.

I've been testing in this project's collections:
https://collections-add-delete-multi-subjects.pfe-preview.zooniverse.org/projects/markb-panoptes/test-project-mb/collections

Remaining Open Items:
- [x] card border
- [x] circle drop-shadow
- [x] test with video and flipbook subjects

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://collections-add-delete-multi-subjects.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/collections/show-list.cjsx
+++ b/app/collections/show-list.cjsx
@@ -74,21 +74,26 @@ SubjectNode = React.createClass
     logClick = @context.geordi?.makeHandler? 'about-menu'
     <div className="collection-subject-viewer">
       <SubjectViewer defaultStyle={false} subject={@props.subject} user={@props.user} project={@state.project} isFavorite={@state.isFavorite}>
-          <Link className="subject-link" to={"/projects/#{@state.project?.slug}/talk/subjects/#{@props.subject.id}"} onClick={logClick?.bind(this, 'view-favorite')}>
-            <span></span>
-          </Link>
+          {if !@props.selecting
+            <Link className="subject-link" to={"/projects/#{@state.project?.slug}/talk/subjects/#{@props.subject.id}"} onClick={logClick?.bind(this, 'view-favorite')}>
+              <span></span>
+            </Link>}
           {if @isOwnerOrCollaborator() and !@props.selecting
             <button type="button" className="collection-subject-viewer-delete-button" onClick={@props.onDelete}>
               <i className="fa fa-close" />
             </button>}
           {if @props.selecting and !@props.selected
-            <button className="collection-subject-viewer-circle-open" onClick={@handleSelect.bind @, 'add'}>
-              <i className="fa fa-circle-o" />
-            </button>}
+            <div className="collection-subject-viewer-selecting">
+              <button className="collection-subject-viewer-select" onClick={@handleSelect.bind @, 'add'}>
+                <i className="fa fa-circle-o" />
+              </button>
+            </div>}
           {if @props.selecting and @props.selected
-            <button className="collection-subject-viewer-circle-open" onClick={@handleSelect.bind @, 'delete'}>
+            <div className="collection-subject-viewer-selecting">
+              <button className="collection-subject-viewer-select" onClick={@handleSelect.bind @, 'delete'}>
               <i className="fa fa-check" />
-            </button>}
+              </button>
+            </div>}
       </SubjectViewer>
     </div>
 
--- a/app/collections/show-list.cjsx
+++ b/app/collections/show-list.cjsx
@@ -4,6 +4,7 @@ Translate = require 'react-translate-component'
 apiClient = require 'panoptes-client/lib/api-client'
 intersection = require 'lodash.intersection'
 pick = require 'lodash.pick'
+classNames = require 'classnames'
 counterpart = require 'counterpart'
 alert = require '../lib/alert'
 Paginator = require '../talk/lib/paginator'
@@ -64,6 +65,11 @@ SubjectNode = React.createClass
 
   render: ->
     logClick = @context.geordi?.makeHandler? 'about-menu'
+    subjectSelectClasses = classNames({
+      "collection-subject-viewer-circle": true,
+      "fa fa-check-circle": !!@props.selected,
+      "fa fa-circle-o": !@props.selected
+    })
     <div className="collection-subject-viewer">
       <SubjectViewer defaultStyle={false} subject={@props.subject} user={@props.user} project={@state.project} isFavorite={@state.isFavorite}>
           {if !@props.selecting
@@ -77,7 +83,7 @@ SubjectNode = React.createClass
           {if @props.selecting
             <label className="collection-subject-viewer-select">
               <input type="checkbox" checked={@props.selected} onChange={@toggleSelect}/>
-              <i className={"collection-subject-viewer-circle " + if @props.selected then "fa fa-check-circle" else "fa fa-circle-o"} />
+              <i className={subjectSelectClasses} />
             </label>}
       </SubjectViewer>
     </div>
@@ -150,9 +156,7 @@ module.exports = React.createClass
 
   promptCollectionManager: ->
     alert (resolve) =>
-      <CollectionsManager user={@props.user} project={@props.project} subjectIDs={@state.selected} onSuccess={resolve} />
-
-    @toggleSelecting()
+      <CollectionsManager user={@props.user} project={@props.project} subjectIDs={@state.selected} onSuccess={() => @toggleSelecting(); resolve();} />
 
   handleDeleteSubject: (subject) ->
     subjects = @state.subjects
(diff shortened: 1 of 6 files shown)


app/collections/show.jsx
```diff
@@ -85,41 +83,59 @@ const CollectionPage = React.createClass({
     const profileLink = `${baseLink}/users/${this.state.owner.login}`;
     const collectionsLinkMessageKey = `collectionPage.${baseType}Link`;
 
+    let roles;
+    if (this.props.roles.some(collectionRole => collectionRole.links.owner.id === this.props.user.id)) {
```
srallen: This check doesn't seem necessary. You could eliminate the `roles` variable, set `userRole` to the returned value of the filter and L102 check for the length of `userRole`, then call a join on `userRole[0].roles`

app/collections/show-list.cjsx
```diff
@@ -45,20 +47,32 @@ SubjectNode = React.createClass
...
+          {if !@props.selecting
+            <Link className="subject-link" to={"/projects/#{@state.project?.slug}/talk/subjects/#{@props.subject.id}"} onClick={logClick?.bind(this, 'view-favorite')}>
+              <span></span>
+            </Link>}
+          {if @isOwnerOrCollaborator() and !@props.selecting
+            <button type="button" className="collection-subject-viewer-delete-button" onClick={@props.onDelete}>
+              <i className="fa fa-close" />
+            </button>}
+          {if @props.selecting and !@props.selected
+            <button className="collection-subject-viewer-select-button" onClick={@handleSelect.bind @, 'add'}>
```
srallen: Instead of two set of buttons, could a checkbox type input be considered with a single handler and the checked value be used to determine whether or not to add or remove? It could be wrapped in a label with the font awesome icons and a conditional to determine which to show based on the presence of the selected subject id in the state?

app/collections/show-list.cjsx
```diff
@@ -140,14 +198,37 @@ module.exports = React.createClass
...
+              {if @state.collectionsManaging
+                <Dialog tag="div" closeButton={true} onCancel={@closeCollectionsManager}>
+                <CollectionsManager user={@props.user} project={@props.project} subjectIDs={@state.selected} onSuccess={@closeCollectionsManager} />
+                </Dialog>}
+              <button className="select-images-button" onClick={@deleteSubjects} disabled={@state.selected.length < 1}>Remove from Collection</button>
+              <button className="select-images-button" onClick={@toggleSelect}>Cancel</button>
+            </div>
+          else
+            <div className="collection-buttons-container">
+              <button className="select-images-button" onClick={@toggleSelect}>Select Images</button>
```
srallen: This should probably say `Select Subjects`

app/collections/show-list.cjsx
```diff
@@ -140,14 +198,37 @@ module.exports = React.createClass
...
+                className="select-images-button"
+                onClick={@openCollectionsManager}
+                disabled={@state.selected.length < 1}>
+                <span>Add to Collection</span>
+              </button>
+              {if @state.collectionsManaging
+                <Dialog tag="div" closeButton={true} onCancel={@closeCollectionsManager}>
+                <CollectionsManager user={@props.user} project={@props.project} subjectIDs={@state.selected} onSuccess={@closeCollectionsManager} />
+                </Dialog>}
+              <button className="select-images-button" onClick={@deleteSubjects} disabled={@state.selected.length < 1}>Remove from Collection</button>
```
srallen: The remove from collection button doesn't make sense if I'm viewing someone else's collection. It probably should only be available if I have appropriate permissions to edit that collection. I was able to use it and it looked like I _could_ and _did_ delete some subjects from a collection that I do not have permissions for. Refreshing the page showed they weren't actually deleted.

mcbouslog:
Questions:
1. after selecting a few images and adding (actually adding or canceling) or deleting currently it's 
  A.coded so the subjects are deselected (if adding, N/A if deleting) and it's no longer in selecting state (top-right button is back to Select Subjects), however other options are
  B. the subjects are deselected but remains in selecting state
  C. the subjects remain selected and it remains in selecting state
2. the Collections Manager dialog doesn't allow the select menu datalist to be fully visible. I tried adding `overflow: auto` and variations in different ways including to a `select-menu` class nested under `collections-manager`, but was unable to fix

trouille:
@beckyrother Could you post here whether you want to message in Talk for volunteers in projects where collections are used heavily  about this design/functionality update before this is merged? We'd talked about that earlier this week. Did you find out from @zwolf what projects are top of the list in collection usage? 

If yes, please change this PR title to [WIP] so we know to wait to merge until you give the thumbs up.

app/collections/show-list.cjsx
```diff
@@ -140,14 +184,35 @@ module.exports = React.createClass
           meta = @state.subjects[0].getMeta()
 
           <div>
-            <div className="collection-subjects-list">
+          {if @state.selecting
```
mcbouslog: or maybe disable the Select Subjects button if signed-out?
or maybe a better question is what behavior is expected if signed-out? if I make the Add to Collection button not visible to a signed-out user, they can still click the Select Subjects button and the circles show, but it's not clear what's going on.

srallen: You can only add or remove from collections if you are signed in, so I don't think these buttons should be visible at all if not signed in.

app/collections/show-list.cjsx
```diff
@@ -140,14 +184,35 @@ module.exports = React.createClass
...
+            <div className="collection-buttons-container">
+              <button
```
mcbouslog: fixed, see line below, and other buttons as well

app/collections/show-list.cjsx
mcbouslog: sounds good -> updated so the Select Subjects button won't show unless signed in.

srallen:
Re: Collections Manager. UI fixes for that can come in a separate PR. Let's keep this scoped to the page itself.

srallen:
One other thing: the sub nav on that page for settings and collaborators are missing an active classname, so without it, there is no indication that you are on the settings or collaborators page in the nav.

css/collections-page.styl
```diff
@@ -27,10 +70,16 @@
       .subject-video-controls
         display: initial
 
+      .subject-image-frame
```
mcbouslog: this is per here: https://github.com/zooniverse/Panoptes-Front-End/blob/c6a6a0e9569adeaeb9a613d358e94e0f235558eb/app/components/frame-viewer.jsx#L51 and I think `image` is intended

srallen: Ah ok. Thanks for clearing that up.

srallen:
Couple more UI/UX suggestions, but not blockers: 

There probably should be a confirmation on removing the subjects in case it is an accidental button press. Just an alert with a confirm and cancel should be sufficient. 

@eatyourgreens also reminded me of this in #3717. The buttons and checkbox input that use font awesome icons instead of text should have an `aria-label` attribute for accessibility.

mcbouslog:
argh I should have remembered the aria-labels! sorry about that.

I like the confirm before deleting multiple subjects. I copied the delete confirm from the collections settings and updated for this purpose, but let me know if something else preferred.

app/collections/show-list.cjsx
```diff
@@ -168,7 +176,16 @@ module.exports = React.createClass
       .then =>
         @props.collection.uncacheLink 'subjects'
 
-  deleteSubjects: () ->
+  confirmDeleteSubjects: ->
+    alert (resolve) =>
+      <div className="confirm-delete-dialog content-container">
+        <p>Are you sure you want to delete {@state.selected.length} subjects from this collection? This action is irreversible!</p>
```
srallen: Let's make this wording consistent with the button. How about `remove` instead of `delete`?

Also, I don't think `This action is irreversible` is necessary. Seems a bit dramatic :)

app/collections/show-list.cjsx
```diff
@@ -168,7 +176,16 @@ module.exports = React.createClass
...
+        <button className="minor-button" autoFocus={true} onClick={resolve}>No, don&apos;t delete the subjects.</button>
```
srallen: How about a simple `Cancel`?

app/collections/show-list.cjsx
```diff
@@ -168,7 +176,16 @@ module.exports = React.createClass
...
+        {' '}
+        <button className="major-button" onClick={() => @deleteSubjects(); resolve();}>Yes, delete the subjects selected!</button>
```
srallen: How about just `Yes`?

beckyrother:
@srallen, agree with those copy changes 👍 

Let's merge this on Monday morning (CST) around the same time we post the message about these changes.

PR #3689, opened 2017-04-04
Fix null or undefined DOM node in text viewer

eatyourgreens:
Rewrites the text viewer to fix a bug where it manipulates the DOM before it has rendered.
Sets a minimum width of 30em for text subjects.
Adds content caching, by URL, for text subjects.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-text-viewer.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/css/subject-viewer.styl
+++ b/css/subject-viewer.styl
@@ -51,6 +51,7 @@
     .text-viewer
       background-color: #404040
       color: white
+      min-width: 30em
       padding: 2%
       white-space: pre-wrap
 
--- a/css/subject-viewer.styl
+++ b/css/subject-viewer.styl
@@ -51,7 +51,7 @@
     .text-viewer
       background-color: #404040
       color: white
-      min-width: 30em
+      width: 96%
       padding: 2%
       white-space: pre-wrap
 
--- a/app/components/text-viewer.jsx
+++ b/app/components/text-viewer.jsx
@@ -35,7 +35,7 @@ class TextViewer extends Component {
       this.setState({ content: cachedContent });
     } else {
       this.setState({ content: 'Loading…' });
-      fetch(src, { mode: 'cors' })
+      fetch(src + '?=')
       .then((response) => {
         return response.text();
       })
--- a/app/components/text-viewer.jsx
+++ b/app/components/text-viewer.jsx
@@ -3,12 +3,14 @@ import React, { Component } from 'react';
 const SUPPORTED_TYPES = ['text'];
 const SUPPORTED_FORMATS = ['plain'];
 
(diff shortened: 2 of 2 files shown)


simensta:
In Chrome and Safari, the width for the subject seems to be set before the subject content is retrieved:
text subject width example
https://fix-text-viewer.pfe-preview.zooniverse.org/projects/scribe-2/textsubjectproject/classify?reload=1&workflow=2885

eatyourgreens:
@simensta this turned into a slightly bigger rewrite of the `TextViewer` component to make it more robust. I think it's ready now.

css/subject-viewer.styl
```diff
@@ -51,6 +51,7 @@
     .text-viewer
       background-color: #404040
       color: white
+      min-width: 30em
```
simensta: The `min-width` property is causing [the text-viewer div background to overlap the task area](https://fix-text-viewer.pfe-preview.zooniverse.org/projects/scribe-2/textsubjectproject/classify?reload=0&workflow=2885) in Chrome, Edge, and Safari. This may only be an issue when the first frame is text, instead of an image. One option for a workaround is #3712.

eatyourgreens: No overlap here, I'm afraid. Could this be caused by viewport width? In that case the sizing should maybe use `vw` instead of `em`.
Screenshot showing the text viewer side-by-side with the task area in Chrome

eatyourgreens: Aha, managed to reproduce it! Should be fixed by 8b663f5, which matches the width of the viewer to the width of its container.

eatyourgreens: Oops, that should have been f43d6e8. I missed a couple of uncommitted changes.

simensta:
Commit [f43d6e8](https://github.com/zooniverse/Panoptes-Front-End/pull/3689/commits/f43d6e804eecce63657ad644ee9aea45a3407a9b) fixes the size of the text viewer in Chrome and Edge. In Safari, the text viewer width is narrower than expected for [subjects that present a text frame first](https://fix-text-viewer.pfe-preview.zooniverse.org/projects/scribe-2/textsubjectproject/classify?reload=0&workflow=2885). 

Besides the Safari issue I think the PR is ready to merge. Since this PR already addresses current issues, do you think we should write up an issue for the Safari bug to be addressed separately?

eatyourgreens:
:+1: to merging this and getting it out of the way. I think we're going to have to address styling separately once we a set of real text subjects to work with anyway.

PR #3694, opened 2017-04-04
First to classify experiment

### LINKED ISSUE #3466: Setup subject.first-to-classify experiment in classifier itself for Gravity Spy
Experiment will happen in the classifier itself somewhere instead of summary since Gravity Spy does not use standard classification summaries.

## Additional information

- **Operating system:**
- **Browser:**

srallen:
Fixes #3466

This is the novelty experiment for Gravity Spy. Rather than remove the text split that was setup previously, this adds another kind of visibility split experiment that shows a banner informing the volunteer they are the first to see the subject prior to classification. Labeled as WIP since experiment still needs setting up on Seven-Ten staging. Will add staging link once it is working.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -111,24 +111,6 @@ export default class Classifier extends React.Component {
     this.setState({ annotations });
   }
 
-  loadClassificationsCount(subject) {
-    let query = {};
-    if (this.props.splits && this.props.splits['subject.first-to-classify']) {
-      query = {
-        workflow_id: this.props.workflow.id,
-        subject_id: subject.id
-      };
-    
-
-      apiClient.type('subject_workflow_statuses')
-      .get(query)
-      .then(([sws]) => {
-        const classificationCount = sws.classifications_count ? sws.classifications_count : 0;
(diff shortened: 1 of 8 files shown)


srallen:
https://first-to-classify-experiment.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing/classify is the staging project setup with a split running right now. You should see:

- Logged out: no splits
- Logged in: a split visbility variant of either `div: true` or `div: false`
  - If true and the subject has an empty array subject workflow status response, the `classificationCount` state is set to 0 and you should see a yellow banner above the classifier saying you're first to classify.
  - if true, but the subject workflow status response has a count, then you should not see the banner.
  - if false, then you should not see the banner

Split variant is set for the user when visiting the project and is what your split is for the entire duration of the experiment (meaning a refresh won't get you a different split). You can try logging in as a different user to try to get the other split or ask @zwolf to change the split for your user.

I have a subject set with subjects that have never been classified unlinked. Whenever you're ready to check this out, let me know, and I'll link up that subject set for this tester project.

PR #3701, opened 2017-04-06
Update webpack and associated dependencies

srallen:
While trying to figure out why the compiling for this [markdownz bug fix](zooniverse/markdownz/pull/53) I'm working on is failing, I decided to bump all of the webpack dependencies since webpack v. 1 is deprecated. The config is updated according to the [migration guide](https://webpack.js.org/guides/migrating/).

This is WIP because it is compiling locally, but I'm seeing the following error in console:
`Uncaught TypeError: asn.define is not a function`
In the stack trace this is in the `certificates.js` in the `parse-asn1` package (which is a dependency that is for the crypto shim webpack uses)

All I can reason is that webpack is failing to correctly resolve the require for the `asn1.js` dependency: `var asn = require('asn1.js')` in `certificates.js`. Here are the docs for [webpack resolve](https://webpack.js.org/configuration/resolve/). I haven't found a resolution yet, but have confirmed that the latest versions of the dependencies down the chain are installing and have scoured their respective github repo open issues. 

I looked into bypassing webpack's shimming of crypto and writing a bit of code to handle whether or not to use native node crypto or a browser shim (`crypto-browserify`), but I don't think this is the right direction to go since it is the module resolving appears to be failing.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


eatyourgreens:
I'm seeing that same error. Also, my user account never loads, top right, so maybe something in the API client auth file?

eatyourgreens:
Investigating this a little further, if I bypass `auth.checkCurrent()`, I can sign in manually without running into that error.

webpack.build.js
```diff
@@ -49,33 +46,54 @@ module.exports = {
...
+      }),
     }, {
       test: /\.styl$/,
-      loader: ExtractTextPlugin.extract('style-loader', 'css-loader?root=../public!stylus-loader'),
+      loader: ExtractTextPlugin.extract({
+        fallback: 'style-loader', 
+        use: [{
+          loader: 'css-loader'
+          options: { root: '../public' },
+        }, {
```
eatyourgreens: This line is throwing an error when I try a staging build. Should this block match the same block in `webpack.dev.js`?

webpack.build.js
```diff
@@ -49,33 +46,54 @@ module.exports = {
...
+        }
+      }),
     }, {
       test: /\.styl$/,
-      loader: ExtractTextPlugin.extract('style-loader', 'css-loader?root=../public!stylus-loader'),
+      loader: ExtractTextPlugin.extract({
+        fallback: 'style-loader', 
+        use: [{
+          loader: 'css-loader'
+          options: { root: '../public' },
```
eatyourgreens: This line is throwing an error when I try a staging build. Should this block match the same block in `webpack.dev.js`?

eatyourgreens:
Staged at https://update-webpack.pfe-preview.zooniverse.org
Submitting a classification fails because of the session ID bug.

webpack.build.js
```diff
@@ -49,33 +46,54 @@ module.exports = {
...
+          options: { root: '../public' }
+        }
+      }),
     }, {
       test: /\.styl$/,
-      loader: ExtractTextPlugin.extract('style-loader', 'css-loader?root=../public!stylus-loader'),
+      loader: ExtractTextPlugin.extract({
+        fallback: 'style-loader', 
+        use: [{
+          loader: 'css-loader'
```
eatyourgreens: never mind previous comment. There's a missing comma here.

eatyourgreens:
Updated to use a different implementation of SHA256. I think that should fix the `crypto` problem.

srallen:
Thanks @eatyourgreens I think I was going to look into somehow aliasing it an older version since this wasn't a problem with webpack 1 or replacing the crypto shim webpack is using, but you beat me to it. This was the only relevant issue I could find on the webpack repo: https://github.com/webpack/webpack/issues/4378

The `asn1.js` dependency wasn't resolving properly and was just an empty object. :woman_shrugging: 

The only thing left that I didn't take a look at was the hot-module configuration, though, I'm not sure who is using that. If no one is using it, I'd like to just remove it.

srallen:
I looked over the update for hot module replacement and it requires the tree shaking configuration which if I'm understanding correctly I think will break with our coffeescript components. Fortunately, the older hot module config still works. I'll leave it as is.

I tested the standard npm `start` script as well as `serve-static`, `serve-hot`, `stage`, and `test`. All seem ok. I didn't try the production deploy script or the docker deploy.

@wgranger could you test the change made to WWT to make sure it doesn't break anything?

srallen:
Staged link: https://update-webpack.pfe-preview.zooniverse.org

wgranger:
This appears to work fine with Worldwide Telescope. However, the Worldwide Telescope summary currently isn't showing up due to a change that needs to be made on [this](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/classification-summary.jsx#L80) line of the classification summary component. I'll make the change on a separate PR. Bottom line - this looks fine with WWT.

PR #3703, opened 2017-04-07
allow inactive workflows to be shown on the stats page

### LINKED ISSUE #3683: Statistics page not removing workflows that were active but are now inactive
## Expected behavior

Steller Watch PI brought up an interesting point that it would be useful to still be able to see the stats for workflows that were active, but are turned to inactive when their associated subject sets are completely retired. 

Could we find a way to keep those now inactive workflows showing within the stats page? Note: important will be to not show all 'inactive' workflows, since some projects will have inactive workflows that haven't had any classifications and so they won't want to clutter up the stats page with these non-interesting workflows. 

## Current behavior

Currently PFE is using an `active: true` filter on those workflow requests for stats

## Additional information

- **Operating system:**
- **Browser:**

CKrawczyk:
Fixes #3683.

This PR allows inactive workflows to be visible on a project's stats page by setting the visibility in the lab.  This also changes the check box on the lab visibility page from `Hide on Stats Page` to `Show on Stats Page`.  I have used the same `workflow.configuration.stats_hidden` parameters that was being used before, so this *should* not change the current look of any live project's stats pages.

Note: Some of the logic is a bit confusing since it defaults inactive workflows that don't have `workflow.configuration.stats_hidden` to be not visible.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://show-inactive-workflow-stats.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -61,9 +61,9 @@ module.exports = React.createClass
       ).then(() => @forceUpdate())
 
   handleWorkflowStatsVisibility: (workflow, e) ->
-    checked = !e.target.checked
+    hidden = !e.target.checked
 
-    workflow.update({ 'configuration.stats_hidden': checked }).save()
+    workflow.update({ 'configuration.stats_hidden': hidden }).save()
       .catch((error) =>
         @setState { error }
       ).then(() => @forceUpdate())
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -169,9 +169,9 @@ module.exports = React.createClass
             {@state.workflows.map (workflow) =>
               setting = workflow.active
               stats_completeness_type = workflow.configuration.stats_completeness_type ? 'retirement'
-              stats_visible = !workflow.configuration.stats_hidden
-              if !workflow.active
-                stats_visible = !(workflow.configuration.stats_hidden == undefined) && stats_visible
+              statsVisible = workflow.active
+              if workflow.configuration.stats_hidden != undefined
+                statsVisible = !workflow.configuration.stats_hidden
               <tr key={workflow.id}>
                 <td>
                   {workflow.id}
@@ -217,8 +217,8 @@ module.exports = React.createClass
                     <input
                       type="checkbox"
                       name="stats_hidden.#{workflow.id}"
-                      value={stats_visible}
-                      checked={stats_visible}
+                      value={statsVisible}
+                      checked={statsVisible}
                       onChange={@handleWorkflowStatsVisibility.bind(this, workflow)}
                     />
                     Show on Stats Page
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -61,7 +61,7 @@ module.exports = React.createClass
       ).then(() => @forceUpdate())
 
   handleWorkflowStatsVisibility: (workflow, e) ->
-    checked = e.target.checked
+    checked = !e.target.checked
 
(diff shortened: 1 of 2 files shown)


app/pages/lab/visibility.cjsx
```diff
@@ -61,7 +61,7 @@ module.exports = React.createClass
       ).then(() => @forceUpdate())
 
   handleWorkflowStatsVisibility: (workflow, e) ->
-    checked = e.target.checked
+    checked = !e.target.checked
```
eatyourgreens: This is very confusing. For clarity, I would name this variable `hidden` or something similar.

app/pages/lab/visibility.cjsx
```diff
@@ -169,6 +169,9 @@ module.exports = React.createClass
             {@state.workflows.map (workflow) =>
               setting = workflow.active
               stats_completeness_type = workflow.configuration.stats_completeness_type ? 'retirement'
+              stats_visible = !workflow.configuration.stats_hidden
+              if !workflow.active
+                stats_visible = !(workflow.configuration.stats_hidden == undefined) && stats_visible
```
eatyourgreens: I'm not sure what this comparison does. My reading is: if `stats_hidden` is undefined, then `stats_visible` will be true and this comparison is redundant. If `stats_hidden` is defined then this becomes `false && !stats_hidden` which is always false.

app/pages/project/stats/index.jsx
```diff
@@ -29,22 +29,26 @@ class ProjectStatsPageController extends React.Component {
...
     };
     getWorkflowsInOrder(project, query)
       .then((workflows) => {
         const workflowsSetToBeVisible =
-          workflows.filter((workflow) => { 
-            return (!workflow.configuration.stats_hidden ? workflow : null);
+          workflows.filter((workflow) => {
+            let stats_visible = !workflow.configuration.stats_hidden
+            if (!workflow.active) {
+              stats_visible = !(workflow.configuration.stats_hidden === undefined) && stats_visible
```
eatyourgreens: Duplicates the previous block of if logic. I'm not quite following what these comparisons do.

app/pages/lab/visibility.cjsx
CKrawczyk: This is the confusing bit that is needed so *inactive* workflows are hidden by default (i.e. `stats_hidden` is not set, but *active* workflows should be shown by default.

app/pages/lab/visibility.cjsx
CKrawczyk: Good point

CKrawczyk:
Just to clarify what is intended by these changes:

+ Inactive workflows should not be visible by default
+ Active workflows should be visible by default
+ Both defaults are overridden when the `Show on Stats Page` box is toggled in the lab

app/pages/lab/visibility.cjsx
eatyourgreens: If `stats_hidden` is not set then this line becomes `stats_visible=true` because of line 172, no? `!undefined` is true in JavaScript. That's what's confusing me.

eatyourgreens: I think you want to set the default using `stats_visible = workflow.active` then override from the checkbox setting.

CKrawczyk: There is a `!` in front of the first condition, so it becomes `false && true` and sets `stats_visible` to `false` (now that I think of it I should have used `!=`, my brain is being slow today).  But you make a good point that using `workflow.active` as the initial value is less confusing.

CKrawczyk:
I have cleaned up the logic and re-staged the branch.

PR #3712, opened 2017-04-10
fixes sizing of text viewer after loading content

simensta:
This is a work around to get the text-viewer to be the right size after loading the content of text frames. Setting a `min-width` property on the `text-viewer` is causing the subject to overlap the task area in Chrome, Safari, and Edge.

Describe your changes.
Instead of using `min-width` to control the size of the text-viewer, the text-viewer renders a different div when the content is loading than when the content is loaded. 

# Review Checklist

- [ ] Does it work on mobile?
- [X] Did you deploy a staging branch? https://text-viewer-sizing-from-fix-text-viewer.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/components/text-viewer.jsx
+++ b/app/components/text-viewer.jsx
@@ -59,11 +59,19 @@ class TextViewer extends Component {
     if (SUPPORTED_FORMATS.indexOf(this.props.format) === -1) {
       content = `Unsupported format: ${this.props.format}`;
     }
-    return (
-      <div ref={(element) => { this.element = element; }} className="text-viewer" >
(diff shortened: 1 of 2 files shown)


eatyourgreens:
I think this is going to mess up the event handlers. The listener on the old div should be removed before it's destroyed, and one added to the new div.

Alternatively, why not have one div and dynamically change the class? Or use a min width based on viewport dimensions (vw).

simensta:
Closing this as the solution in #3689 is preferable. Thanks @eatyourgreens!

PR #3713, opened 2017-04-11
Reset the default frame on subject change

### LINKED ISSUE #3700: When loading next subject, the frame should either be 0 or the default_frame property
## Expected behavior
In multi-frame subjects, when a new subject loads in the classification interface, the frame displayed should either be the first frame or the frame indicated by the `default_frame` property.

## Current behavior
When loading a new subject in the classification interface, the index of the last frame displayed of the previous subject is the index of the first frame displayed of the new subject.

## Steps to replicate
In a workflow with multi-frame subjects, select the last frame before pressing "Next" or "Done". When the new subject loads, it will display the last frame of that subject.

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Fixes #3700 .

Describe your changes.
Calls `getInitialFrame` if the subject changes in the subject viewer.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3700.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -71,6 +71,10 @@ module.exports = React.createClass
       clearTimeout @signInAttentionTimeout
       @setState loading: true
 
+  componentDidUpdate: (prevProps) ->
+    if @props.subject isnt prevProps.subject
+      @setState frame: @getInitialFrame()
+
   logSubjClick: (logType) ->
     @context.geordi?.logEvent
       type: logType


app/components/subject-viewer.cjsx
```diff
@@ -71,6 +71,10 @@ module.exports = React.createClass
       clearTimeout @signInAttentionTimeout
       @setState loading: true
 
+  componentDidUpdate: (prevProps) ->
```
simensta: Why place line 75-76 in `componentDidUpdate` instead of `componentWillReceiveProps`?

eatyourgreens: Because `getInitialFrame` acts on `this.props` so we wait until after the value updates.

PR #3714, opened 2017-04-11
Makes clearer that subjects don't get deleted

simoneduca:
Follow up to https://github.com/zooniverse/Panoptes/issues/1037#issue-91752692

Makes the text on the subject-set delete button less ambiguous.
Also, adds a note to the alert box.

- [x] Did you deploy a staging branch? https://edit-subjectset-text.pfe-preview.zooniverse.org/

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -241,7 +241,7 @@ EditSubjectSetPage = React.createClass
       <p>
         <small>
           <button type="button" className="minor-button" disabled={@state.deletionInProgress} onClick={@deleteSubjectSet}>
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks good to me, but raises the question of what does the end user do with those subjects, if they're no longer linked to a subject set, so not accessible via the UI.

@marten @camallen Do we delete orphaned subjects (ones that are no longer linked to any subject sets)?

amyrebecca:
I think that question probably is larger than this issue, although we should certainly not lose sight of it!

camallen:
> Do we delete orphaned subjects (ones that are no longer linked to any subject sets)?

Yes we do under normal conditions but only if the subject has not been used in the project, e.g. has no classifications and not talk comments, etc.

PR #3716, opened 2017-04-11
Fix workflow selection edge cases

srallen:
This fixes some workflow selection bugs in a couple of edge cases found with some non-launched approved projects:

- Fixes a `TypeError` when there are no active workflows for a project
- Fixes a loop attempting to fetch an inactive project default

This is a project that I purposely set an inactive workflow as default via console so you can see it now just errors once and stops attempting to load the project default forever:
https://fix-no-active-workflows-error.pfe-preview.zooniverse.org/projects/srallen086/more-testing

I don't want to clear out the project default from the project page since I can't assume the intention of the project owner (maybe they accidentally made the default workflow inactive before it was fixed in the lab?). The lab was fixed so this _can't_ happen in the lab anymore in #3552. There are possibly some edge case projects out there with inactive project default workflows, but they should go into the lab and fix it (make their default active or activate it, remove the default, and inactivate it). 

We should think about adding into the lab an automatic switch to dev mode if all of the workflows are made inactive for some reason and fix the messaging on the project home page if that happens (right now stuck with a loading message).

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -198,13 +198,13 @@ ProjectPageController = React.createClass
 
   selectRandomWorkflow: (project) ->
     linkedActiveWorkflows = project.links.active_workflows
-    if linkedActiveWorkflows.length is 0
-      throw new Error "No workflows for project #{project.id}"
-      project.uncacheLink 'workflows'
-    else
+    if linkedActiveWorkflows.length > 0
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks good to me, but I'll leave open for @mcbouslog in case I've missed some way to break this.

Issue #921, opened 2015-07-01
Going above the subject limit causes an error for each subject over


rogerhutchings:
![screen shot 2015-07-01 at 11 08 30](https://cloud.githubusercontent.com/assets/2725841/8452327/f419f5d2-1fe1-11e5-9f92-155f8a0f0f96.png)

Related to #316

PR #3723, opened 2017-04-13
Remove subject upload module if subject limit has been reached

simoneduca:
Removes subject upload module if subject limit has been reached.
It's a way of getting around https://github.com/zooniverse/Panoptes-Front-End/issues/921

However, the current code means users over the limit need to reload the page to get the upload module back, which is less than ideal.

So, this an interim solution while I work on the refactor of the subject upload related files. Once the flow of props is clearer and `ChangeListener` removed, I believe we can update the view without manual refresh.

If this doesn't seem very helpful/necessary, feel free to close and I'll submit a new PR with the refactor.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://disable-upload-when-limit-reached.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -199,17 +199,19 @@ EditSubjectSetPage = React.createClass
 
       <hr />
 
-      <p>
-        <UploadDropTarget accept={"text/csv, text/tab-separated-values, image/*#{if isAdmin() then ', video/*' else ''}"} multiple onSelect={@handleFileSelection}>
-          <strong>Drag-and-drop or click to upload manifests and subject images here.</strong><br />
-          Manifests must be <code>.csv</code> or <code>.tsv</code>. The first row should define metadata headers. All other rows should include at least one reference to an image filename in the same directory as the manifest.<br />
-          Headers that begin with "#" or "//" denote private fields that will not be visible to classifiers in the main classification interface or in the Talk discussion tool.<br />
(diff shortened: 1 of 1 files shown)


app/pages/lab/subject-set.cjsx
```diff
@@ -199,17 +199,19 @@ EditSubjectSetPage = React.createClass
...
-          Headers that begin with "#" or "//" denote private fields that will not be visible to classifiers in the main classification interface or in the Talk discussion tool.<br />
-          Headers that begin with "!" denote fields that <strong>will not</strong> be visible to classifiers in the main classification interface but <strong>will be </strong> visible after classification in the Talk discussion tool.<br />
-          Subject images can be up to {MAX_FILE_SIZE / 1024}KB and any of: {<span key={ext}><code>{ext}</code>{', ' if VALID_SUBJECT_EXTENSIONS[i + 1]?}</span> for ext, i in VALID_SUBJECT_EXTENSIONS}{' '}
-          and may not contain {<span key={char}><kbd>{char}</kbd>{', ' if INVALID_FILENAME_CHARS[i + 1]?}</span> for char, i in INVALID_FILENAME_CHARS}<br />
-        </UploadDropTarget>
-      </p>
-
+      {if @props.user.uploaded_subjects_count >= @props.user.subject_limit and !localStorage.adminFlag
+        <p>You've reached your subject upload limit. Please <a href='/about/contact'> contact us</a> to request changes to your allowance.'</p>
```
eatyourgreens: I think this should be using the `isAdmin()` helper rather than looking at local storage directly.

simoneduca:
@eatyourgreens done, thank you.

PR #3724, opened 2017-04-13
Default guideIcons to an empty object

eatyourgreens:
Fixes: field guide disappears from project pages if there are no attached images.

Describe your changes.
Changes the default value of `guideIcons` from `null` to `{}`

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-field-guide.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/project/index.cjsx
+++ b/app/pages/project/index.cjsx
@@ -30,7 +30,7 @@ ProjectPageController = React.createClass
     background: null
     error: null
     guide: null
-    guideIcons: null
+    guideIcons: {}
     loading: false
     loadingSelectedWorkflow: false
     owner: null


eatyourgreens:
Here's a link to the project that was having this problem: https://fix-field-guide.pfe-preview.zooniverse.org/projects/eatyourgreens/diagnosis-london/classify?env=production

The field guide doesn't show up on https://www.zooniverse.org/projects/eatyourgreens/diagnosis-london/classify

eatyourgreens:
I'm not 100% sure why this fix works, since [`guideIcons` is set to {}](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/index.cjsx#L267) after the guide is fetched, anyway.

Issue #3441, opened 2017-01-30
Add generalized feedback functionality for Project Builder projects


lcjohnso:
A number of projects (Space Warps, Gravity Spy; SW, GSpy) have started testing an experimental feature of PFE that provides volunteers feedback on their submitted annotation at the summary screen. This `metadata-based-feedback` feature has been hard-coded on a project-specific basis, but generalizing this tool would allow many interested projects/teams to use this helpful feature.  This issue expands on ideas discussed in SW-related  Issues https://github.com/zooniverse/Panoptes-Front-End/issues/2834 and https://github.com/zooniverse/Panoptes-Front-End/issues/3361, and implemented in PR https://github.com/zooniverse/Panoptes-Front-End/pull/3202. Also, similar feedback has been implemented by Planet Hunters to reveal locations of simulated and known transits.

**My interest:** I'm developing a number of Project Builder-based star cluster identification projects, similar to the completed Andromeda Project (AP).  These projects will use synthetic clusters to characterize the catalog completeness properties (and receiver operating characteristic curve - ROC).  We found it important to clarify to the volunteers the purpose of these synthetic objects, so they didn't feel that they were being tested or studied.  In addition, volunteers commented that they liked receiving positive feedback when they successfully identified synthetic clusters, reassuring them that they were accomplishing the task successfully.  Here, feedback serves a dual purpose: disclosure and training.

**Specifications for Feedback:**
For this use case, it makes sense to include information about the synthetic clusters in the subject metadata, using similar manifest columns as currently implemented for SW:

#Type #X #Y #Tol
SIM 100.5 200.5 50.0

When an image with a synthetic cluster is shown, the following would be the desired behavior for the feedback:

If volunteer marks a cluster (not galaxy or image artifact marker types) within #Tol pixels of synthetic's (#X, #Y) input position:
1) Display circular marker at (#X, #Y) with radius #Tol.
2) Display text either in text bubble next to synthetic object marker, or as part of the summary screen text on the right side of the UI. Here, the `subjectSuccessMessage` could be the same text every time, defined in `workflow_contents`: "Congratulations, you found a synthetic cluster!"

Optional Features:
- The `subjectSuccessMessage` text could rotate through a small number of randomly chosen messages to mix things up, or in the future could provide tailored feedback based on the specifics of the individual synthetic.  This latter enhancement could be implemented using a #SuccessMessageIndex column that refers back to an array of responses that could be stored in `workflow_contents`.

[...]

PR #3725, opened 2017-04-13
Generalised feedback

rogerhutchings:
Adds in generalised feedback for the single question type to the project builder / classifier. It's been written in a fairly modular way, so adding in new question types (I'll be doing drawing points next) should be straightforward. Unsupported question types show a message to that effect.

__Demo URL__: https://general-feedback.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test

Related issues: #3441, #3361

## How to use

Feedback is enabled per-task - there's a checkbox under the answer editing box that toggles it on and off.

screen shot 2017-04-13 at 14 29 36

Once enabled, you can add a feedback type by hitting the button:

screen shot 2017-04-13 at 14 30 40

You can then add information for that feedback type. All fields are required.

- `ID` - a unique field to identify the type of feedback, e.g. SIM, DUD etc.
- `Default success message` - shown if the correct answer is selected
- `Default success message` - shown if an incorrect answer is selected
- `Answer` - links the correct answer to that feedback type

In the subject manifest, you can then add the following columns:

- `#feedback_N_type` - corresponds to the ID field above
- `#feedback_N_successMessage` - overrides the default success message
- `#feedback_N_failureMessage` - overrides the default failure message

`N` is an integer - you can have as many feedback fields as you like (currently they'll all be checked when running feedback). Only `#feedback_N_type` is required.

For the "single question required" workflow, the manifest looks like this:

```csv
image_name_1,#feedback_1_type,#feedback_1_successMessage,#feedback_1_failureMessage,
foo.jpg,"FOO","","",
bar.jpg,"BAR","","",
baz.jpg,"BAZ","this is a custom success message","this is a custom failure message",
```

## Feedback summary

The summary is generated at the end of a classification, based on the subject feedback fields, the active rules, and the task:

screen shot 2017-04-13 at 14 41 11

## To do:

- [x] Tweak the UI to look a bit more finished
- [x] Add a flag to the classification to say whether the user was shown feedback or not
- [x] Work out why the default classification summary isn't being shown
- [x] Add proptypes
- [x] Write tests
- [ ] Write documentation

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://general-feedback.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test

## Optional


trouille:
Really good stuff. @rogerhutchings Could you also include a file showing what the output looks like in the 'Data Export' raw classification .csv file? 

Also, could you clarify what you mean by "N is an integer - you can have as many feedback fields as you like (currently they'll all be checked when running feedback)."? 

Do you mean you could have a metadata file that looks like:

image_name_1,#feedback_1_type,#feedback_1_successMessage,#feedback_1_failureMessage,#feedback_2_type,#feedback_2_successMessage,#feedback_2_failureMessage, etc.?

Is this for the scenario of being able to choose multiple responses and getting feedback in that scenario? Likely you have a use case for this in mind. Could you share that? Thanks!

rogerhutchings:
@trouille 

1. There's no change to the classification data export at the moment. I think it includes subject metadata fields, however, so backend processing can be done there (although it requires knowledge of the workflow to match up answer indices and annotation values).

1. This is for future expansion - there may be multiple known sims in a point-based drawing task, for example, or multiple animals in a survey task. You could also have different sim data for different workflow steps; I'm planning to optionally restrict which fields are checked per task to prevent collisions with a lot of feedback items. E.g. task one only checks `#feedback_1_type`, task two only checks `#feedback_2_type` and so on

eatyourgreens:
If a subject defines custom feedback messages, in English, in the subject metadata, I think that will create problems with future translation of a project into other languages.

rogerhutchings:
__pain__

Hmm. We could add language codes to the end of the message fields, so `#feedback_1_successMessage` can also have `#feedback_1_successMessage_cn`, which will become the preferred message. It would make adding new translations etc a bit of an arse, though.

rogerhutchings:
I've refactored a bunch of stuff, and added point-based feedback: https://general-feedback.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test

lcjohnso:
@rogerhutchings Looking good!  The drawing feedback seems to be working sensibly, but placement of feedback message for small Y SIM locations should be tested.

For drawing task feedback, will the extension to circle and ellipse points be much trouble?  Point, circle, and ellipse markers are all compatible with the `isWithinTolerance` function, but not sure whether you'll need to iterate through each of these marker types individually.  I'm happy with point-like representation of the success/failure marking (via `feedback-point.jsx`) for all three marker types with (x, y, r) set by (x, y, tol).

Re: custom feedback messages and translation troubles -- my use case does not require custom messages, and I'm happy with type-defined (via `feedback_N_type`) messages. It seems that one could implement customization of feedback messages via multiple feedback types.  For example: using SIM1, SIM2, etc. to pass multiple messages instead of needing subject-specific custom feedback messages.

Two requests that would jive with the specifics of my particular project:
1) Is it possible to disable failure messages?  I do NOT want to display feedback if SIM is not found -- can this be the default behavior if the "Default Failure Message" is ""?
2) Only the on-subject message is required -- I'm happy for the default classification summary to display upon submission, but no strong feelings one way or the other.

I'm happy to provide my set of test subjects for additional testing -- just let me know what the new metadata column name syntax is for the manifest: e.g., `#feedback_1_X`, `#feedback_1_Y`, `#feedback_1_TOL`?

rogerhutchings:
@lcjohnso For 1, not currently, although I can add that in. 2 has been fixed since.

rogerhutchings:
Okay, this is ready to go. Apologies again for the size of the PR.

Things that may be of interest:

- [Redux!](https://github.com/zooniverse/Panoptes-Front-End/pull/3725/files#diff-58ed19aaca2d7ef6489531b4b3594890R1)
- [Ducks!](https://github.com/zooniverse/Panoptes-Front-End/pull/3725/files#diff-203372400dbefd07144c79c0b95b7fb1R1)
- [Using Rewire to mock components](https://github.com/zooniverse/Panoptes-Front-End/pull/3725/files#diff-8f0f2a362d2e0a90fdc1af93356b10dfR16)
- [Exporting redux-aware components as a named export for testing](https://github.com/zooniverse/Panoptes-Front-End/pull/3725/files#diff-4f50e696bbeb5a1d05bae0ef30dbf9edR44)

srallen:
I'll take a look at this within the context of how Gravity Spy uses feedback. Thanks.

app/classifier/classification-summary.jsx
```diff
@@ -91,12 +115,13 @@ class ClassificationSummary extends React.Component {
 
     if (this.showExoplanetSimFeedback()) {
       return (
-        <p style={{ fontWeight: 'bold' }}>Well done! You've found a simulated planet. We include these to help calibrate the project. Keep going to discover a real planet!</p>
+        <p style={{ fontWeight: 'bold' }}>Well done! Youve found a simulated planet. We include these to help calibrate the project. Keep going to discover a real planet!</p>
```
srallen: Was this change intentional?

app/classifier/feedback/feedback-point.spec.js
```diff
@@ -0,0 +1,45 @@
...
+  x: '10',
+  y: '20',
+  tol: '30',
+  success: true
+};
+
+describe('<FeedbackPoint />', function() {
+  it('should return a circle element', function () {
+    const wrapper = shallow(<FeedbackPoint point={POINT} />);
+    assert(wrapper.type() === 'circle');
```
srallen: assert has a strict equal method: `assert.strictEqual(actual, expected[, message])`

app/classifier/feedback/feedback-ruleset.js
```diff
@@ -0,0 +1,61 @@
...
+  _getActiveTypes() {
+    // Find the feedback types required in the subject metadata and check for valid active rules in the task definition.
+    const { metadata } = this._subject;
+    const { feedback } = this._task;
+    return Object.keys(metadata).reduce((types, key) => {
+      const match = key.match(/#feedback_(\d*?)_type/);
+      if (match) {
+        const index = match[1];
+        const type = metadata[key];
+        const typeInTaskDefinition = feedback.types.find((feedbackItem) => feedbackItem.id === type);
```
srallen: `Array.find` isn't supported in IE, so this will need a polyfill. Polyfill.io can polyfill it, just needs to be added to the param list in the source url.

app/classifier/feedback/feedback-summary-container.jsx
```diff
@@ -0,0 +1,61 @@
...
+};
+
+class FeedbackSummaryContainer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.generateFeedbackItems = this.generateFeedbackItems.bind(this);
+  }
+
+  componentWillMount() {
+    this.generateFeedbackItems();
```
srallen: I think you can just call this in the class constructor and the React docs state doing this is preferred: https://facebook.github.io/react/docs/react-component.html#componentwillmount

FWIW, I'm not sure when `componentWillMount` would be preferred over the constructor, though this discussion suggests a use case is if a Redux dispatch needs to happen before mount: http://stackoverflow.com/questions/40828004/constructor-vs-componentwillmount-what-a-componentwillmount-can-do-that-a-const

app/classifier/classification-summary.jsx
rogerhutchings: nope ><

app/classifier/feedback/feedback-summary-container.jsx
rogerhutchings: 👍

app/classifier/feedback/feedback-ruleset.js
rogerhutchings: 👍

rogerhutchings:
I left a cheeky `it.only()` in the tests ~~because I'm an idiot~~ to keep you on your toes.

simoneduca:
No feedback shown for drawing points, @rogerhutchings?
screen shot 2017-05-15 at 10 36 31

simoneduca:
Also, when I select the single question, skip summaries workflow, I don't see feedback for foo, neither when answer is correct nor incorrect.

rogerhutchings:
I think this is due to recent change not passing down the project as a prop to the FrameAnnotator.

simoneduca:
Still can't see feedback for foo on skip summaries wf

app/classifier/classification-summary.jsx
```diff
@@ -44,8 +47,29 @@ class ClassificationSummary extends React.Component {
     );
   }
 
+  isFeedbackEnabled() {
+    if (this.props.project && this.props.project.experimental_tools.includes('general feedback')) {
```
simoneduca: This line is quite difficult to read. Deconstruct to improve that: perhaps `const { project } = this.props` would help.

app/classifier/classifier.jsx
```diff
@@ -148,6 +149,16 @@ export default class Classifier extends React.Component {
   }
 
   completeClassification() {
+    if (this.props.project.experimental_tools && this.props.project.experimental_tools.includes('general feedback')) {
```
simoneduca: This line is very hard to read. Deconstruct would help.

eatyourgreens:
There's a bunch of broken tests and a merge conflict. Bad rebase somewhere along the line?

rogerhutchings:
Yeah, I'm fixing it up now.

rogerhutchings:
I've pushed some fixes. The test breakages were caused by the FrameAnnotator tests not mocking its dependency components, so one of them tried to use the SVG DOM element (which doesn't exist when shallowly rendered). To fix it, I've changed the `screenCTM` object to a function prop which gets called further down the chain.

eatyourgreens:
I've had problems with the SVG tasks breaking when I tried passing the function as a prop, so that's something to beware of.

rogerhutchings:
Good to know. I've tested it locally and it seems legit - it's just used to position the tooltips.

app/classifier/frame-annotator.jsx
```diff
@@ -114,8 +102,6 @@ export default class FrameAnnotator extends React.Component {
           {!!AfterSubject && (
             <AfterSubject {...hookProps} />)}
 
-          <TooltipLayer screenCTM={screenCTM} />
```
eatyourgreens: Did you mean to take out the tooltip layer here?

eatyourgreens: Oh wait, it's moved to the SVG component, right?

rogerhutchings: Yes, I've moved it to the SVG renderer - it's only designed to work with image subjects at the moment.

rogerhutchings: SNAP

eatyourgreens:
Cool. Yeah the problem I'd had was also passing the ScreenCTM directly, rather than a reference to `getScreenCTM`.

rogerhutchings:
@simoneduca @srallen Your comments should be all addressed.

app/pages/admin/project-status/experimental-features.jsx
```diff
@@ -24,6 +24,7 @@ const experimentalFeatures = [
   'freehandSegmentShape',
   'enable subject flags',
   'sim notification',
+  'general feedback'
```
simoneduca: Missing comma

app/pages/lab/feedback/feedback-editor.jsx
```diff
@@ -0,0 +1,103 @@
...
+    this.isFeedbackAvailable = this.isFeedbackAvailable.bind(this)
+    this.renderFeedbackEnable = this.renderFeedbackEnable.bind(this)
+    this.renderFeedbackForm = this.renderFeedbackForm.bind(this)
+  }
+
+  render() {
+    // Unfortunately, the JSON client mutates an object on update, rather than
+    // returning a new instance. So prevProps will always be identical to current
+    // props, breaking some of the lifecycle methods. This ensures we always have
+    // a new object.
```
simoneduca: Is it worth opening an issue for JSON client regarding this?

rogerhutchings: I don't think it can be helped. It'll be less of a problem once state starts getting moved into Redux, since the reducer can be configured to always return a new object (or more likely, since best practice requires Redux objects to be serialisable, the static data is kept in state and components create a new Resource instance based on that).

eatyourgreens: I don't think `.update` should create a new instance of a resource, and there's probably a bunch of code in PFE which relies on that behaviour.

simoneduca: Makes sense.

srallen: `.update` will mutate the cached resource and `.save` is what actually makes the API call. What I've been doing in pfe-lab is dispatching a redux action to update the local store with the resource from the POST response. I don't think we'll be able to get around this without a major refactor on the json-api client.

simoneduca:
@srallen re: the skip summaries wf, I noticed that too and Rog pointed to [another PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3769) that addresses that behaviour.

srallen:
@simoneduca thanks. I missed that comment.

simoneduca:
@srallen I clicked on the first staged version and I was able to reproduce the error you mention, but that version has been superseded. In fact, after clearing the cache, the error disappeared.

This LGTM.

rogerhutchings:
I've fixed the conflicts, will hit merge once Travis completes. Thanks for the help, everyone!

Issue #3458, opened 2017-02-08
highlighter tool for text subjects


simensta:
## Expected behavior
A highlighter tool for selecting metadata in text subjects.
<img width="921" alt="highlighterExample" src="https://cloud.githubusercontent.com/assets/7684729/22743697/9540f546-ede0-11e6-81db-d3537ae9074c.png">

A volunteer should be able to select a highlighter label. Then, select a portion of text. After the text is selected, the background color for that selection should match the color of the label.

The annotation can be represented by an anchorIndex and a focusIndex.

```
annotation: {
     _toolIndex: 0,
    task: "T1",
    value: {
       0: {
           anchorIndex: 41,
           focusIndex: 49,
           labelInformation: {
                color: "#00f00",
                label: "Sender"
        }
    }
}
```

eatyourgreens:
What do `anchorIndex` and `focusIndex` refer to?

simensta:
The `anchorIndex` is the character index for the beginning of the selection.`focusIndex` is the character index for the end of the selection. I chose the labels based on the [Selection API](https://developer.mozilla.org/en-US/docs/Web/API/Selection). However, it might be clearer to use `start` and `end` as labels instead.

eatyourgreens:
`start` and `end` sound more sensible to me. I suspect researchers using this tool will be confused by the terminology used by the Selection API.

eatyourgreens:
>A volunteer should be able to select a highlighter label. Then, select a portion of text. After the text is selected, the background color for that selection should match the color of the label.

I've found you can't reliably detect when a selection is complete, particularly when selecting text by tapping on iOS or Android, or selecting text from the keyboard with a screenreader. So, you might need an additional action (probably pressing a button) to trigger `document.getSelection()`.

eatyourgreens:
Noting that the UI should allow for deleting an annotation too.

simensta:
Closed by #3738.

PR #3738, opened 2017-04-18
Text Highlighter Task

simensta:
Work towards #3458.
Requesting feedback on this approach for a highlighting tool. [Example of the highlighting tool](https://highlighter-tool-v2.pfe-preview.zooniverse.org/projects/scribe-2/textsubjectproject/classify?reload=0&workflow=2831).

On the project above, a user highlights some text and then clicks that text selection to record the annotation. That is not a task flow we want in the end product. @eatyourgreens [suggested a task flow of 1) highlighting text, 2) selecting label for the highlighted text (like Diagnosis London)](https://github.com/zooniverse/Panoptes-Front-End/issues/3458). I am working on implementing this structure for the task. 

Describe your changes.
- Adds a highlighter task to the workflow editor. The highlighter task has a similar structure to the drawing task.
- Starts adding a highlighter tool to the /classify page. I tried to replicate some of the same structure of the drawing tools. For example, the highlighter's label-renderer is suppose to serve a similar function as the markings-renderer.

Todo List
- [x] Task flow of 1. select text, 2. select label for text
- [x] Support highlighting text both forward and backward. At the moment, a user can only highlight forward.
- [x] Use more accessible highlighting colors
- [x] Support selecting text by tapping on a touchscreen
- [x] Support selecting text by keyboard shortcut
- [x]  In the annotation, change `anchorIndex` and `focusIndex` to `start` and `end`, respectively
- [x] Delete selection annotations
- [x] Fix tests
- [x] The classification summary should display the selected words instead of the character indices
- [x] If the start index and end index of a selection are the same, the tool should not create an annotation.
- [x] Annotation values should only be created for selections of subject content, not selections from anywhere on the page.
- [x] Text can be selected from inside a highlighted phrase, causing a duplication of the selected text. 
- [ ] Labeling more than one selection at a time?
- [x] Access to this tool should be enabled under the admin flag (#3873)

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Did you deploy a staging branch? https://highlighter-tool-v2.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/frame-annotator.jsx
+++ b/app/classifier/frame-annotator.jsx
@@ -91,9 +91,9 @@ export default class FrameAnnotator extends React.Component {
           {!!BeforeSubject && (
             <BeforeSubject {...hookProps} />)}
 
-          <AnnotationRenderer type={type} {...rendererProps} />
-
-          {this.props.children}
+          <AnnotationRenderer type={type} {...rendererProps}>
+            {this.props.children}
+          </AnnotationRenderer>
 
           {!!warningBanner && (
             warningBanner
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -44,9 +44,10 @@ export default class Highlighter extends React.Component {
   }
 
   selectableArea(selection, range, offset, start, end){
+    const spansNodes = range.startContainer !== range.endContainer
     const noAreaSelected = start === end;
     const subjectSelection = selection.anchorNode.parentElement.parentElement.className === 'label-renderer';
-    const isSelectable = !noAreaSelected && subjectSelection;
+    const isSelectable = !spansNodes && !noAreaSelected && subjectSelection;
     return isSelectable;
   }
 
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -31,7 +31,8 @@ export default class Highlighter extends React.Component {
     newAnnotation.value.push({
       labelInformation: labelInformation,
       start: start,
-      end: end
+      end: end,
+      text: range.toString()
     });
     this.props.onChange(newAnnotation);
     selection.collapseToEnd();
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -53,16 +53,15 @@ export default class Highlighter extends React.Component {
     const { anchorNode } = selection;
     const { parentNode } = anchorNode;
     let start = 0;
-    let counting = true;
-    parentNode.childNodes.forEach((node) => {
+    for (const node of parentNode.childNodes) {
       // ignore comments (nodeType 8) when calculating offset distance
-      if (counting && node.nodeType !== 8 && node !== anchorNode) {
+      if (node.nodeType !== 8 && node !== anchorNode) {
         start += node.textContent.length;
       }
       if (node === anchorNode) {
-        counting = false;
+        break;
       }
-    });
+    }
     return start;
   }
 
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -3,6 +3,7 @@ import { Markdown } from 'markdownz';
 import GenericTaskEditor from '../generic-editor';
 import GenericTask from '../generic';
 import LabelRenderer from './label-renderer';
+import HighlighterSummary from './summary';
 
 export default class Highlighter extends React.Component {
   constructor(props) {
@@ -66,6 +67,8 @@ export default class Highlighter extends React.Component {
 Highlighter.Editor = GenericTaskEditor;
 
 Highlighter.InsideSubject = LabelRenderer;
+
+Highlighter.Summary = HighlighterSummary;
 
 Highlighter.getDefaultTask = () => {
   return (
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -12,6 +12,7 @@ export default class Highlighter extends React.Component {
     this.handleChange = this.handleChange.bind(this);
     this.createButtons = this.createButtons.bind(this);
     this.createLabelAnnotation = this.createLabelAnnotation.bind(this);
+    this.selectableArea = this.selectableArea.bind(this);
   }
 
   handleChange(toolIndex, e) {
@@ -27,15 +28,28 @@ export default class Highlighter extends React.Component {
     const end = offset + range.endOffset;
     const task = this.props.workflow.tasks[this.props.annotation.task];
     const labelInformation = task.highlighterLabels[toolIndex];
-    const newAnnotation = Object.assign({}, this.props.annotation, { _toolIndex: toolIndex });
-    newAnnotation.value.push({
-      labelInformation: labelInformation,
-      start: start,
-      end: end,
-      text: range.toString()
-    });
-    this.props.onChange(newAnnotation);
+    const selectable = this.selectableArea(selection, range, offset, start, end);
+    if (selectable) {
+      const newAnnotation = Object.assign({}, this.props.annotation, { _toolIndex: toolIndex });
+      newAnnotation.value.push({
+        labelInformation: labelInformation,
+        start: start,
+        end: end,
+        text: range.toString()
+      });
+      this.props.onChange(newAnnotation);
+    }
     selection.collapseToEnd();
+  }
+
+  selectableArea(selection, range, offset, start, end){
+    const noAreaSelected = start === end;
+    const subjectSelection = selection.anchorNode.parentElement.parentElement.className === 'label-renderer';
+    if (!noAreaSelected && subjectSelection) {
+      return true;
+    } else {
+      return false;
+    }
   }
 
   getOffset(selection) {
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import { Markdown } from 'markdownz';
 import GenericTaskEditor from '../generic-editor';
 import GenericTask from '../generic';
+import LabelRenderer from './label-renderer';
 
 export default class Highlighter extends React.Component {
   constructor(props) {
@@ -11,6 +12,7 @@ export default class Highlighter extends React.Component {
   }
 
   handleChange(toolIndex, e) {
+    debugger;
     if (e.target.checked) {
       const newAnnotation = Object.assign({}, this.props.annotation, { _toolIndex: toolIndex });
       this.props.onChange(newAnnotation);
@@ -52,6 +54,8 @@ export default class Highlighter extends React.Component {
 }
 
 Highlighter.Editor = GenericTaskEditor;
+
+Highlighter.InsideSubject = LabelRenderer;
 
 Highlighter.getDefaultTask = () => {
   return (
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -45,11 +45,8 @@ export default class Highlighter extends React.Component {
   selectableArea(selection, range, offset, start, end){
     const noAreaSelected = start === end;
     const subjectSelection = selection.anchorNode.parentElement.parentElement.className === 'label-renderer';
-    if (!noAreaSelected && subjectSelection) {
-      return true;
-    } else {
-      return false;
-    }
+    const isSelectable = !noAreaSelected && subjectSelection;
+    return isSelectable;
   }
 
   getOffset(selection) {
--- a/app/classifier/tasks/highlighter/index.jsx
+++ b/app/classifier/tasks/highlighter/index.jsx
@@ -101,12 +101,23 @@ export default class Highlighter extends React.Component {
 Highlighter.Editor = GenericTaskEditor;
 
 Highlighter.InsideSubject = (props) => {
-  const onClick = (e) => {
-    console.log(e.target)
+  function onClick(e) {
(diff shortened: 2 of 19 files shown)


eatyourgreens:
Noting here that we should probably also support selecting text by tapping on a touchscreen, or [selecting words using keyboard shortcuts](http://osxdaily.com/2012/02/23/keyboard-shortcuts-to-navigate-select-text-mac-os-x/).

app/classifier/frame-annotator.jsx
```diff
@@ -186,38 +191,50 @@ export default class FrameAnnotator extends React.Component {
...
+                return <PersistInsideSubject key={taskName} {...hookProps} />;
+              }
+            })}
+          </g>
+        </svg>
+    }
+
+    if (type === 'text' && TaskComponent) {
+      const currentFrame = this.props.subject.locations[this.props.frame];
+      textAnnotator = <InsideSubject src={currentFrame['text/plain']} {...hookProps} />
```
CKrawczyk: This should be something like 
```js
textAnnotator = InsideSubject ? <InsideSubject src={currentFrame['text/plain']} {...hookProps} /> : undefined;
```
since `InsideSubject` might not exist

eatyourgreens:
I've made a start on breaking up the frame annotator in #3743.

eatyourgreens:
A couple of other things that occurred to me:

1. The annotations will vanish for subsequent tasks in the workflow, unless the tool uses the `PersistInsideSubject` hook.
2. There's an implicit assumption in PFE at the moment that those `PersistInsideSubject` hooks generate SVG, not HTML. If a workflow contains a drawing task and a highlighting task, then rendering will break because of the mixture of HTML and SVG. I can look at this in #3743, where I'm trying to separate SVG and HTML rendering for annotations.

eatyourgreens:
I've rebased to bring in changes to how annotations are rendered from #3743 and added a renderer for this task.

There might be some empty commits in the history now, so that might need tidying up.

eatyourgreens:
@simentsta could you add another task (maybe a question) to the test project so that we can check whether markings persist properly?

app/classifier/tasks/highlighter/index.jsx
```diff
@@ -0,0 +1,125 @@
...
+    const start = range.startOffset;
+    const end = range.endOffset;
+    const task = this.props.workflow.tasks[this.props.annotation.task];
+    const labelInformation = task.highlighterLabels[toolIndex];
+    const newAnnotation = Object.assign({}, this.props.annotation, { _toolIndex: toolIndex });
+    newAnnotation.value.push({
+      labelInformation: labelInformation,
+      start: start,
+      end: end
+    });
```
eatyourgreens: The annotation should probably include the selected text as well as the start and end points.

app/classifier/tasks/highlighter/label-renderer.jsx
```diff
@@ -0,0 +1,119 @@
...
+      }
+      return(newContent);
+    } else {
+      return([this.state.content]);
+    }
+  }
+
+  render() {
+    const labeledContent = this.createNewContent();
+    const invisibleDivStyles = {
```
eatyourgreens: Use CSS classes instead of inline styles here.

app/classifier/tasks/highlighter/summary.jsx
```diff
@@ -0,0 +1,25 @@
...
+
+export default class HighlighterSummary extends React.Component {
+  constructor(props){
+    super(props);
+    this.createSummary = this.createSummary.bind(this);
+  }
+  createSummary(value, index){
+    return(
+      <div key={index} className="answer">
+        <p>{value.labelInformation.label} : {value.start} to {value.end}</p>
```
eatyourgreens: It would be good to display the selected text for each selection here.

app/classifier/tasks/highlighter/label-renderer.jsx
```diff
@@ -0,0 +1,119 @@
...
+    if (newProps.src !== this.props.src) {
+      this.loadText(newProps.src);
+    }
+  }
+
+  componentDidMount(){
+    this.loadText(this.props.src);
+  }
+
+  loadText(src) {
```
eatyourgreens: This is duplicated from the `TextViewer` component. It would be good to get rid of it and rely on the subject viewer to load our subject for us.

eatyourgreens:
A couple of bugs from playing around with the text project:

1. Text can be selected from anywhere on the page. Select 'Highlight and click' from the question text and click a button. The first 19 characters of the telegram are highlighted.

2. Text can be selected from inside a highlighted phrase. Select characters from inside a highlighted string and click a button. Those characters are duplicated and added to the rendered text twice.

app/classifier/tasks/highlighter/summary.jsx
simensta: Definitely!

app/classifier/tasks/highlighter/label-renderer.jsx
eatyourgreens: 127f562 should fix this by allowing `TextViewer` to update `LabelRenderer` on load.

eatyourgreens:
We might be able to restrict text selection to the subject only by [checking the selection's parent node](/zooniverse/diagnosislondon/blob/master/app/lib/annotation-tool.coffee#L61). That will only work if the subject is plain text. If we add future support for parsed markdown, then the parent won't necessarily be the subject viewer.

eatyourgreens:
@simensta rebased to bring in a bug fix from #3751

eatyourgreens:
If I'm selecting text using VoiceOver to read out the subject, both div elements are read, so the subject is read out twice and it's not obvious that one is for selecting text and the other is for displaying the annotated text.

Maybe we should just handle selecting text selection and displaying selections on a single div, like Diagnosis London. [The code for figuring out the offsets etc.](/zooniverse/diagnosislondon/blob/master/app/lib/selection-tool.coffee#L41-L51) might be adapted for use here.

eatyourgreens:
The latest commit hides the text overlay so that all selection is done on the label renderer div. It's a pretty big change, so needs some checking to make sure I haven't broken anything.

With this change, we could make the highlighted text clickable, so that we could add delete buttons to remove annotations, for example. Or maybe popup forms to add details to a selection, like with the drawing tools.

app/classifier/tasks/highlighter/index.jsx
```diff
@@ -36,6 +37,23 @@ export default class Highlighter extends React.Component {
...
+    const { parentNode } = anchorNode;
+    let start = 0;
+    let counting = true;
+    parentNode.childNodes.forEach((node) => {
+      // ignore comments (nodeType 8) when calculating offset distance
+      if (counting && node.nodeType !== 8 && node !== anchorNode) {
+        start += node.textContent.length;
+      }
+      if (node === anchorNode) {
+        counting = false;
```
eatyourgreens: This is a bit of a hack to get around `NodeList` not being an array, so we can't use `break` to get out of the loop once we find the node we're working on.

eatyourgreens:
I'm trying out some new colours, chosen using a bit of colour theory rather than just barfing random CSS colours at the screen (like the drawing tools do.)
Screenshot showing four possibilities for new highlight tool colours

The colours were generated using the picker at http://paletton.com/, generating complementary colours for the PFE button colours.

simensta:
The hidden-text-overlay/counting-node-offset approach from Diagnosis London works nicely! The http://paletton.com tool is cool, the new selection colors are a big improvement. 

Another selection edge case for us to address: if the start index and end index of a selection are the same, the tool should not create an annotation.

simensta:
Noting that this tool should probably be released under the admin flag.

eatyourgreens:
I think we're going to need an `InsideSubject` hook that allows for editing/deleting the marks, so that editing can only happen during this task and not during other steps.

eatyourgreens:
I've made a start on an `InsideSubject` hook that can listen for events on the highlighted text. Haven't quite figured out yet how that translates to actually deleting an annotation from the annotations array.

I've been looking at the drawing tools, to see how they work, but haven't had much luck figuring them out, I'm afraid.

CKrawczyk:
I am not sure how helpful this is, but drawing tasks handle removal here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/marking-initializer.cjsx#L110-L115

eatyourgreens:
@CKrawczyk yeah, that's the code I've been looking at. Do you know how it gets invoked from clicking on a delete button?

eatyourgreens:
Figured something out. It's mouse only at the moment, so still needs work, but you can focus a selection by clicking it, then delete the annotation with backspace.

The trick seems to be having the event listeners in the `InsideSubject` hook, so that they are only invoked when the task is current.

CKrawczyk:
It looks like the "x" button uses `delete-button` here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/drawing-tools/delete-button.cjsx#L37-L39

And that calls the code linked above: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/markings-renderer.cjsx#L81

app/classifier/tasks/highlighter/index.jsx
simensta: Sadly, `NodeList.forEach` isn't supported in Edge or IE.
https://developer.mozilla.org/en-US/docs/Web/API/NodeList

srallen: Maybe use `Array.from` to convert it to an array? You'll need a polyfill for it. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from?v=control

eatyourgreens: The [polyfill that we're using](https://polyfill.io/v2/docs/features/) might have NodeList iterators. I'll double-check.

eatyourgreens: In the end, I've just removed `forEach` and [replaced it with a `for… of` loop](https://developer.mozilla.org/en-US/docs/Web/API/NodeList#Example).

eatyourgreens:
Updated to use the delete button from survey tasks as a delete button for text selections.

@CKrawczyk I decided copying the drawing tools was a bad idea and went for an approach where the rendered annotations, in `PersistInsideSubject`, fire events, then the `InsideSubject` component adds listeners so that those events are only handled during the highlighter task.

I think this will still break if there is more than one highlighter task in a workflow.

app/classifier/tasks/highlighter/index.jsx
```diff
@@ -0,0 +1,159 @@
...
+    const subjectSelection = selection.anchorNode.parentElement.parentElement.className === 'label-renderer';
+    const isSelectable = !spansNodes && !noAreaSelected && subjectSelection;
+    return isSelectable;
+  }
+
+  getOffset(selection) {
+    const { anchorNode } = selection;
+    const { parentNode } = anchorNode;
+    let start = 0;
+    for (const node of parentNode.childNodes) {
```
eatyourgreens: `for…of` will probably break in IE11.

simensta: I can confirm that `for...of` breaks in IE11 (tested on Windows 10, IE11).

simensta: I'll work on replacing `for...of` with something IE11 compatible.

eatyourgreens: This fixed it for the account menu. https://github.com/zooniverse/Panoptes-Front-End/pull/3826/commits/aff7d27e363dc63b9066297658bf27bbc77b9f7d

eatyourgreens:
Rebased (and staged) to fix merge conflicts.

app/classifier/tasks/highlighter/index.jsx
simensta: I addressed this and another IE issue in #3873. I tried to replace `for...of` with `forEach`, but I couldn't find an equivalent to the `break` behavior using `forEach`. So, I replaced `forEach` with a simple `for` loop.

simensta:
#3873 resolved most of the IE11 issues. However, in IE11 there is still an issue with highlighting the end of a line. See @shaunanoordin's review of #3873.

simensta:
Rebased and staged to bring in updates.

eatyourgreens:
Rebased and staged to fix conflicts with the audio player update.

app/classifier/tasks/highlighter/index.jsx
```diff
@@ -0,0 +1,160 @@
...
+    let selection = document.getSelection();
+    this.createLabelAnnotation(selection, toolIndex);
+  }
+
+  createLabelAnnotation(selection, toolIndex) {
+    // currently we only deal with one selection at a time
+    const range = selection.getRangeAt(0);
+    const offset = this.getOffset(selection);
+    const start = offset + range.startOffset;
+    const end = offset + range.endOffset;
```
simensta: In both Edge and IE, `range.endOffset` returns an inconsistent number when highlighting the last word of a line. This happens whether the newline character is `\n` or `\n\r`. However, if the new line is indicated by a `<br>` tag, I can produce consistent values for `range.endOffset`.

eatyourgreens:
Looks like I missed something in that rebase and broke the FrameViewer.

app/classifier/tasks/highlighter/index.jsx
simensta: I think I have a solution for this issue that substitutes newline characters with break tags in the LabelRenderer.

eatyourgreens:
@simensta Any objections if I squash and merge this, and we deal with IE in a separate issue/PR?

simensta:
Sounds good to me!

eatyourgreens:
@simensta 👍 I've just been sat down for a couple of hours with a couple of medical historians, building some workflows that use this tool, so it will be good to get an initial version of the tool into master.

PR #3740, opened 2017-04-19
Collection default subject preview feature

### LINKED ISSUE #3593: Set default subject for collection
## Expected behavior
A user should be able to set a default for their collection which is the subject that is then shown on the collections page.

## Current behavior
Collections page just shows the first subject in the linked subjects array.

srallen:
Closes #3593 

This adds the new feature for setting a custom subject preview for a collection. WIP because I need to add tests for the components I converted to ES6 along the way. Open PR so @beckyrother can begin reviewing the styles.

Staging link: https://collection-preview-es6.pfe-preview.zooniverse.org/

**Do not merge until zooniverse/Panoptes#2296 is in production.**

# Review Checklist

- [ ] Does it work on mobile?

## Optional


beckyrother:
The buttons look great! The images in the collection look like they're stretched, though – can they be cropped rather than resized? 

ghosts_ _another_ _zooniverse

trouille:
A good thing to look into is what happens if you delete the subject from your collection that was chosen as your 'collection cover'. It looks like it remains as your collection cover, but it might be better if the collection cover were to switch back to the first subject in the collection instead.

srallen:
The image stretching only happens in Chrome. I'm looking into it.

srallen:
The img has a `width: 100%; max-height: 250px;` set. For images that have really different aspect ratios than typical, like those images on staging, then those CSS style properties are being interpreted by Chrome to stretch the image to fit those constraints. Removing `width: 100%` is not feasible because images can overflow the parent if the original source is larger. The `max-height` could be removed, but with images that are really tall, the subject viewer will not be uniform in size as it will dynamically be the height of the source image. The other options would be to set the image as a background and use `background-size: cover` or use the `object-fit` property, but `object-fit` is not supported by Edge. I'm inclined to leave as is because in production, I really doubt there are subjects with these kinds of bizarre aspect ratios. It's something to consider when we get around to writing a simplified subject viewer for use on Talk and in collections.

beckyrother:
True, there aren't a ton of projects with weird ratios like this. If the plan is to make a new subject viewer, I guess it doesn't make sense to spend a ton of time making this one look perfect. Just checked it out in Firefox and Safari and it looks fine there, so I'll approve it.

srallen:
@eatyourgreens feel free to reassign if you're busy. Mark's out on vacation for the next couple of weeks.

eatyourgreens:
I've made a note to get to this this week.

For subjects with very tall images, I'd suggest looking at transcription projects like Decoding the Civil War.

srallen:
Also noting that zooniverse/Panoptes#2306 needs to be deployed before this gets merged into production. Once it's available on Panoptes staging, there's a few lines in this PR I'll need to change, too.

srallen:
zooniverse/Panoptes#2306 is now merged into staging, so I'm going to follow up on the changes needed on this branch for it. This PR still would need to wait for both Panoptes PRs to be deployed to production.

srallen:
Ok, this branch has the updates needed and is ready for review. Still waiting on the production Panoptes deploy for merging, though.

I'm working on some basic unit tests for the subject page ES6 conversion in a separate branch, but I'm running into an issue with `coffee-react` throwing when running tests: `Error: Parse Error: Line 6: Illegal import declaration`. The pages all seem to work by manually testing. I'm not sure if you want to wait for automated tests or not. Let me know.

srallen:
Panoptes production deployed today, so WIP label is removed.

https://collection-preview-es6.pfe-preview.zooniverse.org/?env=production

eatyourgreens:
Re. the test error, sounds like a coffeescript file is using `import…` when it should be using `require().default` because `coffee-react` doesn't like backticked import statements.

srallen:
Yep, I haven't figured out where that was happening yet, though I haven't spent a whole lot of time on it.

eatyourgreens:
The slideshow frame controls look a bit weird for subjects that have multiple frames.
https://collection-preview-es6.pfe-preview.zooniverse.org/collections/eatyourgreens/diagnosis-london-favourites?env=production
Should they be styled as small buttons?

srallen:
I see. I had wanted to decouple talk styles from being reused elsewhere. I'll see if I can get these styles correctly applied again.

srallen:
@eatyourgreens I think I have the styles I broke fixed now. The reasoning to begin to decouple the styles was that @beckyrother has a design for the subject viewer in collections that we weren't able to implement easily since the subject viewer is shared. I'm hoping this is one tiny step toward separation of concerns with it. If we do find some styles should be shared, we should leverage stylus variables or functions.

PR #3743, opened 2017-04-21
Break out SVG annotations from frame annotations.

eatyourgreens:
Describe your changes.
Separate SVG drawings out of the frame annotator so that we can add renderers for annotations for subjects other than images.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://generic-subject-marking.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/annotation-renderer/image.jsx
+++ b/app/classifier/annotation-renderer/image.jsx
@@ -1,0 +1,224 @@
+import React from 'react';
+import Draggable from '../../lib/draggable';
+import tasks from '../tasks';
+import getSubjectLocation from '../../lib/get-subject-location';
+import SVGImage from '../../components/svg-image';
+
+export default class ImageAnnotator extends React.Component {
+  constructor(props) {
+    super(props);
+    this.getScreenCurrentTransformationMatrix = this.getScreenCurrentTransformationMatrix.bind(this);
+    this.getEventOffset = this.getEventOffset.bind(this);
+    this.normalizeDifference = this.normalizeDifference.bind(this);
+    this.eventCoordsToSVGCoords = this.eventCoordsToSVGCoords.bind(this);
+  }
+
+  getSizeRect() {
+    const clientRect = this.sizeRect && this.sizeRect.getBoundingClientRect();
+
+    if (clientRect) {
+      const { width, height } = clientRect;
+      let { left, right, top, bottom } = clientRect;
+      left += pageXOffset;
+      right += pageXOffset;
+      top += pageYOffset;
+      bottom += pageYOffset;
+      return { left, right, top, bottom, width, height };
+    }
+    return { left: 0, right: 0, top: 0, bottom: 0, width: 0, height: 0 };
+  }
+
+  getScale() {
+    const sizeRect = this.getSizeRect();
+    const horizontal = sizeRect ? sizeRect.width / this.props.naturalWidth : 0;
+    const vertical = sizeRect ? sizeRect.height / this.props.naturalHeight : 0;
+
+    return { horizontal, vertical };
+  }
+
+  // get current transformation matrix
+  getScreenCurrentTransformationMatrix() {
+    const svg = this.refs.svgSubjectArea;
+    return svg.getScreenCTM();
+  }
+
+  // find the original matrix for the SVG coordinate system
+  getMatrixForWindowCoordsToSVGUserSpaceCoords() {
+    const transformationContainer = this.refs.transformationContainer;
+    return transformationContainer.getScreenCTM().inverse();
+  }
+
+  // get the offset of event coordiantes in terms of the SVG coordinate system
+  getEventOffset(e) {
+    return this.eventCoordsToSVGCoords(e.clientX, e.clientY);
+  }
+
+  // transforms the difference between two event coordinates
+  // into the difference as represent in the SVG coordinate system
+  normalizeDifference(e, d) {
+    const difference = {};
+    const normalizedPoint1 = this.eventCoordsToSVGCoords(e.pageX - d.x, e.pageY - d.y);
+    const normalizedPoint2 = this.eventCoordsToSVGCoords(e.pageX, e.pageY);
+    difference.x = normalizedPoint2.x - normalizedPoint1.x;
+    difference.y = normalizedPoint2.y - normalizedPoint1.y;
+    return difference;
+  }
+
+  // transforms the event coordinates
+  // to points in the SVG coordinate system
+  eventCoordsToSVGCoords(x, y) {
+    const svg = this.refs.svgSubjectArea;
+    const newPoint = svg.createSVGPoint();
+    newPoint.x = x;
+    newPoint.y = y;
+    const matrixForWindowCoordsToSVGUserSpaceCoords = this.getMatrixForWindowCoordsToSVGUserSpaceCoords();
+    const pointforSVGSystem = newPoint.matrixTransform(matrixForWindowCoordsToSVGUserSpaceCoords);
+    return pointforSVGSystem;
+  }
+
+  render() {
+    let taskDescription;
+    let TaskComponent;
+    const { type, src } = getSubjectLocation(this.props.subject, this.props.frame);
+    const createdViewBox = `${this.props.viewBoxDimensions.x} ${this.props.viewBoxDimensions.y} ${this.props.viewBoxDimensions.width} ${this.props.viewBoxDimensions.height}`;
+
+    if (this.props.annotation) {
+      taskDescription = this.props.workflow.tasks[this.props.annotation.task];
+      TaskComponent = tasks[taskDescription.type];
+    }
+
+    const svgStyle = {};
+    if (type === 'image' && !this.props.loading) {
+      // Images are rendered again within the SVG itself.
+      // When cropped right next to the edge of the image,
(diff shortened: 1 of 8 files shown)


eatyourgreens:
`InsideSubject` hooks could be SVG or HTML. There needs to be a filter that sends SVG to the SVG rendering component, HTML to any HTML rendering component. Otherwise, rendering is going to break when it tries to render SVG `<g>` elements inside a HTML `<div>`.

Maybe the components should render based on task type, rather than subject type? How would that work for persistent tasks, such as drawing tasks? Perhaps a task adds its renderer to the classifier and the renderer then persists for subsequent tasks, regardless of their type.

app/classifier/annotation-renderer/image.jsx
```diff
@@ -0,0 +1,224 @@
...
+
+  render() {
+    let taskDescription;
+    let TaskComponent;
+    const { type, src } = getSubjectLocation(this.props.subject, this.props.frame);
+    const createdViewBox = `${this.props.viewBoxDimensions.x} ${this.props.viewBoxDimensions.y} ${this.props.viewBoxDimensions.width} ${this.props.viewBoxDimensions.height}`;
+
+    if (this.props.annotation) {
+      taskDescription = this.props.workflow.tasks[this.props.annotation.task];
+      TaskComponent = tasks[taskDescription.type];
```
CKrawczyk: `TaskComponent` is never used after assignment.

app/classifier/frame-annotator.jsx
```diff
@@ -114,11 +46,11 @@ export default class FrameAnnotator extends React.Component {
     let InsideSubject;
     let AfterSubject;
     const { type, src } = getSubjectLocation(this.props.subject, this.props.frame);
```
CKrawczyk: `src` is no longer used in this file.

app/classifier/annotation-renderer/image.jsx
```diff
@@ -0,0 +1,224 @@
...
+      }
+    });
+
+    const children = React.Children.map(this.props.children, (child) => {
+      return React.cloneElement(child, hookProps);
+    });
+
+    return (
+      <svg
+        ref="svgSubjectArea"
```
CKrawczyk: Can this `ref` be safely changed to the new format or is it still in use elsewhere in the code?

app/classifier/annotation-renderer/image.jsx
```diff
@@ -0,0 +1,224 @@
...
+        className="subject"
+        style={svgStyle}
+        viewBox={createdViewBox}
+        {...svgProps}
+      >
+        <g
+          ref="transformationContainer"
```
CKrawczyk: Same as above.

CKrawczyk:
I just noticed that one of the `FrameAnnotator` tests will have to be updated as well: https://github.com/zooniverse/Panoptes-Front-End/blob/f46f178024d9ba8ce896e7cf8d050f3567c3f19c/app/classifier/frame-annotator.spec.js#L59-L85

app/classifier/annotation-renderer/image.jsx
eatyourgreens: Hmmm, updating the refs seems to have broken drawing tasks. `this.svgSubjectArea` is null but I can't see why.

eatyourgreens:
Yep. I'm leaving that test broken for the time being, so that it stops this PR from being merged.

app/classifier/annotation-renderer/image.jsx
eatyourgreens: Turns out the ref callback is invoked with `null` as an argument when the component updates. https://www.reactenlightenment.com/basic-react-components/6.9.html

eatyourgreens:
Fixed up the refs commit to handle null arguments, rebased and staged again.

eatyourgreens:
@CKrawczyk 
>As for the InsideSubject hooks, could a prop be passed down indicating if it is in an SVG element and have each InsideSubject render method check this prop before returning an element? That way the VIEWERS just pass down a prop and each InsideSubject component decides if they can render based on that prop.
Maybe the marking tool renderers should specify whether their output is HTML or SVG? I'll play around with that idea.

The hooks are slightly confusing because `BeforeSubject` and `AfterSubject` should return HTML, but `InsideSubject` is currently expected to return SVG. So the rendering method varies even within a particular task.

app/classifier/annotation-renderer/index.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
+import SVGAnnotator from './svg';
+
+const VIEWERS = {
+  image: SVGAnnotator
+};
+
+function AnnotationRenderer(props) {
+  const Renderer = VIEWERS[props.type];
```
eatyourgreens: Probably should have a default component for the cases where this returns null.

app/classifier/frame-annotator.jsx
```diff
@@ -137,54 +67,34 @@ export default class FrameAnnotator extends React.Component {
...
+      .keys(tasks)
+      .map((taskName) => {
+        const PersistInsideSubject = tasks[taskName].PersistInsideSubject;
+        if (PersistInsideSubject) {
+          return <PersistInsideSubject key={taskName} {...hookProps} />;
+        }
+        return null;
+      })
+      .filter(Boolean);
+    subjectChildren = subjectChildren.concat(persistentHooks);
```
eatyourgreens: These hooks should be pushed down to the individual renderers, since `TaskComponent.InsideSubject` and `TaskComponent.PersistInsideSubject` might generate HTML or SVG.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -0,0 +1,235 @@
...
+
+    Object.keys(tasks).map((task) => {
+      const Component = tasks[task];
+      if (Component.getSVGProps) {
+        Object.assign(svgProps, Component.getSVGProps(hookProps));
+      }
+    });
+
+    let children = [];
+    const isDrawingTask = taskDescription && (taskDescription.type === 'drawing' || taskDescription.type === 'crop');
```
eatyourgreens: Not sure about the extensibility of this, as we add more tasks. Might be better if `taskDescription` specified its output format somehow.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -0,0 +1,235 @@
...
+    if (isDrawingTask && InsideSubject && !this.props.panEnabled) {
+      children.push(<InsideSubject key="inside" {...hookProps} />);
+    }
+    const persistentHooks = ['drawing', 'crop']
```
eatyourgreens: Same comment here re. hardcoding task names into this component.

eatyourgreens:
Tests should be fixed now too.

CKrawczyk:
It looks like the `InsideSubject` and `PersistInsideSubject` elements are not being rendered after the latest changes.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -0,0 +1,237 @@
...
+
+    Object.keys(tasks).map((task) => {
+      const Component = tasks[task];
+      if (Component.getSVGProps) {
+        Object.assign(svgProps, Component.getSVGProps(hookProps));
+      }
+    });
+
+    let children = [];
+    const isDrawingTask = taskDescription && (tasks[taskDescription.type].AnnotationRenderer === SVGRenderer);
```
CKrawczyk: For some reason `tasks[taskDescription.type].AnnotationRenderer === SVGRenderer` always evaluates to `false` for all tasks.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -0,0 +1,237 @@
...
+    if (isDrawingTask && InsideSubject && !this.props.panEnabled) {
+      children.push(<InsideSubject key="inside" {...hookProps} />);
+    }
+    const persistentHooks = Object
+      .keys(tasks)
+      .filter((key) => { return tasks[key].AnnotationRenderer === SVGRenderer; })
```
CKrawczyk: See above.

eatyourgreens:
Interesting. They are rendering in the tests and in Chrome.

eatyourgreens:
@CKrawczyk which browser isn't rendering the crop and drawing tasks? They are both working for me in Firefox, Chrome and Safari.

CKrawczyk:
Both chrome and firefox, but as I just found out it is an issues with the react-hot-reloading, not the base code.

srallen:
FYI, our current hot reload configuration is the old, now unsupported way of doing it. Migrating to the new configuration I'm pretty sure would mean converting the rest of the app to ES6 because it depends on the babel/webpack tree-shaking feature, so I didn't really dig into trying to make it work.

app/classifier/annotation-renderer/svg.jsx
simensta: After a few renders, when `taskDescription` is no longer null, I am seeing `tasks[taskDescription.type].AnnotationRenderer === SVGRenderer` evaluate to true for drawing tasks.
@CKrawczyk, maybe `taskDescription` is still null?

simensta:
I am happy with these changes, but I don't use `npm run serve-hot` often, if at all. Is it okay if that breaks until PFE is fully converted to ES6?

PR #3749, opened 2017-04-27
[RFC] Add Mobile section to PFE lab builder

robinschaaf:
Add a mobile section to the PFE lab builder workflow page to both promote the new swipe app functionality and allow project owners to opt in or out of having their workflow in the mobile app.
  *  Displays only for tasks with a question type
  *  Displays a message with criteria for getting into the swipe app.  If the workflow fails any of these criteria it will display a message, disable the checkbox and indicate which criteria it's failing

Note that there is a chance that a user could opt in, and then later change the workflow so that it no longer works.  There is an additional check in the mobile app to filter out these workflows.

# Review Checklist

- [x] Does it work on mobile?

## Optional


wgranger:
Here is a preview of the section:

screen shot 2017-05-02 at 11 11 07 am

**Items to consider:**

- There was discussion that a project should only show up in the mobile app if listed `launch_approved: true`. Looks like a change will need to be made on the mobile app to filter for launch approved projects when calling the API to keep the mobile app from filling up with non-approved projects.

- This check box will set the _workflow_ resource to `swipe_enabled: true` under the `configuration` tag. We had also discussed that the project resource would also be toggled to `mobile_friendly: true` if not currently true.

wgranger:
Staged at: https://add-mobile.pfe-preview.zooniverse.org/

wgranger:
Another item that needs to be clarified:

Projects will show up in the native mobile app if the project resource has the `mobile_friendly: true` tag. Once a project is in the native mobile app, _all_ workflows will be available to the user. Workflows that are _not_ swipe enabled will show up in the native app webview. 

Workflows will open in the swipe classifier if the workflow resource has the `configuration.swipe_enabled: true` tag.

app/classifier/tasks/mobile-section.jsx
```diff
@@ -0,0 +1,145 @@
...
+    super(props);
+    this.toggleSwipeEnabled = this.toggleSwipeEnabled.bind(this);
+  }
+
+  toggleSwipeEnabled(e) {
+    const changes = {};
+    changes['configuration.swipe_enabled'] = e.target.checked;
+    this.props.workflow.update(changes).save();
+    if (e.target.checked) {
+      this.props.project.update({ mobile_friendly: e.target.checked }).save();
```
wgranger: Note: This part of the code also sets the project to `mobile_friendly: true` if not currently active.

app/classifier/tasks/mobile-section.jsx
```diff
@@ -0,0 +1,145 @@
...
+    return (
+      <div>
+        <span className="form-label">Mobile App</span>
+        <br />
+        <div className="workflow-mobile-form">
+          <div className="workflow-mobile-form-left-panel">
+            <label
+              className="pill-button"
+              htmlFor="swipe_enabled"
+              title={mobileHelp}
```
wgranger: This will not be visible if the style of the element sets `pointerEvents: none`. Might be best to disable this another way.

app/classifier/tasks/mobile-section.spec.js
```diff
@@ -0,0 +1,412 @@
...
+    before(function () {
+      wrapper = shallow(<MobileSection
+        task={otherTypeTask}
+        workflow={defaultWorkflow}
+        project={defaultProject}
+      />);
+    });
+
+    it('should not render for other task type', function () {
+      assert.equal(wrapper.find('input').props().disabled, false);
```
wgranger: This is a bit odd looking because it's the same check for the above 'does not disable checkbox'. Checking for a particular prop on an item that shouldn't exist seems counter intuitive.

srallen:
> There was discussion that a project should only show up in the mobile app if listed launch_approved: true. 
This should be happening.

> Should this section only become visible for launch approved projects?
Yes

> We also discussed an email notification system to a Zooniverse team member once checked (similar to how beta review emails are sent) so that we are aware a swipe friendly workflow has been initiated should we need to remove that project from the mobile app in case the workflow isn't appropriate for the native app.
This would require a lot of overhead on our part and I don't think we have the time to manage this in addition to the standard beta reviews. We should consider doing as much validation either in the front end client or with the API as much as possible.

rogerhutchings:
Right - this PR needs some love and needs merging. I'm going to close it in favour of a branch on this repo (rather than @robinschaaf's fork, just for ease of admin), rebase it and address the items mentioned in review.

rogerhutchings:
Closed in favour of #3894.

PR #3757, opened 2017-05-03
Add James Artz bio.

jelliotartz:
Add bio and pic for me.

https://james-bio.pfe-preview.zooniverse.org/about/team

# Review Checklist

- [ ] Does it work on mobile?

## Optional


app/pages/about/team-page.cjsx
```diff
@@ -122,6 +122,11 @@ counterpart.registerTranslations 'en',
       hughDickinson:
         title: 'Researcher'
         bio: '''Hugh is a postdoc studying the Zooniverse.'''
+      jamesArtz:
+        title: 'Developer'
+        bio: '''James joined the Zooniverse team in 2017. Before discovering his passion for code,
```
amyrebecca: Not a big deal, but since you're using the `heredoc` syntax, these line breaks are actually meaningful. It'd be better to keep it on one line, even if it is really long.

jelliotartz: Ah, ok, thank you! I'll keep that in mind.

PR #3768, opened 2017-05-08
Stop playing on subject change

### LINKED ISSUE #3729: Echo: After finishing the classification, new subject loads but the previous one is actually displayed
## Expected behavior
When you finish the classification, new subject should load and display. When you visit the Talk, you should see the same subject you have just classified.

## Current behavior
After finishing the classification, new subject does load but the previous one is actually displayed. When you go to the Talk, you see something else than you saw on the Classification page.
(Happened to me 3 times in a row on AmazonCam Tambopata, and another volunteer is experiencing the same from two different computers, even with a fast internet connection). It doesn't happen all the time.
_Will provide some browser console screenshots when it happens the next time._

## Steps to replicate:

1. Classify a subject (call it subject A) on AmazonCam Tambopata.
2. When finished classifying, click "Done".
3. If you're "lucky", you'll get a new subject (subject B) as expected but you'll actually see the same as in the step 1 (subject A). (You won't see how the subject B looks like).
4. Classify the subject (you see subject A but it is actually subject B), and click "Done & Talk". 
5. Subject B will load on the Talk although you were expecting to see the subject A there (It was the one you classified).

## Additional information:
It may look like a repeated subject (with no *Already seen* banner) but when you classify it and go to the Talk, it displays the new subject and not the one you saw in the classification. 
It is not always obvious that the subject you see is the same one as you have just finished. Especially when there are no animals present...
Discussed also here: https://www.zooniverse.org/projects/rainforestexpeditions/amazoncam-tambopata/talk/683/286699

- **Operating system: Linux x86_64** (for me; will ask what the other volunteer uses)
- **Browser: Chromium 48** (for me; will ask what the other volunteer uses)

eatyourgreens:
Fixes #3729.

Describe your changes.
Stop the slideshow player on subject change, and reset the frame counter.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3729.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -69,10 +69,15 @@ module.exports = React.createClass
   componentWillReceiveProps: (nextProps) ->
     unless nextProps.subject is @props.subject
       clearTimeout @signInAttentionTimeout
-      @setState loading: true
+      @setState
+        playing: false
+        loading: true
+        frame: 0
 
   componentDidUpdate: (prevProps) ->
     if @props.subject isnt prevProps.subject
+      # turn off the slideshow player and reset any counters
+      @setPlaying false
       @setState frame: @getInitialFrame()
 
   logSubjClick: (logType) ->


eatyourgreens:
Outside the scope of this bug fix: I think the slideshow player might be better extracted to its own component.

PR #3769, opened 2017-05-08
Allow feedback to override skipping classification summaries

### LINKED ISSUE #3762: Allow feedback summary to override no classification summary experimental feature
The no classification summary experimental feature prevents any classification summary from being seen, even if there is feedback. This should be overridden if there's feedback for the user.

cc #3725 @aprajita

rogerhutchings:
Fixes #3762. Requires #3725 merging first.

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -17,7 +17,7 @@ import experimentsClient from '../lib/experiments-client';
 import TaskNav from './task-nav';
 import ExpertOptions from './expert-options';
 import { connect } from 'react-redux';
-import { isThereFeedback } from './feedback/helpers';
+import * from './feedback/helpers';
 
 // For easy debugging
 window.cachedClassification = CacheClassification;
@@ -152,7 +152,7 @@ class Classifier extends React.Component {
   }
 
   completeClassification() {
-    const feedbackActive = this.props.project.experimental_tools && this.props.project.experimental_tools.includes('general feedback');
+    const feedbackActive = isFeedbackActive(this.props.project);
     
     if (feedbackActive) {
       const classificationMetadata = Object.assign({}, this.props.classification.metadata, {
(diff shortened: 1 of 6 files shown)


rogerhutchings:
Cool, thanks for taking a look :) In theory, all that needs to happen is:
- add a new form for survey tasks to the lab workflow editor
- add a new processor to https://github.com/zooniverse/Panoptes-Front-End/tree/master/app/classifier/feedback

The lab forms could probably be DRYd out a bit, but the processor's probably not a big deal; I've got a PR incoming that refactors a lot for point tools, which has made life easier.

eatyourgreens:
This PR has removed the Done & Talk button from Gravity Spy and Steller Watch. I'm not completely sure I understand [the new logic that checks to remove that button](/zooniverse/Panoptes-Front-End/pull/3769/files#diff-eb97effae530e7721df8de3dc206a73eR169).

app/classifier/classifier.jsx
```diff
@@ -162,7 +165,9 @@ class Classifier extends React.Component {
     }
 
     if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
-      this.props.onCompleteAndLoadAnotherSubject();
+      if (!feedbackActive || (feedbackActive && !isThereFeedback(this.props.subject, this.props.workflow))) {
```
eatyourgreens: This check is different from the later check in the TaskNav component.

Probably not a good idea to have those two different checks in place. I think the code only really works properly if both return the same result, which isn't guaranteed.

app/classifier/task-nav.jsx
```diff
@@ -165,7 +166,7 @@ class TaskNav extends React.Component {
             >
               Back
             </button>}
-          {(!nextTaskKey && this.props.workflow.configuration.hide_classification_summaries && this.props.project && !disableTalk) &&
+          {(!nextTaskKey && this.props.workflow.configuration.hide_classification_summaries && this.props.project && !disableTalk && (!isFeedbackActive(this.props.project) && isThereFeedback(this.props.subject, this.props.workflow))) &&
```
eatyourgreens: This will return a different value than the check that's been added to the `completeClassification` method in the classifier.

Issue #3363, opened 2016-12-31
adding in a button to go directly to the project's project builder page if you're a project owner or collaborator


mschwamb:
It would be nice if you're logged in and a project collaborator or owner that you could go from the project pages or Talk directly to  project builder for that project rather than having to click on the zooniverse menu bar and and the lab icon and then select the project. Maybe there could be a gear icon on the top right bar where the login/logout and profile buttons are that you could click.

mschwamb:
Maybe I'm the only odd usecase for this, but having three projects on panoptes, I cycle between them checking talk, etc, that I find it cumbersome to always have to click on the 'Build a project' and then go find the project in the list to be able to get access to the project's project builder page. It would be really nice if there was a 'gear' icon or something that I could click on the top navigation bar and it would take me directly my project's project builder page. I could also see this being  really handy building a  project and you want to check how things look and then make a change. Any chance a shortcut like this could get added to the navigation bar for project admins?

mschwamb:
It's low priority, but generally I see something on talk or the project page and then need to go into the project builder so it would be nice if there was a gear icon or something so that the project owner/research team could save  multiple clicks.  Also very useful when you're designing the project because you go back and forth between the interface and the workflow editor often

PR #3775, opened 2017-05-10
[2 of 4] New Signed Out Homepage (new navbar)

wgranger:
**Describe your changes:**
This branch includes changes to the nav bar to be merged into the new-signed-out-homepage branch. 

_Note_: This branch will merge into that other branch and not master.

https://new-navbar-style.pfe-preview.zooniverse.org/

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/layout/account-bar.jsx
+++ b/app/layout/account-bar.jsx
@@ -91,14 +91,14 @@ export default class AccountBar extends React.Component {
       <span className="account-bar">
 
         <NotificationsLink isMobile={this.props.isMobile} params={this.props.params} user={this.context.user} linkProps={{
-          className: 'site-nav__link site-nav__icon site-nav__icon--notifications',
+          className: 'site-nav__link',
           activeClassName: 'site-nav__link--active',
           onClick: this.logClick ? this.logClick.bind(this, 'accountMenu.notifications') : null
         }} />
 
         <Link
           to="/inbox"
-          className="site-nav__link site-nav__icon site-nav__icon--inbox"
+          className="site-nav__link site-nav__link--inbox"
           activeClassName="site-nav__link--active"
           aria-label={`
             Inbox ${this.state.unread ? 'with unread messages' : ''}
--- a/app/layout/account-bar.jsx
+++ b/app/layout/account-bar.jsx
@@ -21,7 +21,7 @@ counterpart.registerTranslations('en', {
   }
 });
 
-class AccountBar extends React.Component {
+export default class AccountBar extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
@@ -71,10 +71,26 @@ class AccountBar extends React.Component {
     auth.signOut();
   }
 
+  renderMessages() {
+    const mobile = this.state.unread ? <i className="fa fa-envelope fa-fw" /> : <i className="fa fa-envelope-o fa-fw" />;
+
+    if (this.props.isMobile) {
+      return mobile;
+    } else {
+      return (
+        <div>
+          <Translate content="accountMenu.messages" />
+          {this.state.unread && (` (${this.state.messageCount})`)}
+        </div>
+      );
+    }
+  }
+
   render() {
     return (
       <span className="account-bar">
-        <NotificationsLink params={this.props.params} user={this.context.user} linkProps={{
+
+        <NotificationsLink isMobile={this.props.isMobile} params={this.props.params} user={this.context.user} linkProps={{
           className: 'site-nav__link site-nav__icon site-nav__icon--notifications',
           activeClassName: 'site-nav__link--active',
           onClick: this.logClick ? this.logClick.bind(this, 'accountMenu.notifications') : null
@@ -95,8 +111,7 @@ class AccountBar extends React.Component {
               ${this.state.unread ? 'site-nav__inbox-link--unread' : ''}
             `.trim()}
           >
-            <Translate content="accountMenu.messages" />
-            {this.state.unread && (` (${this.state.messageCount})`)}
+            {this.renderMessages()}
           </span>
         </Link>
 
@@ -120,7 +135,6 @@ class AccountBar extends React.Component {
                 className="site-nav__link"
                 onClick={this.logClick ? this.logClick.bind(this, 'accountMenu.profile') : null}
               >
-                <i className="fa fa-user fa-fw" />{' '}
                 <Translate content="accountMenu.profile" />
               </Link>
               <br />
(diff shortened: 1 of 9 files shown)


eatyourgreens:
This looks really ugly on a project page. The black doesn't fill the width of the viewport.
Screenshot of the new nav bar on a project, showing that it only fills about half the viewport width.

app/layout/account-bar.jsx
```diff
@@ -17,55 +16,57 @@ const DOWN = 40;
...
-  },
+
+    this.handleAccountMenuOpen = this.handleAccountMenuOpen.bind(this);
+    this.navigateMenu = this.navigateMenu.bind(this);
+    this.handleSignOutClick = this.handleSignOutClick.bind(this);
+  }
 
   componentDidMount() {
     addEventListener('locationchange', this.lookUpUnread);
     this.lookUpUnread();
```
eatyourgreens: For this listener to work, `this.lookUpUnread` should be bound to the component's `this` context in the constructor.

app/layout/account-bar.jsx
```diff
@@ -77,47 +78,88 @@ const AccountBar = React.createClass({
...
-      type: 'logout',
+      type: 'logout'
     });
 
     auth.signOut();
-  },
+  }
+
+  renderMessages() {
+    const mobile = this.state.unread ? <i className="fa fa-envelope fa-fw" /> : <i className="fa fa-envelope-o fa-fw" />;
```
eatyourgreens: This should have a text label for VoiceOver to read out.

app/layout/account-bar.jsx
```diff
@@ -77,47 +78,88 @@ const AccountBar = React.createClass({
...
+        </Link>
+
         <TriggeredModalForm
           ref="accountMenuButton"
           className="site-nav__modal"
           trigger={
             <span className="site-nav__link">
               <strong>{this.context.user.display_name}</strong>{' '}
-              <Avatar className="site-nav__user-avatar" user={this.context.user} size="2em" />
+              <i className="fa fa-chevron-down" />
```
eatyourgreens: People might miss having their avatar displayed here 😞

srallen:
Is there really any reason to keep doing the demoted/promoted functionality and styling anymore with the projects with this design? @beckyrother can you share your design thoughts on the navbar changes?

eatyourgreens:
True, that. Also, it's kind of disconcerting that the black navigation changes back to the teal design when you log in. This is one component that's going to have to change globally, for the whole site.

beckyrother:
Can this be a global change? @eatyourgreens is right, it's odd for the navigation to change when you log in. 
@srallen I'm not sure what you mean about demoted/promoted functionality, but my thinking behind this is that a darker bar separate from the content will fade more into the background and be less distracting. I'd like this to be visible all the time, so it replaces the hamburger menu with the motion. In the classify/project section it'll also allow the project's background image and avatar to shine and really obviously differentiate between project navigation and site navigation. Here's an example of that: 
page-dark_png

srallen:
The demoted/promoted is the hamburger menu and that animation (demoted/promoted are the variables and class names that are being used in the code). Thanks for clarifying.

We should go forward with removing that, since it'll make styling and the javascript simpler.

eatyourgreens:
On OSX, the blue focus ring is lost against the teal background, making it difficult to see which item is selected when you are using a keyboard to select from the menu.
User account menu showing blue OSX is barely visible against a teal background
I think any `:hover` styles should also apply to the `:focus` pseudo-class.

wgranger:
This has been restaged with the demoted functionality removed. That functionality hit a good bit of code, but I couldn't find any bugs after removing it from this branch.

CKrawczyk:
The user dropdown is causing horizontal scrolling on my Mac (both chrome and firefox).

screen shot 2017-05-15 at 9 42 54 am

wgranger:
@CKrawczyk Let me know if you're still experiencing the horizontal scrolling issue from above. I'm unable to replicate it on the [staged branch above](https://new-navbar-style.pfe-preview.zooniverse.org/). I thought I took care of that in a previous commit, but I could be missing something.

mschwamb:
On Mac OS when using Firefox, the first item in the account dropdown is always selected (I think - it's got the white box around it) even after you move past it, which is confusing about what's really selected. 

screen shot 2017-05-15 at 11 35 58 am

I didn't experience the horizontal scrolling.

mschwamb:
Maybe not the right place for this... Also if you're updating the navigation bar, can you please consider adding a shortcut to the specific project lab page if you're a project owner or researcher on the project. Having three panoptes projects, I find it a little cumbersome to always have to click on the lab then  click on the list to the project I want. Usually I see something needs editing when I'm at the site on Talk.  See ticket #3363

CKrawczyk:
I am still seeing the scrolling on my computer (I even tried incognito mode).

wgranger:
Ok. A fix might need to be added to the modal-form repo, because this is also happening when I open the main site on my phone.

wgranger:
@eatyourgreens Do you have a suggestion (considering accessibility) to the hover/focus comment made by Meg above? The functionality on this branch is similar to how we currently have it on master.

eatyourgreens:
@wgranger that's there for keyboard accessibility. Check it out on https://www.zooniverse.org.

The first item is selected when you open the menu. Use the up/down cursor keys to move up and down the menu, enter to activate, esc to close the menu and focus the original button.

You should style the menu so that any `:hover` styles also use `:focus`, I think.

wgranger:
The hover and focus styles on the menu are currently shared. I'm mainly curious if there's a way to only focus/hover one item at a time (rather than one item receiving focus while another has hover). I think that was the main issue brought up above.

eatyourgreens:
Oh right. No, hover doesn't focus an item. Only clicking will do that (and then not always depending on the browser.)

OSX usually has a bright blue ring to indicate focus, but if I remember rightly, the original issue was that that ring doesn't show up against the teal background.

Maybe have a chat with @beckyrother about how focus should be indicated here, as compared to mouse hover or touch.

Another option is to leave focus with the original element when the menu opens, then enter the menu with the down arrow. See the search box on twitter.com for an example of this. I can't remember if our components allow for this, though.

beckyrother:
Good point @eatyourgreens. Here's a pdf with what I'm thinking for the hover vs selected:

[nav-signedin-dropdown.pdf](https://github.com/zooniverse/Panoptes-Front-End/files/1019585/nav-signedin-dropdown.pdf)

First page is before you hover over anything else, second shows the hover state. Things to note, and things I've noted from the current PR:
- highlight box goes across the entire dropdown box
- no line between Favorites and Sign Out
- letter-spacing should be 0.15em across the entire nav bar

Let me know if there are any questions. Thanks!

wgranger:
There's currently a [PR](https://github.com/zooniverse/modal-form/pull/31) open in the modal-form repo to fix the issue with the menu being shown offscreen.

beckyrother:
That's great @eatyourgreens, thanks!

PR #3778, opened 2017-05-10
Pass annotations explicitly to the markings renderer

eatyourgreens:
Fixes the `fauxClassification` hack in the combo task.

Describe your changes.
Adds an annotations prop for passing annotations to `MarkingsRenderer`.
Allows the combo task to override the classification and pass its own set of marks to render.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://drawing-annotations.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/combo/markings-renderer.jsx
+++ b/app/classifier/tasks/combo/markings-renderer.jsx
@@ -1,0 +1,60 @@
+import React from 'react';
+
+export default function MarkingsRenderer(props) {
+  // a list that holds the annotations for the current combo task
+  let currentComboAnnotations = [];
+  const allTaskTypes = props.classification.annotations.map( annotation => props.workflow.tasks[annotation.task].type);
+  const i = allTaskTypes.lastIndexOf('combo');
+  if (i > -1) {
+    currentComboAnnotations = props.classification.annotations[i].value;
+  }
+  // a list that holds the annotations for all combo tasks
+  let allComboAnnotations = [];
+  const allComboTypes = [];
+  props.classification.annotations.map( (annotation) => {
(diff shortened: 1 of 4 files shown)


app/classifier/tasks/combo/markings-renderer.jsx
```diff
@@ -0,0 +1,60 @@
...
+        .map((taskType) => {
+          const TaskComponent = props.taskTypes[taskType];
+          if (TaskComponent.PersistInsideSubject) {
+            // when a combo annotation changes make sure the combo annotation updated correctly with only the
+            // curreny combo task's annotatons.  This is a hack to make drawing tasks work in a combo task.
+            function fauxChange(annotation) {
+              props.onChange(Object.assign({}, props.annotation, { value: currentComboAnnotations }));
+            };
+            let fauxAnnotation = props.annotation;
+            if (props.annotation && props.annotation.task && props.workflow.tasks && props.workflow.tasks[props.annotation.task].type === 'combo') {
```
eatyourgreens: This is pretty horrible code. It would be good to get rid of it too.

CKrawczyk: I know at least one of those checks was placed there to make a test pass (the test passes an `annotation` without a `task`, something I don't think can happen in production).

eatyourgreens: `annotation` defaults to `null` in a few places in production, so this guarding against that.

CKrawczyk: The `props.annotation` is, but the `&& props.annotation.task` to check that a non-null annotation has a `task` key.

PR #3780, opened 2017-05-10
Don't let deleted workflows be default

srallen:
Fixes a scenario where a project sets a workflow as default and then deletes it. This removes the project workflow default if it is the default. 

https://remove-default-if-deleted.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -661,6 +661,10 @@ EditWorkflowPage = React.createClass
     if confirmed
       @setState deletionInProgress: true
 
+      if @props.workflow.id is @props.project.configuration?.default_workflow
+        @props.project.update 'configuration.default_workflow': null
+        @props.project.save()
+
       @props.workflow.delete().then =>
         @props.project.uncacheLink 'workflows'
         @context.router.push "/lab/#{@props.project.id}"
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -624,7 +624,7 @@ EditWorkflowPage = React.createClass
       @props.project.update 'configuration.default_workflow': @props.workflow.id
       @props.project.save()
     else
-      @props.project.update 'configuration.default_workflow': null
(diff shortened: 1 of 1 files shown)


app/pages/lab/workflow.cjsx
```diff
@@ -661,6 +661,10 @@ EditWorkflowPage = React.createClass
     if confirmed
       @setState deletionInProgress: true
 
+      if @props.workflow.id is @props.project.configuration?.default_workflow
+        @props.project.update 'configuration.default_workflow': null
+        @props.project.save()
```
eatyourgreens: Would `delete @props.project.configuration.default_workflow` be better than setting the value to null?

eatyourgreens:
LGTM. I'm not sure if it would be better to delete that configuration key outright, but that isn't a blocking issue.

srallen:
I copied what I did in the checkbox handler: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/workflow.cjsx#L627-L628

I don't recall if I'm checking for null explicitly anywhere. If it's not needed, I can change both to remove the property entirely.

eatyourgreens:
Deleting the property seems better to me. Off the top of my head, I can't think of a case where we'd explicitly want to check this against `null`, particularly since workflow ID is a string but `null` is an Object.

srallen:
Using the `delete` operator was for some reason causing a 422 on the PUT for the project. Setting the property to undefined seems to work though.

eatyourgreens:
Weird. 😕

camallen:
@srallen ooi what was the payload for the 422 after deleting the key?

srallen:
@camallen the payload for the PUT on the project was `http_cache: true, projects: {}`. My guess is coffeescript was doing something odd with the `delete` operator by making the whole object empty instead of just deleting the one property in the config object.

camallen:
intriguing, one for the memory banks.

PR #3788, opened 2017-05-11
Mark up project status tabs as tabs

eatyourgreens:
Adds tab roles and states to the Active/Paused/Finished buttons on the Projects page. Styles them according to selected state.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3784.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/css/secondary-page.styl
+++ b/css/secondary-page.styl
@@ -62,6 +62,7 @@
               background: rgba(255,255,255, 0.2)
               color: rgba(255,255,255,1)
             &.active
+            &[aria-selected=true]
               color: white
               border-bottom: solid 2px white
               border-top: 0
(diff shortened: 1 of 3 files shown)


css/secondary-page.styl
```diff
@@ -62,6 +62,7 @@
               background: rgba(255,255,255, 0.2)
               color: rgba(255,255,255,1)
             &.active
+            &[aria-selected=true]
```
rogerhutchings: Should these also be styled to match the top nav active states? (Different padding, no bold)

eatyourgreens: Dunno, these are the styles that are already in use on the live site. :man_shrugging:

rogerhutchings: fair enough

eatyourgreens: By the way, I added the aria roles to these because the Projects page is using buttons to filter the projects list, whereas other pages use links for the secondary nav.

PR #3789, opened 2017-05-11
Check for image src before setting a collection preview

eatyourgreens:
Without this small change, empty collections break the collections list page and preview images don't appear.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-collection-cards.pfe-preview.zooniverse.org/

## Optional


eatyourgreens:
Before:
Screenshot showing previews missing for two collections that do have subjects

After:
Screenshot showing preview images for all collections containing at least one subject

PR #3473, opened 2017-02-10
Remove PromiseRenderer from subject-set.cjsx

simensta:
Removes the PromiseRenderer from subject-set.cjsx, related to #3216. Staged [here](https://subject-set-remove-promise-render.pfe-preview.zooniverse.org/).

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] Did you deploy a staging branch?
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #3800, opened 2017-05-15
Hide aggregation export button

### LINKED ISSUE #3742: Hide aggregation export button
## Expected behavior
Aggregation exports should work, but since getting them to work requires a large effort on the back end, the button should be hidden. Should it be hidden for those who have requested it in the past?

## Current behavior
Aggregation exports silently fail for several task types including text and survey. It's unreliable for others.

jelliotartz:
Fixes #3742.

I didn't want to remove code, and the issue did not say to create a experimental feature flag, so I just added `display:none` styling on the button.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/data-dumps.cjsx
+++ b/app/pages/lab/data-dumps.cjsx
@@ -24,11 +24,22 @@ module.exports = React.createClass
   getInitialState: ->
     avatarError: null
     backgroundError: null
+    showAggregationExportButton: false
+
+  componentDidMount: ->
+    @showAggregationExport @props.project
 
   showWorkflowExport: ->
     Dialog.alert(
       <ExportWorkflowsDialog project={@props.project} />
     )
+
+  showAggregationExport: (project) ->
+    project.get('aggregations_export').then (export) =>
+      if export[0].metadata.state is 'ready'
+        @setState showAggregationExportButton: true
+    .catch (error) =>
+      console.error error
 
   render: ->
     <div className="data-exports">
@@ -67,16 +78,19 @@ module.exports = React.createClass
               buttonKey="projectDetails.workflowContentsExport"
               exportType="workflow_contents_export"  />
           </div>
-           <div className="row" style={"display":"none"}>
-            <DataExportButton
-              project={@props.project}
-              buttonKey="projectDetails.aggregationExport"
-              contentType="application/x-gzip"
-              exportType="aggregations_export"
-              newFeature=true
-            />
-            <small className="form-help">Text tasks and survey tasks cannot be aggregated at this time.</small>
-          </div>
+          {
+           if @state.showAggregationExportButton
+             <div className="row">
+              <DataExportButton
+                project={@props.project}
+                buttonKey="projectDetails.aggregationExport"
+                contentType="application/x-gzip"
+                exportType="aggregations_export"
+                newFeature=true
+              />
+              <small className="form-help">Text tasks and survey tasks cannot be aggregated at this time.</small>
+            </div>
+          }
           <hr />
 
           Talk Data<br />
--- a/app/pages/lab/data-dumps.cjsx
+++ b/app/pages/lab/data-dumps.cjsx
@@ -24,22 +24,11 @@ module.exports = React.createClass
   getInitialState: ->
     avatarError: null
     backgroundError: null
-    showAggregationExportButton: false
-
-  componentDidMount: ->
-    @showAggregationExport @props.project
 
   showWorkflowExport: ->
     Dialog.alert(
       <ExportWorkflowsDialog project={@props.project} />
     )
-
-  showAggregationExport: (project) ->
-    project.get('aggregations_export').then (export) =>
-      if export[0].metadata.state is 'ready'
-        @setState showAggregationExportButton: true
-    .catch (error) =>
-      console.error error
 
   render: ->
     <div className="data-exports">
@@ -78,19 +67,15 @@ module.exports = React.createClass
               buttonKey="projectDetails.workflowContentsExport"
               exportType="workflow_contents_export"  />
           </div>
-          {
-           if @state.showAggregationExportButton
-             <div className="row">
-              <DataExportButton
-                project={@props.project}
-                buttonKey="projectDetails.aggregationExport"
-                contentType="application/x-gzip"
-                exportType="aggregations_export"
-                newFeature=true
-              />
-              <small className="form-help">Text tasks and survey tasks cannot be aggregated at this time.</small>
-            </div>
-          }
+          <div className="row">
+            <DataExportButton
+              project={@props.project}
+              buttonKey="projectDetails.aggregationExport"
+              contentType="application/x-gzip"
+              exportType="aggregations_export"
+              newFeature=true
+            />
+          </div>
           <hr />
 
           Talk Data<br />
--- a/app/pages/lab/data-dumps.cjsx
+++ b/app/pages/lab/data-dumps.cjsx
@@ -67,7 +67,7 @@ module.exports = React.createClass
               buttonKey="projectDetails.workflowContentsExport"
               exportType="workflow_contents_export"  />
           </div>
-           <div className="row">
+           <div className="row" style={"display":"none"}>
             <DataExportButton
               project={@props.project}
               buttonKey="projectDetails.aggregationExport"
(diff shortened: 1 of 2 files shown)


jelliotartz:
Resolves #3742.

marten:
First of all, welcome!

The original ticket #3742 wasn't clear about this, but I think we probably do need to keep the button visible for those projects where an export has been requested previously.

The bio update should possible be its own pull request, so that it could be merged irrespective of your work on hiding the aggregations. Because that commit seems fine to me.

rogerhutchings:
The other thing is not to worry about deleting code - that's what version control is for.

amyrebecca:
i don't understand why we can't just make our minds up whether this should be on or off? it seems like we're introducing unnecessary complexity: if we don't trust aggregation, we should turn it off.

simoneduca:
The fact some projects have requested it before doesn't imply that they want it again or that they got back meaningful data.

camallen:
Let's remove the button and ability to request a new aggregation run if there is no linked resource. 

If a linked resource does exist is leave the button functional (it may not even work in the backend) and provide a link to the existing resource.

astopy:
I think when we discussed this before the idea was that if someone's used it before and got results back then they might be relying on it. Rather than taking away something that people might need, we should only hide it for people who have never used it successfully.

srallen:
Inline styles should be avoided regardless if we hide the button entirely or not. It shouldn't be too involved to add a conditional to show the button conditionally which will give us time to communicate to folks who have used it that the feature is being removed for now. 

In the future, this is just a reminder that experimental features should be behind experimental feature flags. Not doing this makes it harder to remove the feature if we decide that's the best course of action.

app/pages/lab/data-dumps.cjsx
```diff
@@ -24,12 +24,23 @@ module.exports = React.createClass
...
+
+  componentDidMount: ->
+    @showAggregationExport @props.project
 
   showWorkflowExport: ->
     Dialog.alert(
       <ExportWorkflowsDialog project={@props.project} />
     )
 
+  showAggregationExport: (project) ->
```
srallen: This is duplicating the API request that is handled by `<DataExportButton />`. Instead, you'll need to conditionally determine if the API call should be made, hide it if it never has, disable the button for any new exports, but still render the link to download the previous export. Basically, the changes need to happen in the `<DataExportButton />` which is not ideal, but should only be temporary until we communicate to project builders that this feature is going to be disabled at least until we have a better aggregation solution. The help text in this component can also probably be removed.

jelliotartz:
@mcbouslog and I discussed removing the PromiseRenderer component from DataExportButton and rendering this conditional display of the aggregation-export-button via componentDidMount. I decided to push this solution in the meantime, and to perhaps circle back to this issue in the future.

amyrebecca:
Technically we probably should remove the `PromiseRenderer` from the `DataExportButton` if we're gonna be touching it at all. It's a little extra time but we really want to get those out of the project sooner than later, which is why it's in the checklist.

srallen:
Removing `PromiseRenderer` would be a good exercise for working with React's lifecycle methods, so I say go for it. The export request will move out of the props of the `PromiseRenderer` and into `componentDidMount` and `componentWillReceiveProps`.

A later pass on this could be converting it to ES6 and writing tests.

amyrebecca:
@jelliotartz see me if you need any guidance on how to proceed with this

jelliotartz:
thanks @amyrebecca, @mcbouslog pointed me towards a couple helpful resources that I'm looking at for reference, but I'll let you know if I run into blockers :) 

- https://github.com/zooniverse/Panoptes-Front-End/issues/3216

- https://github.com/zooniverse/Panoptes-Front-End/pull/3473/files

eatyourgreens:
If it's useful, here's an old PR where I [removed the PromiseRenderer frrom the Talk data export button](https://github.com/zooniverse/Panoptes-Front-End/pull/3080/files#diff-8e0db9df06ec8099ac71f8d5b0c76183).

app/partials/data-export-button.cjsx
```diff
@@ -14,15 +14,37 @@ module.exports = React.createClass
...
     exportRequested: false
     exportError: null
+    showButton: true
+    mostRecent: null
+
+  componentDidMount: ->
+    @exportGet()
+    @showDataExportBtn()
+
+  componentWillReceiveProps: ->
```
srallen: I don't think defining `componentWillReceiveProps` is necessary since none of the props defined are expected to change while the component is in use (project will only change if the user navigates away to their project list or directly modify the URL anyway at which point the component will mount the first time and call `componentDidMount`)

As an aside, if there is a prop that is expected to change, you typically would want to add a conditional to determine whether or not to call other methods in `componentWillReceiveProps` since it may be a different prop that is changing (let's say the boolean prop in this case if it was changing based on a parent component's state change and if it was written that way). For example (note I'm writing this in coffeescript since this component is still written using it):
```
  if @props.project isnt nextProps.project
    @doSomething()
```
Without the conditional, the method will always be called regardless of which prop is changing.

app/partials/data-export-button.cjsx
```diff
@@ -14,15 +14,37 @@ module.exports = React.createClass
   getInitialState: ->
     exportRequested: false
     exportError: null
+    showButton: true
+    mostRecent: null
+
+  componentDidMount: ->
+    @exportGet()
+    @showDataExportBtn()
```
srallen: This is being called in `@exportGet()` already.

app/partials/data-export-button.cjsx
```diff
@@ -14,15 +14,37 @@ module.exports = React.createClass
...
+
+  componentWillReceiveProps: ->
+    @exportGet()
+    @showDataExportBtn()
 
   exportGet: ->
-    @_exportsGet or= @props.project.get(@props.exportType).catch( -> [])
+    @props.project.get(@props.exportType)
+      .then ([exported]) =>
+        @setState mostRecent: exported
```
srallen: `setState` is asynchronous (explanation in React docs: https://facebook.github.io/react/docs/react-component.html#setstate) so if you're calling a method that depends on the state you've just set, you can't reliably depend on it being set by the time this method is called. Three ways you can deal with this:
- call the method in `setState` as a callback. `setState` can take a callback.
- Pass `exported` in as an argument and then use it on line 39.
- Remove `showDataExportBtn` entirely and just directly add the conditional in the render `if @props.exportType isnt 'aggregations_export' or @state.mostRecent?.metadata?.state?`. When `setState` is called, it causes a re-render.

jelliotartz:
I got the conditional display to work with the callback and passing `exported` as an argument, but I liked the elegance and simplicity of moving the conditional to render, so went with that in the commit.

app/partials/data-export-button.cjsx
```diff
@@ -14,15 +14,22 @@ module.exports = React.createClass
...
+  componentDidMount: ->
+    @exportGet()
 
   exportGet: ->
-    @_exportsGet or= @props.project.get(@props.exportType).catch( -> [])
+    @props.project.get(@props.exportType)
+      .then ([exported]) =>
+        @setState mostRecent: exported
+      .catch (error) =>
+        console.error error
```
srallen: Instead of using `console.error`, can this set `exportError` state?

jelliotartz:
Newbie question: is it ok if I merge this, or is it better practice to have another team member merge the PR?

srallen:
On PFE usually the person who reviews will merge it. We also try to avoid merging on Fridays unless it's for something critical.

srallen:
@jelliotartz could you remove line 2 for the `PromiseRenderer` require now that is removed?

app/partials/data-export-button.cjsx
```diff
@@ -14,15 +13,22 @@ module.exports = React.createClass
...
+  componentDidMount: ->
+    @exportGet()
 
   exportGet: ->
-    @_exportsGet or= @props.project.get(@props.exportType).catch( -> [])
+    @props.project.get(@props.exportType)
+      .then ([exported]) =>
+        @setState mostRecent: exported
+      .catch (error) =>
+        @setState exportError: error
```
srallen: Sorry one last thing, could this only set the error state if the error status is not a 404? We don't need to report 404s to the user.

jelliotartz:
I wasn't sure if I should catch the 404s and display them in the console. I can take that out if it's not what you want.

srallen:
Nope, we don't care about 404s. It just means they've never requested the export before.

PR #3804, opened 2017-05-16
better ETC for finished workflows

camallen:
Fixes https://www.zooniverse.org/talk/17/331857?comment=553695

don't calc and eta for complete workflows, just report 0. 

This looks like i should be extracting the ETA to a class as i've got a props data clump here. Thoughts on best practice for this instead of just 2 functions to create the html. 

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/stats/stats.jsx
+++ b/app/pages/project/stats/stats.jsx
@@ -177,35 +177,12 @@ GraphSelect.propTypes = {
   handleWorkflowChange: React.PropTypes.func,
 };
 
-function calcEta(props) {
-  let eta = undefined;
-  if (props.completeness) {
-    eta = 0;
-  }
-  else if (props.data.length > 1) {
-    let value;
-    let days;
-    if (props.data.length > 15) {
-      // don't count the last bin since the current day is not over yet
-      value = props.data.slice(props.data.length - 15, props.data.length - 1);
-      days = 14;
-    } else {
-      value = props.data.slice(0, props.data.length - 1);
-      days = props.data.length - 1;
-    }
-    const rate = value.reduce((a, b) => (a + b));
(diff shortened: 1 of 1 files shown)


eatyourgreens:
What @srallen said. Calculate the ETA in the parent then pass the result into the ETA component as a prop.

camallen:
updated - i've moved the calculation up to the parent component (thanks @srallen and @eatyourgreens), leaving the functional component as is. OOI should we report 0 or just nothing?

app/pages/project/stats/stats.jsx
```diff
@@ -177,46 +177,20 @@ GraphSelect.propTypes = {
...
-      days = props.data.length - 1;
-    }
-    const rate = value.reduce((a, b) => (a + b));
-    eta = Math.max(0, Math.ceil(days * (props.totalCount - props.currentCount) / rate));
-  }
-  return eta
- }
-
-export const Eta = (props) => {
+export const Eta = (eta) => {
```
eatyourgreens: Small syntax comment. `eta` here should be named `props`.

app/pages/project/stats/stats.jsx
```diff
@@ -177,46 +177,20 @@ GraphSelect.propTypes = {
...
   let output = <div />;
-  const eta = calcEta(props);
   if (eta !== undefined) {
     output = (
       <div>
-        <span className="progress-stats-label">ETC*</span> {`${eta} days`}
+        <span className="progress-stats-label">ETC*</span> {`${eta.numDays} days`}
```
eatyourgreens: `props.numDays` instead of `eta.numDays`.

camallen:
props are indeed passed to components! updated.

PR #3806, opened 2017-05-16
Workflow exports links

### LINKED ISSUE #3693: Make exports by workflow public
We have three launched approved projects now using exports by workflow without any bug reports. It should be made public.

jelliotartz:
Fixes #3693.

Describe your changes.
- Replace select box with radio button list.
- API request to fetch export media.
- Links to previous exports.
- Only export every 24 hours.

https://workflow-exports-links.pfe-preview.zooniverse.org/lab/1663/data-exports

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/export-workflows-dialog.jsx
+++ b/app/pages/lab/export-workflows-dialog.jsx
@@ -43,16 +43,19 @@ class ExportWorkflowsDialog extends React.Component {
       .then((workflows) => {
         this.setState({ workflows, loading: false });
 
-        // TODO: load the exports for each workflow
-        workflows.forEach((wf) => {
-          wf.get('classifications_export')
+        workflows.forEach((workflow) => {
+          workflow.get('classifications_export')
             .then((media) => {
               const mediaState = this.state.media;
-              mediaState[wf.id.toString()] = media[0];
+              mediaState[workflow.id] = media[0];
               this.setState({ media: mediaState });
             })
-            .catch(() => {
-              // this is gonna happen, oh well
+            .catch((error) => {
+              if (error.status === 404) {
+                this.props.onFail(error);
+              } else {
+                console.error(error); // eslint-disable-line no-console
+              }
             });
         });
       });
@@ -102,6 +105,12 @@ class ExportWorkflowsDialog extends React.Component {
   }
 }
 
+ExportWorkflowsDialog.defaultProps = {
+  project: {},
+  onSuccess: () => {},
+  onFail: () => {}
+};
+
 ExportWorkflowsDialog.propTypes = {
   project: React.PropTypes.shape({ links: React.PropTypes.object }).isRequired,
   onSuccess: React.PropTypes.func.isRequired,
@@ -118,8 +127,16 @@ const ExportWorkflowListItem = ({ workflow, media, onChange }) => {
   const titleString = lockout ? 'This item can only be exported every 24 hours' : '';
 
   return (
-    <li>
-      <input type="radio" title={titleString} name="which-workflow" id={`export-${workflow.id}`} disabled={lockout} onChange={onChange} />
+    <li className="workflow-export-list__item">
+      <input
+        type="radio"
+        title={titleString}
+        name="which-workflow"
+        id={`export-${workflow.id}`}
+        className="workflow-export-list__input"
+        disabled={lockout}
+        onChange={onChange}
+      />
       {workflow.display_name}
       <small>
         <ExportWorkflowLink media={myMedia} />
@@ -128,23 +145,44 @@ const ExportWorkflowListItem = ({ workflow, media, onChange }) => {
   );
 };
 
-ExportWorkflowListItem.propTypes = {
-  workflow: React.PropTypes.shape({ id: React.PropTypes.string, display_name: React.PropTypes.string }).isRequired,
-  media: React.PropTypes.shape({}).isRequired,
-  onChange: React.PropTypes.func.isRequired
+ExportWorkflowListItem.defaultProps = {
+  workflow: {
+    id: '',
+    display_name: ''
+  },
+  media: {
+    id: ''
+  },
+  onChange: () => {}
 };
 
+ExportWorkflowListItem.propTypes = {
+  workflow: React.PropTypes.shape({
+    id: React.PropTypes.string,
+    display_name: React.PropTypes.string
+  }).isRequired,
+  media: React.PropTypes.object.isRequired, // eslint-disable-line react/forbid-prop-types
+  onChange: React.PropTypes.func.isRequired
+};
 
 /* eslint-disable multiline-ternary, no-confusing-arrow */
 const ExportWorkflowLink = ({ media }) =>
   media ?
-    <a title={media.updated_at} href={media.src}>{Moment(media.updated_at).fromNow()}</a> :
-    <span>No exports have been requested.</span>;
+    <a
+      title={media.updated_at}
+      href={media.src}
+      className="workflow-export-list__link"
+    >
+      {Moment(media.updated_at).fromNow()}
+    </a> :
+    <span className="workflow-export-list__span">No exports have been requested.</span>;
 /* eslint-enable */
 
-
 ExportWorkflowLink.propTypes = {
-  media: React.PropTypes.shape({ src: React.PropTypes.string, updated_at: React.PropTypes.string })
+  media: React.PropTypes.shape({
+    src: React.PropTypes.string,
+    updated_at: React.PropTypes.string
+  })
 };
 
 ExportWorkflowLink.defaultProps = {
--- a/app/pages/lab/export-workflows-dialog.jsx
+++ b/app/pages/lab/export-workflows-dialog.jsx
@@ -10,7 +10,8 @@ class ExportWorkflowsDialog extends React.Component {
       loading: false,
       workflows: [],
       media: {},
-      selectedWorkflowId: null
+      selectedWorkflowId: null,
+      workflowError: {}
     };
 
     this.requestDataExport = this.requestDataExport.bind(this);
@@ -54,7 +55,9 @@ class ExportWorkflowsDialog extends React.Component {
               if (error.status !== 404) {
                 this.props.onFail(error);
               } else {
-                console.error(error); // eslint-disable-line no-console
+                const workflowErrorState = this.state.workflowError;
+                workflowErrorState[workflow.id] = error[0];
+                this.setState({ workflowError: workflowErrorState });
               }
             });
         });
@@ -79,10 +82,9 @@ class ExportWorkflowsDialog extends React.Component {
           <ul className="workflow-export-list">
             {this.state.workflows.map((result) => {
               const boundHandler = this.setSelectedWorkflowId.bind(this, result.id);
-              return <ExportWorkflowListItem key={result.id} workflow={result} media={this.state.media} onChange={boundHandler} />;
+              return <ExportWorkflowListItem key={result.id} workflow={result} media={this.state.media} onChange={boundHandler} workflowError={this.state.workflowError} />;
             })}
           </ul>
-          { this.props.exportError ? <div className="form-help error">We had a problem requesting your export data: {this.props.exportError.toString()}</div> : null }
         </div>
       );
     }
@@ -118,7 +120,7 @@ ExportWorkflowsDialog.propTypes = {
   onFail: React.PropTypes.func.isRequired
 };
 
-const ExportWorkflowListItem = ({ workflow, media, onChange }) => {
+const ExportWorkflowListItem = ({ workflow, media, onChange, workflowError }) => {
   const myMedia = workflow ? media[workflow.id] : null;
   const now = new Date();
   const lockoutTime = new Date();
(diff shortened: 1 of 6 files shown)


app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -36,17 +42,25 @@ class ExportWorkflowsDialog extends React.Component {
...
+        // TODO: load the exports for each workflow
+        workflows.forEach((wf) => {
+          wf.get('classifications_export')
+            .then((media) => {
+              const mediaState = this.state.media;
+              mediaState[wf.id.toString()] = media[0];
+              this.setState({ media: mediaState });
+            })
+            .catch(() => {
+              // this is gonna happen, oh well
```
srallen: Is there going to be a handler here and any sort of error messaging to the user for failure?

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -36,17 +42,25 @@ class ExportWorkflowsDialog extends React.Component {
...
         this.setState({ workflows, loading: false });
-      });
-  }
 
-  toggleExport() {
-    if (this.workflowList.selectedIndex >= 0) {
-      this.setState({ workflowSelected: true });
-    }
+        // TODO: load the exports for each workflow
+        workflows.forEach((wf) => {
```
srallen: For consistency and clarity, I'd prefer `workflow` instead of `wf`

css/workflow-export-dialog.styl
```diff
@@ -0,0 +1,11 @@
+.workflow-export-list
+  padding-left: 0
+
+  li
```
srallen: Let's actually target these with a class rather than the tag. We've been working toward using a BEM convention for class names.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -90,7 +105,50 @@ class ExportWorkflowsDialog extends React.Component {
...
+      <small>
+        <ExportWorkflowLink media={myMedia} />
+      </small>
+    </li>
+  );
+};
+
+ExportWorkflowListItem.propTypes = {
+  workflow: React.PropTypes.shape({ id: React.PropTypes.string, display_name: React.PropTypes.string }).isRequired,
+  media: React.PropTypes.shape({}).isRequired,
```
srallen: This seems to be missing a shape definition.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -90,7 +105,50 @@ class ExportWorkflowsDialog extends React.Component {
...
+      <input type="radio" title={titleString} name="which-workflow" id={`export-${workflow.id}`} disabled={lockout} onChange={onChange} />
+      {workflow.display_name}
+      <small>
+        <ExportWorkflowLink media={myMedia} />
+      </small>
+    </li>
+  );
+};
+
+ExportWorkflowListItem.propTypes = {
```
srallen: The props that are set as required should have defined defaults.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -90,7 +105,50 @@ class ExportWorkflowsDialog extends React.Component {
 ExportWorkflowsDialog.propTypes = {
```
srallen: The required prop types should have defaults.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -90,7 +105,50 @@ class ExportWorkflowsDialog extends React.Component {
...
   project: React.PropTypes.shape({ links: React.PropTypes.object }).isRequired,
   onSuccess: React.PropTypes.func.isRequired,
-  onFail: React.PropTypes.func.isRequired,
+  onFail: React.PropTypes.func.isRequired
+};
+
+const ExportWorkflowListItem = ({ workflow, media, onChange }) => {
+  const myMedia = workflow ? media[workflow.id.toString()] : null;
```
srallen: The workflow id is already a string. This conversion does not need to happen.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -36,17 +42,25 @@ class ExportWorkflowsDialog extends React.Component {
...
-  toggleExport() {
-    if (this.workflowList.selectedIndex >= 0) {
-      this.setState({ workflowSelected: true });
-    }
+        // TODO: load the exports for each workflow
+        workflows.forEach((wf) => {
+          wf.get('classifications_export')
+            .then((media) => {
+              const mediaState = this.state.media;
+              mediaState[wf.id.toString()] = media[0];
```
srallen: Workflow id is already a string. `toString()` is not necessary.

app/pages/lab/export-workflows-dialog.jsx
amyrebecca: This will fail with a 404 for any workflow that hasn't been exported yet. I'm not crazy about that design decision because it makes us swallow this error. It doesn't affect the user experience at all, which is why we don't do any messaging.

app/pages/lab/export-workflows-dialog.jsx
amyrebecca: We don't use any parts of the shape in this component, because we just pass it on to the child component. The child component validates the props that it cares about, so it would probably be redundant to do it here. The linter rules don't let us simply specify that it's an object and is required, so this is the best solution we could think of.

css/workflow-export-dialog.styl
amyrebecca: I feel weird about adding styles like list-style-type to classes instead of an `li` tag. We did intend to use BEM styles for this component, but it didn't seem necessary because we were able to use styling on the semantic elements. Does BEM really prefer classes to tags?

amyrebecca:
I really dislike making separate files for components like ExportWorkflowListItem that don't make sense outside the context of a parent component--it just seems like it clutters the filesystem for no good reason. But if other folks feel strongly the opposite way I'm not going to be dogmatic about it obviously :)

app/pages/lab/export-workflows-dialog.jsx
srallen: We fixed the error handling in the json-api-client so you can appropriately handle different error codes. So this could be something like `if (error.status !== 404) doSomething`. See example with the classifier: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/classify.cjsx#L281-L288

app/pages/lab/export-workflows-dialog.jsx
srallen: I'd rather have it defined as an object then and add an inline linter rule exception rather than define it as a shape since that just looks like a mistake.

css/workflow-export-dialog.styl
srallen: Yes: http://getbem.com/faq/#custom-tags-for-blocks

srallen:
Separate files for each component is the Airbnb style that I think we should follow, but the rule was disabled because a lot of our old coffeescript files were written that way. IMO, it really does make it hard to follow what component is doing what, especially if you're trying to debug something you didn't write, and bloats individual files. I'd rather have the bloat in the filesystem as it's much more clear.

app/pages/lab/export-workflows-dialog.jsx
amyrebecca: Oh that's handy! Still, even if the error were anything other than 404, I'm not sure we'd want to message the user about it, given that we're not performing the API request in response to a user interaction, just doing it automatically.

app/pages/lab/export-workflows-dialog.jsx
amyrebecca: Works for me, I've done that in the past too.

css/workflow-export-dialog.styl
amyrebecca: Sounds good, I'll do that from here on out.

app/pages/lab/export-workflows-dialog.jsx
srallen: Yes, I see what you mean, so perhaps `if (error.status === 404) return`? FWIW, in a future iteration, all of the export API calls could be refactored to use an include on the project.

amyrebecca: That's a good idea--even though that's what would happen by default anyways, it's a bit more self-documenting.

And you're right, it might also be nice to take a little load off the API to get one request (for the project with its associated export media) instead of five.

app/pages/lab/export-workflows-dialog.jsx
srallen: @jelliotartz It looks like this comment still needs to be addressed.

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -36,17 +42,28 @@ class ExportWorkflowsDialog extends React.Component {
...
-    }
+        workflows.forEach((workflow) => {
+          workflow.get('classifications_export')
+            .then((media) => {
+              const mediaState = this.state.media;
+              mediaState[workflow.id] = media[0];
+              this.setState({ media: mediaState });
+            })
+            .catch((error) => {
+              if (error.status === 404) {
```
srallen: I think this is the reverse of what we were chatting about. The 404 just indicates the project owner hasn't requested this export before, so we don't really care about reporting that back to the user. What I think this should be is:
```
if (error.status !== 404) {
  this.props.onFail.(error);
}
```

srallen:
I found that when the export GET requests were randomly returning a 500 error, that the error message is showing up in the UI on the page rather than in the modal. I think a generic error message should be displayed in the modal for the GET request, maybe inline with the particular workflow export that failed? In the below image, the Journey workflow is the GET request that erred. 

screen shot 2017-05-31 at 9 48 34 am

jelliotartz:
I'm having trouble reproducing the 500 error - I've tried emptying cache and hard reload in Chrome about 10 times, and all I'm seeing in the console are 404 errors from the GET request (and no error message in the UI on the page). Am I missing something? or is there another way I can produce the 500 error besides empty cache & hard reload && wait for random occurrence?

jelliotartz:
I passed `exportError` from `WorkflowExportClassificationButton`'s state to `ExportWorkflowsDialog` as a prop, but its currently displaying `exportError` beneath the list items, rather than inline - I realize this isn't what you're asking for, but I'm at a loss to figure out how to add `exportError` to an individual `ExportWorkflowListItem` (and display it inline, beneath it). 

When I pass `exportError` as a prop to `ExportWorkflowListItem`, it displays it beneath every item. Do I need to find a way to append `exportError` to an individual item? since it starts as part of `WorkflowExportClassificationButton`'s state, it seems like it therefore belongs to all its children if I pass it to them... I could use a little help figuring out what I'm missing to get the inline display with a failed workflow.

srallen:
I think you're on the right track. Instead of trying to reuse the error handler from the parent, you could add a handler in the `ExportWorkflowDialog` component class and a state property. If a particular workflow errors, then update the error state in the catch of the promise for the get request maybe like:
```
let workflowError = {};
workflowError[workflow.id] = error;
this.setState({ workflowError })
```
(I'm using the [object property shorthand syntax](http://eslint.org/docs/rules/object-shorthand) in setState in the above example)

Then pass down that error state as a prop to the list item components and determine which error belongs to which workflow using the id. I'd probably rework those two components into one and instead of using a ternary, have a few inline conditionals (examples here: https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator) for the various rendering cases.

amyrebecca:
Would it be cleaner to simply leave the error where it is and dismiss the modal after the "export" button is clicked? If not, then the stuff you've outlined would be an ideal solution, I just thought I'd suggest something simple in case it worked :)

srallen:
I think for the export POST, yes, that's how the error handling should work, but for the initial GET requests when the modal opens, I think showing inline which workflow GET request failed would be a better UI. In my example, if the request for the Journey workflow fails like it did when I tried it, then I shouldn't be stopped from being able to click to download the export from the Master workflow and should also be able to clearly understand that the Journey workflow link is failing.

jelliotartz:
I'm having trouble testing whether this displays errors inline with the correct workflow. In addition to following @srallen's suggestions, I used the setState of media as a guide, and because this is an object, I'm not getting the ability to update `workflowError` in the React dev tools console like I could when `exportError` was initially set as an empty string. Is there a way I can test whether this is handling and displaying errors correctly, if the prop isn't a string? Or should I be approaching this differently?

Also, happy to refactor `ListItem` and `Dialog` into one component once the error is handled and displayed correctly, just wanted to get that sorted out first :)

app/pages/lab/export-workflows-dialog.jsx
```diff
@@ -36,17 +43,30 @@ class ExportWorkflowsDialog extends React.Component {
...
+            .then((media) => {
+              const mediaState = this.state.media;
+              mediaState[workflow.id] = media[0];
+              this.setState({ media: mediaState });
+            })
+            .catch((error) => {
+              if (error.status !== 404) {
+                this.props.onFail(error);
+              } else {
+                const workflowErrorState = this.state.workflowError;
```
srallen: This is the error handling that should be done instead of `onFail` for the get request for errors that are not 404. This should be moved into the true block of the if statement.

PR #3826, opened 2017-05-24
Account menu keyboard focus

eatyourgreens:
Describe your changes.
Doesn't focus the first menu item when the account menu opens. Instead, enter the menu, from the account button, with the down arrow key.

Also cleans up the account bar javascript and removes `React.CreateClass`.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://account-menu-focus.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/layout/account-bar.jsx
+++ b/app/layout/account-bar.jsx
@@ -6,8 +6,8 @@ import auth from 'panoptes-client/lib/auth';
 import talkClient from 'panoptes-client/lib/talk-client';
 import counterpart from 'counterpart';
 import Translate from 'react-translate-component';
-import Avatar from '../partials/avatar';
 import TriggeredModalForm from 'modal-form/triggered';
+import Avatar from '../partials/avatar';
 import PassContext from '../components/pass-context';
 import NotificationsLink from '../talk/lib/notifications-link';
 
@@ -23,20 +23,20 @@ counterpart.registerTranslations('en', {
     settings: 'Settings',
     signOut: 'Sign Out',
     collections: 'Collections',
-    favorites: 'Favorites',
-  },
+    favorites: 'Favorites'
+  }
 });
 
 const AccountBar = React.createClass({
   contextTypes: {
     user: React.PropTypes.object,
     router: routerShape,
-    geordi: React.PropTypes.object,
+    geordi: React.PropTypes.object
   },
 
   getInitialState() {
     return {
-      unread: false,
+      unread: false
     };
   },
 
@@ -60,10 +60,10 @@ const AccountBar = React.createClass({
     talkClient.type('conversations').get({
       user_id: this.context.user.id,
       unread: true,
-      page_size: 1,
+      page_size: 1
     }).then((conversations) => {
       this.setState({
-        unread: conversations.length > 0,
+        unread: conversations.length > 0
       });
     });
   },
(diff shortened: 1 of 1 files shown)


shaunanoordin:
## PR Notes
* AFAIU the current behaviour is that, if we use keyboard navigation (Tab key to navigate, then Enter key to expand) to open the Account Menu (pictured below), the first sub-item ('Profile') is automatically selected/focused on.
screen shot 2017-05-24 at 15 12 41
* This PR ensures that the first sub-item isn't auto-selected/auto-focused.
  * Side note: Esc key closes the opened sub-menu, as expected.
* I can confirm that this **works** on OSX+Chrome58.
  * OSX+FF53 continues to have tab navigation problems, but this has been raised previously as its own problem.
  * OSX+Safari couldn't be tested, sorry - I still can't log in.
  * Win7+IE11 is an interesting case:
    * The first sub-item isn't auto-selected/auto focused, which is correct.
    * However, Up/Down arrow key navigation (the new navigation capability added in this PR) doesn't work, so I suspect the event listeners may not be firing in IE11.
* Weird observation: the News menu item - which also expands into a sub-menu with sub-items - can't be navigated to, using a keyboard, in any sane manner. In Chrome, you'd have to open the News sub-menu, then open the Account menu, then press the Down key to select the first sub-item in the Account menu, then alt-Tab your way in reverse into the News sub-menu. Logic!
  * However, this is clearly a separate component so shouldn't affect the approval of this PR. It's just the weird interplay of the tab-able components that should be noted.

app/layout/account-bar.jsx
```diff
@@ -22,92 +23,83 @@ counterpart.registerTranslations('en', {
...
+        focusables.push(item);
+      }
+    }
+    const focusIndex = focusables.indexOf(document.activeElement);
 
     const newIndex = {
       [UP]: Math.max(0, focusIndex - 1),
-      [DOWN]: Math.min(focusables.length - 1, focusIndex + 1),
+      [DOWN]: Math.min(focusables.length - 1, focusIndex + 1)
     }[event.which];
```
shaunanoordin: I'm going to check if `Event.which` works properly in IE11.

eatyourgreens: It's supposed to be supported from IE9+ but might be returning a different code for the up/down arrows.

shaunanoordin:
Huh... so apparently this is the IE11 problem?

```
    if (this.accountMenu) {
      try {  //Let's try adding a try...
        for (const item of this.accountMenu.querySelectorAll(FOCUSABLES)) {
          focusables.push(item);
        }
      } catch (err) {  //...and see what IE11 complains about
        console.error(err);
      }
    }
```

And the console error reads...
```
TypeError: Expected _accountMenu$querySel to be iterable, got {
  "0": {
    __reactInternalInstance$p1iksplcm8y6: ReactDOMComponent {
       ...
}
```

eatyourgreens:
Turns out [IE11 does not support `for…of`](https://docs.microsoft.com/en-us/scripting/javascript/reference/for-dot-dot-dot-of-statement-javascript).

shaunanoordin:
Ah, IE11 - the enemy of elegance! That said, I'm happy (sad?) to report that the change to `.forEach` fixes the issue with IE11.

shaunanoordin:
Guess what? This fix has also been confirmed to work properly on Win10+FF53, woot!

So tomorrow I need to figure out what's wonky with Zooniverse's tab navigation in OSX+FF53, and if it's only something weird with my machine or consistent across other Mac users.

eatyourgreens:
No problems for me with OSX+FF 53.0.2. I can also tab through the SVG project graphics on the home page without problems.

Thanks for catching the News menu. All expandable menus should interact with the keyboard in the same way. I'll take a look at that. VoiceOver announce it as "News, collapsed, popup button" so it has the correct roles and state, it just needs the keyboard event handlers.

Issue #3551, opened 2017-03-07
Freehand drawing tools generate too much data


rogerhutchings:
The tools should:

- Debounce updates to the annotation
- Ensure there are no consecutive dupe coords

Fair play to @marten for accurately pointing the finger

rogerhutchings:
We're still getting a lot of data. There are still some options:

- Round down from `n` decimal places
- Smooth out, e.g. by taking every other / 3rd / whatever classification
- Save each point as `[x,y]` instead of a hash, and store schema in classification metadata
- Use compression (thanks @astopy)

cc @camallen

CKrawczyk:
For smoothing the data care should be taken that the Poly-freehand tools don't loose information when used in polygon mode (e.g. make sure the point removed is less than some "delta" from each point on either side).

PR #3827, opened 2017-05-25
Limit accuracy for freehand drawing to 1 decimal place

rogerhutchings:
Contributes to #3551. Uses the `lodash.round` function to round coord data to 1 decimal place. This is then filtered for duplicates as normal.

https://freehand-limit-accuracy.pfe-preview.zooniverse.org/projects/rogerhutchings/etch-a-cell/classify

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/drawing-tools/freehand-helpers.coffee
+++ b/app/classifier/drawing-tools/freehand-helpers.coffee
@@ -1,3 +1,5 @@
+round = require 'lodash/round'
+
 createPathFromCoords = (coordsArray) ->
   [firstCoord, otherCoords...] = coordsArray
   path = "M #{firstCoord.x},#{firstCoord.y} "
@@ -11,6 +13,11 @@ filterDupeCoords = (coordsArray) ->
     filtered
   , []
 
+roundCoords = ({x, y}) ->
+  x: round x, 2
+  y: round y, 2
+
 module.exports =
   createPathFromCoords: createPathFromCoords
   filterDupeCoords: filterDupeCoords
+  roundCoords: roundCoords
--- a/app/classifier/drawing-tools/freehand-helpers.coffee
+++ b/app/classifier/drawing-tools/freehand-helpers.coffee
@@ -13,9 +13,9 @@ filterDupeCoords = (coordsArray) ->
     filtered
   , []
 
-roundCoords = ({x, y}) ->
-  x: round x, 2
-  y: round y, 2
+roundCoords = ({x, y}, precision = 1) ->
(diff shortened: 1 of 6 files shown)


shaunanoordin:
## PR Notes
So far, the changes are looking good. Here's a comparison of the Freehand Shape tool on `master` vs `freehand-limit-accuracy`:

screen shot 2017-05-25 at 14 06 25
`master`: lots of decimal places

screen shot 2017-05-25 at 14 09 02
`freehand-limit-accuracy`: not as many

Tested on OSX+Chrome

Strangely, the decimal points for every x-coordinate is the same for every annotation, and the decimal point for every y-coordinate is 0 for all instances. (On master and otherwise.) This is kinda weird? But not exactly feature-breaking.

Testing will continue with a few other tools and one a few other resolutions. (I noted something weird on the mobile layout, but I need to confirm that.)

shaunanoordin:
## PR Notes
Okay, here's the weird thing. On OSX+Chrome on the mobile layout, the truncation goes to 2 digits.

screen shot 2017-05-25 at 14 18 35
`freehand-limit-accuracy`

It's weird, but it technically fulfils the criteria for making the classification data set smaller.  Perhaps it has something to do with the SVG element itself being squished/stretched on mobile view, altering the dimensions... or something... no, that doesn't make sense either. For comparison, this is the mega string of decimals on master:

screen shot 2017-05-25 at 14 20 37
`master`

Also, ignore this previous comment as I just had a poor sample of data:
> Strangely, the decimal points for every x-coordinate is the same for every annotation, and the decimal point for every y-coordinate is 0 for all instances

shaunanoordin:
@rogerhutchings I can't find any issue with the code nor the functionality - I'm actually happy to approve this PR and get those Classifications downsized ASAP, unless you want us to deep dive into the 2-decimal places mystery.

rogerhutchings:
AH. I know what's happening. I _think_ that everything's actually being rounded to two decimal places. But with the transform at that larger size, it happens to consistently round everything to `x.y0`. And then the last digit is removed, because it adds no precision.

simoneduca:
I haven't looked at the code, but from what @shaunanoordin has observed rounding in this case may not be a good decision. Do we know what are the potential effects of such a big rounding of decimals on a large scale? Will there be any loss of information for the researchers?

rogerhutchings:
No, it's an unnecessary level of precision. I've run it past @hrspiers, who wasn't bothered by it either.

PR #3830, opened 2017-05-26
Add support for another custom Gravity Spy feedback message

srallen:
In anticipation of the announcement next Thursday, this adds support for subjects that have the specified metadata fields to show a custom message. We should get this merged by Tuesday next week at the latest.

CC @rogerhutchings we'll need to consider adding into the generalized feedback feature a way for projects to show a custom feedback message regardless of the annotation. I would have liked Gravity Spy to use the new generalized feedback feature, but I don't think it is designed to handle this use case even if we had survey task support added. 

This test project has a couple of subjects labelled as 'special' that comes up and has a feedback message that says 'congrats': https://gs-special-event-feedback-message.pfe-preview.zooniverse.org/projects/markb-panoptes/gravity-spy-gold-standard-notice-test-project/classify

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/gs-gold-standard-summary.jsx
+++ b/app/classifier/gs-gold-standard-summary.jsx
@@ -26,6 +26,14 @@ const GSGoldStandardSummary = ({ classification, subject, workflow }) => {
     </div>);
   }
 
+  if (subject.metadata['#Label'] === 'special') {
+    return (
+      <div>
+        <p>{subject.metadata['#feedback_1_message']}</p>
+      </div>
+    );
+  }
+
   return (
     <div>
       <p>You responded {choiceLabels.join(', ')}.</p>


rogerhutchings:
Sorry, not sure I understand exactly what circumstances you need a special message for. Any message, success or failure, can be custom set on a subject metadata field with generalised feedback. It doesn't support survey tasks yet, however.

srallen:
They want this message to show up for these subjects regardless of success or failure of the annotation. I might have missed it, but I think the way the feedback works now checks the annotations against the correct answers?

rogerhutchings:
Got it. So actually, maybe this needs to be a feature outside of feedback _anyway_. We could create a metadata field called something like `#post_classification_message`, and add in an additional hook for it in the classification summary component if it happens to exist?

rogerhutchings:
Took me a while to get through to a 'special' subject, but saw the message as expected. LGTM

srallen:
Yeah that would work. I'll open an issue up about it. Thanks!

eatyourgreens:
Is this going to be another piece of UI text that can't be translated, since it's stored in the subject metadata?

srallen:
Yeah, this would have the same issue with translation as the other metadata based content.

PR #3831, opened 2017-05-30
Pan zoom refs

eatyourgreens:
Removes string refs from the PanZoom component.

Moves the `wheel` and `keyDown` event listeners to the component's root element, so that default scrolling behaviour isn't overridden for the window.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://pan-zoom-refs.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -41,9 +41,9 @@ const PanZoom = React.createClass({
   componentDidMount() {
     // these events enable a user to navigate an image using arrows, +, and - keys,
     // while the user is in pan and zoom mode.
-    if (this.props.enabled) {
-      addEventListener('keydown', this.frameKeyPan);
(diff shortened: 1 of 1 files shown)


eatyourgreens:
This might fix the bug in Edge, where scrolling is overridden for the field guide unless it has focus.

shaunanoordin:
screen shot 2017-05-30 at 13 47 03
_The open Field Guide showing two scroll bars: one for the Field Guide content, the other for the main window content._

shaunanoordin:
One quirk with Win7+IE11: there are _three_ scroll bars.

One for the main window content, one for the Field Guide content (the white bit with text) and... one for the whole Field Guide element (the white bit AND the black header that says "Field Text")

This is a non-breaking issue however, just weird, so I'm not too concerned about it.

PR #3845, opened 2017-06-02
Edit a few existing linting rules and add a few more.

simoneduca:
Rules added/edited as warnings:

- `comma-dangle`
- `import/no-unresolved`
- `indent`
- `max-len`
- `no-else-return`
- `no-trailing-spaces`
- `no-unused-vars`
- `no-useless-constructor`
- `no-useless-return`


.eslintrc
```diff
@@ -10,15 +10,20 @@
...
     "arrow-body-style": ["warn", "as-needed"],
-    "comma-dangle": ["warn", "never"],
+    "comma-dangle": ["warn", "always"],
     "import/extensions": "off",
-    "import/no-unresolved": "off",
+    "import/no-unresolved": 1,
+    "indent": ["warn", 2],
     "react/jsx-no-bind": "off",
-    "max-len": "off",
+    "max-len": 1,
```
srallen: If we turn this back on, I'd suggest we add a few options to ignore URLs and comments, similar to the rule setup on the lab app: https://github.com/zooniverse/pfe-lab/blob/master/.eslintrc#L34-L37

.eslintrc
```diff
@@ -10,15 +10,20 @@
     "react/prefer-es6-class": 0,
     "react/prefer-stateless-function": 0,
     "react/no-multi-comp": 0,
-    "array-callback-return": "off", 
+    "array-callback-return": "off",
     "arrow-body-style": ["warn", "as-needed"],
-    "comma-dangle": ["warn", "never"],
+    "comma-dangle": ["warn", "always"],
```
srallen: I don't think we should require trailing comments. Parrish made a good point when we turned this off in #3328 that no other language allows this. I set it as a warning because it had been default enabled with the airbnb ruleset and some folks had been using them. The warning was so folks know to stop using it.

.eslintrc
```diff
@@ -10,15 +10,20 @@
...
     "import/extensions": "off",
-    "import/no-unresolved": "off",
+    "import/no-unresolved": 1,
```
srallen: This was off I think because it was erroring on coffeescript common js requires.

.eslintrc
simoneduca: I'm not too bothered, but I do think [the rule makes it easier to add to multiline lists](https://medium.com/@nikgraf/why-you-should-enforce-dangling-commas-for-multiline-statements-d034c98e36f8). How about turning it on just for multiline?

.eslintrc
simoneduca: Ok cool

.eslintrc
simoneduca: Sure, I like that.

simoneduca: Although, shouldn't it be `"ignoreComments": true`, if you want to ignore comments?

.eslintrc
eatyourgreens: Now that we've got into the habit of removing dangling commas, would it be annoying to ask people to start adding them back in again?

simoneduca: You're right, it would.

eatyourgreens:
I'm getting this error from eslint on this branch.
```ShellSession
Error: /Users/odonnell/zooniverse/Panoptes-Front-End/.eslintrc:
	Configuration for rule "no-unused-vars" is invalid:
	Value "always" no (or more than one) schemas match.
```

eatyourgreens:
More errors 😞  I'm using eslint 3.19
```ShellSession
Error: /Users/odonnell/zooniverse/Panoptes-Front-End/.eslintrc:
	Configuration for rule "no-useless-constructor" is invalid:
	Value "always" has more items than allowed.
```

.eslintrc
```diff
@@ -10,15 +10,20 @@
...
+    "indent": ["warn", 2],
     "react/jsx-no-bind": "off",
-    "max-len": "off",
+    "max-len": 1,
     "multiline-ternary": ["warn", "never"],
-    "no-else-return": "off",
+    "no-else-return": 1,
```
eatyourgreens: I don't see a problem with [using return inside conditionals](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/restart-button.jsx#L4-L12). Is this rule useful?

simoneduca: I turned it on based on [a comment of yours else where](https://github.com/zooniverse/pandora/pull/4/files/42b46a4289376c997e07988e1f449d8749c4774b#r119328415), but happy to switch it off.

eatyourgreens: Oh right, that comment is picking up on a different issue. The `if` block didn't return a value and `return alert()` means you're passing the alert back to the calling function, which doesn't seem right.

I think the linter is already set to warn if it expects a function to return a value and there's no return statement.

simoneduca: OK, cool.

eatyourgreens:
Looks good to me.

PR #3847, opened 2017-06-02
Combo task editor

eatyourgreens:
Fixes #3840 .

Describe your changes.
Checks `workflow.tasks` for task keys that have been deleted, when the ComboEditor mounts.

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch? https://combo-task-editor.pfe-preview.zooniverse.org/
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #3848, opened 2017-06-02
Mutate copies of mock workflow and task data

eatyourgreens:
Otherwise, this will break other tests that import the mock workflow data.

Fixes tests that were broken in #3847


simoneduca:
All tests pass 👍 , LGTM.

PR #3851, opened 2017-06-02
Add featured project tests

simoneduca:
I could not for the life of me find how to test that `img` tag has both non-empty `alt` and `src` tags. Please send help.

### DIFF
--- a/app/pages/home-common/featured-project.jsx
+++ b/app/pages/home-common/featured-project.jsx
@@ -11,21 +11,18 @@ Featured Project Component
 import React from 'react';
 import { Link } from 'react-router';
 
-export default class FeaturedProject extends React.Component {
-  render() {
-    return (
-      <section className="home-featured">
-        <h1>Featured Project</h1>
-        <div className="home-featured-images">
-          <img
-            alt="The direct image (left) and dispersed spectrum (right) of a real galaxy from the WISP survey. The white arrow shows the bright light produced by an emission line."
-            src="./assets/featured-projects/featured-project-20170531-galaxy-nurseries.jpg"
-          />
-        </div>
-        <h2>Introducing the 100th Zooniverse project - Galaxy Nurseries</h2>
-        <p>Help researchers figure out how our universe has changed over time by finding baby galaxies.</p>
-        <Link to="projects/hughdickinson/galaxy-nurseries" className="alternate-button">Get involved!</Link>
-      </section>
-    );
-  }
-}
+const FeaturedProject = () => (
+  <section className="home-featured">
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Have you tried `const {alt, href} = wrapper.find('img').props()`?

app/pages/home-common/featured-project.spec.js
```diff
@@ -0,0 +1,26 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import assert from 'assert';
+import FeaturedProject from './featured-project';
+
+describe.only('FeaturedProject', () => {
```
eatyourgreens: `.only` here means this branch only runs the four tests in this file.

eatyourgreens:
Travis CI is [skipping most of the PFE tests](https://travis-ci.org/zooniverse/Panoptes-Front-End/builds/238799919?utm_source=github_status&utm_medium=notification).

app/pages/home-common/featured-project.spec.js
simoneduca: sorry my bad, that wasn't meant to be there.

simoneduca:
@eatyourgreens I'm doing 
```
const { alt, src } = wrapper.find('img').props();
assert.equal(alt.length, 1);
assert.equal(src.length, 1);
But the test still fails.

**Comment by eatyourgreens** [comment]
You're testing the length of each string there, so maybe test if it's greater than 0.

**Comment by simoneduca** [comment]
Thanks, I thought about that but I don't see any direct way to express that here https://nodejs.org/api/assert.html

**Comment by simoneduca** [comment]
@eatyourgreens that should be ready to go now.

**Comment by eatyourgreens** [comment]
@simoneduca a couple of small comments. Nothing major.

File: app/pages/home-common/featured-project.jsx [hunk 2]
```diff
@@ -11,21 +11,21 @@ Featured Project Component
...
-    );
-  }
+function FeaturedProject() {
+  return (
+    <section className="home-featured">
+      <h1>Featured Project</h1>
+      <div className="home-featured-images">
+        <img
+          alt="The direct image (left) and dispersed spectrum (right) of a real galaxy from the WISP survey. The white arrow shows the bright light produced by an emission line."
+          src="./assets/featued-projects/featured-project-20170531-galaxy-nurseries.jpg"
```
eatyourgreens: Typo: `featued` instead of `featured` in the file path.

simoneduca: Good catch, thanks! Fixed.

PR #3852, opened 2017-06-02
Update Collection Collaborator Settings

### LINKED ISSUE #3745: Update Collections Collaborator settings
## A few additions to the collaborator settings page
- Add the owner's username at the top of the list of collaborators, with "owner" checked and the other boxes disabled
- For other users, the checked option should be in bold
- Add an "are you sure?" modal when unchecking a user's box 

## Additional questions
How is the "Remove" button different from unchecking boxes? If it's not, please remove the button

wgranger:
Fixes #3745 

Describe your changes.
This PR makes some changes to collaborator settings based on suggestions in the issue above. It also writes the components into ES6.

# Review Checklist

- [x] Does it work on mobile?

https://collect-collab-settings.pfe-preview.zooniverse.org/

## Optional


wgranger:
Yes, I can prob swing that today.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+
+    this.handleSubmit = this.handleSubmit.bind(this);
+  }
+
+  handleSubmit(e) {
+    e.preventDefault();
+
+    const checkboxes = this.node.querySelectorAll('[name="role"]');
+    const users = this.userSearch.value().map(option => parseInt(option.value, 10));
+    const checkedBoxes = Object.keys(checkboxes).filter(key => checkboxes[key].checked);
```
srallen: I think it'll be clearer what's happening here if you converted the `checkboxes` NodeList using [`Array.from`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from?v=control) instead of implicitly doing it with `Object.keys()` so you can use `Array.prototype.filter()`.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+    Promise.all((roleSets.map((roleSet) => {
+      return roleSet.save().then(role => role);
+    })))
+      .then((args) => {
+        this.userSearch.clear();
+        Object.keys(checkboxes).map((key) => {
+          checkboxes[key].checked = false;
+        });
+        if (this.props.onAdd) {
+          this.props.onAdd.apply(null, args);
```
srallen: I don't think this is your intention, but `args` here will be the returned value of the above `Promise.all`. Looking at what the prop handler does, I don't think you need to worry about passing arguments.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+      style = {
+        opacity: 0.5,
+        pointerEvents: 'none'
+      };
+    }
+
+    if (this.state.error) {
+      const rawErrorMessage = this.state.error.toString();
+
+      if (rawErrorMessage.indexOf('No User with id')) {
```
srallen: This was written well before we fixed the client to just pass the error object rather than returning a string. I would presume that a user not found would 404, so you can check specifically for that, `if (error.status === 404)` instead of looking for a string of text.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+        {this.state.error && (
+          <p className="form-help error">{errorMessage}</p>
+        )}
+
+        <form style={style}>
+          <div>
+            <UserSearch ref={(component) => { this.userSearch = component; }} />
+          </div>
+
+          <table className="standard-table">
```
srallen: This is non-blocking since I know this was in the original, but a table being used here is not semantic (it's using the table for formatting rather than representing tabular data). I think a definition list would work better and use CSS for formatting.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+            <button className="minor-button" autoFocus={true} onClick={resolve}>No</button>
+            {' '}
+            <button className="major-button" onClick={handleDelete}>Yes</button>
+          </div>
+        </div>
+      );
+    });
+  }
+
+  updateRoles(newRoles = [], callback = () => {}) {
```
srallen: Again, I know it was originally written that way, but I don't think this method needs to support a callback. It's not being used.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+    const { owner } = this.state;
+    const roleSets = this.ownerRoleFirst();
+
+    if (this.state.hasSettingsRole) {
+      return (
+        <div className="collection-settings-tab">
+          {this.state.error && (<p className="form-help error">{this.state.error.toString()}</p>)}
+
+          {owner && (
+            <div>
```
srallen: It seems like this should be an unordered list with the `RoleRow` as list items.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+
+  render() {
+    const { owner } = this.state;
+
+    return (
+      <p>
+        {owner && (<strong>{owner.login}</strong>)}
+
+        { owner.display_name !== this.props.collectionOwner && (
+          <button type="button" className="pill-button" onClick={this.removeRoles}>Remove</button>
```
srallen: You might want to check with @beckyrother, but it doesn't make sense to me to list the owner along with the collaborators and have those checkboxes disabled as I think it would be confusing. This page is almost identical in functionality to the project builder's collaborator page so we may want to think about making them in UI more consistent with each other: 

screen shot 2017-06-12 at 11 30 31 am

Perhaps instead state `You are the collection owner`, a header above the collaborators, and use an `x` for the remove button.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+
+  componentDidMount() {
+    checkIfCollectionOwner(this.props.user, this.props.collection)
+    .then((hasSettingsRole) => {
+      this.setState({ hasSettingsRole });
+    });
+    this.update();
+  }
+
+  update(callback = () => {}) {
```
srallen: Similar to the above comment: I don't think this needs to support a callback. It's not being used.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+    const promise = Promise.all([
+      apiClient.type('collection_roles').get({ collection_id: this.props.collection.id }),
+      this.props.collection.get('owner')
```
srallen: The collection owner is already fetched for in the parent `CollectionPage`. Pass it down as a prop instead of fetching it again.

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+
+          {owner && (
+            <div>
+              {roleSets.map(roleSet =>
+                <RoleRow collectionOwner={owner.display_name} key={roleSet.id} roleSet={roleSet} onRemove={this.handleCollaboratorChange} />
+              )}
+            </div>
+          )}
+
+          <br />
```
srallen: Using a `<br />` seems  a bit odd. Perhaps use CSS instead?

app/collections/collaborators.jsx
```diff
@@ -0,0 +1,375 @@
...
+        roles,
+        links: {
+          collection: this.props.collection.id,
+          user: id
+        }
+      })
+    );
+
+    Promise.all((roleSets.map((roleSet) => {
+      return roleSet.save().then(role => role);
```
wgranger: Ahh. Good catch. I don't think it's necessary. I was initially having issues with this `Promise.all` returning correctly, but I can safely remove the `then()`.

app/collections/collaborators.jsx
wgranger: The callback is being used right before the `setState` in `updateRoles()` to call the `this.props.onRemove` that is passed to it from the `removeRoles()` function.

srallen: Oh I missed that. Why don't we just return the promise instead of using callbacks?

app/collections/collaborators.jsx
beckyrother: @srallen @wgranger yep, that sounds good. Maybe `You are the collection owner` should be more bold so it doesn't get lost up there.
just as long as you, the collection owner, is treated differently to the other collection members and there's no chance of somehow removing yourself, the collection owner, from the collection.

app/collections/collaborators.jsx
wgranger: I might be missing something here. `updateRoles` is called twice in this component, once with a callback and once without, which is why a default no-op is provided. I'm thinking if I return a promise to each call I'll wind up repeating the same `.then`.`.catch`, which would be redundant. Feel free to point out if my understanding is incorrect.

srallen: What I was thinking was that line 222 could instead be:
```
return promise.then(() => {
    this.setState({ saving: false });
  }).catch((error) => {
     this.setState({ error });
  });
```

And the `removeRoles` handler could then be:
```
this.updateRoles([]).then(() => {
  this.props.onRemove();
});
```

wgranger: Ahh. Ok, that makes sense. I was getting a bit confused with all the async action going on.

rogerhutchings:
Doesn't look like there's any conflicts :)

PR #3853, opened 2017-06-02
Restyle Filmstrip and Convert to ES6

### LINKED ISSUE #3829: Update project categories tiles
Just making these tiles match the project cards a bit closer.

## Expected behavior
- Remove rounded corners
- Text label updates: 
font-size: 11px;
font-weight: 700;
text-transform: uppercase;
letter-spacing: 0.1em;

## Here's a photo
projects_ _zooniverse

wgranger:
Fixes #3829 

Describe your changes.
I addressed the styling points on the issue above and also converted the component to ES6, because why not?

# Review Checklist

- [x] Does it work on mobile?

https://fix-3829.pfe-preview.zooniverse.org/

## Optional


app/components/filmstrip.jsx
```diff
@@ -0,0 +1,120 @@
+import React from 'react';
+import { DISCIPLINES } from './disciplines';
+
+export default class Filmstrip extends React.Component {
+  constructor(props) {
```
simoneduca: Just a FYI, passing `props` in the constructor is not necessary https://stackoverflow.com/questions/30571875/whats-the-difference-between-super-and-superprops-in-react-when-using-e

app/components/filmstrip.jsx
```diff
@@ -0,0 +1,120 @@
...
+      list.push('filmstrip--disciplines__discipline-card--active');
+    }
+    if (this.props.value === '' && filterName === 'all') {
+      list.push('filmstrip--disciplines__discipline-card--active');
+    }
+
+    return list.join(' ');
+  }
+
+  adjustPos(increment) {
```
simoneduca: Perhaps a bit picky, but would you mind using names that are a bit more descriptive, e.g. adjustPosition. Same for all the other "pos".

app/components/filmstrip.jsx
```diff
@@ -0,0 +1,120 @@
...
+    this.props.onChange(filterName);
+  }
+
+  calculateClasses(filterName) {
+    const mangledFilterName = this.mangleFilterName(filterName);
+
+    const list = ['filmstrip--disciplines__discipline-card'];
+    list.push(`filmstrip--disciplines__discipline-card--discipline-${mangledFilterName}`);
+
+    if (this.props.value === filterName) {
```
simoneduca: To avoid repetition of `this.props.value`, perhaps have `const { value } = this.props`, after `const mangledFilterName`?

app/components/filmstrip.jsx
```diff
@@ -0,0 +1,120 @@
...
+    );
+  }
+
+}
+
+Filmstrip.propTypes = {
+  filterCards: React.PropTypes.arrayOf(React.PropTypes.object),
+  increment: React.PropTypes.number.isRequired,
+  onChange: React.PropTypes.func.isRequired,
+  value: React.PropTypes.string
```
simoneduca: Should this be required?

app/components/filmstrip.jsx
wgranger: Ahh, nice. Thanks for the clear explanation on that link.

app/components/filmstrip.jsx
wgranger: Yes, this can be `isRequired`. I thought this might not be necessary, since the value is an empty string when all disciplines are shown, from my reading, even an empty string is valid as a required prop.

wgranger:
Thanks for the comments above. I'm also going to add a few unit tests since this component is getting a makeover.

app/components/filmstrip.jsx
```diff
@@ -0,0 +1,120 @@
...
+    const list = ['filmstrip--disciplines__discipline-card'];
+    list.push(`filmstrip--disciplines__discipline-card--discipline-${mangledFilterName}`);
+
+    if (value === filterName) {
+      list.push('filmstrip--disciplines__discipline-card--active');
+    }
+    if (!value && filterName === 'all') {
+      list.push('filmstrip--disciplines__discipline-card--active');
+    }
+    if (value === '' && filterName === 'all') {
```
rogerhutchings: you probably don't need this conditional, as `''` is falsy

PR #3859, opened 2017-06-09
Make feedback optional for success or failure

### LINKED ISSUE #3761: Allow feedback success / failure messages to be optional
Currently all fields are required for feedback, and feedback is always shown where possible. However, there's a requirement for only success / only failure messages to be shown independently.

cc #3725 @lcjohnso

rogerhutchings:
Fixes #3761 

This allows a project builder to selectively enable positive and negative feedback independently. If an annotation meets a feedback rule and showing a success/failure message is enabled, then it works as normal. If not, nothing's returned by the feedback processor.

This also refactors the lab feedback forms to use a common higher order component for common fields, so type-specific ones only need to include those fields relevant to that type.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://make-feedback-optional.pfe-preview.zooniverse.org/

## Optional


package.json
```diff
@@ -119,7 +119,7 @@
     "stage-with-docker": "./bin/run-through-docker.sh 'npm run stage'",
     "_deploy": "export PATH_ROOT=www.zooniverse.org; npm run _build && npm run _publish",
     "deploy": "export NODE_ENV=production; npm run _deploy",
-    "test": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js) --reporter nyan --compilers js:babel-core/register,coffee:coffee-script/register,cjsx:coffee-react/register || true",
+    "test": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name process-feedback-drawing.spec.js) --reporter nyan --compilers js:babel-core/register,coffee:coffee-script/register,cjsx:coffee-react/register || true",
```
srallen: Leftover from local dev?

app/pages/lab/feedback/types/common-form-hoc.jsx
```diff
@@ -0,0 +1,135 @@
...
+    handleInputChange({ target }) {
+      this.handleUpdateState({
+        form: {
+          [target.name]: (target.type === 'checkbox') ? target.checked : target.value,
+        }
+      });
+    }
+
+    handleUpdateState(update) {
+      const newState = merge({}, this.state, update);
```
srallen: Just clarifying what I think is going on here: lodash's merge is being used because a deep merge with state is happening and `setState`'s shallow merge is not sufficient?

app/pages/lab/feedback/types/feedback-input.jsx
```diff
@@ -0,0 +1,61 @@
...
+  if (type === 'checkbox') {
+    propsWrapper.inputProps.checked = value;
+  }
+
+  return (type === 'checkbox')
+    ? <FeedbackInputCheckbox {...propsWrapper} />
+    : <FeedbackInputEverythingElse {...propsWrapper} />;
+};
+
+const FeedbackInputCheckbox = ({ id, title, help, inputProps }) => {
```
srallen: I'm seeing this warning with `FeedbackInputCheckbox`:
```
FeedbackInputCheckbox is changing an uncontrolled input of type checkbox to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components
```

package.json
rogerhutchings: Yup, that's removed.

app/pages/lab/feedback/types/common-form-hoc.jsx
rogerhutchings: That's right

rogerhutchings:
@srallen Everything's done apart from the controlled input warning - I cannot figure that one out for the life of me, as far as I can tell it's always set by a prop...?

app/pages/lab/feedback/types/common-form-hoc.jsx
```diff
@@ -0,0 +1,152 @@
...
+    handleInputChange({ target }) {
+      this.handleUpdateState({
+        form: {
+          [target.name]: (target.type === 'checkbox') ? target.checked : target.value
+        }
+      });
+    }
+
+    handleSave() {
+      this.props.onSubmit(this.state.form, this.props.index)
```
srallen: This line is throwing `Uncaught TypeError: Cannot read property 'catch' of undefined`

app/classifier/feedback/feedback-summary-container.jsx
```diff
@@ -35,24 +35,30 @@ class FeedbackSummaryContainer extends React.Component {
...
+
+  render() {
+    return (this.props.feedback.length) ? <FeedbackSummary feedback={this.props.feedback} /> : null;
+  }
 }
 
 FeedbackSummaryContainer.propTypes = {
-  actions: PropTypes.object,
+  actions: PropTypes.shape({
+    feedback: PropTypes.func
```
srallen: I'm seeing ```Failed prop type: Invalid prop `actions.feedback` of type `object` supplied to `FeedbackSummaryContainer`, expected `function`.```

app/classifier/feedback/feedback-summary-container.jsx
```diff
@@ -35,24 +35,30 @@ class FeedbackSummaryContainer extends React.Component {
     }, []);
     actions.feedback.setFeedback(feedbackItems);
```
srallen: This line is throwing a warning about setState being called during a state transition```setState(...): Cannot update during an existing state transition (such as within `render` or another component's constructor). Render methods should be a pure function of props and state; constructor side-effects are an anti-pattern, but can be moved to `componentWillMount`.``` I'm seeing it both with feedback enabled and disabled.

app/pages/lab/feedback/types/common-form-hoc.jsx
simoneduca: @srallen do you know how to trigger this? I can't reproduce with neither feedback enabled nor disabled.

srallen: It comes up when using the editor form for the point based feedback in the lab.

simoneduca: This is weird because the changes are actually saved and I can see they propagate correctly up to the workflow, which in turn gets updated correctly.

app/classifier/feedback/feedback-summary-container.jsx
simoneduca: You're allowed to call static method in the constructors, but I don't think `generateFeedbackItems()` is static. That's why it's complaining about side-effects. I've moved the call out of the constructor and into `componentWillMount()`, which should have the same behaviour.

app/classifier/feedback/feedback-summary-container.jsx
simoneduca: `feedback` is the object containing all the actions, so `object` is the correct type expected.

app/pages/lab/feedback/types/common-form-hoc.jsx
simoneduca: Fixed with https://github.com/zooniverse/Panoptes-Front-End/pull/3859/commits/6fa8d9fd8f5ddf022307fd0bdddb303489ee9d7d

app/classifier/feedback/feedback-summary-container.jsx
```diff
@@ -35,24 +38,30 @@ class FeedbackSummaryContainer extends React.Component {
...
   }
+
+  render() {
+    return (this.props.feedback.length) ? <FeedbackSummary feedback={this.props.feedback} /> : null;
```
simoneduca: `this.props.feedback.length` always returns `true`; so I think it should be `this.props.feedback.length > 0`.

eatyourgreens: 0 is false in javascript, so aren't those equivalent?

simoneduca: Ah, you're right. I mistakenly thought `this.props.feedback.length` would return `true` but it returns `0`. Thanks.

eatyourgreens: You're right that `this.props.feedback.length > 0` would be more explicit, rather than relying on implicit conversion of a number to a boolean.

simoneduca: Yeah and I just got used to always checking it that way.

simoneduca:
@srallen I'm not seeing this when the feedback is disable. Do you still see it?
> I'm seeing a unique key warning for the FeedbackSummaryContainer when the feedback is disabled. I'm not sure why this is coming up since the render is null. When feedback is enabled, it's specifically the <ul> on line 34 in FeedbackSummary. Maybe item.question won't work as an unique identifier?

simoneduca:
I'm seeing another key error on `classificationSummary`, but I think that's been there before this PR and I haven't been able to fix it.

simoneduca:
Apart from those two comments. Everything should be fixed.

app/classifier/feedback/feedback-summary-container.jsx
srallen: I would prefer explicit conditional statements.

simoneduca: Do you mean `if/then` blocks, instead of ternary op?

srallen: No I mean using `this.props.feedback.length > 0` as @eatyourgreens suggests since that is an explicit statement that doesn't rely on implicit conversion. Javascript can have unintuitive truthy/falsy behavior, so we shouldn't rely on implicit conversion when possible.

simoneduca: Actually I suggested that, so I'm happy to change it.

package.json
```diff
@@ -115,7 +116,9 @@
     "stage-with-docker": "./bin/run-through-docker.sh 'npm run stage'",
     "_deploy": "export PATH_ROOT=www.zooniverse.org; npm run _build && npm run _publish",
     "deploy": "export NODE_ENV=production; npm run _deploy",
-    "test": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js) --reporter nyan --compilers js:babel-core/register,coffee:coffee-script/register,cjsx:coffee-react/register || true",
-    "test-travis": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js) --compilers js:babel-core/register,coffee:coffee-script/register,cjsx:coffee-react/register"
+    "test": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js*) --reporter nyan || true",
+    "test-travis": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js)",
```
eatyourgreens: Should this also be using `*.spec.js*`?

simoneduca: Ah yes, it'd miss out about 11 tests otherwise. Good catch!

simoneduca:
Rebased and installed `eslint-watch`.
Fixed typo in the `test-travis`command.

PR #3861, opened 2017-06-12
Add editorconfig

rogerhutchings:
I was having weird issues with my editor defaulting to 4 spaces on a new machine, and then I realised I hadn't set up my editorconfig properly. And then I realised we didn't have one for PFE.

More info: http://editorconfig.org/


wgranger:
LGTM

PR #3863, opened 2017-06-13
Audio subject

### LINKED ISSUE #3809: Request for audio file (.mp3) subject support.
## Expected behavior
The project builder and/or Python client will accept .mp3 format subjects and the subject viewer will "render" them appropriately using the HTML5 audio player.

## Current behavior
Not implemented.

hughdickinson:
Fixes #3809 .

Implemented audio subjects for users with Admin access using a similar approach to that used for video subjects.

Replicated the tests that are allocated to video subjects.

Example project staged [here](https://audio-subject.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects).

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/components/file-viewer/audio-player.jsx
+++ b/app/components/file-viewer/audio-player.jsx
@@ -7,12 +7,9 @@ class AudioPlayer extends React.Component {
     this.player = null;
 
     this.playAudio = this.playAudio.bind(this);
-    // this.setPlayRate = this.setPlayRate.bind(this);
 
     this.state = {
       playing: false
-      //,
-      // playbackRate: 1
     };
   }
 
@@ -21,12 +18,7 @@ class AudioPlayer extends React.Component {
   }
 
   componentDidUpdate() {
-    //if (this.player) this.player.playbackRate = this.state.playbackRate;
   }
-
-  // setPlayRate(e) {
-  //   this.setState({ playbackRate: parseFloat(e.target.value) });
-  // }
 
   playAudio(playing) {
     const { player } = this;
@@ -41,28 +33,9 @@ class AudioPlayer extends React.Component {
     }
   }
 
-  endVideo() {
+  endAudio() {
     this.setState({ playing: false });
   }
-
-  // renderSpeedControls(rates) {
-  //   return rates.map((rate, i) => {
-  //     return (
-  //       <label key={`rate-${i}`} className="secret-button">
-  //         <input
-  //           type="radio"
-  //           name={`playRate${this.props.frame}`}
-  //           value={rate}
-  //           checked={rate === this.state.playbackRate}
-  //           onChange={this.setPlayRate}
-  //         />
-  //         <span>
-  //           {rate}&times;
-  //         </span>
-  //       </label>
-  //     );
-  //   });
-  // }
 
   render() {
     // const rates = [0.25, 0.5, 1];
@@ -82,15 +55,7 @@ class AudioPlayer extends React.Component {
           Your browser does not support the audio format. Please upgrade your browser.
         </audio>
 
-        {this.props.showControls
-        //   && (
-        //   <span className="subject-video-controls">
-        //     <span className="video-speed">
-        //     Speed: {this.renderSpeedControls(rates)}
-        //     </span>
-        //   </span>
-        // )
-      }
+        {this.props.showControls}
         {this.props.children}
       </div>
     );
--- a/app/components/file-viewer/audio-player.jsx
+++ b/app/components/file-viewer/audio-player.jsx
@@ -38,7 +38,6 @@ class AudioPlayer extends React.Component {
   }
 
   render() {
-    // const rates = [0.25, 0.5, 1];
     return (
       <div className="subject-audio-frame">
         <audio
@@ -55,7 +54,6 @@ class AudioPlayer extends React.Component {
           Your browser does not support the audio format. Please upgrade your browser.
         </audio>
 
-        {this.props.showControls}
         {this.props.children}
       </div>
     );
@@ -74,7 +72,6 @@ AudioPlayer.propTypes = {
 };
 
 AudioPlayer.defaultProps = {
-  showControls: true
 };
 
 export default AudioPlayer;
--- a/app/components/file-viewer/audio-player.jsx
+++ b/app/components/file-viewer/audio-player.jsx
@@ -1,0 +1,115 @@
+import React from 'react';
+
+class AudioPlayer extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.player = null;
+
+    this.playAudio = this.playAudio.bind(this);
+    // this.setPlayRate = this.setPlayRate.bind(this);
+
+    this.state = {
+      playing: false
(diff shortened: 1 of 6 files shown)


app/pages/lab/subject-set.cjsx
```diff
@@ -13,9 +13,9 @@ isAdmin = require '../../lib/is-admin'
 
 NOOP = Function.prototype
 
-VALID_SUBJECT_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif', '.svg']
+VALID_SUBJECT_EXTENSIONS = ['.jpg', '.jpeg', '.png', '.gif', '.svg', '.mp3']
 INVALID_FILENAME_CHARS = ['/', '\\', ':', ',']
-MAX_FILE_SIZE = 600 * 1024
+MAX_FILE_SIZE = 4000 * 1024
```
camallen: Do we really want to bump the file size this much? 4MB seems too big for downloading with preload to auto, especially on mobile devices.

app/components/file-viewer/audio-player.jsx
```diff
@@ -0,0 +1,80 @@
...
+    // const rates = [0.25, 0.5, 1];
+    return (
+      <div className="subject-audio-frame">
+        <audio
+          className="subject"
+          controls={true}
+          ref={(element) => { this.player = element; }}
+          src={this.props.src}
+          type={`${this.props.type}/${this.props.format}`}
+          preload="auto"
```
camallen: This is only a browser hint but should this be auto? Considering asset sizes it may be best to avoid automatic download of the media asset unless the user presses play. 

'metadata' might be a better fit, https://developer.mozilla.org/en/docs/Web/HTML/Element/audio

eatyourgreens: From memory, `autoplay="metadata"` never triggers the `canplay` event in some browsers, so I'd be cautious about changing this. See #2162

app/pages/lab/subject-set.cjsx
eatyourgreens: This does seem excessive since it would correspond to a few minutes of audio playback, I think.

app/components/file-viewer/audio-player.jsx
camallen: Be nice to test this again @hughdickinson - optimal solution is to not download automatically (avoid bandwidth costs on client / server sides) but have the media resource playable ;)

app/components/file-viewer/audio-player.jsx
```diff
@@ -0,0 +1,80 @@
...
+          onCanPlay={this.props.onLoad}
+          onClick={this.playAudio.bind(this, !this.state.playing)}
+          onEnded={this.endAudio}
+        >
+          Your browser does not support the audio format. Please upgrade your browser.
+        </audio>
+
+        {this.props.showControls}
```
eatyourgreens: `showControls` is a boolean. Why is it being rendered here?

app/components/file-viewer/audio-player.jsx
```diff
@@ -0,0 +1,80 @@
...
+  format: React.PropTypes.string,
+  frame: React.PropTypes.number,
+  onLoad: React.PropTypes.func,
+  showControls: React.PropTypes.bool,
+  src: React.PropTypes.string,
+  type: React.PropTypes.string
+};
+
+AudioPlayer.defaultProps = {
+  showControls: true
```
eatyourgreens: I think you can get rid of this. `showControls` is used to render custom controls in the video player, but there are no custom controls in this component.

app/components/file-viewer/audio-player.jsx
```diff
@@ -0,0 +1,80 @@
...
+      player.pause();
+    }
+  }
+
+  endAudio() {
+    this.setState({ playing: false });
+  }
+
+  render() {
+    // const rates = [0.25, 0.5, 1];
```
eatyourgreens: Delete this line?

app/lib/get-subject-location.coffee
```diff
@@ -1,6 +1,7 @@
 READABLE_FORMATS =
   image: ['jpeg', 'png', 'svg+xml', 'gif']
   video: ['mp4']
+  audio: ['mpeg']
```
eatyourgreens: Probably not important, but this doesn't match the MIME type used for the test subject.

app/components/file-viewer/audio-player.jsx
eatyourgreens: Modern browsers are pretty good about conserving bandwidth, particularly on mobile. `preload="auto"` shouldn't download the whole file, just enough to start playing when the play button is pressed, unless `autoplay` is set.

camallen: i couldn't see this documented anywhere when reading this morning, happy to defer to your browser knowledge. 

We should be aware and thinking about this issue as I know i'd be annoyed if my quota was used up downloading large media resources.

eatyourgreens: I believe each browser has its own idea of how to implement the Media API. Fun times. ☹️ 

In Chrome, at least, you can check the GET request for the Arizona Batwatch mp4 files in the network tab and, I think, only the first portion of the file is downloaded.

camallen: in my chrome & firefox i get 3 mp4 media requests, the first two small portions the third retrieves the remaining content all just on page load.

eatyourgreens:
I'm getting an error for the staging project.
`Please try again. Something went wrong: Error: no subject set is associated with this workflow`

hughdickinson:
I've implemented the suggested changes and fixed the subject set on the demo project. Sorry about the latter, I was trying to use the same project to test an audio+image branch I'm working on!

eatyourgreens:
👍 
One small thing: If I favourite an audio file, there's no preview defined for collections, but I don't think that's a blocking issue.

eatyourgreens:
We should probably also think about how audio files are displayed in Talk. https://audio-subject.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/talk/subjects/50246

srallen:
Videos don't have previews either. As we expand on non-image subject types, we'll need to re-think the subject viewer previews in talk and in collections.

eatyourgreens:
Videos render a `video` tag for [Talk previews](https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/230). I think it's defined in the `Thumbnail` component somewhere.

There's also a [larger view for hashtag lists](https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/talk/tags/insect), but I think that uses the standard subject viewer.

srallen:
Ah right, then it's just a problem for collections: https://www.zooniverse.org/projects/zooniverse/arizona-batwatch/collections

eatyourgreens:
Yeah, they might be using CSS background images.

hughdickinson:
Can I just check whether I need to make additional changes. If not will somebody else take care of actually doing the merge?

PR #3878, opened 2017-06-20
[WIP] Audio with image subject

hughdickinson:
Fixes #3863. Now allows an image to be associated with the audio file. The image is assumed to be a visual representation of the sound (i.e. a spectrograph) with time elapsed running in the left to right direction. An SVG marker is overlaid on the image to illustrate the position within the audio track. 

The normal drawing tools can also be used on the image to mark features etc.

Example project and workflow is staged at:

https://audio-with-image-subject.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/classify

The audio-image data are treated as multi-image subjects and forced into non-flipbook mode. The best way to identify or request this behaviour is not clear. Two potential methods are hidden metadata fields in a manifest or a switch in the project builder. All advice is welcome.

Need advice on how best to include audio position in annotations.

# Review Checklist

- [ ] Does it work on mobile?

## Optional


eatyourgreens:
This breaks the tests for `<SVGImage/>` 😞 

The commit history also has the commits to add audio subjects, which I think have already been merged, so I think there might some bad rebase or merge weirdness going on here.

eatyourgreens:
Panoptes supports multiple files for a single frame in the locations array so maybe it would be better to set the subject locations up as one frame with multiple representations
```javascript
locations: [{'audio/mp4': (audio file URL), 'image/jpeg': (image file URL}]
```
instead of multiple frames
```javascript
locations: [{'audio/mp4': (audio file URL)}, {'image/jpeg': (image file URL}]
```
The frame viewer would then need to be updated to handle frames that are represented by multiple files, but that might be better than hacking multi-frame subjects by implying an association between two frames.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -189,7 +191,7 @@ export default class SVGRenderer extends React.Component {
                 />
               </Draggable>
             )}
-
+            <ProgressMarker />
```
eatyourgreens: I'm not sure about adding the progress marker to every subject image. I think that's why the tests are failing.

app/components/file-viewer/audio-player.jsx
```diff
@@ -68,10 +71,14 @@ AudioPlayer.propTypes = {
...
   src: React.PropTypes.string,
-  type: React.PropTypes.string
+  type: React.PropTypes.string,
+  progressListener: React.PropTypes.func,
+  registerProgressObject: React.PropTypes.func
 };
 
 AudioPlayer.defaultProps = {
+  progressListener: null,
+  registerProgressObject: null
```
eatyourgreens: This is declared as a function, but `null` is an object.

app/components/file-viewer/audio-player.jsx
```diff
@@ -68,10 +71,14 @@ AudioPlayer.propTypes = {
...
   showControls: React.PropTypes.bool,
   src: React.PropTypes.string,
-  type: React.PropTypes.string
+  type: React.PropTypes.string,
+  progressListener: React.PropTypes.func,
+  registerProgressObject: React.PropTypes.func
 };
 
 AudioPlayer.defaultProps = {
+  progressListener: null,
```
eatyourgreens: This is declared as a function, but null is an object.

app/components/frame-viewer.jsx
```diff
@@ -120,5 +129,6 @@ FrameViewer.defaultProps = {
   },
   workflow: {
     configuration: {}
-  }
+  },
+  progressListener: null
```
eatyourgreens: Default value should also be a function here.

app/components/progress-updater.jsx
```diff
@@ -0,0 +1,62 @@
...
+    this.progressMonitoredObject = this.registerProgressingObject.bind(this);
+  }
+
+  setSubjectViewer(subjectViewer){
+    this.state.subjectViewer = subjectViewer;
+  }
+
+  updateSubjectViewer(){
+    if(this.state.subjectViewer){
+      this.state.subjectViewer.forceUpdate();
```
eatyourgreens: Calling `forceUpdate` on a component is bad practice. A component should react to external changes via props, so a better pattern here might be a callback to the subject viewer.

app/components/progress-updater.jsx
```diff
@@ -0,0 +1,62 @@
...
+    }
+  }
+
+  handleProgressUpdate(event) {
+    var xPosition = 100 * (this.state.progressObject.currentTime / this.state.progressObject.duration)
```
eatyourgreens: I'm very uneasy about directly accessing the state of an external component here. The internal implementation of the progress object should be opaque, so that changes in implementation don't break components, like this one, that use that component. 

It would be better if the audio player passed down changes, as props, to a child component that rendered the spectrogram.

app/components/progress-updater.jsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
```
eatyourgreens: This isn't a React component, so I'm not sure why it imports React.

app/components/progress-updater.jsx
```diff
@@ -0,0 +1,62 @@
...
+    };
+
+    var progressMarkerStyle = {
+      fill: 'transparent',
+      stroke: 'red',
+      strokeWidth: 4
+    };
+
+    return (
+      <g id='progress_marker' {...progressMarkerStyle}>
```
eatyourgreens: Since this class renders JSX, would it be better to rewrite it as a React component?

app/components/subject-viewer.cjsx
```diff
@@ -101,7 +118,7 @@ module.exports = React.createClass
       mainDisplay = @renderFrame @state.frame
     else
       mainDisplay = @props.subject.locations.map (frame, index) =>
-        @renderFrame index, {key: "frame-#{index}"}
+        @renderFrame index, {key: "frame-#{index}", progressListener: @progressUpdater.progressUpdateListener, progressMarker: @progressUpdater.progressMarkerRenderer, registerProgressObject: @progressUpdater.progressMonitoredObject}
```
eatyourgreens: Looks like you really should just be passing `progressUpdater` as an argument here.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -242,5 +245,6 @@ SVGRenderer.defaultProps = {
   project: null,
   subject: null,
   user: null,
-  workflow: null
+  workflow: null,
+  progressMarker: null
```
eatyourgreens: 'null' used as default value for a function here too.

eatyourgreens:
Re. the failing tests, the same error that fails the tests is also thrown when I try my test drawing project at https://audio-with-image-subject.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify

I believe it's due to `ProgressMarker` being undefined or null when the SVG component tries to render.

eatyourgreens:
The image in the test subject for the audio project also fails to render, with `TypeError: this.props.registerProgressObject is not a function` (Chrome on OSX.)

app/components/progress-updater.jsx
eatyourgreens: Just to expand on that, I was thinking something like `this.props.onProgressChange()` would be better, where the callback is passed down as a prop, and triggers a state change in the parent, which can then be passed down to other children.

app/classifier/annotation-renderer/svg.jsx
eatyourgreens: One thought: you could conditionally push the `ProgressMarker` into the `children` array, when the subject hooks are processed.

li-yp:
Thanks! This is exactly the functionality that I want in the project builder. How can I use it to upload sound clips and spectrograms in the project builder?

hughdickinson:
Refactored component structure to render top-down using new props. Audio player in Safari does not include a position slider which makes tracking through the audio track cumbersome. Be aware of this if you use try to use this functionality.

Subject viewer will attempt to render a subject with an audio and image component if it has exactly two media types with one in image format and the other in audio format.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -19,6 +19,8 @@ export default class SVGRenderer extends React.Component {
   getSizeRect() {
     const clientRect = this.sizeRect && this.sizeRect.getBoundingClientRect();
 
+    console.log('clientRect', clientRect);
```
eatyourgreens: stray console log here

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -36,6 +38,8 @@ export default class SVGRenderer extends React.Component {
     const horizontal = sizeRect ? sizeRect.width / this.props.naturalWidth : 0;
     const vertical = sizeRect ? sizeRect.height / this.props.naturalHeight : 0;
 
+    console.log('scale', { horizontal, vertical });
```
eatyourgreens: console log

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -76,6 +80,7 @@ export default class SVGRenderer extends React.Component {
     newPoint.y = y;
     const matrixForWindowCoordsToSVGUserSpaceCoords = this.getMatrixForWindowCoordsToSVGUserSpaceCoords();
     const pointforSVGSystem = newPoint.matrixTransform(matrixForWindowCoordsToSVGUserSpaceCoords);
+    console.log('pointforSVGSystem', pointforSVGSystem);
```
eatyourgreens: console log

app/classifier/frame-annotator.jsx
```diff
@@ -49,6 +49,7 @@ export default class FrameAnnotator extends React.Component {
 
     if (this.props.annotation.task) {
       taskDescription = this.props.workflow.tasks[this.props.annotation.task];
+      // console.log(taskDescription, tasks);
```
eatyourgreens: delete this comment

app/components/file-viewer/audio-player.jsx
```diff
@@ -9,22 +11,25 @@ class AudioPlayer extends React.Component {
...
+
   }
 
   componentDidMount() {
     this.player.controlsList = "nodownload"; // Non-spec option for Chrome browsers to hide the display of a download button
   }
 
-  componentDidUpdate() {
-  }
+  componentDidUpdate() {}
```
eatyourgreens: empty method

app/components/file-viewer/audio-player.jsx
```diff
@@ -9,22 +11,25 @@ class AudioPlayer extends React.Component {
...
 
   playAudio(playing) {
-    const { player } = this;
-    if (!player) return;
+    const {player} = this;
+    if (!player)
+      return;
```
eatyourgreens: what is this returning here?

app/components/file-viewer/audio-player.jsx
```diff
@@ -34,27 +39,87 @@ class AudioPlayer extends React.Component {
...
+    this.setState({progressPosition: this.player.currentTime});
+  }
+
+  onAudioLoad(e) {
+    e.stopPropagation();
+    this.state.trackDuration = this.player.duration;
+  }
+
+  renderProgressMarker() {
+    // console.log(this.state.progressPosition, this.state.trackDuration);
```
eatyourgreens: delete this comment

app/components/file-viewer/image-viewer.jsx
```diff
@@ -8,6 +8,7 @@ class ImageViewer extends React.Component {
     this.state = {
       loading: true
     };
+    // console.log(props.src);
```
eatyourgreens: delete this comment

app/components/file-viewer/index.jsx
```diff
@@ -23,8 +23,19 @@ const VIEWERS = {
...
+    }
+    // ... add outher here if neccessary
+  }
+  return VIEWERS[props.type] || DefaultViewer;
+
+}
+
 function FileViewer(props) {
-  const Viewer = VIEWERS[props.type] || DefaultViewer;
+  const Viewer = subjectViewerSelector(props);//VIEWERS[props.type] || DefaultViewer;
```
eatyourgreens: delete the comment here?

app/components/file-viewer/progress-indicator.jsx
```diff
@@ -0,0 +1,65 @@
...
+      y2: '100%'
+    };
+
+    const progressMarkerStyle = {
+      fill: 'transparent',
+      stroke: 'red',
+      strokeWidth: 2
+    };
+
+    // console.log(points);
```
eatyourgreens: delete this comment

app/components/frame-viewer.jsx
```diff
@@ -20,6 +22,8 @@ export default class FrameViewer extends React.Component {
     const width = e.target.videoWidth || e.target.naturalWidth || 100;
     const height = e.target.videoHeight || e.target.naturalHeight || 100;
 
+    console.log('handleLoad called', width, height, e, e.target);
```
eatyourgreens: stray console log

app/components/subject-viewer.cjsx
```diff
@@ -7,19 +7,31 @@ getSubjectLocation = require '../lib/get-subject-location'
 FlagSubjectButton = require './flag-subject-button'
 SignInPrompt = require '../partials/sign-in-prompt'
 `import FrameViewer from './frame-viewer';`
-`import CollectionsManagerIcon from '../collections/manager-icon';`
```
eatyourgreens: Is this supposed to be deleted?

eatyourgreens: The staging link gives this error `Uncaught (in promise) ReferenceError: CollectionsManagerIcon is not defined`.

app/lib/get-subject-locations.js
```diff
@@ -0,0 +1,24 @@
+const READABLE_FORMATS = {
+  image: [
+    'jpeg', 'png', 'svg+xml', 'gif'
+  ],
+  video: ['mp4'],
+  audio: ['mp3', 'm4a', 'mpeg']
```
eatyourgreens: should 'mpeg' be audio or video? or both?

PR #3883, opened 2017-06-23
Hackday - rewrite Admin as a module using Grommet

rogerhutchings:
For my hackday project, I had a stab at converting the Admin area into a more separate module, as per zooniverse/rfc#15, as it's the smallest logical part I thought I could do in the time available. The idea was to create something that would be trivial to eventually put into its own folder / monorepo folder / package / whatever.

- Created a new module which exports its routes
- Rewrote Coffeescript into ES6
- Used Grommet components for layout, controls
- Separated up files into containers / components

### What I'd like from a review

- Do you think this is on the right track?
- Is this worth finishing?

### Caveats

- It's not finished. I ran out of time around the `project-status` section.
- It's a bit rough and ready. I haven't totally got the hang on Grommet, and I think there are some unwanted styles interfering a bit (ticks in checkboxes not centred, for one thing).
- It's not totally standalone. It still requires some components in the main `/lib` and `/components` folders, although these would probably be extracted into a common components package.
- It's not using redux for state management. But it totally could.

https://hack-admin.pfe-preview.zooniverse.org/

### Still to do

- Finish separating containers and components up
- Finish styling
- Remove old `admin.styl` file

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://hack-admin.pfe-preview.zooniverse.org/

## Optional


simensta:
@rogerhutchings, I think the admin section is a great place to start incorporating zoo-grommet. 

What do you think about splitting these changes into two PRs, one for incorporating grommet and another for the more structural changes (i.e., Coffeescript to ES6, module which exports its routes, containers / components separation)?

rogerhutchings:
Sure, I can do that. I'll ping you when I have something...

PR #3888, opened 2017-06-23
Make Cross-browser fixes and general adjustments

### LINKED ISSUE #3886: Additional home page updates
### Please notify me before merging these changes

---

1) Background of video player should be #007482 in case of awkward screen sizes 
video

2) Please look into the issue with commas not showing up in Safari on the number of classifications 
commas

3) Border of project cards should be #CBCCCB 

4) Remove the "listen to the zooniverse" link in the paragraph under number of classifications

5) Remove the full stop at the end of "Meet the researchers" headline

6) Add a line between the first and second Daily Zoo post, 30px under the white box and 30px above the next item

7) Change "Newest project" to "Latest project" in the social section

8) Change "Recent project updates" to "Project updates"

wgranger:
Fixes #3886 

Describe your changes.
More landing page adjustments based on the issue. This also fixes an issue that is currently visible on mobile view where the project cards don't switch to column view correctly.

# Review Checklist

- [x] Does it work on mobile?

https://homepage-updates.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/home-not-logged-in/research.jsx
+++ b/app/pages/home-not-logged-in/research.jsx
@@ -16,10 +16,6 @@ counterpart.registerTranslations('en', {
     `,
     classifications: 'Classifications so far by',
     labs: 'Zooniverse Labs',
-    listen: `
-      The projects created with the Zooniverse project builder
-      inspire and empower researchers, designers, and artists.
-    `,
     meetResearchers: 'Meet the researchers who\'ve created projects for free on the Zooniverse',
     options: 'Sign in or register to get started',
     real: 'Real researchers, real results',
@@ -43,7 +39,7 @@ counterpart.registerTranslations('en', {
 const HomePageResearch = (({ count, screenWidth, showDialog, volunteerCount }) =>
   <section className="home-research">
     <Translate className="tertiary-kicker" component="h2" content="researchHomePage.works" />
-    <span className="class-counter">{count.toLocaleString()}</span>
+    <span className="class-counter">{count.toString().replace(/.(?=(?:.{3})+$)/g, '$&,')}</span>
     <Translate className="main-kicker" component="h3" content="researchHomePage.classifications" />
     <div>
       <h3 className="main-kicker">{volunteerCount.toLocaleString()}</h3>{' '}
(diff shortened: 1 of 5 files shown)


app/pages/home-not-logged-in/research.jsx
```diff
@@ -43,7 +39,7 @@ counterpart.registerTranslations('en', {
 const HomePageResearch = (({ count, screenWidth, showDialog, volunteerCount }) =>
   <section className="home-research">
     <Translate className="tertiary-kicker" component="h2" content="researchHomePage.works" />
-    <span className="class-counter">{count.toLocaleString()}</span>
+    <span className="class-counter">{count.toString().replace(/.(?=(?:.{3})+$)/g, '$&,')}</span>
```
wgranger: I was unable to recreate this issue on Safari (browsers not inserting commas in large numbers with `.toLocaleString()`, but this regex should work just the same.

css/home-page.styl
```diff
@@ -439,6 +440,10 @@ $flex-section
           &:nth-child(odd)
             justify-content: flex-end
 
+        @media (max-width: 400px)
+          flex-basis: 100%
+          flex-direction: column
+
```
wgranger: This extended media query was necessary as `flex-basis` is used primarily in the above media queries to keep project cards breaking two at a time instead of stranding a sole card on a line alone. Mobile Safari wasn't breaking into column with the small viewport.

app/pages/home-not-logged-in/research.jsx
srallen: Some languages use periods rather than commas for the thousands separator and `toLocaleString` will properly format that. I think it'd be best to stick with that.

wgranger: Yeah, originally this issue came up because it was reported `.toLocaleString()` wasn't working on a Safari browser, and I'm suspecting it was an older version of the browser. The `.toLocaleString()` function appears to work on current Safari. 

This brings up a question about browser support. Should we make concessions to support older versions of a browser or expect browsers to be up to date?

css/home-page.styl
srallen: Maybe CSS multi-column spec would work better than flex in this case? https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Columns/Using_multi-column_layouts

app/pages/home-not-logged-in/research.jsx
srallen: Our policy generally is we support current version and the most recent previous version only. MDN reports that Safari 10 should support it. If the issue is being seen in Safari 9, we could look into polyfilling it. Looks like the service we use will do it, if we specify: https://github.com/Financial-Times/polyfill-service/issues/359

css/home-page.styl
rogerhutchings: Or [Grommet Tiles](https://grommet.github.io/docs/tiles)?

mcbouslog:
code and [the staged branch](https://homepage-updates.pfe-preview.zooniverse.org/) LGTM. 
@beckyrother - thoughts on final review?

beckyrother:
Looks good! Thanks for checking.

PR #3894, opened 2017-06-26
Add mobile section to project builder - the do-over

rogerhutchings:
Replaces #3749, as it saves having to ask Robin to make changes to her fork!

Adds the mobile section into the project builder, addressing the items @wgranger picked up on in his review of the original PR. I've also rewritten the test suite to be more BDD-like, and remove a lot of the code duplication.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://robin-add-mobile.pfe-preview.zooniverse.org/

## Optional


app/pages/lab/mobile-section.jsx
```diff
@@ -0,0 +1,178 @@
...
+      <div>
+        <span className="form-label">Mobile App</span>
+        <br />
+        <div className="workflow-mobile-form">
+          <div className="workflow-mobile-form-left-panel">
+            <label
+              className="pill-button"
+              htmlFor="swipe_enabled"
+              title={mobileHelp}
+              style={(swipeEligible && launchApproved) ? null : disabledStyle}
```
wgranger: Oh nice. My thought was that this would have to be a change on the mobile repo, but this seems to work well also. In mobile, I'll prob also add `launch_approved` to the api query once we have enough fixes/patches to warrant an update.

app/pages/lab/mobile-section.jsx
```diff
@@ -0,0 +1,178 @@
...
+            >
+              <AutoSave resource={this.props.workflow}>
+                <input
+                  id="swipe_enabled"
+                  type="checkbox"
+                  disabled={!swipeEligible || !launchApproved}
```
wgranger: Perhaps this is being a little paranoid, but I can change the props from dev tools to enable this input. An added check on the backend (that `launch_enabled: true`) and in mobile would be an added safeguard against this condition.

wgranger:
Thanks for opening this back up with a few changes. I think only a couple changes are needed to functionality to merge this. I will open an issue on the mobile repo to make a change for only `launch_approved` projects, and that should add further protection from any project sneaking in. 

Also, there was an edge case of a workflow changing back and forth between being swipe eligible. For example, what if a workflow is `swipe_enabled: true` (two questions, one task, etc.), but then an extra task is added? The workflow would need to change back to `swipe_enabled: false`. I'm thinking this might be easy to do with a lifecycle check. Maybe when new props are received, a check is made for swipe eligibility, and if it changes, then `swipe_enabled` will switch back to `false`.

rogerhutchings:
@wgranger So I had to rewrite the component, since everything was being done in the render. This should now work as described :)

srallen:
@rogerhutchings I'm not sure if I'm gonna have time to get to this soon, so I'm gonna remove myself and maybe @wgranger can do the full review?

wgranger:
@srallen Yes, I should be able to get to this in a couple days.

wgranger:
Wasn't there originally a validation in this PR to check that a project was `launch_approved` before allowing a workflow to be tagged as `swipe_enabled`? If not, I think it should be included.

wgranger:
Overall, this looks good. Once the change above is made I will test it out a bit more to make sure I can't add some random, non-approved project to the mobile app. If all looks good there, I can approve the PR and we can merge Monday.

simoneduca:
Waiting to merge until referenced issue is closed.

wgranger:
@simoneduca Note the most recent commit pushed to reflect the changed location of the `swipe_enabled` tag.

simoneduca:
That looks good, @wgranger. But looking at the Panoptes issue, we're still waiting on the backend, right?

wgranger:
Yes, we are. Just wanted to give the heads up on changes to this branch. Also, there is a PR on the mobile repo reflecting the same change. https://github.com/zooniverse/mobile/pull/132

wgranger:
Let's try to get this merged. 

There was initially a worry of DOM hacking and a desire to only allow workflows to set `swipe_enabled: true` in the backend if their project was `launch_approved`. Instead of making any backend changes, we are going to adjust the mobile code to only return `launch_approved` projects. 

We're getting that change made on the mobile repo today. Then I can rebase and get this merged once the app store accepts the change. @simoneduca

wgranger:
Ok. I think this should be all set. I have put the `swipe_enabled` tag back in the workflow configuration blob, and I had to update a test to pass with Enzyme v3. 

I think everything should be all set, and we are going to deploy the updated mobile code to the app store and Google Play.

mcbouslog:
really sorry to further complicate, but this'll have to be updated to also toggle `workflow.mobile_friendly`. I can attempt a PR into this branch unless someone would prefer otherwise.

per https://github.com/zooniverse/mobile/issues/151

wgranger:
@mcbouslog I just handled the `mobile_friendly` toggle on the workflow resource, so this should be good to go.

simoneduca:
@wgranger nothing urgent, but something I've just noticed I missed before merging: with the new translations we don't have the text in the component but we extract it a locale file in the `locales` folder. Would you mind taking care of that? No worries if you've got too much on already.

eatyourgreens:
We aren’t translating the Project Builder so don’t worry about using counterpart here.

simoneduca:
Good point, cheers 👍

PR #3907, opened 2017-06-30
Reveal Project Social Icons in Lab

### LINKED ISSUE #489: social media links on the project front page with a form to fill in on the project builder
low priority - if science teams are going to be encouraged to communicate their project's happenings and results with social media it would be nice if you could enter those links in the project builder and the twitter icon, g+ icon, and facebook icon would appear on the front page of the site. I think explicitly having those text boxes to fill in in the project builder would encourage project teams to set up such accounts.

wgranger:
Fixes #489 

Describe your changes.
This adds a social links editor to the project lab page. More info. can be found in the issue.

This proved to be a bit trickier than I first suspected, since a projects external links and social links are both contained in the same `project.urls` array, and I had to find a way to identify, edit, and separate a project's social links from regular external links. 

# Review Checklist

- [x] Does it work on mobile?

https://reveal-social-icons.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/lab/external-links-editor.jsx
+++ b/app/pages/lab/external-links-editor.jsx
@@ -95,6 +95,9 @@ export default class ExternalLinksEditor extends React.Component {
         link._key = Math.random();
       }
     }
+    const filterSocial = tableUrls.filter((url) => {
+      return !url.path;
+    });
 
     return (
       <table className="external-links-table">
@@ -106,11 +109,11 @@ export default class ExternalLinksEditor extends React.Component {
         </thead>
         <DragReorderable
           tag="tbody"
-          items={tableUrls}
+          items={filterSocial}
           render={this.renderRow}
           onChange={this.handleLinkReorder}
         />
-      </table>      
+      </table>
     );
   }
 
--- a/app/pages/lab/external-links-editor.jsx
+++ b/app/pages/lab/external-links-editor.jsx
@@ -1,7 +1,7 @@
 import React from 'react';
+import DragReorderable from 'drag-reorderable';
 import AutoSave from '../../components/auto-save.coffee';
 import handleInputChange from '../../lib/handle-input-change.coffee';
-import DragReorderable from 'drag-reorderable';
 
 export default class ExternalLinksEditor extends React.Component {
   constructor(props) {
@@ -17,8 +17,8 @@ export default class ExternalLinksEditor extends React.Component {
     const changes = {
       [`urls.${this.props.project.urls.length}`]: {
         label: 'Example',
-        url: 'https://example.com/',
-      },
+        url: 'https://example.com/'
+      }
     };
     this.props.project.update(changes);
   }
@@ -37,7 +37,7 @@ export default class ExternalLinksEditor extends React.Component {
     if (indexToRemove > -1) {
       urlList.splice(indexToRemove, 1);
       const changes = {
-        urls: urlList,
+        urls: urlList
       };
       this.props.project.update(changes);
       // Remove link is handeld outside of the AutoSave so save directly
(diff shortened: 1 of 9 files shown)


test/setup.js
```diff
@@ -13,3 +13,4 @@ Object.keys(document.defaultView).forEach((property) => {
 global.navigator = {
   userAgent: 'node.js'
 };
+global.document.documentElement.dataset = {};
```
wgranger: This was added because the `DragReordable` component will error when Enzyme uses `mount` instead of shallow, because the component searches for `document.documentElement.dataset`.

mschwamb:
I wanted to note that if you have a external link or two then you pretty much can only add two social media icons before the navigation bar goes onto two pages, maybe there doesn't need to be that much space between the social media icons
Example from the P4:T development version on the preview staging area 

screen shot 2017-07-01 at 5 06 02 pm

wgranger:
@beckyrother 

I changed icon spacing, so check the icons on [this staged branch](https://reveal-social-icons.pfe-preview.zooniverse.org/) when you get a chance.

beckyrother:
Looks good, thanks for checking @wgranger!

app/pages/lab/social-links-editor.jsx
```diff
@@ -0,0 +1,147 @@
...
+          url: `https://${site}${e.target.value}`
+        }
+      };
+      this.props.project.update(changes);
+    } else {
+      this.handleRemoveLink(site);
+    }
+  }
+
+  handleLinkReorder(newLinkOrder) {
```
mcbouslog: I liked how you used `.sort` in a different file, that I coincidentally reviewed before this one. I wasn't familiar with it, had the related docs up and it felt like could apply here. After some help from Google and a little messing around I'm not sure if the following will cause any issue (doesn't appear to, could use confirmation), but the following might work(?):
```
handleLinkReorder(newLinkOrder) {
    const externalUrls = this.props.project.urls.filter(url => !url.path);
    const socialUrls = this.props.project.urls.filter(url => url.path);
    const newSocialUrls = socialUrls.sort((a, b) => newLinkOrder.indexOf(a.site) - newLinkOrder.indexOf(b.site));
    const changes = { urls: externalUrls.concat(newSocialUrls) };
    this.props.project.update(changes).save();
    this.setState({ socialOrder: newLinkOrder });
  }
```
If think helps please feel free to rename variables.

app/pages/lab/external-links-editor.jsx
```diff
@@ -95,6 +95,9 @@ export default class ExternalLinksEditor extends React.Component {
         link._key = Math.random();
       }
     }
+    const filterSocial = tableUrls.filter((url) => {
```
wgranger: Good catch.

app/pages/lab/social-links-editor.jsx
```diff
@@ -0,0 +1,147 @@
...
+
+  handleDisableDrag(event) {
+    event.target.parentElement.parentElement.setAttribute('draggable', false);
+  }
+
+  handleEnableDrag(event) {
+    event.target.parentElement.parentElement.setAttribute('draggable', true);
+  }
+
+  indexFinder(toSearch, toFind) { //eslint-disable-line
```
wgranger: You're right about why I included this disabled line, but it's not consistent that I have this disabled line in only one of the methods. Probably a good idea to remove the comment. This is probably worth bringing up in a channel (how to handle these static methods, or removing the linter rule).

app/pages/lab/social-links-editor.jsx
wgranger: That's a great solution. Thanks a ton for the suggestion. This was one part of the code that I wracked my brain about, so it's good that it can be refactored into the few lines above.

Issue #3904, opened 2017-06-29
workflows request and page_size issue


mcbouslog:
Notes from Nature now has over 100 workflows (or expeditions). Eventually, NfN will become an organization, with workflows/expeditions spread across many projects, but until then the over 100 workflows need to be listed or accessed in various ways in various components, but [the current default API query limits the `page_size` response to 100](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/get-workflows-in-order.coffee#L6):

- [x] 1. [lab/workflows](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/workflows-container.jsx#L44) - create dynamic paginator in #3908, **but will need revisit if max = 20**
- [x] 2. lab/visibility - add current paginator - Done in #3913 
- [ ] 3. [project/home](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/home/index.jsx#L87) (volunteer selects from all active workflows)
- [x] 4. [project/stats](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/stats/index.jsx#L43) - addressed in #3955, **but will need revisit if max = 20**
- [ ] 5. [admin/project](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/admin/project-status.jsx#L60) - add current paginator(?)
- [ ] 6. [lab/data-exports](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/export-workflows-dialog.jsx#L44) - by workflow, add current paginator(?)

Related
- #2820
- #3144
- #3903

srallen:
@mcbouslog Thanks for opening this. The two most relevant for NfN are probably 1 and 2, so they should be fixed first. For 3, since they have their own landing, I imagine the redirect would send volunteers back to that? I think 3 will need some design input from @beckyrother because a paginator might be awkward, but I don't think listing out 100+ workflows on the page is good UI. 

With the project stats, we probably should fetch the workflows that have been set to be displayed which I think is in the project config. For 5, just normal pagination should work for us and is the lowest priority since we're the only ones to use that.

beckyrother:
I think a paginator would be ok, or if you do list out all 100+ we should add Search and Sort functionality, i.e. sort by date added, name, etc.

eatyourgreens:
[...]

PR #3908, opened 2017-06-30
Reworked workflows lab page to toggle between reorder and list views

srallen:
For #3904.

I've add a button to toggle between a default list view with pagination and a view that uses `DragReorderable`. I've updated the tests to reflect the UI changes. This needs to be checked to see if it works for Notes from Nature.

https://reworked-workflows-ui.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/workflows-container.jsx
+++ b/app/pages/lab/workflows-container.jsx
@@ -6,6 +6,7 @@ export default class WorkflowsContainer extends React.Component {
     super(props);
     this.state = {
       loading: true,
+      reorder: false,
       workflowCreationInProgress: false,
       workflows: []
     };
@@ -15,17 +16,35 @@ export default class WorkflowsContainer extends React.Component {
     this.hideCreateWorkflow = this.hideCreateWorkflow.bind(this);
     this.showCreateWorkflow = this.showCreateWorkflow.bind(this);
     this.handleWorkflowReorder = this.handleWorkflowReorder.bind(this);
+    this.toggleReorder = this.toggleReorder.bind(this);
+    this.onPageChange = this.onPageChange.bind(this);
   }
 
   componentDidMount() {
     this.getWorkflowList();
   }
 
-  getWorkflowList() {
-    getWorkflowsInOrder(this.props.project, { fields: 'display_name' })
-    .then((workflows) => {
-      this.setState({ workflows, loading: false });
+  onPageChange(page) {
+    const nextQuery = Object.assign({}, this.props.location.query, { page });
+    this.context.router.push({
+      pathname: this.props.location.pathname,
+      query: nextQuery
     });
+    this.getWorkflowList(page);
+  }
+
+  getWorkflowList(page = 1) {
+    if (this.state.reorder) {
+      getWorkflowsInOrder(this.props.project, { fields: 'display_name', page_size: this.props.project.links.workflows.length })
+      .then((workflows) => {
+        this.setState({ workflows, loading: false });
+      });
+    } else {
+      this.props.project.get('workflows', { page })
+        .then((workflows) => {
+          this.setState({ workflows, loading: false });
+        });
+    }
   }
 
   labPath(postFix = '') {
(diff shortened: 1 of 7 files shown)


app/pages/lab/workflows-container.jsx
```diff
@@ -60,13 +79,18 @@ export default class WorkflowsContainer extends React.Component {
     this.getWorkflowList();
   }
 
+  toggleReorder() {
+    this.setState((prevState) => { return { reorder: !prevState.reorder }; });
+  }
```
mcbouslog: should this call `getWorkflowList`? 

in NfN's case it works as expected initially, with 20 (of 100) workflows listed with pagination at the bottom, but once clicking the Reorder View button should it show all 100 workflows within the DragReorderable component?

app/pages/lab/workflows-container.jsx
```diff
@@ -15,17 +16,35 @@ export default class WorkflowsContainer extends React.Component {
...
+  getWorkflowList(page = 1) {
+    if (this.state.reorder) {
+      getWorkflowsInOrder(this.props.project, { fields: 'display_name', page_size: this.props.project.links.workflows.length })
+      .then((workflows) => {
+        this.setState({ workflows, loading: false });
+      });
+    } else {
+      this.props.project.get('workflows', { page })
+        .then((workflows) => {
+          this.setState({ workflows, loading: false });
```
mcbouslog: clicking Next doesn't advance to page 2 like I'd expect in https://reworked-workflows-ui.pfe-preview.zooniverse.org/lab/1613/workflows. I've been trying to figure out why without luck, the url is properly updated, maybe a `componentWillReceiveProps` that compares `location.query.page`?

app/pages/lab/workflows-container.jsx
srallen: Yep. Just need to add it to the callback of setState

app/pages/lab/workflows-container.jsx
srallen: I forgot to pass `onPageChange` down as a prop, so the pagination buttons don't have a handler that does anything at the moment.

srallen:
@mcbouslog the two things you pointed out are fixed now. Thanks.

app/pages/lab/workflows.jsx
```diff
@@ -19,23 +20,47 @@ const WorkflowsPage = (props) => {
...
 
-      <DragReorderable tag="ul" className="nav-list" items={props.workflows} render={renderWorkflow} onChange={props.handleWorkflowReorder} />
+      {(!props.reorder && props.workflows.length > 0) &&
+        <div>
+          <ul className="nav-list">
+            {props.workflows.map(workflow => renderWorkflow(workflow))}
+          </ul>
+          <hr />
+          <Paginator
+            className="talk"
```
mcbouslog: wouldn't have noticed if hadn't pointed out in different PR, but do we want to avoid using the "talk" class outside of Talk? not sure.

mcbouslog:
are there major pros or cons to implementing as this PR does with `onPageChange` provided from container to Paginator and no `componentWillReceiveProps` vs [this PR](https://github.com/zooniverse/Panoptes-Front-End/pull/3913/files) that uses the Paginator default `onPageChange` and `componentWillReceiveProps` comparing location query param?

app/pages/lab/workflows.jsx
srallen: Good catch. I copy and pasted this component over from another file and didn't notice.

srallen: Ok I did some bigger CSS changes, but I added some more sensible base styles to the `paginator` classes, which will be helpful for #3913 too. I moved the talk style variables in to the common stylus file too.

srallen:
Answering your question, I don't think there are any major pros/cons, just depends on what you're trying to do and either works.

Issue #3218, opened 2016-11-11
Components that still use mixins


rogerhutchings:
- [ ] `app/collections/collaborators.cjsx`
- [ ] `app/collections/settings.cjsx`
- [ ] `app/collections/show.cjsx`
- [ ] `app/components/promise-renderer.cjsx`
- [ ] `app/components/workflow-toggle.cjsx`
- [ ] `app/lib/handle-prop-changes.coffee`
- [ ] `app/lib/promise-to-set-state.coffee`
- [ ] `app/pages/admin/project-status.cjsx`
- [ ] `app/pages/admin/project-status.cjsx`
- [ ] `app/pages/collections/collections-list.cjsx`
- [ ] `app/pages/collections/favorites-list.cjsx`
- [ ] `app/pages/collections/index.cjsx`
- [ ] `app/pages/lab/project.cjsx`
- [ ] `app/pages/lab/project.cjsx`
- [ ] `app/pages/lab/visibility.cjsx`
- [ ] `app/pages/lab/workflow-viewer/index.cjsx`
- [ ] `app/pages/profile/index.cjsx`
- [ ] `app/pages/project/about/education.cjsx`
- [ ] `app/pages/project/about/faq.cjsx`
- [ ] `app/pages/project/about/index.cjsx`
- [ ] `app/pages/project/about/research.cjsx`
- [ ] `app/pages/project/about/results.cjsx`
- [ ] `app/pages/project/about/team.cjsx`
- [ ] `app/pages/project/classify.cjsx`
- [ ] `app/pages/project/index.cjsx`
- [ ] `app/pages/sign-in.cjsx`
- [ ] `app/partials/project-page-editor.cjsx`
- [ ] `app/partials/register-form.cjsx`
- [ ] `app/subjects/comment-form.cjsx`
- [ ] `app/talk/add-zoo-team-form.cjsx`
- [ ] `app/talk/comment-box.cjsx`
- [ ] `app/talk/comment-report-form.cjsx`
- [ ] `app/talk/comment.cjsx`
- [ ] `app/talk/inbox-conversation.cjsx`
- [ ] `app/talk/init.cjsx`

rogerhutchings:
...and the mixins are:

- `CollectionRole`
- `Feedback`
- `HandlePropChanges`
- `History`
- `Navigation`
- `PromiseToSetState`
- `SetToggle`
- `TitleMixin`
- `ToggleChildren`

PR #3911, opened 2017-07-03
Remove Title Mixin

wgranger:
Removes the TitleMixin from several components (#3218): 

`app/collections/show.jsx `
`app/pages/collections/collections-list.cjsx`
`app/pages/collections/favorites-list.cjsx `
`app/pages/collections/index.cjsx`
`app/pages/lab/project.cjsx`
`app/pages/project/classify.cjsx`
`app/pages/project/index.cjsx`
`app/pages/sign-in.cjsx`

Describe your changes.
I rewrote the mixin as a higher-order component (HOC), and applied changes to each component it touched. Some changes were made to `project.cjsx` to remove the `PromiseRenderer` so the title could be set solely from the wrapper instead of using the HOC in the wrapper and the wrapper's child. However, determining when the user prop is received seemed a bit tricky (see `getProjectAndOwners()`.

# Review Checklist

- [x] Does it work on mobile?

https://remove-title-mixing.pfe-preview.zooniverse.org/

## Optional


amyrebecca:
This fails to build for me right now under node 8.1.2

screen shot 2017-07-05 at 12 48 00 pm

srallen:
Rewriting the mixin as a HOC is good practice, but I think long term, we could use [react-helmet](https://github.com/nfl/react-helmet). Apologies if I didn't mention this at the start of the PFE work week.

wgranger:
Ok, rewrote that component a bit in a more straightforward way that should be easier to read. The components still all include HOC, but if a change needs to be made to React Helmet I can do so, either in this PR or a subsequent one.

wgranger:
Closed in favor of #3938

PR #3913, opened 2017-07-03
Lab visibility paginator

jelliotartz:
Fixes part of #3904.

Describe your changes.
- Add paginator component for workflows in project builder visibility page.
- Uses page_size of 20. This can be removed when getWorkflowsInOrder does not override the default page_size.

Staged at https://lab-visibility-paginator.pfe-preview.zooniverse.org/lab/1613/visibility

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -83,6 +83,7 @@ module.exports = React.createClass
       opacity: 0.7
       pointerEvents: 'none'
 
+
     <div>
       <p className="form-label">Project state and visibility</p>
 
@@ -240,17 +241,16 @@ module.exports = React.createClass
           </table>
         </div>}
 
-      <br/>
+      {if @state.workflows.length > 0
+        meta = @state.workflows[0].getMeta()
 
-      {@state.workflows.length > 0 && (
         <Paginator
-          className="talk"
-          page={@state.workflows[0]._meta.workflows.page}
-          pageCount={@state.workflows[0]._meta.workflows.page_count}
-        />
-      )}
+          className="lab-visibility-paginator"
+          page={meta.page}
+          pageCount={meta.page_count}
+        />}
 
-      <hr/>
+      <hr />
 
       <p className="form-label">Status</p>
       <p className="form-help">In a live project active workflows are available to volunteers and cannot be edited. Inactive workflows can be edited if a project is live or in development.</p>
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -40,6 +40,8 @@ module.exports = React.createClass
       pathname: @props.location.pathname
       query: nextQuery
     @getWorkflowList page
+    .then () =>
+      @workflowList?.scrollIntoView()
 
   getWorkflowList: (page) ->
     @setState { loadingWorkflows: true }
@@ -49,9 +51,6 @@ module.exports = React.createClass
         @setState
           workflows: workflows
           loadingWorkflows: false
-      .then () =>
-        if @props.location?.query?.page?
-          @workflowList?.scrollIntoView()
 
   setRadio: (property, value) ->
     @set property, value
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -1,5 +1,4 @@
 React = require 'react'
-ReactDOM = require 'react-dom'
 apiClient = require 'panoptes-client/lib/api-client'
 WorkflowToggle = require '../../components/workflow-toggle'
 SetToggle = require '../../lib/set-toggle'
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -11,9 +11,12 @@ Paginator = require '../../talk/lib/paginator'
 module.exports = React.createClass
   displayName: 'EditProjectVisibility'
 
+  contextTypes:
+    router: React.PropTypes.object.isRequired
+
   getDefaultProps: ->
     project: null
-    location: query: page: 1
+    # location: query: page: 1
 
   getInitialState: ->
     error: null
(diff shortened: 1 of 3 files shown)


app/pages/lab/visibility.cjsx
```diff
@@ -229,6 +239,19 @@ module.exports = React.createClass
             </tbody>
           </table>
         </div>}
+
+      <br/>
```
srallen: Maybe use css or put the paginator in a div (which should have some margin defaults) instead of using a line break tag.

app/pages/lab/visibility.cjsx
```diff
@@ -229,6 +239,19 @@ module.exports = React.createClass
...
+
+      {@state.workflows.length > 0 && (
+        <Paginator
+          className="talk"
+          page={@state.workflows[0]._meta.workflows.page}
+          pageCount={@state.workflows[0]._meta.workflows.page_count}
+        />
+      )}
+
+      <hr/>
```
srallen: Just a minor style thing, but could you put a space between the tag name and its close slash?: `<hr />`

app/pages/lab/visibility.cjsx
```diff
@@ -229,6 +240,18 @@ module.exports = React.createClass
             </tbody>
           </table>
         </div>}
+
+      {if @state.workflows.length > 0
+        meta = @state.workflows[0].getMeta()
+
+        <Paginator
+          className="lab-visibility-paginator"
```
srallen: I wasn't clear, but what I meant is adding the `paginator` class as defined in this [file](https://github.com/zooniverse/Panoptes-Front-End/blob/master/css/paginator.styl) and then modifying in a new sub-class defined in the lab file if needed. You shouldn't need to copy and paste over the styles from the original talk class. 

The CSS in this project is a bit of a mess, but we're trying to use BEM for new classes, so try to follow that if you can and it makes sense for the changes.

srallen: Also, if it makes sense as a basic style, feel free to update the paginator.styl file.

srallen: Yeah, this a bit of my bad after taking a closer look again. There are a few styles that probably should move into the paginator stylus file as base styles for the buttons.

srallen: If you're ok with waiting, #3908 is close to merging and I'll get them in with that and you can rebase after that is merged.

jelliotartz: Sure, no problem! :)

app/pages/lab/visibility.cjsx
```diff
@@ -229,6 +240,18 @@ module.exports = React.createClass
             </tbody>
           </table>
         </div>}
+
+      {if @state.workflows.length > 0
+        meta = @state.workflows[0].getMeta()
+
+        <Paginator
+          className="paginator"
```
mcbouslog: it looks like the Paginator component gets the paginator class by default [here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/talk/lib/paginator.cjsx#L72), so I don't think we need to add on like 248, but please confirm on your local branch.

mcbouslog:
LGTM

jelliotartz:
I've got the visibility page to scroll to the top of the workflow section using the code below, which is implemented in `main.cjsx`. But would you prefer that the custom scroll logic be implemented in the `EditProjectVisibility` component, rather than in `shouldUpdateScroll`? I can push up the current progress if that's helpful.

```
shouldUpdateScroll = (prevRouterProps, routerProps) ->
  pathname = routerProps.location.pathname.split('/')
  isStats = ('stats' in pathname) and ('projects' in pathname)
  isSubjectTalk = ('talk' in pathname) and ('talk' in pathname)
  isLabVisibility = ('lab' in pathname) and ('visibility' in pathname)
  hashChange = routerProps.location.hash.length
  if isStats or hashChange or isSubjectTalk
    false
  else if isLabVisibility
    workflowList = document.getElementById("workflow-list")
    if workflowList
      workflowList.scrollIntoView()
  else
    true
```

srallen:
@jelliotartz It should be specifically for the pagination click event. Although what you changed seems to work, you're modifying the scroll behavior for that entire page at the routing level, so if that page ever has other kinds of hash changes, it'll have that behavior and that may not make sense. So what we want to do:
- Override the default scrolling behavior at the router to not scroll to top on that page
- Then in the click handler for the paginator on page change, scroll into view the top of the workflow list.

app/pages/lab/visibility.cjsx
```diff
@@ -27,8 +31,22 @@ module.exports = React.createClass
...
+    page = if @props.location?.query?.page? then @props.location.query.page else 1
+    @getWorkflowList page
+
+  onPageChange: (page) ->
+    nextQuery = Object.assign {}, @props.location.query, { page }
+    @context.router.push
+      pathname: @props.location.pathname
+      query: nextQuery
+    @getWorkflowList page
+    workflowList = document.getElementById("workflow-list")
```
srallen: Use a [`ref`](https://facebook.github.io/react/docs/refs-and-the-dom.html) instead of directly interacting with the DOM.

app/pages/lab/visibility.cjsx
```diff
@@ -27,8 +31,22 @@ module.exports = React.createClass
...
+    workflowList?.scrollIntoView()
```
srallen: How about we put this in a `then` for the promise of the API request so that it only happens after the successful return of the next or previous page of workflows.

app/pages/lab/visibility.cjsx
jelliotartz: I've been working on trying to use `ref` to target the DOM node that I want for the custom scroll to a specific element, but I'm not getting it to scroll correctly though. 

Should I be using a string, eg`ref="workflow-list"`? or something more like `ref={(node) => {this.workflowList = node; }}`, or `ref={ (node) => @workflowList = ReactDOM.findDOMNode(node) }`?

And in the `.then`, I'm not getting the DOM node to store in a variable. I've been trying things like `node = ReactDOM.findDOMNode(workflowList)` or `ReactDOM.findDOMNode("workflow-list")?.scrollIntoView()`, but this isn't working, and I'm kinda out of ideas for how to get the desired DOM element stored so that I can call `scrollIntoView()` on it. 

If you have a moment @srallen, it'd be great if you could take a look/point me towards how I'm missing the correct syntax.

srallen: Using string literals in refs is deprecated. Be aware of this when googling how to use refs in React, which is why I linked you to the current docs. We can chat about this tomorrow morning.

eatyourgreens: @jelliotartz `ref={node => this.workflowList = node` runs the callback when the component first renders. `this.workflowList` holds a reference to the DOM node, if it's a HTML element, or the component, if it's a React component. You don't need to use `ReactDOM` to access nodes from refs to HTML elements, you can just use `this.workflowList.scrollIntoView()`. 

I'm on my phone, so there may be syntax errors in that answer.

jelliotartz: Thanks @eatyourgreens! That did the trick!

PR #3925, opened 2017-07-10
Collections subject adding bug fixes

### LINKED ISSUE #3821: Collections – adding multiple subjects to a collection causes an error message?
## Expected behavior
Select multiple subjects and add to an existing collection

## Current behavior
From [Talk](https://www.zooniverse.org/talk/14/295770?comment=559967):
>I observe that if I select all items on a screen to add to a collection, most recently I am getting an ugly message that an / some items are already part of the collection and the process fails. 1) I don't reckon the images are part of the collection so I think the error message is not true. 2) Maybe this message is being generated because the collection is already quite large. If I go through the images one at a time, they generally add to the collection okay. 3) If the images are part of the collection, couldn't the process be smart and assign to the collection the images not already added?

srallen:
Fixes #3821 by preemptively filtering out the ids of the subjects to add array by comparing it to the serialized subjects linked of the collection. This also fixes when the add button is disabled as well as improve the error messaging UI. If all of the selected subjects are already added to the target collection, a clearer message is displayed and the modal doesn't auto-close so the user has a chance to read it.

https://fix-3821.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/collections/collection-search.cjsx
+++ b/app/collections/collection-search.cjsx
@@ -6,7 +6,7 @@ module.exports = React.createClass
   displayName: 'CollectionSearch'
 
   propTypes:
-    multi: React.PropTypes.bool
+    multi: React.PropTypes.bool.isRequired
     onChange: React.PropTypes.func
 
   getDefaultProps: ->
--- a/app/collections/collection-search.cjsx
+++ b/app/collections/collection-search.cjsx
@@ -7,17 +7,19 @@ module.exports = React.createClass
 
   propTypes:
     multi: React.PropTypes.bool
-    project: React.PropTypes.object
+    onChange: React.PropTypes.func
 
   getDefaultProps: ->
     multi: false
-    project: null
 
   getInitialState: ->
     collections: []
 
   onChange: (collections) ->
-    @setState {collections}
+    @setState({ collections }, ->
+      if @props.onChange
+        @props.onChange()
+    )
 
   searchCollections: (value) ->
     query =
@@ -28,15 +30,12 @@ module.exports = React.createClass
 
     apiClient.type('collections').get query
       .then (collections) ->
-
-        opts = collections.map (collection) ->
-          {
+        options = collections.map (collection) -> {
             value: collection.id,
             label: collection.display_name,
             collection: collection
           }
-
-        {options: opts}
+        { options }
 
   getSelected: ->
     @state.collections
@@ -52,4 +51,5 @@ module.exports = React.createClass
       searchPromptText="Type to search Collections"
       className="collection-search"
       closeAfterClick={true}
-      loadOptions={@searchCollections} />
+      loadOptions={@searchCollections}
+    />
--- a/app/collections/collections-manager.jsx
+++ b/app/collections/collections-manager.jsx
@@ -1,46 +1,80 @@
 import React from 'react';
 import CollectionsCreateForm from './create-form';
 import CollectionSearch from './collection-search';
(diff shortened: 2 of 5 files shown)


app/collections/collection-search.cjsx
```diff
@@ -7,17 +7,19 @@ module.exports = React.createClass
 
   propTypes:
     multi: React.PropTypes.bool
```
eatyourgreens: Has a default defined, so should be `isRequired`?

app/collections/collections-manager.jsx
```diff
@@ -1,46 +1,80 @@
...
-    if (!(collections.length > 0)) return;
+    if (collections.length === 0) return;
 
-    const promises = collections.map(collection =>
-      collection.collection.addLink('subjects', this.props.subjectIDs)
-    );
+    const promises = collections.map((searchResult) => {
+      const collection = searchResult.collection;
+      const subjectsToAdd = this.props.subjectIDs.filter((id) => {
+        if (!collection.links.subjects.includes(id)) return id;
```
eatyourgreens: I'm not sure if I'm reading this correctly, but  the callback in `Array.filter()` should return a boolean for each array item, but this returns the item.

I think you mean to return `!collection.links.subjects.includes(id)`.

app/collections/collections-manager.jsx
```diff
@@ -1,46 +1,80 @@
...
+        if (this.state.errors.length === 0) this.props.onSuccess();
+        this.setState({ adding: false });
+      }).catch(error => this.setState({ adding: false, errors: [error.toString()] }));
   }
 
   render() {
     return (
       <div className="collections-manager">
-        <h1>Add Subject to Collection</h1>
+        <h1>Add Subject to Collection{' '}{this.state.adding && <LoadingIndicator />}</h1>
```
eatyourgreens: This shouldn't be `h1` unless `Add Subject to Collection` is the top-level heading for the current page.

app/collections/collections-manager.jsx
```diff
@@ -1,46 +1,80 @@
...
     this.addToCollections = this.addToCollections.bind(this);
+    this.onChange = this.onChange.bind(this);
+  }
+
+  onChange() {
+    const collections = this.search.getSelected();
+
+    if (collections.length > 0) {
+      this.setState({ hasCollectionSelected: false });
+    }
```
eatyourgreens: Out of interest, if `collections` is an array of selected collections, why would `hasCollectionSelected` be false? I'm not sure I follow the logic here, which might be a problem for someone who comes along to maintain this later.

srallen: Right. I'll reverse the logic since it's being used to manage the button disabled state.

app/collections/collections-manager.jsx
srallen: Yep, I do. I tend to forget that `Array.filter()` returns a boolean.

srallen:
@eatyourgreens all the comments should be addressed. Changing the header to a `h2` required me to add a bit of hacky css to make sure the dialog stayed the size it was before (modal-form doesn't take classname or style props!). Switching to `h2` made the dialog slightly smaller enough to cause some odd wrapping with the search input, so adding min-width the header makes it the right size again. Adding a min-width to the whole collection-manager class caused some overflow issues and the dialog itself wasn't expanding. This is just another +1 to switch to a true modal component in the future at some point.

PR #3937, opened 2017-07-19
Enable users to choose preferred language

simoneduca:
Describe your changes.
Add a dropdown for users to select the languages available
Stores users preference for next visits
# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://add-locales-menu.pfe-preview.zooniverse.org/

## Optional


app/lib/language-engine.js
```diff
@@ -0,0 +1,24 @@
+import counterpart from 'counterpart';
+
+const load = (language) => {
+  const defaultLanguage = { label: 'English', value: 'en' };
```
eatyourgreens: Should this default to the project primary language if the project isn't in English?

app/lib/language-selector.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component, PropTypes } from 'react';
+import Select from 'react-select';
+import languages from './languages';
+import { select } from './language-engine';
+
+const propTypes = {
+  props: PropTypes.object.isRequired
+};
+const storedLanguage = JSON.parse(localStorage.getItem('preferred-language'));
+const defaultLanguage = { label: 'English', value: 'en' };
```
eatyourgreens: default language is defined in more than one place, which could make it harder to change.

app/lib/language-selector.jsx
```diff
@@ -0,0 +1,51 @@
+import React, { Component, PropTypes } from 'react';
+import Select from 'react-select';
```
eatyourgreens: Why react-select instead of a dropdown menu?

simoneduca:
@eatyourgreens given that this a fairly self-contained feature, I was thinking of using Redux, like Feedback does.

app/lib/language-engine.js
simoneduca: Good point, yes. As you said, that would accommodate people creating projects with a different default language than English.

eatyourgreens:
When the language changes, the [document language](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/lang) should be changed in the HTML too.

Then we can do nifty things like [styling Hebrew as RTL](/zooniverse/Galaxy-Zoo/blob/master/css/locales.styl#L26-L27).

simoneduca:
Neat

eatyourgreens:
There's [a request on Talk for RTL language support](https://www.zooniverse.org/talk/18/392214?comment=655055).

simoneduca:
@eatyourgreens This is so old it's probably safe to close it and start again if this feature is required in the future. Is that ok with you?

simoneduca:
I take that as a 'yes'. Closing.

PR #3938, opened 2017-07-19
Use React-Helmet to Change Titles

wgranger:
Supercedes #3911 

Describe your changes.
Per suggestion by @srallen, this PR uses a third-party library to change doc titles. This was initially done to remove the titleMixin. 

Aside from removing the titleMixin and using react-helmet on components it touches, I've also used react-helmet on some other components to have some sort of standardized doc titles across the site.

https://use-react-helmet.pfe-preview.zooniverse.org/

# Review Checklist

- [ ] Does it work on mobile?

## Optional


rogerhutchings:
Soooooo should all these title strings actually be in counterpart for translation?

_ducks_

wgranger:
@rogerhutchings Yes... Probably so. I can make that change pretty quickly.

app/pages/settings/index.cjsx
```diff
@@ -20,6 +21,7 @@ UserSettingsPage = React.createClass
 
   render: ->
     <section className="user-profile-content">
+      <Helmet title="#{counterpart 'userSettingsPage.title'} » #{@props.user.display_name}" />
       <div className="secondary-page settings-page">
         <h2><Translate content="userSettingsPage.header" /></h2>
```
srallen: Changing the counterpart object broke this header: 
screen shot 2017-07-24 at 11 57 49 am

app/talk/inbox.cjsx
```diff
@@ -14,17 +16,21 @@ PAGE_SIZE = talkConfig.inboxPageSize
 
 promptToSignIn = -> alert (resolve) -> <SignInPrompt onChoose={resolve} />
 
+counterpart.registerTranslations 'en',
+  messagesPage:
+    title: 'Messages'
```
srallen: This is inconsistent with the header and the url of the page. Maybe this should be `Inbox` instead?

wgranger:
Good catch on the above. Title functions have been removed, aside from one with a bit more logic.

eatyourgreens:
Ha! I've approved this but GitHub still says no to merging.

Thanks for moving some more hardcoded English strings into counterpart, by the way.

eatyourgreens:
Maybe we can get an approval from everyone on the front-end team before this is merged 😉

simoneduca:
My comment was only meant as a little nudge to help reducing the time PRs lie around. I assume you were being sarcastic @eatyourgreens, but just in case you weren't: although I think a review from everyone would be unnecessary, another pair of eyes can't hurt, can it?

eatyourgreens:
Pointing out that you and I have both approved this today, so it's very much good to go, but github still makes us wait on @srallen, which is fine but would be a problem if she was away for some reason.

simoneduca:
Gotcha, apologies for the misunderstanding. There is the [dismiss review option](https://github.com/blog/2265-dismissing-reviews-on-pull-requests), but it feels a bit rude to bypass somebody.

eatyourgreens:
That looks really useful if someone is off sick, and a change has to be merged urgently. I agree, rude otherwise to bypass someone like that.

srallen:
@wgranger could you update the mixin issue that this one is done? Thanks!

wgranger:
done

PR #3943, opened 2017-07-20
Audio with image subject

hughdickinson:
Small change to fix an undefined reference to “player”.

Also fixed several linter issues.


eatyourgreens:
@hughdickinson this wants to add 26 commits so looks like something is still out of synch with your branch.

app/components/file-viewer/audio-player.jsx
```diff
@@ -105,9 +103,11 @@ class AudioPlayer extends React.Component {
           {imageElement}
         </div>
         <div className="audio-player-component">
-          <audio className="subject" controls={true} ref={(element) => {
-            this.player = element;
-          }} src={this.audioSrc()} type={this.audioTypeString()} preload="auto" onCanPlay={this.onAudioLoad.bind(this)} onClick={this.playAudio.bind(this, !this.state.playing)} onEnded={this.endAudio} onTimeUpdate={this.updateProgress.bind(this)}>
+          <audio
+            className="subject" controls={true} ref={(element) => {
```
eatyourgreens: maybe break this line up so there's one prop per line?

PR #3946, opened 2017-07-21
Prevent muliple classifications of same choice

### LINKED ISSUE #3727: Survey task - same identification allowed multiple times
## Expected behavior
Identification should get updated if same animal is selected
## Current behavior
Two separate identifications are allowed when same animal is selected.
screen shot 2017-04-13 at 17 02 29

## Steps to replicate
Select same animal from filter.

simoneduca:
Fixes #3727 .

Describe your changes.
If the selected choice matches an existing one, the latter is passed as a prop to the choice component. 

**Expected behaviour**: see the same choice displayed, same answers, etc. Also classification submitted for same choice should be one.

**Current behviour**: Tested with Western Montana Wildlife: if choice already exists, the existing one is correctly displayed. _However two annotations with the same choice are still submitted in the classification._

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-3727.pfe-preview.zooniverse.org/?env=production

## Optional

### DIFF
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -52,7 +52,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotation?.answers || {}
     focusedAnswer: ''
 
   checkFilledIn: ->
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -48,7 +48,8 @@ module.exports = React.createClass
       {if @state.selectedChoiceID is ''
         <Chooser task={@props.task} filters={@state.filters} onFilter={@handleFilter} onChoose={@handleChoice} onRemove={@handleRemove} annotation={@props.annotation} focusedChoice={@state.focusedChoice} />
       else
-        <Choice task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
+        existingAnnotation = _.find(@props.annotation.value, 'choice', @state.selectedChoiceID) || {}
+        <Choice annotation={existingAnnotation} task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
     </div>
 
   handleFilter: (characteristicID, valueID) ->
@@ -75,7 +76,7 @@ module.exports = React.createClass
     @setState filters: {}
 
   clearSelection: ->
-    @setState 
+    @setState
       selectedChoiceID: ''
       focusedChoice: @state.selectedChoiceID
     newAnnotation = Object.assign {}, @props.annotation, _choiceInProgress: false


app/classifier/tasks/survey/choice.cjsx
```diff
@@ -52,7 +52,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotation?.answers || {}
```
eatyourgreens: If you add `annotation` to the default props, then you don't need to set a default for answers here.

app/classifier/tasks/survey/index.cjsx
```diff
@@ -48,7 +48,8 @@ module.exports = React.createClass
       {if @state.selectedChoiceID is ''
         <Chooser task={@props.task} filters={@state.filters} onFilter={@handleFilter} onChoose={@handleChoice} onRemove={@handleRemove} annotation={@props.annotation} focusedChoice={@state.focusedChoice} />
       else
-        <Choice task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
+        existingAnnotation = _.find(@props.annotation.value, 'choice', @state.selectedChoiceID) || {}
```
eatyourgreens: Could this use `Array.filter` instead of `_.find`?

app/classifier/tasks/survey/index.cjsx
```diff
@@ -48,7 +48,8 @@ module.exports = React.createClass
...
+        <Choice annotation={existingAnnotation} task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
```
eatyourgreens: `@handleAnnotation` is called on confirm, so you should update that to handle existing annotations correctly too.

eatyourgreens:
Leaving this open because multiple annotations are still being created for the same choice.

srallen:
Should we find out if there are any projects that relies on or wants this behavior before changing it? While most survey task projects don't want this, some might, like Chimp & See that wants each individual chimp identified in addition to any other species.

simoneduca:
I'll ask on #general.

app/classifier/tasks/survey/index.cjsx
simoneduca: Probably, but we have lodash so why not use it?

app/classifier/tasks/survey/index.cjsx
simoneduca: Good catch, thanks

app/classifier/tasks/survey/choice.cjsx
simoneduca: Do you mean it should go back to `answers: {}`?

app/classifier/tasks/survey/index.cjsx
eatyourgreens: I think this is the only line in PFE that's using Underscore. If Underscore is dropped in a future version of lodash, this will break.

Arrays have come along way since Underscore was first written, and you should favour native Javascript methods where possible.

app/classifier/tasks/survey/choice.cjsx
eatyourgreens: Set the default `annotation` prop to `{ answers: {}}`.

app/classifier/tasks/survey/index.cjsx
eatyourgreens: `@handleAnnotation` is causing the bug where each choice is pushed to the end of the `values` array as a new annotation, even if that choice already exists.

srallen:
What do you think about changing the survey task styling to be more like Wildcam's survey task when a selection is made instead of listing out the selections on top the subject viewer?

app/classifier/tasks/survey/index.cjsx
eatyourgreens: Alternatively, you could import `find` at the top of the file so that it's obvious that the dependency is there.

simoneduca: Absolutely agree. I just didn't realise it's not used anywhere else. I'd be happy to remove the package then, which was already installed.

simoneduca:
WG's styling for that is so much neater, so I'm happy to do that.

app/classifier/tasks/survey/index.cjsx
eatyourgreens: Actually, I'm wrong. I searched PFE for `_.` but I don't think anything accesses it directly like that. There are components that import from `lodash`.

srallen: lodash is being used in other parts of the codebase, but for things like deep merges.

simoneduca: We went full circle now.

eatyourgreens: I'd still favour `[existingAnnotation] = annotation.value.filter( value => value.choice === this.state.selectedChoiceID )` to pluck out a single answer from `annotation.value`.

eatyourgreens: Incidentally, since there are multiple answers with the same choice ID at the moment, does the current line just pick out the first one? Or does `existingAnnotation` get set to an array if it finds two annotations?

simoneduca: `_.find` returns the first match https://lodash.com/docs/4.17.4#find
In hindsight, `.filter` is better for this.

simoneduca:
I've incorporated the comments on this into a new incoming PR. Re. styling I will request a separate PR. Closing.

PR #3947, opened 2017-07-25
Survey task: Prevent multiple classifications of same choice

### LINKED ISSUE #3727: Survey task - same identification allowed multiple times

simoneduca:
Fixes #3727 .

Replaces https://github.com/zooniverse/Panoptes-Front-End/pull/3946

Describe your changes.
If the current choice of a survey task already exists, this allows to edit it and submit it as _one_ classification.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://remove-multiple-choice.pfe-preview.zooniverse.org/?env=production Tested with Western Montana Widlife

### DIFF
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -45,7 +45,12 @@ module.exports = React.createClass
   displayName: 'Choice'
 
   propTypes:
-    annotation: React.PropTypes.object.isRequired
+    annotation: React.PropTypes.shape({
+      answers: React.PropTypes.object,
+      choice: React.PropTypes.string,
+      filters: React.PropTypes.object
+    })
+
 
   getDefaultProps: ->
     annotation: { answers: {} }
@@ -56,7 +61,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: @props.annotation?.answers || {}
+    answers: @props.annotation.answers
     focusedAnswer: ''
 
   checkFilledIn: ->
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -44,7 +44,11 @@ ImageFlipper = React.createClass
 module.exports = React.createClass
   displayName: 'Choice'
 
+  propTypes:
+    annotation: React.PropTypes.object.isRequired
+
   getDefaultProps: ->
+    annotation: { answers: {} }
     task: null
     choiceID: ''
     onSwitch: Function.prototype
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -45,15 +45,14 @@ module.exports = React.createClass
   displayName: 'Choice'
 
   propTypes:
-    annotation: React.PropTypes.shape({
+    annotationValue: React.PropTypes.shape({
       answers: React.PropTypes.object,
       choice: React.PropTypes.string,
       filters: React.PropTypes.object
     })
 
-
   getDefaultProps: ->
-    annotation: { answers: {} }
+    annotationValue: { answers: {} }
     task: null
     choiceID: ''
     onSwitch: Function.prototype
@@ -61,7 +60,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: @props.annotation.answers
+    answers: @props.annotationValue.answers
     focusedAnswer: ''
 
   checkFilledIn: ->
--- a/app/classifier/tasks/survey/choice.cjsx
+++ b/app/classifier/tasks/survey/choice.cjsx
@@ -52,7 +52,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotation?.answers || {}
     focusedAnswer: ''
 
   checkFilledIn: ->
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -50,8 +50,8 @@ module.exports = React.createClass
       else
         # @ is undefined within the scope of find
         currentSelection = @state.selectedChoiceID
-        existingAnnotation = @props.annotation.value.find (value) -> value.choice is currentSelection
-        <Choice annotation={existingAnnotation} task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
+        existingAnnotationValue = @props.annotation.value.find (value) -> value.choice is currentSelection
+        <Choice annotationValue={existingAnnotationValue} task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
     </div>
 
   handleFilter: (characteristicID, valueID) ->
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -86,8 +86,7 @@ module.exports = React.createClass
 
   handleAnnotation: (choice, answers, e) ->
     filters = JSON.parse JSON.stringify @state.filters
-    currentSelection = @state.selectedChoiceID
-    value = @props.annotation.value?.filter (value) -> value.choice isnt currentSelection
+    value = @props.annotation.value?.filter (value) -> value.choice isnt choice
     value.push {choice, answers, filters}
     @clearFilters()
     @clearSelection()
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -86,14 +86,9 @@ module.exports = React.createClass
 
   handleAnnotation: (choice, answers, e) ->
     filters = JSON.parse JSON.stringify @state.filters
-    value = @props.annotation.value?.slice?(0) ? []
-    # @ is undefined within the scope of findIndex
     currentSelection = @state.selectedChoiceID
-    annotationIndex = value.findIndex (value) -> value.choice is currentSelection
-    if value[annotationIndex]
-      value[annotationIndex] = @props.annotation.value[annotationIndex]
-    else
-      value.push {choice, answers, filters}
+    value = @props.annotation.value?.filter (value) -> value.choice isnt currentSelection
+    value.push {choice, answers, filters}
     @clearFilters()
     @clearSelection()
     setTimeout => # Wait for filters and selection to clear.
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -50,7 +50,7 @@ module.exports = React.createClass
       else
         # @ is undefined within the scope of find
         currentSelection = @state.selectedChoiceID
-        existingAnnotation = @props.annotation.value.find (value) -> value.choice is currentSelection || {}
+        existingAnnotation = @props.annotation.value.find (value) -> value.choice is currentSelection
         <Choice annotation={existingAnnotation} task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
     </div>
 
--- a/app/classifier/tasks/survey/index.cjsx
+++ b/app/classifier/tasks/survey/index.cjsx
@@ -48,7 +48,10 @@ module.exports = React.createClass
       {if @state.selectedChoiceID is ''
         <Chooser task={@props.task} filters={@state.filters} onFilter={@handleFilter} onChoose={@handleChoice} onRemove={@handleRemove} annotation={@props.annotation} focusedChoice={@state.focusedChoice} />
       else
-        <Choice task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
+        # @ is undefined within the scope of find
+        currentSelection = @state.selectedChoiceID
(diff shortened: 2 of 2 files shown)


eatyourgreens:
[Array.findIndex](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) and [Array.find](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex) aren't supported in IE. This will work better, cross-browser, if uses `indexOf` and `filter` to find items in arrays.

Other than that, looks good to me. Pinging @amyrebecca for a review since she is more familiar with the survey task than I am.

app/classifier/tasks/survey/choice.cjsx
```diff
@@ -52,7 +52,7 @@ module.exports = React.createClass
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotation?.answers || {}
```
eatyourgreens: Is there a `propType` (and `defaultProp` if necessary) defined for `annotation`?

app/classifier/tasks/survey/index.cjsx
```diff
@@ -84,7 +87,13 @@ module.exports = React.createClass
   handleAnnotation: (choice, answers, e) ->
     filters = JSON.parse JSON.stringify @state.filters
     value = @props.annotation.value?.slice?(0) ? []
-    value.push {choice, answers, filters}
+    # @ is undefined within the scope of findIndex
+    currentSelection = @state.selectedChoiceID
+    annotationIndex = value.findIndex (value) -> value.choice is currentSelection
```
eatyourgreens: this might be better using `value.indexOf(currentSelection)`

app/classifier/tasks/survey/index.cjsx
```diff
@@ -84,7 +87,13 @@ module.exports = React.createClass
...
+    if value[annotationIndex]
```
eatyourgreens: `if annotationIndex > -1`

app/classifier/tasks/survey/index.cjsx
```diff
@@ -84,7 +87,13 @@ module.exports = React.createClass
...
+      value[annotationIndex] = @props.annotation.value[annotationIndex]
```
eatyourgreens: Should this be `value[annotationIndex] = { choice, answers, filters }`?

app/classifier/tasks/survey/index.cjsx
simoneduca: Aren't `.find` and others methods polyfilled?

eatyourgreens:
Looking a bit closer, I'm not sure that this allows you to edit an annotation. Looks to me like it never replaces an existing value with new answers.

app/classifier/tasks/survey/index.cjsx
eatyourgreens: I'm not sure. This looks fine if they are, though though I'm not sure why using an anonymous callback is better than `value.indexOf(currentSelection)`.

simoneduca:
Perhaps I'm missing something, but on WMW I was able to edit answers and submitting the edited annotations.

app/classifier/tasks/survey/index.cjsx
srallen: `Array.find` has been added to the features we're polyfilling from the polyfill service we're using. `findIndex` might come with the default set, but should be double checked: https://github.com/zooniverse/Panoptes-Front-End/blob/master/views/index.ejs#L74

Or `value.indexOf()` could be used like @eatyourgreens is suggesting.

simoneduca: I'm not too fussed. I will change it.

eatyourgreens:
@simoneduca weird. Just reading the code, line 94 looks like it puts the old annotation, from `@props.annotation`, back into the new array.

app/classifier/tasks/survey/choice.cjsx
simoneduca: There isn't. I didn't remove this line and replaceit with `annotation : { answers: {} }` because I didn't want to change all the occurences of `@@state.answers`. I can do that, if you prefer though.

eatyourgreens: I think a `propType` definition would be helpful for anyone who has to maintain this code. I'm not fussed about a default value, but eslint nags you to add one when a prop isn't required, so that's probably a good practice to get into.

eatyourgreens: Looking at the coffeescript for `Choice`, looks like the original practice on PFE was to document props by adding them to `getDefaultProps` so maybe `annotation: { answers: {} }` should go in there after all.

app/classifier/tasks/survey/index.cjsx
simoneduca: @eatyourgreens btw just passing `currentSelection` as it is doesn't work. There's probably a way to make good old `indexOf` work here, but to me it seems easier to just add `findIndex` to the polyfill url. Unless you show me a clean method to get `indexOf` to work.

eatyourgreens: That's odd. [indexOf](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf) should return -1 if the argument doesn't exist.

eatyourgreens: Oh wait, `selectedChoice` isn't an item in that array, is it.

How about removing the `if` check in favour of
```coffeescript
value = value.filter((value) -> { value.choice !== selectedChoiceID})
value.push({choice, answers, filters)
```

Remove the old answer, if it exists, then add the new one to the end of the array as usual.

So many variables called `value` there 😞

simoneduca: Nice one, your last suggestion is definetely the way to go, thanks.

simoneduca:
I've removed `.findIndex` which [doesn't come as default with the polyfill service](https://github.com/Financial-Times/polyfill-service/issues/1087#issuecomment-279973954). I left `.find` though, which is polyfilled and seems more suitable here, because I don't need an array returned but a value.

app/classifier/tasks/survey/index.cjsx
```diff
@@ -75,15 +78,16 @@ module.exports = React.createClass
...
 
   clearSelection: ->
-    @setState 
+    @setState
       selectedChoiceID: ''
       focusedChoice: @state.selectedChoiceID
     newAnnotation = Object.assign {}, @props.annotation, _choiceInProgress: false
     @props.onChange newAnnotation
 
   handleAnnotation: (choice, answers, e) ->
```
eatyourgreens: `choice` here is the selected choice ID, so you don't need `selectedChoiceID` later.

app/classifier/tasks/survey/index.cjsx
```diff
@@ -75,15 +78,16 @@ module.exports = React.createClass
...
     filters = JSON.parse JSON.stringify @state.filters
-    value = @props.annotation.value?.slice?(0) ? []
+    currentSelection = @state.selectedChoiceID
+    value = @props.annotation.value?.filter (value) -> value.choice isnt currentSelection
```
eatyourgreens: This might not do what you expect if, for some reason, `currentSelection` and `choice` have different values. They should be the same, but I'm not sure about relying on that assumption.

simoneduca: Good point, thanks. Testing suing`choice` now.

simoneduca: That works nicely, pushed.

simoneduca:
@eatyourgreens happy for me to merge or want to allow more eyes on this first?

app/classifier/tasks/survey/index.cjsx
```diff
@@ -48,7 +48,10 @@ module.exports = React.createClass
       {if @state.selectedChoiceID is ''
         <Chooser task={@props.task} filters={@state.filters} onFilter={@handleFilter} onChoose={@handleChoice} onRemove={@handleRemove} annotation={@props.annotation} focusedChoice={@state.focusedChoice} />
       else
-        <Choice task={@props.task} choiceID={@state.selectedChoiceID} onSwitch={@handleChoice} onCancel={@clearSelection} onConfirm={@handleAnnotation} />}
+        # @ is undefined within the scope of find
+        currentSelection = @state.selectedChoiceID
+        existingAnnotation = @props.annotation.value.find (value) -> value.choice is currentSelection || {}
```
eatyourgreens: Delete `|| {}`

app/classifier/tasks/survey/choice.cjsx
```diff
@@ -44,15 +44,19 @@ ImageFlipper = React.createClass
...
+    annotation: { answers: {} }
     task: null
     choiceID: ''
     onSwitch: Function.prototype
     onConfirm: Function.prototype
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotation?.answers || {}
```
eatyourgreens: Now that you've set a default value for the prop, this line becomes `answers: @props.annotation.answers`.

app/classifier/tasks/survey/index.cjsx
eatyourgreens: `undefined` is a valid value for `existingAnnotation` here (there might not be an annotation yet) so you don't need to set a default value of `{}`. The `Choice` component is responsible for setting a default value if the annotation doesn't exist.

eatyourgreens:
Sorry, I've got a couple more small comments on things that could be tidied up.

app/classifier/tasks/survey/choice.cjsx
```diff
@@ -44,15 +44,19 @@ ImageFlipper = React.createClass
 module.exports = React.createClass
   displayName: 'Choice'
 
+  propTypes:
+    annotation: React.PropTypes.object.isRequired
```
eatyourgreens: You've set a default value for this, so remove `.isRequired`.

eatyourgreens: Also, it's better to define a shape for the object here. The linter would warn that `object` isn't allowed as a prop type.

simoneduca: I agree, but I opted for `object` because it didn't seem right to have so many nested shapes:
```
annotation: React.PropTypes.shape({
      answers.propTypes.shape({
          // what goes in here? Type-checking all the possible answers seems crazy
      }),
      choice.propTypes.string,
      filters.propTypes.shape({
         // what goes in here?
      })
    })
```
Thoughts?

eatyourgreens:
```javascript
annotation: React.PropTypes.shape({
  answers: React.PropTypes.object,
  choice: React.PropTypes.string
})
```
shouldn't generate any warnings from eslint. I'm not 100% sure if that's exactly what a survey task annotation looks like, but I think it's right.

eatyourgreens: Is `filters` present on the annotations that get passed in? I'm not sure what that looks like, but I guess you could figure it out from the code in `Index`, maybe?

simoneduca: It is and I will add it to the type check. Technically `annotation` here was a misnomer, because it refers to only the `value` of an annotation. Understanding the variable names was part of the challenge for me.

eatyourgreens: I thought you added this prop? Renaming it is an excellent idea if the name is confusing.

Re. confusing names, `choice` and `choiceID` being used for the same thing in `Index` always confuses me, but `choice` is baked into the annotations now, so renaming it would break aggregation for the survey task 😞

eatyourgreens: Looks like [`filters` is also a generic object](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/survey/index.cjsx#L42).

simoneduca: I did add it to `choice.cjsx`, but not to `index.cjsx` https://github.com/blog/2265-dismissing-reviews-on-pull-requests. That's were I found it confusing, but it's not a biggie. I pushed latest changes.

eatyourgreens: I'm pretty sure the `annotation` property that's passed into a task is the actual annotation property from the current classification, not one of its values.

eatyourgreens: You're right, the same name is being used for two completely different things in `Choice` and `Index`.

simoneduca: That was partly my fault. Should be a better now.

app/classifier/tasks/survey/choice.cjsx
```diff
@@ -44,15 +44,23 @@ ImageFlipper = React.createClass
...
+    annotationValue: { answers: {} }
     task: null
     choiceID: ''
     onSwitch: Function.prototype
     onConfirm: Function.prototype
     onCancel: Function.prototype
 
   getInitialState: ->
-    answers: {}
+    answers: @props.annotationValue.answers
```
eatyourgreens: If this component mounts before the annotation has updated in the parent, then the value here will be from the previous annotation.

I think the solution to that is to check `annotationValue` in `componentWillReceiveProps` and update state if necessary.

simoneduca: On it ;)

simoneduca: Oddly enough
```
    console.log 'componentWillReceiveProps', nextProps.annotationValue, @props.annotationValue
    if nextProps.annotationValue isnt @props.annotationValue
      @setState
        answers: {}
```
put here, just after `getinitialState` doesn't seem to fire :/

simoneduca: Experimenting with `componentWillMount` but nope.

PR #3955, opened 2017-07-27
edit stats workflow request to page_size of project.links.workflows.length

mcbouslog:
Towards #3904.

Currently NfN's most recent workflow, 4667 Butterfly_Hummingbird Moths, does not appear on [the project stats page](https://www.zooniverse.org/projects/zooniverse/notes-from-nature/stats) because the `getWorkflowsInOrder` request's `page_size` is 100, and this workflow is over that `page_size`.

This PR changes the stats page `getWorkflowsInOrder` request's `page_size` to the `project.links.workflows.length`, resulting in the inclusion of the recent workflow - https://edit-stats-workflows-request.pfe-preview.zooniverse.org/projects/zooniverse/notes-from-nature/stats?env=production.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/stats/index.jsx
+++ b/app/pages/project/stats/index.jsx
@@ -38,7 +38,8 @@ class ProjectStatsPageController extends React.Component {
       'retirement,subjects_count'
     ];
     const query = {
-      fields: fields.join(',')
+      fields: fields.join(','),
+      page_size: project.links.workflows.length
     };
     getWorkflowsInOrder(project, query)
       .then((workflows) => {


eatyourgreens:
Can this make repeated paged requests instead, looping until `workflows.length` is reached? I'm thinking something like the project preference requests that fill up the project stats circle on the logged-in homepage.

I'm not sure about the implications of requesting 100 resources from Panoptes in a single request.

camallen:
it really depends on the end point and how much data it has to serialize. If it's a lot then you run into issues with ELB timouts (60s) and serializing data over the network (MBs?).It's a balancing operation, if the api end point is fast and data volumes small then go for it. If it's not the switch to pages of data instead of heavy loading the api and tying up server resources. 

Final note, HTTP caching can help with this, it's on my list to cycle back to.

eatyourgreens:
@camallen I think discussion has moved [over here](/zooniverse/Panoptes-Front-End/issues/3904#issuecomment-318665526)

PR #3956, opened 2017-07-27
exception for _toolIndex in no-underscore-dangle eslint rule

simensta:
**Describe your changes.**
In the eslint file, I added an exception for the no-underscore-dangle rule, so that we don't get lint errors when using `_toolIndex`.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -24,6 +24,7 @@
     "multiline-ternary": ["warn", "never"],
     "no-else-return": "off",
     "no-trailing-spaces": 1,
+    "no-underscore-dangle": ["error", { "allow": ["_toolIndex"] }],
     "no-unused-vars": 1,
     "no-useless-constructor": 1,
     "no-useless-return": 1,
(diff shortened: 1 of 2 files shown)


simoneduca:
I could only find two occurrences of `_toolIndex`. Is the point of adding the rule, just to tackle those? I don't mind adding the rule either way, just want to clarify.

simensta:
In a previous PR, I added this rule modification at the [file level](https://github.com/zooniverse/Panoptes-Front-End/blob/b4f12a3ce84dccc1929fc9b8ff409fdf3082aff8/app/classifier/tasks/highlighter/index.jsx#L1) for `tasks/highlighter/index.jsx`. Seeing `_toolIndex` occur in multiple files (and possibly in future files) prompted me to propose this rule change at the project level. For example, this rule will affect `drawing/index.cjsx` and `drawing/marking-initializer.cjsx` once those files are rewritten in JavaScript.

PR #3957, opened 2017-07-27
organizations admin section

mcbouslog:
Adds admin organizations section similar to admin projects section.

Staged branch - https://orgs-admin.pfe-preview.zooniverse.org/admin/organization-status

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/admin.cjsx
+++ b/app/pages/admin.cjsx
@@ -43,7 +43,7 @@ AdminPage = React.createClass
                 activeClassName="active">
                 <Translate content="userAdminPage.nav.grantbot" />
               </Link>
-              <Link to="/admin/organization_status"
+              <Link to="/admin/organization-status"
                 type="button"
                 className="secret-button admin-button"
                 activeClassName="active">
--- a/app/pages/admin.cjsx
+++ b/app/pages/admin.cjsx
@@ -12,6 +12,7 @@ counterpart.registerTranslations 'en',
       createAdmin: "Manage Users"
       projectStatus: "Set Project Status"
       grantbot: "Grantbot"
+      organizationStatus: "Set Organization Status"
 
 AdminPage = React.createClass
   displayName: 'AdminPage'
@@ -41,6 +42,12 @@ AdminPage = React.createClass
                 className="secret-button admin-button"
                 activeClassName="active">
                 <Translate content="userAdminPage.nav.grantbot" />
+              </Link>
+              <Link to="/admin/organization_status"
+                type="button"
+                className="secret-button admin-button"
+                activeClassName="active">
+                <Translate content="userAdminPage.nav.organizationStatus" />
               </Link>
             </nav>
           </aside>
--- a/app/pages/admin/organization-status-list.jsx
+++ b/app/pages/admin/organization-status-list.jsx
@@ -1,5 +1,4 @@
 import React, { Component } from 'react';
-import { Link } from 'react-router';
 import apiClient from 'panoptes-client/lib/api-client';
 import LoadingIndicator from '../../components/loading-indicator';
 import ProjectIcon from '../../components/project-icon';
@@ -29,8 +28,16 @@ class OrganizationStatusList extends Component {
   }
 
   getOrganizations() {
+    const { query } = this.props.location;
+
+    const organizationsQuery = {
+      include: 'avatar'
+    };
+
+    const mergedQuery = Object.assign({}, organizationsQuery, query);
+
     this.setState({ loading: true, error: null });
-    return apiClient.type('organizations').get({ include: 'avatar' })
+    return apiClient.type('organizations').get(mergedQuery)
       .then((organizations) => { this.setState({ organizations, loading: false }); })
       .catch((error) => { this.setState({ error: `Error requesting organizations:, ${error}`, loading: false }); });
   }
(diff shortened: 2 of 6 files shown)


app/pages/admin/organization-status-list.jsx
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  getOrganizations() {
+    const { query } = this.props.location;
+
+    const organizationsQuery = {
+      include: 'avatar'
+    };
+
+    const mergedQuery = Object.assign({}, organizationsQuery, query);
```
mcbouslog: this is more than necessary, but leaving for eventual sorting and filtering once functionality available from back-end.

app/pages/admin/organization-status.jsx
```diff
@@ -0,0 +1,107 @@
...
+
+import ProjectIcon from '../../components/project-icon';
+import LoadingIndicator from '../../components/loading-indicator';
+
+import Toggle from './project-status/toggle';
+
+class OrganizationStatus extends Component {
+  constructor(props) {
+    super(props);
+    this.forceUpdate = this.forceUpdate.bind(this);
```
simoneduca: Where is `forceUpdate` defined?

simoneduca: I've never used the client's methods for watching changes, so I might just be missing something obvious.

eatyourgreens: `forceUpdate` is [a React component API method](https://facebook.github.io/react/docs/react-component.html#forceupdate).

simoneduca: Oh, thought that's a no go for new code. Is that necessary as a`ChangeListener` replacement?

eatyourgreens: It is generally a bad idea, since it causes pain for people like you and me when we try and understand the component. What's it being used for here?

simoneduca: No idea, but perhaps this comment gives a clue? https://github.com/zooniverse/Panoptes-Front-End/pull/3957/files#diff-31cef41494be24db197d9701f5abac41R37

eatyourgreens: I think you're justified in blocking this PR if you don't know what a piece of code does.

simoneduca: @mcbouslog ?

eatyourgreens: @simoneduca from what I can tell from the [json-api-client README](/zooniverse/json-api-client/blob/master/README.md) that listener you linked too will fire when `organisation` changes. So the component renders when the organisation resource is changed inside or outside this component. I think.

eatyourgreens: A better pattern would be to have the component react to organisation changes in component state.

simoneduca: Yep

eatyourgreens: And handle changes outside the component via `componentWillReceiveProps`.

mcbouslog: The bad reasoning for code as is was duplicating the current project admin components, but I'll rewrite these organization admin components to use `componentWillReceiveProps` and also refactor the project admin components in a separate PR.

app/pages/admin.cjsx
```diff
@@ -42,6 +43,12 @@ AdminPage = React.createClass
                 activeClassName="active">
                 <Translate content="userAdminPage.nav.grantbot" />
               </Link>
+              <Link to="/admin/organization_status"
```
srallen: URLs should use dashes instead of underscores as they have better readability.

app/router.cjsx
```diff
@@ -232,6 +234,8 @@ module.exports =
       <Route path="project_status" component={ProjectStatusList} />
       <Route path="project_status/:owner/:name" component={ProjectStatus} />
       <Route path="grantbot" component={Grantbot} />
+      <Route path="organization_status" component={OrganizationStatusList} />
+      <Route path="organization_status/:owner/:name" component={OrganizationStatus} />
```
srallen: Same comment about the URLs here. Dashes should be preferred. URLs for projects should fixed too, but that's out of scope for this PR.

app/pages/admin/organization-status.jsx
```diff
@@ -0,0 +1,138 @@
...
+        this.setState({ organization });
+        return organization;
+      });
+  }
+
+  getProjects() {
+    this.state.organization.get('projects', { sort: 'display_name' })
+      .then((projects) => {
+        this.setState({ projects });
+      });
```
srallen: This should have a catch.

app/pages/admin/organization-status-list.jsx
```diff
@@ -0,0 +1,81 @@
...
+    if (prevProps.location.query !== this.props.location.query) {
+      this.getOrganizations();
+    }
+  }
+
+  getOrganizations() {
+    const { query } = this.props.location;
+
+    const organizationsQuery = {
+      include: 'avatar'
```
srallen: There's a bug with the json-api-client that if you are including a single resource that it must be in an array: https://github.com/zooniverse/json-api-client/issues/35

app/pages/admin/organization-status.jsx
```diff
@@ -0,0 +1,138 @@
...
+  renderProjects() {
+    if (this.state.projects.length === 0) {
+      return <div>No projects found</div>;
+    }
+
+    return (
+      <ul className="project-status__section-list">
+        {this.state.projects.map(project =>
+          <li key={project.id} className="section-list__item">
+            <span>{`${project.id} - ${project.display_name}`}</span>
```
srallen: Not a blocker, but maybe this can link to the project's admin status page?

PR #3964, opened 2017-08-01
Remove type-check checkbox

simoneduca:
This resumes the conversation about the PR template I interrupted by hastily closing https://github.com/zooniverse/Panoptes-Front-End/pull/3963

Describe your changes.
As suggested by @srallen, I removed the type-check checkbox, because the linter catches that.


simoneduca:
Sorry @srallen getting to this only now. I assume the checkbox about tests should be in the main section and _not_ optional. Is that right?

srallen:
Yes, I think so. Thanks!

PR #3970, opened 2017-08-02
Autofocus question tasks

eatyourgreens:
Adds autofocus for question tasks by focusing the question, rather than the first answer, if no answer has been selected yet.

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? https://autofocus-questions.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/generic.jsx
+++ b/app/classifier/tasks/generic.jsx
@@ -13,11 +13,14 @@ export default class GenericTask extends React.Component {
   }
 
   componentDidMount() {
-    this.props.autoFocus && this.container.focus && this.container.focus();
+    if (this.props.autoFocus && this.container.focus) {
+      this.container.focus();
+    }
   }
 
-  componentWillReceiveProps(nextProps) {
-    if (nextProps.answers !== this.props.answers && nextProps.autoFocus && this.container.focus) {
+  componentDidUpdate(prevProps) {
+    // autofocus the container again if the question changes
+    if (this.props.autoFocus && this.props.question !== prevProps.question && this.container.focus) {
       this.container.focus();
     }
   }
--- a/app/classifier/tasks/generic.jsx
+++ b/app/classifier/tasks/generic.jsx
@@ -25,12 +25,14 @@ export default class GenericTask extends React.Component {
   showHelp() {
     alert(
       (resolve, reject) =>
-        <div className="content-container">
-          <Markdown className="classification-task-help">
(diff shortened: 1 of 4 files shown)


eatyourgreens:
This should allow you to answer question tasks from the keyboard, without raising the issues in #2563

eatyourgreens:
https://autofocus-questions.pfe-preview.zooniverse.org/projects/mschwamb/comet-hunters/classify?env=production looks like a good project for trying this out.

eatyourgreens:
Good point about the tutorial. It might be launching before focus is set, so that there's no active element to return focus to once it closes. I'll have a look.

eatyourgreens:
I've fixed the tutorial focus, which reminded me that the `Tutorial` component is still written in coffeescript. I wasn't sure if converting it would be in scope here, so I might leave that to a separate PR.

Mini-courses should have similar behaviour, if they don't already.

mcbouslog:
Looks good to me. 

Not sure if should hold anything up, but if consecutive question tasks the second won't have focus like the first - https://autofocus-questions.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?workflow=3023.

eatyourgreens:
I thought that was fixed in 63dad58c but maybe not?

eatyourgreens:
Checking the task sounds like a good idea. I'm not sure why I didn't do that.

eatyourgreens:
@mcbouslog turns out that the `GenericTask` component doesn't receive `task` as a prop so I'll give this some further thought.

eatyourgreens:
@mcbouslog should be fixed in 590d454

mcbouslog:
Jim your prediction is correct, if a multiple question task's option(s) are selected, then unselected, a following question task's focus will be on the Done button instead of the task title. However, if the first task requires at least one answer, this issue doesn't come up.

I'm not sure if this should block things though, it might be an edgy enough case to punt on?

Regardless, I'm trying to wrap my head around a possible solution (sorry for the false start on a `task` prop). Maybe we could compare the `question` prop between `prevProps` and `this.props`, presuming they wouldn't be the same between back to back question tasks?

eatyourgreens:
Thanks @mcbouslog you are absolutely right. I should be checking to see if the question has changed.

eatyourgreens:
Updated to look for question changes when determining whether to autofocus the container.

PR #3976, opened 2017-08-03
Allow to unsubscribe for older browser users

### LINKED ISSUE #3969: Allow unsubscribe form submission if checkValidity isn't supported
## Expected behavior

The user should be able to submit the unsubscribe form even if their browser doesn't support email validation.

## Current behavior

Form submission is only enabled if the [email address is valid](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/unsubscribe.cjsx#L21)

## Additional information

- **Operating system:**
- **Browser:**

simoneduca:
Fixes #3969.

https://fix-unsubscribe.pfe-preview.zooniverse.org/unsubscribe

Describe your changes.
- Rewrite component in JS
- Default `checkValidity()` to true when `undefined` to support older browsers.

**Note**: I've tried to reproduce the original functionality as close as possible.
However, does anyone know if url `/unsubscribe?email=something` is ever returned by the server? If not, I can get rid of more code.

# Review Checklist
- [ ] Does it work on mobile?

## Optional


app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
+import React, { Component, PropTypes } from 'react';
+import Auth from 'panoptes-client/lib/auth';
```
eatyourgreens: This isn't a class so should be `auth` instead of `Auth`.

app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+  handleSubmit(event) {
+    event.preventDefault();
+    const email = this.state.emailValue;
+    this.setState({
+      inProgress: true,
+      emailSuccess: false,
+      emailError: false
+    });
+    Auth.unsubscribeEmail({ email })
```
eatyourgreens: `auth.unsubscribeEmail(…`

app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
...
+
+class UnsubscribeFromEmails extends Component {
+  constructor() {
+    super();
+    this.handleChange = this.handleChange.bind(this);
+    this.handleSubmit = this.handleSubmit.bind(this);
+    this.renderForm = this.renderForm.bind(this);
+    this.renderIconFeedback = this.renderIconFeedback.bind(this);
+    this.renderTextFeedback = this.renderTextFeedback.bind(this);
+    this.renderTokenFlowFeedback = this.renderTokenFlowFeedback.bind(this);
```
eatyourgreens: You don't have to bind the render methods here: they're already bound to the instance. Only the event handlers need to be bound.

app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
...
+      inProgress: false,
+    };
+  }
+  componentDidMount() {
+    return this.handleChange;
+  }
+
+  handleChange(event) {
+    const emailIsValid = event.target.checkValidity();
+    const emailValue = event.target.value || event.target.defaultValue;
```
eatyourgreens: I'm not sure about `defaultValue` here. It isn't a standard DOM node property.

simoneduca: Noted thanks. `defaultValue` is the bit I'd like to get rid of because I don't think `/unsubscribe?email=something` is ever passed.

app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
...
+  }
+
+  render() {
+    const { location } = this.props;
+    const defaultValue = location.query ? location.query.email : '';
+    const disabledButton = !this.state.emailIsValid || this.state.emailSuccess;
+    const query = location.query ? location.query : {};
+    return (
+      <div className="centered-grid">
+        {query && query.processed
```
eatyourgreens: `query` here is always true, because it was set on line 118, so this check is redundant.

app/pages/unsubscribe.jsx
```diff
@@ -0,0 +1,138 @@
...
+      </div>
+    );
+  }
+}
+
+UnsubscribeFromEmails.propTypes = {
+  location: PropTypes.shape({
+    query: PropTypes.shape({
+      email: PropTypes.string,
+      processed: PropTypes.bool
```
eatyourgreens: URLs are always strings, so `bool` here might throw prop type warnings in the browser.

simoneduca: Ah good catch, I forgot to change that when I actually realised what processed was. cheers

app/pages/unsubscribe.jsx
eatyourgreens: Turns out [`defaultValue` is a React thing](https://facebook.github.io/react/docs/uncontrolled-components.html#default-values).

simoneduca: Mm, Uncontrolled component aren't recommended though.

eatyourgreens: Following up on that, it might give you warnings about switching an input between being controlled and uncontrolled.

eatyourgreens:
Why is this changing the package lock file?

simoneduca:
The previous build failed because of a missing module. So I've `rm -rf node_modules/ && npm install` and pushed. Checks are passing now.

eatyourgreens:
I'm not sure about this. This PR doesn't add or remove any dependencies so I would prefer if you deleted that commit from the PR.

simoneduca:
Done. Travis is failing because module `graceful-readlink` is missing. Mm.

eatyourgreens:
But passing on [this build](https://travis-ci.org/zooniverse/Panoptes-Front-End/builds/260617962). I don't think this PR is breaking Travis, since PRs with the same package lock file were passing earlier today.

simoneduca:
Rebased, tests passing now.

app/pages/unsubscribe.jsx
simoneduca: No warnings. I might leave it as it is until it's clear whether that url is ever used.

eatyourgreens: I suspect `event.target.value` always exists, so the righthand side of the or comparison is never reached here.

simoneduca: Correct, it should have been the other way round `event.target.defaultValue || event.target.value`

eatyourgreens: `defaultValue` isn't a property of DOM elements, so I'm 99% sure that first one is always false.

simoneduca: Argh. I thought we've established it's a React property.

eatyourgreens: Yeah, that's why I think you won't find it on the DOM node returned by `event.target`. I'm not 100% sure about that.

simoneduca: I see. How come it works then?

eatyourgreens: If both do exist, and they have different values, which one should you unsubscribe? I'd go with the input value.

simoneduca: I'm confused. The code is supposed to prepopulate the field when [`location.query.email` isn't empty](https://github.com/zooniverse/Panoptes-Front-End/blob/98986cd8ca96cb63fdb38c63f8495dd69caf17a7/app/pages/unsubscribe.jsx#L112) and it that case I should unsubscribe the value in the url.

eatyourgreens: It works because the input element always has a value, stored in `event.target.value`. Looking at the original code, the email address is always taken from the input value, which I think is the desired behaviour.

`element.defaultValue` sets the initial value of `element.value`, but `element.value` can change from that initial value.

simoneduca: Check latest commit

eatyourgreens: 👍🏻 I think it was working for you because you were looking at a case where they both have the same value.

simoneduca:
I would have liked to write a test for the page, before merging but I can do that in separate PR.

eatyourgreens:
Keep this branch open. You can continue to push to it, and open new PRs as you add commits.

Issue #3122, opened 2016-10-12
Classify as a user group


srallen:
## Expected behavior

Select from a drop down a user group one is a member of and classify as a part of a group.
## Current behavior

_Please include any error messages from the browser console and/or screenshots_
Classifier does not have the UI to select and classify as a group

This functionality is related to Intro2Astro, but is being generalized for PFE.

srallen:
After discussing with @marten and @zwolf, this UI feature is unnecessary. Classifications are counted toward all groups you are a member of. Education API then determines what to display based on assignment.

srallen:
Reopening because caesar needs to know the user group membership for reductions. Classification by user group will be selected by query param similar to how workflow id is used for assignments with Wildcam Labs.

srallen:
Added query param and to classification metadata in #3979. Validation is in #4061

PR #3979, opened 2017-08-04
Add group query param and group id to classification metadata

srallen:
For #3122 and to support the Intro2Astro use case with the EducationAPI front end being built.

This adds a check for the presence of the url query param `?group=id` and if present, adds the `user_group` id to the classification metadata on create as well as adds a UI notification banner that you are a student classifying with your classroom.

Can be seen working on staging: https://classify-by-group.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats/classify?group=2534

Note: I'm not validating the id in the query param that it is actually a group that exists nor am I validating that the user is authenticated or that the user is in that group. We may want to consider these validations, but I can add those in subsequent PRs. 

I know we talked in the last retrospective being more strict about coffeescript to javascript conversions, so I am working on that, but in a separate branch.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -145,6 +145,9 @@ module.exports = React.createClass
           workflow: workflow.id
           subjects: [subject.id]
 
+      if @props.location.query?.group?
+        classification.update({ 'metadata.user_group': @props.location.query.group })
+
       # If the user hasn't interacted with a classification resource before,
       # we won't know how to resolve its links, so attach these manually.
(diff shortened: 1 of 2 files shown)


srallen:
@amyrebecca is adding the `user_group` id to the classification metadata going to work for caesar's new grouping feature?

amyrebecca:
@srallen that should work. we'll need to make sure to write extractors that harvest that field and figure out how to specify to the reducer how to find it.

Issue #3763, opened 2017-05-04
Allow feedback in classifier to be shown in a specified subject frame


rogerhutchings:
For multi-frame subjects, you should be able to specify which frame drawing feedback will appear in.

cc #3725 @aprajita

eatyourgreens:
Perhaps the summary should flip the subject to the default frame, if one is specified in the subject metadata.

eatyourgreens:
NB this might be easier if we redesign the classifier component hierarchy so that the subject is a child of the task, not a sibling. That way, the task components can pass props down to the subject to change how it displays.

eatyourgreens:
Linking this to #3858

PR #3993, opened 2017-08-18
Fix general feedback series of small bugs

### LINKED ISSUE #3858: Space Warps Generalised Feedback Issues
## Expected behaviour

Generalised feedback messages expected to behave differently for
- SIM success & missed
  - appear at sim marker location
  - default messages overridden by metadata
  - with multiple images showing and clone markers on only expect one to contain feedback message

- DUD success & missed
  - would be nice to have a pop-up message like in SIM at the centre of the image for no marker placed (i.e. success), and at the location of the marker for missed SIMS. users will be expecting and probably would only register the same kind of message for duds as for sims especially for a fast 

## Current behaviour
- messages appear at strange locations
- both default and metadata messages appearing together
- dud messaging only in classification summary text, request pop-up messages for duds
- with clone marker on, all images show feedback, only one should be enough for this setup

image

## Steps to replicate
@rogerhutchings please try out workflow https://www.zooniverse.org/projects/aprajita/swdemo/classify?reload=0&workflow=4257

## Additional information

- **Operating system:**
mac os x yosemite v10.10.5
- **Browser:**
Chrome Version 58.0.3029.110 (64-bit)

simoneduca:
Fixes #3858 

Describe your changes.
This PR addresses three main points:

- [ ] fix tooltips locations
- [x] metadata messages override defaults
- [ ] with clone marker on, all images show feedback, only one should be enough for this setup

And, possibly, a nice to have:
- [x] dud messaging only in classification summary text, request pop-up messages for duds

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-sw-feedback.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/feedback/helpers.js
+++ b/app/classifier/feedback/helpers.js
@@ -16,4 +16,8 @@ const isThereFeedback = (subject, workflow) =>
     return newResult;
   }, false);
 
-export { isFeedbackActive, isThereFeedback };
+const renderUniqueFeedback = (annotations, feedback) => {
+  const annotationFrame = annotations.map(item => item.value.map(value => value.frame));
+  return feedback.filter(item => item.frame === annotationFrame);
+};
+export { isFeedbackActive, isThereFeedback, renderUniqueFeedback };
--- a/app/classifier/feedback/process-feedback-drawing.js
+++ b/app/classifier/feedback/process-feedback-drawing.js
@@ -11,6 +11,7 @@ const isWithinTolerance = (annotationX, annotationY, feedbackX, feedbackY, toler
 
 const createFeedbackItem = (success, annotation, rule) => {
   const feedbackItem = {
+    frame: annotation.value.map(item => parseInt(item.frame)),
     task: annotation.task,
     type: 'drawing',
     success,
--- a/app/classifier/feedback/process-feedback-drawing.js
+++ b/app/classifier/feedback/process-feedback-drawing.js
@@ -16,16 +16,10 @@ const createFeedbackItem = (success, annotation, rule) => {
     success,
     message: (success) ? rule.successMessage : rule.failureMessage
   };
-
-  if (rule.dud) {
-    feedbackItem.target = 'summary';
-  } else {
-    feedbackItem.x = rule.x;
-    feedbackItem.y = rule.y;
-    feedbackItem.tol = rule.tol;
-    feedbackItem.target = 'classifier';
-  }
-
+  feedbackItem.x = rule.x;
+  feedbackItem.y = rule.y;
+  feedbackItem.tol = rule.tol;
+  feedbackItem.target = 'classifier';
   return feedbackItem;
 };
 
--- a/app/classifier/annotation-renderer/svg.jsx
+++ b/app/classifier/annotation-renderer/svg.jsx
@@ -194,10 +194,10 @@ export default class SVGRenderer extends React.Component {
               </Draggable>
             )}
             {children}
-            {(showFeedback) && (<SVGFeedbackViewer />)}
+            {(showFeedback) && (<SVGFeedbackViewer annotations={annotations} />)}
           </g>
         </svg>
-        {(showFeedback) && (<SVGToolTipLayer getScreenCTM={this.getScreenCurrentTransformationMatrix} />)}
+        {(showFeedback) && (<SVGToolTipLayer annotations={annotations} getScreenCTM={this.getScreenCurrentTransformationMatrix} />)}
         {this.props.children}
       </div>
     );
--- a/app/classifier/feedback/process-feedback-drawing.spec.js
+++ b/app/classifier/feedback/process-feedback-drawing.spec.js
@@ -114,7 +114,7 @@ describe('processDrawingFeedback', function () {
     assert.strictEqual(result[0].task, DUD_ANNOTATION.task);
     assert.strictEqual(result[0].success, true);
     assert.strictEqual(result[0].message, SUCCESS_MESSAGE);
-    assert.strictEqual(result[0].target, 'summary');
+    assert.strictEqual(result[0].target, 'classifier');
   });
 
   it('should return nothing if the subject is a dud and there are no annotations, but showing a success message is disabled', function () {
@@ -131,7 +131,7 @@ describe('processDrawingFeedback', function () {
     assert.strictEqual(result[0].task, NORMAL_ANNOTATION.task);
     assert.strictEqual(result[0].success, false);
     assert.strictEqual(result[0].message, FAILURE_MESSAGE);
-    assert.strictEqual(result[0].target, 'summary');
+    assert.strictEqual(result[0].target, 'classifier');
   });
 
   it('should return nothing if the subject is a dud and there are annotations, but showing a failure message is disabled', function () {
--- a/app/classifier/feedback/process-feedback-single.js
+++ b/app/classifier/feedback/process-feedback-single.js
@@ -13,7 +13,7 @@ const processSingleFeedback = (annotation, subject, task) => {
         question: task.question,
         success: result,
         message: (result) ? rule.successMessage : rule.failureMessage,
-        target: 'summary'
+        target: 'classifier'
       });
     }
 
--- a/app/classifier/feedback/process-feedback-single.spec.js
+++ b/app/classifier/feedback/process-feedback-single.spec.js
@@ -49,7 +49,7 @@ describe('processSingleFeedback', function () {
     assert.strictEqual(result[0].success, true);
     assert.strictEqual(result[0].message, SUCCESS_MESSAGE);
     assert.strictEqual(result[0].question, QUESTION);
-    assert.strictEqual(result[0].target, 'summary');
+    assert.strictEqual(result[0].target, 'classifier');
   });
 
   it('should return nothing if a rule and annotation match, but showing a success message is disabled', function () {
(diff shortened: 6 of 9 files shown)


simoneduca:
@eatyourgreens @aprajita would you mind having a look at the first checklist above and tell me if you're still experiencing any of the points. Thanks.

eatyourgreens:
I think all of those are still problems.

eatyourgreens:
Here are the individual issues for that list:
#4022 Tooltip locations
#4023 All feedback should appear in popup boxes
#3763 Feedback showing across all subject frames

simoneduca:
Legend, thanks @eatyourgreens

simoneduca:
Fix #4023.

eatyourgreens:
Here's a link to the test project on staging: https://fix-sw-feedback.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test

eatyourgreens:
I'm able to reproduce #4022 if I resize the window on the test project. The vertical position of the feedback seems to depend on viewport height, and the feedback vanishes off the top of the viewport for small viewport heights.

Also on the test project:  feedback isn't showing up for duds, or for the question workflows, since d981234. I think the feedback for those might be missing (x,y) coords that are needed to display the tooltip.

aprajita:
Hi @simoneduca 
thanks for working on this & apologies for the delayed response.
regarding  metadata messages override defaults
which i note has been ticked. I think the metadata did override the defaults it was jsut when I accidentally had feedback_2_type defined but no  #feedback_2_successMessage #feedback_2_failureMessage defined so it was reverting to defaults as it had nothing else to go on, which seems like an appropriate fallback in the user forgets to define the feedback.

for #4022 @eatyourgreens when you reproduced this, did you also see the oversized feedback boxes as well as the offset position?

thanks!

simoneduca:
@eatyourgreens you're right, the question workflows are missing coordinates in the feedback item object, hence they don't display. Sorry my bad. I don't think they were intended to originally, but I'll try to change that. Same for DUDs.

simoneduca:
I still can't figure out how to add the coordinates though. Perhaps @rogerhutchings can shed some light?

aprajita:
@simoneduca shall i upload a manifest with the centre of the field or some other default for the duds - would that help?
In the existing manifest i switched the feedback_1_successMessage to correspond to no marker, and a marker being place to trigger the the feedback_1_successMessage

simoneduca:
Thanks @aprajita; tbh I'm not sure. What would the expected behaviour be in the case of centre of the field?

aprajita:
I am not sure!!  but you can try it at
https://www.zooniverse.org/projects/aprajita/swdemo/classify?reload=0&workflow=4817
i've put 160,45 as the feedback coordinate for duds (not sure what to put for tolerance?). no sure yet whether the triggers are reversed for mark or no mark.
the sim/lens feedback appears but it does have extra large boxes still.
the dud messages still appear on he right hand side in this workflow though. perhaps i should upload it to a test staging @eatyourgreens pasted in?
thanks

simoneduca:
Interesting. Have we seen this behaviour before? I don't think I have 
screen shot 2017-09-19 at 16 13 11

aprajita:
hi @simoneduca 
no, i haven't seen this before
is this on the workflow i sent a link for as I didn't get that with LEN/SIM they seemed to behave well.

thanks
aprajita.

aprajita:
sorry @simoneduca ignore my last post, i just noticed that no markers were placed so the negative feedback is correct.
this behaviour is as expected, there's one subject where i included feedback_2 for 2 possible features either side of the galaxy to test how this works.

eatyourgreens:
@aprajita the link I posted is a test project, on the staging database, which has workflows and simple test subjects for each of the different types of feedback that are available in Panoptes.
https://master.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test
It allows @simoneduca to test code changes to make sure changes for SW aren't affecting other types of feedback.

Space Warps is using point-based feedback, so the subjects on that workflow should be test cases for feedback in SW. You don't need to upload any subjects to staging, but do let us know if there are test cases that are missing.

aprajita:
thanks for explaining @eatyourgreens 
I tried the 2 test cases, looks fine with a couple of comments
- I dont see success message for unmarked duds here
- also another case could be to test could be like what @simoneduca pasted in the case there are 2 possible locations denoted by feedback_1_* and feedback_2_* to see how & where feedback is presented.

simoneduca:
I'm a bit stuck with trying to get feedback to display only in the marked frame. I know that it's a matter of adding[ this check](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/markings-renderer.cjsx#L58) somewhere, I just can't find where. Any suggestions @eatyourgreens?

eatyourgreens:
Did you see my note on the original issue about maybe using the default frame for the subject?

simoneduca:
Yes but I don't know what that means.

eatyourgreens:
Feedback within the image gets rendered on [line 199 of the SVGRenderer](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/annotation-renderer/svg.jsx#L199) and the current frame number is passed in as a prop to that component.

eatyourgreens:
@simoneduca @rogerhutchings added a default frame to the subject viewer, for Etch A Cell, so that it doesn't always start on the first frame.

simoneduca:
I'm still lost. I can see that frame 0 is set as default, but the issue originally said that we should be able to specify frames. That is something to change in the project builder.

eatyourgreens:
So here's a question: what's the difference between [SVGFeedbackViewer](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/annotation-renderer/svg.jsx#L195) and [SVGTooltipLayer](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/annotation-renderer/svg.jsx#L199)?

It looks like both those lines are places where you should check the frame number before rendering feedback.

eatyourgreens:
If 0 is the default then I guess only display feedback on frame 0. Otherwise, only display it on the specified default frame.

eatyourgreens:
Forgot to mention: I think the default frame is specified on the subject metadata, and I think the subject viewer method is called `getInitialFrame`. Etch A Cell are using it, so maybe look at that project, if it isn't clear from the subject viewer code.

simoneduca:
It's a good question. I raked my brains yesterday, but the code of both components is pretty much identical and couldn't figure out why needed both.

simoneduca:
Re Etch A Cell, do you mean looking at the workflow?

eatyourgreens:
I'd start by looking at `getInitialFrame`. That should point you towards where in Panoptes the default frame is coming from.

eatyourgreens:
Actually, simplest/quickest implementation would be to code feedback for frame 0 now, then add a variable default frame later.

srallen:
Could we break this PR up into the PRs to fix the individual issues so that this doesn't get stalled and too far behind master?

simoneduca:
Feedback is fairly self contained, but sure.

simoneduca:
I'm fixing the tests I broke with the last commit. Please note though: the last commit actually doesn't fix the duplicate feedback issue. I just pushed it in order to ask for help.

eatyourgreens:
@srallen has a good point. Maybe trash this branch and start separate branches for each of the issues, since this branch has broken the test project and fails the tests.

eatyourgreens:
Also, this branch has fallen behind the latest version of the SVG annotation component.

app/classifier/feedback/helpers.js
```diff
@@ -16,4 +16,8 @@ const isThereFeedback = (subject, workflow) =>
     return newResult;
   }, false);
 
-export { isFeedbackActive, isThereFeedback };
+const renderUniqueFeedback = (annotations, feedback) => {
+  const annotationFrame = annotations.map(item => item.value.map(value => value.frame));
```
eatyourgreens: `Array.map` maps one array onto another array, so either name this `annotationFrames` or reduce the array down to a single frame value.

app/classifier/feedback/helpers.js
```diff
@@ -16,4 +16,8 @@ const isThereFeedback = (subject, workflow) =>
...
+  return feedback.filter(item => item.frame === annotationFrame);
```
eatyourgreens: `annotationFrame` is an array, so this will never return true.

app/classifier/feedback/process-feedback-drawing.js
```diff
@@ -11,6 +11,7 @@ const isWithinTolerance = (annotationX, annotationY, feedbackX, feedbackY, toler
 
 const createFeedbackItem = (success, annotation, rule) => {
   const feedbackItem = {
+    frame: annotation.value.map(item => parseInt(item.frame)),
```
eatyourgreens: Again, `Array.map` returns an array, so reduce it down to a single value to get a frame number.

eatyourgreens:
@simoneduca I left some comments on that commit. Is that the help you were looking for?

simoneduca:
It is, thanks.

eatyourgreens:
@simoneduca [Array.reduce](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/Reduce?v=a) is possibly what you need, but I'm not sure how you intended your code to behave when the map returns more than one frame number (or none.)

simoneduca:
@rogerhutchings is rewriting the feedback feature to address these issues as well as other feature requests. Closing this.

PR #3998, opened 2017-08-24
Resize home page images

### LINKED ISSUE #3994: Optimize home page media assets
## Expected behavior
The site doesn't weigh ~10MB on the home page.  
## Current behavior
Some of the home-* images are big, ranging from 781KB to 564KB. Also the featured project image is 239KB. These can all be resized / optimized for the view port.

## Steps to replicate
Open dev tools, turn off caching and look at the network requests for Img assets. 
The reason i was surprised was the cost / weight of data on our logged out home page -
 https://whatdoesmysitecost.com/test/170822_HC_6e97d744f72cdfc2b6439a51c8418afc#usdCost

We should check all our assets are optimized (both size / format / quality) to reduce any network overhead on our site. This should apply for any new PR's that add media assets to the site.

eatyourgreens:
Fixes #3994 .

Describe your changes.
Renders home page images using the Thumbnail component, with a reasonable maximum width set.

Ideally, we should not be committing large images into this repo in the first place, so the originals might need some optimisation.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://resize-images.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/pages/home-common/featured-project.jsx
+++ b/app/pages/home-common/featured-project.jsx
@@ -10,20 +10,23 @@ Featured Project Component
 
 import React from 'react';
 import { Link } from 'react-router';
+import Thumbnail from '../../components/thumbnail';
 
-const FeaturedProject = (() =>
+const FeaturedProject = (() => (
   <section className="home-featured">
     <h1 className="secondary-kicker">Featured Project</h1>
     <div className="home-featured-images">
-      <img
+      <Thumbnail
         alt="Planetary Response Network & Rescue Global: Caribbean Storms 2017"
-        src="/assets/featured-projects/featured-project-20170913-irma.jpg"
+        src="//zooniverse.org/assets/featured-projects/featured-project-20170913-irma.jpg"
+        width={800}
       />
     </div>
     <h2 className="secondary-headline">Caribbean Storms 2017</h2>
     <p className="display-body">Check back daily to join the relief effort to help the victims of Hurricane Irma.</p>
     <Link to="projects/vrooje/planetary-response-network-and-rescue-global-caribbean-storms-2017">View Project!</Link>
   </section>
+  )
 );
 
 export default FeaturedProject;
--- a/app/pages/home-common/featured-project.jsx
+++ b/app/pages/home-common/featured-project.jsx
@@ -18,7 +18,7 @@ const FeaturedProject = (() => (
     <div className="home-featured-images">
       <Thumbnail
         alt="Planetary Response Network & Rescue Global: Caribbean Storms 2017"
-        src="//zooniverse.org/assets/featured-projects/featured-project-20170913-irma.jpg"
+        src="//www.zooniverse.org/assets/featured-projects/featured-project-20170913-irma.jpg"
         width={800}
       />
     </div>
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Thumbnail component should be updated so that it doesn't require the full URL for an image.

app/pages/home-common/featured-project.jsx
```diff
@@ -10,14 +10,16 @@ Featured Project Component
...
 
 const FeaturedProject = (() =>
   <section className="home-featured">
     <h1 className="secondary-kicker">Featured Project</h1>
     <div className="home-featured-images">
-      <img
+      <Thumbnail
         alt="Steller Sea Lion"
-        src="/assets/featured-projects/featured-project-20170801-steller.jpg"
+        src="//www.zooniverse.org/assets/featured-projects/featured-project-20170801-steller.jpg"
```
camallen: This won't work for preview branch deploys if the assets aren't in the www.zooniverse.org bucket path due to the nginx rewrites.

eatyourgreens: Turns out mapping staging branches to S3 bucket paths isn't simple. I'm not sure if that's something that should be handled with some extra logic in the `Thumbnail` component, or deferred to the backend somewhere.

If it's handled in the javascript, we'll need a new component method that can take a path from the image src, determine whether or not it contains a hostname, map that hostname to an S3 bucket path then return the S3 path.

eatyourgreens:
Rebased to updated the featured project.

simoneduca:
Is this ready to go or Cam's comment about deploy preview branches is a blocker?

eatyourgreens:
Git hub hid my comment about staging branches (from last week), because it was made on an outdated commit, so here it is again.

> Turns out mapping staging branches to S3 bucket paths isn't simple. I'm not sure if that's something that should be handled with some extra logic in the Thumbnail component, or deferred to the backend somewhere.

> If it's handled in the javascript, we'll need a new component method that can take a path from the image src, determine whether or not it contains a hostname, map that hostname to an S3 bucket path then return the S3 path.

eatyourgreens:
Rebased (and staged again) to update the featured project.

app/pages/home-not-logged-in/discover.jsx
```diff
@@ -53,7 +54,7 @@ const HomePageDiscover = (({ showDialog }) =>
     </div>
 
     <div className="home-discover__image">
-      <img role="presentation" src="/assets/home-computer.png" />
+      <Thumbnail role="presentation" src="//www.zooniverse.org/assets/home-computer.png" width={450} />
```
simoneduca: Why did you switch to an absolute url?

eatyourgreens: You can't pass a relative URL to the Thumbnail component.

eatyourgreens: The generated image URL ends up being something like https://thumbnails.zooniverse.org/450x999/www.zooniverse.org/assets/home-computer.png

Although I'm not seeing that for the featured project so I guess thumbnails.zooniverse.org expects the first part to be an actual bucket name on S3.

simoneduca: I don't think it's ideal in general to use absolute urls. But /shrug.

eatyourgreens: That was Cam's comment too.

It turns out to  be a bucket path on S3. I'm interested in hearing ideas on how to fix that for staging branches. See my comment above.

PR #4004, opened 2017-08-24
refactor slider task to stateless component

### LINKED ISSUE #4003: slider task state issues
## Expected behavior
clicking up or down arrows in number input increases/decreases value by task step amount

## Current behavior
takes two clicks to increase/decrease value by task step amount, and if click combination of ups/downs input displays value different than `annotation.value`

## Steps to replicate
https://www.zooniverse.org/projects/achmorrison/steelpan-vibrations/classify

Steps here added by @camallen
1. from the number input, use up / down arrows to verify step functionality works as expected (currently a 0.5 step)
0. use your mouse to drag / click around on the slider to change the value while using up / down to step
0. observe step over and step back values being unexpected. 
    + i press up (right) / down (left) and numbers change with strange steps, e.g.
    +  up only results in 4,6,5,7
    +  down 6,8,5,7
0. check submitted annotation differs from the UI value (only happens when the slider is whack)

mcbouslog:
Fixes #4003.

Refactor slider task to stateless component.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://refactor-slider-no-state.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/slider/index.jsx
+++ b/app/classifier/tasks/slider/index.jsx
@@ -22,7 +22,7 @@ const SliderSummary = (props) => {
       </div>
     </div>
   );
-}
+};
 
 SliderSummary.propTypes = {
   task: React.PropTypes.shape(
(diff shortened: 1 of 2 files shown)


app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
+  return (
+    <div>
+      <GenericTask
+        question={task.instruction}
+        help={task.help}
+        required={task.required}
+      >
+        <div className="slider-task-container full ">
+          <div className="slider-task-range">
+            <label htmlFor="slider-range" className="answer">
```
eatyourgreens: HtmlFor isn't doing anything here and should be removed.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
-              />
+              </div>
             </label>
           </div>
-        </GenericTask>
-      </div>
-    );
-  }
-}
+          <label htmlFor="slider-number" className="answer slider-task-number">
```
eatyourgreens: htmlFor should also be removed here.

eatyourgreens:
The input labels aren't being used correctly, but I think that was a problem with the original code too. 

Is there a test that changes the slider and checks for a corresponding change in the annotation value?

eatyourgreens:
The slider task works properly in the dev classifier on master and #4003 seems to be caused by the slider having a step size of .5, but only generating integer numbers. So you have to click/slide twice to increment/decrement the counter.

I'm not 100% sure why this PR would fix that, so I'm leaving off merging this until I better understand what the underlying bug is.

eatyourgreens:
Turns out there are a couple of bugs in the slider task code, which I've fixed in #4005.

eatyourgreens:
Looks like this includes the same changes I made in #4005 👍🏻

I like the removal of state. Setting state from props is what caused the original bug. 

This needs testing to make sure the component isn't  flipping between controlled and uncontrolled when it sets the input values.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
-      const newAnnotation = Object.assign({}, this.props.annotation, { value: this.props.task.defaultValue });
-      this.props.onChange(newAnnotation);
-    } else if (this.props.annotation.value !== this.state.value) {
-      this.setState({ value: this.props.annotation.value });
-    }
-  }
-
-  handleChange(e) {
+const SliderTask = ({ task, annotation, onChange, autoFocus }) => {
+  const handleChange = (e) => {
```
eatyourgreens: Not a blocking issue, but I find explicit function declarations `function handleChange(e) {…}` easier to read, and debug, than assigning an anonymous function to a variable.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
     let value = e.target.value;
-    if (parseFloat(value, 10) > parseFloat(this.props.task.max, 10)) {
-      value = this.props.task.max;
-    } else if (parseFloat(value, 10) < parseFloat(this.props.task.min, 10)) {
-      value = this.props.task.min;
+    if (parseFloat(value, 10) > parseFloat(task.max, 10)) {
+      value = task.max;
```
eatyourgreens: `value = Math.min(parseFloat(value, 10), parseFloat(task.max, 10));` would also work here.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
+    } else if (parseFloat(value, 10) < parseFloat(task.min, 10)) {
```
eatyourgreens: value = Math.max(parseFloat(value, 10), parseFloat(task.min, 10));

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
-        </GenericTask>
-      </div>
-    );
-  }
-}
+          <label htmlFor="slider-number" className="answer slider-task-number">
+            <input
+              className="standard-input"
+              type="number"
+              autoFocus={autoFocus}
```
eatyourgreens: Either the range input or the number input should have autofocus, but not both.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
+              <div>
+                <input
+                  className="standard-input"
+                  type="range"
+                  autoFocus={autoFocus}
+                  onChange={handleChange.bind(this)}
+                  max={task.max}
+                  min={task.min}
+                  step={task.step}
+                  value={annotation.value || task.defaultValue}
```
eatyourgreens: Will setting the value like this throw a React warning about setting a controlled input as uncontrolled (or vice versa)?

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,68 @@ SliderSummary.propTypes = {
...
+          <label htmlFor="slider-number" className="answer slider-task-number">
+            <input
+              className="standard-input"
+              type="number"
+              autoFocus={autoFocus}
+              onChange={handleChange.bind(this)}
+              max={task.max}
+              min={task.min}
+              step={task.step}
+              value={annotation.value || task.defaultValue}
```
eatyourgreens: Same comment here about setting the value of a controlled input.

eatyourgreens:
Some small comments about code style, but I can't see anything that would block merging this.

mcbouslog:
noting from Slack:
- best to start fresh branch after fixes merged with #4005 
- check if providing an initial value to a numeric input locks it up

CKrawczyk:
I seem to recall there was an issue with setting the default value (as specified in the workflow editor) correctly if the component did not have state, but if the tests are passing it is likely no longer an issue.

Issue #1913, opened 2015-11-17
Add "delete" button under "Manage Users" on admin page


astopy:
The decision in #956 was to handle account deletions manually on a case-by-case basis. We need a UI for actually doing that.

PR #4010, opened 2017-08-29
Delete users

marten:
Fixes https://github.com/zooniverse/Panoptes-Front-End/issues/1913

Added a button for deleting a user in the admin section.

Refactored to move the editting page under a different route. Without this, there was no way to tell the router to go back to the admin index and stop showing the deleted resource. This also means you can now send URLs for specific users around in Slack etc, and can refresh the page when you're devving on it without needing to re-type the user login every time.

TODO:

- [ ] Figure out how to add tests (loads user dynamically from API, which is making it difficult)
- [x] Update to the correct `json-api-client`

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/admin/user-settings/delete-user.jsx
+++ b/app/pages/admin/user-settings/delete-user.jsx
@@ -1,0 +1,63 @@
+import React, { Component } from 'react';
+
+import LoadingIndicator from '../../../components/loading-indicator';
+
+class DeleteUser extends Component {
+    constructor(props) {
+        super(props);
+        this.state = {deletionInProgress: false};
+        this.handleClick = this.handleClick.bind(this)
+        this.performDelete = this.performDelete.bind(this)
+    }
+
+    render() {
+        return (
+            <div>
+                <button type="button" className="minor-button" disabled={this.state.deletionInProgress} onClick={this.handleClick} style={{background: "red", color: "white"}}>
+                    Delete this user
+                    <LoadingIndicator off={!this.state.deletionInProgress} />
+                </button>
+                {this.state.deletionError && (
+                     <div className="form-help error">{this.state.deletionError.message}</div>
+                )}
+            </div>
+        );
+    }
+
+    handleClick() {
+        this.setState({deletionError: null});
+
+        var phrase = prompt("You are about to delete this user and all their projects! This cannot be reversed.\nEnter this user's login to confirm.");
+
+        if (phrase === this.props.user.login) {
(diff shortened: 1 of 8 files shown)


marten:
Currently blocked by https://github.com/npm/npm/issues/18337

app/pages/admin/user-settings/delete-user.jsx
```diff
@@ -0,0 +1,63 @@
...
+           <div className="form-help error">{this.state.deletionError.message}</div>
+        )}
+      </div>
+    );
+  }
+
+  handleClick() {
+    this.setState({deletionError: null});
+
+    var phrase = prompt("You are about to delete this user! This cannot be reversed.\nEnter this user's login to confirm.");
```
marten: I'm not terribly excited about this solution, but on the other hand, it's only an admin tool?

marten:
@srallen @eatyourgreens @rogerhutchings @simoneduca Any of you have a second to give me some feedback on this?

simoneduca:
Sure. 
Regarding the unchecked boxes above. AFAIK, people are not entirely happy with testing the API. Look at [this example](https://github.com/zooniverse/Panoptes-Front-End/blob/05d3315da360c528e26fd26122797c93574044e2/app/pages/lab/apply-for-beta-form.spec.js). I haven't tried myself, but `rewire` is what you need.
This is the correct [json-api-client](https://github.com/zooniverse/json-api-client).
Also I thought this is blocked by the npm issue?

marten:
No, Jim managed to push a commit that works for that.

app/pages/admin/user-settings/delete-user.jsx
eatyourgreens:
```
if (window.confirm(`Do you really want to delete ${this.props.user.login}?`) {
  this.performDelete();
}
```

simoneduca: It's fine by me.

simoneduca: Only thing perhaps, is there any ambiguity between user's `login` and `display_name`?

simoneduca: Ah, I like Jim's idea too, which removes my concerns.

eatyourgreens: Depends how difficult you want to make deleting a user, I guess.

PR #4012, opened 2017-08-30
shrink steel pan featured project image

camallen:
Keep the site light, shrink the feature project media resource we put up. 

I used imageoptim (`brew cask install imageoptim`) to shrink the jpeg as much as i could. The resulting image is ~6x smaller for not much difference in the resulting image (slight lightening).

We should all try and shrink images to something reasonable (dims / quality) to keep the site as light as possible.

Pinging @jelliotartz @beckyrother since this change came in from #4011


eatyourgreens:
#3998 will fix large home page images too.

jelliotartz:
Thanks @camallen! I'll keep image size in mind when reviewing future PRs that include images!

beckyrother:
thanks @eatyourgreens and @camallen !

PR #4019, opened 2017-09-04
Add a simple translations model

eatyourgreens:
Project and workflows front-end for the translations API from zooniverse/Panoptes#2435.

Gets localised strings from the translations API.
Uses Redux to store the resource translations.
Updates some project and task components to use localised strings instead of API resource properties (like `project.display_name`.)

# Review Checklist

- [ ] Does it work on mobile?

## Optional


eatyourgreens:
Some links for testing:
Project default language (English): https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground

Spanish translation: https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?language=es

Italian translation: https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?language=it

Greek translation (doesn't exist): https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?language=el

Note that requesting a translation that doesn't exist simply breaks everything.

eatyourgreens:
Moving the localisation of the interface text into #4027

eatyourgreens:
Rebased off #4027 instead of master.

eatyourgreens:
Rebased to latest master.

eatyourgreens:
Rebased to bring in updates to the survey task summary.

eatyourgreens:
TODO: mock up some translations for the tests. Survey task tests are failing because default translations don't exist.

eatyourgreens:
Following discussions on Slack yesterday, components might be loaded in from zooniverse/Zooniverse-React-Components. In that case components shouldn't have to decide where strings are coming from (a translation resource or the original Panoptes resource.)

In that case, translations might be better handled in a wrapper and passed down to individual components as a prop, to avoid awkward code like ``translations.strings.project.title || this.props.project.display_name` when rendering strings.

app/classifier/tasks/survey/translations.jsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
+import translations from '../../../pages/project/translations';
+
+function SurveyTranslations(props) {
```
eatyourgreens: I find it really hard to follow what this function does 😞 (and I wrote it!)

app/classifier/tasks/survey/translations.jsx
```diff
@@ -0,0 +1,66 @@
...
+  const { task } = props;
+  const translation = {
```
eatyourgreens: Maybe try `const translation = Object.assign({}, task)` here, so that translation labels automatically default to the workflow task labels, and the following code can be greatly simplified.

I can't remember if `Object.assign()` does a deep merge of nested objects.

app/classifier/tasks/survey/translations.jsx
simoneduca: It is really hard. I lost track at the 6th "characteristics". I know I always say this but some destructuring may help?

app/pages/project/translations.js
```diff
@@ -0,0 +1,30 @@
+import apiClient from 'panoptes-client/lib/api-client';
+import counterpart from 'counterpart';
+
+counterpart.setFallbackLocale('en');
+
+
+const translations = {
```
eatyourgreens: This all works, but should probably be replaced with a Redux store, and `load` action, at some point.

app/pages/project/home/project-home.jsx
```diff
@@ -69,7 +73,7 @@ const ProjectHomePage = (props) => {
 
             <div>
               <img role="presentation" src={avatarSrc} />
-              <span>&quot;{props.project.researcher_quote}&quot;</span>
+              <span>&quot;{translations.strings.project.researcher_quote || props.project.researcher_quote}&quot;</span>
```
eatyourgreens: This should be updated to use the translation wrapper pattern that's being used by workflow tasks.

app/classifier/tasks/survey/translations.jsx
eatyourgreens: Any specific thoughts on how destructuring would work here? I had the same thought.

app/classifier/tasks/survey/translations.jsx
eatyourgreens: Specifically, this needs to be done in such a way that [properties on `translation` aren't pointers to the original properties on `task`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Deep_Clone). 🤔

simoneduca: I don't think it does
>The Object.assign() method only copies enumerable and own properties from a source object to a target object.

app/classifier/tasks/survey/translations.jsx
eatyourgreens: I've started to rewrite the first block and I've got this
```javascript
const taskStrings = translations.strings.workflow.tasks;
  const { characteristics, choices, questions } = task;
  Object.keys(characteristics).map((characteristicId) => {
    translation.characteristics[characteristicId] = Object.assign({}, characteristics[characteristicId]);
    if (taskStrings[`${props.taskKey}.characteristics.${characteristicId}.label`]) {
      translation.characteristics[characteristicId].label = taskStrings[`${props.taskKey}.characteristics.${characteristicId}.label`];
    }
    translation.characteristics[characteristicId].values = Object.assign({}, characteristics[characteristicId].values);
    Object.keys(task.characteristics[characteristicId].values).map((valueId) => {
      translation.characteristics[characteristicId].values[valueId] = Object.assign({}, characteristics[characteristicId].values[valueId]);
      if (taskStrings[`${props.taskKey}.characteristics.${characteristicId}.values.${valueId}.label`]) {
      translation.characteristics[characteristicId].values[valueId].label =
        taskStrings[`${props.taskKey}.characteristics.${characteristicId}.values.${valueId}.label`];
      }
    });
  });
```
which is slightly less convoluted, but I'd appreciate any advice you can offer to simplify it further.

app/classifier/tasks/survey/translations.jsx
eatyourgreens: Yeah, the MDN article I linked to says that it creates references, not actual copies. We want to be able to change `translation` without changing `task`.

srallen: lodash has a deep merge method and we have it as a dependency. I think it might be used in the codebase somewhere already.

eatyourgreens: I'll have a look. I've rewritten the object cloning in cc2114356a8aea888a3a2bbe066b1180c7d44908 to make better use of temp variables and destructuring.

Here, we want to overwrite a key's value if a corresponding key exists in the Panoptes translation resource. The translation wrapper basically gives us something that looks like a task (but contains translated text when available) for backwards compatibility with projects that haven't been migrated over to use translations yet.

I'm also thinking that it might allow is to wrap ZRC components, and pass text into them via a translation prop. Code that doesn't use translations yet should still work by passing a task object as the translation (since the structures are identical) `<Task translation={workflow.tasks[taskKey]} />`.

eatyourgreens: I'm also hoping that these adapters can be ripped out, or greatly simplified, once every Panoptes resource has been migrated across to use translations.

srallen: Hm, maybe translations is a good case for a generalized higher order component in ZRC?

eatyourgreens: Possibly. I'm finding each resource needs its own, specific adapter component. There isn't much consistency between tasks, for example.

app/classifier/tasks/survey/translations.jsx
simoneduca: The only thing I can think of is moving the second `.map` block out to its own method. Just to give the eyes a little rest. I'll keep thinking.

simoneduca: Roger wrote this for WGE at some point. I think it does what you want.

```
/**
 * Simple is object check.
 * @param item
 * @returns {boolean}
 */
export function isObject(item) {
  return (item && typeof item === 'object' && !Array.isArray(item) && item !== null);
}

/**
 * Deep merge two objects.
 * @param target
 * @param source
 */
export function mergeDeep(target, source) {
  if (isObject(target) && isObject(source)) {
    Object.keys(source).forEach(key => {
      if (isObject(source[key])) {
        if (!target[key]) Object.assign(target, { [key]: {} });
        mergeDeep(target[key], source[key]);
      } else {
        Object.assign(target, { [key]: source[key] });
      }
    });
  }
  return target;
}
```

eatyourgreens: Deep merge will get rid of all the `Object.assign` calls.

I'll have a think about how to reduce the number of maps. Maybe I'm looking at this the wrong way round and the code should be looping through `taskStrings` in order to assign translation strings to the `translation` object.

app/classifier/tasks/survey/translations.jsx
eatyourgreens: If I can figure out a generalised map from `taskStrings` to `translation` then all of this can be turned into a generic adapter for workflow task translations.

Generic translations for all resources might require some reworking of Panoptes internals. I remember @camallen saying that workflow translations have kept the awkward structure that they have here, with the long string keys for tasks, because changing those keys would break anything that relies on them, like classification exports.

app/classifier/tasks/survey/translations.jsx
```diff
@@ -0,0 +1,64 @@
+import React from 'react';
+import merge from 'lodash/merge';
+import translations from '../../../pages/project/translations';
+
+function SurveyTranslations(props) {
+  const { task } = props;
+  const taskStrings = translations.strings.workflow.tasks;
+  const translation = merge({}, task);
```
eatyourgreens: @simoneduca updated to use deep merge to copy the task structure over to the translation.

The maps are still here to copy the translated strings across, if they exist, but there's probably a better way to do that too.

simoneduca: Of course, I forgot about lodash. Should read comments better. It's made a nice difference 👍

app/classifier/tasks/survey/translations.jsx
```diff
@@ -0,0 +1,41 @@
...
+    while (translationKeys.length) {
+      temp[translationKeys[0]] = (translationKeys.length === 1) ? value : {};
+      temp = temp[translationKeys[0]];
+      translationKeys.shift();
+    }
+    return translationObject;
+  }
+
+  Object.keys(taskStrings).map((translationKey) => {
+    const newTranslation = explodeTranslationKey(translationKey, taskStrings[translationKey]);
```
eatyourgreens: @simoneduca is this easier to understand than the nested maps? I've added a function that explodes the translation task keys out to nested objects, which can be merged into the translation.

eatyourgreens:
In theory, the latest version of the survey task wrapper should act as a generic adapter for all workflow tasks. I haven't tested that, though.

app/classifier/tasks/survey/translations.jsx
simoneduca: Yeah! I mean, there's lots going on, but it's looking pretty neat now.

eatyourgreens: Cheers! I've added tests, which hopefully clarify what this component does.

app/classifier/tasks/translations.spec.js
```diff
@@ -0,0 +1,79 @@
...
+
+describe('Task translation', function () {
+  const wrapper = mount(
+    <TaskTranslation taskKey='survey' task={task} >
+      <StubTask task={task} />
+    </TaskTranslation>
+  );
+  const stubTask = wrapper.find(StubTask);
+
+  it('should not mutate the original task', function () {
```
eatyourgreens: I should check that this actually fails if `TaskTranslation` does mutate the `task` object.

eatyourgreens:
@srallen @rogerhutchings @simoneduca I think I'm mostly happy with this now. It fails gracefully when the translations endpoint isn't available, so merging to master should not cause problems.

My big question is whether to prefer the implementation in 68a653a, which mutates the project resource in order to pass down the translated strings, or the implementation in 5049821, which creates a new object and passes that down as a prop to be rendered by child components.

The second is much safer, IMO, but would require some work on ZRC components so that they accept a new prop and render strings from that prop.

I'm also completely open to suggestions for better names for these things.

app/classifier/tasks/translations.spec.js
eatyourgreens: It didn't fail when `task` was mutated, so I've fixed that.

simoneduca:
Intuitively, after a quick look at the code, I would go with the second implementation. Changing the original object sounds something like it can come back to bite us later on. Also the work on ZRC doesn't seem too much to accommodate that approach.

eatyourgreens:
@simoneduca 00631db updates project components to use a translation prop for rendered strings, instead of the project resource.

simoneduca:
Nice. Also I forgot to ask which parts do you need a hand (re)naming?

eatyourgreens:
Not sure about nice: the staging links now give undefined prop errors.

srallen:
I hope to take a look at this today or maybe tomorrow.

simoneduca:
Where are the errors? Give me a shout, if you think I could help.

eatyourgreens:
@simoneduca staging is fixed with the latest commit.

eatyourgreens:
@srallen Thanks! I think I am pretty much done with this for the moment, pending review comments.

eatyourgreens:
Updated to use a Redux store for the current locale and project and workflow language strings.

app/redux/ducks/translations.js
```diff
@@ -0,0 +1,63 @@
...
+
+counterpart.setFallbackLocale('en');
+// Actions
+const ERROR = 'pfe/translations/ERROR';
+const LOAD = 'pfe/translations/LOAD';
+const SET_LOCALE = 'pfe/translations/SET_LOCALE';
+const SET_STRINGS = 'pfe/translations/SET_STRINGS';
+
+const initialState = {
+  locale: 'en',
```
eatyourgreens: Should this be `counterpart.getLocale()` rather than `en`?

eatyourgreens:
Rebased to include project page changes from #4112.

simoneduca:
Is this just waiting on rebasing?

eatyourgreens:
Rebased (and staged) to fix a conflict in the survey task choice component.

app/redux/ducks/translations.js
simoneduca: I think it should. As with any other hard-coded `'en'` string, right?

eatyourgreens: I think so. I'm not sure if this should be set to the default counterpart locale, or use `getLocale()` in which case initial state can be changed by `counterpart.setLocale()`.

simoneduca: Would setting it to the default counterpart locale be a hurdle for allowing researchers to have projects where the default is different from `en`, like this https://www.zooniverse.org/projects/stefanodf/occhio-alle-penne?

eatyourgreens: Good question. Does this PR update the locale when the project loads? It should change to the project's primary language.

eatyourgreens: My reading of the counterpart docs is that counterpart's default locale is always English, regardless of the browser's locale.

simoneduca: [I think you're right](https://github.com/martinandert/react-translate-component#usage)
>Whenever the user selects a different locale from the drop-down, we correspondingly set the new drop-down's value as locale in the Counterpart library, which in turn triggers an event that our (soon to be integrated) Translate component listens to. As initially active value for the select element we specify Counterpart's current locale ("en" by default).

eatyourgreens: My comment was wrong: when a project loads, we should respect the user's preferred locale, currently set via a query param. If that's not present, use the project's locale (which is always `en` at the moment, regardless of the actual text stored with the resource.)

eatyourgreens: Either way, I don't think that has anything to do with the initial Redux state, because nothing has loaded when that is set. I think.

app/pages/project/index.cjsx
```diff
@@ -125,7 +133,8 @@ ProjectPageController = React.createClass
             awaitProjectAvatar,
             awaitProjectCompleteness,
             awaitProjectRoles,
-            awaitPreferences
+            awaitPreferences,
+            this.props.actions.translations.load('project', project.id, this.props.translations.locale)
```
simoneduca: Should `this.` be `@`

simoneduca:
@eatyourgreens I've had another look through the code, which is fine and pretty clear. Tests seem to fail where they're supposed to and I can't see anything wrong with the Redux side of things.

The only thing I noticed is that [redux-logger](https://github.com/evgenyrodionov/redux-logger) is missing. It's one of the most helpful redux tool, so perhaps we could think of adding it in. Also I don't think anybody uses the [redux-devtools-extension](https://github.com/zooniverse/Panoptes-Front-End/blob/master/package.json#L38), so I think we should remove that.

With the view of documenting the process of adding translations to other components and translation in general, the steps involved so far are:
1. Write a generic (possibly HOC) component whose main function is passing the translation object as prop from the Redux store to all the children components that need it.
2. Wrap the components supposed to support translations with the generic component, like `TaskTranslations` in this PR.

I've probably missed a step, please let me know.

Note: for now creating and editing a translation is taken care by [Pandora](https://github.com/zooniverse/pandora). 

LGTM, but I'd like another pair of eyes to look at it before merging.

app/pages/project/index.cjsx
srallen: I'm not too picky on adhering to coffeescript conventions or shorthand these days, as long as it doesn't break compiling, since eventually this should be rewritten in javascript anyway.

app/pages/project/home/index.jsx
```diff
@@ -132,5 +133,12 @@ ProjectHomeContainer.propTypes = {
     researcher_quote: React.PropTypes.string
   }).isRequired,
   projectIsComplete: React.PropTypes.bool.isRequired,
-  splits: React.PropTypes.object
+  splits: React.PropTypes.object,
+  translation: React.PropTypes.shape({
```
srallen: This PropType is failing with `translation` as `undefined` when testing with this project using the English default: https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground

srallen:
Just so I have an idea of what I should be seeing, with the test projects I see console warnings like `ES workflow 2824 translation fetch error: Could not find translation with id=''`. Is this expected?

srallen:
Oh one last comment I forgot, this should get a rebase since the redux devtools dependency is removed from master.

simoneduca:
@srallen those fetch errors are expected because the translation resource doesn't exists. I share your uncertainty on how to check that this works as it should, without a full translation available. Perhaps @eatyourgreens could list the parts that are supposed to be translated. I'd be happy to translate into Italian all the necessary bits so that we can see the PR at work.

eatyourgreens:
https://translation-api.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground has English, Italian and Spanish translations available for the project resource.

https://translation-api.pfe-preview.zooniverse.org/projects/srallen086/survey-test has English translations for the project and default workflow.

Anything else should warn about translations not being available, I think, but that shouldn't break anything. If it does, there's something wrong in this PR.

@simoneduca we're waiting on zooniverse/pandora#46 and https://github.com/zooniverse/pandora/pull/46 before you can add or edit translations. At the moment, the translations API is effectively read-only.

eatyourgreens:
I've added tests for those cases where resource translations don't exist in Panoptes.

app/pages/project/home/index.jsx
eatyourgreens: Not sure what's going on here. There's a default, empty translation set as the initial state in the translations ducks file.

simoneduca: Is this the same issue we had before with some other props? They seemed to error before rendering...

eatyourgreens: Aha! This warning is being thrown because there's no translation prop defined in `router.cjsx`. I'm not sure how to handle that.

eatyourgreens: @simoneduca Yes! If you look at the stack trace, the calling context is the `RouterContext` component.

simoneduca: Silly but can we not define the missing props in `router.cjsx`?

eatyourgreens: @simoneduca How do you do that?

simoneduca: Mm, that [doesn't seem to have a straight answer](https://github.com/ReactTraining/react-router/issues/1857#issuecomment-180879045).

eatyourgreens: @simoneduca I know. I thought maybe you'd come across a way to pass default props in the routes definitions.

eatyourgreens: Looks like required props are not recommended on top-level components.

app/pages/project/project-translation.spec.jsx
```diff
@@ -0,0 +1,96 @@
...
+    project: {
+      title: 'A translated title',
+      display_name: 'A translated display name',
+      introduction: 'A translated introduction'
+    }
+  }
+};
+
+const project = {
+  id: 12345,
```
srallen: Not a blocker, but for accuracy with these mocks, the ids should be strings.

eatyourgreens:
@srallen Thanks for taking a look. I will have a look to see why the Shortcut task is failing like that.

eatyourgreens:
@srallen I can see the required prop warning, but the `translation` prop is defined in the Shortcut task, as far as I can tell, both in the component constructor and in `componentWillMount`. I'm at a loss as to how to fix that warning.

srallen:
@eatyourgreens ok, it's not a blocker since it didn't seem to break anything, but it's nice to remove warning messages when we can. I'll do one last look over today.

srallen:
I think this will ready to merge after a final rebase.

PR #4021, opened 2017-09-05
Adds a README for feedback

simoneduca:
Fixes # .

Describe your changes.
Adds basic instructions on how to use feedback. As feedback is an experimental feature, I placed the readme in the feedback folder. In the future, we may consider moving it somewhere on the actual site.

### DIFF
--- a/app/classifier/feedback/README.md
+++ b/app/classifier/feedback/README.md
@@ -2,11 +2,11 @@
 
       Feedback is enabled per-task - there's a checkbox under the answer editing box that toggles it on and off.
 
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Looks good to me. Is it worth adding instructions for the point tool feedback too, or leave that until later?

If the images are edited/deleted on github, then they'll change (or vanish) in this readme. Should we move them into PFE?

simoneduca:
I'm not sure how the point tool feedback works, so I'd leave it for now. I've added the images but they're not showing yet in the README.

PR #4027, opened 2017-09-07
Localise project pages

eatyourgreens:
Extract all the English strings, in project UI, to counterpart.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://localise-projects.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/locales/en.js
+++ b/app/locales/en.js
@@ -1,0 +1,40 @@
+export default {
+  loading: '(Loading)',
+  classifier: {
+    back: 'Back',
+    backButtonWarning: 'Going back will clear your work for the current task.',
+    done: 'Done',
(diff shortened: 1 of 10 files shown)


eatyourgreens:
I don't think this has caught everything, but it should be easy to add keys to the locale files as we come across strings in the code.

TODO: load in translations when needed, as per #3937, rather than requiring in all the possible locales at once.

app/locales/en.js
```diff
@@ -0,0 +1,40 @@
+export default {
```
simoneduca: Perhaps not at this stage, but is it worth moving language files to their own repo? https://github.com/zooniverse/translate-pfe

PR #4030, opened 2017-09-08
Introduces functionality for galaxy modelling custom project

tingard:
This pull request introduces code allowing in-browser model creation, specifically for a galaxy modelling project. A user can draw shapes on a subject (i.e. galaxy components like a disk, bar and spiral arms), tweak parameters using the slider task and are shown the difference between the provided subject and their model. 

# Describe your changes.
Creates a hidden img and canvas, and allows parsing of classification into rendering commands which are rendered to the canvas using webgl and regl. The img is used to load the subject into a regl texture to be fed to the rendering instance.

## File introductions:
- `/app/components/model-render.jsx`
  - Creates the `canvas` and `img` elements. Creates model instance and triggers model updating when annotation/subject changes
- `/app/components/modelling/index.js`
  - Determines which model is to be used for this workflow, introduced as a basic means of futureproofing
- `/app/components/modelling/baseModel.js`
  - provides the `baseModel` class, to be extended with custom models for projects
- `/app/components/modelling/galaxyModel.js`
  - extends `baseModel`, providing a function for parsing an annotation to rendering commands and an internal "model", which contains default values and regl rendering functions.
- `/app/components/modelling/galaxyRegls.js`
  - provides galaxy modelling regl rendering functions
- `/app/components/modelling/postProcessingRegl.js`
  - provides post processing commands to be run after the annotation is complete (should be moved to be associated with the galaxy modelling project specifically?) . Contains PSF convolution and difference calculation, both of which are galaxy modelling specific.

# Review Checklist

  - Will not work properly in safari as crossorigin not supported (to the best of my knowledge). Does not interfere with other tasks (requires workflow.metadata.modelling
- [ ] Does it work on mobile?
  - Might work technically, not a pleasant user experience

## Optional


app/classifier/classifier.jsx
```diff
@@ -239,7 +246,15 @@ class Classifier extends React.Component {
           onChange={this.handleAnnotationChange.bind(this, currentClassification)}
           playIterations={this.props.workflow.configuration.playIterations}
         />
-
+        {this.props.workflow.configuration.metadata &&
+          this.props.workflow.configuration.metadata.type === 'modelling' ?
+            <ModelCanvas
```
eatyourgreens: This is an instance of the `ModelRenderer` component, so the name is a bit confusing.

app/classifier/classifier.jsx
```diff
@@ -239,7 +246,15 @@ class Classifier extends React.Component {
           onChange={this.handleAnnotationChange.bind(this, currentClassification)}
           playIterations={this.props.workflow.configuration.playIterations}
         />
-
+        {this.props.workflow.configuration.metadata &&
```
eatyourgreens: We've re-written the classifier to get rid of project-specific, or workflow-specific, hacks. This conditional makes me nervous that those are creeping back in again.

app/classifier/tasks/slider/index.jsx
```diff
@@ -49,13 +49,13 @@ class SliderTask extends React.Component {
   }
 
   componentDidMount() {
-    const newAnnotation = Object.assign({}, this.props.annotation, { value: this.props.task.defaultValue });
+    const newAnnotation = Object.assign({}, this.props.annotation, { value: this.state.value });
```
eatyourgreens: See #4004 which removes the component lifecycle methods from the slider task.

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+  }
+}
+
+class ModelRenderer extends React.Component {
+  constructor(props) {
+    super(props);
+    this.getDifference = this.getDifference.bind(this);
+    this.setTexture = this.setTexture.bind(this);
+
+    this.imgUrl = Object.values(this.props.subject.locations[0])[0];
```
eatyourgreens: Why is this setting an instance property instead of component state?

eatyourgreens: I don't think `this.props.subject.locations[0])[0]` is guaranteed to be the image you want, or even an image, for subjects in general.

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+    }
+  }
+
+  // don't want to re-render whenever props update
+  shouldComponentUpdate() {
+    if (this.imgUrl !== Object.values(this.props.subject.locations[0])[0]) {
+      // we have a new subject. Refresh the model calculator
+      const im0 = this.props.subject.locations[0];
+      if (this.props.subject.locations.length < 2) {
+        this.props.subject.locations[1] = Object.assign({}, im0);
```
eatyourgreens: Should the subject resource, from the API, be modified directly like this?

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+    // component has mounted, initialise the regl canvases
+    this.model = new Model(this.canvas, this.props.subject.metadata);
+    // if the image has loaded
+    if (this.imageLoaded) {
+      // provide the differece caluculator with the image
+      this.model.setBaseTexture(this.im);
+      this.model.update(this.props.classification.annotations);
+    } else {
+      // tell the setTexture function that it should update the difference calc
+      this.imageLoaded = true;
```
eatyourgreens: Why isn't this stored in component state?

app/classifier/classifier.jsx
tingard: Good point, how would you suggest I phrased this so that the model renderer is only loaded for modelling workflows?

app/components/model-renderer.jsx
tingard: This is a hacky part - need to send the calculated difference between subject + model and send to a viewer. Would be better if a dedicated webgl viewer was introduced?

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+    if (this.props.subject.locations.length < 2) {
+      const im0 = this.props.subject.locations[0];
+      this.props.subject.locations[1] = Object.assign({}, im0);
+    }
+  }
+  componentDidMount() {
+    if (!webGLCompatibilityTest()) {
+      // TODO: is there a fallback option?
+      /* eslint-disable no-console */
+      console.error('WebGL is not available, aborting');
```
CKrawczyk: It might be useful to render a message to the so the user realizes that the project will not work in their browser.  Something like: "It looks like your browser will not work for this project, please use XXX or YYY"

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+      /* eslint-enable no-console */
+      return;
+    }
+    // component has mounted, initialise the regl canvases
+    this.model = new Model(this.canvas, this.props.subject.metadata);
+    // if the image has loaded
+    if (this.imageLoaded) {
```
CKrawczyk: This block is confusing.  The `componentDidMount` will only be called once in the life cycle, and `imageLoaded` can only be set to `true` inside of it.  So `imageLoaded` can never be `true` before `componentDidMount` is called.  When would the first part of this `if` block ever be called?

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+    // check if a model has been assigned
+    if (this.model !== null) {
+      // if so, update the render function list to the new annotation
+      window.requestAnimationFrame(this.getDifference);
+    }
+  }
+
+  // don't want to re-render whenever props update
+  shouldComponentUpdate() {
+    if (this.imgUrl !== Object.values(this.props.subject.locations[0])[0]) {
```
CKrawczyk: If `imgUrl` is in state you can check against `nextProps` and `nextState` directly instead of re-fetching the URL from the subject.

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,127 @@
...
+      /* eslint-enable no-console */
+      return;
+    }
+    // component has mounted, initialise the regl canvases
+    this.model = new Model(this.canvas, this.props.subject.metadata);
+    // if the image has loaded
+    if (this.imageLoaded) {
+      // provide the differece caluculator with the image
+      this.model.setBaseTexture(this.im);
+      this.model.update(this.props.classification.annotations);
```
CKrawczyk: Couldn't you call `this.setTexture()` since it is extracted as its own method?

app/classifier/classifier.jsx
eatyourgreens: Maybe incorporate the `ModelCanvas` into the subject viewer as a new type of frame viewer?

eatyourgreens: That would remove the need to check the workflow at all, since the model would become just another frame in the subject locations array.

app/components/modelling/galaxyModel.js
```diff
@@ -0,0 +1,144 @@
...
+    ];
+  }
+  parse(annotation) {
+    // TODO: make this explicitly related to the galaxy rendering model.
+    // TODO: link spiral arm to disk? Link centres?
+    // we have disk, then bulge then bar
+    const renderFunctions = [];
+    const annotationLength = annotation.length;
+    let disk = null;
+    if (annotationLength > 0 && annotation[0].value[0].value.length > 0) {
```
CKrawczyk: Having the `annotation` index for each of the model components does not seem like a good idea.  If the order of the workflow questions change you should not have to change the drawing code.

We should have a way to auto-detect this from the workflow structure.

app/components/modelling/galaxyModel.js
```diff
@@ -0,0 +1,144 @@
...
+        const spiralArm = Object.assign(
+          { name: this.model[3].name },
+          this.model[3].default,
+          {
+            disk,
+            i0: parseFloat(annotation[3].value[0].value[i].details[0].value),
+            spread: parseFloat(annotation[3].value[0].value[i].details[1].value),
+            falloff: parseFloat(annotation[3].value[1].value),
+            // spread: parseFloat(annotation[3].value[1].value[0].value),
+            // i0: parseFloat(annotation[3].value[1].value[1].value),
```
CKrawczyk: Remove old commented code.

app/components/modelling/baseModel.js
```diff
@@ -0,0 +1,139 @@
...
+  kill() {
+    this.regl.destroy();
+  }
+  setBaseTexture(image) {
+    try {
+      this.baseTexture = this.regl.texture(image);
+    } catch (err) {
+      if (err.name === 'SecurityError') {
+        // TODO: shouldn't really be using alert here. Does Zooniverse have a modal?
+        alert('CORS error encountered, please make sure you are using Google Chrome and try refreshing the page');
```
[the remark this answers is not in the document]
eatyourgreens: PFE has its own alert box in `/app/lib/alert`. An example of usage is the [task help](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/generic.jsx#L28-L40).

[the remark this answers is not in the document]
srallen: Grommet's been added in the admin UI as a test usage, but it's still a ways off before I'd say we'd really start implementing it elsewhere on PFE. I'd suggest using the alert component the lib folder for this use case as well.

tingard:
Made some changes, but some things still left to do:

- renamed ModelCanvas to ModelRenderer
- Moved checks for worflow.metadata.modelling into a wrapper component for ModelRenderer and ModelScore
- Apparently made some changes to slider task, which will **I need to check and rebase carefully from master**
- Started work on new frameviewer called `ModelViewer` (probably should be `CanvasViewer`?)
- moved things to state where suggested
- used the provided alert component
- fixed a persistent CORS error apparently being caused by chrome caching subject images but not the Acces-Control data.
- Made changes inside ModelScore to remove unnecessary math code (as per Coleman's request)
- changed default colour of difference display

TODO:
- [ ] Frameviewer
- [ ] Rebase again
- [ ] Remove duplicate modelscore files

eatyourgreens:
Looks like your Slider task changes revert some changes that were made to fix bugs in Steelpan Vibrations. Maybe just overwrite the slider task component with the latest version from master, since it's been rewritten as a stateless function.

tingard:
@eatyourgreens I'm planning on rebasing from master today so hopefully that'll be picked up as a conflict?

eatyourgreens:
GitHub's flagging it as a conflict so git should flag it locally.

eatyourgreens:
There might even be a git flag to ignore locally changes and overwrite conflicts from master. I don't think you've made any major changes to the slider.

tingard:
Rebase done, I'm a twit and missed one of the conflicts so there's an extra commit that might want to be squashed in somewhere for neatness (unless PRs are squashed together anyway)?

eatyourgreens:
I don't think we need `yarn.lock`. Anyway, Travis is complaining about yarn when it tries to run the tests.

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,148 @@
...
+    if (this.state.imgUrl !== Object.values(this.props.subject.locations[0])[0]) {
+      // we have a new subject. Refresh the model calculator
+      const im0 = this.props.subject.locations[0];
+      if (this.props.subject.locations.length < 2) {
+        this.props.subject.locations[1] = Object.assign({}, im0);
+      }
+      this.model.kill();
+      const metadata = this.props.subject.metadata;
+      this.model = new Model(this.canvas, metadata);
+      this.setState({ imgUrl: Object.values(im0)[0] });
```
CKrawczyk: `shouldComponentUpdate` should only contain logic around returning `true` or `false`. This other code should go in `componentWillUpdate` (code called *before* an update happens).

app/components/modelling/baseModel.jsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+  kill() {
+    this.regl.destroy();
+  }
+  setBaseTexture(image) {
+    try {
+      this.baseTexture = this.regl.texture(image);
+    } catch (err) {
+      if (err.name === 'SecurityError') {
+        // TODO: shouldn't really be using alert here. Does Zooniverse have a modal?
```
CKrawczyk: This comment is out of date.

app/components/modelling/baseModel.jsx
```diff
@@ -0,0 +1,147 @@
...
+        alert(
+          (resolve, reject) =>
+            (
+              <div className="content-container">
+                <Markdown className="classification-task-help">
+                  CORS error encountered, please make sure you are using Google Chrome and try refreshing the page
```
CKrawczyk: Is this CORS error still happening, I thought you fixed it?  If it is fixed this entire `alert` can go away.

tingard: CORS will still trigger on browsers that don't support `crossorigin="anonymous"` on image tags (i.e. most recent safari, I think). The bug I spotted and fixed was displaying as CORS but actually something else that I hadn't been able to catch.

tingard:
@eatyourgreens any updates?

eatyourgreens:
I have some small comments. I'll add them later today.

app/classifier/classifier.jsx
```diff
@@ -147,14 +150,18 @@ class Classifier extends React.Component {
     this.props.classification.update(changes);
   }
 
+  handleModelScoreUpdate(newScore) {
+    this.props.workflow.configuration.metadata.modelScore = newScore;
```
eatyourgreens: This looks dangerous to me. Volunteers shouldn't be modifying the workflow configuration.

If the intention here is to update the state of the score, either update component state or update the score as part of the task annotation.

app/classifier/classifier.jsx
```diff
@@ -267,7 +279,7 @@ class Classifier extends React.Component {
               toggleExpertClassification={this.toggleExpertClassification}
             />
           }
-
+          <ModelScore workflow={this.props.workflow} />
```
eatyourgreens: Does this component need the entire workflow object if all it does is render a score?

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,144 @@
...
+          crossOrigin="anonymous"
+          hidden={true}
+        />
+      </div>
+    );
+  }
+}
+/* eslint-disable react/forbid-prop-types */
+ModelRenderer.propTypes = {
+  classification: React.PropTypes.object,
```
eatyourgreens: Use `React.PropTypes.shape({})` here and define the properties of `classification` that are used in this component.

app/components/model-renderer.jsx
```diff
@@ -0,0 +1,144 @@
...
+  subject: React.PropTypes.object,
```
eatyourgreens: Use `React.PropTypes.shape({})` here and define the properties of `subject` that are used in this component.

app/components/modelling/galaxyScore.jsx
```diff
@@ -0,0 +1,60 @@
+import React from 'react';
+
+// grabs the model score (calculated in modelCanvas) and displays it with a
+// progress bar
+const ModelScore = (props) => {
+  if (props.workflow && props.workflow.configuration &&
+    props.workflow.configuration.metadata &&
+    props.workflow.configuration.metadata.modelScore) {
+    let s = props.workflow.configuration.metadata.modelScore;
```
eatyourgreens: This component just renders a score, so pass the score directly as a prop and bypass the mess of conditionals here.

app/components/modelling/galaxyScore.jsx
```diff
@@ -0,0 +1,60 @@
...
+      s = s.toFixed(2);
+    } else if (s < 95) {
+      s = (s).toFixed(3);
+    } else {
+      s = (s).toFixed(4);
+    }
+    return (
+      <div className="answer undefined">
+        <div className="answer-button">
+          <div className="answer-button-label">Score: {s}</div>
```
eatyourgreens: Why is this using a `button` class if it isn't a button? Will this give the visual impression that it is a clickable element?

app/components/modelling/galaxyScore.jsx
```diff
@@ -0,0 +1,60 @@
...
+    let s = props.workflow.configuration.metadata.modelScore;
+    if (s < 80) {
+      s = s.toFixed(2);
+    } else if (s < 95) {
+      s = (s).toFixed(3);
+    } else {
+      s = (s).toFixed(4);
+    }
+    return (
+      <div className="answer undefined">
```
eatyourgreens: What does `undefined` do here?

app/classifier/classifier.jsx
tingard: I tried updating component state, but it was a huge performance hit (not sure why) so not feasible. 
Where in the annotation would you recommend? Would `classification.metadata` be safe enough? I don't know what gets sent back to the api.

app/classifier/classifier.jsx
tingard: the workflow was being passed as it contains information used by the wrapper inside ModelScore to check if it's a "modelling workflow".

eatyourgreens: Maybe pass that in as a `modellingEnabled` prop instead? Similar to how pan-zoom is handled: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/frame-viewer.jsx#L59-L65

It would keep all the complicated conditional code in one place by checking once and passing the boolean result down. I think feedback does something similar.

app/classifier/classifier.jsx
eatyourgreens: The annotation will be in `classification.annotations` and I think the annotation for the current task is passed down as a prop, so maybe `this.props.annotation`. The actual structure of the annotation varies from task to task, so depends on what sort of task modelling is. I'm guessing maybe some sort of drawing task?

app/components/modelling/galaxyScore.jsx
tingard: I have no idea where these classes materialised from. Don't think they have any impact whatsoever

app/classifier/classifier.jsx
tingard: The only workflow so far is series of combo tasks with drawing and sliders, but I could potentially see just a drawing task being used.
It's the kind of thing that sits perfectly on state, I'll try and debug the performance problems. I think it's just triggering a huge number of re-renders.

eatyourgreens: This sounds familiar. @rogerhutchings might be helpful here because he ran into a similar problem with the freehand drawing tool triggering hundreds, if not thousands, of re-renders when you drew a line. I think his solution was to buffer changes until drawing was complete, then trigger an annotation change, which in turn triggers a re-render of the classifier.

srallen: React's `shouldComponentUpdate` is helpful in stopping unnecessary re-renders. Not sure if it would help in this case.

eatyourgreens: I know @rogerhutchings is a big fan of that method, so that might be what he used.

eatyourgreens:
Looks good to me. There's a couple of places where it relies on the image URL being in `subject.locations[0])[0]`, which might be dangerous in general for projects, but is probably ok for the galaxy project.

I'm happy to merge this next week, after a final check on staging to see if it has any side-effects for existing projects.

tingard:
Brilliant, thanks @eatyourgreens 
The only thing I can think of would be to check if 'image/jpeg' or 'image/png' is set, and use it if so? And if not don't render.
I'd also like to change the isModelling from checking for a "workflow type" to a boolean flag, but that won't work for the current workflow and I've made changes in the editors that aren't in my local workflow JSON, can I download the JSON using the python client?

eatyourgreens:
Setting up the subject locations (after the subject loads) as an array something like:
```
[{ 'image/jpeg': image_URL }, { 'image/webGL': data_URL }]
```
might work. There's no requirement in PFE, as far as I can tell, for the key to be a valid MIME type, so we can get away with inventing our own format and type here. Then the image viewer code would just need to be extended to use a different renderer when the passed type is `webGL`.

That might be worth experimenting with in a separate PR, to see if it does work.

PR #4031, opened 2017-09-08
Add recent classifications to projects

eatyourgreens:
Add a /recents route to projects 
eg. https://recents.pfe-preview.zooniverse.org/projects/sandiegozooglobal/wildwatch-kenya/recents?env=production
Add a basic component to render subjects from the user recents API endpoint.

Allows volunteers to see their recently classified subjects and link through to Talk.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://recents.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/locales/en.js
+++ b/app/locales/en.js
@@ -18,7 +18,8 @@ export default {
       about: 'About',
       classify: 'Classify',
       talk: 'Talk',
-      collections: 'Collect'
+      collections: 'Collect',
+      recents: 'Recents'
     },
     home: {
       researcher: 'Words from the researcher',
--- a/app/locales/en.js
+++ b/app/locales/en.js
@@ -6,6 +6,7 @@ export default {
     done: 'Done',
     doneAndTalk: 'Done & Talk',
     next: 'Next',
+    recents: 'Your recent classifications',
     talk: 'Talk',
     tutorialButton: 'Show the project tutorial',
     miniCourseButton: 'Restart the project mini-course'
--- a/app/locales/it.js
+++ b/app/locales/it.js
@@ -6,6 +6,7 @@ export default {
     done: 'Done',
     doneAndTalk: 'Done & Talk',
     next: 'Next',
+    recents: 'Your recent classifications',
     talk: 'Talk',
     tutorialButton: 'Show the project tutorial',
     miniCourseButton: 'Restart the project mini-course'
--- a/app/locales/it.js
+++ b/app/locales/it.js
@@ -18,7 +18,8 @@ export default {
       about: 'A proposito',
       classify: 'Classifica',
       talk: 'Forum',
-      collections: 'Colleziona'
+      collections: 'Colleziona',
+      recents: 'Recents'
     },
     home: {
       researcher: 'Words from the researcher',
--- a/app/pages/profile/recents.jsx
+++ b/app/pages/profile/recents.jsx
@@ -26,30 +26,32 @@ class Recents extends React.Component {
   render() {
     const { project } = this.props;
     return (
-      <div className="secondary-page has-project-context">
+      <div className="collections-page secondary-page has-project-context">
         <div className="hero collections-hero">
           <div className="hero-container">
             <Translate content="classifier.recents" component="h1" />
           </div>
         </div>
-        <ul className="collections-card-list">
-          {this.state.recents.map((recent) => {
-            const { type, format, src } = getSubjectLocation(recent);
-            return (
-              <li key={recent.id} className="collection-card">
-                <Link to={`/projects/${project.slug}/talk/subjects/${recent.links.subject}`}>
-                  <Thumbnail
-                    alt={`Subject ${recent.links.subject}`}
-                    src={src}
-                    type={type}
-                    format={format}
-                    height={250}
-                  />
-                </Link>
-              </li>
-            );
-          })}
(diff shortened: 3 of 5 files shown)


trouille:
@beckyrother Tagging you here. Good to take a look.

camallen:
@eatyourgreens i get an error when not logged in and page dies... otherwise it looks awesome.
screen shot 2017-09-12 at 13 51 42

eatyourgreens:
@camallen Oops. The link in the project navigation doesn't show up for anonymous users, so I didn't put any checks in the view itself.

I'm not crazy about the styling of the list, but it does help everyone on Talk who asks how they can go back and comment on something they just classified.

beckyrother:
Yessss this is awesome. Can we remove the corner radius though? Looking to phase that out as we make updates.

eatyourgreens:
@beckyrother sure. I was just lazy and copied the collection cards CSS classes.

simoneduca:
Does this only need rebasing and removing the border radius?

eatyourgreens:
Rebased and fixed up to bring in the new locale files.

eatyourgreens:
The styling isn't great, but I think it looks slightly better if it copies the CSS classes from the collection listing page.

Screenshot of the recents page, showing square subject thumbnails roughly centred on a light grey background.

simoneduca:
This is what I see with no classification made on a project. Is it intended?
screen shot 2017-09-18 at 13 41 28

eatyourgreens:
@simoneduca Probably not. Maybe I missed a CSS class in bb66ee2?

eatyourgreens:
Quick fix seems to be to hide the list if `recents.length` is 0.

eatyourgreens:
Rebased to fix a merge conflict in the routes config.

simoneduca:
One more rebase needed I think.

eatyourgreens:
Rebase (and staged) to fix a conflict.

beckyrother:
Hi sorry looking at this again – sorry if it's too late to add this, but would it be possible to add the same controls as on the collections images? That way they can quickly add a subject to favorites/collections if they meant to and forgot.

(like this)
beautiful_ _zooniverse

eatyourgreens:
Depends how the subject viewer for collections works. Recents objects have the subject ID, but are not instances of Subject so probably can't be passed directly to the subject viewer, which has all the nifty controls.

You raise a very good point, though, that representations of subjects should be the same wherever they appear in the site. It would make sense to update that subject viewer component so that it can take a `recents` resource as a prop, in place of a `subjects` resource.

camallen:
@eatyourgreens what does the recents resource need to talk like a subject resource? I assume the linked subject_id (for adding to collections) as well as the metadata? Would side loading the subjects resource make more sense here and using that combined response instead?

eatyourgreens:
@camallen good question. I haven't looked at those subject buttons to see what they use. I suspect they require a Subject resource to be passed in, but only actually use the ID, locations and metadata.

Sideloading the subject makes sense. What sort of impact would it have on performance if I display, say, the 20 most recently classified subjects?

eatyourgreens:
@camallen Looks like collections [pass a subject to a SubjectViewer](/zooniverse/Panoptes-Front-End/blob/master/app/collections/show-list.cjsx#L81) but any object with the expected shape might work. I can have a look when I've got some time to look at this PR again.

camallen:
After checking the restpack serializer code it looks like the it would be a an extra 1 db query to fetch the joined subject instances via id (index scan) and then it would construct the extra serializer responses for the subject. This response construction looks like it'll cause multiple location queries for each subject we side-load, so not ideal. 

I suggest we test this approach and if it's not ideal then we either fix the side load code or replace the make the recent model quack like a subject by adding metadata and subject_id as well to the serialized response.

eatyourgreens:
@camallen I'm trying out passing a fake subject resource to the subject viewer and I think that would work. I'll try to have something up later.

camallen:
@eatyourgreens let me know what you need. I can't wait to let people see recent images they've been working on.

eatyourgreens:
I've just pushed, and staged, an update that copies @mcbouslog's code from the collections pages. I fake a subject by passing in an object with the subject ID and locations. Seems to work.
https://recents.pfe-preview.zooniverse.org/projects/sandiegozooglobal/wildwatch-kenya/recents?env=production

eatyourgreens:
With the Subject Viewer added, is this good to go?

beckyrother:
lgtm! thank you for making this change!

camallen:
@eatyourgreens we could pretty easily serialize the subject metadata with the recent response. That shouldn't block this getting out though, just noting the api can provide this if / when people complain.

eatyourgreens:
Updated so it only shows one preview frame for subjects with multiple frames.
https://recents.pfe-preview.zooniverse.org/projects/rainforestexpeditions/amazoncam-tambopata/recents?env=production

eatyourgreens:
cheering_minions

eatyourgreens:
Someone should let @yshish know that recent subjects are back!

yshish:
@eatyourgreens Thank you, I'm watching you working on this all the time :D (Hope my thank-you-box will arrive soon!) Can't wait to get on Zooniverse to see it working!

PR #4032, opened 2017-09-11
Localise survey task

eatyourgreens:
Depends on #4027 

Rewrites the survey task components in JavaScript and extracts English strings to the locale files.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://localise-survey-task.pfe-preview.zooniverse.org/

## Optional


app/classifier/tasks/survey/chooser.jsx
```diff
@@ -0,0 +1,290 @@
...
+      .map((button) => {
+        const choiceID = button.getAttribute('data-choiceID');
+        const index = this.props.task.choicesOrder.indexOf(choiceID);
+        newChoiceButtons[index] = button;
+      });
+    this.choiceButtons = newChoiceButtons.filter(Boolean);
+  }
+
+  whatSizeThumbnails({ length }) {
+    if (length <= 5) {
```
simoneduca: This chain of ifs seems insane to me. Could we use `while` perhaps?

eatyourgreens: Isn't `while` used for loops?

eatyourgreens: There's a ton of code in the survey task that made me scratch my head while I was converting it.

simoneduca: Indeed, I thought that was a loop. Read too quickly, sorry.
When I've translated components before, I tried to improve the code if possible. You're right though, looks like quite a task in this case.

simoneduca: `switch` is what I wanted to say. Would that be a bit neater?

eatyourgreens: I went for a straight rewrite to avoid changing any existing functionality by accident. In the absence of tests, it's hard to tell if a change has broken a function.

simoneduca: I appreciate that. I'll review this as is and we should rewrite the survey task in a separate PR.

eatyourgreens:
This didn't merge to master. My bad, I based it off the `localise-projects` branch for #4027.

PR #4036, opened 2017-09-12
Reset PanZoom on subject change

### LINKED ISSUE #4034: Pan-and-zoom resets when switching or playing through frames
## Expected behavior
I'm not sure this is expected from a code perspective, but if I've got a multi-frame subject and I use pan and zoom to zoom in on one frame, I expect that switching to the other frame will keep the same zoom/pan.

## Current behavior
Playing through frames or manually switching frames resets the pan-and-zoom tool to the default. It doesn't preserve the zoom in the original frame; if I zoom in frame 1, then switch to frame 2 (where the zoom level resets), then switch back to frame 1, the zoom doesn't go back to the modified version.

## Steps to replicate
I'm seeing this in the latest [PRN project](https://www.zooniverse.org/projects/vrooje/planetary-response-network-and-rescue-global-caribbean-storms-2017/classify).

## Additional information
I'm not sure this is quite the same as #3465 ? Even if the zoom was never intended to be subject-wide, it seems like the zoom level not being remembered from frame to frame is a bug.

- **Operating system:**
OS X 10.11.6

- **Browser:**
Chrome Version 59.0.3071.115 (Official Build) (64-bit)

eatyourgreens:
Fixes #4034.

Describe your changes.
Removes React.createClass from `PanZoom`.
Deletes the old `componentWillUpdate` method.
Resets the zoom level and rotation if subject or frame width change.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://pan-zoom.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -1,29 +1,18 @@
 /* eslint no-unused-expressions: ["error", { "allowTernary": true }] */
 import React from 'react';
 
-const PanZoom = React.createClass({
-
-  propTypes: {
-    children: React.PropTypes.node,
-    enabled: React.PropTypes.bool,
-    frameDimensions: React.PropTypes.shape({
-      height: React.PropTypes.number,
-      width: React.PropTypes.number
-    })
-  },
-
-  getDefaultProps() {
-    return {
-      enabled: false,
-      frameDimensions: {
-        height: 0,
-        width: 0
-      }
-    };
-  },
-
-  getInitialState() {
-    return {
+class PanZoom extends React.Component {
+  constructor() {
+    super();
+    this.frameKeyPan = this.frameKeyPan.bind(this);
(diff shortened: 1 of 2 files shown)


app/components/pan-zoom.jsx
```diff
@@ -45,115 +35,41 @@ const PanZoom = React.createClass({
...
-    this.setState({
-      viewBoxDimensions: {
-        width: newProps.frameDimensions.width,
-        height: newProps.frameDimensions.height,
-        x: 0,
-        y: 0
-      }
-    });
-  },
+  componentDidUpdate(oldProps) {
```
eatyourgreens: `zoomReset` triggers an update and render, so `componentWillUpdate` might be safer here to avoid recursive loops.

Issue #2970, opened 2016-09-05
Organizations: public landing page


trouille:
## Expected behavior

A landing page from which a volunteer will access all ~~workflows~~ projects associated with the ~~Federated Project~~ Organization (e.g., Urban Wildlife Watch.org, Notes from Nature.org, etc.)

Current design: https://zooniverse.slack.com/files/beckyrother/F683P4SQ1/organization_page_-_talk.png

UPDATES TO DESIGN LINKED:
- introduction section won't need Read More button
- stats show Volunteers, Classifications, Subjects and Completed Subjects, not horizontal bar graph(s)
- bottom left teal section will be about page content
- bottom right section will list social media links and external URLs (like Blog)

(edited by JA 6/15/2017, MB various times)

trouille:
Will be populated by info cards for each project in the federation, see https://github.com/zooniverse/Panoptes-Front-End/issues/2971

jelliotartz:
![org-home-page](https://user-images.githubusercontent.com/12473119/27198168-b347477a-51d6-11e7-8627-99eb6cbb2942.png)

@beckyrother - nothing new here, just wanted to keep you aware that there's a specific issue for the organization landing page.

jelliotartz:
Current state of progress staged [here](https://org-pages.pfe-preview.zooniverse.org/organizations/5)

jelliotartz:
Next steps:
- [ ] Create demo org with projects in various states
- [ ] Organization avatar next to organization title
- [ ] Background Image behind title
- [ ] Project avatar on [project cards](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/project-card.cjsx)
- [ ] remove space between stats/about & footer

beckyrother:
Thanks James! I also made a little style guide: https://zooniverse.slack.com/files/beckyrother/F55KWSPQQ/organizations-style.pdf

mcbouslog:
Projects have the Announcement Banner functionality, might this be something we'll want to implement for organizations?

mcbouslog:
Initial organization landing page merged, closing this issue in favor of specific issues as needed.

PR #4037, opened 2017-09-12
add initial organization landing page

mcbouslog:
Towards #2970.

Initial organization landing page work.

Goals
- basic landing page with org hero, projects and organization details
- show different organization affiliated projects list (see Projects List section below) based on user status (admin, collaborator, volunteer)

Projects List
- collaborator or admin - see Volunteer View toggle (unchecked), initially projects shown are all associated with org, if check Volunteer View toggle then show only launch approved projects
- volunteer, listed org - see launch approved projects linked to org
- volunteer, unlisted org - no landing, message does not have access

Org Details
- researcher quote - placeholder until different design or org Talk comment
- stats - placeholder, how project stats compiled needs discussion
- about - initial add
- links - placeholder for url input and design help

Demo
- (listed org with projects) https://org-landing-page.pfe-preview.zooniverse.org/organizations/markb-panoptes/nfnorgtest
- (unlisted org without projects) https://org-landing-page.pfe-preview.zooniverse.org/organizations/markb-panoptes/galaxyzooorgtest

In Process
- [x] tests
- [x] translations

Future Iterations Will Include
- improved design
- categories to filter projects
- org details placeholders replaced

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://org-landing-page.pfe-preview.zooniverse.org/


eatyourgreens:
"There was an error retrieving {org. name}" shows up while waiting for the organisation to load from the API. Is that expected?

app/pages/organizations/organization-page.jsx
```diff
@@ -0,0 +1,194 @@
...
+            {this.props.organizationAvatar &&
+              <Thumbnail
+                src={this.props.organizationAvatar.src}
+                className="avatar organization-hero__avatar"
+                width={AVATAR_SIZE}
+                height={AVATAR_SIZE}
+              />}
+            <div>
+              <h1 className="organization-hero__title">{this.props.organization.display_name}</h1>
+              <h5 className="organization-hero__description">{this.props.organization.description}</h5>
```
srallen: `h5` doesn't seem like the right tag for the description. Maybe `<p>`?

mcbouslog:
A few brief notes on comments so far:
- "There was an error..." should not be displayed while loading an org, I'll fix
- `Promise.all` sounds much better, I'll refactor accordingly
- collaborator view is so an org builder can see the org landing with related projects regardless of what's officially listed or launched, which while they can get that info from the lab, we're thinking org builders will want to see the cards on the org landing and interact with like a volunteer would, before they list or launch things. I think the same refactor for Proise.all can clean this code up too.
- agreed projects and about page should be separate states, will change

srallen:
> we're thinking org builders will want to see the cards on the org landing and interact with like a volunteer would, before they list or launch things.

Do we really know they'll want that or will use it? I think we're getting ahead of ourselves again building complexity when it may not be needed. We don't let project builders preview their card in the project listing during betas. And, why would they need to? I guess I'm still not understanding the utility here. There isn't really anything to test and we don't let them modify the design or CSS of the organization page.

app/pages/organizations/organization-container.jsx
```diff
@@ -0,0 +1,199 @@
...
+    apiClient.type('organizations').get({ slug, include: 'avatar,background,organization_roles,pages' })
+      .then(([organization]) => {
+        this.setState({ organization });
+
+        if (organization) {
+          const awaitAvatar = apiClient.type('avatars').get(organization.links.avatar.id)
+            .catch(error => console.error('error loading avatar', error)); // eslint-disable-line no-console
+          const awaitBackground = apiClient.type('backgrounds').get(organization.links.background.id)
+            .catch(error => console.error('error loading background', error)); // eslint-disable-line no-console
+          const awaitRoles = organization.get('organization_roles')
```
srallen: Can this be retrieved out the client cache since it was a part of the include instead?

app/pages/organizations/organization-container.jsx
```diff
@@ -0,0 +1,199 @@
...
+            .catch(error => console.error('error loading roles', error)); // eslint-disable-line no-console
+          const awaitPages = organization.get('pages')
```
srallen: Same question about getting this out of the cache.

app/pages/organizations/organization-page.jsx
```diff
@@ -0,0 +1,176 @@
...
+        </section>
+
+        <section className="resources-container">
+          {this.props.collaborator &&
+            <label className="organization-page__toggle" htmlFor="collaborator view">
+              <input onChange={this.handleViewToggle.bind(this)} type="checkbox" value={this.props.collaboratorView} />
+              <span>Collaborator View</span>
+            </label>}
+          <div className="project-card-list">
+            {this.props.organizationProjects.map(project =>
```
srallen: We should have a UI if there are no projects linked yet and the array is empty, since we are allowing organization owners and collaborators to view the org page before it is listed.

srallen: I think having a loading indicator and possibly an error state UI for projects separate from the org loading would be a good idea too.

app/pages/organizations/organization-page.jsx
```diff
@@ -0,0 +1,176 @@
...
+            </div>
+          </div>
+
+          <div className="organization-page__container">
+            <div className="organization-details__content">
+              <h4 className="organization-details__heading">About {this.props.organization.display_name}</h4>
+              {aboutPage &&
+                <div>
+                  <Markdown className={aboutContentClass} project={this.props.organization}>
+                    {aboutPage.content}
```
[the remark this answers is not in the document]
srallen: IE 11 has the image way too big and is causing horizontal scroll. Might be a issue with the known IE bugs with flex-grow.
screen shot 2017-09-22 at 3 16 40 pm

mcbouslog:
Can I revisit the the about section styling issues in the next iteration? The next iteration will include a lot of improvements to the entire details section.

I also plan to add tests for which projects (launched vs all) are shown based on user permission and view state, but would prefer to wait to see exactly what we do with cards (add launch_approved or status attributes) as that may make those tests more straightforward.

Otherwise I think this is in good shape, let me know what you think. Thanks!

app/pages/organizations/organization-container.spec.js
```diff
@@ -0,0 +1,99 @@
...
+
+  describe('with a succesfully fetched organization', function () {
+    it('should render OrganizationPage if organization is listed', function () {
+      const listedOrg = organization;
+      listedOrg.listed = true;
+
+      wrapper.setState({ organization: listedOrg });
+      const orgPage = wrapper.find('OrganizationPage');
+
+      assert.ok(orgPage, 'OrganizationPage render');
```
srallen: Maybe explicitly compare length instead? `assert.equal(orgPage.length, 1)`

app/pages/organizations/organization-container.spec.js
```diff
@@ -0,0 +1,99 @@
...
+    it('should render OrganizationPage if organization not listed and user is collaborator', function () {
+      const isCollabSpy = sinon.stub(OrganizationContainer.prototype, 'isCollaborator').returns(true);
+
+      const unlistedOrg = organization;
+      unlistedOrg.listed = false;
+
+      wrapper.setState({ organization: unlistedOrg });
+      const orgPage = wrapper.find('OrganizationPage');
+
+      assert.ok(orgPage, 'OrganizationPage render');
```
srallen: Same comment with a more explicit comparison here.

app/pages/organizations/organization-page.jsx
```diff
@@ -0,0 +1,245 @@
...
+
+    const aboutContentClass = classnames(
+      'organization-details__about-content',
+      { 'organization-details__about-content--expanded': this.state.readMore });
+
+    return (
+      <div className="organization-page">
+        <section
+          className="organization-hero"
+          style={{ backgroundImage: `url(${this.props.organizationBackground.src})` }}
```
srallen: Not a blocker, but something to think about with the style pass. This will probably error if the organization doesn't have a background.

app/pages/organizations/organization-page.jsx
```diff
@@ -0,0 +1,245 @@
...
+        >
+          <div className="organization-hero__container">
+            {this.props.organizationAvatar &&
+              <Thumbnail
+                src={this.props.organizationAvatar.src}
```
srallen: Same comment about erroring if the organization doesn't have an avatar set yet.

app/pages/organizations/organization-page.spec.js
```diff
@@ -0,0 +1,143 @@
...
+          organization={organization}
+          toggleCollaboratorView={toggle}
+        />);
+      label = wrapper.find('label.organization-page__toggle');
+      checkbox = label.find('input[type="checkbox"]');
+    });
+
+    it('should show project view toggle', function () {
+      assert.ok(label.length, 'Label exists');
+      assert.ok(checkbox.length, 'Checkbox exists');
```
srallen: I'd prefer an explicit comparison here too.

app/pages/organizations/organization-page.spec.js
```diff
@@ -0,0 +1,143 @@
...
+        collaborator={false}
+        organization={organization}
+      />);
+    const label = wrapper.find('label.organization-page__toggle');
+    assert.equal(label.length, 0);
+  });
+
+  it('should render OrganizationProjectCards', function () {
+    const wrapper = shallow(<OrganizationPage organization={organization} />);
+    assert.ok(wrapper.find('OrganizationProjectCards'), 'OrganizationProjectCards exists');
```
srallen: Same comment about explicit comparison.

app/pages/organizations/organization-page.jsx
mcbouslog: great catch! I'll improve in next iteration if ok? the avatar has a fallback stripey look, background has gradient, https://org-landing-page.pfe-preview.zooniverse.org/organizations/markb-panoptes/noavnoback, definitely could use some refinement.

srallen: Yeah totally ok to fix this with the style fixes.

PR #4038, opened 2017-09-13
Fix a couple of pan zoom bugs

eatyourgreens:
Checks image height and width for changes before resetting zoom for a new subject.

Fixes some eslint warnings.

Adds an eslint rule for switch/case indentation.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://pan-zoom.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -128,7 +128,7 @@ class PanZoom extends React.Component {
   }
 
   zoomReset(props) {
-    props = props ? props : this.props;
+    props = props || this.props;
     this.setState({
       viewBoxDimensions: {
         width: props.frameDimensions.width,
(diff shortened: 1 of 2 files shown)


app/components/pan-zoom.jsx
```diff
@@ -126,16 +127,17 @@ class PanZoom extends React.Component {
...
   }
 
-  zoomReset() {
+  zoomReset(props) {
+    props = props || this.props;
     this.setState({
       viewBoxDimensions: {
-        width: this.props.frameDimensions.width,
-        height: this.props.frameDimensions.height,
+        width: props.frameDimensions.width,
```
srallen: Testing with the current PRN project (https://pan-zoom.pfe-preview.zooniverse.org/projects/vrooje/planetary-response-network-and-rescue-global-caribbean-storms-2017/classify?env=production), if you rotate or zoom in the subject image and hit the reset button, it'll throw ` TypeError: Cannot read property 'width' of undefined`. I think the issue is this line, but it's a little hard for me to be sure looking at the minified file.

eatyourgreens:
@srallen that's the right workflow. The issue showed up when you finished a classification and loaded the next subject, because the subjects have different aspect ratios and sizes. I used demo mode to reproduce it.

srallen:
@eatyourgreens thanks for the clarification. I'll take a look after the zoominar.

app/components/pan-zoom.jsx
eatyourgreens: My bad. I broke the handler for that button.

eatyourgreens: Should be fixed by f5066ed

eatyourgreens:
Prop warnings are fixed for the tests too.

PR #4044, opened 2017-09-14
refactor slider to stateless component

mcbouslog:
Refactor slider task to stateless component. Mentioning @eatyourgreens, @tingard and @CKrawczyk. 

IMPORTANT - see PR #4004 for original comments and notes.

Had trouble creating tests to make sure annotation properly updated, related to controlled vs uncontrolled component comments. Added tests for `onChange`, though I'm unsure if that's sufficient. If we can't be comfortable annotation is properly updated with this stateless component, I say we scrap this and leave slider task as is (with state).

I see tests for if task is required, but don't see ability to toggle required in editor?

If there are two slider tasks back to back in a workflow, the second doesn't receive intended focus, unsure how to fix, though I don't think necessary and unrelated to this PR's specific scope, happy to add if simple solution.

Staging workflow I've been tinkering on: https://refactor-slider-stateless.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project/classify?reload=0&workflow=3030

# Review Checklist

- [x] Does it work in all major browsers: ~~Firefox~~, ~~Chrome~~, Edge, ~~Safari~~?
- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://refactor-slider-stateless.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/tasks/index.spec.js
+++ b/app/classifier/tasks/index.spec.js
@@ -7,14 +7,14 @@ for (const key in tasks) {
   const TaskComponent = tasks[key];
   describe('Task ' + key, function() {
     let wrapper;
-  
+
     it('should render with default props', function() {
       wrapper = shallow(<TaskComponent />);
     });
-  
+
     it('should update on annotation change', function() {
       wrapper = shallow(<TaskComponent />);
-      let annotation = TaskComponent.getDefaultAnnotation()
+      let annotation = TaskComponent.getDefaultAnnotation(tasks[key])
       wrapper.setProps({annotation});
     });
   });
--- a/app/classifier/tasks/index.spec.js
+++ b/app/classifier/tasks/index.spec.js
@@ -2,10 +2,12 @@ import { shallow } from 'enzyme';
 import React from 'react';
 import assert from 'assert';
 import tasks from './';
+import { workflow } from '../../pages/dev-classifier/mock-data';
 
-for (const key in tasks) {
-  const TaskComponent = tasks[key];
-  describe('Task ' + key, function() {
+for (const key in workflow.tasks) {
+  const task = workflow.tasks[key];
+  const TaskComponent = tasks[task.type];
+  describe('Task ' + task.type, function() {
     let wrapper;
 
     it('should render with default props', function() {
@@ -14,7 +16,7 @@ for (const key in tasks) {
 
     it('should update on annotation change', function() {
       wrapper = shallow(<TaskComponent />);
-      let annotation = TaskComponent.getDefaultAnnotation(tasks[key])
+      let annotation = TaskComponent.getDefaultAnnotation(task, workflow, tasks);
       wrapper.setProps({annotation});
     });
   });
--- a/app/classifier/tasks/slider/index.jsx
+++ b/app/classifier/tasks/slider/index.jsx
@@ -5,6 +5,16 @@ import SliderTaskEditor from './editor';
 
 const NOOP = Function.prototype;
 
+const SLIDERTASKDEFAULT = {
+  defaultValue: '0',
+  help: '',
+  instruction: 'Enter an Instruction.',
+  max: '1',
+  min: '0',
+  step: '0.1',
+  type: 'slider'
+};
+
 const SliderSummary = (props) => {
   let answer = 'No answer';
   if (props.annotation.value !== null) {
@@ -32,7 +42,10 @@ SliderSummary.propTypes = {
   ).isRequired,
   annotation: React.PropTypes.shape(
     {
-      value: React.PropTypes.string
+      value: React.PropTypes.oneOfType([
+        React.PropTypes.string,
+        React.PropTypes.number
+      ])
     }
   ).isRequired
 };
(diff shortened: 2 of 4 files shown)


app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,64 @@ SliderSummary.propTypes = {
...
+                <div className="slider-task-range__label-container__right">
+                  {task.max}
                 </div>
-              </label>
-            </div>
-            <label className="answer slider-task-number">
-              <input
-                className="standard-input"
-                type="number"
-                autoFocus={this.props.autoFocus}
```
mcbouslog: removed autofocus from number input, left on slider, let know if opposite preferred

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,64 @@ SliderSummary.propTypes = {
...
+              <div>
+                <input
+                  className="standard-input"
+                  type="range"
+                  autoFocus={autoFocus}
+                  onChange={handleChange.bind(this)}
+                  max={task.max}
+                  min={task.min}
+                  step={task.step}
+                  value={annotation.value === null ? task.defaultValue : annotation.value}
```
CKrawczyk: The issue with setting the value like this is when the slider is not moved from the default position `annotation.value` is `null` meaning the data dump for the user's classification would be `null` instead of the default value.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,64 @@ SliderSummary.propTypes = {
...
-}
+          <label className="answer slider-task-number">
+            <input
+              className="standard-input"
+              type="number"
+              onChange={handleChange.bind(this)}
+              max={task.max}
+              min={task.min}
+              step={task.step}
+              value={annotation.value === null ? task.defaultValue : annotation.value}
```
CKrawczyk: Same as above.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,64 @@ SliderSummary.propTypes = {
...
+      >
+        <div className="slider-task-container full ">
+          <div className="slider-task-range">
+            <label className="answer">
+              <div>
+                <input
+                  className="standard-input"
+                  type="range"
+                  autoFocus={autoFocus}
+                  onChange={handleChange.bind(this)}
```
eatyourgreens: I don't think you need to bind `handleChange` to the `SliderTask` function here.

app/classifier/tasks/slider/index.jsx
```diff
@@ -37,94 +37,64 @@ SliderSummary.propTypes = {
...
-        </GenericTask>
-      </div>
-    );
-  }
-}
+          <label className="answer slider-task-number">
+            <input
+              className="standard-input"
+              type="number"
+              onChange={handleChange.bind(this)}
```
eatyourgreens: I don't think bind is needed here, either.

eatyourgreens:
I think that null annotation value is being set by [`SliderTask.getDefaultAnnotation()`](/zooniverse/Panoptes-Front-End/blob/bad325c12628e45070102f24af78ecd6714b7846/app/classifier/tasks/slider/index.jsx#L115-L116)

app/classifier/tasks/index.spec.js
```diff
@@ -7,14 +7,14 @@ for (const key in tasks) {
...
+
     it('should render with default props', function() {
       wrapper = shallow(<TaskComponent />);
     });
-  
+
     it('should update on annotation change', function() {
       wrapper = shallow(<TaskComponent />);
-      let annotation = TaskComponent.getDefaultAnnotation()
+      let annotation = TaskComponent.getDefaultAnnotation(tasks[key])
```
mcbouslog: this now breaks when testing the combo, test fail message:
```
1) Task combo should update on annotation change:
TypeError: Cannot read property 'map' of undefined
at Function.getDefaultAnnotation (app/classifier/tasks/combo/index.cjsx:20:35)
at Context.<anonymous> (app/classifier/tasks/index.spec.js:37:38)
```

CKrawczyk: I don't think `getDefaultAnnotation()` takes in a parameter, if it can be set up so it can, that would solve the issue.

eatyourgreens: I think the combo task changed `getDefaultAnnotation` so that it could be invoked with a task, but I forget how that was implemented.

The error is something to do with an undefined array somewhere.

eatyourgreens:
Apart from the failing test, the slider task seems to be working when I try Steelpan Vibrations on this branch.

app/classifier/tasks/index.spec.js
eatyourgreens: This code change has changed a generic test for all tasks, so `tasks[key]` would have to be a valid task for each task type ie. it should be an array of tasks for the combo task.

eatyourgreens: From this line, earlier in the test, `const TaskComponent = tasks[key];` `task[key` is a React component, and not a task description.

eatyourgreens: After a bit of trial and error with the combo task, I think 7ef3266 fixes the tests by using the mock tasks from the dev classifier.

CKrawczyk:
Everything is working as intended and the tests are passing.

PR #4053, opened 2017-09-21
HomePageRoot ES6 conversion

jelliotartz:
Describe your changes.
- Converts HomePageRoot component from coffeescript to ES6.
- Adds tests for HomePageRoot component.
- Rather than passing in all props as was done in the coffeescript file, I looked at what props were being used in `HomePageForUser` and only passed in what was being used (`user` and `location`).
- Similarly, no props are used in `HomePageNotLoggedIn`, so I didn't pass any in.
- Question: is there an issue created for ES6 conversion? I searched Issues and couldn't find one to append this PR to.
- Staged [here](https://home-cs-to-js.pfe-preview.zooniverse.org/).

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/pages/home.spec.js
```diff
@@ -0,0 +1,60 @@
...
+
+    beforeEach(function () {
+      wrapper = shallow(<HomePageRoot user={user} location={location} />)
+    });
+
+    it('without crashing', function () {
+      assert.equal(wrapper, wrapper)
+    });
+
+    it('with a user prop', function() {
```
srallen: This is testing the child component, not really `HomePageRoot`. This test would be more appropriate for testing `HomePageLoggedIn`.

app/pages/home.spec.js
```diff
@@ -0,0 +1,60 @@
...
+      const userDisplayName = wrapper.nodes[0].props.user.display_name
+      assert.equal(userDisplayName, 'Test User')
+    });
+
+    it('with a location prop', function() {
```
srallen: Same comment about testing the child, not `HomePageRoot`

app/pages/home.spec.js
```diff
@@ -0,0 +1,60 @@
...
+      wrapper = shallow(<HomePageRoot />)
+    })
+
+    it('without crashing', function () {
+      assert.equal(wrapper, wrapper)
+    });
+
+    it('for a not authenticated user', function() {
+      const child = wrapper.children().root.length
+      assert.equal(child, 1)
```
srallen: What you could do instead is `assert.equal(wrapper.find('HomePageNotLoggedIn').length, 1)`

app/pages/home.spec.js
```diff
@@ -0,0 +1,60 @@
...
+    })
+
+    it('without a user prop', function() {
```
srallen: This is testing the child component, not `HomePageRoot`

jelliotartz:
Thanks for your comments on how to test just `HomePageRoot` and not its child components. I appreciate the feedback!

I'm not sure what why, but the syntax you're suggesting

```js
it('renders <HomePageNotLoggedIn /> when there isn\'t a user', function() {
  assert.equal(wrapper.find('HomePageNotLoggedIn').length, 1);
});
```

is not passing. It looks like it's trying to check the length of a ShallowWrapper object, and it's returning 0 instead of 1.

Adding `first()` or `root` eg

```js
it('renders <HomePageNotLoggedIn /> when there isn\'t a user', function() {
  assert.equal(wrapper.find('HomePageNotLoggedIn').root.length, 1)
});

it('renders <HomePageLoggedIn /> when there is a user', function() {
  wrapper.setProps({ user, location });
  assert.equal(wrapper.find('HomePageLoggedIn').first().length, 1)
});
```

creates passing tests, but I still think there's something wonky going on with the syntax or the assert library - I can interchange `HomePageLoggedIn` and `HomePageNotLoggedIn` in the assert statements, and both still pass, so it doesn't seem like it's doing what it's supposed to.

I thought maybe it has to do with not importing `HomePageLoggedIn` and `HomePageNotLoggedIn`, but I tried importing them directly, and did not find a difference in terms of tests passing and failing.

I've been looking at [collaborator.spec](https://github.com/zooniverse/Panoptes-Front-End/blob/9fc241547630148bd10a3fe97e5272b61ca8d6d7/app/collections/collaborators.spec.js) as an example of how to use the `assert` library's syntax. I'm at a loss to explain why `wrapper.find('Component').length` seems to be working differently in my spec than in collaborator spec.

app/pages/home.spec.js
```diff
@@ -0,0 +1,29 @@
...
+};
+
+const location = {
+  hash: '#page'
+};
+
+describe('HomePageRoot', () => {
+  let wrapper;
+
+  beforeEach(() => {
```
srallen: This can just be `before()`

jelliotartz:
Gah! Thank you!! I'm tempted to put a 'Common Gotchas' PR up on the Enzyme docs to put that little detail out there!

PR #3995, opened 2017-08-23
Optimise subject viewer CSS and fix Firefox image rendering bug

eatyourgreens:
Fixes #3989 
Describe your changes.
Some small changes to the subject viewer. Inline styles are moved to the style sheet. Background colour and positioning are applied to a `div`, instead of to the `svg` element.

The combo task now checks whether a task renders SVG before rendering SVG marks.

These changes were made while trying to discover why Firefox 55 isn't displaying SVG images properly in the classifier. Unfortunately, it doesn't fix that bug.

UPDATE: checking the image size before rendering seems to fix the SVG image in FF 55+.

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch? https://fix-3989.pfe-preview.zooniverse.org/
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [x] Are the tests passing locally and on Travis?

## Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4055, opened 2017-09-21
Fix audio player SVG rendering

### LINKED ISSUE #4054: Audio player doesn't render SVG progress marker
## Expected behavior
Audio subjects with an associated image should show a vertical SVG line that advances as the audio plays.
## Current behavior
The progress marker is rendered in the DOM, but appears to be rendered behind the image component.
## Steps to replicate
Play the test audio subject at https://master.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/classify

### LINKED ISSUE #4067: Audio player doesn't work for subjects in Talk
## Expected behavior
Talk subjects should render the audio player in the same way that it's rendered in the classifier.

## Current behavior
Talk subjects show up as a slideshow with two frames: one for the audio file and one for the image file. This is likely a result of the way the code tests the subject locations array to guess whether or not a subject should use the audio player.

## Steps to replicate
Compare
https://master.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/talk/subjects/50252

with
https://master.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/classify

eatyourgreens:
Fixes #4054 
Fixes #4067 

Describe your changes.
Reverses one of the changes in #3995 so that the subject frame viewer is rendered after the SVG marking surface in the DOM.

I'm not 100% sure why this works, so it may need careful checking.

This PR also fixes ESLint warnings in the `ProgressIndicator` component and removes some redundant HTML from the audio player.

Some if logic is also reversed in the subject viewer so that we check for a mixed media audio file before deciding to render the flipbook slideshow. This fixes audio subject rendering on Talk.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://fix-audio-player.pfe-preview.zooniverse.org/

## Optional

### DIFF
--- a/app/classifier/annotation-renderer/svg.jsx
+++ b/app/classifier/annotation-renderer/svg.jsx
@@ -161,7 +161,6 @@ export default class SVGRenderer extends React.Component {
 
     return (
       <div>
-        {this.props.children}
         <div className="subject svg-subject">
           <svg
             ref={(element) => { if (element) this.svgSubjectArea = element; }}
@@ -197,6 +196,7 @@ export default class SVGRenderer extends React.Component {
           </svg>
         </div>
         {(showFeedback) && (<SVGToolTipLayer getScreenCTM={this.getScreenCurrentTransformationMatrix} />)}
+        {this.props.children}
       </div>
     );
   }
--- a/app/components/file-viewer/audio-player.jsx
+++ b/app/components/file-viewer/audio-player.jsx
@@ -107,7 +107,7 @@ class AudioPlayer extends React.Component {
       );
     }
     return (
-      <div>
+      <div className="subject-audio-frame">
         {imageElement}
         <div className="audio-player-component">
           <audio
@@ -123,8 +123,6 @@ class AudioPlayer extends React.Component {
           >
             Your browser does not support the audio format. Please upgrade your browser.
           </audio>
-
-          {this.props.children}
         </div>
       </div>
     );
@@ -133,7 +131,6 @@ class AudioPlayer extends React.Component {
 }
 
 AudioPlayer.propTypes = {
-  children: React.PropTypes.node,
   format: React.PropTypes.oneOfType([React.PropTypes.array, React.PropTypes.string]),
   frame: React.PropTypes.number,
   onLoad: React.PropTypes.func,
--- a/app/components/file-viewer/audio-player.jsx
+++ b/app/components/file-viewer/audio-player.jsx
@@ -11,9 +11,12 @@ class AudioPlayer extends React.Component {
     this.endAudio = this.endAudio.bind(this);
     this.playAudio = this.playAudio.bind(this);
     this.onAudioLoad = this.onAudioLoad.bind(this);
+    this.onImageLoad = this.onImageLoad.bind(this);
     this.updateProgress = this.updateProgress.bind(this);
 
     this.state = {
+      naturalHeight: 0,
+      naturalWidth: 0,
       playing: false,
       progressPosition: 0,
       trackDuration: 0
@@ -27,6 +30,12 @@ class AudioPlayer extends React.Component {
   onAudioLoad(e) {
     e.stopPropagation();
     this.state.trackDuration = this.player.duration;
+  }
+
+  onImageLoad(e) {
+    const { naturalHeight, naturalWidth } = e.target;
+    this.setState({ naturalHeight, naturalWidth });
+    this.props.onLoad(e);
   }
 
   imageSrc() {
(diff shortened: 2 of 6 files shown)


eatyourgreens:
Here's a test subject that has an audio file with linked spectrogram image
https://fix-audio-player.pfe-preview.zooniverse.org/projects/hughdickinson/testing-audio-subjects/classify

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -197,6 +196,7 @@ export default class SVGRenderer extends React.Component {
           </svg>
         </div>
         {(showFeedback) && (<SVGToolTipLayer getScreenCTM={this.getScreenCurrentTransformationMatrix} />)}
+        {this.props.children}
```
eatyourgreens: Why doesn't this render the SVG behind everything else, preventing you from seeing your marks during drawing tasks?

app/components/file-viewer/audio-player.jsx
```diff
@@ -83,43 +83,32 @@ class AudioPlayer extends React.Component {
...
+            src={imageSrc}
+            type={this.imageTypeString()}
+            format={this.imageFormatString()}
+            frame={this.props.frame}
+            onLoad={this.props.onLoad}
+            onFocus={this.props.onFocus}
+            onBlur={this.props.onBlur}
+          />
+        </ProgressIndicator>
       );
```
eatyourgreens: Removed a bunch of redundant `div` elements here, in favour of nested React components.

app/classifier/annotation-renderer/svg.jsx
srallen: I'm not exactly sure. I've been spending some time to understand what the various children are at different levels of the viewer tree and it's hard to keep track. What I have found is that having the `this.props.children` back on line 164 with a z-index will get it to render on top and won't interfere with the drawing SVG layer too. I think it has to do with not totally understanding what's getting priority with the box model and the `position: absolute` being used on both the div with the `svg-subject` class, the audio player, and the svg progress marker. It's confusing to me to have the audio player rendered as a child of the SVGRenderer component too.

eatyourgreens: Thanks for taking a look. The positioning here puzzles me too. This change resets the classifier back to the way it originally worked before I messed about with it in #3995. I can't understand why drawing marks rendered above the subject in the old classifier setup. CSS `position` declarations reset [the stacking context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context), so it may have something to do with that.

eatyourgreens:
The subject being a child of the renderer component comes from the [generic Renderer](/zooniverse/Panoptes-Front-End/blob/master/app/classifier/annotation-renderer/index.jsx#L26-L28) I'm happy to take another look at that. That pattern was mostly used for the text highlighter tools, but the general idea was that you pass an array of annotations to the renderer, which then wraps a single frame of the subject so that you can draw on the subject frame, or highlight it, or interact with it in general.

The confusing DOM in the SVG component goes back to the original classifier design, for backwards compatibility.

app/classifier/annotation-renderer/svg.jsx
eatyourgreens: Oh and the audio player is just plain broken. If you add a subject which is only an mp3 file, the player renders underneath the subject metadata and favourite buttons, and is totally useless.

eatyourgreens:
I'm having another look at this. IIRC, the strange set up in the audio player (absolutely positioned SVG rendered above a HTML image) happened because React doesn't seem to support the `onload` attribute on SVG `image`. I'm not 100% sure browsers support that either.

eatyourgreens:
Removed `this.props.children` from the audio and video players. I think this was there to support a generic loading indicator, originally, but I can't see children being passed to [the FileViewer](/zooniverse/Panoptes-Front-End/blob/master/app/components/frame-viewer.jsx#L83-L91) now.

eatyourgreens:
I tried removing the HTML image from the DOM, once the SVG image had loaded, but the layout broke so I think I'll just leave this for the time being.

eatyourgreens:
I've also noticed that if I remove `position:absolute` from the progress marker, in order to fix the layout, I can either display the red progress marker line, or the svg marking surface for the drawing task, but I can't have both at the same time.

Definitely leaving this now, as it seems to be at a point where any change breaks it.

eatyourgreens:
Latest commit has broken the drawing task. Marks are hidden behind the subject now. 😢

eatyourgreens:
But all the tests are passing, so that's good.

eatyourgreens:
Hacky fix in 912b3f8: lower the opacity of the foreground image.

eatyourgreens:
Related issue: #4067 the audio player doesn't render at all in Talk.

app/classifier/annotation-renderer/svg.jsx
srallen: I suppose that can be fixed as needed. Do we have any audio only projects? The projects that want audio at the moment also seem to want the spectrogram image paired with it.

eatyourgreens:
Replacing `height` with `max-height` on the the progress marker seems to fix that. I'm not completely sure what's going on there either. Something to do with the calculated height of the image container including the audio player.

eatyourgreens:
I've played around a bit with using `position: relative` on the progress marker's container, which would be better CSS, in my opinion. That creates a new stacking context for the progress marker, which breaks the drawing task by rendering the SVG progress marker in front of the SVG marking surface.

eatyourgreens:
I came up with a much nicer implementation this afternoon, with much cleaner CSS. It worked in Talk but didn't work at all with the drawing tasks in the classifier, so I ended up binning it. :man_shrugging:

PR #90, opened 2015-02-01
Sign in dialog

brian-c:
New strategy for handling dialog returns with promises. You can now do normal alerts:

```coffee
alert <div>Whatever</div>
```

or pass in a promise executor function if you want to get some value out:

```coffee
what = alert (resolve, reject) ->
  <div>
    <button onClick={resolve.bind null, 'Awesome'}>OK</button>
    <button onClick={reject}>Cancel</button>
  </div>

what.then (value) ->
  console.log 'Dialog closed:', value
```

Also snuck in a privacy policy page.

PR #4063, opened 2017-09-22
PrivacyPolicy ES6 conversion

jelliotartz:
Describe your changes.
- Converts PrivacyPolicy from coffeescript to ES6.
- Rename component from PrivacyPage to PrivacyPolicy.
- Removes `forceUpdate()` called via lifecycle hooks and `counterpart.on` and `offLocaleChange`. Component and site are working without it, and privacy policy text is only written in English.
- Refactors PrivacyPolicy to functional component.
- Moves privacy policy text to `locales/en`.
- Breaks up privacy policy text into smaller pieces
  - Each bolded section into separate strings, for easier readability.
- Adds unit tests for heading and content.
- Staged [here](https://privacy-policy-cs-to-js.pfe-preview.zooniverse.org/privacy).
- Related: https://github.com/zooniverse/Panoptes-Front-End/pull/90 (original creation of `privacy-policy.cjsx` file).

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+  beforeEach(() => {
+    wrapper = shallow(<PrivacyPolicy />);
+  });
+
+  it('renders without crashing', () => {
+    const privacyPolicyContainer = wrapper.find('div.content-container');
+    assert.equal(privacyPolicyContainer.length, 1);
+  });
+
+  describe('heading', () => {
```
srallen: This isn't really testing what is shallow rendered but rather that the Translate component has a `h1` prop. A unit test that you could do instead would be that a `Translate` component is shallow rendered in `PrivacyPolicy`:

```
it('renders a <Translate /> component', function() {
  assert.equal(wrapper.find('Translate').length, 1);
});
```

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+    it('renders the page title', () => {
+      const headingElement = wrapper
+        .find('div.content-container').children()
+        .first()
+        .props().component;
+      assert.equal(headingElement, 'h1');
+    });
+
+    it('renders page title content', () => {
```
srallen: If you're wanting to check the value of a prop using enzyme, you can pass the key of the prop into the method `.prop('key')`: `wrapper.prop('content')`

The selectors for enzyme's find method works a lot like jQuery. Right now it's a bit hard to follow what is being tested here. This instead could be:
```
assert.equal(wrapper.find('Translate').prop('content'), 'privacy.title');
```

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
+import React from 'react';
+import assert from 'assert';
+import { shallow } from 'enzyme';
+import PrivacyPolicy from './privacy-policy';
+
+describe('PrivacyPolicy', () => {
+  let wrapper;
+
+  beforeEach(() => {
```
srallen: I don't think this needs to be a `beforeEach`. `beforeEach` runs before each test in the describe block and I don't think these unit tests need the wrapper to be recreated each time.

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+        .first()
+        .props().content;
+      assert.equal(headingContent, 'privacy.title');
+    });
+  });
+
+  describe('user agreement section', () => {
+    let markdownElements;
+
+    beforeEach(() => {
```
srallen: Again, I think a `before` is more fitting here than `beforeEach`

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+      markdownElements = wrapper
```
srallen: This again is hard to figure out what exactly is being targeted. Try to simplify the selectors.

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+        .first()
+        .props().children;
+    });
+
+    it('renders four markdown elements', () => {
+      assert.equal(markdownElements.length, 4);
+    });
+
+    it('renders each of its sections', () => {
+      const sections = [
```
srallen: Maybe import the actual keys from the locale file?

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+        assert.ok(element.props.children.match(sections[index]));
+      });
+    });
+  });
+
+  describe('privacy policy section', () => {
+    let markdownElements;
+
+    beforeEach(() => {
+      markdownElements = wrapper
```
srallen: Again try to simplify the selectors. You should be able to more directly target what you're trying to use here.

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,101 @@
...
+        .last()
+        .props().children;
+    });
+
+    it('renders ten markdown elements', () => {
+      assert.equal(markdownElements.length, 10);
+    });
+
+    it('renders each of its sections', () => {
+      const sections = [
```
srallen: Also here, maybe import the actual keys from the locale file.

app/pages/privacy-policy.spec.js
```diff
@@ -0,0 +1,68 @@
...
+
+    it('renders four markdown elements', () => {
+      assert.equal(markdownElements.length, 4);
+    });
+
+    it('renders each of its sections', () => {
+      const userAgreementSections = Object.keys(englishTranslations.privacy.userAgreement);
+
+      markdownElements.forEach((element, index) => {
+        assert.ok(element.props().children.match(userAgreementSections[index]));
```
srallen: On my first pass I didn't quite understand what you were trying to test, but it's more clear now. I think you're trying to test that the english translation gets rendered in the DOM? However, that isn't what is happening. What you're asserting here is that the the child of the Markdown component is matching a regex. You're giving it the key from the locales file, so it'll try to match literally for instance `'en.privacy.userAgreement.summary'`. It's passing because the shallow render output is actually throwing a counterpart error, for example: `missing translation: en.privacy.userAgreement.summary`. I'm not totally sure but it might be because shallow rendering doesn't run the full react component lifecycle.

jelliotartz: Yeah, I did see the counterpart error, and my thought process was that matching on the regex would test whether each of the markdown elements at least text-matched a key in the correct part of the translations. so if something changes and the markdown element doesn't match what's in the translation key, the test will fail. Are you ok with this approach? or is it outside the responsibility of the tests for this component?

srallen: It's not obvious at a cursory glance that the reason it passes is because counterpart is failing in the shallow render. I think this could be confusing to other devs who may look at this later. Ultimately, I don't think we don't want to test if something is failing accidentally. You'd need to fix the test rendering so that what we actually expect, the parsed markdown into HTML with the expected English text, and then make an assertion that the rendered HTML and text equals the expected HTML and text. Or refactor the tests to actually test just the component as I suggest, not its children or the dependencies.

Testing the specific rendered text after all the parsing and transpiling done by Markdownz, markdown-it, counterpart, and React is not something we should be doing since that is essentially retesting dependencies that are already tested.

PR #4072, opened 2017-09-29
Localise the survey task summary

eatyourgreens:
Describe your changes.
Convert the survey task summary to ES6.
Move embedded English strings out into the counterpart locale file.
Add tests for the survey summary.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://localise-survey-summary.pfe-preview.zooniverse.org/

## Optional


app/classifier/tasks/survey/summary.jsx
```diff
@@ -0,0 +1,83 @@
+import React from 'react';
+import Translate from 'react-translate-component';
+import Utility from './utility';
+
+class SurveySummary extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      expanded: props.expanded
```
simoneduca: I think passing props to state isn't an anti-pattern, if you were just seeding data that doesn't need to be updated during the lifecycle. Is that the case?

app/classifier/tasks/survey/summary.jsx
```diff
@@ -0,0 +1,83 @@
...
+          {this.state.expanded &&
+            choiceSummaries.map((choiceSummary, i) => (
+              <div key={i} className="answer">
+                {choiceSummary}
+              </div>
+            ))
+          }
+        </div>
+      </div>
+    );
```
simoneduca: I feel all this block could use some destructuring to reduce the number of `this.props`

app/classifier/tasks/survey/summary.jsx
```diff
@@ -0,0 +1,83 @@
...
+      <div>
+        <div className="question">
+          <Translate
+            content="tasks.survey.surveyOf"
+            with={{
+              count: this.props.task.choicesOrder.length
+            }}
+          />
+          {this.state.expanded ?
+            <button type="button" className="toggle-more" onClick={this.setState.bind(this, { expanded: false }, null)}>
```
simoneduca: This is just me not having seen/used this before, but could the `onClick` method be moved to its own function, so we can avoid repetition too?

app/classifier/tasks/survey/summary.jsx
eatyourgreens: Could be. I think this is just copied straight across from the original code.

app/classifier/tasks/survey/summary.jsx
eatyourgreens: Add two new methods, one to expand the section and one to contract it? Or one new method with bound arguments?

app/classifier/tasks/survey/summary.jsx
eatyourgreens: Expanded state is changed by clicking the buttons to open/close the summary. Is that what you were asking about?

eatyourgreens:
Updated and staged at https://localise-survey-summary.pfe-preview.zooniverse.org/

eatyourgreens:
Fixed the summary saying 'Survey of %(count)' too! Oops.

app/classifier/tasks/survey/summary.jsx
simoneduca: As far as I can see, the expand button works properly, so this isn't blocking. However, for being extra cautious, could we just have `false` as a default in the constructor, instead of feeding the props?

eatyourgreens: But the tests pass in `expanded=true` since you can render with the summary initially open, or am I misunderstanding?

eatyourgreens:
Added a test for that 'Survey of…' bug.

app/classifier/tasks/survey/summary.jsx
simoneduca: I think I'm just being rubbish at being clear. My thought: if `props.expanded` is never going to be `true` on first render, why don't just assign `false` to the local initial state and avoid a possible antipattern?

eatyourgreens: And my reply was that `expanded` can be true on first render, hence there's a test for it. I feel like I'm missing something obvious here.

PR #4075, opened 2017-10-03
dropdown subTask

### LINKED ISSUE #3962: Add dropdown as a subTask option for drawing tools
## Expected behavior
A volunteer draws a shape on an image. Then, the volunteer is prompted to label that mark with an option from a dropdown menu.

## Current behavior
Currently, you can achieve a similar task flow by asking a volunteer to generate a text label through a text input subTask. For projects with a set of anticipated labels, free text entry may complicate the data and the aggregation process.

simensta:
Fixes #3962.

**Describe your changes.**
This PR adds the dropdown task as a possible drawing subTask. The dropdown subTask can only be added to a workflow if the dropdown task is enabled for the project.

You can try it out with this demo project: https://dropdown-subtask.pfe-preview.zooniverse.org/projects/scribe-2/demo-for-dropdown-subtask/classify

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://dropdown-subtask.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -70,6 +70,7 @@ module.exports = React.createClass
               <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
             <hr />
             <p style={textAlign: 'center'}>
+              {# The tabIndex is set on the button as a workaround to allow the OK button to gain focus when the subtask is a dropdown.}
               <button tabIndex="0" autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
             </p>
           </ModalFocus>
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -70,7 +70,7 @@ module.exports = React.createClass
               <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
             <hr />
             <p style={textAlign: 'center'}>
-              <button autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
+              <button tabIndex="0" autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
             </p>
           </ModalFocus>
         </StickyModalForm>}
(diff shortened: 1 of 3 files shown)


eatyourgreens:
The layout goes wrong on a phone screen. I'm not sure why it's rendering so wide. dropdown task on an iPhone 6+, cut off on left and right of screen.

app/classifier/drawing-tools/root.cjsx
```diff
@@ -70,7 +70,7 @@ module.exports = React.createClass
               <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
             <hr />
             <p style={textAlign: 'center'}>
-              <button autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
+              <button tabIndex="0" autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
```
eatyourgreens: Buttons automatically have tabindex 0 so I'm not sure why this is here. The commit message says it's a fix, but not what it was fixing. Is it a workaround for my bad autofocus code on the button?

simensta: The commit message could definitely be clearer. If the tabIndex is not indicated on the button, the dropdown task won't relinquish focus. I don't think this is an issue with the other subTasks.

simensta:
Thanks for catching the issue with the dialog size on mobile, @eatyourgreens! I'll look into a fix for that.

app/classifier/drawing-tools/root.cjsx
simensta: Would the thing to do here be to rebase and edit the commit message for 002e98a so that it is clearer?

eatyourgreens: I think a comment in the code would help, so that the tabindex doesn't get deleted during a code cleanup. 

I'll make a note to have a look at the weird focus behaviour. The dropdown menus don't do that in Notes from Nature, so it might be caused by code I added to handle tabbing through drawing subtasks.

simensta: Sounds good, I'll add a comment in the code.

mcbouslog:
For me worked on Chrome v61, Firefox v56, Safari v11 and EdgeHTML v13. Not sure about my VM's Edge version, but I think that's good enough?

simensta:
Thanks @mcbouslog! The demo project works for me on a VM with Windows 10 and Microsoft Edge 15.

simensta:
@eatyourgreens, what do you think about the new dropdown dialog size on mobile?

eatyourgreens:
@simensta works for me. By the way, the focus bug is due to [line 18 of the ModalFocus component](/zooniverse/Panoptes-Front-End/blob/master/app/components/modal-focus.jsx#L18) ignoring the OK button, because it's disabled when the component mounts. The `focusables` array doesn't get refreshed when the button changes state.

PR #4079, opened 2017-10-04
Always submit a classification

### LINKED ISSUE #4078: SW Demo - classification not recording
## Expected behavior
The SW Demo workflows have 'hide classification summaries selected so the classification window should show the "done" and "done & talk" buttons

## Current behavior
Only a 'done' button is available and classification summaries are shown in the next window (see images below)
However, the more key problem is that the classification is not being recorded in the database (see notes below)
Also one can go through the images indefinitely, no 'already seen' flag is shown (this used to happen quickly for this small test set of 16 subjects)
image
image

## Steps to replicate
Launch workflow
https://www.zooniverse.org/projects/aprajita/swdemo/classify?reload=0&workflow=4817

## Additional information
Hi @simoneduca, when trying out the project with @marten today, he noticed that no classification have been recorded since 22 Sept, and "FeedbackMessage"(?) was persistently False in the classifications made before this date.  We both tried to make classifications at the time, some new ones did appear but any training subject with the feedback message shown as the second image above was not recorded, only the non-training subjects were recorded.
 
I think the 2 main points are
- classification summaries are being shown even though this is set to "hide"
- classifications on training images are not being recorded

Not sure what has changed, as I think this workflow used to behave as expected.

- **Operating system:**
OS X 10.10.5

- **Browser:**
Chrome Version 61.0.3163.91 (Official Build) (64-bit)

eatyourgreens:
Fixes #4078 .

Describe your changes.
Kind of a hacky fix, but makes sure that `props.onComplete` is always called by the classifier for a completed classification.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -168,6 +168,8 @@ class Classifier extends React.Component {
     if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
       if (!feedbackActive || (feedbackActive && !isThereFeedback(this.props.subject, this.props.workflow))) {
           this.props.onCompleteAndLoadAnotherSubject();
+        } else {
+          this.props.onComplete();
         }
     } else {
       this.props.onComplete()


app/classifier/classifier.jsx
```diff
@@ -168,6 +168,8 @@ class Classifier extends React.Component {
     if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
       if (!feedbackActive || (feedbackActive && !isThereFeedback(this.props.subject, this.props.workflow))) {
```
eatyourgreens: I don't understand how this can ever be true, since it ORs `!feedbackActive` and `feedbackActive`.

rogerhutchings: It should be true if feedback is disabled, OR if feedback is enabled but this subject doesn't have any feedback metadata (the latter is a nested expression)

eatyourgreens: OK, so if feedback is enabled and the subject has feedback, then #4078 is triggered because nothing is submitted.

PR #4081, opened 2017-10-05
ResetPasswordPage ES6 conversion

### LINKED ISSUE #2255: Password reset issues
We've been getting a lot of people writing to contact@zoo to tell us they are having issues with the password reset. Most of these are probably people making mistakes with their usernames or something, but there are two things that could be fixed that currently lead to confusion:

~~1) You can't go through the password reset process if you're already logged-in. You just get an error.~~

1) You shouldn't be able to attempt password reset when logged-in. We should show a message that lets the user know they are actually logged in.

2) Once you log in after resetting you password, you still see a white page that just says "You have successfully reset your password, please login to get started." We should instead redirect people to their profile page, or projects view.

jelliotartz:
# Describe your changes.
- Fixes #2255.
- If a logged-in user navigates to `/reset-password`, page displays a message informing the user that they are logged in, and to please log out if they would like to change their password.
- After a user resets their password and logs in with their new password, they are redirected to `/projects`.
- Converts ResetPasswordPage component to ES6.
- Adds unit tests.
- [Staged here](https://reset-password-es6-conversion.pfe-preview.zooniverse.org/).
  - Note: If you go to this branch and enter your email to reset your password, the link in the email will send you to `pfe-preview.zooniverse.org/reset-password?reset_password_token=...`, not `reset-password-es6-conversion.pfe-preview.zooniverse.org/`. You will need to copy/paste the token back to the staging branch to complete the reset-password process on this branch.

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/pages/reset-password.jsx
```diff
@@ -0,0 +1,222 @@
...
+        });
+        alert(resolve => <LoginDialog onSuccess={resolve} />);
+        this.context.router.push('/projects');
+      })
+      .catch((error) => {
+        this.setState({ resetError: error });
+      });
+  }
+
+  renderNewPasswordForm() {
```
srallen: Maybe just make this its own functional component?

app/pages/reset-password.jsx
```diff
@@ -0,0 +1,222 @@
...
+              className="form-help error"
+              component="small"
+              content="resetPassword.resetError"
+            />}
+        </p>
+      </form>
+    );
+  }
+
+  renderSubmitEmailForm() {
```
srallen: Make this its own functional component.

app/pages/reset-password.jsx
```diff
@@ -0,0 +1,222 @@
...
+        this.context.router.push('/projects');
+      })
+      .catch((error) => {
+        this.setState({ resetError: error });
+      });
+  }
+
+  renderNewPasswordForm() {
+    return (
+      <form method="POST" onSubmit={this.handlePasswordResetSubmit}>
```
srallen: The method isn't really being used here since the onSubmit default behavior is being prevented. The method prop could be removed.

app/pages/reset-password.jsx
```diff
@@ -0,0 +1,222 @@
...
+  }
+}
+
+ResetPasswordPage.propTypes = {
+  location: PropTypes.shape({
+    query: PropTypes.shape({
+      reset_password_token: PropTypes.string
+    })
+  }),
+  user: PropTypes.shape({})
```
srallen: This makes it look like some PropType definitions are missing. It's ok for this to be just `PropTypes.object` if all we're doing is checking for the presence of the user. If eslint warning is bothersome, then you can put in an inline exception.

app/pages/reset-password.spec.js
```diff
@@ -0,0 +1,71 @@
...
+    assert.equal(ResetPasswordPageContainer.length, 1);
+  });
+
+  describe('#renderSubmitEmailForm', () => {
+    it('does not allow a signed in user to submit their email', () => {
+      wrapper.setProps({ user });
+      assert.equal(wrapper.find('Translate').prop('content'), 'resetPassword.loggedInDialog');
+    });
+
+    it('renders an email input form for a not logged in user', () => {
```
srallen: When you extract out the forms into their own functional components you can rework these tests to check for the presence of the correct form.

app/pages/reset-password.jsx
srallen: The partials folder should go away eventually. New components shouldn't go into that folder and existing components in there should move closer to where they're actually used. This folder is left over from the organizational structure when the app first launched, but it doesn't make sense anymore. 

The components folder is intended for reusable components.

I'd prefer them as separate files. You could make a folder for this page and its set of components for organizational purposes.

jelliotartz:
Was a little unsure about how to properly test the form components - should I create separate spec files for them? or is just checking which component renders, based on what props are passed in, enough?

app/pages/reset-password/reset-password.spec.js
```diff
@@ -0,0 +1,50 @@
...
+    assert.equal(ResetPasswordPageContainer.length, 1);
+  });
+
+  describe('SubmitEmailForm', () => {
+    it('renders without crashing', () => {
+      assert.equal(wrapper.find('SubmitEmailForm').length, 1);
+    });
+
+    it('renders with no user if none is passed in', () => {
+      assert.equal(wrapper.find('SubmitEmailForm').nodes[0].props.user, null)
```
srallen: If you find yourself trying to use `node[i]` in your test, you probably either need to use a better selector to find what you're trying to test or you are trying to test something in a child component and this test should be written for a set a tests for that child component.

jelliotartz:
Added unit tests for the success, error, and in progress props :)

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
+
   it('renders a new password form for a user', () => {
-    const wrapper = shallow(<NewPasswordForm />);
     assert.equal(wrapper.find('form').length, 1);
   });
 
   it('should call onSubmit when user submits a new password', () => {
     const newPasswordSpy = sinon.spy();
-    const wrapper = mount(<NewPasswordForm onSubmit={newPasswordSpy} />);
+    wrapper = mount(<NewPasswordForm onSubmit={newPasswordSpy} />);
```
srallen: You shouldn't need to mount this to test a spy. The spy can be setup from the beginning at in a const at the start of the describe block and passed in as a prop in the initial shallow render.

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
     wrapper.find('button').simulate('submit');
```
srallen: The `onSubmit` handler is on the form, not the button, so you'll want to `wrapper.find('form').simulate('submit').

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
     assert.equal(newPasswordSpy.calledOnce, true);
   });
+
+  it('conditionally shows and hides an in-progress icon', () => {
+    wrapper.setProps({ inProgress: true });
+    assert.equal(wrapper.find('i').length, 1);
```
srallen: Let's be more specific about which `<i>` you're trying to find here: `wrapper.find('i.form-help')`

Enzyme uses a selectors in a very similar way to jQuery, so it'd be good to read up on that.

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
+    wrapper.setProps({ inProgress: false });
+    assert.equal(wrapper.find('i').length, 0);
```
srallen: Same comment about specificity here.

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
+  });
+
+  it('conditionally shows and hides a reset success icon', () => {
+    wrapper.setProps({ resetSuccess: true });
+    assert.equal(wrapper.find('i').length, 1);
```
srallen: Same comment about specificity

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
+    wrapper.setProps({ resetSuccess: false });
+    assert.equal(wrapper.find('i').length, 0);
```
srallen: Same comment about specificity

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -5,15 +5,41 @@ import sinon from 'sinon';
...
+  });
+
+  it('conditionally shows and hides a reset password error message', () => {
+    wrapper.setProps({ resetError: 'test error message' });
+    assert.equal(wrapper.find('Translate').last().prop('content'), 'resetPassword.resetError');
+    wrapper.setProps({ resetError: null });
+    assert.notEqual(wrapper.find('Translate').last().prop('content'), 'resetPassword.resetError');
```
srallen: This doesn't exactly tell us that the error message isn't displaying, but just that the last Translate component found doesn't equal the expected content. One way to clarify this is to maybe test the number of Translate components expected.

app/pages/reset-password/submit-email-form.spec.js
```diff
@@ -32,4 +32,27 @@ describe('SubmitEmailForm', () => {
...
+    assert.equal(wrapper.find('i').length, 0);
+  });
+
+  it('conditionally shows and hides an email success icon and message', () => {
+    wrapper.setProps({ emailSuccess: true });
+    assert.equal(wrapper.find('i').length, 1);
+    assert.equal(wrapper.find('Translate').last().prop('content'), 'resetPassword.emailSuccess');
+    wrapper.setProps({ emailSuccess: false });
+    assert.equal(wrapper.find('i').length, 0);
+    assert.notEqual(wrapper.find('Translate').last().prop('content'), 'resetPassword.emailSuccess');
```
srallen: Same comment about testing the expected number of Translate components to be found.

app/pages/reset-password/submit-email-form.spec.js
```diff
@@ -32,4 +32,27 @@ describe('SubmitEmailForm', () => {
...
+  });
+
+  it('conditionally shows and hides an email error message', () => {
+    wrapper.setProps({ emailError: 'test error message' });
+    assert.equal(wrapper.find('Translate').last().prop('content'), 'resetPassword.emailError');
+    wrapper.setProps({ emailError: null });
+    assert.notEqual(wrapper.find('Translate').last().prop('content'), 'resetPassword.emailError');
```
srallen: Same comment about testing the expected number of Translate components to be found.

app/pages/reset-password/new-password-form.spec.js
```diff
@@ -0,0 +1,45 @@
...
+    wrapper = shallow(<NewPasswordForm />);
+  });
+
+  it('renders a new password form for a user', () => {
+    assert.equal(wrapper.find('form').length, 1);
+  });
+
+  it('should call onSubmit when user submits a new password', () => {
+    const newPasswordSpy = sinon.spy();
+    wrapper = mount(<NewPasswordForm onSubmit={newPasswordSpy} />);
```
srallen: Same comments here about not needing mount, when to setup the spy, and simulated the onSubmit using the form.

jelliotartz:
Thanks, I appreciate your comments on these tests, and specifically on sinon spies - it's been good practice writing spy and stub tests lately!

jelliotartz:
Rebase done, thanks!

Issue #4041, opened 2017-09-13
Don't let an admin accidentally make a default workflow inactive


srallen:
## Expected behavior
Either don't allow a default workflow to be made inactive in the admin UI or remove it as default when made inactive. Current behavior in the project builder UI is to remove it as default.

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Admin UI allows admins to make any workflows active or inactive at will. This can break a project if the default workflow is made inactive.

srallen:
We probably shouldn't allow admins to make default workflows inactive because we should be in conversation with the project owner and collaborators.

srallen:
Alternatively we could warn the admin user what this change will do and ask for confirmation.

simoneduca:
How do you feel about combining both your suggestions: having a check that prevents a default workflow to be made inactive and if whenever that happens an alert would say something like "You can't make a default workflow inactive. Perhaps you could change default workflow first."

srallen:
@simoneduca We can definitely think about what the best UI is for that and I think folks who use the Admin interface the most often should weigh in. Perhaps @mrniaboc would have a preference. 

Also, I think this is going to be on deck for @jelliotartz to tackle in the next week or so. Thanks.

mrniaboc:
I'm generally in favour of admins having as much power as possible, but being super responsible with it. Admins should never change something that will have a major effect on the project without consulting the owner first, but they should be able to do everything the own can.

jelliotartz:
Unless anyone objects, I'll create an alert to display if an admin user attempts to make the default workflow for a project inactive, but will continue to allow this functionality to exist, if the admin user confirms that they want to do it. Happy to consider alternatives if others weigh in :)

PR #4092, opened 2017-10-12
WorkflowToggle ES6 conversion

jelliotartz:
# Describe your changes
- Towards #4041.
- Converts WorkflowToggle component from coffeescript to ES6.
- Adds unit tests.
- Removes project prop that wasn't being used.
- Adds Translation and moves label text to `locales/en`.
- [Staged here](https://workflow-toggle-es6-conversion.pfe-preview.zooniverse.org/).

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,67 @@
...
+  constructor(props) {
+    super(props);
+    this.handleToggleActive = this.handleToggleActive.bind(this);
+    this.state = {
+      error: null,
+      setting: {}
+    };
+  }
+
+  handleToggleActive(property, value) {
```
srallen: Can this just be named `handleToggle`?

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,67 @@
...
+    const setting = workflow[this.props.field];
+
+    return (
+      <span>
+        { this.props.workflow.id } - { this.props.workflow.display_name}:
+        <label>
+          <input
+            type="checkbox"
+            name={this.props.field}
+            value={setting}
```
srallen: The value is being set from a property from the workflow prop. The input onChange handler is also setting it in the state, but `this.state.setting` isn't being used anywhere.

There are two ways to handle inputs in React: either make them controlled or uncontrolled or in other words either have React handle the state of the input or have the native DOM behavior handle the state of the input. There's sort of a mix of both ways going on here. [Read more about controlled inputs](https://reactjs.org/docs/forms.html). We need to pick one or the other, otherwise, bugs can arise because we're not sure which state is accurate, the workflow prop vs the local component state.

I would suggest simplifying the state a bit here too. We don't want to exactly replicate what the `SetToggle` mixin did. Mixins were a precursor to higher order components, so this mixin was in the past used in several components and could be managing the state of several toggles. We're removing it now, so we're not really managing multiple settings at a time anymore. The component state could just be the current value of the checkbox, if you decide to go the route of a controlled input.

There also has to be consideration of not being out of sync of the state of the workflow too.

Let me know when you've read up on this and we discuss about how to solve all of these state management issues.

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,67 @@
...
+            checked={setting}
+            onChange={this.handleToggleActive.bind(this, this.props.field, !setting)}
```
srallen: You shouldn't need to bind this to the handler. The DOM Event object will have what you need ( here's a generic example is of a controlled input):
```
onChange(event) {
  const checked = event.target.checked;

  this.setState({ checked })
}

render(
  return (
    <input type="checkbox" name="active" value="active" checked={this.state.checked} onChange={this.onChange} />
  )
)
```

app/components/workflow-toggle.spec.js
```diff
@@ -0,0 +1,39 @@
...
+};
+
+describe('WorkflowToggle', () => {
+  let wrapper;
+  let handleToggleActiveStub;
+
+  before(() => {
+    handleToggleActiveStub = sinon.stub(WorkflowToggle.prototype, 'handleToggleActive');
+    wrapper = shallow(<WorkflowToggle project={project} workflow={workflow} field="active" />);
+    handleToggleActiveStub.restore();
```
srallen: Calling restore here shouldn't be necessary. You haven't done anything with the stub yet.

jelliotartz:
I removed the sinon stub test for handleToggle in the spec file, since this is now handled in ProjectStatus - no spec file currently exists for ProjectStatus, will make one if you'd like me to, but wasn't sure if that is outside of the scope of this PR and testing the WorkflowToggle component.

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,36 @@
...
+  handleToggle: () => {},
+  name: null,
+  workflow: {}
+};
+
+WorkflowToggle.propTypes = {
+  checked: React.PropTypes.bool,
+  handleToggle: React.PropTypes.func,
+  name: React.PropTypes.string,
+  workflow: React.PropTypes.object // eslint-disable-line react/forbid-prop-types
```
srallen: In this case, this could actually be a validation using the shape propType because `workflow.id` and workflow.display_name` are used in this component. It also makes sense to make it required.

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,36 @@
...
+          onChange={handleToggle}
+        />
+        <Translate content="workflowToggle.label" />
+      </label>
+    </span>
+  );
+}
+
+WorkflowToggle.defaultProps = {
+  checked: null,
```
srallen: The default should be false here.

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,36 @@
...
+  handleToggle: () => {},
+  name: null,
```
srallen: Default should be an empty string.

app/components/workflow-toggle.jsx
```diff
@@ -0,0 +1,36 @@
...
+
+const WorkflowToggle = ({ workflow, handleToggle, name, checked }) => {
+  return (
+    <span>
+      { workflow.id } - { workflow.display_name}:
+      <label>
+        <input
+          type="checkbox"
+          name={name}
+          value={checked}
```
srallen: [Checkbox input values should be strings](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/checkbox)

app/pages/admin/project-status.jsx
```diff
@@ -72,6 +64,24 @@ class ProjectStatus extends Component {
...
+    let selected = event.target.value;
+    selected = selected === 'none' ? undefined : selected;
+    return workflow.update({ 'configuration.level': selected }).save()
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
+  handleToggle(event, workflow) {
+    this.setState({ error: null });
+    let checked = event.target.checked;
```
srallen: The assignment of this variable doesn't change, so it could be a `const`

app/components/workflow-toggle.jsx
jelliotartz: Ok to go with "checked" as the string for input value here?

srallen: Check out the example in the MDN article, along with the explanation: 
```
  <div>
    <input type="checkbox" id="subscribeNews" name="subscribe" value="newsletter">
    <label for="subscribeNews">Subscribe to newsletter?</label>
  </div>
```

> A DOMString representing the value of the checkbox. This is never seen on the client-side, but on the server this is the value given to the data submitted with the checkbox's name. In this example, we've got a name of subscribe, and a value of newsletter. When the form is submitted, the data name/value pair will be subscribe=newsletter. If the value attribute was omitted, the submitted data would be given a default value of on, so the submitted data in that case would be subscribe=on.

However, we're not using the normal browser behavior (no form is actually being submitted). So the value could be a string that makes sense for what the checkbox is toggling or it could be left off entirely. You can decide which makes the most sense in this case. 

I don't think the original component got this quite right either, so it's good to be mindful of errors in the old code.

jelliotartz: Cool, thanks. I was thinking it wasn't necessary, but wasn't sure if it was included originally because of best practices, or for some other reason. I'll remove it, since that seems like the cleanest solution.

jelliotartz:
Rebased :)

PR #4096, opened 2017-10-14
Update mock data file

simoneduca:
Describe your changes.
Add user and preferences

### DIFF
--- a/app/pages/dev-classifier/mock-data.coffee
+++ b/app/pages/dev-classifier/mock-data.coffee
@@ -560,7 +560,17 @@ project = apiClient.type('projects').create
   experimental_tools: []
 
 preferences = apiClient.type('project_preferences').create
-  preferences: {}
+  preferences: [
+    {
+      activity_count: 'one million'
+      email_communication: false
+      id: 'MOCK_PREFERENCES_ID_FOR_CLASSIFIER'
+    }
+  ]
+
+user = apiClient.type('users').create
+  id: 'MOCK_USER_ID_FOR_CLASSIFIER'
+  login: 'MOCK_USER_LOGIN_FOR_CLASSIFIER'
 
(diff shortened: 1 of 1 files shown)


simoneduca:
Sorry, my bad. I changed the code order to be alphabetical, to make it easier to look for the definitions. Fix coming.

eatyourgreens:
Can you please test changes before committing and pushing?

eatyourgreens:
One small question about `activity_count`. Otherwise, looks good. However, I don't think `mockData.user` is used in the [dev classifier](/zooniverse/Panoptes-Front-End/blob/master/app/pages/dev-classifier/index.cjsx#L62-L64).

simoneduca:
You're right. I was thinking of mock-data as a general purpose file, not restricted to dev-classifier. If that's not its intended purpose, perhaps I should just remove `user`.

eatyourgreens:
`mock-data` is the config script for the dev classifier. It sets up a mock project, workflow, subject and classification for testing out the classification tools.

If you're declaring variables to use in tests, [declare those with the test functions themselves](/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/home/home-workflow-buttons.spec.js#L26-L28).

simoneduca:
Makes sense. I'll remove the polluting `user` then.

eatyourgreens:
Are the new preferences used by the dev classifier? It seemed to be working without them, but I haven't looked at it for a while so maybe I'm missing something.

simoneduca:
Not really https://github.com/zooniverse/Panoptes-Front-End/blob/5bbd29f44fa354a735f0a9edcb08dc4ca26de00c/app/pages/dev-classifier/index.cjsx#L64
Feels like this PR wasn't really necessary. I think I'll jut close it.

PR #4100, opened 2017-10-17
Workflow default dialog alert

### LINKED ISSUE #4041: Don't let an admin accidentally make a default workflow inactive

jelliotartz:
# Describe your changes
- Fixes #4041.
  - Related: #4092
- Adds a modal popup (WorkflowDefaultDialog) when an admin user sets a project's default workflow to inactive on the ProjectStatus page.
  - RFC: styling on modal popup is currently pretty generic. Not sure what we'd like, would appreciate input.
- Adds an asterisk on ProjectStatus page to indicate the default workflow, along with explanatory text.
  - I didn't add a Translate component for this text, as I wasn't sure if it was desirable to have some of the text in Translate and some not, and if moving all of ProjectStatus' text to Translate was outside of the scope of this PR.
- RFC: if a user clicks the background or presses escape, the workflow retains its default status, but `workflow.active` still gets toggled (ie, it is still possible to have a default workflow that is inactive). Not sure if this is the desired UX, would appreciate input.
- Still need to add tests for:
  - RFC: Do I need to add tests for the conditional display of WorkflowDefaultDialog modal in ProjectStatus?
  - ~~Content of WorkflowDefaultDialog.~~
- [Staged here](https://default-workflow-alert.pfe-preview.zooniverse.org/).

# Review Checklist

- [x] Does it work on mobile?

## Optional

### DIFF
--- a/app/components/workflow-default-dialog.jsx
+++ b/app/components/workflow-default-dialog.jsx
@@ -1,14 +1,12 @@
 import React from 'react';
+import Translate from 'react-translate-component';
 
 const WorkflowDefaultDialog = ({ onSuccess }) => {
 
   return (
     <div>
-      <div>
-        You are about to make the default workflow inactive,
-        which will remove the 'default' setting from this workflow.
-        Go to project builder => workflows to set a default workflow.
-      </div>
+      <Translate content="workflowDefaultDialog.text" />
+      <br />
       <button type="submit" onSubmit={onSuccess}>ok</button>
     </div>
   )
--- a/app/components/workflow-default-dialog.jsx
+++ b/app/components/workflow-default-dialog.jsx
@@ -1,0 +1,17 @@
+import React from 'react';
+
+const WorkflowDefaultDialog = ({ onSuccess }) => {
+
+  return (
+    <div>
+      <div>
+        You are about to make the default workflow inactive,
+        which will remove the 'default' setting from this workflow.
+        Go to project builder => workflows to set a default workflow.
+      </div>
+      <button type="submit" onSubmit={onSuccess}>ok</button>
+    </div>
+  )
+}
+
+export default WorkflowDefaultDialog;
--- a/app/components/workflow-default-dialog.jsx
+++ b/app/components/workflow-default-dialog.jsx
@@ -1,15 +1,20 @@
 import React from 'react';
 import Translate from 'react-translate-component';
 
-const WorkflowDefaultDialog = ({ onSuccess }) => {
+const WorkflowDefaultDialog = ({ onSuccess }) => (
+  <div>
+    <Translate content="workflowDefaultDialog.text" />
+    <br />
+    <button type="submit" onSubmit={onSuccess}>ok</button>
+  </div>
+  );
 
-  return (
-    <div>
-      <Translate content="workflowDefaultDialog.text" />
-      <br />
-      <button type="submit" onSubmit={onSuccess}>ok</button>
-    </div>
-  )
-}
+WorkflowDefaultDialog.defaultProps = {
+  onSuccess: () => {}
+};
+
+WorkflowDefaultDialog.propTypes = {
+  onSuccess: React.PropTypes.func
+};
 
 export default WorkflowDefaultDialog;
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -22,6 +22,7 @@ class ProjectStatus extends Component {
     this.renderError = this.renderError.bind(this);
     this.renderWorkflows = this.renderWorkflows.bind(this);
     this.handleToggle = this.handleToggle.bind(this);
+    this.getProjectAndWorkflows = this.getProjectAndWorkflows.bind(this);
 
     this.state = {
       project: null,
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -10,7 +10,7 @@ import VersionList from './project-status/version-list';
 import ExperimentalFeatures from './project-status/experimental-features';
 import Toggle from './project-status/toggle';
 import RedirectToggle from './project-status/redirect-toggle';
-import WorkflowDefaultDialog from './workflow-default-dialog'
+import WorkflowDefaultDialog from './workflow-default-dialog';
 
 class ProjectStatus extends Component {
   constructor(props) {
@@ -18,7 +18,6 @@ class ProjectStatus extends Component {
     this.onChangeWorkflowLevel = this.onChangeWorkflowLevel.bind(this);
     this.getWorkflows = this.getWorkflows.bind(this);
     this.forceUpdate = this.forceUpdate.bind(this);
-    this.renderError = this.renderError.bind(this);
     this.renderWorkflows = this.renderWorkflows.bind(this);
     this.handleToggle = this.handleToggle.bind(this);
     this.getProjectAndWorkflows = this.getProjectAndWorkflows.bind(this);
@@ -87,22 +86,18 @@ class ProjectStatus extends Component {
   }
 
   handleDialogSuccess(workflow) {
-    const promises = [
-      workflow.update({ active: false }).save(),
-      this.state.project.update({ 'configuration.default_workflow': undefined }).save()
-    ];
-
-    Promise.all(promises)
+    this.state.project.update({ 'configuration.default_workflow': undefined }).save()
+      .then(() => {
+        workflow.update({ active: false }).save()
+          .then(() => {
+            this.getProjectAndWorkflows();
+          })
+          .catch(error => this.setState({ error }));
+      })
       .then(() => {
         this.setState({ dialogIsOpen: false });
-        this.getProjectAndWorkflows();
       })
-      .catch(error => {
-        this.setState({
-          dialogIsOpen: false,
-          error
-        });
-      });
+      .catch(error => this.setState({ error }));
   }
 
   handleToggle(event, workflow) {
@@ -114,16 +109,9 @@ class ProjectStatus extends Component {
       this.setState({ dialogIsOpen: true });
     }
 
-    if (defaultWorkflowId !== workflow.id) {
-      workflow.update({ active: isChecked })
-        .save()
+    if ((defaultWorkflowId !== workflow.id) || (defaultWorkflowId === workflow.id && !workflow.active)) {
+      workflow.update({ active: isChecked }).save()
         .catch(error => this.setState({ error }));
-    }
-  }
-
-  renderError() {
-    if (this.state.error) {
-      return <div>{this.state.error}</div>;
     }
   }
 
@@ -143,7 +131,7 @@ class ProjectStatus extends Component {
                 workflow={workflow}
                 name="active"
                 checked={workflow.active}
-                handleToggle={(event) => this.handleToggle(event, workflow)}
+                handleToggle={event => this.handleToggle(event, workflow)}
               />{' | '}
               {this.state.dialogIsOpen &&
                 <WorkflowDefaultDialog
@@ -212,7 +200,8 @@ class ProjectStatus extends Component {
           <small>The workflow level dropdown is for the workflow assignment experimental feature.</small>
           <br />
           <small>An asterisk (*) denotes a default workflow.</small>
-          {this.renderError()}
+          {this.state.error &&
+            `Error ${this.state.error.status}: ${this.state.error.message}`}
           {this.renderWorkflows()}
         </div>
         <hr />
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -1,5 +1,6 @@
 import React, { Component } from 'react';
 import apiClient from 'panoptes-client/lib/api-client';
+import Dialog from 'modal-form/dialog';
 
 import ProjectIcon from '../../components/project-icon';
 import getWorkflowsInOrder from '../../lib/get-workflows-in-order';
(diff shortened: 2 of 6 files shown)


app/components/workflow-default-dialog.jsx
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import Translate from 'react-translate-component';
+
+const WorkflowDefaultDialog = ({ onSuccess }) => (
+  <div>
+    <Translate content="workflowDefaultDialog.text" />
```
srallen: I'm not sure if this needs to be translated. The admin pages are only for Zooniverse team members.

app/pages/admin/project-status.jsx
```diff
@@ -25,7 +27,8 @@ class ProjectStatus extends Component {
       project: null,
       error: null,
       usedWorkflowLevels: [],
-      workflows: []
+      workflows: [],
+      workflowSetting: {}
```
srallen: Is this going to be used somewhere? I don't see it used in the code below.

app/pages/admin/project-status.jsx
```diff
@@ -76,10 +79,28 @@ class ProjectStatus extends Component {
...
+    const defaultWorkflowId = this.state.project.configuration.default_workflow;
+
+    const promises = [this.getWorkflows(), this.getProject()];
+
+    if (defaultWorkflowId === workflow.id && workflow.active) {
+      Dialog.alert(
+        <WorkflowDefaultDialog closeButton={true} required={true} />
+      )
+      .then(() => {
+        Promise.all(promises)
```
srallen: This is related to your question in the PR: 
> if a user clicks the background or presses escape, the workflow retains its default status, but workflow.active still gets toggled (ie, it is still possible to have a default workflow that is inactive). Not sure if this is the desired UX, would appreciate input.

My suggestion to use `Promise.all()` was for the project _update_ and the workflow _update_. If a workflow is active and the project default, we want to prompt the admin user asking to confirm the switch and if they confirm, then remove that workflow as the default and make it inactive. If they cancel, then no action is taken.

At this point, you shouldn't first need to do a get request for the project and workflows again as you should already have them since they are called in `componentDidMount`. The get request for the project and workflows should be after the successful `Promise.all()`. What's defined in `componentDidMount` could be moved into a class method to be reused here as well as in `componentDidMount`.

app/pages/admin/project-status.jsx
```diff
@@ -76,10 +79,28 @@ class ProjectStatus extends Component {
...
+          .then(() => {
+            if (!this.state.error) {
```
srallen: If there is an error, then the promise is rejected, which is what `catch()` is for.

app/pages/admin/project-status.jsx
```diff
@@ -76,10 +79,28 @@ class ProjectStatus extends Component {
...
+          })
+          .catch(error => this.setState({ error }));
+      })
+      .catch(error => this.setState({ error }));
+    }
 
-    return workflow.update({ 'active': checked }).save()
-      .then(() => this.getWorkflows())
-      .catch(error => this.setState({ error }))
+    workflow.update({ active: checked }).save();
```
srallen: Also related to the question in the PR: `if (defaultWorkflowId !== workflow.id)` then update the workflow. We don't want to always update the active state for the workflow.

jelliotartz:
Thanks :) Regarding the tests, do I need to create a spec file for ProjectStatus and then check whether WorkflowDefaultDialog component renders as expected, i.e. simulate a click on WorkflowToggle, set state so that `defaultWorkflowId === workflow.id && workflow.active`, then test whether the modal is present when it's supposed to be? (and not present when it shouldn't be there?)

app/pages/admin/project-status.jsx
```diff
@@ -76,10 +83,30 @@ class ProjectStatus extends Component {
...
+          .catch(error => this.setState({ error }));
+      })
+      .catch(error => this.setState({ error }));
+    }
 
-    return workflow.update({ 'active': checked }).save()
-      .then(() => this.getWorkflows())
-      .catch(error => this.setState({ error }))
+    if (defaultWorkflowId !== workflow.id) {
+      workflow.update({ active: checked }).save();
```
srallen: This should still have a catch

jelliotartz:
I changed the `.catch` block on `Dialog.alert` to `console.error` instead of `this.setState({ error })` because clicking on the background or typing escape to cancel the modal is throwing this error: 
image

`console.error` throws [these errors], not sure if they're any better... 
image

Not sure if there's a better way to gracefully handle the onCancel functionality without throwing errors, although the behavior of aborting a check upon reading the modal message is now working as desired when the user hits escape or clicks the background.

jelliotartz:
RFC: I added an `id` on the `Dialog` success and cancel buttons, so that I could target them in unit tests - happy to change this if there's a better or preferred way to target buttons that don't have classes (or to change the ids to classes).

app/pages/admin/project-status.jsx
```diff
@@ -10,6 +10,7 @@ import VersionList from './project-status/version-list';
 import ExperimentalFeatures from './project-status/experimental-features';
 import Toggle from './project-status/toggle';
 import RedirectToggle from './project-status/redirect-toggle';
+import WorkflowDefaultDialog from './workflow-default-dialog'
```
srallen: Non-blocking, but this is missing a semi-colon.

app/pages/admin/project-status.jsx
```diff
@@ -164,6 +210,8 @@ class ProjectStatus extends Component {
         <div className="project-status__section">
           <h4>Workflow Settings</h4>
           <small>The workflow level dropdown is for the workflow assignment experimental feature.</small>
+          <br />
+          <small>An asterisk (*) denotes a default workflow.</small>
           {this.renderError()}
```
srallen: There's a React error being thrown, `Error: Objects are not valid as a React child` when an error is set to the component because this is expecting a `this.state.error` to be a string. We changed how errors are handled in the javascript client to send along the full error object instead, so how the error renders should be modified too. Perhaps the status and message together?:
```
{`Error ${this.state.error.status}: ${this.state.error.message}`}
```

app/pages/admin/project-status.jsx
```diff
@@ -73,13 +82,43 @@ class ProjectStatus extends Component {
...
+    this.setState({ dialogIsOpen: false });
+  }
+
+  handleDialogSuccess(workflow) {
+    const promises = [
+      workflow.update({ active: false }).save(),
+      this.state.project.update({ 'configuration.default_workflow': undefined }).save()
+    ];
+
+    Promise.all(promises)
```
srallen: I'm rethinking the use of `Promise.all()` since the possibility of a failure condition on the project PUT. We don't want the workflow changed if updating the default workflow fails. It made sense initially to have a single `then()` which `Promise.all()` allows us, but I'm thinking this will be safer:
```
this.state.project.update({ 'configuration.default_workflow': undefined }).save()
  .then(() -> {
     workflow.update({ active: false }).save()
      .then(this.getProjectAndWorkflows() // Only get project and workflows after successful project and workflow PUT
     ).catch(// its own catch);
  }).catch(// its own catch)
  .then(() => {
    this.setState({ dialogIsOpen: false }); 
  })
```
This is just a rough idea without testing, so it might need tweaks.

app/pages/admin/project-status.jsx
```diff
@@ -73,13 +82,43 @@ class ProjectStatus extends Component {
...
+    const defaultWorkflowId = this.state.project.configuration.default_workflow;
 
-    return workflow.update({ 'active': checked }).save()
-      .then(() => this.getWorkflows())
-      .catch(error => this.setState({ error }))
+    if (defaultWorkflowId === workflow.id && workflow.active) {
+      this.setState({ dialogIsOpen: true });
+    }
+
+    if (defaultWorkflowId !== workflow.id) {
```
srallen: I went back and looked at the lab code and if a project isn't live, then the default workflow button can still be enabled for inactive workflows. I think this should change, but to make sure we still allow admins to toggle the checkbox as needed, this conditional needs modification:
`((defaultWorkflowId !== workflow.id) || (defaultWorkflowId === workflow.id && !workflow.active))`

app/pages/admin/workflow-default-dialog.jsx
```diff
@@ -0,0 +1,39 @@
...
+import Dialog from 'modal-form/dialog';
+
+const WorkflowDefaultDialog = ({ onSuccess, onCancel }) => (
+    <Dialog tag="div" closeButton={true} onCancel={onCancel}>
+      <div>
+        You are about to make the default workflow inactive,
+        which will remove the default workflow setting from this project.
+        The default workflow can be set in the edit workflows page of the project builder.
+      </div>
+      <br />
```
srallen: I missed seeing the `<br />` last review. If the goal is to have a bit of space, then perhaps use a `<p>` around the text instead of a`<div>` and a `<br />`. Paragraphs are also block-level and browsers default to given them margins too.

app/pages/admin/workflow-default-dialog.spec.js
```diff
@@ -0,0 +1,32 @@
...
+    wrapper = shallow(<WorkflowDefaultDialog onSuccess={onSuccessSpy} onCancel={onCancelSpy}/>);
+  });
+
+  it('renders without crashing', () => {
+    const WorkflowDefaultDialogContainer = wrapper.find('div').first();
+    assert.equal(WorkflowDefaultDialogContainer.length, 1);
+  });
+
+  it('calls the onSuccess handler', () => {
+    wrapper.find('button#workflowDefaultDialogSuccess').simulate('click');
```
srallen: I'd prefer not adding CSS ids or classes to elements just for the purposes of testing. Since there are only two buttons using enzyme's `first()` or `last()` is fine or [`at()`](http://airbnb.io/enzyme/docs/api/ShallowWrapper/at.html)

jelliotartz:
I'm seeing some weird behavior when toggling a default workflow from active to inactive - when there is more than one workflow, it toggles the last workflow, rather than the one that was the default. The default setting is getting removed though. Wanted to check and see if you can reproduce this issue though, in case it's a cache bug again (I tried on Firefox though, got same result, so I don't think it is).

Also, on my local, the workflow being passed into `handleDialogSuccess` is always the last workflow. I'm trying different ways to pass the workflow to `handleDialogSuccess`, but so far haven't found a way to modify `onSuccess={this.handleDialogSuccess.bind(null, workflow)}`

app/pages/admin/project-status.jsx
```diff
@@ -85,10 +86,11 @@ class ProjectStatus extends Component {
     this.setState({ dialogIsOpen: false });
   }
 
-  handleDialogSuccess(workflow) {
+  handleDialogSuccess() {
+    const defaultWorkflow = this.state.workflows.filter(workflow => workflow.id === this.state.defaultWorkflowId);
     this.state.project.update({ 'configuration.default_workflow': undefined }).save()
       .then(() => {
-        workflow.update({ active: false }).save()
+        defaultWorkflow[0].update({ active: false }).save()
```
jelliotartz: I changed the name of this variable because of a linter error (`'workflow' is already declared in the upper scope`)

jelliotartz:
Got a working solution to this ^^ issue (with a little input from Mark - thx!) - I think what was happening was that because `dialogIsOpen` is initally false, binding `handleDialogSuccess` to the workflow wasn't doing anything on initial render. When I change state of `dialogIsOpen` in `handleToggle`, I think it's only rerendering `WorkflowDefaultDialog`, and not the whole page - and thus the workflows don't get mapped again, and the bind is getting set on the last workflow every time.

To fix this, I stored the default workflow id in state, and no longer pass it in the onSuccess event. In `handleDialogSuccess` I get the desired workflow by filtering all the workflows and matching the state-stored default workflow id.

app/pages/admin/project-status.jsx
```diff
@@ -85,10 +86,11 @@ class ProjectStatus extends Component {
     this.setState({ dialogIsOpen: false });
   }
 
-  handleDialogSuccess(workflow) {
+  handleDialogSuccess() {
+    const defaultWorkflow = this.state.workflows.filter(workflow => workflow.id === this.state.defaultWorkflowId);
```
jelliotartz: I wanted to call this variable `workflow`, but that raised a linter error (`'workflow' is already declared in the upper scope`), so I went with defaultWorkflow.

app/pages/admin/project-status.jsx
```diff
@@ -85,17 +86,21 @@ class ProjectStatus extends Component {
...
```
jelliotartz: I wanted to call this variable `workflow`, but that raised a linter error (`'workflow' is already declared in the upper scope`), so I went with defaultWorkflow.

srallen: Good use of `array.filter()`, however, we can sidestep storing the default workflow id in the component state since it's already stored in the project resource:
`const defaultWorkflow = this.state.workflows.filter(workflow => workflow.id === this.state.project.configuration.default_workflow);`

app/pages/admin/project-status.jsx
```diff
@@ -106,7 +111,10 @@ class ProjectStatus extends Component {
     const defaultWorkflowId = this.state.project.configuration.default_workflow;
 
     if (defaultWorkflowId === workflow.id && workflow.active) {
-      this.setState({ dialogIsOpen: true });
+      this.setState({
+        defaultWorkflowId,
```
srallen: This is redundant with the `default_workflow` already a property on the project resource which is already stored in the component state.

Issue #4089, opened 2017-10-11
org landing details section follow-up


mcbouslog:
this is issue is specific to the sections below the organization affiliated project cards section.

- [x] stats section functionality
- [x] about section clean-up
- [x] links section functionality
- [x] researcher quote, cycle through projects' researcher quotes

see PR #4037 
continuation of Issue #2970

beckyrother:
Looks great! Thanks Mark!

beckyrother:
Re: your Slack question about quotes – Let's cycle through the project quotes, possibly with a link to each project from that researcher's name. So every time the page reloads it displays a random quote.

PR #4101, opened 2017-10-17
display organization links

mcbouslog:
Towards #4089.

Displays organization urls - first external links with their labels, then social links with fa icons and paths.

Staged - https://org-links.pfe-preview.zooniverse.org/organizations/markb-panoptes/nfnorgtest

# Review Checklist

- [x] Does it work on mobile?

## Optional


app/pages/organizations/organization-page.jsx
```diff
@@ -183,11 +193,39 @@ class OrganizationPage extends React.Component {
             </div>
             <div className="organization-researcher-words">
               <h4 className="organization-details__heading"><Translate content="organization.home.links" /></h4>
-              <ul>
-                <li className="organization-details__link">Blog</li>
-                <li className="organization-details__link"><i className="fa fa-facebook fa-fw" />- @orgFacebook</li>
-                <li className="organization-details__link"><i className="fa fa-twitter fa-fw" />- @orgTwitter</li>
-              </ul>
+              <div className="organization-details__links">
+                {!!rearrangedLinks.length && rearrangedLinks.map((link, i) => {
```
jelliotartz: Curious as to why we need the `!!` at the beginning of `!!reaarrangedLinks.length` here. Does the check not work without it?

app/pages/organizations/organization-page.jsx
```diff
@@ -183,11 +193,39 @@ class OrganizationPage extends React.Component {
...
+                  }
+                  return (
+                    <a
+                      key={i}
+                      className="organization-details__link organization-details__link--external"
+                      href={link.url}
+                      target="_blank"
+                      rel="noopener noreferrer"
+                    >
+                      {link.label}
```
jelliotartz: Design question - have we considered adding a link icon for the external links (maybe, eg, [`fa-link`](http://fontawesome.io/icon/link/)? Having the social icons but nothing for the links looks a little uneven to me, but not a blocker if you like the links without an icon.

app/pages/organizations/organization-page.jsx
eatyourgreens: It's a shorthand for writing `Boolean(rearrangedLinks.length)` and explicitly casts an expression to `true/false`. Here are [some examples](https://www.sitepoint.com/javascript-double-negation-trick-trouble/), including a few that regularly trip people up.

eatyourgreens: Slightly related: sometimes you will come across `x = +someVariable` in JavaScript. The plus sign tells the interpreter to explicitly evaluate `someVariable` as a number.

jelliotartz: Thanks, yeah, I'm familiar with the concept, was more curious as to why it's needed here.

eatyourgreens: That's a really good question. I _think_ it's needed here because there's no explicit `if`, but I would have thought that the `&&` takes care of casting the expressions to Booleans.

srallen: `if (array.length)` is an implicit check on a truthy or falsy value. `!!` casts it to an explicit boolean. Relying on truthy/falsy values can lead to unexpected behavior: https://www.sitepoint.com/javascript-truthy-falsy/

Using explicit values in conditionals is pretty much always preferred in javascript.

srallen: FWIW, I tend to always write explicit length checks on arrays in conditionals, `if (array.length > 0)` or `if (array.length > -1)`, etc.

mcbouslog: I'll change to the explicit length check.
An explicit check or the `!!` solved an issue where if just `rearrangedLinks.length` while it did prevent the `.map` it also returned 0, the length of the empty array, and actually showed the 0 in the DOM.

app/pages/organizations/organization-page.jsx
mcbouslog: good question, I think I like it as is where external links don't have an icon, but defer to @beckyrother's opinion.

beckyrother: That's a great point James, thanks. It would be great to give an indication that the link is leaving the site – let's do icons. We've used [this one](http://fontawesome.io/icon/external-link/) elsewhere.

srallen: I hate to open up the link behavior can of worms, but we should be consistent across the site on link behavior. Do project external links open in a new tab? Should any link open in a new tab? We also don't enforce any particular way links are opened that are submitted in markdown from projects (and in fact let them choose with a kind of markdown).

mcbouslog: currently project sub-nav links open in a new tab, though [the related code](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/project-page.cjsx#L168) looks strange to me, not sure exactly why the links open in a new tab.

srallen: OK, as long as this is consistent with projects, though if the icon is added, it should probably be added there too. 

The project external link is using a named target, so it opens in a new tab, and when clicked again will open in the same tab rather than a new one.

mcbouslog: Ah a named tab, that's a good way to do it, I'll do the same here unless there's a reason not to.

app/pages/organizations/organization-page.jsx
```diff
@@ -183,11 +193,39 @@ class OrganizationPage extends React.Component {
...
+                  if (link.path) {
+                    let iconForLabel;
+                    if (Object.keys(SocialIcons).indexOf(link.site) >= 0) {
+                      iconForLabel = SocialIcons[link.site];
+                    } else {
+                      iconForLabel = 'globe';
+                    }
+                    return (
+                      <a
+                        key={i}
```
mcbouslog: For the key I think I should use `link.key` for external and `link.path` for social links instead of the array index.

app/pages/organizations/organization-page.jsx
```diff
@@ -183,11 +193,39 @@ class OrganizationPage extends React.Component {
...
+                        target="_blank"
+                        rel="noopener noreferrer"
+                      >
+                        <i className={`fa fa-${iconForLabel} fa-fw fa-2x`} />
+                        <span> - @{link.path}</span>
+                      </a>);
+                  }
+                  return (
+                    <a
+                      key={i}
```
mcbouslog: see other comment regarding key and index

jelliotartz:
## PR Update

Good feedback from a number of contributors here. Will summarize action points from inline comments below:
- [ ] `!!rearrangedLinks.length` check: change to `(rearrangedLinks.length > 0)`
- [ ] add link icon to external links, for consistency with social link icons.
  - side note: to maintain consistency with project social and external links, we should probably add icons in project external links as well. Outside of scope for this PR, will open up issue for this.
- [ ] link behavior: both project links and org links open new tabs. Use named target for consistency with project links
- [ ] keys: use `link.key` for external and `link.path` for social links instead of the array index.

mcbouslog:
I think this reflects requested changes, pending final review from @jelliotartz and @beckyrother, and if so to be merged after Autumnwatch's initial night.

app/pages/organizations/organization-page.jsx
```diff
@@ -194,35 +194,26 @@ class OrganizationPage extends React.Component {
...
+                    label = <span> - {link.label}</span>;
                   }
                   return (
                     <a
-                      key={i}
-                      className="organization-details__link organization-details__link--external"
-                      href={link.url}
-                      target="_blank"
+                      key={link.key || link.path}
+                      className="organization-details__link organization-details__link"
```
jelliotartz: Is there a reason `organization-details__link` is repeated in this class?

mcbouslog: nope, forgot to delete!

jelliotartz:
Loaded up my test org here and noticed the Google+ icon margin is almost overlapping with the `-`. Sorry for the huge copy/paste screen shot.

image

css/organization-page.styl
```diff
@@ -101,10 +101,20 @@
...
+    display: flex
     font-size: 1.2em
-    list-style: none
-    margin-bottom: 1.2em
+    margin-left: .8em
+    margin-top: 20px
+    text-decoration: none
+
+  &__icon
+    padding-right: .1em
```
mcbouslog: I think this solves the icon spacing issue, though there may be a more elegant solution

jelliotartz:
## PR Review (Update)
- All issues resolved via action points. Thanks!

## Status
Approved and merging.

PR #4108, opened 2017-10-20
Add a Spanish locale

eatyourgreens:
Adds a locale file for Spanish UI translations.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/locales/es.js
+++ b/app/locales/es.js
@@ -1,0 +1,147 @@
+export default {
+  loading: '(Loading)',
+  classifier: {
+    back: 'Back',
+    backButtonWarning: 'Going back will clear your work for the current task.',
+    done: 'Done',
+    doneAndTalk: 'Done & Talk',
+    next: 'Next',
+    talk: 'Talk',
+    tutorialButton: 'Show the project tutorial',
+    miniCourseButton: 'Restart the project mini-course'
+  },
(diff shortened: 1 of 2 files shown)


app/locales/es.js
```diff
@@ -0,0 +1,147 @@
...
+      about: 'About %(title)s',
+      metadata: {
+        statistics: '%(title)s Statistics',
+        classifications: 'Classifications',
+        volunteers: 'Volunteers',
+        completedSubjects: 'Completed Subjects',
+        subjects: 'Subjects'
+      },
+      talk: {
+        zero: 'Noone is talking about <strong>%(title)s</strong> right now.',
```
jelliotartz: Pretty sure there's a typo here - this should be no one instead of noone, right?

eatyourgreens: 'Noone' in UK English, I think. I guess it will ultimately get translated into `Nadie` for this file, but the English should be double checked for the project home page.

jelliotartz: Ah, sorry, wasn't aware of that difference between American and UK English! Thanks! Happy to make a quick PR on the English text, if we're in favor of it - but don't mean to imply the American 'no one' is correct and UK 'noone' is not :)

PR #4118, opened 2017-10-25
Add check for default workflow

jelliotartz:
# Describe your changes
- Adds a check for presence of configuration attribute (where `default_workflow` is stored) on a project before conditionally rendering an asterisk next to the default workflow on the admin/project-status page.
  - If no default workflow is set, it will display 'No workflows found' (same behavior as before this PR), but now should throw no errors in the console or interfere with the UI.
  - Issues were found with toggles freezing on new projects that have no default workflow set. (tagging @mrniaboc here for reference).
- [Staged here](https://default-workflow-check.pfe-preview.zooniverse.org/).
- Related: https://github.com/zooniverse/Panoptes-Front-End/pull/4100.

# Review Checklist

- [x] Does it work on mobile?

## Optional


jelliotartz:
## PR Update
Was able to produce the same error (can't read property `default_workflow` of undefined) in the console when attempting to toggle the workflow active setting of a test project with one workflow and no default_workflow set - so adding a check for `this.state.project.configuration` in the `handleToggle` function to handle this use case.

jelliotartz:
sure, will do!

PR #4127, opened 2017-10-30
ProjectStatus unit tests

jelliotartz:
## Describe your changes
- Adds tests for the ProjectStatus component in Admin.
- Tests are divided into a group that checks the presence of the basic page layout elements, and a group that focuses on the Workflow Settings section (where I've recently implemented changes  regarding toggling active status of default workflows)
  - RFC: Not sure if testing page layout elements is necessary, or a good practice. Happy to remove if not needed/wanted.
- RFC: So far I haven't been able to write tests that simulate a user clicking the WorkflowToggle checkbox, getting the WorkflowDefaultDialog modal to open, and testing its functionality.
  - Are tests of the modal functionality (eg, user clicks the 'active' toggle for a default workflow, modal opens, user clicks 'ok', and default_workflow gets removed from project) outside the scope of testing the ProjectStatus component?
  - Would appreciate input on a) are tests of modal functionality necessary/desired/good practice? b) if yes, any suggestions on where my syntax is off? 
    - Is there anywhere in the PFE codebase where we use Enzyme to test modal dialogs being triggered in parent components?
    - I thought perhaps I needed to use `mount` instead of `shallow` render, but haven't had success with mount either.
    - Do I need to use a `stub` instead of a `spy`, and control/force the code down a specific path (ie, to click the toggle for a default workflow, then force 'ok'/'cancel')
  - Is it possible that testing modals in Enzyme is difficult/buggy? some SO and github issues I've found while trying to get modal tests working:
    - https://github.com/airbnb/enzyme/issues/252
    - https://stackoverflow.com/questions/45644244/content-of-react-modal-dialogs-is-not-available-to-enzyme-tests-using-mount
    - https://github.com/airbnb/enzyme/issues/535
- Related PRs (my recent work related to ProjectStatus component and toggling default workflow status):
  - #4118 (default workflow existential check)
  - #4100 (default workflow dialog alert)
  - #4092 (WorkflowToggle ES6 conversion)

## Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -25,6 +25,7 @@ class ProjectStatus extends Component {
     this.handleDialogSuccess = this.handleDialogSuccess.bind(this);
 
     this.state = {
+      defaultWorkflowId: null,
       dialogIsOpen: false,
       error: null,
       project: null,
@@ -47,9 +48,13 @@ class ProjectStatus extends Component {
 
     return apiClient.type('projects').get({ slug }).then((projects) => {
       const project = projects[0];
+      const defaultWorkflowId = project.configuration && project.configuration.default_workflow
       // TODO: We ought to improve this ChangeListener replacement
       project.listen('change', this.forceUpdate);
-      this.setState({ project });
+      this.setState({
+        defaultWorkflowId: defaultWorkflowId,
+        project
+      });
       return project;
     });
   }
@@ -108,19 +113,14 @@ class ProjectStatus extends Component {
   handleToggle(event, workflow) {
     this.setState({ error: null });
     const isChecked = event.target.checked;
-    let defaultWorkflowId;
 
-    if (this.state.project.configuration) {
-      defaultWorkflowId = this.state.project.configuration.default_workflow;
-    }
-
-    if (defaultWorkflowId === workflow.id && workflow.active) {
+    if (this.state.defaultWorkflowId === workflow.id && workflow.active) {
       this.setState({
         dialogIsOpen: true
       });
     }
 
-    if ((defaultWorkflowId !== workflow.id) || (defaultWorkflowId === workflow.id && !workflow.active)) {
+    if ((this.state.defaultWorkflowId !== workflow.id) || (this.state.defaultWorkflowId === workflow.id && !workflow.active)) {
       workflow.update({ active: isChecked }).save()
         .catch(error => this.setState({ error }));
     }
@@ -144,7 +144,7 @@ class ProjectStatus extends Component {
                 checked={workflow.active}
                 handleToggle={event => this.handleToggle(event, workflow)}
               />{' | '}
-              {this.state.dialogIsOpen &&
+              {this.state.dialogIsOpen && workflow.id === this.state.defaultWorkflowId &&
                 <WorkflowDefaultDialog
                   onCancel={this.handleDialogCancel}
                   onSuccess={this.handleDialogSuccess}
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -25,7 +25,6 @@ class ProjectStatus extends Component {
     this.handleDialogSuccess = this.handleDialogSuccess.bind(this);
 
     this.state = {
-      defaultWorkflowId: null,
       dialogIsOpen: false,
       error: null,
       project: null,
@@ -48,13 +47,9 @@ class ProjectStatus extends Component {
 
     return apiClient.type('projects').get({ slug }).then((projects) => {
       const project = projects[0];
-      const defaultWorkflowId = project.configuration && project.configuration.default_workflow
       // TODO: We ought to improve this ChangeListener replacement
       project.listen('change', this.forceUpdate);
-      this.setState({
-        defaultWorkflowId: defaultWorkflowId,
-        project
-      });
+      this.setState({ project });
       return project;
     });
   }
@@ -113,14 +108,19 @@ class ProjectStatus extends Component {
   handleToggle(event, workflow) {
     this.setState({ error: null });
     const isChecked = event.target.checked;
+    let defaultWorkflowId;
 
-    if (this.state.defaultWorkflowId === workflow.id && workflow.active) {
+    if (this.state.project.configuration) {
+      defaultWorkflowId = this.state.project.configuration.default_workflow;
+    }
+
+    if (defaultWorkflowId === workflow.id && workflow.active) {
       this.setState({
         dialogIsOpen: true
       });
     }
 
-    if ((this.state.defaultWorkflowId !== workflow.id) || (this.state.defaultWorkflowId === workflow.id && !workflow.active)) {
+    if ((defaultWorkflowId !== workflow.id) || (defaultWorkflowId === workflow.id && !workflow.active)) {
       workflow.update({ active: isChecked }).save()
         .catch(error => this.setState({ error }));
     }
@@ -133,6 +133,11 @@ class ProjectStatus extends Component {
 
     return (
       <ul className="project-status__section-list">
+      {this.state.dialogIsOpen &&
+        <WorkflowDefaultDialog
+          onCancel={this.handleDialogCancel}
+          onSuccess={this.handleDialogSuccess}
+        />}
         {this.state.workflows.map((workflow) => {
           return (
             <li key={workflow.id} className="section-list__item">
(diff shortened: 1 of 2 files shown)


jelliotartz:
## PR Update
Thanks for your input @srallen :) RFC: I was writing a test for rendering the WorkflowDefaultDialog component, and came upon an issue - the component is currently rendered once for every workflow, rather than just once, for the default workflow. (you can test this by inspecting the modal element on a project with multiple workflows).

This seems like an issue that should be resolved, and I've refactored the code in ProjectStatus to only render one WorkflowDefaultDialog component when the modal is triggered (on a separate branch, not currently included here). However, updating ProjectStatus to only render one WorkflowDefaultDialog seems like it's outside of the scope of this PR. But, the issue was discovered in the process of writing the test for the presence of the component, which is within the scope of the PR... 

I pushed up changes that include a passing test for the WorkflowDefaultComponent (it tests whether it renders one for each workflow), with a comment to refactor ProjectStatus so that we only render one component. Happy to either update ProjectStatus and the test within this PR, or to open up a separate PR to address the issue and fix the test there, just wanted to confer before continuing :)

srallen:
Let's get the fix in with this. Just keep the tests and the bug fix as separate commits.

app/pages/admin/project-status.jsx
```diff
@@ -108,19 +113,14 @@ class ProjectStatus extends Component {
...
 
-    if (defaultWorkflowId === workflow.id && workflow.active) {
+    if (this.state.defaultWorkflowId === workflow.id && workflow.active) {
       this.setState({
         dialogIsOpen: true
       });
     }
 
-    if ((defaultWorkflowId !== workflow.id) || (defaultWorkflowId === workflow.id && !workflow.active)) {
+    if ((this.state.defaultWorkflowId !== workflow.id) || (this.state.defaultWorkflowId === workflow.id && !workflow.active)) {
```
jelliotartz: As I was updating this conditional, I was wondering if it's necessary - I tried changing it to just be an `else` attached to the conditional above it, and the UI functionality seems to be the same (tests all still passing as well). 

Would appreciate input on whether it's necessary here to explicitly check for the opposite of the conditions described [above](https://github.com/zooniverse/Panoptes-Front-End/blob/2a53e8b2172c3ed39eaa5f9847839131b1c2ba20/app/pages/admin/project-status.jsx#L123).

jelliotartz:
## PR Update
- Bug fix for modal added on a separate commit.
- I've added unit tests for the custom components, which I put in their own `describe` block.
- I created a 'workflow settings' `describe` block, where I unit test each of the custom components associated with workflow settings, as well as the #onChangeWorkflowLevel method.
- I'm not sure if I need to write tests for other methods in this component, (e.g., project and workflow GET requests, `componentDidMount`, and the event handler methods). Please let me know if more tests need to be written to cover unit testing for this component :)

jelliotartz:
## PR Update
- Added unit tests for all of the event handlers, except for `handleToggle`
- RFC: I've been stuck on how to test the call of `handleToggle` for a while now - pretty sure I'm missing something in the syntax, but can't figure it out.
  - I'm getting `AssertError: expected handleToggle to be called once but was called 0 times`
  - My hunch is that the issue is a syntax error [here](https://github.com/zooniverse/Panoptes-Front-End/pull/4127/commits/088153e834ede8c1a1f87928b2ed593cc5d50212#diff-65d3138def24203beacf09aea3888c99R115), but I could be off base on that.
  - Additionally, this event handler is different in that it's an event that is passed to a custom component, which then calls it via an `onChange` event on its input - I've been looking on [Sinon's docs](https://codeutopia.net/blog/2016/05/23/sinon-js-quick-tip-how-to-stubmock-complex-objects-such-as-dom-objects/) for how to stub complex objects, but so far no luck applying their solutions to this scenario.
  - Would appreciate suggestions on how to test the `handleToggle` event handler :)

app/pages/admin/project-status.spec.js
```diff
@@ -0,0 +1,142 @@
...
+
+  before(() => {
+    onChangeWorkflowLevelSpy = sinon.stub(ProjectStatus.prototype, 'onChangeWorkflowLevel');
+    handleDialogCancelSpy = sinon.stub(ProjectStatus.prototype, 'handleDialogCancel');
+    handleDialogSuccessSpy = sinon.stub(ProjectStatus.prototype, 'handleDialogSuccess');
+    handleToggleSpy = sinon.stub(ProjectStatus.prototype, 'handleToggle');
+
+    wrapper = shallow(<ProjectStatus />);
+
+    wrapper.setState({
```
srallen: This doesn't need to happen in the before. You're setting state in your tests already for the specific condition that's being tested.

app/pages/admin/project-status.spec.js
```diff
@@ -0,0 +1,142 @@
...
+    });
+  });
+
+  it('renders without crashing', () => {
+    assert.equal(wrapper, wrapper);
+  });
+
+  describe('custom components', () => {
+    it('LoadingIndicator renders if project is not in state', () => {
+      wrapper.setState({ project: null });
```
srallen: This doesn't need to happen once you remove the setState out of the before block. You do want to test what is rendered for the initial state. There should also be a unit test for when the workflows state is null (the initial state for workflows)

app/pages/admin/project-status.spec.js
```diff
@@ -0,0 +1,142 @@
...
+      loadingIndicator = wrapper.find('LoadingIndicator');
+      assert.equal(loadingIndicator.length, 1);
+    });
+
+    it('LoadingIndicator does not render if project is in state', () => {
```
srallen: Perhaps put the unit tests that rely upon there being a project in the component state in its own describe block.

app/pages/admin/project-status.spec.js
```diff
@@ -0,0 +1,142 @@
...
+  let handleDialogSuccessSpy;
+  let handleToggleSpy;
+  let event;
+
+
+  before(() => {
+    onChangeWorkflowLevelSpy = sinon.stub(ProjectStatus.prototype, 'onChangeWorkflowLevel');
+    handleDialogCancelSpy = sinon.stub(ProjectStatus.prototype, 'handleDialogCancel');
+    handleDialogSuccessSpy = sinon.stub(ProjectStatus.prototype, 'handleDialogSuccess');
+    handleToggleSpy = sinon.stub(ProjectStatus.prototype, 'handleToggle');
```
srallen: Let's be consistent with variable names. If they're stubs, then name the variables stubs.

jelliotartz:
Thanks, yeah, I was thinking perhaps the issue was actually that I was still getting integration and unit tests mixed up - thanks for clearing that up.

Tests now refactored into `describe` blocks that set state in a rational way - start with the initial null states for project and workflow, test for loading indicator, then add the project mock object, test what should/shouldn't be there, then add workflows, and appropriate unit tests.

app/pages/admin/project-status.spec.js
```diff
@@ -33,68 +33,73 @@ const workflows = [
...
     });
 
-    it('VersionList renders', () => {
+    it('renders a VersionList component', () => {
       const versionListComponent = wrapper.find('VersionList');
       assert.equal(versionListComponent.length, 1);
     });
+
+    it('renders a no workflows found message when no workflow is in component state', () => {
+      const noWorkflowsMessage = wrapper.find('div[children="No workflows found"]');
```
jelliotartz: I was able to target this element using a solution [suggested here](https://github.com/airbnb/enzyme/issues/270), but, as one person comments, this seems tightly coupled to the implementation - would it be better to move this text to a Translate component, and then test whether what is rendered matches what is in the Translate component's `content` prop? ([something more like this](https://github.com/zooniverse/Panoptes-Front-End/blob/ba4b0ca140dccbb325264ff66b0b6af5ddb67ef8/app/pages/security-policy.spec.js#L24))

srallen: We don't need translations here. I'm of the opinion that finding specific DOM elements for a unit test will always be coupled with the implementation. It's unavoidable, however, I would avoid using a prop that is internal to the React API as well as using a selector that looks like a CSS selector when it's not. Here's how I'd write this test and the DOM traversal:
```
      const noWorkflowsMessage = wrapper.find('.project-status__section').at(1).children().find('div');
      assert.equal(noWorkflowsMessage.text(), 'No workflows found');
```
It also better matches the description of the test, which is that it renders a specific message.

jelliotartz:
Thanks, I'll keep that in mind going forward :) 

Just rebased this against master.

PR #4130, opened 2017-10-31
redo logging in ES6 and using google analytics as well as geordi

amyrebecca:
Staged: https://google-analytics.pfe-preview.zooniverse.org/

Fixes # .

Describe your changes.

# Review Checklist

- [x] Does it work on mobile?

## Optional


amyrebecca:
Just noticed that I left some things in `index.ejs` that shouldn't be there, don't merge this yet

amyrebecca:
@wgranger just brought to my attention the other day that this hadn't been merged yet. my apologies!

wgranger:
Before this is merged a change will need to be made to the Zooniverse Global accounts on the Google Analytics interface to account for the dimension map in `ga-log-adaptor.js`

wgranger:
Zooniverse Global (where this is logging) is now set up with the correct custom dimensions. This will be handy when testing this PR.

app/lib/ga-log-adapter.js
```diff
@@ -0,0 +1,48 @@
+const dimensionMap = {
+  experiment: 1,
+  cohort: 2,
+  projectToken: 3,
+  userID: 4,
+  subjectID: 5,
+  workflowID: 6
```
wgranger: Does this need the extra `data` key?

amyrebecca: I just added the key while we were talking yesterday to show how easy it was--it's not actually used yet :)

app/partials/app.cjsx
```diff
@@ -47,6 +50,7 @@ PanoptesApp = createReactClass
 
   componentWillMount: ->
     @geordiLogger = new GeordiLogger
+    @geordiLogger.subscribe(new GeordiLogAdapter(), new GALogAdapter(window.dataLayer)
```
wgranger: This needs a closing parentheses

amyrebecca: oh for goodness-- hang on one second lol

app/lib/ga-log-adapter.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  configure(keys) {
+    this.onRemember(keys);
+  }
+
+  onRemember(newKeys) {
+    Object.keys(newKeys).forEach((key) => {
+      if (Object.keys(dimensionMap).includes(key)) {
+        this.gaLayer('set', `dimension${dimensionMap[key]}`, newKeys[key]);
```
wgranger: It looks like this is firing when the page is loaded, and I'm getting `this.gaLayer is not a function`

amyrebecca: oh that makes sense; that will happen in places where the google analytics tracking code hasn't been installed. on production it should work fine though

wgranger: The error won't let the page load.

wgranger:
The GA layer is found on initialization; however, gaLayer is an array, which is what causes `this.gaLayer is not a function` to appear. This is because `window.dataLayer`, that is passed into the adapter's constructor, is an array.

amyrebecca:
It sounds like a quick workaround might be simply to include the google
analytics code after all, with the appropriate tracking bug. We might be
looking at a version incompatibility between GA and GTM; I never tested the
two in combination before. I'll be OOO until Wednesday morning so if you
need this sooner feel free to fork the PR.

wgranger:
The most recent commit adds GA script to index.ejs. However, there are a couple concerns with the current logging.

- In order to set up GA logging based on the environment, some configuration will need to be made through Google Tag Manager.
- In this PR, events are logged to GA twice. Once through the Geordi client and once through the work on this PR. GA logging should be removed from the Geordi client.
- It looks like custom dimensions will also need to be configured through GTM. 
- Once the changes are made, the `dataLayer` object should be used instead of `window.ga`

srallen:
@amyrebecca @wgranger It was decided in the ops channel today to just go ahead and shut down geordi because it takes up a lot of database storage. So, with that in mind, it seems the work to add Google Analytics might need some refactoring?

amyrebecca:
this is great news. i'll try to find out what my prioritization is so i can get to this as soon as possible, or whether someone else should carry it on from here.

amyrebecca:
@trouille this is the issue blocking the Snapshot WI google analytics stuff

Issue #3469, opened 2017-02-09
Interface for adding/editing a collection description


srallen:
Input box for collection owner or collaborator to add a description. Back end will need to support this first, zooniverse/Panoptes#2170

## Additional information

- **Operating system:**
- **Browser:**

PR #4133, opened 2017-11-01
CollectionCreateForm ES6 conversion

jelliotartz:
## Describe your changes
- Converts CollectionsCreateForm from coffeescript to ES6.
- Adds unit tests.
- Towards #3469.
- [Staged here](https://collection-create-form-es6-conversion.pfe-preview.zooniverse.org/).

## Known Issues
- ~~Eslint is throwing errors that the use of string references (`ref="name"` in `input`) is deprecated.~~  I've been working on replacing the string references with `ref={(node) => { this.name = node; }}`, but so far getting syntax errors. ~~Wanted to get this PR up, but will continue working on refactoring to remove string references from the inputs.~~ (edit: taken care of)
- I'm thinking there may be other parts of this conversion that can be refactored. Input is welcome/appreciated, but wanted to get a PR in as is.

## Review Checklist

- [x] Does it work on mobile?
 

### Optional


jelliotartz:
## PR Update
- Refactored form and `onSubmit` to remove string references and refer to components by setting a property on the `this` object in the reference callback.
- RFC: I've noticed that the `project` prop is a string referring to the projectId. Would it make sense to change the name of the prop [here](https://github.com/zooniverse/Panoptes-Front-End/blob/4b0fabb604949e59a659b00dd325d2040dd014c0/app/collections/collections-manager.jsx#L95) to `projectID` (or `projectId`), and then rename the prop in `CollectionsCreateForm` to more accurately reflect what this prop represents?

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    onSubmitSpy = sinon.stub(CollectionsCreateForm.prototype, 'onSubmit');
+    wrapper = shallow(<CollectionsCreateForm />);
+  });
+
+  it('should render without crashing', () => {
+    assert.equal(wrapper, wrapper);
+  });
+
+  it('should render a text input for collection name', () => {
+    const collectionNameInput = wrapper.find('input.collection-name-input');
```
srallen: This is re-testing React/JSX and not needed.

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    assert.equal(collectionNameInput.length, 1);
+  });
+
+  it('should render an private checkbox', () => {
+    const privateCheckbox = wrapper.find('input[type="checkbox"]');
```
srallen: Same comment about re-testing React/JSX

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,49 @@
...
+    assert.equal(privateCheckbox.length, 1);
+  });
+
+  it('renders a disabled button when there is no text in the input', () => {
+    wrapper.setState({ collectionNameLength: 0 });
```
srallen: The initial state is set to 0, so calling `setState` here is not necessary.

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+  it('renders a submit button when there is text in the input', () => {
+    wrapper.setState({ collectionNameLength: 5 });
+    addCollectionButton = wrapper.find('button');
+    assert.equal(addCollectionButton.prop('disabled'), undefined);
+  });
+
+  it('calls onSubmit when the add collection button is clicked', () => {
+    const event = {};
+    wrapper.setState({ collectionNameLength: 5 });
```
srallen: The state was set in the previous test to have length, and there was no resetting of that at the end, so setting the state again here is not necessary.

app/collections/create-form.jsx
```diff
@@ -0,0 +1,108 @@
...
+            <Translate content="collections.createForm.private" />
+          </label>
+          <div className="submit-button-container">
+            {this.state.collectionNameLength < 1 ?
+              <button type="submit" disabled={true}>
+                <Translate content="collections.createForm.submit" />
+              </button> :
+              <button type="submit">
+                <Translate content="collections.createForm.submit" />
+              </button>}
```
srallen: This whole block can be more concise since the conditional only applies to the disabled state of the button:

app/collections/create-form.jsx
```diff
@@ -0,0 +1,108 @@
+import React from 'react';
+import Translate from 'react-translate-component';
+import apiClient from 'panoptes-client/lib/api-client';
+
+class CollectionsCreateForm extends React.Component {
+  constructor(props) {
+    super(props);
```
srallen: FYI, passing props into super isn't necessary if props aren't used in the constructor. Admittedly we've been inconsistent with this.

app/collections/create-form.jsx
```diff
@@ -0,0 +1,108 @@
...
+      collectionNameLength: this.name.value.length
+    });
+  }
+
+  render() {
+    return (
+      <form onSubmit={this.onSubmit} className="collections-create-form">
+        <div className="form-help error">
+          {this.state.error &&
+              `Error ${this.state.error.status}: ${this.state.error.message}`}
```
srallen: The conversion here is missing the more user friendly messages. We should still use those. Now that we have the status codes available to us, that could be useful in determining what message should be.

jelliotartz:
Thanks for your suggestions :) Any thoughts on renaming the `project` prop to `projectID` or `projectId` (see previous PR Update above)

srallen:
Yeah, that's fine as it improves clarity with the propType.

app/collections/create-form.jsx
```diff
@@ -0,0 +1,108 @@
...
+          </div>
+        </div>
+      </form>
+    );
+  }
+}
+
+CollectionsCreateForm.defaultProps = {
+  onSubmit: () => {},
+  projectID: '',
```
srallen: This default prop should probably be null since it's set as null in the parent if there's no project.

eatyourgreens: React might warn about that being the wrong type, because `null` is an Object and it looks like the prop should be a string.

srallen: Hm, that's probably why it got left out of the default prop definitions originally. Regardless, we should be consistent with types used. Either default it to null and define the allowed proptypes to allow string and null or the parent needs to be edited to set it as an empty string when there's no project.

eatyourgreens: My preference is for this to be consistently defined as a string everywhere that it appears. Allowing object or string opens up the possibility of uncaught errors if someone accidentally passes the project object when they mean to use the project ID.

jelliotartz:
## PR Update
- Thanks for this input! projectID prop now consistently defined as a string - either the project id as a string, or an empty string.
- The conditional to assign the collection to a project is working as expected.
- Staging branch updated.

app/collections/create-form.jsx
```diff
@@ -0,0 +1,112 @@
...
+    const collection = {
+      display_name: displayName,
+      private: isPrivate,
+      links
+    };
+
+    apiClient.type('collections').create(collection).save()
+      .then((newCollection) => {
+        this.name.value = '';
+        this.isPrivate.value = true;
```
srallen: This is non-blocking because it was written this way before, but I don't understand why this is being set true after a successful collection POST. This is something to consider for future refactoring when the description field is added.

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+  it('renders a submit button when there is text in the input', () => {
+    wrapper.setState({ collectionNameLength: 5 });
+    addCollectionButton = wrapper.find('button');
+    assert.equal(addCollectionButton.prop('disabled'), false);
+  });
+
+  it('calls onSubmit when the add collection button is clicked', () => {
+    const event = {};
+    wrapper.find('form').simulate('submit', event);
```
srallen: Passing the simulated submit an empty event object isn't necessary since the class method handling the submit event is stubbed.

app/collections/create-form.spec.js
```diff
@@ -0,0 +1,37 @@
...
+import { shallow } from 'enzyme';
+import CollectionsCreateForm from './create-form';
+
+describe('<CollectionsCreateForm />', () => {
+  let wrapper;
+  let onSubmitSpy;
+  let addCollectionButton;
+
+  before(() => {
+    onSubmitSpy = sinon.stub(CollectionsCreateForm.prototype, 'onSubmit');
```
srallen: Let's be consistent with variable naming here, `onSubmitStub`

app/collections/create-form.jsx
jelliotartz: Agreed, I was wondering the same thing. Happy to change to whatever makes the most sense - `false`, `null`, or `undefined`, just let me know what you'd prefer :)

app/collections/create-form.jsx
```diff
@@ -40,7 +40,7 @@ class CollectionsCreateForm extends React.Component {
     apiClient.type('collections').create(collection).save()
       .then((newCollection) => {
         this.name.value = '';
-        this.isPrivate.value = true;
+        this.isPrivate.value = null;
```
srallen: We shouldn't change this from the original for the conversion. I left a comment because I didn't understand it, but that doesn't mean it should change. For the upcoming bugfixes and features that will be implementing, some time should be spent to understand it and if it's cruft, then it should be removed.

jelliotartz:
Sounds good, value reverted :)

srallen:
This just needs a rebase to get the updated package.json, then should be ready.

jelliotartz:
Rebase done

PR #4135, opened 2017-11-02
add categories to organization landing

mcbouslog:
Towards #2970.
Related to https://github.com/zooniverse/pfe-lab/pull/137.

If categories added to an organization, shows category buttons on organization landing.

If no categories on an organization, organization landing shows no category buttons.

On initial load the "All" button should be active, then when a category is clicked the projects should be filtered by the category selected.

https://org-categories.pfe-preview.zooniverse.org/organizations/markb-panoptes/nfnorgtest

# Review Checklist

- [x] Does it work on mobile?
- [x] Did you deploy a staging branch? - https://org-categories.pfe-preview.zooniverse.org

## Optional

### DIFF
--- a/app/pages/organizations/organization-container.jsx
+++ b/app/pages/organizations/organization-container.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { browserHistory } from 'react-router';
 import apiClient from 'panoptes-client/lib/api-client';
 import Translate from 'react-translate-component';
 import isAdmin from '../../lib/is-admin';
@@ -23,6 +24,7 @@ class OrganizationContainer extends React.Component {
     };
 
     this.toggleCollaboratorView = this.toggleCollaboratorView.bind(this);
+    this.updateQuery = this.updateQuery.bind(this);
   }
 
   componentDidMount() {
@@ -42,6 +44,10 @@ class OrganizationContainer extends React.Component {
         this.fetchOrganization(nextProps.params.name, nextProps.params.owner);
       }
     }
+
+    if (this.state.organization && (nextProps.location.query !== this.props.location.query)) {
+      this.fetchProjects(this.state.organization, (isAdmin() || this.isCollaborator()), nextProps.location.query);
+    }
   }
 
   isCollaborator() {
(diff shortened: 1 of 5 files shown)


mcbouslog:
just noticed if org collab, click View as Volunteer, click a category or two, then go back to All category, it shows all project cards even though View as Volunteer is still checked (it should only show launch approved).

app/pages/organizations/organization-container.jsx
```diff
@@ -60,13 +66,31 @@ class OrganizationContainer extends React.Component {
...
-  fetchProjects(organization, collaboratorView) {
-    this.setState({ errorFetchingProjects: null, fetchingProjects: true });
+  updateQuery(newParams) {
+    const query = Object.assign({}, this.props.location.query, newParams);
+    const results = [];
+    Object.keys(query).map((key) => {
+      if (query[key] === '') {
+        results.push(delete query[key]);
+      }
+      return results;
```
srallen: Does the results array get used somewhere again? If not, maybe `forEach` could be used instead of `map`?

app/pages/organizations/organization-container.jsx
```diff
@@ -60,13 +66,31 @@ class OrganizationContainer extends React.Component {
...
+    });
+    const newLocation = Object.assign({}, this.props.location, { query });
+    newLocation.search = '';
+    browserHistory.push(newLocation);
```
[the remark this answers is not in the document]
mcbouslog: using context looks better to me, great suggestion!

mcbouslog:
thanks @srallen - great point about collection roles! I added a related card to the zooniverse organizations project to be sure to revisit

app/pages/organizations/organization-page.jsx
```diff
@@ -110,6 +124,23 @@ class OrganizationPage extends React.Component {
...
+            <div className="organization-page__categories">
+              <button
+                className={this.calculateClasses('All')}
+                onClick={this.handleCategoryChange.bind(this, '')}
+              >
+                All
+              </button>
+              {this.props.organization.categories.map(category =>
+                <button
+                  key={category}
```
srallen: Not gonna require this for this PR since it's more of a question, but is there a guarantee that the categories are unique? I'm not sure if the API validates uniqueness for tags. Could be a potential problem for use w/ keys here.

eatyourgreens: We just ran into this using link URL for keys in the project navigation links. In a real project the key will always be unique but there is an edge case in which a project owner can accidentally create a duplicate key, in which case PFE will not render the second item. 

I think it's a sufficiently rare case that you shouldn't spend a lot of time worrying about it.

mcbouslog: good point. could adding `index` and then defining the key as 
```
key={`${category}-${index}`}
```
help?

eatyourgreens: I'd strongly advise against using array index in keys, since [that breaks your component if the array order changes](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318).

mcbouslog: sounds good

mcbouslog:
yes great point regarding scroll position, will do, thanks!

PR #4141, opened 2017-11-08
Clear out user group query and state on sign out

srallen:
This is follow up to #4061 and in support of the Intro2Astro use of projects with classroom user groups. If the user signs out, the group state should be set to false and the query param cleared.

I am a bit unsure about clearing the query param in case the student first arrives to the project not signed in and it may be difficult to remind them to make sure to use the project URL with the correct query param so they get credit for their classification. For now, I'll try it out with the query param cleared too.

This can be tested using the zootester1 account and at this staged path:

https://clear-user-group-on-logout.pfe-preview.zooniverse.org/projects/srallen086/intro2astro-hubble-testing/classify?group=1326233

I can share the account credentials in person when you're ready to test.

# Review Checklist

- [ ] Does it work on mobile?

## Optional

### DIFF
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -49,7 +49,9 @@ module.exports = React.createClass
 
 
   contextTypes:
-    geordi: React.PropTypes.object
+    geordi: React.PropTypes.object,
+    initialLoadComplete: React.PropTypes.bool,
+    router: React.PropTypes.object
 
   propTypes:
     loadingSelectedWorkflow: React.PropTypes.bool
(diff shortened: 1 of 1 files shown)


app/pages/project/classify.cjsx
```diff
@@ -321,6 +326,13 @@ module.exports = React.createClass
       apiClient.type('user_groups').get(props.location.query.group).then (group) =>
         @setState({ validUserGroup: group and group.links.users.includes(props.user.id) })
```
mcbouslog: if there is a group query param in the url and a signed in user, but the user isn't included in the group, do we want to clear the group query param from the url? could conditionally call `clearUserGroupForClassification` within `validateUserGroup` if user not included in group.

app/pages/project/classify.cjsx
```diff
@@ -321,6 +326,13 @@ module.exports = React.createClass
...
 
+  clearUserGroupForClassification: (props, context) ->
+    if (props.location.query?.group)
+      @setState({ validUserGroup: false })
+      newLocation = Object.assign({}, props.location, { query: '' });
```
mcbouslog: not relevant for initial use cases, but I think this would clear `workflow=1234` query param if present in project with experimental allow workflow query feature, not sure if more surgical way to clear just `group`.

srallen:
Thanks for the questions and suggestions. I improved the conditions where the group query could be removed and improved the removal class method to only remove the group query param.

PR #4142, opened 2017-11-09
Rewrite profile editor without PromiseRenderer

eatyourgreens:
Rewrites the account profile settings page, replacing the PromiseRenderer component with component state.

# Review Checklist

- [ ] Does it work on mobile?
- [x] Did you deploy a staging branch? https://edit-profile.pfe-preview.zooniverse.org

## Optional


app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,142 @@
...
+
+  handleMediaChange(type, file) {
+    const { user } = this.props;
+    const errorProp = `${type}Error`;
+    this.setState({ [errorProp]: null });
+    apiClient.post(user._getURL(type), { media: { content_type: file.type }})
+    .then(([resource]) => {
+      putFile(resource.src, file, { 'Content-Type': file.type });
+    })
+    .then(() => {
```
eatyourgreens: Why was this block in the original component? What did it do? Can we just delete it?

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  handleMediaClear(type) {
+    const errorProp = `${type}Error`;
+    this.setState({ [errorProp]: null });
+    this[`${type}Get`]
+    .then((resource) => {
+      !!resource && resource.delete();
+    })
+    .then(() => {
```
eatyourgreens: Ditto for this block. Why was the original code rewriting component methods on-the-fly?

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  avatarGet() {
+    this.props.user.get('avatar')
+    .then(([avatar]) => {
+      this.setState({ avatar });
+    });
+  }
+
+  headerGet() {
```
eatyourgreens: should be `profile_headerGet`

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,142 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {
+      avatar: {},
+      avatarError: null,
+      header: {},
+      headerError: null
+    };
+    this.avatarGet();
+    this.headerGet();
```
eatyourgreens: `this.profile_headerGet`

app/pages/settings/profile.jsx
simoneduca: I have no idea. Do you mean this also doesn't do anything?

app/pages/settings/profile.jsx
srallen: I originally wrote these files and I'm pretty sure I copied what was being done in the lab for the image media uploads. Maybe this commit is a clue? https://github.com/zooniverse/Panoptes-Front-End/commit/627021199705ff433439812cd69e0a7f5f4667bb

app/pages/settings/profile.jsx
eatyourgreens: It does do something. It deletes the event handler then forces a re-render, effectively making the button useless after one successful click. But why?

eatyourgreens: The original code generated a new event handler on each render.

app/pages/settings/profile.jsx
eatyourgreens: Or maybe not. I'm confused by my own code.

eatyourgreens:
Rebased, re-staged and, most importantly, re-written. I'm a lot happier with this version.

eatyourgreens:
I'm seeing the avatar being retrieved for both the avatar and profile header. Not sure why that's happening.

eatyourgreens:
Waiting on zooniverse/Panoptes#2528 to fix profile images.

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,133 @@
...
+      return resource;
+    });
+  }
+
+  handleMediaChange(type, file) {
+    const { user } = this.props;
+    const errorProp = `${type}Error`;
+    this.setState({ [errorProp]: null });
+    apiClient.post(user._getURL(type), { media: { content_type: file.type }})
+    .then(([resource]) => {
```
eatyourgreens: `resource.src` here is a signed S3 URL, for uploading the file, so we can't store this resource in component state.

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,133 @@
...
+      putFile(resource.src, file, { 'Content-Type': file.type })
+      .then(() => this.getUserResource(type))
```
eatyourgreens: Fetch the resource from Panoptes, again, in order to get the public S3 URL.

app/pages/settings/profile.jsx
```diff
@@ -0,0 +1,133 @@
...
+import putFile from '../../lib/put-file';
+
+const MAX_AVATAR_SIZE = 65536;
+const MAX_HEADER_SIZE = 256000;
+
+class CustomiseProfile extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      avatar: {},
```
eatyourgreens: Uses empty objects for defaults, to avoid `undefined property of null` errors later.

eatyourgreens:
@srallen thanks for your comments. I think that code was there to force an update when the media file changes. Should all be working in the new code now, by fetching the new resource from Panoptes and storing it in component state.

eatyourgreens:
The Panoptes bug is fixed, so I think this is ready for review now @srallen @simoneduca

eatyourgreens:
Added some tests for the initial load of the user images and the two delete buttons.

eatyourgreens:
Rebased to latest master.

app/pages/settings/profile.jsx
simoneduca: Is that a better approach in general?

app/pages/settings/profile.spec.js
```diff
@@ -0,0 +1,51 @@
...
+  });
+
+  it('renders the user background', () => {
+    const avatar = wrapper.find('ImageSelector[src="//zooniverse.org/images/header.jpg"]');
+    assert.equal(avatar.length, 1);
+  });
+
+  it('deletes the avatar when Clear Avatar is pressed', () => {
+    wrapper.find('button').first().simulate('click');
+    setTimeout(() => assert.equal(wrapper.update().state().avatar, {}));
```
simoneduca: @eatyourgreens Is this intended to simulate the avatar resource being an empty object when deleted? If so, shouldn't the test fail if I add`src: 'avatar'` inside `{}`?

eatyourgreens: Yep.

app/pages/settings/profile.jsx
eatyourgreens: Depends on your code.

app/pages/settings/profile.spec.js
simoneduca: It doesn't.

eatyourgreens: Interesting. I'll take another look at this when I'm free. Did you find any clues as to why it isn't working?

simoneduca: I've never seen `setTimeout` used in test so I was checking if it plays well with `wrapper.update()`.

Issue #4139, opened 2017-11-07
Creating a collection in a project context gets the wrong project id


srallen:
## Expected behavior
The collection create form within the project context should use the project id of the project you're in the context of. The collection create form should probably include in its UI a message that the collection will be linked to the project you're viewing.

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Viewing a collection in a project context means the collection is linked to that project. Usually you'd expect all the subjects in that collection to be from that project, but we do not validate this. Any subject from any project can be linked to a collection. Consider this collection that is linked and being viewed within the project context:
https://master.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats/collections/wgranger-test/test-collect

It contains subjects from other projects. If you click the collection manager button, the collection create form is using the project id presumably from the default subject.

srallen:
Related to this issue is that project ids, likely from the default subject, are being used outside of the project context too for the form: https://master.pfe-preview.zooniverse.org/collections/zootester1/hello-cats

If you're creating a collection outside of a project context, it should not be linked to any project.

srallen:
@jelliotartz could you take a look at this while working on implementing the description field for the form?

jelliotartz:
Sure, no problem! :)

jelliotartz:
I've been working on isolating specific issues that are problematic regarding collections and projects, and have a few questions:

Scenario 1:
A user is on a collection page that is not in a project context. They click the CollectionsManagerIcon on a subject, to create a new collection. Should this new collection get the project context from that subject's project?

if yes, follow up: what if user selects multiple subjects, from different projects?

Scenario 2:
A user is in a collection that is in a project context (let's call it Collection A, in Project B). User selects 2 subjects from Collection A, but they are NOT subjects from Project B, and wants to create a new collection. Should the new collection be linked to Project B?

follow up: what if both subjects are from the same (separate) project? Link to that project? what if they are not?

eatyourgreens:
> Should this new collection get the project context from that subject's project?
Subjects can belong to more than one project too, just to make this more complicated. In practice, I don't think anyone has yet used this feature of Panoptes. A practical, but still theoretical, use case might be two projects sharing a subject set within the same organisation.

camallen:
[...]

PR #4146, opened 2017-11-09
Collection description

### LINKED ISSUE #3469: Interface for adding/editing a collection description

jelliotartz:
## Describe your changes
  - Adds a description `textarea` field for a user on the CollectionsCreateForm.
  - Edit functionality: user can edit a description in the CollectionSettings page.
  - Display: 
    - Collection card name and description redesigned per discussion with @beckyrother.
    - Rollover transition with hover state for collection description.
  - Collection description is limited to 300 characters. `Charlimit` component added to notify user of character count as they add/edit their description.
    - If the input field on the create form goes over 300 characters, the submit button is disabled.
    - If the input field on the edit form goes over 300 characters, an error message is displayed beneath the field, and more text cannot be entered.
  - Fixes #3469.
    - Related: 
      - https://github.com/zooniverse/Panoptes/issues/2170.
      - ES6 conversion of CollectionsCreateForm https://github.com/zooniverse/Panoptes-Front-End/pull/4133
  - [Staged here](https://collection-description.pfe-preview.zooniverse.org).

## RFCs
  - `handleDescriptionInputChange` works in its current form, but it seems likely that it could use some optimization/refactoring.
    - conditional display of error message: is there a simpler or more elegant way to accomplish this? Also, is this the UI/UX that we want?
    - AutoSave and `update` API call - is there a cleaner way to handle updating? We use AutoSave on the `DisplayNameSlugEditor` field above description, so it seems logical that we should use it here as well, right? I played around with removing/changing the `.update` API call, but have not yet figured out whether it can be simplified. 
  - If a description is more than ~125 characters, the remaining text is not displayed (ie, it can't fit on the card, as it's currently designed). This current UI seems relatively ok to me, just thought I'd point it out in case others want to weigh in on how to handle the display of long descriptions.

## TODOs
  - [x] Write tests for changes/additions to CollectionsCreateForm and CollectionSettings.
  - [x] Creating a collection in a project context gets the wrong project id [(Issue#4139)](https://github.com/zooniverse/Panoptes-Front-End/issues/4139).

## Review Checklist

- [x] Does it work on mobile?

### Optional


app/collections/create-form.jsx
```diff
@@ -77,6 +89,18 @@ class CollectionsCreateForm extends React.Component {
             placeholder="Collection Name"
           />
         </label>
+        <label>
+          <textarea
+            className="collection-name-input"
```
jelliotartz: RFC: I'm currently using the `.collection-name-input` class for both the name input and the description input styles - which seems not right. Would it be better to change the name to something like `.collection-inputs`? or make a separate css class for `.collection-description-input`? 

Right now, styling the name and description inputs with the same selector seems to be working just fine, but I'm not sure what's best practice for this.

app/collections/create-form.jsx
```diff
@@ -54,6 +60,12 @@ class CollectionsCreateForm extends React.Component {
     });
   }
 
+  handleDescriptionInputChange() {
```
srallen: Let's use the event object here:
```
handleDescriptionInputChange(event) {
  this.setState({ description: event.target.value })
}
```

app/collections/create-form.jsx
```diff
@@ -33,13 +37,15 @@ class CollectionsCreateForm extends React.Component {
...
+      description,
       private: isPrivate,
       links
     };
 
     apiClient.type('collections').create(collection).save()
       .then((newCollection) => {
         this.name.value = '';
+        this.description.value = '';
         this.isPrivate.value = true;
```
srallen: As I mentioned in the conversion PR, I think this is in error. This is invalid HTML. If we're resetting the form state after submission, then this should be resetting what the checked attribute is set to.

app/collections/create-form.jsx
```diff
@@ -1,24 +1,28 @@
...
 import apiClient from 'panoptes-client/lib/api-client';
+import CharLimit from '../components/char-limit';
 
 class CollectionsCreateForm extends React.Component {
   constructor() {
     super();
 
     this.state = {
       collectionNameLength: 0,
+      descriptionText: '',
```
srallen: Let's just call this `description`

app/collections/create-form.jsx
```diff
@@ -77,6 +89,18 @@ class CollectionsCreateForm extends React.Component {
...
+          <textarea
+            className="collection-name-input"
+            onChange={this.handleDescriptionInputChange}
+            ref={(node) => { this.description = node; }}
+            placeholder="Collection Description (less than 300 characters)"
+          />
+        </label>
+        <CharLimit
+          limit={300}
+          string={this.state.descriptionText ? this.state.descriptionText : ''}
```
srallen: The ternary is unnecessary. The default state is already an empty string.

app/collections/create-form.jsx
```diff
@@ -77,6 +89,18 @@ class CollectionsCreateForm extends React.Component {
             placeholder="Collection Name"
           />
         </label>
+        <label>
```
srallen: This doesn't actually have a text label, so I'm not sure of the utility of wrapping the input with one. Same with the name input. We should add labels or remove the label tags.

app/collections/create-form.jsx
srallen: This is something that the BEM can help us with. Let's do a few things:

- Rename this file to be consistent with the name of the component
- The top level class name `collections-create-form` is the block
- Inputs are elements in the block, so following BEM convention: `collections-create-form__input`
- Any variants should use the modifier convention: `collections-create-form__input--description`

app/collections/create-form.spec.js
```diff
@@ -21,24 +23,65 @@ describe('<CollectionsCreateForm />', () => {
     assert.equal(wrapper, wrapper);
   });
 
-  it('renders a disabled button when there is no text in the input', () => {
-    addCollectionButton = wrapper.find('button');
-    assert.equal(addCollectionButton.prop('disabled'), true);
+  describe('should render a disabled submit button', () => {
+    it('when there is no text in the name input', () => {
+      wrapper.setState({
```
srallen: You don't need to set state here because the initial state is 0.

app/collections/create-form.spec.js
```diff
@@ -21,24 +23,65 @@ describe('<CollectionsCreateForm />', () => {
...
+      sinon.assert.calledOnce(handleDescriptionInputChangeStub);
+    });
   });
 
-  it('calls #handleNameInputChange when a user changes text in the name input', () => {
-    wrapper.find('input.collection-name-input').simulate('change');
-    sinon.assert.calledOnce(handleNameInputChangeStub);
+  describe('should handle errors', () => {
+    it('when there is an error', () => {
+      wrapper.setState({ error: ['it broke!'] });
```
srallen: If we're mocking an error, then it should be an error object, not an array.

app/collections/settings.jsx
```diff
@@ -84,6 +89,37 @@ export default class CollectionSettings extends React.Component {
...
+        descriptionLengthError: true
+      });
+      return;
+    } else {
+      this.setState({
+        descriptionLengthError: false
+      });
+    }
+
+    this.props.collection.update(changes).save()
```
srallen: The AutoSave component is bad UX most of the places we've used it. There's a refactored DisplayNameSlugEditor that doesn't use the AutoSave component in ZRC. But we're not quite ready to add ZRC here yet. In the meantime, then to be consistent with behavior, using the AutoSave component for the description is ok. 

Now, since we're using the AutoSave component, it handles the update and save of the resource for you, so these lines are duplicating the request.

app/collections/settings.jsx
```diff
@@ -84,6 +89,37 @@ export default class CollectionSettings extends React.Component {
       });
   }
 
+  handleDescriptionInputChange() {
+    const property = 'description';
```
srallen: Add a name attribute to the textarea and then you can use the event object: `event.target.name`

app/collections/settings.jsx
```diff
@@ -84,6 +89,37 @@ export default class CollectionSettings extends React.Component {
...
+    const setting = this.state.setting;
+    setting[property] = true;
+    this.setState({ error: null, setting });
+
+    const descriptionText = this.description.value;
```
srallen: Use the event object: `event.target.value`

app/collections/settings.jsx
```diff
@@ -84,6 +89,37 @@ export default class CollectionSettings extends React.Component {
       });
   }
 
+  handleDescriptionInputChange() {
```
srallen: Since the AutoSave component is handling the resource update and save, then a lot of what this method is doing isn't necessary. Tracking the description length and communicating that error is helpful because all the AutoSave component does on error is turn the background red around the input and doesn't actually say what's wrong.

app/collections/settings.jsx
```diff
@@ -84,6 +89,37 @@ export default class CollectionSettings extends React.Component {
...
+    const property = 'description';
+    const setting = this.state.setting;
+    setting[property] = true;
```
srallen: Is this a copy and paste from the `handleToggle` event handler? This handler is dealing with text, not boolean values of a checkbox or radio button input type, so setting this to true doesn't make sense!

app/collections/settings.spec.js
```diff
@@ -19,48 +19,61 @@ const collectionWithoutDefaultSubject = {
   display_name: 'A collection',
   private: false,
   slug: 'username/a-collection'
-}
+};
 
-describe('<CollectionSettings />', function() {
+describe('<CollectionSettings />', () => {
```
srallen: I haven't been a stickler about this for PFE tests, but the mocha docs specifically say that arrow functions are discouraged: https://mochajs.org/#arrow-functions

Changing this isn't necessary.

css/collection-card.styl
```diff
@@ -1,26 +1,53 @@
 .collections-card-list
   @extends $owned-card-list
-  
+
 .collection-card
   @extends $owned-card
   background: url('/assets/simple-pattern.png') center center repeat
   border-radius: 1em
```
srallen: Should the borders still be rounded? 

screen shot 2017-11-10 at 4 42 08 pm

css/collection-card.styl
```diff
@@ -1,26 +1,53 @@
...
+  .details
+    padding: 1em
+    text-align: left
+    height: 80px
+    transition: height 0.5s ease
+    position: absolute
+    bottom: 0
+    left: 0
+    right: 0
+    justify-content: space-between
```
srallen: This appears to largely be a copy/paste of the CSS from project cards. If they're going to be shared, then let's get the shared bits into the owned card styles in common.styl and the divergent styles would go in either collection-card.styl and project-card.styl respectively. That being said, there's some CSS in the project cards that doesn't make sense anymore. All of the flex stuff on the details container for instance. We should be removing cruft styles as the styles change.

app/collections/settings.spec.js
jelliotartz: Happy to not use arrow functions, as this is what Mocha prefers :) I was using arrow functions because function expressions are generating the linter errors and warning below for every test: 
```
    warning  Unexpected unnamed function                               func-names
    error    Unexpected function expression                            prefer-arrow-callback
    error    Missing space before function parentheses                 space-before-function-paren
```
should we modify our eslint config to not throw these? or just ignore them?

srallen: At some point we can setup the linter to have different rules for mocha, but that's out of scope here. Just ignore them or you can put exceptions at the top of the file if it's bothersome.

jelliotartz:
## PR Update
I addressed a lot of the requested changes, but still am working on a few things:
- `handleDescriptionInputChange`: 
  - I've tried removing the update and save methods, but so far am having issues - ie, it breaks when I remove that API call. Is there an issue with how I'm connecting my `textarea` to the event handler? or maybe is something not working as expected with AutoSave?
  - I can remove `.save()`, and the `.then()` and `.catch()` blocks, but removing `this.props.collection.update(changes)` currently breaks the update functionality.
- still have work to do on the css refactoring.
- haven't updated sad, failing tests.
- staging branch updated.

app/collections/settings.jsx
```diff
@@ -89,35 +88,18 @@ export default class CollectionSettings extends React.Component {
...
   }
 
-  handleDescriptionInputChange() {
-    const property = 'description';
-    const setting = this.state.setting;
-    setting[property] = true;
-    this.setState({ error: null, setting });
-
-    const descriptionText = this.description.value;
+  handleDescriptionInputChange(event) {
```
srallen: This whole method could be a bit more concise and readable. 
This can be reduced to:
```
handleDescriptionInputChange(event) {
  const description = event.target.value;
  this.setState({ description })

  if (description.length <= 300) {
     this.props.collection.update({ description });
  }
}
```

Using the name attribute from an input is useful, but not necessary here since we're only concerned with updating the description.

css/collection-card.styl
```diff
@@ -36,13 +29,16 @@
...
     .description
-      text-overflow: ellipsis
       color: COPY
       font-size: 0.9em
       height: 150px
       margin: 0.6em 0
       overflow: hidden
+      text-overflow: ellipsis
       text-align: left
+      display: -webkit-box;
```
jelliotartz: I added the styles below to get the ellipsis to display, following the example [here](https://css-tricks.com/line-clampin/#article-header-id-0). Not sure if there's a better way, but just having `overflow: hidden` and `text-overflow: ellipsis` was not rendering the ellipsis for text that was longer than the area allotted for the description.

Also, as I was working on this, it occurred to me to ask: we're currently limiting description text to 300 characters - is there anywhere besides on the collection card where the description is displayed? Approximately 125 characters fit on the card before the text overflows. Would it make sense to limit the text to, say, 125? Or if I'm just missing where the full 300 character description is displayed, never mind.

srallen: Well, no, not yet, since the description is new. The description probably should also be displayed on the collection view page too. Maybe somewhere below the name but above the subjects? You could check in with @beckyrother

jelliotartz:
## PR Update
When a user creates a new collection, the CollectionsCreateForm now displays a message telling them what project the collection will be linked to, or that the collection will not be linked to a project.

This was accomplished by passing in `@props.project` to each subject node (named projectContext here, partly to differentiate it from `@state.project` in show-list, and partly because I think it accurately defines what it is) - when show-list is within a project context, it passes in a project resource. When it's not, it passes in null.

Potential issue: 
- SubjectViewer no longer checks for `@props.project` when rendering the CollectionsManagerIcon and FavoritesButton (it can't because sometimes SubjectNode passes in null as the project prop - ie, when user is not in a project context). I've been checking and trying to find any instances where this breaks something, so far I have not found any breaking changes.

[Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

app/components/subject-viewer.cjsx
```diff
@@ -179,7 +179,7 @@ module.exports = React.createClass
                 <i className="fa fa-info-circle fa-fw"></i>
               </button>{' '}
             </span>}
-          {if @props.project? and @props.subject?
+          {if @props.subject?
```
srallen: This needs to be tested a bit more. The check for the project is necessary for the FavoritesButton. The FavoritesButton will only work if there's a project; the favorites collection is for a project. It won't work outside of project context. 

Before, the project set to the parent state was the one being passed down. It was determined in either two ways: the project linked to the project or the project the subject belonged to. I think we'll have to be careful for the FavoritesButton that it gets the correct project, that is the project  that the favorites collection belongs to for that user. I think perhaps the code as it was was written to get the FavoritesButton to work correctly, but was causing buggy behavior for the Collections Manager/Creating other collections. Perhaps we should separate out the props and naming for project context vs the project the favorites collection belongs to and make sure the FavoritesButton is only visible when in a project context.

app/pages/collections/collection-card.jsx
```diff
@@ -45,15 +45,18 @@ export default class CollectionCard extends React.Component {
...
             <div className="owner">
               {this.props.shared ?
                 <span><i className="fa fa-users" />{' '}</span> : null}
               {this.props.collection.links.owner.display_name}
             </div>
+            <div className="name">
+              <span>{this.props.collection.display_name}</span>
+              {this.props.collection.private ? <i className="fa fa-lock" /> : null}
+            </div>
+            <div className="description">
```
srallen: How about just a `<p>` here?

jelliotartz:
## PR Update
Addressed some/most of the requested changes in this commit. Some details about things that are still WIP:
  - Cleaning up the css styles:
    - Tried my best to organize `$owned-card` with the common styles, and to leave alterations on project and colleciton card. Apologies in advance for anything I missed.
    - > For example, using flex doesn't make sense anymore if something is also absolutely positioned now.
    - removing `display: flex` and `flex-direction: column` from `$owned-card` sets part of the project card description to be visible at the bottom of the card when it's not rolled over (\*\), and also interferes with the display of the description text (ellipsis no longer working as expected) (*). Not sure why this is the case, and will continue looking for ways to clean that up.
      - *Including screen shots for reference
  - For the Favorites button issue, I've been trying to implement two different properties to handle the different requirements for FavoritesButton and CollectionManagerIcon - the project prop for favorites, and some sort of boolean that changes based on whether the user is in a project context or not. So far I haven't been able to get it to work though. Will keep at it though.

[Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

image

image

srallen:
Just responding to the CSS questions:
Removing the flex box CSS properties aren't causing that issue with the description text overflow. I'm seeing it present regardless. It's the `text-overflow` property that isn't working like we would expect. Reading up a bit more on `text-overflow`, apparently it doesn't work on multi-line strings: https://software.intel.com/en-us/html5/hub/blogs/ellipse-my-text/ So it wasn't a great suggestion on my part.

Second thing: I understand that `-webkit-line-clamp` was probably added to get the text trimming to work, but I'm not sure why `-webkit-box` or `-webkit-box-orient` is present. The CSS box properties are deprecated early version of flex box. Also, generally speaking you shouldn't need to add vendor specific CSS properties (unless it's something only that vendor supports and it's to fix a vendor specific CSS issue). We have nib setup to handle this for us. 

The project cards are just letting descriptions get cut off, so that's probably fine here for now too. We'll have to adding in some string trimming via javascript at some point.

One last thing, the description should be added to the collection page itself too. I'm not sure if you saw that comment because that somehow ended up on an outdated diff too.

srallen:
@mcbouslog if it's ok, could you help do some reviewing on this too since I'll be out this week?

jelliotartz:
## PR Update

Per convo with @beckyrother, we designed the collection description display on the collection view page with the following specifications:
  - 50% width
  - lined up with the select subject button
  - font: Open Sans, 14px

@beckyrother et al, please feel free add more to the styles as needed.

Above styles added - apologies if additions/changes to CSS don't follow our desired conventions. Happy to refactor, but would appreciate direction if there's a simpler/better way to handle the design requirements. 

Working on trying to find good/BEM-compliant naming for the collection description and select subject buttons and their container, will pick up here tomorrow. 

Also still WIP on fav button issue.

[staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

beckyrother:
Thanks @jelliotartz ! If you'd like to use an existing style rather than creating a new one, it could just be `p` from the project About pages: 

treeversity_ _about_ _zooniverse

(not sure if that's helpful or not)

jelliotartz:
## PR Update

CSS wrangling:
- Doing my best to make the CSS naming BEM compliant, eliminate cruft, and keep the selectors logical and easy to understand. Apologies if it's off the mark.
- Added `color: COPY` from the About page styles, thanks @beckyrother :)

Project ID bug and favorites project context:
- Added a check to make sure the FavoritesButton is only visible when in a project context.
- Testing on my local with staging projects, the following appears to be working:
  - in project context:
    - user can add subject to a project-affiliated collection
    - user can add subject to favorites (which are project-affiliated)
  - not in project context
    - user can add subject to a not-project-affiliated collection
    - favorites button is not visible
  - in both of the above contexts, a message displays on the collection create form, telling the user whether the collection will be created in a project context or not (if yes, it displays the project name).

Looking back through the outdated comments, not sure if I'm missing any requested changes at this point - please let me know if anything jumps out as still needing attention.

[Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

css/common.styl
```diff
@@ -166,6 +166,8 @@ $owned-card
   overflow: hidden
   text-align: left
   text-decoration: none
+  height: 270px
+  position: relative
 
   svg
```
srallen: The project card has its own height for the svg element defined in its style file, so this is only applying to the collection card. Let's move this into the collection card style file.

css/common.styl
```diff
@@ -166,6 +166,8 @@ $owned-card
   overflow: hidden
   text-align: left
   text-decoration: none
+  height: 270px
```
srallen: Since a fixed height is defined, then the max-height css property doesn't make sense anymore. It can be removed.

css/common.styl
```diff
@@ -177,12 +179,21 @@ $owned-card
     display: flex
     flex-direction: column
     padding: 20px
```
srallen: The collections card appears to have its own padding defined, so this `padding: 20px` is only applicable to the project card and should be moved into its style file.

jelliotartz:
## PR Update
Changes addressed:
  - Adding a collection not affiliated with a project no longer displays a message informing the user about the unaffiliated status
  - Description placeholder message edited for clarity
  - Collection description moved to `CollectionPage` component and rendered with `<p>` element
  - `.collections-show` returned to previous scope over entire CollectionShowList component
  - Remove unnecessary `<span>` element from description on collection card
  - CSS wrangling: 
    - Move styles shared by project and collection cards to `common`
    - Move not-shared card styles to the appropriate card style sheets.
    - Clean up and remove old styles and code
      - RFC: @eatyourgreens I removed `.card-button` from the HTML and CSS, as we are currently using a `display: none` css property to hide it on project cards. Were we keeping the button element around for accessibility? I'm happy to put it back in if we want it for accessibility reasons, just wasn't sure. And if yes we do want it, I'll add something similar for the collection card.
      - RFC: @beckyrother changes I made to the location of collection description on the collection show page have interfered with lining up the description and the buttons - I'm conginuing to tweak the CSS to get the alignment back, but wanted to check and see how strongly you feel about that - ie, is having the description above the buttons a deal breaker? (see image below for clarification)

image

[Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

beckyrother:
@jelliotartz it would be great if it were possible to line up those two elements, since right now there's quite a bit of awkward white space up there. Thank you!

jelliotartz:
## PR Update
I've got the alignment issue fixed, and I've kept description in the `CollectionPage` component/outside of the `if @state.subjects?` check in `CollectionShowList` - however, the CSS used to accomplish this feels hacky. 

I'm guessing there's a more elegant way to get description and the select subject buttons in alignment, but I'm finding it tough because one component (CollectionShowList) is the child of the other (CollectionPage). Open to suggestions for other approaches.

### [Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

srallen:
@jelliotartz The description displaying really shouldn't be conditional based on the presence of subjects, but odds are that this will be a rare circumstance. Let's revert back to the styles and layout previous rather than try to guess on the positioning using the top CSS property. These components need some serious layout restructuring and that's much bigger than the scope of this PR.

jelliotartz:
### [Staging branch updated](https://collection-description.pfe-preview.zooniverse.org/).

css/collections-page.styl
```diff
@@ -70,40 +70,56 @@ $collection-viewer-button
     @media (max-width: 900px)
       margin: 0.6em 0
 
-  .collections-show
-    .collection-buttons-container
-      display: flex
-      justify-content: flex-end
-      margin-bottom: 2em
+  .collection__description-container
```
srallen: Can we add a bottom margin to this, maybe 1em? Depending on the subjects and description length, the description can run right up against the subject:

screen shot 2017-11-29 at 10 11 10 am

jelliotartz:
`margin-bottom` added, and rebased in case it was needed.

beckyrother:
@srallen would you like me or James to write that note for Talk?

srallen:
@beckyrother I asked @jelliotartz to draft something. @jelliotartz could you share with @beckyrother ? Then if you like, you can update the collection feature thread you started before after this deploys? https://www.zooniverse.org/projects/zooniverse/gravity-spy/talk/729/295767?comment=489915&page=1

I'm wondering if we should also announce on Zoo wide talk too?

jelliotartz:
@srallen @beckyrother wanted to follow up on posting about this, now that it's merged - I'll paste the text I drafted below, let me know if you'd like me to make the zoo-wide post :) 


## New Feature: Collection Description

A new feature has been added to Collections - there is now a Description text field (up to 300 characters) that can be used to provide more information about a collection. You can find the description field on the Collection Settings page and it will display on both the collection card and page.

srallen:
Just a minor edit suggestion: I don't think collection, description, or settings should be capitalized. Also the field is available when you create a new collection too. 

@jelliotartz @beckyrother Yep, it's deployed. It definitely needs to be announced on Gravity Spy. I think @beckyrother to update the GS Talk since you posted the last collection feature announcement? Maybe @jelliotartz you can post to Zoo Talk, maybe just the Zooniverse general board?

jelliotartz:
Thanks for the editing suggestions, text updated :) I'll put up the post on Zooniverse general board (that's [this board](https://www.zooniverse.org/talk/14), yeah?

PR #4168, opened 2017-11-20
Make promotion messaging work for SW. Fix minor CSS bug

### LINKED ISSUE #4167: SW promotion message
Congratulations! You've unlocked the next level and can now access even more challenging questions about what you see in the trail camera images.  If you prefer to stay at this level, you can choose to stay.  

Button options: 
--Try the new level
--No thanks

srallen:
Staging branch URL: https://sw-promotion-message.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy?env=production

Fixes #4167. We can't quite test this for Snapshot Wisconsin Challenge yet, so the staging link is to test that this doesn't break Gravity Spy. This can be tested by going to a previous level. The promotion dialog will pop up when the subject queue is empty. I did enough classifications to get the zootester1 account promoted to level 2 on gravity spy in case you don't have an account that's been promoted yet. If you go to level 1, do enough classifications to clear the subject queue, you should see Gravity Spy's promotion dialog with their specific text. 

The other change was to make it so that the custom sign in prompt will only show for Gravity Spy. Since SW:C will be done with focus groups, they'll make sure the users are signed in. 

I also fixed some minor linter errors and a CSS bug with the mini-course.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/locales/en.js
+++ b/app/locales/en.js
@@ -9,7 +9,12 @@ export default {
     recents: 'Your recent classifications',
     talk: 'Talk',
     tutorialButton: 'Show the project tutorial',
-    miniCourseButton: 'Restart the project mini-course'
+    miniCourseButton: 'Restart the project mini-course',
+    workflowAssignmentDialog: {
+      promotionMessage: "Congratulations! You've unlocked the next workflow. If you prefer to stay on this workflow, you can choose to stay.",
+      acceptButton: 'Try the new workflow',
+      declineButton: 'No, thanks'
+    }
   },
   project: {
     loading: 'Loading project',
@@ -183,7 +188,7 @@ export default {
     },
     project: {
       section: 'Project email preferences',
-      header: 'Project'
+      header: 'Project',
     }
   },
   about: {
--- a/app/pages/project/classify.cjsx
+++ b/app/pages/project/classify.cjsx
@@ -314,7 +314,7 @@ module.exports = React.createClass
     # The API needs to add the promotion message as a field in project contents...
     gravitySpy = '1104'
     snapshotWiChallenge = '5371'
-    defaultMessage = "Congratulations! You've unlocked the next workflow. If you prefer to stay on this workflow, you can choose to stay"
+    defaultMessage = props.translation.promotion_message;
     gravitySpyMessage = "Congratulations! Because you're doing so well, you can level up and access more types of glitches, have more options for classifying them, and see glitches that our computer algorithms are even less confident in. If you prefer to stay at this level, you can choose to stay."
     snapshotWiMessage = "Congratulations! You've unlocked the next level and can now access even more challenging questions about what you see in the trail camera images. If you prefer to stay at this level, you can choose to stay."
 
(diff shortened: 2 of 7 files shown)


app/pages/project/classify.cjsx
```diff
@@ -305,9 +305,25 @@ module.exports = React.createClass
 
     @loadAppropriateClassification(@props) if @props.workflow?
 
+  getPromotionMessage: (props) ->
```
eatyourgreens: Could this be set in [the translations wrapper for projects](/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/project-translations.jsx), then referenced as `props.translation.promotion_message`:

srallen: I've moved the default promotion message into the project translation wrapper. I'm hesitant to put the custom messages for Gravity Spy and Snapshot Wisconsin in there because these being hard coded is temporary until it gets moved into the project contents in Panoptes. Once those are there, we can check for the presence of those project content strings or fallback to the default messaging. Does this work?

eatyourgreens: The translation wrapper should take care of that fallback for you, shouldn't it?

I don't think it makes a big difference here, but I'd prefer if components expect their language strings to be provided externally, via the `translation` prop, rather than defining them themselves.

srallen: Ok, I can do another quick refactor to have the handling of this in the translation wrapper.

srallen:
@eatyourgreens I refactored to actually determine which message to use in the `WorkflowAssignmentDialog` and the default message is in the en.js file. I also added the button strings to the en.js file. Once the promotion message is added to the project contents on the API, then it can be passed down properly through the project translations wrapper. 

I dismissed the previous approval since this has changed enough to warrant another full review. This can be tested a couple of ways:

For gravity spy, use an account that has been promoted at least to level two: https://sw-promotion-message.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy?env=production

The zootester1 account has been so you can use that if your own account hasn't been promoted yet on the GS project. If you do classifications on level 1, you should see the dialog with the correct message once the subject queue is clear.

On staging, you can use my test project to see the default message: https://sw-promotion-message.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing

You'll have to promote your own account to test this. As an admin, update your preferences's settings property, `settings.workflow_id`,  set to `'2334'` and classify on the beginner level. Once the queue is empty, you'll see the dialog.

app/locales/en.js
```diff
@@ -9,7 +9,12 @@ export default {
     recents: 'Your recent classifications',
     talk: 'Talk',
     tutorialButton: 'Show the project tutorial',
-    miniCourseButton: 'Restart the project mini-course'
+    miniCourseButton: 'Restart the project mini-course',
```
eatyourgreens: These should be added to the other language files, so that people know they're available to be translated.

eatyourgreens:
Looks good to me. The new language strings should be added to the other language files at some point, but I don't think that should hold up merging this.

PR #4181, opened 2017-11-27
Project page controller tests

### LINKED ISSUE #4180: Allow the tester role to be able to access inactive workflows
I was told to repost this as a front-end issue since the API is capable of doing this. I wanted to test out with a small group of users our new workflow, but apparently the tester role can't see a inactive workflow even with the direct link.

Can it be changed to allow testers to access inactive workflows via the direct url. It would be great for beta testing of new workflows or future upgrades to a project.

Note this is for P4: Ridges. I thought the tester role would have access to an inactive workflow just like the tester role can see a private project.

eatyourgreens:
Staging URL: https://project-page-controller.pfe-preview.zooniverse.org

Adds tests for some workflow selection cases in the project page controller.
Adds 'tester' to the roles that are allowed to load inactive workflows.
Fixes #4180
 
# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/index.spec.js
+++ b/app/pages/project/index.spec.js
@@ -57,7 +57,7 @@ const project = apiClient.type('projects').create({
     return Promise.resolve(resources[type]);
   },
   links: {
-    active_workflows: [1, 2, 3, 4, 5],
+    active_workflows: ['1', '2', '3', '4', '5'],
     avatar: { id: '1' },
     background: { id: '1' },
     owner: { id: '1' }
--- a/app/pages/project/index.spec.js
+++ b/app/pages/project/index.spec.js
@@ -170,4 +170,21 @@ describe('ProjectPageController', () => {
       sinon.assert.calledWith(workflowSpy, '2', true);
     });
   });
+
+  describe('without a saved workflow', () => {
+    beforeEach(() => {
+      location.query = {};
+      project.update({ 'configuration.default_workflow': '1' });
+      preferences.update({ settings: {}, preferences: {}});
+      const user = apiClient.type('users').create({ id: '4' });
+      wrapper.setProps({ user });
+    });
+
+    it('should load the project default workflow', () => {
+      wrapper.setState({ project, preferences });
+      controller.getSelectedWorkflow(project, preferences);
+      sinon.assert.calledOnce(workflowSpy);
+      sinon.assert.calledWith(workflowSpy, '1', true);
+    });
+  });
 });
--- a/app/pages/project/index.spec.js
+++ b/app/pages/project/index.spec.js
@@ -80,9 +80,6 @@ const preferences = apiClient.type('project_preferences').create({
   preferences: {},
   links: {
     project: project.id
-  },
-  save() {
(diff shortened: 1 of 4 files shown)


eatyourgreens:
I'm seeing a bunch of Promise rejection warnings when these tests run:
```
(node:12780) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: Cannot read property 'preferences' of null
(node:12780) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```
But the tests run, and pass, and I can't see where those null object property errors are being thrown.

eatyourgreens:
> LGTM code change is pretty small, seems some other changes crept in
@camallen I'm not sure I follow. I don't see any extra changes.

camallen:
The router change and the nav bar import seemed unrelated to checking the tester role to me. Ignore me if they were necessary.

eatyourgreens:
@camallen The router change loads [the Redux-conntected component](/zooniverse/Panoptes-Front-End/pull/4181/files#diff-71de65fb566f57e480dfd8b2ef49c7abR369), rather than the undecorated component that is now exported for these tests.

Navbar change is there because backticked import statements crash our tests, so was just added for safety.

eatyourgreens:
I'll merge this later unless @srallen can see anything I've missed.

app/pages/project/index.spec.js
```diff
@@ -0,0 +1,190 @@
...
+
+const project = apiClient.type('projects').create({
+  id: '1',
+  display_name: 'A test project',
+  experimental_tools: [],
+  get(type) {
+    return Promise.resolve(resources[type]);
+  },
+  links: {
+    active_workflows: [1, 2, 3, 4, 5],
```
srallen: If these represent ids, should they be strings?

app/pages/project/index.spec.js
```diff
@@ -0,0 +1,190 @@
...
+  id: project.links.avatar.id
+});
+
+const preferences = apiClient.type('project_preferences').create({
+  preferences: {},
+  links: {
+    project: project.id
+  },
+  save() {
+    Promise.resolve(preferences);
```
srallen: Can you reference a variable inside its own declaration? Also, the promise should be returned if this is mocking the API client.

eatyourgreens: This was added to stop `preference.save()` throwing warnings somewhere in the component. I don't think the return value is actually used in the code, so it might be safer to just return `null` here.

eatyourgreens: JavaScript is fine with circular references inside a declaration, but it probably is a terrible thing to do.

eatyourgreens:
@srallen latest two commits should address your comments.

eatyourgreens:
Rebased to latest master.

eatyourgreens:
Cheers. I think those warnings might be fixed by adding catch blocks to API requests in old code.

PR #4182, opened 2017-11-27
Convert email settings page

simoneduca:
Remove promise renderer and change listener

Staging branch URL: https://email-settings.pfe-preview.zooniverse.org/

Promise-renderer is still not quite dead, I don't think.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


simoneduca:
This is WIP because I need to figure out this
screen shot 2017-11-27 at 14 20 01

eatyourgreens:
This is still running promises in the component render. Promises should be run outside of render, probably when the component mounts, and the responses stored in state. Rendering should then render the component state.

Have a look at #4142 for an example of how to do this.

The specific error you've got seems to be because one of your functional components is returning a promise, rather than a React component.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+              {this.renderPagination()}
+            </td>
+          </tr>
+        );
+      }
+    });
+  }
+
+  renderTalkPreferences() {
+    return talkClient.type('subscription_preferences').get()
```
eatyourgreens: This is returning a promise instead of a React component.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+  constructor(props) {
+    super(props);
+    console.log(props)
+    this.state = {
+      page: 1,
+      projects: [],
+      projectPreferences: []
+    };
+    this.handlePreferenceChange = this.handlePreferenceChange.bind(this);
+    this.nameOfPreference = this.nameOfPreference.bind(this);
```
eatyourgreens: You only need to bind event handlers here. The other methods will have been automatically bound to the component context.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  handleProjectEmailChange(projectPreference, args) {
+    handleInputChange.apply(projectPreference, args);
+    projectPreference.save();
+    this.setState();
+  }
+
+  handlePreferenceChange(preference, event) {
+    return preference.update({
```
eatyourgreens: Why is this function returning a value?

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+      email_digest: event.target.value
+    })
+    .save();
+  }
+
+  nameOfPreference(preference) {
+    switch (preference.category) {
```
eatyourgreens: This whole block can probably be replaced by a dictionary object and a lookup.
```
{
  participating_discussions: "When discussions I'm participating in are updated.',
  followed_discussions: "When discussions I'm following are updated."
  etc.
}
```

Also, the English text strings should be moved into counterpart, but maybe not in this PR?

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+
+    }
+  }
+
+  renderProjectPreferences() {
+    const { projects, projectPreferences } = this.state;
+    return projectPreferences.map((projectPreference, i) => {
+      if (projects[i]) {
+        return (
+          <tr key={i}>
```
eatyourgreens: [Don't use array indexes as React keys.](https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318) `project.id` or `projectPreference.id` would be better here.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+      sort: 'display_name'
+    })
+    .then((preferences) => {
+      if (preferences.length > 0) {
+        const projectIDs = preferences.map(pref => pref.links.project);
+        apiClient.type('projects').get({
+          id: projectIDs
+        })
+        .then((projects) => {
+          return this.setState({
```
eatyourgreens: Why is this function returning a value?

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+          <thead>
+            <tr>
+              <th>Send me an email</th>
+              <th>Immediately</th>
+              <th>Daily</th>
+              <th>Weekly</th>
+              <th>Never</th>
+            </tr>
+          </thead>
+          {this.renderTalkPreferences()}
```
eatyourgreens: This will try to render a Promise as a React component, and throw the error you're seeing.

simoneduca:
Thanks for your comments @eatyourgreens. Actioning them now.

eatyourgreens:
Looking good @simoneduca Just need to get those project tests passing now.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,300 @@
...
+    };
+    this.handlePreferenceChange = this.handlePreferenceChange.bind(this);
+    this.nameOfPreference = this.nameOfPreference.bind(this);
+    this.sortPreferences = this.sortPreferences.bind(this);
+    this.talkPreferenceOption = this.talkPreferenceOption.bind(this);
+    this.getProjectForPreferences(props.user);
+  }
+
+  componentDidMount() {
+    talkClient.type('subscription_preferences').get()
```
eatyourgreens: Out of interest, why are you running the API request for project preferences in the constructor, but the API request for Talk preferences after the component mounts?

simoneduca:
Not quite, @eatyourgreens. Changing preferences is broken. Working on that.

eatyourgreens:
I think the tests are broken because the mock preferences don't have links defined, and you're now using links to find the project for each preference.

eatyourgreens:
Tests are passing for changing Talk preferences. Is it project preferences that don't update correctly? There's a test for those too but I think the tests might need updating for your new code.

app/pages/settings/email.jsx
simoneduca: Oh, that's just me using `componentDidMount()` as the [default place where to fetch data](https://daveceddia.com/where-fetch-data-componentwillmount-vs-componentdidmount/). Project prefs were explicitly fetched in `componentWillMount`, which can be effectively replaced by the constructor. There's probably not much in it, but I'm happy to either move this to the constructor or the talk preferences call inside here. Have you got a preference?

eatyourgreens: No strong preference, and the tests don't seem to be affected by using `componentDidMount`, but both calls should probably go in the same place.

simoneduca:
Mm interesting, I can neither update Talk nor projects.

eatyourgreens:
Is this set of tests passing for you? https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/settings/email.spec.js#L115-L121

simoneduca:
Yes, only project listing is failing. But I don't think that's right because I can't update talk prefs when running locally.
screen shot 2017-11-28 at 13 17 49

eatyourgreens:
The test checks that the preference resource is updated, but doesn't check that the radio buttons are rendered correctly after the update. Is that the problem you're seeing (radio buttons don't reflect the new state of the preferences.)

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,300 @@
...
+    .catch((error) => {
+      console.log('Something went wrong. Error: ', error);
+    });
+  }
+
+  handlePreferenceChange(preference, event) {
+    preference.update({
+      email_digest: event.target.value
+    })
+    .save();
```
eatyourgreens: This event handler should update Talk preferences in component state, if you want the change to be rendered in the form.

simoneduca:
Ah that's it, sorry my bad. Talk prefs are updated correctly. It's just the radio buttons not working properly.

app/pages/settings/email.jsx
simoneduca: Yes, I think a `this.setState()` should do it.

eatyourgreens:
@simoneduca you're missing a `setState` in the event handler, so `state.talkPreferences` never gets updated.

eatyourgreens:
#4183 fixes the tests for the project listing. Project preference event handlers still fail tests though.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,299 @@
...
+  }
+
+  handlePreferenceChange(preference, event) {
+    if (event.target.type === 'radio') {
+      const newTalkPrefs = Object.assign({}, this.state.talkPreferences);
+      preference.update({
+        email_digest: event.target.value
+      })
+      .save();
+      this.setState();
```
eatyourgreens: `setState` must always be called with the state changes as an argument.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,299 @@
...
+    } else {
+      preference.update({
+        email_communication: !!event.target.value
+      })
+      .save();
+      this.setState();
```
eatyourgreens: Same here.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,299 @@
...
+        });
+      }
+    })
+    .catch((error) => {
+      console.log('Something went wrong. Error: ', error);
+    });
+  }
+
+  handlePreferenceChange(preference, event) {
+    if (event.target.type === 'radio') {
```
eatyourgreens: This is very fragile, in that it assumes that noone will ever add a radio button to this component in future. A better test would be to switch based on the target name, or to just have two event handler functions: one for the Talk preferences and another one for the project preferences.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,299 @@
...
+      const newTalkPrefs = Object.assign({}, this.state.talkPreferences);
+      preference.update({
+        email_digest: event.target.value
+      })
+      .save();
+      this.setState();
+    } else {
+      preference.update({
+        email_communication: !!event.target.value
+      })
```
eatyourgreens: These are checkboxes, I think, so this should use `event.target.checked`.

eatyourgreens: See the tests for an example.

eatyourgreens:
I'm seeing a duplicate key warning when I run the tests.
```
Warning: Each child in an array or iterator should have a unique "key" prop. Check the render method of `EmailSettingsPage`.
```

app/pages/settings/email.jsx
eatyourgreens: Fixed in #4183

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,299 @@
...
+    const sortedPrefs = this.sortPreferences(talkPreferences);
+    return (
+      (sortedPrefs.length > 0) ?
+        <tbody>
+          {sortedPrefs.map((pref) => {
+            if (pref.category !== 'system' && pref.category !== 'moderation_reports') {
+              return (
+                <tr key={pref.id}>
+                  <td>{this.nameOfPreference(pref)}</td>
+                  {this.talkPreferenceOption(pref, 'immediate')}
```
eatyourgreens: Each of these options will need a unique key.

eatyourgreens: Also, as of React 14, functions that return components, like this one, can also be written as React components.

eatyourgreens: Ignore me. The key warning is bad test data in the tests. I've fixed it on the `email-settings-tests` branch.

eatyourgreens:
Key warnings are fixed in #4184

simoneduca:
I unnecessarily rebased master, sorry about the commit history @eatyourgreens

eatyourgreens:
I was wondering why all our commits have been added twice.

eatyourgreens:
The PR diff is now showing changes to files that aren't related to your work.

eatyourgreens:
Cleaned up, force pushed and re-staged.

eatyourgreens:
The project preferences list doesn't update when I change pages from the dropdown menu. I'm not seeing any errors in the console, though.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+
+  renderPagination() {
+    const { meta, page } = this.state;
+    if (meta) {
+      return (
+        <nav className="pagination">{'Page'}
+          <select
+            value={page}
+            disabled={meta.page_count < 2}
+            onChange={e => this.setState({ page: e.target.value })}
```
eatyourgreens: This changes the page number in component state, but that change doesn't trigger an API request for that page of projects/project preferences.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+    talkClient.type('subscription_preferences').get()
+    .then((preferences) => {
+      this.setState({
+        talkPreferences: preferences
+      });
+    });
+  }
+
+  getProjectForPreferences(user) {
+    return user.get('project_preferences', {
```
eatyourgreens: I think this is supposed to be called if `state.page` changes.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+      console.log('Something went wrong. Error: ', error);
+    });
+  }
+
+  handleProjectPreferenceChange(preference, event) {
+    preference.update({
+      email_communication: !!event.target.checked
+    })
+    .save()
+    .then(updatedPref => this.setState(updatedPref));
```
eatyourgreens: This overwrites component state with a preference resource, leading to the creation of `state.id`, `state._type` etc. You can check this in the React inspector in dev tools.

I think what you mean to update here is `state.projectPreferences`.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  handleTalkPreferenceChange(preference, event) {
+    preference.update({
+      email_digest: event.target.value
+    })
+    .save()
+    .then(updatedPref => this.setState(updatedPref));
```
eatyourgreens: Same here, except I think you mean to update `state.talkPreferences` here.

app/pages/settings/email.jsx
simoneduca: Yeah if block in `componentDidUpdate` is not executed, so I might have screwed up the condition conversion. Also I think two pages of projects seem too few. I have 5 in production. 🤔

simoneduca: Surely enough, I did. It should be `!(this.state.page === prevState.page)` instead of `!this.state.page === prevState.page`

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,289 @@
...
+    };
+    this.handleProjectPreferenceChange = this.handleProjectPreferenceChange.bind(this);
+    this.handleTalkPreferenceChange = this.handleTalkPreferenceChange.bind(this);
+    this.talkPreferenceOption = this.talkPreferenceOption.bind(this);
+    this.getProjectForPreferences(props.user);
+    this.getTalkPreferences();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (!this.state.page === prevState.page) {
```
eatyourgreens: Should be `this.state.page !== prevState.page`.

app/pages/settings/email.jsx
simoneduca: Ah my bad, thanks. I'm struggling to merge the updated resource into the existing array. How should I do that?

eatyourgreens: `preference` is a reference to an item in the `projectPreferences` array, not a copy, so you're already updating the array. 

If you want to make it this function more explicit, you could pass the array index instead of the preference, then `projectPreferences[index].update(changes).save()` and `setState({ projectPreferences })`.

eatyourgreens: Oh wait, you want to store the updated preference from the API as `projectPreferences[index]` before you set state. The key is still to pass the array index, rather than the preference itself. 

Or you could get the index from `projectPreferences.indexOf(preference)`, if you leave the handler as it is and pass the preference in.

simoneduca: Thanks, just pushed latest fixes.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,297 @@
...
+    };
+    this.handleProjectPreferenceChange = this.handleProjectPreferenceChange.bind(this);
+    this.handleTalkPreferenceChange = this.handleTalkPreferenceChange.bind(this);
+    this.talkPreferenceOption = this.talkPreferenceOption.bind(this);
+    this.getProjectForPreferences(props.user);
+    this.getTalkPreferences();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (!(this.state.page === prevState.page)) {
```
eatyourgreens: Better to use `!==` here.

simoneduca: Sure, I can see it's more readable. Are there any other reasons?

app/pages/settings/email.jsx
simoneduca: @eatyourgreens I'm happy to move the English strings into a counterpart. They'd include zoo-wide preferences and talk preferences. Anything else?

app/pages/settings/email.jsx
eatyourgreens: Because you're checking whether two variables are not equal, and `!==` is the strict not equal comparison operator in JavaScript. 

Here you're testing if two values are strictly equal, then performing a second test to invert the first test.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,297 @@
...
+    super(props);
+    this.state = {
+      page: 1,
+      projects: [],
+      projectPreferences: [],
+      talkPreferences: []
+    };
+    this.handleProjectPreferenceChange = this.handleProjectPreferenceChange.bind(this);
+    this.handleTalkPreferenceChange = this.handleTalkPreferenceChange.bind(this);
+    this.talkPreferenceOption = this.talkPreferenceOption.bind(this);
```
eatyourgreens: I don't think this line is necessary. Only event handlers need to be bound in the constructor. `this.talkPreferenceOption` is a React component.

simoneduca: Sorry, missed that.

app/pages/settings/email.jsx
eatyourgreens: All the form labels, I guess. 

Could you add Italian translations too? I think you can test by adding `language=it` as a query parameter.

simoneduca: I probably need to be reminded how to do that. I'll push something shortly and we discuss from there, thanks.

eatyourgreens: Italian strings are in `locales/it.json`.

app/pages/settings/email.jsx
```diff
@@ -1,9 +1,28 @@
...
+    beta: 'Get beta project email updates and become a beta tester'
+  },
+  talkPrefs: {
+    title: 'Send me an email',
+    frequency: {
+      immediate: 'Immediately',
+      day: 'Daily',
+      week: 'Weekly',
+      never: 'Never'
+    }
```
simoneduca: @eatyourgreens the talk preferences options are missing here because they're inside the lookup function, so I haven't figured out how to add them yet.

eatyourgreens: Language strings should be added to `app/locales/en.js`, not embedded in the component.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
+import React from 'react';
+import Translate from 'react-translate-component';
+import counterpart from 'counterpart';
+import apiClient from 'panoptes-client/lib/api-client';
+import talkClient from 'panoptes-client/lib/talk-client';
+import AutoSave from '../../components/auto-save';
+import handleInputChange from '../../lib/handle-input-change';
+
+counterpart.registerTranslations('en', {
```
eatyourgreens: This whole block should be moved to [the English locale file](/zooniverse/Panoptes-Front-End/blob/master/app/locales/en.js).

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
...
+    });
+  }
+
+  render() {
+    return (
+      <div className="content-container">
+        <p>
+          <AutoSave resource={this.props.user}>
+            <span className="form-label">
+              Email address
```
eatyourgreens: This should be translated.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
...
+        </table>
+
+        <p><strong>Project email preferences</strong></p>
+        <table>
+          <thead>
+            <tr>
+              <th>
+                <i className="fa fa-envelope-o fa-fw" />
+              </th>
+              <th>Project</th>
```
eatyourgreens: This should be translated.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
...
+              <Translate component="th" content="talkPrefs.frequency.immediate" />
+              <Translate component="th" content="talkPrefs.frequency.day" />
+              <Translate component="th" content="talkPrefs.frequency.week" />
+              <Translate component="th" content="talkPrefs.frequency.never" />
+            </tr>
+          </thead>
+          {this.renderTalkPreferences()}
+        </table>
+
+        <p><strong>Project email preferences</strong></p>
```
eatyourgreens: This should be translated.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
...
+            <input
+              type="text"
+              className="standard-input full"
+              name="email"
+              value={this.props.user.email}
+              onChange={handleInputChange.bind(this.props.user)}
+            />
+          </AutoSave>
+        </p>
+        <p><strong>Zooniverse email preferences</strong></p>
```
eatyourgreens: This should be translated.

app/pages/settings/email.jsx
eatyourgreens: For the Talk preferences, you need to remove the `nameOfPreference` method and replace it with references to counterpart keys instead eg. `<Translate content="settings.talk.participating_discussions" />`.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,315 @@
+import React from 'react';
+import Translate from 'react-translate-component';
+import counterpart from 'counterpart';
```
eatyourgreens: You don't need to import counterpart.

simoneduca: Don't I? I was following this https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/index.cjsx#L13

eatyourgreens: No, you don't. That line imports counterpart translations globally for the whole app.

eatyourgreens: There's an example of this in #4072, which extracted out the English language strings for the survey task summary component.

simoneduca:
@eatyourgreens I've added the relevant Italian strings and tried testing with `?language=it`, but it does nothing 😭

eatyourgreens:
Hmm. You might need to navigate to a project to actually change the language. I think the code is in the project page at the moment.

eatyourgreens:
Looking good! I've added some more tests in #4188

simoneduca:
I think I caught all the strings. Should be good to go.

eatyourgreens:
Excellent! I'll give it another look over after lunch, but this should be good now.

app/pages/settings/email.jsx
```diff
@@ -0,0 +1,309 @@
...
+      email_digest: event.target.value
+    })
+    .save()
+    .then((updatedPref) => {
+      talkPreferences[index] = updatedPref;
+      this.setState({ talkPreferences });
+    });
+  }
+
+  nameOfPreference(preference) {
```
eatyourgreens: Can this be deleted now?

simoneduca: Done

PR #4189, opened 2017-11-29
Extract functional components

eatyourgreens:
Extracts components for Talk preferences, project preferences and pagination.
Fixes array sort being called more than once for Talk preferences.
Adds a footer to the project listing table.
Fixes page number being passed as a string, instead of a number, in the page change event handler.

### DIFF
--- a/app/pages/settings/email.jsx
+++ b/app/pages/settings/email.jsx
@@ -124,7 +124,7 @@ Pagination.defaultProps = {
   meta: {
     page_count: 1
   },
-  page: '1'
+  page: 1
 };
 
 class EmailSettingsPage extends React.Component {
--- a/app/pages/settings/email.jsx
+++ b/app/pages/settings/email.jsx
@@ -70,7 +70,7 @@ function ProjectPreferences({ projects, projectPreferences, onChange }) {
                   type="checkbox"
                   name="email_communication"
                   checked={projectPreference.email_communication}
-                  onChange={(e) => onChange(i, e)}
+                  onChange={e => onChange(i, e)}
                 />
               </td>
               <td>
@@ -124,8 +124,8 @@ Pagination.defaultProps = {
   meta: {
     page_count: 1
   },
-  page: 1
-}
+  page: '1'
+};
 
 class EmailSettingsPage extends React.Component {
   constructor(props) {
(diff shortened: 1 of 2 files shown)


app/pages/settings/email.jsx
```diff
@@ -5,11 +5,133 @@ import talkClient from 'panoptes-client/lib/talk-client';
...
+  },
+  page: '1'
+}
+
 class EmailSettingsPage extends React.Component {
   constructor(props) {
     super(props);
     this.state = {
-      page: 1,
+      page: '1',
```
simoneduca: Why a string? It was an integer in [the original](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/settings/email.cjsx#L16)

eatyourgreens: Try changing it to `React.PropTypes.number` and watch the invalid prop warnings pile up in the dev console.

eatyourgreens: Or change this back to a number and watch the warnings appear when you run the tests.

simoneduca: Mm that's annoying, thanks.

eatyourgreens: Do you want me to fix the email settings component so that page number is always passed around as a number?

eatyourgreens: Fixed in 2fa1281

simoneduca:
Waiting for the tests to pass before merging.

app/pages/settings/email.jsx
```diff
@@ -5,6 +5,128 @@ import talkClient from 'panoptes-client/lib/talk-client';
...
+  }),
+  page: React.PropTypes.number,
+  onChange: React.PropTypes.func
+};
+
+Pagination.defaultProps = {
+  meta: {
+    page_count: 1
+  },
+  page: '1'
```
eatyourgreens: Didn't I just fix this? Page is a number, not a string.

eatyourgreens:
You've broken the page prop type warning that I fixed.

simoneduca:
Oops, sorry. I've stash my changes, pulled your changes and then pop them on. Then pushed. How do I fix it?

eatyourgreens:
Change the default page value back to a number. See my last comment on the code.

simoneduca:
Here you go.

Issue #4187, opened 2017-11-28
pages/about ES6 conversion


jelliotartz:
#### Purpose of issue: an overview of work needed to convert all the files in `pages/about` to ES6.

Most of these conversions look pretty straightforward - ie, they should be stateless functional components that follow a pretty standard workflow:
- write failing unit tests for each component
- update router for ES6 syntax
- move text to translation file
- make tests pass
- address any linter errors
- create staging branch
- pass all required manual testing and review checklist items

Happy to adjust if desired :) 

## Files to be converted to ES6
`about/`
- [x] `about-home`
- [x] `acknowledgements`
- [x] `contact-page`
- [x] `faq-page`
- [x] `index`
- [ ] `publications-page`
- [ ] `team-page`

PR #4190, opened 2017-11-29
AboutHome ES6 conversion

jelliotartz:
## Describe your changes
- Converts AboutHome from coffeescript to ES6.
- Adds unit tests for AboutHome component.
- Moves text to `locales/en`.
- Staging branch URL: https://about-home-es6-conversion.pfe-preview.zooniverse.org/
- Towards #4187.

### Required Manual Testing

### Review Checklist

- [x] Does it work on mobile?

### Optional


app/locales/en.js
```diff
@@ -185,5 +185,14 @@ export default {
...
+      anyoneCanResearch: '### At the Zooniverse, anyone can be a researcher\n\nYou don’t need any specialised background, training, or expertise to participate in any Zooniverse projects. We make it easy for anyone to contribute to real academic research, on their own computer, at their own convenience.\n\nYou’ll be able to study authentic objects of interest gathered by researchers, like images of faraway galaxies, historical records and diaries, or videos of animals in their natural habitats. By answering simple questions about them, you’ll help contribute to our understanding of our world, our history, our Universe, and more.\n\nWith our wide-ranging and ever-expanding suite of projects, covering many disciplines and topics across the sciences and humanities, there\'s a place for anyone and everyone to explore, learn and have fun in the Zooniverse. To volunteer with us, just go to the [Projects](/projects) page, choose one you like the look of, and get started.',
```
rogerhutchings: Instead of using newline characters, you can create multi-line strings using backticks:

```
`
Paragraph 1

Paragraph 2
`
```

app/pages/about/about-home.spec.js
```diff
@@ -0,0 +1,23 @@
...
+
+describe('AboutHome', function () {
+  let wrapper;
+
+  before(function () {
+    wrapper = shallow(<AboutHome />);
+  });
+
+  it('renders without crashing', function () {
+    assert.equal(wrapper, wrapper);
```
rogerhutchings: This is always going to be true. I'm pretty sure this test can just be an empty function, as the `before` block renders the component.

app/locales/en.js
jelliotartz: Using the backticks doesn't work with our Markdown components though, which we're using to render these sections. If we stop using Markdown, will we need to introduce our own styles if we want to keep the look of the site the same across all these static pages (see also, eg, privacy and security policy). Happy to entertain other possibilities, but not sure if using backticks/no Markdown is the easiest way to get what we want here.

image


Backticks render like this ^^

image


Markdown elements render like this ^^

rogerhutchings: Not suggesting not using Markdown, but I didn't know the markdown components couldn't handle backticks. ¯\_(ツ)_/¯

rogerhutchings: By the way - did you use the backticks instead of the single quotes around the strings, or inside?

jelliotartz: Instead of. Using them inside fails to compile.

srallen: That makes sense because that's telling the javascript to use a template literal by using back ticks.

srallen: Since these are the strings eventually translator volunteers may see we may want to just break it up into more key/value pairs. This is probably fine for now though.

PR #4204, opened 2017-12-05
Add project state radio buttons to admin

rogerhutchings:
Staging branch URL: https://project-status-admin.pfe-preview.zooniverse.org/

Fixes @mrniaboc and @camallen eyeballing me as soon as I entered the office this morning.

Adds some radio buttons to the project admin page so an admin can set a project's state to either live (`''`), paused (`'paused'`) or finished (`'finished'`)

I found some other states on projects on the staging server, e.g. `'live'` and `'development'`. Should these be supported as well @camallen?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -188,7 +188,7 @@ class ProjectStatus extends Component {
     if (!this.state.project) {
       return <LoadingIndicator />;
     }
-    console.info(this)
+
     return (
       <div className="project-status">
         <ProjectIcon project={this.state.project} />
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -199,6 +199,7 @@ class ProjectStatus extends Component {
             <label htmlFor="project-state-active">
               <input
                 id="project-state-active"
+                name="project-state-buttons"
                 type="radio"
                 value=""
                 checked={this.state.project.state === ''}
@@ -209,6 +210,7 @@ class ProjectStatus extends Component {
             <label htmlFor="project-state-paused">
               <input
                 id="project-state-paused"
+                name="project-state-buttons"
                 type="radio"
                 value="paused"
                 checked={this.state.project.state === 'paused'}
@@ -219,6 +221,7 @@ class ProjectStatus extends Component {
             <label htmlFor="project-state-finished">
               <input
                 id="project-state-finished"
+                name="project-state-buttons"
                 type="radio"
                 value="finished"
                 checked={this.state.project.state === 'finished'}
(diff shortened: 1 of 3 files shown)


mrniaboc:
Hahaha great work Roger!

There are already buttons for 'live' and 'development' status on the admin page. Perhaps the new live (' ') status should be called 'active' to avoid confusion and match the projects page?

rogerhutchings:
It's called `Active` already, I'm a muppet and wrote the PR comment wrong

screenshot 2017-12-05 11 28 19

mrniaboc:
Ah ha! I couldn't see the preview page, sorry! Good man.

eatyourgreens:
You can't deselect a radio button, only change the selection to another button in the same named group, so one of these buttons should always be selected by default.

app/pages/admin/project-status.jsx
```diff
@@ -185,6 +192,43 @@ class ProjectStatus extends Component {
...
       <div className="project-status">
         <ProjectIcon project={this.state.project} />
+
+        <div className="project-status__section project-status__section--state">
+          <h4>Project State</h4>
+          <fieldset>
+            <label htmlFor="project-state-active">
+              <input
+                id="project-state-active"
+                type="radio"
```
eatyourgreens: name attribute is missing.

rogerhutchings:
As discussed offline - there are other states that a project may be in, so having a default state may not work.

eatyourgreens:
@rogerhutchings if I accidentally select 'Active' for a project that's still in development, then can't deselect it, does anything break?

Otherwise, this looks good to me.

app/pages/admin/project-status.jsx
```diff
@@ -185,6 +192,46 @@ class ProjectStatus extends Component {
     return (
       <div className="project-status">
         <ProjectIcon project={this.state.project} />
+
+        <div className="project-status__section project-status__section--state">
+          <h4>Project State</h4>
+          <fieldset>
+            <label htmlFor="project-state-active">
```
camallen: Unfortunately this wasn't setup with active in the enum states api side. I apologise for this. The enum states you can set are paused and finished only. Thus the active setting here acts as a reset instead of an actual state. I would assume clicking active would persist that change through the UI. Perhaps we should rename active to reflect this behaviour. 

The state param is a ***** mess api side, if the enum field is set it uses that if not it proxies to the live setting. I have no idea why but it does.

rogerhutchings: Right - it now sets it to an empty string instead of `live`

PR #4207, opened 2017-12-06
[WIP] Use user passed as a prop consistently

srallen:
Staging branch URL: https://use-user-prop.pfe-preview.zooniverse.org/

Fixes a prop type warning and a bug found while reviewing another PR

I refactored a couple of project home components to consistently use the user prop instead of using a mix of the user passed as a prop and passed around in context. I also fixed a Gravity Spy related bug. I found that if you are on the Gravity Spy project home page and you sign out, that the workflow buttons were not only still available, but all the buttons were enabled. What should have happened is that the buttons should be hidden, so this regression may have been introduced during a coffeescript -> ES6 conversion or other refactoring at some point.

How to test:
Camera Catalog is an example of a project that does not use workflow promotion and allows users to choose their workflow. The buttons should be available regardless of the presence of a signed in user: https://use-user-prop.pfe-preview.zooniverse.org/projects/panthera-research/camera-catalogue?env=production 

Gravity Spy uses workflow promotion. When a user is signed in, the buttons should be visible and only the workflows the user has access to are enabled. When there isn't a user, the buttons should not be rendered: https://use-user-prop.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy?env=production

And any other projects that do not use workflow promotion and do not have the "Allow user choose workflow" setting set should not have the buttons visible.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/home/home-workflow-buttons.jsx
+++ b/app/pages/project/home/home-workflow-buttons.jsx
@@ -14,9 +14,6 @@ counterpart.registerTranslations('en', {
 export default class ProjectHomeWorkflowButtons extends React.Component {
   constructor(props) {
     super(props);
-    this.state = {
-      showWorkflows: false
-    };
 
     this.shouldWorkflowBeDisabled = this.shouldWorkflowBeDisabled.bind(this);
     this.renderRedirectLink = this.renderRedirectLink.bind(this);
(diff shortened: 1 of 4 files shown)


srallen:
WIP because I still got to fix the tests.

srallen:
The tests are passing now.

srallen:
I broke the get started button when it's working as a toggle to show/hide the workflow buttons, so I'm gonna rethink this a bit and also talk with @beckyrother about it.

srallen:
Just talked with @beckyrother and what she and I decided is that the Get Started button styled link should only be present for projects that don't let the user choose the workflow. For projects that do let the user choose the workflow, the button styled link will be gone and the call to action will be the header for the workflow buttons div. This eliminates the inconsistent behavior and I don't need to try to keep this working while also trying to fix the Gravity Spy bug I found.

beckyrother:
Just thinking about this more – on cases where we remove the "get started" button, can we center the "learn more" button and add text "get started" with a fontawesome `fa-arrow-down` next to it? Just to make sure people know to scroll.

srallen:
I'm gonna  close this PR for now and reopen a new one with the refactoring as described in the comments.

PR #4218, opened 2017-12-11
Facelift - Project Header Bar

rogerhutchings:
## URLs

Staging branch URL: https://facelift-headerbar.pfe-preview.zooniverse.org/

Staging project: https://facelift-headerbar.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats?facelift=true

### With project background / avatar set

fireshot capture 5 - i fancy cats zooniverse_ - http___localhost_3735_projects_brooke_i-fancy-cats

### Without project background /avatar set

fireshot capture 7 - feedback test _ - http___localhost_3735_projects_rogerhutchings_feedback-test

## Description

Adds the redesigned project header bar from @beckyrother's new classifier designs. It's hidden behind a feature flag - add `?facelift=true` to the URL to enable it.

Adds [styled-components](styled-components.com) for styling, and the [object-rest-spread](https://babeljs.io/docs/plugins/transform-object-rest-spread/) transform for Babel.

### Still to do

- [x] Hover states for nav links
- [x] Adapt to narrow screen sizes
- [ ] Tests

@beckyrother, can you advise me on the first two?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/project/components/ProjectNavbar/ProjectNavbarContainer.jsx
```diff
@@ -0,0 +1,119 @@
...
+      label: counterpart(link.translationPath),
+      url: link.url
+    }));
+  }
+
+  render() {
+    const avatarSrc = _.get(this.props.projectAvatar, 'src', undefined);
+    const backgroundSrc = _.get(this.props.background, 'src', undefined);
+    const navLinks = this.getNavLinks();
+    const projectTitle = _.get(this.props.project, 'title', undefined);
```
eatyourgreens: Language strings should be rendered from a translation, now that we've hooked projects up to the translations API.

rogerhutchings: Is the translations API available on staging?

eatyourgreens: Staging and production. It's already in use for projects e.g. the existing project navigation takes a translation prop.

rogerhutchings: Cool, I've just reused that prop

rogerhutchings:
Also @beckyrother - the current navbar includes links to external social media sites etc. Are these going to be moved to the `Connect with [Project]` widget, or are they duplicated there?

beckyrother:
@rogerhutchings this looks amazing! thanks for getting started so quickly, so exciting to see this coming together. 
- please increase link letter-spacing to 1.5px (seeing the type the rendered in browser, needs a little more space)
**your questions:** 
- hover state should be a teal underline just like in the main nav, and should be #007482 (zooniverse teal)
- mobile view coming! 
- we'll keep the social links the same way they're treated now, as icons in the menu bar. so that may mean duplicating them if the researcher enables the `connect with [Project]` widget, which is fine.

beckyrother:
@rogerhutchings Mobile view! [Here's the InVision link.](https://invis.io/8XEYUIMHS#/269381735_000_Mobile) Please really only look at this for the menu – I realize the image doesn't have a toolbar, which will obviously need to be resolved. On click the menu should slide down from behind the top bar:

classify-mobile-menu-1

srallen:
I'm wrapping up some Intro2Astro stuff today and tomorrow, but I should be able to pick this up to finish next week.

srallen:
I fixed the broken tests and wrote some new ones that uncovered a couple of minor bugs that I also fixed. I'm going to continue to write unit tests so we get good coverage on these new components.

beckyrother:
Would you like design notes here?

srallen:
@beckyrother Yes that'd be great!

srallen:
Just to note here, the tests pass locally for me. Looking into why they fail on Travis.

beckyrother:
Top nav:
- [ ] Increase the space between links to `2rem`
- [ ] Is it possible to make the underline 3px #fff 
- [ ] Looks like the letter spacing isn't working?

srallen:
@beckyrother I've made the few style tweaks we talked about, so this is ready for another design review. Thanks!

srallen:
@eatyourgreens I'd like to get this merged as is now because of the size of the PR and continue working on the tests in another branch. I think this will be safe to do so because the new header isn't loaded unless one uses the query param that is setup.

beckyrother:
@srallen Looking at this again, not sure the underline under the project name is needed when on the project home page. Mobile nav looks great, but I should have done an active link example – can that be #007482 , as in the cell is that color when the menu opens?

srallen:
@beckyrother the staging site is ready for another design review. I removed the active styling from the project title, but retained the underline for its hover and focus states. I also added the active state to the mobile nav link buttons.

beckyrother:
@srallen this looks great! approved from me 👍

srallen:
I've rebased and regenerated the package-lock.json

shaunanoordin:
Quick thoughts before I run tests:
- the code read generally looks good, although I may be missing something with 3K+ new lines.
- [minor] some components (ProjectNavbar) don't specify defaultProps, but I don't imagine this will be an issue.
- This is the first time I've seen tagged template literals, so I can't comment on its usage. TIL, though. 👌 
  ```
  const StyledExternalLink = styled.a`
    ${commonStyles}
  `;
  ```
- ❓ I'm curious about our usage of styled components (or rather, the `styled-compontents` library) instead of setting the styles in their CSS. It's such a different way of doing things (i.e. previous separation of function.js and style.css, vs new implementation of style _inside_ functions) that I'd like to better understand the pros/cons of the approach.

app/pages/project/components/ProjectNavbar/helpers/userHasLabAccess.js
```diff
@@ -0,0 +1,11 @@
+function userHasLabAccess({ projectRoles, user }) {
+  return projectRoles.some((role) => {
+    if (role.links.owner.id === user.id) {
```
shaunanoordin: [minor] Any chance that `role.links.owner` or `role.roles` would be null/undefined for any reason?

shaunanoordin:
Uh, I can't Tab-navigate the new Project header's nav links in Firefox; the header just gets skipped altogether. I'll look into this WTF-ness - I can confirm it works on Chrome and Safari.

shaunanoordin:
OK, never mind - it looks like the Firefox tabbing issue is a separate issue altogether. Tab nav is borked for FF right now, even on our current master on zooniverse.org

eatyourgreens:
Tabbing in Firefox has come as a problem before, but I don't remember how we fixed it.

app/pages/project/components/ProjectNavbar/helpers/userHasLabAccess.js
srallen: If they are, then the API would be pretty broken. I could still add existential checks just to be sure.

srallen:
@shaunanoordin Thanks!

To answer the question about why we're using styled-components here is that it's an experiment. CSS in JS has been popular in the react community for some time, however, all of the implementations until now was just JS objects in the style property of the component which I think has a few issues including losing all of the benefits of a CSS pre-processor like Stylus, a lot of repetitiveness, and being unfriendly to designers that may only know CSS. I didn't find that there were many gains with the modularization of the CSS this way.

The styled-components library instead generates CSS and classes that is added into the head of the page then uses the generated class names on the component. The template literals allow us to write standard CSS. On top of all that, it supports theme files that we can pass into the app's context using the built in `<ThemeProvider />` component, so we can have the best of both worlds of having both global styles defined and encapsulated component specific styles. One of the down sides is that as far as I know, we'll be losing some of the nice syntactic sugar Stylus provides us, however, this may be solved problem with some plugins and extensions I don't know about yet. 

Another motivation is that Grommet v2 is using styled-components. If we write a styled-components theme and a set of shared components to go into ZRC that uses it, then there's great benefit to any projects that start with Grommet v2, plus our custom theme and set of components. It'll take a bit to get there, but that's the idea.

I'm working on a branch to add the theme as well as hopefully getting tests working for the components that use styled-components. 

As an aside, it looks like IE isn't using the blur filter either. It probably doesn't support that either. I can take a look in my follow up PR.

eatyourgreens:
Thanks @srallen for the explanation of styled components!

> The styled-components library instead generates CSS and classes that is added into the head of the page then uses the generated class names on the component.

This is a really bad idea, so we should be very cautious about using them.

eatyourgreens:
Turns out this PR upgraded `react-select`, which has broken Notes from Nature.

PR #4222, opened 2017-12-12
Add project page tests for the project disclaimer

eatyourgreens:
Fixes broken component import.

Fixes broken default props.

Add some tests for the project disclaimer.

### DIFF
--- a/app/pages/project/project-page.spec.js
+++ b/app/pages/project/project-page.spec.js
@@ -4,23 +4,29 @@ import { expect } from 'chai';
 import { project, workflow } from '../dev-classifier/mock-data';
 import ProjectPage from './project-page';
 
+function Page() {
+  return(
+    <p>Hello world!</p>
+  )
+}
+
 describe('ProjectPage', () => {
   const background = {
     src: 'the project background image url'
   };
 
   it('should render without crashing', () => {
-    shallow(<ProjectPage />);
+    shallow(<ProjectPage><Page /></ProjectPage>);
   });
 
   it('should render the project nav bar', () => {
-    const wrapper = shallow(<ProjectPage />);
+    const wrapper = shallow(<ProjectPage><Page /></ProjectPage>);
     const navElement = wrapper.find('ProjectNavbar').first();
     expect(navElement).to.have.lengthOf(1);
   });
 
   it('should receive a background, project and workflow props', () => {
-    const wrapper = mount(<ProjectPage background={background} project={project} workflow={workflow} />);
+    const wrapper = mount(<ProjectPage background={background} project={project} workflow={workflow} ><Page /></ProjectPage>);
     expect(wrapper.props().background).to.equal(background);
     expect(wrapper.props().project).to.equal(project);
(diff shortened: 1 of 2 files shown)


app/pages/project/project-page.spec.js
```diff
@@ -2,27 +2,53 @@ import React from 'react';
...
   it('should receive a background, project and workflow props', () => {
-    const wrapper = mount(<ProjectPage background={background} project={project} workflow={workflow} />);
+    const wrapper = mount(<ProjectPage background={background} project={project} workflow={workflow} ><Page /></ProjectPage>);
     expect(wrapper.props().background).to.equal(background);
     expect(wrapper.props().project).to.equal(project);
     expect(wrapper.props().workflow).to.equal(workflow);
   });
+  
+  describe('with a launch-approved project', () => {
+    const project = { launch_approved: true };
```
simoneduca: `project` is already declared above. Perhaps `approved` could be used instead?

app/pages/project/project-page.spec.js
```diff
@@ -2,27 +2,59 @@ import React from 'react';
...
   });
 
   it('should render the project nav bar', () => {
-    const wrapper = shallow(<ProjectPage />);
+    const wrapper = shallow(<ProjectPage><Page /></ProjectPage>);
     const navElement = wrapper.find('ProjectNavbar').first();
     expect(navElement).to.have.lengthOf(1);
   });
 
   it('should receive a background, project and workflow props', () => {
```
simoneduca: In hindsight I should remove this test. It checks something that is already React's job, i.e. checking components receive props and I don't think we should test React.

app/pages/project/project-page.spec.js
```diff
@@ -2,27 +2,59 @@ import React from 'react';
...
+  describe('without approval', () => {
+    project.launch_approved = false;
+    const wrapper = shallow(<ProjectPage project={project} ><Page /></ProjectPage>);
+    const disclaimer = wrapper.find('Translate[className="project-disclaimer"]');
+    it('should render the Zooniverse disclaimer.', () => {
+      expect(disclaimer).to.have.lengthOf(1);
+    });
+    it('should render the disclaimer immediately after its children.', () => {
+      expect(wrapper.childAt(1).name()).to.equal('Page');
+      expect(wrapper.childAt(2)).to.deep.equal(disclaimer);
```
simoneduca: The test passes but, as we discussed, it's brittle and it'd break if we add anything to the page because the indexes are fixed. I think it's ok for now, but ideally we want to test that the _next_ child after `Page` is `disclaimer`.

simoneduca:
Minor comments, LGTM. I'll merge once tests pass.

app/pages/project/project-page.spec.js
eatyourgreens: Yeah, checking the relative positions of the child page and the disclaimer in the components child nodes would be much better than checking the absolute positions of both.

simoneduca:
Mm Travis says the build hasn't started yet, which is odd. I'll merge anyway, given that it's not a merge into master.

PR #4233, opened 2017-12-14
Add test watch and run npm script

camallen:
avoid the time delay of running the tests manually, re-run the tests on code / test file system changes.

Note: to get this working you'll need to comment out all the api client resource `.create` statements in https://github.com/zooniverse/Panoptes-Front-End/blob/9a13d927365d65bbedaa7b5cc16018f680301ac0/app/pages/settings/email.spec.js#L10 once that is done the tests will run after file system changes on the code.

Those commented lines should be replaced with test doubles instead of the actual API calls to create resources.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,6 +119,7 @@
     "stage-with-docker": "./bin/run-through-docker.sh 'npm run stage'",
     "_deploy": "export PATH_ROOT=www.zooniverse.org; npm run _build && npm run _publish",
     "deploy": "export NODE_ENV=production; npm run _deploy",
+    "test-and-watch": "NODE_ENV=development BABEL_ENV=test mocha --watch test/setup.js $(find app -name *.spec.js*) --reporter nyan || true",
     "test": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js*) --reporter nyan || true",
     "test-travis": "NODE_ENV=development BABEL_ENV=test mocha test/setup.js $(find app -name *.spec.js*)",
     "lint": "eslint ./app --ext .jsx,.js",


eatyourgreens:
Problem is, if you comment out those lines then you break the email settings tests.

camallen:
Yeah i get that - this isn't a solution to fix those tests just a proposal for running tests. Those commented lines should be replaced with test doubles instead of the actual API calls to create resources, so the tests can run without using the api code.

eatyourgreens:
Those aren't API calls. You'd need to add `.save()` to call Panoptes. `.create()` just sets up a local JavaScript object with the passed properties.

camallen:
Sorry i should have said client calls not API calls, though it's good to know it's not hitting the network. I'll look to include the work in #4234 before moving this from WIP.

simoneduca:
Just waiting for come to remove the `WIP` flag

eatyourgreens:
This could do with a rebase to the latest master, but should be all good now.

simoneduca:
I've rebased locally and this is what I've found:

1. the command doesn't run when a source file is changed, but only when a spec.js file is changed.
2. tests break when the command runs the second time, see screenshot
screen shot 2018-01-17 at 11 17 28
It looks like `sinon` isn't used correctly in [Cam's test here](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/project-page.spec.js#L186).

Question re 1: isn't the point of running tests automatically to check tests are passing (or failing), when source code is changed and not/as well as when tests are changed?

camallen:
@simoneduca was sinon updated hence the API changed vs when i added that spec?  You can change the files that are being watched to add the source code as well, i probably forgot that when initially setting it up.

eatyourgreens:
@camallen I updated your tests, using a codemod script, when I upgraded sinon.

eatyourgreens:
Broken sinon spy is here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/project-page.spec.js#L137

From the error, the spy already exists when sinon tries to create it.

eatyourgreens:
Here's how to clean up the client spies once tests have run: https://github.com/zooniverse/Panoptes-Front-End/blob/c446fe91bd33438351aedfcb76e99da538248910/app/pages/settings/email.spec.js#L145-L148

simoneduca:
As expected here's another test that breaks when running `test-and-watch`: https://github.com/zooniverse/Panoptes-Front-End/blob/c446fe91bd33438351aedfcb76e99da538248910/app/pages/settings/email.spec.js#L11
I'll try to compile a list with all the breaking tests and put it here.

simoneduca:
screen shot 2018-01-17 at 14 17 45

screen shot 2018-01-17 at 14 17 00

screen shot 2018-01-17 at 14 17 23
screen shot 2018-01-17 at 14 18 23
screen shot 2018-01-17 at 14 18 40

simoneduca:
I believe those above are all the problematic tests.

simoneduca:
Actually that's false. There's more than that. I'll open a new PR to fix these tests and then come back to this.

simoneduca:
Superseded by https://github.com/zooniverse/Panoptes-Front-End/pull/4302

PR #4249, opened 2017-12-29
Fix Sign-in Prompt from Subject Preview

### LINKED ISSUE #4239: Looking at a subject from the project builder shows "You should sign in"
## Expected behavior

No message.

## Current behavior

"You should sign in!" appears underneath the subject image.

## Steps to replicate

Go to project builder, go to subject set. Click the Eye icon next to a subject.

## Additional information

- **Operating system:**
- **Browser:**

wgranger:
Staging branch URL: https://fix-4239.pfe-preview.zooniverse.org/

Fixes #4239  .

Describe your changes.
The issue above concerns the wiggling "you should sign-in" prompt from appearing in the project builder's subject preview. The issue was that we were reusing the `SubjectViewer` component for the preview without passing in the user prop, so the SV had no idea if a user was present or not. 

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/subject-set.cjsx
+++ b/app/pages/lab/subject-set.cjsx
@@ -65,6 +65,9 @@ SubjectSetListingTable = React.createClass
 SubjectSetListing = React.createClass
   displayName: 'SubjectSetListing'
 
+  contextTypes:
+    user: React.PropTypes.object
+
   getDefaultProps: ->
     subjectSet: {}
 
@@ -123,7 +126,7 @@ SubjectSetListing = React.createClass
 
   previewSubject: (subject) ->
     alert <div className="content-container subject-preview">
-      <SubjectViewer subject={subject} />
+      <SubjectViewer subject={subject} user={@context.user} />
     </div>
 
   removeSubject: (subject) ->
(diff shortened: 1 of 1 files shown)


app/pages/lab/subject-set.cjsx
```diff
@@ -65,6 +65,9 @@ SubjectSetListingTable = React.createClass
 SubjectSetListing = React.createClass
   displayName: 'SubjectSetListing'
 
+  contextTypes:
+    user: React.PropTypes.object
```
simoneduca: Recently a few problems emerged which were caused by `context` not playing ball nicely, so I would stop using it. `user` is available as prop in the tree, so why not just use that instead? You could add it to `<SubjectSetListing ... user={@props.user}/>` `<SubjectViewer subject={subject} user={@props.user} />`.

wgranger: Makes sense. I initially chose context because of ease of addition, but I'm definitely up to make the change if passing down props is more reliable. Will push up the adjustment shortly.

Issue #4039, opened 2017-09-13
Re-adding an object to a collection doesn't seem to work


crowston:
## Expected behavior
An object added and then deleted from a collection can be added back to the collection. 
## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Error message: "There are no subjects to add or subjects are already added to My collection."
## Steps to replicate
Add an object to a collection from its talk page. Delete it from the collection (in another page). Add it back again. (In Gravity Spy).

## Additional information

- **Operating system:**
Mac OS X
- **Browser:**
Firefox.

srallen:
Likely race condition, cache, or state that is out of sync.

eatyourgreens:
I think this was fixed by #4250.

PR #4250, opened 2017-12-29
Allow Subject to Be Added to Empty Collection

### LINKED ISSUE #4070: Empty collection can't add new subjects
## Expected behavior
Empty collections should be able to have new subjects added.

## Current behavior
Error with a spinning wheel on add modal
`Uncaught TypeError: Cannot read property 'includes' of undefined` in zooniverse.org console.

image

## Steps to replicate
1. create a new collection by collecting an image.
2. remove that image from the collection (note this doesn't update the default image src by the way, it probably should if there is only 1 subject in the collection). 
3. find another subject to collect (reload the page) and attempt to add that subject to the now empty collection.

wgranger:
Staging branch URL: https://fix-4070.pfe-preview.zooniverse.org/

Fixes #4070  .

Describe your changes.
The linked issue concerns the inability to add a subject to an empty collections set. This branch solves the issue, however, there seems to also be a caching issue affecting how the collections page is updated. This seemed to be a bigger issue (which I saved for another PR), but I can look into it more with this PR if desired. 

The additional bug mentioned above appears to have already been documented as #4039.

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional

### DIFF
--- a/app/collections/collections-manager.spec.js
+++ b/app/collections/collections-manager.spec.js
@@ -2,18 +2,35 @@ import React from 'react';
 import assert from 'assert';
 import { shallow } from 'enzyme';
 import sinon from 'sinon';
+import apiClient from 'panoptes-client/lib/api-client';
 import CollectionsManager from './collections-manager';
 
 const project = {
   id: '342'
 };
 
+const selectedCollection = [{
+  collection: apiClient.type('collections').create({
+    display_name: 'test collection',
+    description: '',
+    private: false,
+    links: { project: '1234' }
+  })
+}];
+
+const subjectIDs = ['1234'];
+
+const searchNode = {
+  getSelected() { return selectedCollection; }
+};
+
 describe('<CollectionsManager />', function() {
   let wrapper;
   let addToCollectionsSpy;
   before(function() {
-    addToCollectionsSpy = sinon.stub(CollectionsManager.prototype, 'addToCollections');
-    wrapper = shallow(<CollectionsManager project={project} />);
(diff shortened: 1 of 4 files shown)


eatyourgreens:
Looks good to me, but you should also write a test that catches the bug (and passes with this fix.)

wgranger:
Thanks for the suggestion. I used the approach in the email.spec to mock resources from the apiClient. I feel like I have a good handle on spies vs. stubs after the recent commit.

app/collections/collections-manager.spec.js
```diff
@@ -47,4 +64,9 @@ describe('<CollectionsManager />', function() {
     addButton.simulate('click');
     sinon.assert.calledOnce(addToCollectionsSpy);
   });
+
+  it('can add a subject to an empty collection', function() {
+    const addButton = wrapper.find('.search-button');
+    addButton.simulate('click');
```
mcbouslog: missing assert

eatyourgreens: I think this is missing `sinon.assert.calledTwice(addToCollectionsSpy);` (called twice because the previous test already called the spy once.)

eatyourgreens: Also, this test is just repeating the previous test without any changes.

wgranger:
My bad. I failed to adjust that test after figuring out spy/subs. I adjusted it again to detect a subject was added, but it is still somewhat similar to the test before (testing after a button click), so I'm happy to adjust it again if needed.

PR #4253, opened 2018-01-03
Fix warnings from workflow tests

eatyourgreens:
Mock Panoptes resources to stub out API method calls.
Stub API client requests during the tests.
Clean up the test code.

Staging branch URL:

Fixes various API warnings that are echoed to the console when the tests run.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/index.spec.js
+++ b/app/pages/project/index.spec.js
@@ -49,44 +49,99 @@ const resources = {
   ]
 };
 
-const project = apiClient.type('projects').create({
-  id: 'a',
-  display_name: 'A test project',
-  experimental_tools: [],
-  get(type) {
-    return Promise.resolve(resources[type]);
-  },
-  links: {
-    active_workflows: ['1', '2', '3', '4', '5'],
-    avatar: { id: '1' },
-    background: { id: '1' },
-    owner: { id: '1' }
+function mockPanoptesResource(type, options) {
+  const resource = apiClient.type(type).create(options);
+  if (!resource.save.restore) {
+    sinon.stub(resource, 'save', () => Promise.resolve(resource));
+    sinon.stub(resource, 'get', resourceType => Promise.resolve(resources[resourceType]));
+    sinon.stub(resource, 'delete');
   }
-});
+  return resource;
+}
 
-const background = apiClient.type('backgrounds').create({
-  id: project.links.background.id
-});
-
-const owner = apiClient.type('users').create({
-  id: project.links.owner.id
-});
-
-const projectAvatar = apiClient.type('avatars').create({
-  id: project.links.avatar.id
-});
-
-const preferences = apiClient.type('project_preferences').create({
-  preferences: {},
-  links: {
-    project: project.id
+const project = mockPanoptesResource('projects',
+  {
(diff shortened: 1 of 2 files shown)


app/pages/project/index.spec.js
```diff
@@ -49,45 +49,100 @@ const resources = {
...
-    return Promise.resolve(resources[type]);
-  },
-  links: {
-    active_workflows: ['1', '2', '3', '4', '5'],
-    avatar: { id: '1' },
-    background: { id: '1' },
-    owner: { id: '1' }
+function mockPanoptesResource(type, options) {
+  const resource = apiClient.type(type).create(options);
+  if (!resource.save.restore) {
```
camallen: Not a blocker at all but revealing the intention here to stub if not stubbed already could maybe be commented / extracted to function? 

I initially though it would call the save and the code might be unintentionally modified to do so in future, e.g `if (!resource.save().restore)`. That said it's probably just me being paranoid

app/pages/project/index.spec.js
```diff
@@ -49,45 +49,100 @@ const resources = {
...
+    sinon.stub(resource, 'save', () => Promise.resolve(resource));
+    sinon.stub(resource, 'get', resourceType => Promise.resolve(resources[resourceType]));
```
camallen: Not a blocker but `resources` here is a little confusing when reading for the first time. It looks like it's a prebuilt instance with known resource types and this get lookup will just fetch one of them, maybe rename to `factoryResources` or some such to know it's a known prebuilt resource lookup.

app/pages/project/index.spec.js
```diff
@@ -49,45 +49,100 @@ const resources = {
...
+      translations={translations}
+    />,
+    { context }
+  );
   const controller = wrapper.instance();
   const workflowSpy = sinon.spy(controller, 'getWorkflow');
 
+  before(() => {
+    sinon.stub(apiClient, 'request', (method, url, payload) => {
+      if (url === '/workflows') {
```
camallen: This branching when setting up the stub looks like the test context could be located closer to the relevant test instead of at the higher level. Having test setup context like this normally signifies this to me and I appreciate it might be hard to setup the context at a finer level.

app/pages/project/index.spec.js
eatyourgreens: A better approach would be to clean up the stubs in an `after` block. This check is really only here because the stubs in the email spec test interfere with the tests here, which isn't good.

app/pages/project/index.spec.js
eatyourgreens: It should probably be named `RESOURCES`, in keeping with constants in other files.

app/pages/project/index.spec.js
eatyourgreens: This was a workaround for workflow selection going into an infinite loop if you don't return anything from a network request. I think it affects all the tests, but it's not a very good fix.

eatyourgreens:
Thanks for the comments @camallen. I've made some changes which might improve these tests.

PR #4256, opened 2018-01-04
Refactor workflow selection

eatyourgreens:
Create a workflow selection component.
Update workflow selection tests.
Delete workflow selection code from the project page controller.

Staging branch URL: https://refactor-workflow-selection.pfe-preview.zooniverse.org/

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional


eatyourgreens:
Moves workflow selection out of the project page controller and into its own component, which expects to receive a project, preferences and user, and passes a workflow down to its children.

eatyourgreens:
I think this is ready for review now. Workflow selection tests are all passing, which made writing this a lot easier than it might have been otherwise.

Workflow selection should now be a black box that takes user, project and preferences and returns a workflow (or errors out.)

eatyourgreens:
That's weird. I'll have a look tomorrow but I didn't touch the package setup in this PR.

srallen:
Yep, I would think that would be caught in the diff if it was properly deleted, but it is not. I only caught it because I usually delete the node_modules folder and reinstall when doing reviews and the console messaged that a new package-lock.json was created.

srallen:
@eatyourgreens I just deleted my local copy of the branch and refetched it and it is there now, so it appears to have been a issue on my end only I think?

app/pages/project/workflow-selection.jsx
```diff
@@ -0,0 +1,220 @@
...
+    ) {
+      selectedWorkflowID = this.props.location.query.workflow;
+      activeFilter = false;
+      if (preferences && preferences.preferences.selected_workflow !== selectedWorkflowID) {
+        this.handlePreferencesChange('preferences.selected_workflow', selectedWorkflowID);
+      }
+    } else if (this.props.location.query &&
+      this.props.location.query.workflow &&
+      project.experimentalTools &&
+      project.experimentalTools.indexOf('allow workflow query') > -1
```
srallen: Lines 49 and 50 should be `project.experimental_tools && project.experimental_tools.indexOf('allow workflow query') > -1`. NfN workflow query loaded is broken otherwise. Tests should be updated to catch this failing case too.

eatyourgreens: Good catch. `project.experimental_tools` is defined in the tests too, I just never wrote a test that checked it.

eatyourgreens:
I've added a test for projects that use the workflow query param, fixed the component to pass the test, and restaged.

eatyourgreens:
That is weird. Could be a translations bug that only shows up here but not on master. I'll have a look at it tomorrow.

eatyourgreens:
I had a quick look on my phone. I can see the project descriptions, which might make this bug harder to track down.

eatyourgreens:
> This is weird, but the project descriptions are missing on this branch vs master on staging:

Confirmed in a desktop browser that props aren't being properly passed down to the project page component.

PR #4258, opened 2018-01-04
Rewrite ActiveUsers and Add Tests

wgranger:
Staging branch URL: https://active-users-es6.pfe-preview.zooniverse.org/

Describe your changes.
This PR converts the ActiveUsers component to ES6. A .spec file is also included for testing.

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional


app/talk/active-users.jsx
```diff
@@ -0,0 +1,197 @@
...
+    this.fetchUncachedUsers = this.fetchUncachedUsers.bind(this);
+    this.getActiveUserIds = this.getActiveUserIds.bind(this);
+    this.onPageChange = this.onPageChange.bind(this);
+    this.resetTimer = this.resetTimer.bind(this);
+    this.restartTimer = this.restartTimer.bind(this);
+    this.pageCount = this.pageCount.bind(this);
+    this.update = this.update.bind(this);
+    this.userIdsOnPage = this.userIdsOnPage.bind(this);
+    this.userLink = this.userLink.bind(this);
+  }
```
eatyourgreens: Only event handlers need to be explicitly bound to the `this` context. I think you can delete all of these bind calls, because the constructor already binds all your instance methods for you.

eatyourgreens: OK, you need to bind `this.onPageChange`, but should be fine to delete all the other bind calls.

eatyourgreens:
One minor comment about unnecessary calls to `bind` in the constructor. Other than that, this looks good to me.

app/talk/active-users.jsx
wgranger: Oh nice. I had a misunderstanding of how the function binding worked.

eatyourgreens: No problem. You have to bind handlers to the component instance because events set `this` to the element that fired the event, for historical reasons. MDN has a longer explanation. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this

PR #4261, opened 2018-01-05
Add task translations to drawing subtasks

eatyourgreens:
Staging branch URL: https://subtask-translations.pfe-preview.zooniverse.org/

Connects up drawing subtasks to the translations API, by wrapping the task component in a translations wrapper.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -1,6 +1,7 @@
 React = require 'react'
 StickyModalForm = require 'modal-form/sticky'
 ModalFocus = require('../../components/modal-focus').default
+TaskTranslations = require('../tasks/translations').default
 
 STROKE_WIDTH = 1.5
 SELECTED_STROKE_WIDTH = 2.5
@@ -14,6 +15,9 @@ SEMI_MODAL_UNDERLAY_STYLE =
 
 module.exports = React.createClass
   displayName: 'DrawingToolRoot'
+
+  contextTypes:
+    store: React.PropTypes.object
 
   statics:
     distance: (x1, y1, x2, y2) ->
@@ -76,7 +80,9 @@ module.exports = React.createClass
             {for detailTask, i in toolProps.details
               detailTask._key ?= Math.random()
               TaskComponent = tasks[detailTask.type]
-              <TaskComponent autoFocus={i is 0} key={detailTask._key} task={detailTask} translation={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />}
+              <TaskTranslations key={detailTask._key} taskKey={"TASK-KEY.tools.TOOL-INDEX.details.#{i}"} task={detailTask} store={@context.store}>
+                <TaskComponent autoFocus={i is 0} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />
+              </TaskTranslations>}
             <hr />
             <p style={textAlign: 'center'}>
               <button autoFocus={toolProps.details[0].type in ['single', 'multiple']} type="submit" className="standard-button" disabled={not detailsAreComplete}>OK</button>
--- a/app/classifier/drawing-tools/root.cjsx
+++ b/app/classifier/drawing-tools/root.cjsx
@@ -80,7 +80,13 @@ module.exports = React.createClass
             {for detailTask, i in toolProps.details
               detailTask._key ?= Math.random()
               TaskComponent = tasks[detailTask.type]
-              <TaskTranslations key={detailTask._key} taskKey={"TASK-KEY.tools.TOOL-INDEX.details.#{i}"} task={detailTask} store={@context.store}>
+              taskKey = "#{toolProps.taskKey}.tools.#{toolProps.mark.tool}.details.#{i}"
+              <TaskTranslations
+                key={detailTask._key}
+                taskKey={taskKey}
+                task={detailTask}
+                store={@context.store}
+              >
                 <TaskComponent autoFocus={i is 0} task={detailTask} annotation={toolProps.mark.details[i]} onChange={@handleDetailsChange.bind this, i} />
               </TaskTranslations>}
             <hr />
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -69,6 +69,7 @@ module.exports = React.createClass
                 preferences: @props.preferences
 
               toolProps =
+                taskKey: annotation.task
                 classification: @props.classification
                 mark: mark
                 details: details


eatyourgreens:
TODO: figure out how to set the translation keys properly for subtasks.

eatyourgreens:
This is ready for review now, but there are no translated workflows with drawing tasks and subtasks in Panoptes yet.

camallen:
@eatyourgreens the subtask translations strings will look like:
```
"tasks.T1.tools.3.label" => "Purple",
"tasks.T1.tools.3.details.0.answers.0.label" => "Painfully wow",
"tasks.T1.tools.3.details.0.answers.1.label" => "Just wow",
"tasks.T1.tools.3.details.0.question" => "Wow rating:",
```
I'm deploying some code that 'should just work' on staging and will translate the workflows for https://master.pfe-preview.zooniverse.org/projects/brooke/i-fancy-cats for a reference guide.

eatyourgreens:
@camallen Cheers! That's the pattern I've used for the translation key prefix too: https://github.com/zooniverse/Panoptes-Front-End/blob/904dd21403169c6d65970a5873c6c40ae58f4980/app/classifier/drawing-tools/root.cjsx#L83

eatyourgreens:
By the way, I haven't added translations for the drawing task labels in the front-end yet, only for the question tasks in the details pop-up.

app/classifier/drawing-tools/root.cjsx
```diff
@@ -15,6 +16,9 @@ SEMI_MODAL_UNDERLAY_STYLE =
 module.exports = React.createClass
   displayName: 'DrawingToolRoot'
 
+  contextTypes:
+    store: React.PropTypes.object
```
simoneduca: Do we have to use `context` here? Have you checked if `store` is available in the tree?

eatyourgreens: `react-redux` uses `context` to pass the store into connected components, but I'm happy to remove it here if you can suggest a better way to pass it into the modal form.

simoneduca: Doesn't the `connect` API for `react-redux`do that?

eatyourgreens: Yeah, a connected component gets the store from `this.context` inside `connect()`.

eatyourgreens: `TaskTranslations` is the connected component here, by the way.

simoneduca: If `TaskTranslations` is `connect`ed it has `store` already, you don't need to pass it to it. Sorry, I'm probably missing something.

eatyourgreens: Try removing `store={@context.store}` from `TaskTranslations` and see what happens.

eatyourgreens: My best guess here is that the `StickyModalForm` component doesn't pass `context` to its children.

eatyourgreens: If you remove context, you should see the following error in the console when you try a drawing task that pops up a modal form: `Uncaught Error: Could not find "store" in either the context or props of "Connect(TaskTranslations)". Either wrap the root component in a <Provider>, or explicitly pass "store" as a prop to "Connect(TaskTranslations)".`

simoneduca: It's probably me being slow, sorry, but if `TaskTranslations` is `connect`ed it has `store` already, you don't need to pass it to it.

simoneduca: screen shot 2018-01-11 at 10 25 02
The react dev tools show the same properties, whether `store` is passed or not. So I'm not sure what's doing there. As far as I can tell the functionality is not affected by removing `store` and, when I draw I don't get any error, just a `warning.js?8a56:36 Warning: Unknown prop `tool` on <g> tag. Remove this prop from the element`.

eatyourgreens: Are you trying a drawing task that has subtasks, like the cat face tool in I Fancy Cats or the circle tool in the dev classifier?

eatyourgreens: Your screenshot is debugging the drawing task, not the subtasks.

simoneduca: Ah I see what you mean now. This is really annoying: `connect` should do the job and we definitely shouldn't need `contextTypes`, but, as you said, something in between is not playing ball.

eatyourgreens: `Connect` gets the store (and subscription) from `context.store`, then looks for `props.store` if context isn't available.
https://github.com/reactjs/react-redux/blob/e170b2c54fbb796a32c087dcb9a1325abe5a370c/src/components/connectAdvanced.js#L81-L87

Dialogs like `StickyModalForm` don't pass context down to their children, which is why `Connect` throws an error here.

simoneduca: Do you mean all dialogs do that in general? Why? Can you point to a link please?

eatyourgreens: Yes, because they create elements outside of the current tree. Easiest way to see this is to open a tutorial on a zooniverse project, then inspect the popup with the React dev tools and look at the `Connect` component.

eatyourgreens: You should see that `context` is an empty object. If you inspect a `Connect` component elsewhere on a zooniverse project page, you'll see that `context.store` is defined.

srallen: I think he's talking about the dialogs from the modal-form (https://github.com/zooniverse/modal-form) library. We could open an issue in that repo to add context to what is passed down. It should be simple enough to add to the Base class.

eatyourgreens: React 16 introduces [Portals](https://reactjs.org/docs/portals.html) to work with children outside of the current parent node.

srallen: Oh yeah, I forgot about that. I'd really love it if we could get to React 16. One thing at a time though.

simoneduca: Gotcha, thanks. Here's the issue https://github.com/zooniverse/modal-form/issues/32

eatyourgreens: This affects all modals, not just the modal form library. For example, tutorials, mini-courses and field guides also have to be passed the store explicitly as a prop.

srallen: It depends on the design of the modal. I've seen some React modal libraries that are designed to be a child in the tree and use CSS to position it. If it's written to add the React element outside of the tree, then yes. That's how the modal-form is written as well as the dialog.cjsx version in PFE (which Brian used as a base for modal-form I think).

eatyourgreens: Yeah. Field guides are using `react-pullout` which also has the same problem. I thought tutorials were using PFE's built-in Dialog component but maybe not. I remember raising with Brian that modal-form was a bad fit here, because tutorials aren't forms and shouldn't use `<form>` elements. That was on the list of things to fix, but I don't remember if we ever got to it.

eatyourgreens: There's a [`PassContext`](/zooniverse/Panoptes-Front-End/blob/master/app/components/pass-context.cjsx) component which takes the PFE context and passes it to its children. Would using that be a better solution?

PR #4270, opened 2018-01-09
[WIP] Canvas-Viewer

tingard:
Staging branch URL: https://add-canvas-viewer.pfe-preview.zooniverse.org/

## Aim:
Add a canvas-viewer to PFE that allows data-based subject rendering in the browser.
To be used in https://www.zooniverse.org/projects/tingard/galaxy-builder 

This could also enable general data subjects to be plotted in the front-end using something like chart.js (i.e. spectra), and allow more complicated image manipulations in the browser (i.e. gravitational lens reconstruction in zooniverse, which is something a researcher here in Portsmouth is potentially interested in. 

### `components/canvas-viewer.jsx`
New viewer which takes in the subject, annotation and workflow objects and displays a canvas linked to workflow-specific rendering commands.

Some thoughts
- Subject-specific rendering commands rather than workflow
- Is there a better place to put this so that the file-viewer is independent of annotations? How would the result be displayed to the user (in a performant way, without manipulating resources from API or threading lots of unnecessary props through the component hierarchy)
- Could end up with a subject without an associated file, which Panoptes wouldn't be happy with?

### `components/modelling` restructuring
- Broken into multiple model objects which handle rendering. 'components/modelling/index.js' exports a function which accepts the workflow object and returns an appropriate model object.
- at the moment uses `workflow.id`, which is not correct IMO.
- Need to move `baseModel.jsx` to a pure javascript file, and specify it means for [regl](https://github.com/regl-project/regl)-based rendering (`baseReglModel.js`?)

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/components/file-viewer/canvas-viewer.jsx
```diff
@@ -0,0 +1,111 @@
...
+    this.createNewModel(this.props);
+  }
+  shouldComponentUpdate(nextProps, nextState) {
+    // check if a new subject has been provided
+    if (this.state.loading !== nextState.loading || this.props.src !== nextProps.src) {
+      return true;
+    }
+    // Only re-render when annotation has changed, or we have zoomed/panned
+    // JSON is expensive, so reduce the test as much as possible
+    if (JSON.stringify(nextProps.annotations) !== JSON.stringify(this.props.annotations)) {
```
tingard: Easiest way to do a deep comparison between objects

app/components/file-viewer/index.jsx
```diff
@@ -20,39 +21,52 @@ const VIEWERS = {
...
+    frame: props.frame,
+    onLoad: props.onLoad,
+    onFocus: props.onFocus,
+    onBlur: props.onBlur
+  };
+  if (props.type === 'canvas') {
+    Object.assign(
+      viewerProps,
+      {
+        annotations: JSON.parse(JSON.stringify(props.classification.annotations)),
```
tingard: Effectively a deep copy. Without this `this.props.annotations` will always equal `nextProps.annotations` inside `shouldComponentUpdate`

tingard:
I'm also noticing a ~2px offset between annotation and render result in Safari compared to Chrome, and an even larger one in firefox, which I can't explain.

eatyourgreens:
Some initial thoughts:

The file viewer components are designed to render data passed in from subject locations. They probably shouldn't be hacked to respond to changes in annotations.

The JSON parse/stringify hacks to compare arrays need to go.

Have you written any tests for this? Would writing tests help to spec out the behaviour? I'm having trouble understanding whether or not modelling is a task in its own right. If it is a new type of task, then your canvas viewer could be a new type of annotation render component in the classifier, which renders annotations for modelling tasks.

The changes to the SVG component, which renders drawing annotations, make me uneasy, because it seems like there's a strong coupling between the canvas viewer and drawing tasks.

eatyourgreens:
Inviting @srallen to comment here because these changes contradict our idea that the subject media viewers should be read-only components, responsible only for rendering the contents of URLs.

app/components/file-viewer/canvas-viewer.jsx
eatyourgreens: Noting that `annotations` is an array, not an object, but I'm not sure if that makes a difference here.

app/classifier/annotation-renderer/svg.jsx
```diff
@@ -153,7 +153,17 @@ export default class SVGRenderer extends React.Component {
...
+      this.props.children,
+      child => React.cloneElement(
+        child,
+        type === 'canvas' ? {
+          viewBoxDimensions: this.props.viewBoxDimensions,
+          classification: this.props.classification,
+          subject: this.props.subject,
+          workflow: this.props.workflow
+        } : {})
+    );
```
eatyourgreens: This block introduces a strong coupling between drawing tasks and the canvas viewer.

app/components/file-viewer/canvas-viewer.jsx
```diff
@@ -0,0 +1,111 @@
...
+      this.model.update(nextProps.annotations, nextProps.viewBoxDimensions);
```
eatyourgreens: `shouldComponentUpdate` should return true/false to indicate whether a new render is needed. It shouldn't be calling object instance methods like this.

app/components/file-viewer/index.jsx
eatyourgreens: `annotations: props.classification.annotations.map(annotation => annotation)` maybe?

I'm not entirely sure why you are making a copy of the whole annotations array here.

eatyourgreens:
I've staged this at https://add-canvas-viewer.pfe-preview.zooniverse.org/

eatyourgreens:
> Add a canvas-viewer to PFE that allows data-based subject rendering in the browser.

Regarding this, I'd expect to be able to upload a subject to Panoptes with a data file in the locations array, something like `"text/data": data_file_url`, and expect the new viewer to render the contents as a plot of the data.

app/components/file-viewer/index.jsx
tingard: I want to only trigger a re-render when the annotation changes, so think I need to properly deep copy the prop being passed in order to check for changes.
Unfortunately I don't think that'll work as annotations is an array of objects, so the newly constructed array would still refer to the old objects.
i.e.
```javascript
original = [{ a : 1}];
copySlice = original.slice();
copyMap = original.map(i => i);
copyJson = JSON.parse(JSON.stringify(original));
original[0].a = 2;
console.log(original,  copySlice, copyMap, copyJson);
```

tingard: Sorry, bad phrasing. By re-render I mean update the model, rather than the component.

eatyourgreens: In that case, you should be looking at the annotation for the current task, not all the annotations for the current workflow. That might be passed down as `props.annotation`. If not, it should be the last annotation in the annotations array.

app/components/file-viewer/canvas-viewer.jsx
eatyourgreens: I think the model update should be called from `componentWillUpdate` or `componentDidUpdate`, leaving `shouldComponentUpdate` to do what it says on the tin, and decide whether or not those methods should be called.

tingard: The problem is that the model should only update in the case that `shouldComponentUpdate` would return false, and I'm not sure where to do that other than by spawning a Promise.

I've just read in [the react docs](https://reactjs.org/docs/react-component.html#shouldcomponentupdate)

> *"We do not recommend doing deep equality checks or using JSON.stringify() in shouldComponentUpdate(). It is very inefficient and will harm performance."*

So this has to be re-written either way.

tingard:
I was envisioning the JSON for that kind of subject would be along the lines of 

`{'canvas/data-csv': "data_file_url.csv", 'model': "CHART_PLOTTER" }`
or 
`{'canvas/plot-data-csv': "data_file_url.csv" }`.

Where `plot-data` would specify to load in the correct (i.e. chart.js rendering) Model.

eatyourgreens:
@tingard `canvas/data-csv` isn't a valid MIME type, so wouldn't be allowed by Rails for a Panoptes subject. If people want to upload CSV files as subjects, the MIME type is `text/csv`.

app/components/file-viewer/canvas-viewer.jsx
eatyourgreens: I'd advise against updating the model when `shouldComponentUpdate` returns false. That's going to confuse anyone who has to debug model updates, and expects updates to run when `shouldComponentUpdate` returns true.

app/components/file-viewer/canvas-viewer.jsx
tingard: is this a preferable way of doing a deep equality check?
`const deepEquals = (a, b) => (Object.keys(a).length ? Object.keys(a).every(k => deepEquals(a[k], b[k])) : a === b);`
doesn't catch circular references but you could wrap it
```javascript
const deepEquals = (a, b) => {
  const foo = (a, b) => (Object.keys(a).length ? Object.keys(a).every(k => foo(a[k], b[k])) : a === b);
  try {
    return foo(a, b);
  } catch (RangeError) {
    throw('TypeError: Cannot check deep equality of circular structure');
    return false;
  }
}
function Foo() {
   this.circular = this;
   this.abc = "Hello";
}
const foo = new Foo();
deepEquals(foo, foo);
```

eatyourgreens: Are you watching to see if a particular key in the current annotation changes? If so, why not just compare that key on old and new props in `componentWillReceiveProps` or `componentWillUpdate`?

srallen: lodash's `_.isMatch()` might be what you're looking for and we already have lodash installed as a dependency: https://lodash.com/docs/4.17.4#isMatch

eatyourgreens: I'd just compare the old and new annotation, to be honest: https://github.com/zooniverse/Panoptes-Front-End/blob/60a5ff4db7e344b0365da2d288b432b97b44ead4/app/classifier/frame-annotator.jsx#L23

eatyourgreens: Comparing the entire annotations array seems wasteful when the only item that could have changed is the current annotation.

tingard: I'm still getting the problem with object references not changing so would need to do a deep copy somewhere.
The frame annotator check doesn't fire either when a drawing tool is manipulated (`{value: [{...}]}`), only for tasks with basic values (`{ value: 0 }`).

eatyourgreens: Looking more closely, I don't think that line in the frame annotator is supposed to fire when the annotation changes. I think it's there to detect when the task changes. That's confusing.

eatyourgreens: I'm not sure if this is relevant here but #4272 fixes some bugs where annotation changes weren't correctly updating classifier state.

tingard:
@eatyourgreens ah okay, would something like this be okay?
```json
{
  "locations": [
    {"application/json": "testData.json"}, 
    {"text/csv": "testData.csv"}
  ],
  "metadata": {
    "isModelling": true,
    "modelling": [
      { "frame": 0, "model": "LINE_PLOT" },
      { "frame": 1, "model": "SCATTER_PLOT" }
    ]
  }
}
```
As then you could specify a model for each frame. Canvas-viewer could catch all 'application' or 'text' subjects, though I'm not sure that's desirable?

app/classifier/annotation-renderer/svg.jsx
tingard: moved the injection of these props to the frame-viewer, with viewBoxDimensions being added in the frame-annotator and a default case added in canvas-viewer

tingard:
I'm forgetting that there are also scenarios where you want to perform manipulations on an image, for example galaxy-builder. I suppose in those scenarios you could upload the image as some kind of Uint8 array. I suppose canvas-viewer could even handle fits files [like this](http://www.clearskyinstitute.com/fits/), though the file sizes might make that a bad idea.

app/components/file-viewer/index.jsx
```diff
@@ -20,39 +21,52 @@ const VIEWERS = {
...
-      onFocus={props.onFocus}
-      onBlur={props.onBlur}
-    />
+    <Viewer {...viewerProps} />
   );
 }
 
 FileViewer.propTypes = {
   className: React.PropTypes.string,
+  classification: React.PropTypes.object,
```
eatyourgreens: Remember that the file viewer will also be used to view subjects on Talk, where classifications and workflows aren't available.

I'd advise against making any changes to this component that tie it to the classifier.

tingard: I'm not sure where to send the annotation/zoom information to in a way it can be used then. Do you have any suggestions?

eatyourgreens: I thought you'd decided to pass that info down via the frame annotator components?

tingard: It's going to the file-viewer from frame-annotator and frame-viewer, then through to canvas-viewer. Could add a default value for classification, annotation and subject (no longer passing workflow) in file-viewer?

eatyourgreens: Can you remind me why the classification and subject are being passed down? At this level in the component tree, you're dealing with a single frame from a subject, so I'm not really getting why the subject still needs to be available.

If you already have the annotation as a prop, what are you using the classification for?

Sorry for the questions, I'm just very confused by the intent of this component at the moment.

eatyourgreens: Actually, would it make more sense to generate the model somewhere around here, then pass it down as a prop instead of the subject?

tingard: Primary motivation behind this component is to allow in-browser modelling, as is currently done in a hacky way by https://zooniverse.org/projects/tingard/galaxy-builder.

Entire classification is being passed down as some projects like galaxy-builder rely on the whole annotations list to create a model, as each task is for a different part of the model (disk/bar/spiral). Tried doing this in one huge combo task and UX was horrible.

Annotation prop is being used to check whether the most recent annotation has changed, though I could just as easily use `classification.annotations[classification.annotations.length -1]`

Subject is needed as there are many possible tasks that canvas-viewer could do under the umbrella of modelling, and I need a way to identify which model *this* frame of *this* subject should be using, hence storing it in metadata.

tingard: In general I'd need the canvas context (2d or webgl) to render, and while assembling a function which could be called with the context would be okay for 2d contexts, it would be super expensive to repeatedly create and bind webgl rendering commands. I can give it a go but doubt it'd work well.
Is there any movement behind storing annotations in redux? Then I wouldn't need to thread anything through other components and could just grab from the store using a wrapper. Doesn't address the problem of canvas viewer being classification independent though...

eatyourgreens: Have a look at #4277 and #4272 for where we currently ate with annotations. We want to use a store for annotations in the new classifier, which we're about to start building, but the current proposal is that the subject viewer in the new classifier will be read-only ie. it will be designed to render subject frames but won't have any knowledge of annotations.

eatyourgreens:
Storing an object in the subject metadata is a neat idea, but you should check whether it's possible in practice, since the metadata is usually generated from text in CSV manifest files.

app/components/file-viewer/index.jsx
eatyourgreens: > Entire classification is being passed down as some projects like galaxy-builder rely on the whole annotations list to create a model, as each task is for a different part of the model (disk/bar/spiral). Tried doing this in one huge combo task and UX was horrible.

The model only needs to be updated when the current annotation changes, though? The values for previous tasks will be fixed so recalculating the mode for those seems wasteful. Is it possible to store the model and update with annotation changes, rather than creating it from scratch each time? Then pass the model down to a canvas that renders it.

tingard: That does make sense. I suppose the problem is that what I'm proposing results in the classification becoming part of the subject, which goes against that ethos.

I suppose that would be possible, but then you'd still be passing through what is effectively a compiled version of the annotation, breaking the whole annotation-free file-viewer plan.

How about somehow separating into a "dynamic-file-viewer" and a static one?

eatyourgreens: Sounds good. Your idea of a compiled version of the annotations sounds good to me. Would the annotations be compiled into something that could be stored in Redux, then read and rendered by the viewer?

tingard: I could probably organise things into a basic redux compatible object, would need some fiddling to check.
As far as I can tell you'd need to send all annotations through as canvas-viewer will unmount and re-mount when frames are switched, so if the first task is done on an image subject and then the next requires the canvas, canvas-viewer wouldn't know the first annotation to render, which it might need in the general case.
Did you mean the idea of a dynamic-file-viewer sounds good?

eatyourgreens: I was thinking the viewer should be a purely functional component that renders its props, and all the heavy calculation work should be done higher in the tree, stored in state and passed down as props to the viewer. That avoids you redoing all your work when the frame changes, instead reading your data from a prop or, eventually, a data store.

eatyourgreens: Also, avoid coding for the general case unless you actually need it. There's a few bits of code in the classifier that try to accommodate cases which, it turns out, aren't actually needed in real projects.

tingard:
Just a thought, but this feels like a good use case for [react portals](https://reactjs.org/docs/portals.html)?

Would create the dynamic-canvas-renderer as a child of Classifier and it would portal itself to replace the canvas with selector `.canvas-viewer-wrapper>canvas[data-frame="1"]`, or something along those lines?

eatyourgreens:
This could be a good case for portals, but legacy code in PFE uses `React.createClass`, which is holding us back from moving to React 16.

tingard:
I could open a PR to replace `createClass` with the `create-react-class` module, and even do the same for `PropTypes` at the same time? I suppose the time-sensitivity of my project means that may be a bit blue-sky.

eatyourgreens:
That's a nice idea, but if you do a search through this repo, there's a _lot_ of legacy coffeescript that needs replacing, so I'm not sure how practical it is.

tingard:
Ah is it not possible to have a halfway house using 16 where there's coffeescript but it uses `create-react-class`?

tingard:
>Storing an object in the subject metadata is a neat idea, but you should check whether it's possible in practice, since the metadata is usually generated from text in CSV manifest files.

It's possible through the python client - I've uploaded a convolution kernel array as metadata for galaxy builder, not sure about using a csv though...

eatyourgreens:
We should rebase this before it gets too far out of sync with master.

eatyourgreens:
Did you see that file get copied over and published to S3 when the webpack build ran?

eatyourgreens:
Looking at the server config, I don't think `.json` is recognised as an extension for subject URLs.
https://github.com/zooniverse/static/blob/959863c4af1b83f4993a9b75c677ce1585083c60/sites/www.zooniverse.org.conf#L63-L74

tingard:
Coleman can see that it was moved to s3, interesting that subjects uploaded normally do still load fine, but not the static assets

eatyourgreens:
nginx resolves all URLs to `/index.html`, to load the React app, unless they meet certain exception rules and I don't think that includes URLs that end in `.json`. You could try using the fully qualified URL of the file on static.zooniverse.org, I guess.

CKrawczyk:
@tingard I was looking into the image "pip"s not working on the dev classifier and tracked it down to [the format not being detected](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/lib/get-subject-location.coffee) when [the subject view tries to get the subject type](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/subject-viewer.cjsx#L114) and [use that as a `key`](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/subject-viewer.cjsx#L178).

CKrawczyk:
There is also and error `Encountered two children with the same key`.  The two `canvas` frames use the same file, so the `pip`s end up with the same key.

eatyourgreens:
There's an unstated assumption in PFE that each subject location will be a unique URL. Is there a case where two locations might point to the same file?

eatyourgreens:
Working on feedback recently, I discovered that there's [a features directory](zooniverse/Panoptes-Front-End/tree/master/app/features) in PFE now. That looks like a good place to put the `modelling` directory too.

srallen:
I changed the flipbook subject image pips from using the array index to use the URL in zooniverse/Panoptes-Front-End#4342 which the assumption that the file names would be unique. The users on Talk reported that making sure the key is unique fixed the bug. If there is something else that can be used as a unique identifier, then feel free to change it.

CKrawczyk:
Currently the flipbook toggle button only shows up for `image` data types: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/subject-viewer.cjsx#L124

tingard:
changed to `when type is 'image' || type is 'application'`

Does that statement need to be a which? There's only one case I can see.

CKrawczyk:
It is in a switch statement, the coffee script for that line would be `when 'image', 'application'`

app/components/pan-zoom.jsx
```diff
@@ -148,6 +148,7 @@ class PanZoom extends React.Component {
   }
 
   togglePanOn() {
+    console.log(this.state.panEnabled);
```
tingard: this was committed accidentally

app/components/svg-transparent-rect.jsx
```diff
@@ -0,0 +1,75 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const FILTERS = {
+  invert: "url('#svg-invert-filter')"
+};
+
+const INVERT =
```
CKrawczyk: If this is a transparent SVG does it need an invert option?

tingard: Nope, well spotted

tingard:
> There's an unstated assumption in PFE that each subject location will be a unique URL. Is there a case where two locations might point to the same file?
For galaxy-builder the subjects could be re-used to save a small amount of bandwidth (~100KB), but not worth the hassle imo

tingard:
Ah whoops, so did I (same location different frame and different model). The galaxy-builder model json  (frame 2) is a stripped-back version of the full difference calculation json (frame 1)

tingard:
Still do to:
- fix server config to allow json static resources
- Write some tests
- fix canvas sizing (to be set using subject json but using a function in the react component passed to the Model)

tingard:
Sort of waiting on #4405 and #4407 for galaxy-builder project.
Still need to fix server config as above and perform cross-platform compatibility checks.

tingard:
PR opened on zooniverse static https://github.com/zooniverse/static/pull/28 to allow JSON subjects in dev/classifier

tingard:
Will rebase soon and do some final testing. There is a bug when you switch frames while zoomed-in where things do not appear the correct size, and rotation is not supported in galaxy-builder (though I don't think I can flag this in the project anywhere?)

tingard:
@eatyourgreens could you possibly re-stage this today? (CKrawczyk is in Japan)

I think this is ready for another review / testing, @eatyourgreens @srallen feedback is very welcome!

eatyourgreens:
No problem. It's staged at https://add-canvas-viewer.pfe-preview.zooniverse.org/

tingard:
Hmm, weird sizing bug strikes again! https://add-canvas-viewer.pfe-preview.zooniverse.org/projects/tingard/galaxy-builder/classify

tingard:
I also can't unlink certain subject sets, no errors in console provided
https://i.imgur.com/YDFRV7B.png

tingard:
Dev classifier test data still not working, will simply remove for the time being (nice proof of concept for plotting but nvm)

app/features/modelling/dev-classifier/index.js
```diff
@@ -0,0 +1,43 @@
+class DevClassifierModel {
+  constructor(canvas, { frame }) {
+    this.ctx = canvas.getContext('2d');
+    this.frame = frame;
+  }
+  update(annotations, vBD) {
+    this.ctx.clearRect(0, 0, this.ctx.canvas.width, this.ctx.canvas.width);
+    for (let i = 0; i < annotations.length; i += 1) {
```
srallen: Can `forEach` be used here instead of the for loops? 

Here's a good comparison on why you'd usually want to use `forEach` over a for loop: https://thejsguy.com/2016/07/30/javascript-for-loop-vs-array-foreach.html

app/features/modelling/galaxy-builder/index.js
```diff
@@ -0,0 +1,197 @@
...
+        default:
+          break;
+      }
+      if (comp !== null) {
+        comp[0](Object.assign({ texture: this.state.pixels }, comp[1]));
+        this.state.pixels({ copy: true });
+        ret.push(comp);
+      }
+    }
+    for (let i = 0, comp = null; i < annotations.length; i++) {
```
srallen: How about `forEach` here?

app/features/modelling/galaxy-builder/index.js
```diff
@@ -0,0 +1,197 @@
...
+    //       re-calculated
+    this.state.annotations = annotations;
+    const s = {
+      size: [this.state.sizing.width, this.state.sizing.height],
+      // the canvas is not the same size as is visible
+      sizeMultiplier: this.canvas.width / this.state.sizing.width,
+      model: this.model
+    };
+    const ret = [];
+    for (let i = 0, comp = null; i < annotations.length; i++) {
```
srallen: Maybe `Array.map()` here? you could set the variable `ret` to equal `annotations.map(callback)`

app/features/modelling/baseReglModel.jsx
```diff
@@ -0,0 +1,85 @@
...
+    this.setModel();
+  }
+  update(annotations, newViewBox) {
+    // this function parses the new annotation and then triggers a render.
+    // update the render funtions
+    this.regl.poll();
+    this.regl.clear({
+      color: [0, 0, 0, 1]
+    });
+    this.state.pixels({
```
srallen: If I'm reading this correctly, this appears to be a component state property that is a function? Does the returned result of the function become the state value? This is confusing to me and I'm wondering if this is directly mutating state, which is an anti-pattern in React.

app/features/modelling/galaxy-builder/index.js
```diff
@@ -0,0 +1,197 @@
...
+          break;
+        case 'bar':
+          comp = parseBar(annotations[i], s);
+          break;
+        default:
+          break;
+      }
+      if (comp !== null) {
+        comp[0](Object.assign({ texture: this.state.pixels }, comp[1]));
+        this.state.pixels({ copy: true });
```
srallen: Same question here about this potentially directly mutating state.

app/features/modelling/galaxy-builder/parseFunctions.js
```diff
@@ -0,0 +1,102 @@
...
+
+export const parseSpiral = (comp, state, currentComps) => {
+  // has a spiral been drawn?
+  if (!hasComp(comp)) return null;
+  // do we have a disk?
+  const disks = currentComps.filter(i => i[1].name === 'disk');
+  if (disks.length === 0) return null;
+  else {
+    const ret = [];
+    for (let i = 0; i < comp.value[0].value.length; i += 1) {
```
srallen: Would `Array.map()` work better here?

app/features/modelling/galaxy-builder/parseFunctions.spec.js
```diff
@@ -0,0 +1,147 @@
...
+    spiral: {
+      name: 'spiral',
+      func: () => null,
+      default: { mux: 100, muy: 100, rx: 5, ry: 5, spread: 1, roll: 0, i0: 0.75, n: 2, c: 2, falloff: 1 }
+    }
+  }
+};
+
+const annotationWithoutShape = cloneDeep(annotations);
+for (let i = 0; i < annotationWithoutShape.length; i += 1) {
```
srallen: Should this be in a before statement in the describe block? Also, maybe use `forEach` instead?

app/features/modelling/galaxy-builder/parseFunctions.spec.js
```diff
@@ -0,0 +1,147 @@
...
+    assert.equal(hasComp(annotations[0]), true);
+  });
+});
+
+describe('parseDisk', function () {
+  it('should return null for an annotation without a drawn shape', function () {
+    assert.equal(parseDisk(annotationWithoutShape[0], state), null);
+  });
+  it('should return component parameters for an annotation with a drawn shape', function () {
+    assert.ok(parseDisk(annotations[0], state));
```
srallen: I see sinon imported, but not being used. Perhaps you can wrap the tested functions in this test suite with spies so you can test more than just `assert.ok()` for instance if the spy has the expected return value, etc.

app/features/modelling/galaxy-builder/postProcessingRegl.js
```diff
@@ -0,0 +1,194 @@
...
+  return asinh(px / ${a}) / asinh(${a});
+}
+`;
+
+export const convolvePSF = (r, psf, size) => {
+  const psfKeys = {};
+  let convolution = '';
+  let c = [0, 0];
+  for (let i = 0; i < size[0]; i += 1) {
+    for (let j = 0; j < size[1]; j += 1) {
```
srallen: Maybe `forEach`?

app/features/modelling/line-plot/index.js
```diff
@@ -0,0 +1,34 @@
...
+            data,
+            options: {}
+          }
+        );
+      })
+      .catch((e) => {
+        console.warn(e);
+      });
+  }
+  update() {
```
srallen: This method just returns null... is this class a work in progress and is going to be built out more later? If so, could there be a TODO comment left? If not, then does this need to be a class? Perhaps what's happening in the constructor could just be an exported function?

tingard:
@srallen thanks for all the comments, I'll go through them today and it'd be fantastic if we could merge this afterwards

app/components/file-viewer/canvas-viewer.jsx
```diff
@@ -0,0 +1,191 @@
...
+    this.setState({
+      canvasStyle: {
+        width,
+        height
+      }
+    });
+  }
+  createNewModel(props) {
+    if (!this.state.loading) this.setState({ loading: true });
+    return new Promise((resolve, reject) => {
```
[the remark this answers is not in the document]
tingard: I'm not sure I understand what you mean here? 
The promise updates what it needs on `this`, so I don't think so. It's returned from the class method so I can chain `model.update` where needed

app/components/svg-transparent-rect.jsx
```diff
@@ -0,0 +1,42 @@
...
+    super();
+    this.fixWeirdSize = this.fixWeirdSize.bind(this);
+  }
+
+  componentDidMount() {
+    this.fixWeirdSize();
+  }
+
+  componentDidUpdate() {
+    this.fixWeirdSize();
```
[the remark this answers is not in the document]
tingard: I have no idea, it's copy-paste from `app/components/svg-image.jsx`

app/features/modelling/dev-classifier/index.js
tingard: This file is no longer used, so I'll remove it. I used the for loop for performance, but annotations is so short it probably wouldn't matter.
Article was a good read, thanks!

app/features/modelling/baseReglModel.jsx
```diff
@@ -0,0 +1,85 @@
...
+    }
+  }
+  getTexture() {
+    return {
+      data: this.regl.read(),
+      width: this.canvas.width,
+      height: this.canvas.height
+    };
+  }
+  // -------------- Functions to be overwritten by child classes --------------
```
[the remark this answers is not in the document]
tingard: Interesting! Would this still work considering the Models aren't react components but Javascript classes?

app/features/modelling/baseReglModel.jsx
tingard: pixels is a reference to a WebGL texture
As this isn't a React component I've only used a variable called "state" to group things which are related to what the model should display (I think).

app/features/modelling/galaxy-builder/parseFunctions.spec.js
tingard: This was the original plan, however time constraints meant I had to shift focus. I have it on a trello board to come back to 😅

app/features/modelling/galaxy-builder/postProcessingRegl.js
tingard: I'd like to keep this as is, I think it's more performant and I'm not sure nested forEach's would be much more readable?
```javascript
psf.forEach(
  (row, i) => row.forEach(
    (value, j) => {
      psfKeys[`psf[${(i * size[0]) + j}]`] = value;
      c = [i - 5, j - 5];
      convolution += `texture2D(texture, uv + vec2(${(c[0]).toFixed(1)}/texSize[0],`;
      convolution += `${(c[1]).toFixed(1)}/texSize[1])) * psf[${(i * size[0]) + j}]`;
      if (i < size[0] - 1 || j < size[1] - 1) {
        convolution += ' + \n';
      }
    }
  )
)
```

tingard:
In hindsight this should have been multiple commits...?

eatyourgreens:
I think this PR will be easier to squash, rather than merge multiple commits to master.

app/components/file-viewer/index.jsx
```diff
@@ -21,39 +23,53 @@ const VIEWERS = {
...
+    onLoad: props.onLoad,
+    onFocus: props.onFocus,
+    onBlur: props.onBlur
+  };
+  if (props.type === 'application') {
+    Object.assign(
+      viewerProps,
+      {
+        annotation: cloneDeep(props.annotation),
+        annotations: props.classification.annotations,
```
eatyourgreens: Not a blocking comment for this PR, but requiring the classification here will break the subject viewer in Talk for any future projects that use data subjects.

app/components/frame-viewer.jsx
```diff
@@ -45,19 +45,30 @@ export default class FrameViewer extends React.Component {
...
     }
-    const zoomEnabled = this.props.workflow && this.props.workflow.configuration.pan_and_zoom && (type === 'image' || this.props.isAudioPlusImage);
+    const zoomEnabled = (
+      this.props.workflow &&
+      this.props.workflow.configuration.pan_and_zoom &&
+      (type === 'image' || type === 'application' || this.props.isAudioPlusImage)
+    );
+    const modellingProps = type === 'application' ? { // could be a more specific type check here
+      annotation: this.props.annotation,
+      classification: this.props.classification,
```
eatyourgreens: `classification` being used by the subject viewer here too.

app/components/file-viewer/index.jsx
tingard: Should I use a default blank annotation, or wait for the Portal approach once React 16 is through?

tingard: Or do you mean regarding #4484?

app/components/svg-transparent-rect.jsx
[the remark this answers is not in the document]
eatyourgreens: If this is a copy of `SVGImage`, you can replace it with `<SVGImage fill="#000000" fillOpacity="0.0" />` in the parent component.

app/components/file-viewer/index.jsx
eatyourgreens: Sorry, I mean that Talk doesn't have workflows or classifications. Neither do collections and favourites, which also display subjects. The subject viewer can't display classifications. Only the components in `/app/classifier/annotation-renderer` can render annotations.

`classification` and `annotation` should be optional props, with sensible defaults. With #4484 `classification.annotations` will always be an empty array while you're classifying.

app/components/svg-transparent-rect.jsx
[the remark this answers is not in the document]
eatyourgreens: I'm confused that this component uses `SVGImage` code but is actually rendering an SVG rectangle.

app/components/file-viewer/index.jsx
tingard: okay, so I'll provide those defaults in `file-viewer/index.jsx` for the time being, and then clean the tree of references to workflow/classification etc... once I can move canvas-viewer to a portal?

eatyourgreens: I'm not sure that portals would fix this, but React 16 is still a way off anyway. Just be aware that file viewer components will be rendered outside the classifier so any classifier props that are passed down must be optional.

app/features/modelling/galaxy-builder/postProcessingRegl.js
srallen: For loops are faster, though, typically the difference (in milliseconds) won't be noticed by the end user and performance is still variable with the browser. If it can be demonstrated that there is a noticeable impact, using say dev tool's performance analysis tools, then we can make a decisions based on performance. 

Generally the newer array enumerables are preferred since that's the direction javascript is moving and I do think readability, although subjective, is an important consideration for the team. I find seeing named variables like `row` and `value` more immediately readable at a glance than the for loop.

tingard: Good point, premature optimisation is the root of all evil and all that. It also only runs once in a classification rather than once a frame, which I'd forgotten! I feel like this should be commented too as it's about as readable as Daily Mail written backwards in webdings... I'll add this to the list of things to put in my follow-up PR

tingard: Agreed re. readability being paramount. The for loop obsession comes from when I was doing rendering on the CPU about a year ago and was desperately trying to scrape back every possible millisecond.

app/features/modelling/line-plot/index.js
tingard: the `update()`  function is required for a canvas-builder model, and since this doesn't subclass anything I needed to add in a placeholder of some kind. Is there a best practise for this?

PR #4191, opened 2017-11-29
Feedback Rewrite

rogerhutchings:
cc @chrislintott @trouille 

Staging branch URL: https://feedback-update.pfe-preview.zooniverse.org

Test project URL: https://feedback-update.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test/classify

- Rewrites the feedback code to make it more organised and more understandable.
- Adds in comments and documentation.
- Supports radial (tolerance) and dud feedback.
- Supports text feedback on success, failure or both.
- Removes single question feedback (will be added back in future).
- Moves feedback from the classifier to a modal, which opens on task change.
- Saves feedback to the classification per-task

## To do

- Tests
- Add in a new strategy for handling ellipse overlap
- Clone the subject viewer into the modal and allow rendering of point feedback to show where known objects are
- Make the modal prettier

# Required Manual Testing

- [x] Does the non-logged in home page render correctly?
- [x] Does the logged in home page render correctly?
- [x] Does the projects page render correctly?
- [x] Can you load project home pages?
- [x] Can you load the classification page?
- [x] Can you submit a classification?
- [x] Does talk load correctly?
- [x] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4272, opened 2018-01-09
Remove feedback hack from classifier

### LINKED ISSUE #4243: cannot load classifier in different workflow after feedback workflow
## Expected behavior
After making classification in feedback enabled workflow, can navigate to a different workflow and make a classification.

## Current behavior
Made classidication in this feedback enabled workflow - https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/feedback-testing-mb/classify?reload=0&workflow=3135,

then navigated to a different project/workflow and classifier wouldn't load.

Uncaught (in promise) TypeError: Cannot read property 'task' of undefined from feedback.js?3746:26, which is [this line in ducks/feedback.js](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/redux/ducks/feedback.js#L26).

mentioning @rogerhutchings

eatyourgreens:
Staging branch URL: https://remove-feedback-hack.pfe-preview.zooniverse.org/

- Fix a bug in `updateAnnotations` by properly copying the annotations array to component state.
- Remove the taskID hack (see #4202) from `componentDidUpdate` and use state changes to detect when the task changes instead.
- Remove _.last() in favour of using `currentAnnotation` for consistency. We shouldn't be using a library for something as simple as getting the last member of an array.
- Fixes #4243 by setting a default for `currentAnnotation` in `updateFeedback`.

I've tested a couple of tasks in the dev classifier, and annotations seem to be working as expected, but this really does need an extra pair of eyes to check that I haven't broken classification annotations.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -4,7 +4,6 @@ import { VisibilitySplit } from 'seven-ten';
 import Translate from 'react-translate-component';
 import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
-import _ from 'lodash';
 
 import SubjectViewer from '../components/subject-viewer';
 import ClassificationSummary from './classification-summary';
@@ -125,7 +124,7 @@ class Classifier extends React.Component {
   }
 
   checkForFeedback(taskId) {
-    const taskFeedback = _.get(this.props.feedback, `rules.${taskId}`, []);
+    const taskFeedback = this.props.feedback[`rules.${taskId}`] || [];
 
     if (!this.props.feedback.active || !taskFeedback.length) {
       return Promise.resolve(false);
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -124,9 +124,10 @@ class Classifier extends React.Component {
   }
 
   checkForFeedback(taskId) {
-    const taskFeedback = this.props.feedback[`rules.${taskId}`] || [];
+    const { feedback } = this.props;
+    const taskFeedback = (feedback.rules && feedback.rules[taskId]) ? feedback.rules[taskId] : [];
 
-    if (!this.props.feedback.active || !taskFeedback.length) {
+    if (!feedback.active || !taskFeedback.length) {
       return Promise.resolve(false);
     }
 
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -92,7 +92,7 @@ class Classifier extends React.Component {
       this.updateAnnotations();
     }
 
-    if (currentAnnotation.task !== prevAnnotation.task) {
+    if (prevAnnotation && currentAnnotation.task !== prevAnnotation.task) {
       this.checkForFeedback(prevAnnotation.task);
     }
   }
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -77,7 +77,7 @@ class Classifier extends React.Component {
     }
 
     if (nextProps.classification !== this.props.classification) {
-      const { annotations } = nextProps.classification;
+      const annotations = nextProps.classification.annotations.slice();
       this.setState({ annotations });
     }
   }
@@ -224,7 +224,7 @@ class Classifier extends React.Component {
   }
 
   handleAnnotationChange(classification, newAnnotation) {
-    const { annotations } = classification;
+    const annotations  = classification.annotations.slice();
     annotations[annotations.length - 1] = newAnnotation;
     classification.update({ annotations });
     this.setState({ annotations });
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -45,8 +45,7 @@ class Classifier extends React.Component {
       showingExpertClassification: false,
       subjectLoading: false,
       annotations: [],
-      modelScore: null,
-      currentTaskId: null
+      modelScore: null
     };
   }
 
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Test feedback project is here: https://remove-feedback-hack.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test/classify

app/classifier/classifier.jsx
```diff
@@ -175,7 +157,7 @@ class Classifier extends React.Component {
   }
 
   updateAnnotations() {
-    const { annotations } = this.props.classification;
+    const annotations = this.props.classification.annotations.slice();
     this.setState({ annotations });
     if (this.props.feedback.active) {
```
mcbouslog: not directly related to the changes in this PR, but in this demo project with feedback - https://remove-feedback-hack.pfe-preview.zooniverse.org/projects/markb-panoptes/feedback-testing-mb - if you first make a classification in workflow "Feedback 4" then go back to the project landing and try to go into workflow "Feedback 3" it hangs and doesn't load a subject, related to #4243.

I think changing this line to:
`if (annotations.length > 0 && this.props.feedback.active) {`
solves the issue.

app/classifier/classifier.jsx
```diff
@@ -127,7 +109,7 @@ class Classifier extends React.Component {
   }
 
   getActiveTask(state) {
```
mcbouslog: not a review comment, just for my understanding, can move this conversation to Slack if prefer, but how is `getActiveTask` used? I don't see it elsewhere in this file, or in the classifier directory.

eatyourgreens: Very good question. I fixed it, but I don't know where it's called either.

app/classifier/classifier.jsx
eatyourgreens: I can have a look into this bug tomorrow. If I understand correctly, it's a bug on master rather than a bug in this PR?

eatyourgreens:
The classifier confuses me too, mostly because classifications listen to change events that are emitted by other components, like the task navigation buttons. 

There's no hurry on this. I saw @rogerhutchings comment about hacking classifier state, and realised what the problem was, so thought I'd push a fix while it was fresh in my mind.

app/classifier/classifier.jsx
mcbouslog: that's correct, on master, not this PR, figure I'd mention if feel like including, but no worries if not

eatyourgreens: Probably quicker to fix and merge that one separately, rather than fix it here. I'll have a look at it tomorrow but it sounds like a quick fix.

eatyourgreens: Turns out it was easier to fix here. Should be fixed by 4bba343.

app/classifier/classifier.jsx
eatyourgreens: `getActiveTask` was added in #4191 but isn't actually used anywhere. I've removed it.

app/classifier/classifier.jsx
eatyourgreens: This doesn't fix the bug in the feedback action itself, so I've opened #4275.

eatyourgreens:
Rebased to remove the experiments client.

app/classifier/classifier.jsx
```diff
@@ -78,40 +75,23 @@ class Classifier extends React.Component {
...
-    // componentDidUpdate is a linting error, and with good reason.
-    const currentTaskId = _.get(_.last(this.state.annotations), 'task', '');
-
-    if (prevState.currentTaskId !== currentTaskId) {
-      this.setState({ currentTaskId }, () => this.forceUpdate());
-    }
-
-    if (prevState.currentTaskId !== this.state.currentTaskId) {
-      this.checkForFeedback(prevState.currentTaskId);
+    if (prevAnnotation && currentAnnotation.task !== prevAnnotation.task) {
```
mcbouslog: I'm not seeing feedback pop-up anymore, I think because it's not getting past the conditional comparing `currentAnnotation.task !== prevAnnotation.task`

eatyourgreens: My fault, I think. It was working until I rebased.

eatyourgreens: Looking at this now. The Done button at the end of a workflow doesn't change the current task, so feedback can't be triggered for the last task in a workflow.

eatyourgreens: This has now confused me because @rogerhutchings original hack shouldn't work either, since it also triggered feedback when the task changes.

eatyourgreens: I broke the rules lookup when I removed `_.get()`. The correct lookup is `feedback.rules[taskKey]`. Fixed now.

eatyourgreens: That said, I'm still confused as to how the feedback check is triggered at the end of a classification.

mcbouslog: My bad, this wasn't the right line of code to refer to, as noted by your fix with `feedback.rules[taskKey]`, the feedback on single task workflows is triggered through `completeClassification` which is sent through to https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/task-nav.jsx#L92.

eatyourgreens: Ah yes, I missed that `this.checkFeedback` is called in more than one place.

It would make sense to have a callback, from the task navigation, for task changes too. That might be more reliable than listening to annotation changes.

PR #4283, opened 2018-01-12
Replace all createClass calls with createReactClass and add required imports

tingard:
Staging branch URL: https://replace-react-create-class.pfe-preview.zooniverse.org/

Fixes use of `React.createClass` by replacing with `createReactClass` from `create-react-class` as detailed [here](https://reactjs.org/docs/react-without-es6.html).

Used a python script [temporarily hosted here](https://www.dropbox.com/s/xtds2x47te3gydq/fixPfeCreateClassPropTypes.py?dl=0) (TODO set up a repo) to automatically replace all instances of `createClass` in zooniverse.

Others also running the script and checking all required files are changed would be appreciated.

Script can also be used to replace `React.PropTypes` with `PropTypes` from `react-prop-types`.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


tingard:
I've realised it would be better to run the Facebook plugin first, as it cleverly creates classes sometimes, rather than blindly moving to `create-react-class`. Worth re-doing things in that order?

eatyourgreens:
Sounds like it's worth trying.

tingard:
After reading through [all this](https://github.com/reactjs/react-codemod#explanation-of-the-new-es2015-class-transform-with-property-initializers) I've become prohibitively paranoid that I'm not being nearly careful enough, and that something has almost definitely broken here.

eatyourgreens:
I'm happy to stage this on Monday, so we can check if anything broke. Unfortunately, the files that this changes will mostly be ones that haven't had tests written yet.

eatyourgreens:
Staged at https://replace-react-create-class.pfe-preview.zooniverse.org/?env=production

eatyourgreens:
> I've realised it would be better to run the Facebook plugin first, as it cleverly creates classes sometimes, rather than blindly moving to create-react-class. Worth re-doing things in that order?

Would it make sense to run the Facebook codemod across javascript files, and use the Python script only to convert coffeescript?

Staging branch looks good, from a preliminary browse around the site and doing a few classifications.

tingard:
I tried the codemod over the weekend and it did break things, working today to understand why.

This PR is only a first step towards react 16, there's a fair few calls to depreciated functions as far as I can see, so definitely worth going via 15.6 and watching for console warnings, as that should be easier to debug. This is what I'll spend a majority of this week on I think.

tingard:
Cool, ~16 jsx files now using `class __ extends React.Component`, rest on `createReactClass`. All cjsx components on `createReactClass`.
Easy enough to do the same for `PropTypes` in a different (same?) PR, though I'll just use the python script as it's a simple switch.
From there I'll create a new branch and attempt the react-15.6 switch, see what pops up and report back to slack because I'll probably be a little out of my depth.

eatyourgreens:
Cheers. I've pushed the latest changes to https://replace-react-create-class.pfe-preview.zooniverse.org/

If nothing major comes up on this branch, I'd say merge it and deal with the React and PropTypes upgrades in separate PRs.

mcbouslog:
I also tried a handful of what I thought are typical actions on staging (classifying, messaging, collecting, logging in/out, etc.) and encountered no issue. Thanks @tingard!

eatyourgreens:
Looks like this needs an update because `ActiveUsers` was converted to ES6.

PR #4277, opened 2018-01-10
Remove classification change listener

eatyourgreens:
Staging branch URL: https://classification-update.pfe-preview.zooniverse.org/

This builds on #4272 (which needs to be merged first.)

Removes the classification change listener from the classifier. Tasks are updated to modify annotations via the `onChange` callback, instead of calling `classification.update('annotations')`.

# Required Manual Testing

- [x] Does the non-logged in home page render correctly?
- [x] Does the logged in home page render correctly?
- [x] Does the projects page render correctly?
- [x] Can you load project home pages?
- [x] Can you load the classification page?
- [x] Can you submit a classification?
- [x] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [x] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [x] Are the tests passing locally and on Travis?

# Optional

- [x] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [x] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

Issue #4292, opened 2018-01-15
Feedback modal shown twice


mcbouslog:
## Expected behavior
Clicking "ok" in feedback modal dismisses modal.

## Current behavior
After initial click of "ok" in feedback modal, the same modal is shown. Modal is dismissed after second click of "ok."

## Steps to replicate
https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/feedback-testing-mb/classify?reload=0&workflow=3133

eatyourgreens:
I'm only seeing the feedback once when I try this out.

mcbouslog:
Yep, me too! A subsequent PR may have fixed the issue. Thanks,

eatyourgreens:
It might be worth checking on #4295. That's the change that we merged, then later reverted because it broke the combo task.

mcbouslog:
Ah yes, actually that staged branch does show the feedback modal twice. I'll make a comment in the PR.

eatyourgreens:
It's because `checkFeedback()` is called when the classification completes, then again when annotation state changes and #4295 fixes some cases where annotation changes didn't update classifier state.

PR #4295, opened 2018-01-16
Remove classification change listener

### LINKED ISSUE #4296: Hide previous marks not working since #4277
Related to changes in #4277 - the hide previous marks are not working.

## Expected behavior
Hide previous marks should hid current drawing tasks annotations on the subject viewer.
## Current behavior
Hide previous marks does not hide any drawing annotations but can remove them entirely.

## Steps to replicate

Goto https://classification-update.pfe-preview.zooniverse.org/projects/penguintom79/seabirdwatch/classify and draw some marks then try to use hide previous marks checkbox

## Additional information
Related to the change listener PR in #4277.

- **Operating system:**
- **Browser:**

eatyourgreens:
Staging branch URL: https://classification-update.pfe-preview.zooniverse.org

Fixes #4296

Repeat of #4277 but this should properly account for drawing tasks when they are used as part of combo tasks.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/task.jsx
+++ b/app/classifier/task.jsx
@@ -48,15 +48,14 @@ class Task extends React.Component {
       taskTypes: tasks,
       workflow,
       classification,
-      onChange: () => this.handleAnnotationChange
+      onChange: this.handleAnnotationChange
     };
 
     return (
       <div className="task-container" style={this.props.subjectLoading ? disabledStyle : null}>
         <div className="coverable-task-container">
           {persistentHooksBeforeTask.map((HookComponent, i) => {
-            const key = i + Math.random();
-            return (<HookComponent key={key} {...taskHookProps} />);
+            return (<HookComponent key={i} {...taskHookProps} />);
           })}
 
           {!!annotation &&
@@ -77,8 +76,7 @@ class Task extends React.Component {
           }
 
           {persistentHooksAfterTask.map((HookComponent, i) => {
-            const key = i + Math.random();
-            return (<HookComponent key={key} {...taskHookProps} />);
+            return (<HookComponent key={i} {...taskHookProps} />);
           })}
 
           <hr />
--- a/app/classifier/task.jsx
+++ b/app/classifier/task.jsx
@@ -12,8 +12,9 @@ class Task extends React.Component {
 
   handleAnnotationChange(newAnnotation) {
     const { classification } = this.props;
-    classification.annotations[classification.annotations.length - 1] = newAnnotation;
-    classification.update('annotations');
+    const annotations = classification.annotations.slice();
+    annotations[annotations.length - 1] = newAnnotation;
+    this.props.updateAnnotations(annotations);
   }
 
   render() {
@@ -47,7 +48,7 @@ class Task extends React.Component {
       taskTypes: tasks,
       workflow,
       classification,
-      onChange: () => classification.update()
+      onChange: () => this.handleAnnotationChange
     };
 
     return (
@@ -123,12 +124,17 @@ Task.propTypes = {
   task: PropTypes.shape({
     type: PropTypes.string
   }),
+  updateAnnotations: PropTypes.func,
   user: PropTypes.shape({
     id: PropTypes.string
   }),
   workflow: PropTypes.shape({
     id: PropTypes.string
   })
+};
+
+Task.defaultProps = {
+  updateAnnotations: () => null
 };
 
 export default Task;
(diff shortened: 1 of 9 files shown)


app/classifier/task.jsx
```diff
@@ -11,8 +11,9 @@ class Task extends React.Component {
 
   handleAnnotationChange(newAnnotation) {
     const { classification } = this.props;
-    classification.annotations[classification.annotations.length - 1] = newAnnotation;
-    classification.update('annotations');
+    const annotations = classification.annotations.slice();
+    annotations[annotations.length - 1] = newAnnotation;
```
camallen: OOI why does this replace the last existing annotation with the new one, is it because the last one was a default? 

I tried to trace how the new default annotation gets it's actual value stored and it looks like it's in the specific task component on a change event. If that is so then any default will be updated to be a real one, right? Or i am way off track?

app/classifier/tasks/drawing/hide-previous-marks-toggle.cjsx
```diff
@@ -18,7 +18,7 @@ HidePreviousMarksToggle = React.createClass
   setPreviousMarks: (count) ->
     @_lastKnownAnnotationCount = @props.classification.annotations.length
     @props.classification._hideMarksBefore = count
-    @props.onChange @props.classification
+    # @props.onChange @props.classification
```
camallen: remove / re-instate if this is not / required

app/classifier/task.jsx
eatyourgreens: This expects to get an updated annotation when a task changes. It simply overwrites your previous annotation, for this task, with the new one.

eatyourgreens: The callback then forces a re-render by updating `state.annotations` in the classifier.

eatyourgreens:
I've added a fix for Hide Previous Marks (#4296)

app/classifier/task.jsx
camallen: right, got it.

app/classifier/tasks/drawing/hide-previous-marks-toggle.cjsx
```diff
@@ -7,18 +7,17 @@ HidePreviousMarksToggle = React.createClass
...
+  getInitialState: ->
+    checked: false
 
   setPreviousMarks: (count) ->
-    @_lastKnownAnnotationCount = @props.classification.annotations.length
     @props.classification._hideMarksBefore = count
-    @props.onChange @props.classification
+    checked = count > 0
+    @setState { checked }
+    # this is a hack to force the parent to render again with marks hidden
```
camallen: should this be extracted to a function with a name like `force_parent_render_hack()` to signify the intention of it? 

It might allow the removal of the comment and the commit history could store the context of the reason for that code?

eatyourgreens: That's not a bad idea, though the linter would make us call it `forceParentRender` because it complains if names aren't camel-case.

camallen: i can't help myself when writing code snippets these days, i should have 🐫'd it

eatyourgreens:
Probably needs a rebase now that #4283 has merged.

eatyourgreens:
Rebased and re-staged.

eatyourgreens:
I've run through every task in the dev classifier now, and haven't been able to find a problem with this.

simoneduca:
Let's do this (after rebasing)!

eatyourgreens:
@simoneduca does this still need rebasing?

mcbouslog:
It appears this causes the feedback modal to show twice:
https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/feedback-testing-mb/classify?reload=0&workflow=3133

^ moved from #4292

simoneduca:
What's the status of this @eatyourgreens? I don't think so, but has the issue with the modal been fixed?

eatyourgreens:
@simoneduca the issue with the modal is a bug in the experimental feedback feature. This PR is waiting on someone to test it and confirm it works (or request changes.)

Actually reading back, sounds like [you've already tested it](#issuecomment-358609243).

simoneduca:
Yeah I couldn't find anything wrong, but I wasn't sure if the modal issue was caused by this. basically I didn't understand why it was referenced here.

eatyourgreens:
This PR changes how annotation changes are reflected in classifier state. Basically, any change in an annotation updates classifier state now, so that we don't miss annotation changes when we pass annotations down to child components.

Feedback relied on the old behaviour, which I _think_ had bugs, to detect when annotations had changed and feedback should be shown.

simoneduca:
Sounds like it's good to go then.

PR #4305, opened 2018-01-18
[WIP] Upgrading enzyme to 3.x

tingard:
Staging branch URL:

Upgrades enzyme to 3.3.0. Enzyme migration guide can be found [here](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#lifecycle-methods)

Currently have a few tests still failing, have already corrected a few.

### Currently failing tests

- [ ] Shortcut functionality should show an active button with a shortcut
- [ ] ShortcutEditor methods should call toggleShortcut with an input toggle
- [ ] ShortcutEditor methods should call remove handler when clicked
- [ ] ShortcutEditor methods should call add handler when clicked
- [ ] SVGImage it should contain an image element
- [ ] SocialLinksEditor should call handleRemoveLink() when link is removed

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4307, opened 2018-01-18
 Upgrading enzyme to 3.x

eatyourgreens:
Staging branch URL: https://upgrading-enzyme-3.pfe-preview.zooniverse.org

Continues #4305

Upgrades React to 15.6.2 and enzyme to 3.3.0. Enzyme migration guide can be found [here](https://github.com/airbnb/enzyme/blob/master/docs/guides/migration-from-2-to-3.md#lifecycle-methods)

~~Currently have a few tests still failing, have already corrected a few.~~

### Currently failing tests

- [x] Shortcut functionality should show an active button with a shortcut
- [x] ShortcutEditor methods should call toggleShortcut with an input toggle
- [x] ShortcutEditor methods should call remove handler when clicked
- [x] ShortcutEditor methods should call add handler when clicked
- [x] SVGImage it should contain an image element
- [x] SocialLinksEditor should call handleRemoveLink() when link is removed

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
@tingard Does this branch load React 15.4 or 15.6 for you? I'm seeing the package lock file revert back to 15.4 when I run npm install.

eatyourgreens:
I think I've fixed everything except the shortcut editor. I'm having a bit of trouble understanding those tests.

tingard:
Oh awesome, yeah I couldn't make sense of those when I glanced at then. Well check react in a sec, been in a talk

eatyourgreens:
I think the only tests that are broken now are tests that spy on `wrapper.instance()` methods. I'm not sure what's going on there. I thought `wrapper.instance()` gives you back the React component, so spying on its methods should work.

tingard:
@eatyourgreens `rm -rf node_modules && npm install && npm ls` gives react-15.6.2 for me. It also complains about missing deps for markdown and seven-ten

eatyourgreens:
@simoneduca in where, sorry?

simoneduca:
Sorry I was talking about the [first ShortcutEditor test that fails](https://github.com/zooniverse/Panoptes-Front-End/blob/c6e9680822aa280311c374a3026159b536514ea3/app/classifier/tasks/shortcut/editor.spec.js#L87) for me

eatyourgreens:
The error I'm getting is that `toggleStub` is not called, even though it actually is.

eatyourgreens:
I'm also seeing that `workflow.tasks` is actually changed by that test, even though `toggleStub` is stubbed and, therefore, should not actually do anything when called.
```javascript
1 { T1: { answers: [ [Object], [Object] ], type: 'shortcut' } },
2 {} _
```

eatyourgreens:
Fixed now. It seems enzyme 3 is fussy about you stubbing event handlers before the component is instantiated, rather than instantiating the component and then stubbing instance methods.

simoneduca:
Nice one! Well done both!

eatyourgreens:
Re-staged to bring markdownz and seven-ten up to React 15.6 too. I've also removed `React.createClass` and `React.PropTypes` from seven-ten.

app/pages/project/about/index.spec.js
```diff
@@ -9,6 +9,8 @@ function Page() {
   );
 }
 
+// TODO: THIS BREAKS EVERYTHING, WHY?
```
srallen: Is this comment still relevant?

app/pages/lab/social-links-editor.spec.js
```diff
@@ -2,8 +2,18 @@ import React from 'react';
 import assert from 'assert';
 import { shallow, mount } from 'enzyme';
 import sinon from 'sinon';
+import apiClient from 'panoptes-client/lib/api-client';
 import SocialLinksEditor from './social-links-editor';
 
+function mockPanoptesResource(type, options) {
+  const resource = apiClient.type(type).create(options);
+  apiClient._typesCache = {};
```
srallen: Non-blocking comment, but there is an undocumented `uncache()` class method on Resource: https://github.com/zooniverse/json-api-client/blob/c6dd8e406c2fcb02db1a84f85c6b7d85fe068179/src/resource.coffee#L86-L92

I have no idea if it does what is needed for these tests to work though.

app/classifier/tasks/survey/index.spec.js
```diff
@@ -5,68 +5,68 @@ import SurveyTask from './';
 import Choice from './choice';
 import { workflow } from '../../../pages/dev-classifier/mock-data';
 
-describe('Survey Task', function(){
+describe('Survey Task', () => {
```
srallen: Arrow functions are actually discouraged in mocha: https://mochajs.org/#arrow-functions

But with these tests, I don't think it matters. Just an FYI.

app/pages/project/about/index.spec.js
tingard: ah no, it was breaking when I was trying to configure enzyme in setup.js, fixed by making sure globals were set before enzyme configuring.

app/classifier/tasks/survey/index.spec.js
simoneduca: TIL, thanks @srallen

eatyourgreens:
I've removed that outdated comment, upgrading markdownz to 7.4.1 and replaced all the arrow functions in tests with `function` declarations. This should be good to merge once Travis has finished.

Once this is merged, I'll update #4310 to bring in these changes.

app/classifier/tasks/survey/index.spec.js
tingard: Ah okay, interesting! I'll have to hack my linter for personal projects now...

app/pages/lab/social-links-editor.spec.js
eatyourgreens: From the code, it looks like it removes the current resource from the client cache, so `resource.uncache()` probably would work here too.

PR #4309, opened 2018-01-19
Revert "Facelift - Project Header Bar"

camallen:
Reverts zooniverse/Panoptes-Front-End#4218 

To fix broken keyboard navigation, tab behaviour for autocomplete select elements, reported to be broken in NFN https://www.zooniverse.org/talk/17/525366?comment=872032


eatyourgreens:
#4218 upgraded a bunch of unrelated packages, including `react-select`, which then broke NfN.

Issue #4221, opened 2017-12-12
 Adding active info to the workflows page


mschwamb:
Now that there is the workflows page, it seems like having info on which ones are active and the ability to turn them on and off might be now better to have on the workflows page rather than on the visibility page

Workflows page

<img width="897" alt="screen shot 2017-12-11 at 11 53 55 pm" src="https://user-images.githubusercontent.com/753325/33868996-665f79e2-ded4-11e7-90b0-f3ab2588b15c.png">

Part of the Visibility page

<img width="879" alt="screen shot 2017-12-12 at 12 36 04 am" src="https://user-images.githubusercontent.com/753325/33869014-87878f6a-ded4-11e7-8e7f-ce9c819ab30e.png">

mschwamb:
The added completion is nice. I think adding which workflows are active to this area would be really helpful

<img width="1189" alt="screen shot 2018-01-30 at 8 33 43 pm" src="https://user-images.githubusercontent.com/753325/35608380-fb184f56-05fc-11e8-9fd2-ad30caa48dc2.png">

simoneduca:
@mschwamb I'm planning to move the relevant part on the visibility section on the workflow page. So that should take already care of the active bit as it wouldn't make sense to have it there as well in the list. It might look a bit odd to have two separate tables on the same page. I'll open a PR and discuss there.

mschwamb:
sounds good

simoneduca:
@mschwamb you can test latest changes here https://workflows-page.pfe-preview.zooniverse.org/

mschwamb:
Looks good. I'd add the help text about the active/not active from the viability page to the workflow page. 

I'd suggest moving everything about the workflows from the visibility page and put it on the workflow page. Then the visibility page is for the overall project again.

PR #4313, opened 2018-01-22
Improve workflows page

### LINKED ISSUE #4303: Signal *not* to make all workflows inactive when a project is finished.
A few projects have done that. The result is a broken classify page. Some text in the lab should warn builders not make all workflow inactive.
In addition, it would be useful to automatically move a project to "paused" or "finished" when all data is retired.

simoneduca:
Staging branch URL: https://fix-4303.pfe-preview.zooniverse.org/

Fixes #4303, counts toward #4221. 

This is the first step to improve the workflow page. Moving the workflow settings from their current place (the visibility page) to the workflows seems a good idea, but I'll do that in a separate PR.
Changes:

1. Display workflow completeness.
2. Add note about making workflows inactive.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflows.jsx
+++ b/app/pages/lab/workflows.jsx
@@ -9,10 +9,11 @@ import Paginator from '../../talk/lib/paginator';
 
 const WorkflowsPage = (props) => {
   const renderWorkflow = ((workflow) => {
+    let progressPercentage = workflow.completeness * 100;
     return (
       <li key={workflow.id}>
         <Link key={workflow.id} to={props.labPath(`/workflows/${workflow.id}`)} className="nav-list-item" activeClassName="active">
-          {workflow.display_name}
+          {workflow.display_name}{' -- '}{`${progressPercentage.toFixed(0)} % Complete`}
           {(props.project.configuration && workflow.id === props.project.configuration.default_workflow) && (
             <span title="Default workflow">{' '}*{' '}</span>
           )}
@@ -33,8 +34,9 @@ const WorkflowsPage = (props) => {
       <p>A workflow is the sequence of tasks that you’re asking volunteers to perform.</p>
       <p>An asterisk (*) denotes a default workflow.</p>
       <p>If you have multiple workflows you can rearrange the order in which they are listed on your project's front page by clicking the reorder view button and then clicking and dragging on the left gray tab next to each workflow title listed below.</p>
+      <p>NOTE: Please leave at least one active workflow; even if all workflows are 100% complete.</p>
       <p>{reorderButton}</p>
-      
+
       {props.reorder &&
         <DragReorderable tag="ul" className="nav-list" items={props.workflows} render={renderWorkflow} onChange={props.handleWorkflowReorder} />}
 
--- a/app/pages/lab/workflows.jsx
+++ b/app/pages/lab/workflows.jsx
@@ -8,7 +8,7 @@ import LoadingIndicator from '../../components/loading-indicator';
 import Paginator from '../../talk/lib/paginator';
 
 const WorkflowsPage = (props) => {
-  const renderWorkflow = ((workflow) => {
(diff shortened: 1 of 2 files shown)


app/pages/lab/workflows.jsx
```diff
@@ -9,10 +9,11 @@ import Paginator from '../../talk/lib/paginator';
 
 const WorkflowsPage = (props) => {
   const renderWorkflow = ((workflow) => {
+    let progressPercentage = workflow.completeness * 100;
     return (
       <li key={workflow.id}>
         <Link key={workflow.id} to={props.labPath(`/workflows/${workflow.id}`)} className="nav-list-item" activeClassName="active">
-          {workflow.display_name}
+          {workflow.display_name}{' -- '}{`${progressPercentage.toFixed(0)} % Complete`}
```
srallen: I'm not too keen on this. It makes it look like the percentage is a part of the display name.

app/pages/lab/workflows.jsx
```diff
@@ -33,8 +34,9 @@ const WorkflowsPage = (props) => {
       <p>A workflow is the sequence of tasks that you’re asking volunteers to perform.</p>
       <p>An asterisk (*) denotes a default workflow.</p>
       <p>If you have multiple workflows you can rearrange the order in which they are listed on your project's front page by clicking the reorder view button and then clicking and dragging on the left gray tab next to each workflow title listed below.</p>
+      <p>NOTE: Please leave at least one active workflow; even if all workflows are 100% complete.</p>
```
srallen: Perhaps bold lettering can be used for emphasis instead of capitals?

simoneduca: Sure.

app/pages/lab/workflows.jsx
simoneduca: Fair enough. Turning the list into a table should take care of that.

simoneduca:
Looking into why that broke tests.

simoneduca:
The test fails because it's looking for class name `.nav-list li`, which isn't suitable anymore for a table.

eatyourgreens:
Looks like re-ordering workflows doesn't work any more, either.

simoneduca:
`table` really doesn't play well with `DragReorderable` 😭

eatyourgreens:
@simoneduca works for me with `<DragReorderable tag="table"`. What errors are you getting?

simoneduca:
@eatyourgreens works here too, but I get this warning `<tr> cannot appear as a child of <table>. See DragReorderable > table > tr. Add a <tbody> to your code to match the DOM tree generated by the browser.` Not sure if that might lead to something unpredictable, but it seems not.

CKrawczyk:
If it helps, I have used `DragReorderable` for a table for the external links editor page before: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/external-links-editor.jsx

simoneduca:
Aha! thanks @CKrawczyk that's helpful.

simoneduca:
Perhaps not the neatest way but the warning is gone now.

srallen:
The reordering view doesn't need to be a table. I don't think project owners would need the tabular data and settings for the workflows when reordering, probably just the display names. But if you've already gotten it to work, I guess it doesn't need to change back.

simoneduca:
Yeah it should be working now.

simoneduca:
@srallen are you happy with this?

simoneduca:
Good points! Thanks @srallen

simoneduca:
Restaged. Also doesn't look it needs rebasing.

simoneduca:
I borked a rebase. Hang in there.

simoneduca:
Done, latest client version and test workflow-selection are in there now.

PR #4319, opened 2018-01-25
Better signalling for official projects

### LINKED ISSUE #4304: How to signal a project has been officially launched by the Zooniverse?
It's not very clear when a project has been approved by the Zooniverse vs publicised privately by the project owner. This maybe causing confusion amongst our volunteers.

There is a banner at the bottom of the home page of non-official projects, but it's easily missed.
screen shot 2018-01-18 at 10 27 26

An option for solving this issue would be to clearly mark an official project (as well as leaving the existing banner). Perhaps something like "verified" users on Twitter get a tick next to their name.

@beckyrother what are your thoughts about this?

simoneduca:
- [x] fix tests.

Staging branch URLs.

For testing official projects: https://official-projects.pfe-preview.zooniverse.org/projects/ghthomas/project-plumage?env=production

For beta-approved projects (admin-mode): https://official-projects.pfe-preview.zooniverse.org/projects/abpullan/the-edwardian-postcard-project-1901-1910?env=production

For all other projects(admin-mode): https://official-projects.pfe-preview.zooniverse.org/projects/dougbrn/lsst-zoo-test?env=production

Fixes #4304.

Describe your changes.
Move disclaimer up.
Add checkmark for lauch-approved projects.

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional


simoneduca:
Sure. Just a question: where is the best place for them? I'm looking in [this folder](https://github.com/zooniverse/Panoptes-Front-End/tree/cd079d5aa4a0349cab8009ff4215a1fcf3c82261/app/pages/project/components/ProjectNavbar/components), but I can't find the relevant bits, e.g. "under review", etc.

simoneduca:
Btw, where all the classnames coming from? They all have meaningless names in the rendered html.

srallen:
@simoneduca I think the 'under review' label got missed in this updated version. That label and the checkmark probably either fit best in the [ProjectTitle component](https://github.com/zooniverse/Panoptes-Front-End/blob/cd079d5aa4a0349cab8009ff4215a1fcf3c82261/app/pages/project/components/ProjectNavbar/components/ProjectTitle/ProjectTitle.jsx) or as separate components if necessary. 

The classes are generated by [styled-components](https://www.styled-components.com/) that I mentioned are being experimented with these new components.

As I said, I'm working on the tests, so if you could get these two things added into the new navbar, then I'm happy to pick up the testing portion for them.

simoneduca:
Sounds good. Sorry I might have missed the discussion about this. What's the status of the new header? It seems it's still experimental? I ask because I'm not clear if under review and checkmark should be in the new header only or in both navbars (in which case, I'll need to fix these tests anyway).

srallen:
The new navbar is basically done. There are some minor tweaks and fixes I'm working on in addition to the tests. Adding 'Under Review' and the checkmark could be a couple of those tweaks. I think whether or not it should be done to both is a question for @beckyrother and will defer to her opinion, but I think that in terms of effort spent, adding the checkmark could just be rolled into the new version rather than add it to both.

beckyrother:
@simoneduca is working on adding the checkmark: https://github.com/zooniverse/Panoptes-Front-End/issues/4304

simoneduca:
After asking on slack, I decided to remove the original test that checked the position of the disclaimer because it was too brittle. I kept the tests for testing the conditional rendering of the disclaimer.

simoneduca:
I think this is ready to go and other tweaks to the new navbar can be added in a new PR.

@srallen I've noticed the `<h1>` in `ProjectTitle` doesn't play nicely with long title. Also it looks a bit odd to have "under review" as big as the title. In an attempt to make that better, I've `em`phasised it. But probably @beckyrother has a better idea.

Test link for under review: https://official-projects.pfe-preview.zooniverse.org/projects/abpullan/the-edwardian-postcard-project-1901-1910?facelift=true&env=production 

Test link for official projects: https://official-projects.pfe-preview.zooniverse.org/projects/ghthomas/project-plumage?env=production&facelift=true

app/pages/project/home/project-home.jsx
```diff
@@ -21,6 +21,14 @@ const ProjectHomePage = (props) => {
     { 'project-home-page__description--top-padding': !props.organization }
   );
 
+  const launchApproved = (!props.project.launch_approved) ?
```
srallen: If the render when false is null, this can just be done inline on line 77:
```
{!props.project.launched_approved &&
  <Translate />}
```

app/pages/project/components/ProjectNavbar/components/ProjectTitle/ProjectTitle.jsx
```diff
@@ -25,7 +25,31 @@ const StyledLink = styled(IndexLink).attrs({
   }
 `;
 
-function ProjectTitle({ link, title }) {
+function ProjectTitle({ launched, link, title, underReview }) {
+  if (underReview) {
```
srallen: The duplication doesn't seem necessary here. The check for the states can be done inline:
```
<H1>
  <StyledLink to={`${link}?facelift=true`}>
    {underReview &&
      <p><em>Under Review</em></p>
    {title}
    {launched &&
      <img />}
  </StyledLink>
</H1>
```

app/pages/project/home/project-home.spec.js
```diff
@@ -68,6 +66,34 @@ describe('ProjectHome', function() {
...
 
+  describe('with a launch-approved project', function () {
+    project.launch_approved = true;
+    wrapper = shallow(
+      <ProjectHome
+        project={project}
+        background={background}
+        translation={translation}
+      />);
+    const disclaimer = wrapper.find('Translate[className="project-disclaimer"]');
```
srallen: I'd like to move away from relying on class names as selectors as this can be brittle and if we do indeed go forward with styled-components, then we won't be able to use class names since they're generated. Enzyme supports object properties as selectors too:
`wrapper.find({ content: 'project.disclaimer' })`

app/pages/project/home/project-home.spec.js
```diff
@@ -68,6 +66,34 @@ describe('ProjectHome', function() {
...
+
+  describe('without approval', function () {
+    project.launch_approved = false;
+    wrapper = shallow(
+      <ProjectHome
+        project={project}
+        background={background}
+        translation={translation}
+      />);
+    const disclaimer = wrapper.find('Translate[className="project-disclaimer"]');
```
srallen: Same comment about class name use in the selector.

beckyrother:
@simoneduca Sorry, I guess I was confused about the "Under review" part. You're right, that does not look good. I thought we were removing the words entirely and just going with the sentence at the bottom of the top section.
@srallen good idea to use font-awesome, let's go with that so the icon can be bigger and less pixellated.

app/pages/project/components/ProjectNavbar/components/ProjectTitle/ProjectTitle.jsx
simoneduca: Good point thanks.

app/pages/project/home/project-home.spec.js
simoneduca: Nice. I agree moving away is a good idea, regardless of stlyed-components, which I still find pretty awkward. What made you think of trying them out, i.e. advantages?

simoneduca:
@srallen sorry I'm having problems staging my changes, but all comments should have been addressed now.

app/pages/project/home/project-home.spec.js
srallen: A few reasons:
- It'll be better and easier to share styles in shared components using JS in CSS. It would make sure the styles are correctly applied and have no issues with trying to setup to ingest CSS from a package, setup webpack, class name collisions, etc. 
- It has a nice theming feature which will make it easier to manage which theme is used with app state. The classifier redesign will have a light and dark theme.
- Grommet v2 uses styled-components for ease of theming. 
- We're interested in using Grommet with the classifier redesign.

simoneduca:
Done (sorry I still can stage, I'm investigating). I've tried writing the styled components bit. Not sure I did it right, but it works.

simoneduca:
Actually I've just manage to stage the branch :)

app/pages/project/home/project-home.spec.js
simoneduca: Cool, seems like a good idea then, thanks.

srallen:
@beckyrother would you like to do a review pass on this? FWIW, it seems the under review message lost its styling at some point. Not from this PR. It used to be small, under the project title, and teal. I'm happy to fix that in my branch with the rest of the final project nav improvements.

beckyrother:
@srallen Yes, it would be great to reinstate that small label the way that it was – small and teal. Thanks for checking!

simoneduca:
Thanks both! Is this good to go then?

PR #4324, opened 2018-01-29
[WIP] React 16 upgrade

tingard:
Creating a PR to track milestones

Staging branch URL:

Makes the final(?) changes to React 16

List of conflicting modules: 
- [ ] [react-component-resizable](https://www.npmjs.com/package/react-component-resizable)
- [ ] [modal-form](https://www.npmjs.com/package/modal-form)
- [ ] maybe [react-interpolate-component](https://www.npmjs.com/package/react-interpolate-component)
- [ ] ... 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


srallen:
Here's an open PR to update modal-form to use `prop-types` and `create-react-class`: https://github.com/zooniverse/modal-form/pull/33

Is there any chance we don't have to update react-router for React 16? React-router v4's API is pretty awful in my opinion and it'd take a bit of time to update all of the routes to use their new API. Version 3 might work if we do have to upgrade? If not, I have been looking into other routing libraries in a starter test react app.

tingard:
Looks like 3.2.0 of react-router will work with react 16 https://github.com/ReactTraining/react-router/issues/5561

tingard:
Current failing module is [drag-reorderable](https://www.npmjs.com/package/drag-reorderable)

tingard:
So if the idea of JSON subjects goes through I'll probably need to use gzip to keep file sizes down (test file is up to 2.8MB before gzipping and ~350kB after)

eatyourgreens:
zooniverse/drag-reorderable#4 removes `React.PropTypes` and upgrades that package to React 15.6.

tingard:
Good news is that we're now running tests. Bad news is it's still broken (Markdownz is saying `Cannot read property 'querySelectorAll' of null`)

tingard:
from [here](https://reactjs.org/docs/react-dom.html):
> `findDOMNode` only works on mounted components (that is, components that have been placed in the DOM). If you try to call this on a component that has not been mounted yet (like calling `findDOMNode()` in `render()` on a component that has yet to be created) an exception will be thrown.
Maybe should be replaced with refs?

eatyourgreens:
Markdownz and the SevenTen client might still be on React 15 here.

eatyourgreens:
I'm also seeing this warning when I run the tests, so the `DropdownTask` might be doing something that doesn't work in React 16.

```shell
Warning: Functions are not valid as a React child. This may happen if you return a Component instead of <Component /> from render. Or maybe you meant to call this function rather than return it.
    in div (created by DropdownTask) 
    in div (created by DropdownTask)
    in div (created by GenericTask)
    in GenericTask (created by DropdownTask)
    in DropdownTask (created by WrapperComponent)
    in WrapperComponent
```

eatyourgreens:
zooniverse/markdownz#63 should fix the DOM error, and also the same errors that appear in the browser console when you load a new subject.

srallen:
@tingard I think react-component-resizable might be cruft from the new project navbar redesign. I might have added it to see how it worked and forgot to uninstall it. I think it can be safely uninstalled.

tingard:
Thanks for your comments, I'll have a look into things on Monday!

eatyourgreens:
I've tried out zooniverse/markdownz#63 on a local copy of this branch. The tests ran with React warnings but they passed without errors.

eatyourgreens:
@tingard the git repo is just the source code. It would still need to be built after downloading.

tingard:
@eatyourgreens I'm not getting `src`, so nothing to build. Any ideas why?
All I'm doing is running 
```sh
npm install --save-dev https://github.com/zooniverse/markdownz.git#footnote-links
```
And then checking `node_modules/markdownz`

eatyourgreens:
Sorry, I meant you have to clone the repo, build it then install that to PFE. I don't know if it's possible to install it directly from git, since the repo doesn't have the lib directory.

tingard:
@eatyourgreens I'm seeing the same (only react warnings)! Thanks for getting that fixed, this is looking close to completion to me?

tingard:
Current breaking error: `react-router-scroll` requires `history` ^2.1.2, PFE uses ~4.7.2, errors when starting server.

eatyourgreens:
I'd prefer to keep the PropTypes and CreateClass changes separate, since we started those with the upgrade to 15.6, so I've opened #4382.

tingard:
Reverted my commit and rebased from master. Any suggestions regarding what to do about react-router-scroll? (rc-slider is fine once bumped to latest as far as I can tell)

tingard:
Having removed it from main.cjsx I'm not sure what it was trying to accomplish? Chrome seems to restore scroll itself on some pages

eatyourgreens:
It scrolls back to the top on some pages. The exact rules were in `main.cjsx` I think.

tingard:
When it's disabled then (afaik) react 16 runs. Though `isMounted` inside promise-render may be problematic.

eatyourgreens:
`PromiseRenderer` is an old component that was a mistake in retrospect. There’s some explanation in the Readme.

eatyourgreens:
We need that scroll behaviour, by the way, but I don’t remember  which route transitions it was set up for.

tingard:
whoops looking at the repo it's just that I haven't bumped to the latest version...

tingard:
@eatyourgreens if you have a moment to check this out I can't see anything obviously breaking any more. I should clear up the React warnings on test before merge though.

eatyourgreens:
I'll try to make time for this next week.

We should stage it somewhere, and have everyone test it out to see if anything has obviously broken. That's pretty much how we tested the upgrade from React 14 to 15.

tingard:
@srallen That was the case on the 22nd of feb. It'll need a rebase and check-through again now though, as new components/tests/package changes might break things. Could be easy but who knows!
I'd do this but am very pressed for time over the next two weeks, and then start a placement.

eatyourgreens:
The only merge conflict is the addition of `chart.js` to the package file, so that should be easy to take care of.

srallen:
@tingard That's ok. I'm happy to take a look since I'm keen to get upgraded to React 16 too. Could you make sure your fork is editable by us? Here's how to do that: https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

eatyourgreens:
I added @tingard to zooniverse on github a few weeks ago. I'll double check that, but you should be able to push directly to this repo. Let me know on Slack if that doesn't work.

tingard:
(I'm not going to pretend to fully understand how git forks work)
The branch is pushed to my fork, I have no clue if it's synced to the upstream zooniverse repo but can do that if needed

PR #4331, opened 2018-02-01
Adds status to workflows table

### LINKED ISSUE #4221: Adding active info to the workflows page

simoneduca:
Staging branch URL: https://workflows-page.pfe-preview.zooniverse.org/

Fixes #4221 .

Describe your changes.
Adds status tab to the workflows table.
I'm not keen on moving the other workflow settings from the visibility page here, as it'd look pretty messy. Plus the other settings seems already in the right place.
Happy to discuss though.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
-const WorkflowsPage = (props) => {
-  const renderWorkflowTable = ((workflow) => {
+class WorkflowsPage extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleWorkflowSettingChange = this.handleWorkflowSettingChange.bind(this);
+    this.state = {
+      active: false,
+      error: null
+    };
```
srallen: I don't see `this.state.active` nor `this.state.error` used anywhere in the code.

app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
+      .then(() => {
+        if (!workflow.active && workflow.id === defaultWorkflow) {
+          this.props.project.update({ 'configuration.default_workflow': null });
+          this.props.project.save();
+        }
+      })
+      .then(() => this.setState({ active: workflow.active }));
+  }
+
+  renderWorkflowTable(workflow) {
```
srallen: This render method is getting quite large. Perhaps it's time to break this out into its own functional component?

app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
-};
+        <p>A workflow is the sequence of tasks that you’re asking volunteers to perform.</p>
+        <p>An asterisk (*) denotes a default workflow.</p>
+        <p>If you have multiple workflows you can rearrange the order in which they are listed on your project's front page by clicking the reorder view button and then clicking and dragging on the left gray tab next to each workflow title listed below.</p>
+        <p><em>Note</em>: Please leave at least one active workflow; even if all workflows are 100% complete.</p>
+        <p>{reorderButton}</p>
+
+        {this.props.reorder &&
+          <DragReorderable
+            tag="ul" className="nav-list"
```
srallen: These should be on their own lines.

app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
+            Active
+          </label>
+        </td>
       </tr>
     );
-  });
+  }
 
-  const renderWorkflowList = ((workflow) => {
+  renderWorkflowList(workflow) {
```
srallen: I'm not sure how the DragReorderable component would handle it, but also this could possibly be its own functional component.

app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
+  constructor(props) {
+    super(props);
+    this.handleWorkflowSettingChange = this.handleWorkflowSettingChange.bind(this);
+    this.state = {
+      active: false,
+      error: null
+    };
+  }
+
+  handleWorkflowSettingChange(workflow, e) {
```
srallen: I think the intention of this component was just to be a functional view component. API calls should go into `WorkflowsContainer`

app/pages/lab/workflows.jsx
```diff
@@ -7,116 +7,161 @@ import WorkflowCreateForm from './workflow-create-form';
...
+    const checked = e.target.checked;
+    workflow.update({ active: checked }).save()
+      .catch(error => console.log(error))
+      .then(() => {
+        if (!workflow.active && workflow.id === defaultWorkflow) {
+          this.props.project.update({ 'configuration.default_workflow': null });
+          this.props.project.save();
+        }
+      })
+      .then(() => this.setState({ active: workflow.active }));
```
srallen: Why is state being set for the active state of a single workflow when this page is for all the project workflows? It's also not being used anywhere. Is the point to trigger a re-render? I think what you want to do is after this API call is made, then you'd want to call `getWorkflowList(page)` (in `WorkflowsContainer`) for the page that you're on. So, another reason why this should move into that component.

app/pages/lab/workflows.jsx
simoneduca: Done.

app/pages/lab/workflows.jsx
simoneduca: `DragReorderable` breaks when trying to pass a component to it, so I left `renderWorkflowList` alone given that it's not large.

simoneduca:
Restaged https://workflows-page.pfe-preview.zooniverse.org
Thanks @srallen your comments were very helpful and sorry for the rather messy first attempt.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    type="checkbox"
+                  />
+                  Active
+                </label>
+              </td>
+              <td>
+                <label htmlFor="stats_completeness">
+                  <input
+                    id="stats_completeness"
+                    checked={statsCompletenessType === 'classification'}
```
srallen: One thing to check is that this is now doing strict equality, whereas the old code wasn't and so may have been doing an implicit type conversion. Strict equality is always preferable in javascript if we can do that here.

app/pages/lab/workflows.spec.js
```diff
@@ -24,7 +24,7 @@ describe('WorkflowsPage', function () {
   const toggleReorderSpy = sinon.spy();
 
   before(function () {
-    wrapper = shallow(
+    wrapper = mount(
```
srallen: Why did this change to mount?

simoneduca: Because the test wasn't passing with the addition of the new `WorkflowsTable` component.

app/pages/lab/workflows-table.jsx
simoneduca: Both sides of the comparison are of type `String`, so it should be ok.

eatyourgreens: `statsCompletenessType` isn't defined in this function so these checkboxes are never checked.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+          <th>Name</th>
+          <th>Progress</th>
+          <th>Status</th>
+          <th>Completeness statistic</th>
+          <th>Statistic visibility</th>
+        </tr>
+      </thead>
+      <tbody>
+        {workflows.map((workflow) => {
+          const progressPercentage = workflow.completeness * 100;
```
eatyourgreens: What does this map do?

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+          const statsCompletenessType = (workflow.configuration) ? workflow.configuration.stats_completeness_type : 'retirement';
+          let statsVisible = workflow.active;
+          if (workflow.configuration && workflow.configuration.stats_hidden !== undefined) {
+            statsVisible = !workflow.configuration.stats_hidden;
+          }
```
eatyourgreens: My bad. I misread this bracket as closing the map callback.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    onChange={e => handleWorkflowSettingChange(e, meta.page, workflow)}
+                    type="checkbox"
+                  />
+                  Active
+                </label>
+              </td>
+              <td>
+                <label htmlFor="stats_completeness">
+                  <input
+                    id="stats_completeness"
```
eatyourgreens: This code generates multiple elements with the same ID. Element IDs must be unique in the DOM. You can safely remove the `id` here.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    checked={workflow.active}
+                    id="active"
+                    onChange={e => handleWorkflowSettingChange(e, meta.page, workflow)}
+                    type="checkbox"
+                  />
+                  Active
+                </label>
+              </td>
+              <td>
+                <label htmlFor="stats_completeness">
```
eatyourgreens: This label wraps the input that it labels, so the `htmlFor` isn't needed.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+                    name={`stats_completeness_type.${workflow.id}`}
+                    onChange={e => handleSetStatsCompletenessType(e, meta.page, workflow)}
+                    type="radio"
+                    value="classification"
+                  />
+                  Classification Count
+                </label>
+                {' '}
+                <label htmlFor="stats_completeness">
+                  <input
```
eatyourgreens: Ditto regarding `htmlFor` and `id` here.

app/pages/lab/workflows.spec.js
eatyourgreens: That probably means that some of the tests should be moved to the tests for `WorkflowsTable`.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,119 @@
...
+          <th>Progress</th>
+          <th>Status</th>
+          <th>Completeness statistic</th>
+          <th>Statistic visibility</th>
+        </tr>
+      </thead>
+      <tbody>
+        {workflows.map((workflow) => {
+          const progressPercentage = workflow.completeness * 100;
+          const statsCompletenessType = (workflow.configuration) ? workflow.configuration.stats_completeness_type : 'retirement';
```
eatyourgreens: If `workflow.configuration` exists, this will default `statsCompletenessType` to `undefined`, not `retirement`.

app/pages/lab/workflows.spec.js
simoneduca: That's a good idea, I'll do that.

simoneduca:
I'm adding tests for all the new settings.

simoneduca:
This should be ready to go.

app/pages/lab/workflows-container.jsx
```diff
@@ -50,6 +53,33 @@ export default class WorkflowsContainer extends React.Component {
...
   }
 
+  handleSetStatsCompletenessType(e, page, workflow) {
+    workflow.update({ 'configuration.stats_completeness_type': e.target.value }).save()
+      .catch(error => console.log(error))
+      .then(() => this.getWorkflowList(page));
+  }
+
+  handleWorkflowStatusChange(e, page, workflow) {
+    const defaultWorkflow = (this.props.configuration) ? this.props.configuration.default_workflow : null;
```
srallen: I think this should be `const defaultWorkflow = (this.props.configuration && this.props.configuration.default_workflow) ? this.props.configuration.default_workflow : null;`

Also, could the prop be renamed projectConfiguration for clarity? I had to read this line a couple of times because I thought initially it was referring to the workflow configuration object.

app/pages/lab/workflows-table.spec.js
```diff
@@ -0,0 +1,69 @@
...
+    const statusCheckbox = wrapper.find('input[id="active"]').at(0);
+    statusCheckbox.simulate('change');
+    expect(statusChangeSpy.calledOnce).to.be.true;
+  });
+
+  it('should call handleSetStatsCompletenessType on completeness-type-classification change', function () {
+    const completenessRadioButton = wrapper.find('input[value="classification"]').at(0);
+    completenessRadioButton.simulate('change');
+    expect(completenessTypeChangeSpy.calledOnce).to.be.true;
+    completenessTypeChangeSpy.reset();
```
srallen: I'm seeing a warning about this being deprecated in the next sinon version. Could this be done in another way?

app/pages/lab/workflows.spec.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react';
 import assert from 'assert';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
```
srallen: mount isn't being used in these tests.

app/pages/lab/workflows-table.spec.js
```diff
@@ -0,0 +1,69 @@
...
+  });
+
+  it('should call handleWorkflowStatusChange on status checkbox change', function () {
+    const statusCheckbox = wrapper.find('input[id="active"]').at(0);
+    statusCheckbox.simulate('change');
+    expect(statusChangeSpy.calledOnce).to.be.true;
+  });
+
+  it('should call handleSetStatsCompletenessType on completeness-type-classification change', function () {
+    const completenessRadioButton = wrapper.find('input[value="classification"]').at(0);
```
srallen: Shouldn't this return a single wrapper? There's only one DOM element that has `value="classification"` so calling `at()` shouldn't be necessary here.

app/pages/lab/workflows-table.spec.js
```diff
@@ -0,0 +1,69 @@
...
+      />
+    );
+  });
+
+  it('should display the correct amount of workflows', function () {
+    expect(wrapper.find('Link')).to.have.lengthOf(2);
+  });
+
+  it('should call handleWorkflowStatusChange on status checkbox change', function () {
+    const statusCheckbox = wrapper.find('input[id="active"]').at(0);
```
srallen: Same comment about unnecessary use of `at()`

app/pages/lab/workflows-table.spec.js
```diff
@@ -0,0 +1,69 @@
...
+
+  it('should call handleSetStatsCompletenessType on completeness-type-classification change', function () {
+    const completenessRadioButton = wrapper.find('input[value="classification"]').at(0);
+    completenessRadioButton.simulate('change');
+    expect(completenessTypeChangeSpy.calledOnce).to.be.true;
+    completenessTypeChangeSpy.reset();
+  });
+
+  it('should call handleSetStatsCompletenessType on completeness-type-retirement change', function () {
+    const completenessRadioButton = wrapper.find('input[value="retirement"]').at(1);
```
srallen: Same comment about `at()`

app/pages/lab/workflows-table.jsx
```diff
@@ -45,11 +45,11 @@ const WorkflowsTable = ({
                 {`${progressPercentage.toFixed(0)} % Complete`}
               </td>
               <td>
-                <label htmlFor="active">
+                <label>
                   <input
                     checked={workflow.active}
                     id="active"
```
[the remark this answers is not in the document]
simoneduca: That's how I find this checkbox in the test. Can you suggest a different way?

app/pages/lab/workflows-table.spec.js
simoneduca: It doesn't return a single wrapper. Removing `at(0)` results in `Error: Method “simulate” is only meant to be run on a single node. 2 found instead.` because there is a node with `value="classification"` for every workflow (2 in the test).

app/pages/lab/workflows-table.spec.js
eatyourgreens: Only one element can have a given `id` in the DOM, so your HTML is invalid if this find() is returning more than one checkbox.

I think I mentioned in an earlier PR comment that you should remove these `id` attributes.

app/pages/lab/workflows-container.jsx
```diff
@@ -50,6 +53,34 @@ export default class WorkflowsContainer extends React.Component {
...
   }
 
+  handleSetStatsCompletenessType(e, page, workflow) {
+    workflow.update({ 'configuration.stats_completeness_type': e.target.value }).save()
+      .catch(error => console.log(error))
+      .then(() => this.getWorkflowList(page));
+  }
+
+  handleWorkflowStatusChange(e, page, workflow) {
+    const projectConfiguration = (this.props.configuration && this.props.configuration.default_workflow) ?
```
eatyourgreens: This is storing the default workflow as `projectConfiguration`. I think @srallen was asking you to rename the `configuration` prop to `projectConfiguration` in order to distinguish it from the workflow config.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,116 @@
...
+              <td>
+                {`${progressPercentage.toFixed(0)} % Complete`}
+              </td>
+              <td>
+                <label>
+                  <input
+                    checked={workflow.active}
+                    id="active"
+                    onChange={e => handleWorkflowStatusChange(e, meta.page, workflow)}
+                    type="checkbox"
```
eatyourgreens: This checkbox is missing a name and value.

app/pages/lab/workflows-table.jsx
```diff
@@ -0,0 +1,116 @@
...
+                </Link>
+              </td>
+              <td>
+                {`${progressPercentage.toFixed(0)} % Complete`}
+              </td>
+              <td>
+                <label>
+                  <input
+                    checked={workflow.active}
+                    id="active"
```
eatyourgreens: This ID is invalid HTML and not needed.

app/pages/lab/workflows-table.spec.js
eatyourgreens: You could search by name and value if you want to test the button for a particular workflow, but I'm not sure if that makes a difference for these tests.

simoneduca:
I think I got everything now. Thanks for the comments Sarah and Jim!

PR #4333, opened 2018-02-02
Allow survey tasks to clear radio button answers

eatyourgreens:
Staging branch URL: https://survey-radio-toggle.pfe-preview.zooniverse.org

Fixes a bug reported by Snapshot Wisconsin: https://www.zooniverse.org/talk/17/535525

Adds handlers to the survey choice radio buttons so that a selected answer can be cleared on click, space or backspace.

Removes some old code which cleared the current answer but was never actually called.

Removes named refs from the `Choice` component.

# Required Manual Testing

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, ~~Chrome~~, ~~Edge~~, ~~Safari~~?
- [x] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/survey/choice.jsx
+++ b/app/classifier/tasks/survey/choice.jsx
@@ -177,6 +177,7 @@ class Choice extends React.Component {
                         >
                           <input
                             name={questionId}
+                            value={answerId}
                             type={inputType}
                             autoFocus={!hasFocus && i === 0}
                             checked={isChecked}
--- a/app/classifier/tasks/survey/choice.jsx
+++ b/app/classifier/tasks/survey/choice.jsx
@@ -6,6 +6,9 @@ import Translate from 'react-translate-component';
 import ImageFlipper from './image-flipper';
 import Utility from './utility';
 
+const BACKSPACE = 8;
+const SPACE = 32;
+
 
 class Choice extends React.Component {
   constructor(props) {
@@ -71,6 +74,19 @@ class Choice extends React.Component {
 
   handleIdentification() {
     this.props.onConfirm(this.props.choiceID, this.state.answers);
+  }
+
+  handleRadioKeyDown(questionId, answerId, e) {
+    switch (e.which) {
+      case BACKSPACE:
+      case SPACE:
+        if (e.target.checked) {
+          e.preventDefault();
+          this.handleRadioToggle(questionId, answerId, e);
+        }
+        break;
+      default:
+    }
   }
 
   handleRadioToggle(questionId, answerId, e) {
@@ -172,6 +188,7 @@ class Choice extends React.Component {
                             checked={isChecked}
                             onChange={this.handleAnswer.bind(this, questionId, answerId)}
                             onClick={this.handleRadioToggle.bind(this, questionId, answerId)}
+                            onKeyDown={this.handleRadioKeyDown.bind(this, questionId, answerId)}
                             onFocus={this.handleFocus.bind(this, questionId, answerId)}
                             onBlur={this.handleFocus.bind(this, null, null)}
                           />
--- a/app/classifier/tasks/survey/choice.jsx
+++ b/app/classifier/tasks/survey/choice.jsx
@@ -73,6 +73,15 @@ class Choice extends React.Component {
     this.props.onConfirm(this.props.choiceID, this.state.answers);
   }
 
+  handleRadioToggle(questionId, answerId, e) {
+    const { type } = e.target;
+    const { answers } = this.state;
+    if (type === 'radio' && answers[questionId] === answerId) {
+      delete answers[questionId];
+      this.setState({ answers });
+    }
+  }
+
   render() {
     const { choiceID, task, translation } = this.props;
     const choice = task.choices[this.props.choiceID];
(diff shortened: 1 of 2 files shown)


eatyourgreens:
I'm not sure how well `onClick` is supported for radio inputs, so this could do with some cross-browser testing, particularly on Windows.

eatyourgreens:
I've had a quick look at this with IE11 on BrowserStack, and was able to toggle the radio button answers with mouse click or space, and clear them with backspace.

mcbouslog:
Code looks good to me, I'll test on a few projects and report back with final review.

simoneduca:
I'm not sure what I'm supposed to test here. Mind clarifying? Cheers.

mcbouslog:
I think currently if you click an animal, then click a number for how many, you can't unclick it.
https://master.pfe-preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify

But this PR fixes it so you can.
https://survey-radio-toggle.pfe-preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify

app/classifier/tasks/survey/choice.jsx
```diff
@@ -73,6 +72,28 @@ class Choice extends React.Component {
...
+        }
+        break;
+      default:
+    }
+  }
+
+  resetSingleAnswerQuestion(e) {
+    const { type, name, value } = e.target;
+    const { answers } = this.state;
+    if (type === 'radio' && answers[name] === value) {
```
mcbouslog: I think this might need a second conditional where
```
else if (type === 'radio' && answers.name.includes(value)) {
  const newValues = answers.name.filter(item => item !== value);
  answers.name = newValues;
  this.setState({ answers });
}
```

or something similar to handle a question that supports multiple answers like the second question (What behaviors do you see?) in this test workflow:
https://survey-radio-toggle.pfe-preview.zooniverse.org/projects/markb-panoptes/survey-task-test-project/classify

eatyourgreens: I can toggle the checkbox answers without any extra code. There's already code to splice the answers array in `handleAnswers`, I think.

mcbouslog: nope the code noted above isn't fixing it, dang, I'll keep trying a bit

eatyourgreens:
@simoneduca the linked Talk discussion describes the bug on Snapshot Wisconsin. https://www.zooniverse.org/talk/17/535525

They used to be able to clear radio button answers by clicking on them.

app/classifier/tasks/survey/choice.jsx
eatyourgreens: Verified on your test workflow that I can toggle the checkboxes on and off, and the final classification reflects the checked choices.

eatyourgreens: NB Radio buttons will only allow you to select one answer anyway, so `answers[name]` will never be an array for `type=radio`.

mcbouslog: I don't think I can toggle answers for a question that accepts multiple answers with the spacebar on this branch, but can on master. Works fine with the mouse on both branches.

eatyourgreens: Ah right, in that case the error is in the `keyDown` handler, not this function.

app/classifier/tasks/survey/choice.jsx
```diff
@@ -73,6 +72,28 @@ class Choice extends React.Component {
     this.props.onConfirm(this.props.choiceID, this.state.answers);
   }
 
+  handleRadioKeyDown(e) {
+    switch (e.which) {
+      case BACKSPACE:
+      case SPACE:
+        if (e.target.checked) {
+          e.preventDefault();
```
eatyourgreens: This is going to prevent the default action for checkboxes as well as radio buttons, so they can't be toggled off with space.

app/classifier/tasks/survey/choice.jsx
eatyourgreens: I've added a comment to remind myself to fix this.

mcbouslog:
Chrome, Edge and Safari work well for me.

Firefox was strange where if I attempted to use space to deselect an answer it would flash for a second like it was properly deselecting, but then it would stay selected.

app/classifier/tasks/survey/choice.jsx
mcbouslog: yep, adding ` && (e.target.type === 'radio')` to the conditional appears to fix it

eatyourgreens:
Space selects a radio button in all browsers. Sounds like Firefox is ignoring the `preventDefault` and performing the default anyway. `stopPropagation` might fix it. It may just be that Firefox doesn’t let you cancel default behaviour on form elements.

eatyourgreens:
@mcbouslog @simoneduca d08500c should fix the space key for checkboxes.

eatyourgreens:
Firefox is firing a change event, calling `handleAnswer`, when the radio button updates in response to the state change here:
https://github.com/zooniverse/Panoptes-Front-End/blob/d08500cede4ea24240fe53f908ad8dfc37fff1f1/app/classifier/tasks/survey/choice.jsx#L93-L96

PR #4346, opened 2018-02-08
Fix stuck tutorial bug #4343

### LINKED ISSUE #4343: Unable to close project tutorial
## Expected behavior
The close button, or "Let's go!", should dismiss the tutorial when I visit a new project.

## Current behavior
Closing or completing the tutorial throws an error because `this.props.preferences` is null in the Tutorial component at line 131.
https://github.com/zooniverse/Panoptes-Front-End/blob/48d4db306e2e9eb69e70ff0e86476266080e124c/app/classifier/tutorial.cjsx#L131-L136

## Steps to replicate
Visit a project that you have never visited before ie. have no user project preferences for, and try to exit the tutorial.

The bug is a little tricky to replicate, because it only triggers on the very first visit to any new project. On the second and subsequent visits, you have a UPP defined so the bug doesn't trigger.

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Staging branch URL:

Fixes #4343.

Fixes a bug in the project page controller where `fetchProjectData` is called twice on component mount, which leads to user preferences being created for a new project, but then overwritten with `undefined`.
- [x] - Set out expectations for the project page controller.
- [ ] - Write actual tests to isolate the bug.
- [ ] - Fix the bug.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/index.spec.js
+++ b/app/pages/project/index.spec.js
@@ -1,0 +1,36 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import assert from 'assert';
+import sinon from 'sinon';
+import ProjectPageController from './';
+
+describe('ProjectPageController', function () {
+  describe('with initial load complete', function () {
+    it('should fetch project data on mount.', function () {
+    
+    });
+    it('should fetch project data again on project change.', function () {
+      
+    });
+    it('should fetch project data again on user change.', function () {
+      
+    });
+    it('should not fetch project data again on any other prop change.', function () {
+      
+    });
+  });
+  describe('without initial load complete', function () {
+    it('should not fetch project data on mount.', function () {
+    
+    });
+    it('should fetch project data again on project change.', function () {
+      
+    });
+    it('should not fetch project data again on user change.', function () {
+      
+    });
+    it('should not fetch project data again on any other prop change.', function () {
+      
+    });
+  })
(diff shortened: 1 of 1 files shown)


eatyourgreens:
61c9774 attempts to reproduce the bug. When you enter a project page for the first time, `fetchProjectData` is being called twice because `componentWillReceiveProps` is running before the first fetch has set `state.loading` to true.

app/pages/project/index.spec.js
```diff
@@ -0,0 +1,164 @@
...
+    
+    afterEach(function () {
+      apiRequestStub.resetHistory();
+    })
+    
+    it('should not fetch project data on mount.', function () {
+      sinon.assert.notCalled(apiRequestStub);
+    });
+    
+    it('should fetch project data again on project change.', function () {
```
eatyourgreens: Is this expectation correct when `initialLoadComplete` is false? It follows the logic in `componentWillReceiveProps` but it doesn't seem right to me, at least, that the project data should ever be fetched before initial log-in has completed.

srallen: Hm, that doesn't seem right to me either. We don't want any of the data loaded until we know if there's a user or not.

PR #4357, opened 2018-02-14
Edit org roles request to paginate through all roles

mcbouslog:
Staging branch URL: https://all-org-roles.pfe-preview.zooniverse.org/
- demo project: https://all-org-roles.pfe-preview.zooniverse.org/organizations/markb-panoptes/nfnorgtest

To test I edited the org roles request (new line 173) to include `page_size: 2`. Unsure of how to write related actual tests.

Added code to paginate through org role responses to ensure all roles are collected for checking if user is an org collaborator.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/organization/organization-container.jsx
+++ b/app/pages/organization/organization-container.jsx
@@ -26,6 +26,7 @@ class OrganizationContainer extends React.Component {
       quoteObject: {}
     };
 
+    this.fetchAllOrganizationRoles = this.fetchAllOrganizationRoles.bind(this);
     this.toggleCollaboratorView = this.toggleCollaboratorView.bind(this);
     this.updateQuery = this.updateQuery.bind(this);
   }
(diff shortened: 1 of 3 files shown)


app/pages/organization/organization-container.jsx
```diff
@@ -26,6 +26,7 @@ class OrganizationContainer extends React.Component {
       quoteObject: {}
     };
 
+    this.fetchAllOrganizationRoles = this.fetchAllOrganizationRoles.bind(this);
```
srallen: I don't think this binding is necessary this class method isn't an event handler.

mcbouslog: without binding `fetchAllOrganizationRoles` in the `constructor` or alternatively and maybe preferable in the function itself when it's set as a `const` I get an error because the `organizationRoles` becomes undefined.

srallen: I'm not sure why binding would affect default parameter in a class method. What happens if you instead remove the default and define it in inside the method?:
```
fetchAllOrganizationRoles(organization, organizationRoles, _page) {
  const organizationRoles = organizationRoles || [];
  const _page = _page || 1;
  // other stuff
}
```

mcbouslog: didn't realize I don't need to define the method as a `const` within itself, I think because I'm returning the promise, so just using `this.fetchAllOrganizationRoles` to continue to string promises works as desired. but not entirely sure this is the best solution, let me know what you think.

mcbouslog:
please hold on merging until I've added tests as noted

mcbouslog:
please hold on refactor to utilize `meta.next_page`

mcbouslog:
I think this is ready for another review now, let me know what you think.

PR #4362, opened 2018-02-15
Use project instead of workflows in Apply For Beta form

### LINKED ISSUE #4360: Cannot check the "Project has at least one active workflow" box in the project builder.
## Expected behaviour
The project builder should recognise that an active workflow is linked to the project under the  Visibility menu.

## Current behaviour
I am trying to prepare Muon Hunters 2.0 for beta launch but cannot click the "Apply for review" button under the Visibility menu.  I cannot check the "Project has at least one active workflow" box despite having one active workflow linked.

screen shot 2018-02-15 at 09 34 26

screen shot 2018-02-15 at 09 34 45

## Additional information
The project has a subject set with 100 subjects:

screen shot 2018-02-15 at 09 35 06

screen shot 2018-02-15 at 09 54 00

The project has an FAQ:

screen shot 2018-02-15 at 09 35 56

- **Operating system:**
OSX 10.11.6 

- **Browser:**
Chrome 64.0.3282.140

simoneduca:
Staging branch URL: https://visibility-page.pfe-preview.zooniverse.org/

Fixes #4360 .

Describe your changes.
Get active workflows and subject count from project instead of workflows, which isn't available anymore.
I'm not sure this is working properly. I'm seeing some strange behaviour when I pass from no active workflows to even just one active. When I check the visibility page the checkbox takes ages to update or doesn't update at all, even if the project does have `active_workflows`. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/apply-for-beta-form.jsx
+++ b/app/pages/lab/apply-for-beta-form.jsx
@@ -9,7 +9,7 @@ const REQUIRED_PAGES = ['Research', 'FAQ'];
 
 // Static functions
 const projectHasActiveWorkflows = (project) => {
-  if (project.links.active_workflows) {
+  if (project.links.active_workflows && project.links.active_workflows.length > 0) {
     return true;
   }
   return false;
@@ -93,6 +93,7 @@ class ApplyForBetaForm extends React.Component {
     this.updateValidationsFromProps = this.updateValidationsFromProps.bind(this);
 
     this.state = {
+      links: {},
       validations: {
         projectIsPublic: projectIsPublic(props.project),
         projectIsLive: projectIsLive(props.project),
@@ -105,6 +106,7 @@ class ApplyForBetaForm extends React.Component {
       doingAsyncValidation: false
     };
   }
+
 
   componentWillUpdate(nextProps) {
     this.updateValidationsFromProps(nextProps);
--- a/app/pages/lab/apply-for-beta-form.jsx
+++ b/app/pages/lab/apply-for-beta-form.jsx
@@ -8,28 +8,17 @@ const MINIMUM_SUBJECT_COUNT = 100;
 const REQUIRED_PAGES = ['Research', 'FAQ'];
 
 // Static functions
-const projectHasActiveWorkflows = (workflows) => {
-  return workflows.some((workflow) => {
-    return workflow.active;
-  });
+const projectHasActiveWorkflows = (project) => {
+  if (project.links.active_workflows) {
+    return true;
+  }
+  return false;
 };
 
-const projectHasMinimumActiveSubjects = (workflows) => {
-  const activeWorkflows = workflows.filter((workflow) => {
-    return workflow.active;
-  });
-  const uniqueSetIDs = activeWorkflows.reduce((sets, workflow) => {
-    return uniq(sets.concat(workflow.links.subject_sets));
-  }, []);
-  // Second parameter is an empty object to prevent request caching.
-  return apiClient.type('subject_sets', {})
-    .get(uniqueSetIDs)
-    .then((sets) => {
-      const subjectCount = sets.reduce((count, set) => {
-        return count + set.set_member_subjects_count;
-      }, 0);
-      return (subjectCount >= MINIMUM_SUBJECT_COUNT) ? true : `The project only has ${subjectCount} of ${MINIMUM_SUBJECT_COUNT} required subjects`;
-    });
+const projectHasMinimumActiveSubjects = (project) => {
+  const subjectCount = project.subjects_count;
+  return (subjectCount >= MINIMUM_SUBJECT_COUNT) ?
+    true : `The project only has ${subjectCount} of ${MINIMUM_SUBJECT_COUNT} required subjects`;
 };
 
 const projectHasRequiredContent = (project) => {
@@ -107,7 +96,7 @@ class ApplyForBetaForm extends React.Component {
       validations: {
         projectIsPublic: projectIsPublic(props.project),
         projectIsLive: projectIsLive(props.project),
-        projectHasActiveWorkflows: projectHasActiveWorkflows(props.workflows),
+        projectHasActiveWorkflows: projectHasActiveWorkflows(props.project),
         labPolicyReviewed: false,
         bestPracticesReviewed: false,
         feedbackReviewed: false
@@ -126,7 +115,7 @@ class ApplyForBetaForm extends React.Component {
     // error messages.
     this.setState({ doingAsyncValidation: true });
     return Promise.all([
-      projectHasMinimumActiveSubjects(this.props.workflows),
+      projectHasMinimumActiveSubjects(this.props.project),
       projectHasRequiredContent(this.props.project)
     ])
     .catch((error) => {
@@ -194,6 +183,7 @@ class ApplyForBetaForm extends React.Component {
   }
 
   updateValidationsFromProps(props) {
+
     // We need to do a props comparison, otherwise we get a loop where props
     // update state -> updates state repeatedly.
     //
@@ -206,7 +196,7 @@ class ApplyForBetaForm extends React.Component {
     const newValues = {
       projectIsPublic: projectIsPublic(props.project),
       projectIsLive: projectIsLive(props.project),
-      projectHasActiveWorkflows: projectHasActiveWorkflows(props.workflows)
+      projectHasActiveWorkflows: projectHasActiveWorkflows(props.project)
     };
 
     Object.keys(newValues).forEach((key) => {
@@ -223,7 +213,6 @@ class ApplyForBetaForm extends React.Component {
   render() {
     const applyButtonDisabled = !this.canApplyForReview() ||
       this.state.doingAsyncValidation;
-
     return (
       <div>
 
@@ -264,7 +253,6 @@ class ApplyForBetaForm extends React.Component {
 
 ApplyForBetaForm.defaultProps = {
   project: {},
-  workflows: [],
   applyFn: Function.prototype
 };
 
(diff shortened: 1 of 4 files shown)


simoneduca:
I'm looking at fixing the tests.

eatyourgreens:
Re. The delays, are you waiting on a Panoptes background worker to update the project after you change the active workflows. I don’t really understand why you’re seeing a delay either.

app/pages/lab/apply-for-beta-form.jsx
```diff
@@ -8,28 +8,17 @@ const MINIMUM_SUBJECT_COUNT = 100;
 const REQUIRED_PAGES = ['Research', 'FAQ'];
 
 // Static functions
-const projectHasActiveWorkflows = (workflows) => {
-  return workflows.some((workflow) => {
-    return workflow.active;
-  });
+const projectHasActiveWorkflows = (project) => {
+  if (project.links.active_workflows) {
```
eatyourgreens: I think this will be true if `project.links.active_workflows` is an empty array.

simoneduca: AFAIK, the `active_workflows` field gets added to `links` only if its `length > 0`.

eatyourgreens: In that case, could you explicitly check for `project.links.active_workflows.length > 0` here?

app/pages/lab/apply-for-beta-form.jsx
```diff
@@ -8,28 +8,17 @@ const MINIMUM_SUBJECT_COUNT = 100;
...
-    .then((sets) => {
-      const subjectCount = sets.reduce((count, set) => {
-        return count + set.set_member_subjects_count;
-      }, 0);
-      return (subjectCount >= MINIMUM_SUBJECT_COUNT) ? true : `The project only has ${subjectCount} of ${MINIMUM_SUBJECT_COUNT} required subjects`;
-    });
+const projectHasMinimumActiveSubjects = (project) => {
+  const subjectCount = project.subjects_count;
+  return (subjectCount >= MINIMUM_SUBJECT_COUNT) ?
+    true : `The project only has ${subjectCount} of ${MINIMUM_SUBJECT_COUNT} required subjects`;
```
eatyourgreens: Not a blocking comment, since this was in the code already, but do you know why this returns a string if false?

eatyourgreens:
When I log `project.links.active_workflows` in the `WorkflowsTable` component, it doesn't change after I tick or untick the Active box for a workflow. I think that's why you're validation checks aren't changing. Do you update the project after changing a workflow's active status?

eatyourgreens:
Also, if I refresh the workflows page immediately after unchecking a couple of workflows,  `project.links.active_workflows` hasn't changed. I don't think that array updates instantly in Panoptes, after you change a workflow.

app/pages/lab/apply-for-beta-form.jsx
simoneduca: Wouldn't that throw an error when `active_workflows` is `undefined`?

simoneduca:
Yeah, that should be the reasons for the "delays" I'm seeing.

app/pages/lab/apply-for-beta-form.jsx
eatyourgreens: Then check for `project.links.active_workflows && project.links.active_workflows.length > 0`.

simoneduca:
RE. updating the project, good point. I think I can update it [here](https://github.com/zooniverse/Panoptes-Front-End/blob/5e80550403d6025fbbee69fa0a9043b5565bfbcd/app/pages/lab/workflows-container.jsx#L51) and then it should bubble up to the workflows table

eatyourgreens:
`project` is a prop there, and you shouldn't change props. You might need to update the project higher up the tree, so that it's changed in state quite high up then passed down to the Workflows page and the Visibility page.

The client has `addLink` and `removeLink` methods which will probably let you add and remove active workflows from the project itself. There's a little bit of documentation in the JSON client [Readme](/zooniverse/json-api-client/blob/master/README.md).

eatyourgreens:
Of course, it may simply be quicker to pass the updated workflows to the Visibility page, and use the original code to validate them for beta.

simoneduca:
Not sure how I would do that? I've tried `componentWillReceiveProps`, on the applyforbeta component but it's not working.

eatyourgreens:
Have you updated the project in the parent component, so that a new project prop is passed down to ApplyForBeta? `componentWillReceiveProps` won't run until the parent component passes in new props.

srallen:
There could be a callback passed down via props to be called in the then of the update for the workflow activeness that could request for the project resource and set it to the parent state of both the visibility component and the workflows page component? 

I'm taking about here along with the req to get the workflows: 

https://github.com/zooniverse/Panoptes-Front-End/blob/5e80550403d6025fbbee69fa0a9043b5565bfbcd/app/pages/lab/workflows-container.jsx#L74

simoneduca:
@eatyourgreens sorry I meant I've tried `componentWillReceiveProps` on `visibility.cjsx`, not applyforbeta page.
@srallen 
>that could request for the project resource and set it to the parent state of both the visibility component and the workflows page component
I don't understand that bit. Could you clarify please?

eatyourgreens:
@simoneduca ok, did you update the prop in the parent component of `visibility.cjsx`? `ComponentWillReceiveProps` is called when new props are passed down from a parent component.

simoneduca:
I can log `componentWillReceiveProps()` and check that it's indeed called on `visibility.cjsx`. Which one is the parent component of `visibility.cjsx`? I can only seen it being required here https://github.com/zooniverse/Panoptes-Front-End/blob/5e80550403d6025fbbee69fa0a9043b5565bfbcd/app/router.cjsx#L233

srallen:
@simoneduca project.cjsx in the lab folder is the parent of all the project edit pages, which you can find using the react dev tools and investigating the component tree. Looking at those, these are some of the last remaining components using the PromiseRenderer, so properly using state and callbacks here will be more involved since it'll involve refactoring the parent.

simoneduca:
D'oh, thanks @srallen. I'd be happy to refactor the project-page, but that seems out of the scope of this PR.

app/pages/lab/apply-for-beta-form.jsx
simoneduca: No idea, sorry.

srallen:
How often are people making a workflow active and then immediately going to the visibility page to request a beta? I'm going to guess this is fairly uncommon and I'd rather us not add back code that is basically cruft and move forward instead with a refactor of the parent component in a separate PR.

eatyourgreens:
Good point. I'm happy to stick with using the project as-is.

simoneduca:
I'm slightly confused. As far as I understand, we've gone full 360 on this one, which means that no change is needed and I can close this PR?

srallen:
That’s not what I’m saying. There are still bug fixes coming in with this PR because not all of the cruft with removing the api call for the workflows was removed previously. What I’m saying is that we should properly fix the parent component rather than adding back the API calls for the workflows since making a workflow active and immediately going to request a beta will be rare user behavior.

simoneduca:
>we should properly fix the parent component rather than adding back the API calls for the workflows
@srallen totally agree
>making a workflow active and immediately going to request a beta will be rare user behavior
 I'm not sure how rare that behaviour but it seems good practice to let any state changes propagate and [adding a proper link as suggested here](https://github.com/zooniverse/Panoptes/issues/2679#issuecomment-366728772) would allow that. I agree also that that should be done in the context of fixing the parent component.
So, how about I'll finish this PR with removing all the workflow cruft so that we have a less-than-ideal fix and then tackle `project.cjsx`?

eatyourgreens:
Resources have an `uncache` method, so you can call `project.uncache()` to remove stale project data from the client. https://github.com/zooniverse/json-api-client/blob/48426c58de25073142878d81f69ea5383f7f10c2/src/resource.coffee#L86-L92

app/pages/lab/apply-for-beta-form.jsx
```diff
@@ -104,10 +93,11 @@ class ApplyForBetaForm extends React.Component {
     this.updateValidationsFromProps = this.updateValidationsFromProps.bind(this);
 
     this.state = {
+      links: {},
```
eatyourgreens: I can't see this being used anywhere in this component.

simoneduca:
Where do you think `uncache` would be necessary?

eatyourgreens:
After you've changed a workflow's active status, invalidating the current project.

app/pages/lab/apply-for-beta-form.jsx
```diff
@@ -8,28 +8,17 @@ const MINIMUM_SUBJECT_COUNT = 100;
...
 // Static functions
-const projectHasActiveWorkflows = (workflows) => {
-  return workflows.some((workflow) => {
-    return workflow.active;
-  });
+const projectHasActiveWorkflows = (project) => {
+  if (project.links.active_workflows && project.links.active_workflows.length > 0) {
+    return true;
+  }
+  return false;
```
srallen: This function could just `return (project.links.active_workflows && project.links.active_workflows.length > 0)`

simoneduca: Done, thanks.

simoneduca:
Anything else I missed?

app/pages/lab/workflows-container.jsx
```diff
@@ -66,6 +66,7 @@ export default class WorkflowsContainer extends React.Component {
     workflow.update({ active: checked }).save()
       .catch(error => console.log(error))
       .then(() => {
+        this.props.project.uncache();
```
eatyourgreens: Does uncaching the project here cause any problems updating the resource in the next couple of lines? I don't think it will but just checking.

simoneduca: No, as far as I am aware.

eatyourgreens: 👍

Issue #4358, opened 2018-02-15
PFE wrongly assumes that all apps need a `#`


simoneduca:
## Expected behavior
Being redirected to the correct page
## Current behavior
User sees the home page (host address), regardless of what the fragment identifier specifies. E.g. https://www.antislaverymanuscripts.org/#/classify just brings up the site homepage.

## Steps to replicate

1. Go to https://www.antislaverymanuscripts.org
2. Click on the project's talk link
3. Click on the classify link

## Additional information
The culprit is this [line of code](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/components/ProjectNavbar/helpers/getProjectLinks.js#L50), which wrongly assumes that all apps need a `#`.

##Suggestion to fix this

1. Remove `#` from the line above.
2. Add `#`, or `#!` for our angular apps, to the project's `redirect` field as needed.

eatyourgreens:
Protip: Github lets you quote code directly in your issues. Here's that line of code: https://github.com/zooniverse/Panoptes-Front-End/blob/5e80550403d6025fbbee69fa0a9043b5565bfbcd/app/pages/project/components/ProjectNavbar/helpers/getProjectLinks.js#L50

simoneduca:
Just realised the link above is for the new navbar.
Here's the code for the current navbar https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/project-navbar.jsx#L118

eatyourgreens:
Odd that any existing hash in `project.redirect` is stripped out.

PR #4366, opened 2018-02-16
Make new navbar public

srallen:
Staging branch URL: https://make-new-navbar-public.pfe-preview.zooniverse.org/

This makes the new project navbar public. I also did some cruft clean up and test refactoring for the project-page and project-home components. One thing I cleaned up was the double import of the project navbar and the project announcements in the project-home component and in the parent project-page component and then the conditional in the project-page to make sure it didn't render twice. I really don't understand why that was being done, but with the new project navbar, I'm considering it cruft. With that clean up, I refactored the associated tests. There were also a couple of tests for the project-page component that were testing the very deeply nested children or the position of components which just seems like testing React to me, which in my opinion, we shouldn't be writing tests for. 

Now, the tests aren't passing on Travis. They are passing locally. I added some console logs, which you can see in the Travis output to show that the failing tests are because the expected single call of several spies are being called multiple times. I'm honestly not sure why. @eatyourgreens do you have an idea why this might happen?

@simoneduca can we coordinate this with #4358 so we don't have merge conflict and we also shouldn't be doing any more work on the old project navbar since it's being removed.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


srallen:
@beckyrother when this is ready to merge, we can coordinate announcing the new navigation to the volunteers.

app/pages/project/project-page.jsx
```diff
@@ -111,7 +84,11 @@ export default class ProjectPage extends React.Component {
           user: this.props.user,
           workflow: this.props.workflow
         })}
-        {potentialFieldGuide}
+        {(this.props.location.pathname !== `/projects/${this.props.project.slug}`) &&
```
eatyourgreens: This doesn’t check whether we’re actually on the home page. The `onHomePage` check from the old code is more reliable.

srallen: Ok, good to know. It was confusing because that check was only being used with the navbar and I'm not sure if that worked or why the navbar was being added in both components.

eatyourgreens: From memory, the string comparison misses trailing slashes but the router path check will catch those.

eatyourgreens:
I don't think any of these function bindings are needed, since all of these methods are bound to the class context on creation. They might be interfering with the spies, since `bind` returns a new function. I'm not convinced that's the problem here, but it's worth deleting these lines if they do nothing. https://github.com/zooniverse/Panoptes-Front-End/blob/9adda8319dd80e623d4c8823a2e44bd7b3fa80e9/app/pages/project/components/ProjectNavbar/ProjectNavbarContainer.jsx#L14-L18

eatyourgreens:
Actually, `handleClientChange` should be bound, since it's an event handler. The others can probably go.

eatyourgreens:
More unnecessary function binding here too. This line can go.
https://github.com/zooniverse/Panoptes-Front-End/blob/9adda8319dd80e623d4c8823a2e44bd7b3fa80e9/app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx#L22

eatyourgreens:
Looks like removing those `bind` calls didn't fix the spies. I'm out of ideas for the moment. Tests are passing locally for me too.

app/pages/project/project-page.jsx
eatyourgreens: There might be a better check where we look at `props.children` to see what type of component the project page is, but I'm not sure how that would work in practice.

srallen:
@eatyourgreens thanks for taking a look. I'm wondering if it has to do with how sinon setups the spies and if those spies are being added to the constructor of a component from the very start of the test suite. 🤷‍♀️ I'll take another look today.

app/pages/project/project-page.jsx
eatyourgreens: `React.Children.only(this.props.children).type.name === 'ProjectHomeContainer'` might work, using [Function.name](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/name) to get the only child's name.

eatyourgreens:
I've tried explictly running the Travis script locally, but that passes too. 😕

srallen:
@eatyourgreens I got the tests working by calling `resetHistory` in the before block of the tests that except those class methods to only be called a certain number of times. This doesn't really fix why this is happening, but gets the tests to pass. 

I mentioned the background image doubling to @beckyrother and that will be addressed in a follow up PR. This should be ready for review now.

eatyourgreens:
Looks good. I’ll have another look tomorrow morning. Posting this link so I can test language switching. It should show the nav links in Italian. https://make-new-navbar-public.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground?language=it

beckyrother:
Please @ me before this is merged so we can set up a Talk post when the changes go live. Want to make sure we're transparent about the changes and give people a place to give feedback – I'm asking that people either respond to the (upcoming) Talk post, DM me, or email me directly.

eatyourgreens:
@beckyrother I’m happy to assign this PR to you, so that you have final say on pressing the big green button.

app/pages/project/components/ProjectNavbar/ProjectNavbar.spec.js
```diff
@@ -41,23 +43,27 @@ describe('ProjectNavbar', function() {
     });
 
     it('should set the breakpoint state on mount', function() {
+      console.log('setBreakpointStub.callCount', setBreakpointStub.callCount)
```
simoneduca: Why do we need logs in this test?

app/pages/project/components/ProjectNavbar/ProjectNavbarContainer.spec.js
```diff
@@ -39,18 +44,22 @@ describe('ProjectNavbarContainer', function() {
   it('should render without crashing', function() {});
 
   it('should call getNavLinks on render', function() {
+    console.log('getNavLinksSpy.callCount', getNavLinksSpy.callCount);
```
simoneduca: And in this test too?

eatyourgreens:
Translations are working but I don't see a Classify link on my test project. I'm not sure what's going on there. https://make-new-navbar-public.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground

eatyourgreens:
My test project isn't working on this branch because it has no active workflows 😠 @simoneduca have you been editing those workflows? I'm trying to use it to test a couple of PRs at the moment.

simoneduca:
I have, sorry. I forgot to put them back as they were 😄I see you removed me from your project's collaborators. You're a harsh master.

eatyourgreens:
Only while I was testing #4369 in IE. I can put you back now I'm finished.

eatyourgreens:
@srallen the project page tests use mount to test the Sugar subscriptions:
https://github.com/zooniverse/Panoptes-Front-End/blob/f9f78fe07ade37ab4a58543aee46365692fe70db/app/pages/project/project-page.spec.js#L131-L137

Could that be causing the 2 extra renders of the project nav component that were being reported by Travis? The spies replace the prototype methods, so I'd expect them to persist across component instances. I think you're right that we have to be careful about setting up and tearing down our tests properly.

eatyourgreens:
I've removed those console logs and also the unnecessary bind calls.

srallen:
@eatyourgreens oh yeah, that could have been it. I think we'll need to audit the test spies used in our tests so far to make sure there's a parallel resetHistory or restore call to make sure they're doing what we think they are doing. Thanks for removing the logs. I forgot about those.

@simoneduca I noted in the PR I added logs to try to debug why the tests were only failing on Travis. I forgot to remove then after I got the tests passing. If you have questions about the design, I'm sure @beckyrother can answer.

beckyrother:
Hi @simoneduca, happy to answer questions about the design. You can see the entire redesign here on InVision: https://invis.io/F3F2T5K2Q#/270495242_00_Index to give more context about the overall design changes. 

There are a few reasons for the change. First, the space for the actual classification task has been carefully measured to fit within the height of the average browser window (643 px, according to analytics data). When a user scrolls down to fit the classification task in their browser window, the page will not move when they submit a classification. As they get into the task, there should not be a lot of scrolling at all, unless they want to explore the additional features underneath the classification.  Also, you'll notice that the image researchers chose for the background of the project has been moved to the background of the nav bar. (A future PR will remove the background from the project itself and replace it with a clean #EDF0F4 to enhance readability of both the question and the subject image.) (There will also be a dark theme for images which are a similar color to #EDF0F4 like the Planet Four: Ridges images.) Additionally, the increased height opens up the space and feels less cramped and more "designed".

Hope that helps. Please let me know if there are additional concerns about the changes we're making –we'll be implementing the rest of the redesign in the coming weeks, so it would be good to make sure everyone's on board before we get too far into development.

beckyrother:
Is this ready to merge @srallen @simoneduca @eatyourgreens ?

eatyourgreens:
All good here @beckyrother.

simoneduca:
Thanks @beckyrother for clarifying. Ready to go for me, if you want to press the button.

PR #4369, opened 2018-02-19
Classification queue

marten:
Extracts a ClassificationQueue, making the classify page a little lighter on the logic. Also means that this process is now more easily tested.

Staging branch URL: https://classification-queue.pfe-preview.zooniverse.org

Fixes # .

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -1,98 +1,97 @@
+/* eslint no-underscore-dangle: ["error", {"allowAfterThis": true}] */
 import apiClient from 'panoptes-client/lib/api-client';
-import { Split } from 'seven-ten';
 
 const FAILED_CLASSIFICATION_QUEUE_NAME = 'failed-classifications';
 const MAX_RECENTS = 10;
 
 class ClassificationQueue {
-    constructor(storage, api, onClassificationSaved) {
-        this.storage = storage || window.localStorage;
-        this.apiClient = api || apiClient;
-        this.recents = [];
-        this.onClassificationSaved = onClassificationSaved;
+  constructor(storage, api, onClassificationSaved) {
+    this.storage = storage || window.localStorage;
+    this.apiClient = api || apiClient;
+    this.recents = [];
+    this.onClassificationSaved = onClassificationSaved;
+  }
+
+  add(classification) {
+    const queue = this._loadQueue();
+    queue.push(classification);
+
+    try {
+      this._saveQueue(queue);
+      console.info('Queued classifications:', queue.length);
+    } catch (error) {
+      console.error('Failed to queue classification:', error);
     }
 
-    add(classification) {
-        const queue = this._loadQueue();
-        queue.push(classification);
+    this.flushToBackend();
+  }
 
-        try {
+  length() {
+    return this._loadQueue().length;
+  }
+
+  addRecent(classification) {
+    if (this.recents.length > MAX_RECENTS) {
+      const droppedClassification = this.recents.pop();
+      droppedClassification.destroy();
+    }
+
+    this.recents.unshift(classification);
+  }
+
+  flushToBackend() {
+    const queue = this._loadQueue();
+
+    if (queue.length > 0) {
+      console.log('Saving queued classifications:', queue.length);
+      queue.forEach((classificationData) => {
+        this.apiClient.type('classifications').create(classificationData).save().then((actualClassification) => {
+          console.log('Saved classification', actualClassification.id);
+          this.onClassificationSaved(actualClassification);
+          this.addRecent(actualClassification);
+
+          const indexInQueue = queue.indexOf(classificationData);
+          queue.splice(indexInQueue, 1);
+
+          try {
             this._saveQueue(queue);
-            console.info('Queued classifications:', queue.length);
-        }
-        catch(error) {
-            console.error('Failed to queue classification:', error);
-        }
+            console.info('Saved a queued classification, remaining:', queue.length);
+          } catch (error) {
+            console.error('Failed to update classification queue:', error);
+          }
+        })
+        .catch((error) => {
+          console.error('Failed to save a queued classification:', error);
 
-        this.flushToBackend();
-    }
+          if (error.status === 422) {
+            console.error('Dropping malformed classification permanently', classificationData);
+            const indexInQueue = queue.indexOf(classificationData);
+            queue.splice(indexInQueue, 1);
 
-    length() {
-        return this._loadQueue().length;
-    }
-
-    addRecent(classification) {
-        if (this.recents.length > MAX_RECENTS) {
-            let droppedClassification = this.recents.pop();
-            droppedClassification.destroy();
-        }
-
-        this.recents.unshift(classification);
-    }
-
-    flushToBackend() {
-        const queue = this._loadQueue();
-
-        if (queue.length > 0) {
-            console.log('Saving queued classifications:', queue.length);
-            for (let classificationData of queue) {
-                this.apiClient.type('classifications').create(classificationData).save().then((actualClassification) => {
-                    console.log('Saved classification', actualClassification.id);
-                    this.onClassificationSaved(actualClassification);
-                    this.addRecent(actualClassification);
-                    let indexInQueue = queue.indexOf(classificationData);
-                    queue.splice(indexInQueue, 1);
-                    try {
-                        this._saveQueue(queue);
-                        console.info('Saved a queued classification, remaining:', queue.length);
-                    }
-                    catch(error) {
-                        console.error('Failed to update classification queue:', error);
-                    }
-                }).catch(function(error) {
-                    console.error('Failed to save a queued classification:', error);
-
-                    switch(error.status) {
-                    case 422:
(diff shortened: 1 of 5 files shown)


app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+
+        this.recents.unshift(classification);
+    }
+
+    flushToBackend() {
+        var queue = this._loadQueue();
+
+        if (queue.length > 0) {
+            console.log('Saving queued classifications:', queue.length);
+            for (let classificationData of queue) {
```
eatyourgreens: `for…of` isn’t supported by arrays in IE. I think we’ve preferred `Array.forEach` for better browser compatibility so `queue.forEach( function(classificationData) {…})` might be safer here.

app/pages/project/classify.cjsx
```diff
@@ -9,15 +9,14 @@ counterpart = require 'counterpart'
 `import FinishedBanner from './finished-banner';`
 Classifier = require('../../classifier').default
 seenThisSession = require '../../lib/seen-this-session'
+ClassificationQueue = require('../../lib/classification-queue').default
 `import WorkflowAssignmentDialog from '../../components/workflow-assignment-dialog';`
 { Split } = require('seven-ten')
```
eatyourgreens: Is Split still needed here?

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+                this.apiClient.type('classifications').create(classificationData)
+                    .save()
+                    .then((actualClassification) => {
+                        console.log('Saved classification', actualClassification.id);
+                        Split.classificationCreated(actualClassification); // Metric log needs classification id
```
eatyourgreens: We should find out if Split is still needed here.

eatyourgreens: Actually, would it be better to pass a post-classification callback to the constructor, invoke it here and leave `Split` in the calling page?

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+                            try {
+                                this._saveQueue(queue);
+                            }
+                            catch(error) {
+                                console.error('Failed to update classification queue:', error);
+                            }
+                            break;
+                        }
+                    });
+            }
```
eatyourgreens: Indentation is weird here. eslint should be complaining about this.

eatyourgreens:
Tests are warning about an unhandled Promise rejection, so there's a catch block missing somewhere.

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+            for (let classificationData of queue) {
+                this.apiClient.type('classifications').create(classificationData)
+                    .save()
+                    .then((actualClassification) => {
+                        console.log('Saved classification', actualClassification.id);
+                        Split.classificationCreated(actualClassification); // Metric log needs classification id
+                        this.addRecent(actualClassification);
+                        let indexInQueue = queue.indexOf(classificationData);
+                        queue.splice(indexInQueue, 1);
+                        try {
```
eatyourgreens: I'm not sure why this is using `try…catch` instead of handling errors when the promise rejects.

eatyourgreens: nvm. This is catching local storage errors, like 'out of space'.

test/fake-api-client.js
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    save() {
+        let saved = this.resourceType.save(this.attributes);
+
+        if (saved) {
+            return Promise.resolve(this);
+        }
+        else {
+            return new Promise((resolve, reject) => { reject(); });
```
eatyourgreens: `reject( new Error('Save failed!') )` would be better here. The tests currently log `Failed to save a queued classification: undefined`.

app/lib/classification-queue.js
eatyourgreens: Seems to work in IE11 on BrowserStack, so I might be too paranoid about browser compatibility here.

eatyourgreens:
https://classification-queue.pfe-preview.zooniverse.org/dev/classifier throws `Uncaught TypeError: Cannot read property 'catch' of undefined` when you finish a classification. I think that's an existing bug so not a blocking issue here.

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+
+class ClassificationQueue {
+    constructor(storage, api) {
+        this.storage = storage || window.localStorage;
+        this.apiClient = api || apiClient;
+        this.recents = [];
+    }
+
+    add(classification) {
+        var queue = this._loadQueue();
```
eatyourgreens: Should be `const queue`.

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+        }
+        catch(error) {
+            console.error('Failed to queue classification:', error);
+        }
+
+        this.flushToBackend();
+    }
+
+    length() {
+        return this._loadQueue().length
```
eatyourgreens: Missing semicolon.

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+        if (this.recents.length > MAX_RECENTS) {
+            let droppedClassification = this.recents.pop();
+            droppedClassification.destroy();
+        }
+
+        this.recents.unshift(classification);
+    }
+
+    flushToBackend() {
+        var queue = this._loadQueue();
```
eatyourgreens: `const queue`

app/lib/classification-queue.js
eatyourgreens: `const` instead of `let` here.

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+        if (queue.length > 0) {
+            console.log('Saving queued classifications:', queue.length);
+            for (let classificationData of queue) {
+                this.apiClient.type('classifications').create(classificationData)
+                    .save()
+                    .then((actualClassification) => {
+                        console.log('Saved classification', actualClassification.id);
+                        Split.classificationCreated(actualClassification); // Metric log needs classification id
+                        this.addRecent(actualClassification);
+                        let indexInQueue = queue.indexOf(classificationData);
```
eatyourgreens: `const indexInQueue`

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+                            console.error('Failed to update classification queue:', error);
+                        }
+                    })
+                    .catch(function(error) {
+                        console.error('Failed to save a queued classification:', error);
+
+                        switch(error.status) {
+                        case 422:
+                            console.error("Dropping malformed classification permanently", classificationData);
+                            let indexInQueue = queue.indexOf(classificationData);
```
eatyourgreens: `const indexInQueue`

app/lib/classification-queue.js
```diff
@@ -0,0 +1,101 @@
...
+                            }
+                            break;
+                        }
+                    });
+            }
+        }
+    }
+
+    _loadQueue() {
+        var queue = JSON.parse(this.storage.getItem(FAILED_CLASSIFICATION_QUEUE_NAME));
```
eatyourgreens: `const queue`

eatyourgreens:
eslint throws a ton of warnings, mostly about unexpected indentation throughout. I've added comments for some of the syntax changes it suggests.

app/lib/classification-queue.js
marten: But as it's sending stuff out, it's removing elements from the array. I know that we're not changing the "pointer", but marking it const seems like a lie to readers of the code?

eatyourgreens: It's never reassigned so it's `const` in JS.

app/lib/classification-queue.js
srallen: The metric logs Seven Ten makes needs the classification id if the split is tracking the classification created to find out if the split has an effect on classification. The researcher can then compare the split assigned to the user to the classification data export using the id.

app/lib/classification-queue.js
marten: There's a `babel-plugin-transform-es2015-for-of` in the package.lock. Might be that?

eatyourgreens: Could well be.

app/lib/classification-queue.js
srallen: A callback would be fine for that and I think preferable long term.

srallen:
This is approved, but labeled WIP still. Is the merge being delayed?

app/pages/project/classify.cjsx
marten: Yep

marten:
@srallen I wanted to make sure to fix @eatyourgreens comments, but I think it's good to go now. Apologies for the incorrect indentation.

eatyourgreens:
Looks good to me. Tests are passing and the queue is working with my test project on staging.

Both tests throw the 'Save failed!' error, which I wasn't expecting, but everything seems to be working. I'm happy to merge this.

PR #4370, opened 2018-02-20
Reduce workflow request page size

eatyourgreens:
Staging branch URL: https://workflows-page-size.pfe-preview.zooniverse.org

Reduces the default workflow request page size from 100 to 20. Uses a recursive function to make paged calls until all workflows have been fetched from Panoptes and can be processed.

Currently waits on all the network requests to complete before processing workflows. A better version might incrementally update in the calling component, as each request completes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/lib/get-workflows-in-order.js
```diff
@@ -0,0 +1,52 @@
...
+          allWorkflows = allWorkflows.concat(workflows);
+          const meta = workflows[0] ? workflows[0].getMeta() : null;
+          if (meta && meta.page !== meta.page_count) {
+            return getWorkflows(query, page + 1);
+          }
+          return Promise.resolve(allWorkflows);
+        })
+        .catch((error) => {
+          console.info(error);
+          return Promise.resolve(allWorkflows);
```
camallen: could this allWorkflows promise return and the other above be consolidated to a final `.then` chain block after the catch? I.e. always return the final built ordered workflows array.

app/lib/get-workflows-in-order.js
```diff
@@ -0,0 +1,52 @@
...
+        });
+    }
+  }
+
+  return getAllWorkflows(query).then((workflows) => {
+    const workflowsByID = {};
+    workflows.forEach((workflow) => { workflowsByID[workflow.id] = workflow; });
+
+    const workflowsInOrder = order.map(workflowID => workflowsByID[workflowID]).filter(Boolean);
```
camallen: Should we add sorting by id to the API? It should be pretty cheap as it's the primary key index. The sort order will be maintained throughout the pages.

app/lib/get-workflows-in-order.js
```diff
@@ -0,0 +1,52 @@
+function getWorkflowsInOrder(project, query) {
+  query = query || {};
+  const order = (project.configuration && project.configuration.workflow_order) ?
+    project.configuration.workflow_order :
+    [];
+
+  // TODO remove default page_size once pagination solution implemented
+  query.page_size = query.page_size || 20;
```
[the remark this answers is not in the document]
eatyourgreens: Yep. I think this line is in there for cases where `query.page_size` is null, but I'm not sure. The comment implies it's safe to remove, though.

app/lib/get-workflows-in-order.js
eatyourgreens: Sorry, but I'm not following you here. This catches any errors in the `then` block, which won't happen if we move the `then` block after the catch. It's meant to ensure that the promise chain always returns something, even if one of the recursively chained promises fails.

eatyourgreens: Oh wait, do you mean

```
const allWorkflows;
getWorkflows()
.then( workflows => // maybe get some more workflows and update allWorkflows)
.catch( error => // well, that didn't work )
.then( () => // return the final state of allWorkflows)
```

I can try that and see if it works.

app/lib/get-workflows-in-order.js
eatyourgreens: This doesn't sort by ID. It sorts the workflows in the order that's set in the `project.order` array.

app/lib/get-workflows-in-order.js
```diff
@@ -0,0 +1,52 @@
...
+  function getAllWorkflows(query) {
+    let allWorkflows = [];
+    return getWorkflows(query, 1);
+
+    function getWorkflows(query, page) {
+      return project.get('workflows', Object.assign({}, query, { page }))
+        .then((workflows) => {
+          allWorkflows = allWorkflows.concat(workflows);
+          const meta = workflows[0] ? workflows[0].getMeta() : null;
+          if (meta && meta.page !== meta.page_count) {
```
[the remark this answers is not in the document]
eatyourgreens: I did not know that. That changes code like this in a few other places. Pinging @mcbouslog because he could use this for his organisation queries too.

app/lib/get-workflows-in-order.js
camallen: Gotcha, i see what it's doing now. I miss read that.

app/lib/get-workflows-in-order.js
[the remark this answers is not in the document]
mcbouslog: I think it was added to accommodate the quantity of NfN workflows, but can now be safely removed. It's laughably small, but quickly threw together https://github.com/zooniverse/Panoptes-Front-End/compare/remove-page-sizes.

PR #4378, opened 2018-02-21
Clean up page  and header style cruft

srallen:
Staging branch URL: https://clean-up-style-cruft.pfe-preview.zooniverse.org/

This cleans up a lot of accumulated cruft with the page styles and the header. The 'on-secondary-page' class was originally used to scope the header variants for the original header design. On the home page it was transparent and on every other page, the background color was cobalt blue. I've audited all of the instances where a class was being added to the html tag on mount and found thus:
- `on-collection-page` did nothing. Removed it
- `on-secondary-page` just scoped the styles unnecessarily since that header variant no longer exists. Removed it.
- `on-home-page-not-logged-in` was scoping some styles that had to do with the second header iteration that used the 'detached' functionality. This is no longer being used, so I removed it. I also had confirmed with @beckyrother a while ago while making the ZRC version of the header that the links should have full opacity and what color they should be. I've made that update here.
- `on-project-page` defined some defaults for the font color and background color. The `BACKGROUND` variable is white and the `.project-page` class also set the color to be white, so it was redundant. I then moved the default background color to the `.project-page` class since that component is the parent of all the project pages.
- After I removed the styles that unnecessarily absolute positioned the header, the signed in home page was noticeable scrolling on load. The current site isn't doing that, so I refactored the scroll into view code and added a onClick handler to the close button of the sections so it only scrolls when opening or closing one of the signed in home page sections.

Not totally necessary, but finding all this cruft was really bugging me. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


css/project-page.styl
```diff
@@ -62,6 +58,7 @@
   z-index: 2
 
 .project-page
+  background: FOREGROUND
```
eatyourgreens: Wut? `FOREGROUND` is confusing name for the background colour.

css/site-nav.styl
```diff
@@ -16,14 +16,14 @@
 
   &__link
     border-bottom: 2px solid transparent
-    color: LIGHT_GREY
+    color: #b2b2b2
```
eatyourgreens: We should be consistent about naming colours instead of using hex codes.

PR #4396, opened 2018-02-23
add email password reset form errors

### LINKED ISSUE #4390: Password reset form swallows API errors
https://github.com/zooniverse/Panoptes-Front-End/blob/d6c8a183e3a8e6dfe653f426c77e9107a240e446/app/pages/reset-password/new-password-form.jsx#L48-L53

The original error message should be shown here, instead of a generic "Something went wrong" error that doesn't tell volunteers what they did wrong.

camallen:
Staging branch URL:  https://email-reset-errors.pfe-preview.zooniverse.org/reset-password?reset_password_token=H6GtX

Fixes #4390 and lots of email reports of problems with email password resets, specifically no API error reports, no length checking or password match checks. 

This PR adds:
1. the API failure messages to the form
0. HTML5 input password length validation before form submission
0. Checks the passwords match before attempting to contact the API.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/locales/es.js
+++ b/app/locales/es.js
@@ -156,9 +156,9 @@ export default {
   },
   resetPassword: {
     heading: 'Reset Password',
-    newPasswordFormDialog: 'Enter matching passwords that are at least 8 characters long so you can get back to doing some research.',
+    newPasswordFormDialog: 'Enter the same password twice so you can get back to doing some research. Passwords need to be at least 8 characters long.',
     newPasswordFormLabel: 'New password:',
-    newPasswordConfirmationLabel: 'Again, to confirm:',
+    newPasswordConfirmationLabel: 'Repeat your password to confirm:',
     enterEmailLabel: 'Please enter your email address here and we’ll send you a link you can follow to reset it.',
     emailSuccess: 'We’ve just sent you an email with a link to reset your password.',
     emailError: 'There was an error reseting your password.',
--- a/app/locales/es.js
+++ b/app/locales/es.js
@@ -156,7 +156,7 @@ export default {
   },
   resetPassword: {
     heading: 'Reset Password',
-    newPasswordFormDialog: 'Go ahead and enter a new password, then you can get back to doing some research.',
+    newPasswordFormDialog: 'Enter matching passwords that are at least 8 characters long so you can get back to doing some research.',
     newPasswordFormLabel: 'New password:',
     newPasswordConfirmationLabel: 'Again, to confirm:',
     enterEmailLabel: 'Please enter your email address here and we’ll send you a link you can follow to reset it.',
--- a/app/locales/es.js
+++ b/app/locales/es.js
@@ -163,6 +163,7 @@ export default {
     emailSuccess: 'We’ve just sent you an email with a link to reset your password.',
     emailError: 'There was an error reseting your password.',
     resetError: 'Something went wrong, please try and reset your password via email again.',
+    passwordsDoNotMatch: 'The passwords do not match, please try again.',
     loggedInDialog: 'You are currently logged in. Please log out if you would like to reset your password.'
   },
   workflowToggle: {
--- a/app/locales/nl.js
+++ b/app/locales/nl.js
@@ -189,7 +189,7 @@ export default {
   },
   resetPassword: {
     heading: 'Wachtwoord vergeten',
-    newPasswordFormDialog: 'Voer een nieuw wachtwoord in, dan kun je verder met onderzoek doen.',
+    newPasswordFormDialog: 'Voer bijbehorende wachtwoorden in die ten minste 8 tekens lang zijn, zodat u verder kunt gaan met onderzoek.',
     newPasswordFormLabel: 'Nieuw wachtwoord:',
     newPasswordConfirmationLabel: 'Herhaal, ter controle:',
     enterEmailLabel: 'Voer je e-mailadres in en we sturen je een link waarmee je kunt herstellen.',
--- a/app/locales/nl.js
+++ b/app/locales/nl.js
@@ -189,7 +189,7 @@ export default {
   },
   resetPassword: {
     heading: 'Wachtwoord vergeten',
-    newPasswordFormDialog: 'Voer bijbehorende wachtwoorden in die ten minste 8 tekens lang zijn, zodat u verder kunt gaan met onderzoek.',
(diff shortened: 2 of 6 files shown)


app/pages/reset-password/reset-password.jsx
```diff
@@ -64,18 +64,35 @@ class ResetPasswordPage extends React.Component {
     const token = this.props.location.query.reset_password_token;
     const password = event.target[0].value;
     const confirmation = event.target[1].value;
+    const passwordMatch = password === confirmation
```
camallen: I wasn't sure the best way to test form submission behaviour via enzyme. Happy to add testing in here if folks feel strongly about it.

app/locales/es.js
```diff
@@ -156,13 +156,14 @@ export default {
   },
   resetPassword: {
     heading: 'Reset Password',
-    newPasswordFormDialog: 'Go ahead and enter a new password, then you can get back to doing some research.',
+    newPasswordFormDialog: 'Enter matching passwords that are at least 8 characters long so you can get back to doing some research.',
```
marten: I think the "matching passwords" is a bit weird. A confirmation field is quite normal, and I can't remember seeing it labeled like this. I'd be wondering "matching to what?".

app/locales/es.js
```diff
@@ -156,13 +156,14 @@ export default {
...
     newPasswordFormLabel: 'New password:',
     newPasswordConfirmationLabel: 'Again, to confirm:',
```
marten: Perhaps instead "matching passwords", this label should be improved? Think it's often called "Confirm password"?

marten: How about "Retype your new password to confirm"

app/locales/es.js
marten: Perhaps "Enter the same password twice so you can get back to doing some research. Passwords need to be at least 8 characters long."?

app/locales/nl.js
```diff
@@ -189,13 +189,13 @@ export default {
   },
   resetPassword: {
     heading: 'Wachtwoord vergeten',
-    newPasswordFormDialog: 'Voer een nieuw wachtwoord in, dan kun je verder met onderzoek doen.',
+    newPasswordFormDialog: 'Voer bijbehorende wachtwoorden in die ten minste 8 tekens lang zijn, zodat u verder kunt gaan met onderzoek.',
```
marten: Matching my other suggestion, this would be: "Voer hetzelfde wachtwoord tweemaal in, zodat je verder kunt met onderzoek doen. Een wachtwoord moet tenminste 8 tekens lang zijn."

"Bijbehorende" is the wrong term in any case. That'd translate back as "passwords belonging to", not "matching". Let me know if you need a better literal translation instead of for my own suggestion

app/locales/nl.js
```diff
@@ -189,13 +189,13 @@ export default {
...
     newPasswordFormLabel: 'Nieuw wachtwoord:',
     newPasswordConfirmationLabel: 'Herhaal, ter controle:',
     enterEmailLabel: 'Voer je e-mailadres in en we sturen je een link waarmee je kunt herstellen.',
     emailSuccess: 'We hebben je zojuist een e-mail met een herstellink gezonden.',
     emailError: 'Er ging iets mis bij het herstellen van je wachtwoord.',
-    resetError: 'Er ging iets mis, probeer het nogmaals.',
+    passwordsDoNotMatch: 'De wachtwoorden komen niet overeen, probeer het opnieuw.',
```
marten: 👍

app/pages/reset-password/new-password-form.jsx
```diff
@@ -46,29 +54,29 @@ const NewPasswordForm = ({ onSubmit, disabled, inProgress, resetSuccess, resetEr
         <i className="fa fa-check-circle form-help success" />}
 
       {resetError &&
-        <Translate
-          className="form-help error"
-          component="small"
-          content="resetPassword.resetError"
-        />}
+        <small className="form-help error">
```
simoneduca: Why has `Translate` been removed?

camallen: Because the resetError text is an error response from the API now and will vary based on the api error.

eatyourgreens:
Oddly enough, that password reset form does have JS validation. I just tested it and it told me `1111` was too short. 🤔

eatyourgreens:
If you want to consolidate the code, here's the other form.
https://github.com/zooniverse/Panoptes-Front-End/blob/2548ea60e9e940c914a6021c8bd5100edaf17315/app/pages/settings/account.cjsx#L9-L90

camallen:
Ha - right, the other form you found is for changing a password for a logged in user via https://www.zooniverse.org/settings. 

This form is for unauthenticated users that is accessible via the password reset email link we send from the server. 

Request a forgot email password at https://www.zooniverse.org/reset-passwordand the server sends you an email with a reset link. That link takes you to this page and it wires up the api request with the one time use password reset token. 

Also the change password action uses a different API end point to reset/change the password, [reset](https://github.com/zooniverse/panoptes-javascript-client/blob/595e694c71dd877ddbae4a13e670fc466314258e/lib/auth.js#L304) vs [change](https://github.com/zooniverse/panoptes-javascript-client/blob/595e694c71dd877ddbae4a13e670fc466314258e/lib/auth.js#L260). 

I agree we can consolidate the form but can that come in another PR? Our password resets are currently very problematic (see contact@ emails) so shipping this should be a priority.

eatyourgreens:
I've got no problem with merging this and coming back to consolidate the code for the two forms.

By the way, whil elooking into how this works, I found this in the `changePassword` action. The new password is also submitted as the confirmation. I'm not sure if that's a bug, or actually how the form was originally intended to work. 🤔 
https://github.com/zooniverse/panoptes-javascript-client/blob/595e694c71dd877ddbae4a13e670fc466314258e/lib/auth.js#L267-L269

camallen:
good catch @eatyourgreens! I bet that is one of the issues with the reset not working as I can mistype in the first field. I'll fix this then :shipit:

camallen:
Apologies, i miss read that last comment. On authenticated password change it seems that it relies on the JS to ensure they match before submission, so they should be the same.

eatyourgreens:
I wasn't sure because, as you pointed out, there's a separate reset action in that same file. One for each of the two forms. `resetPassword` seems to pass the confirmation correctly.

PR #4404, opened 2018-02-26
Add dropdown task default annotations on update

### LINKED ISSUE #4401: Malformed classifications being saved that break exports
Occasionally, classification export jobs for a given project or workflow will mysteriously vanish and unhelpful NoMethodErrors show up in Honeybadger. This is due to malformed classifications being saved by PFE and when the exporter tries to untangle them, it chokes and dies silently. This seems to happen with huge tasks consisting of many dropdowns and combos. 

I've tracked down a few examples which will hopefully help us track this down. I've formatted them but put them in gists because lordy they're humungous. Project 3414. workflow 5919. There's a significant number of these classifications for this project/workflow and tbh I'm not sure yet how we fix/clean these so that exports run once we stop this from happening.

Workflow task definition:
https://gist.github.com/zwolf/f00b4f67629dc465aecbe8be9ffd9fc8

Here's an annotation that causes the NoMethodError in the exporter:
https://gist.github.com/zwolf/ed952817b4308ca3d64c6c972d299093

Seems like this kind of thing could be the culprit:
```
      {
        "task": "T13",
        "value": [
          null,
          null,
          {
            "value": "Canada",
            "option": false
          }
        ]
      },
```

cc: @mcbouslog / @trouille

mcbouslog:
Staging branch URL: https://dropdown-default-update-fix.pfe-preview.zooniverse.org/projects/markb-panoptes/test-project-1-mb/classify?reload=1&workflow=3165

Closes #4401. 

Describe your changes.
- check and if annotations not present, set default annotations (`{ option: false, value: null }`) on component update

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -90,21 +90,22 @@ module.exports = createReactClass
     @menus = []
 
   componentDidMount: ->
-    @setDefaultAnnotations()
+    @setDefaultValues()
 
     if @props.autoFocus is true
       @handleFocus()
 
   componentDidUpdate: (prevProps) ->
     if prevProps.task isnt @props.task
-      @setDefaultAnnotations()
+      @setDefaultValues()
       if @props.autoFocus is true
         @handleFocus()
 
-  setDefaultAnnotations: ->
+  setDefaultValues: ->
     unless @props.annotation.value.length
-      for select, i in @props.task.selects
-        @onChangeSelect(i)
+      value = @props.task.selects.map () -> { option: false, value: null }
+      newAnnotation = Object.assign @props.annotation, {value}
+      @props.onChange newAnnotation
 
   selectedOptions: () ->
     # return annotation values mapped to react-select option objects
--- a/app/classifier/tasks/dropdown/index.cjsx
+++ b/app/classifier/tasks/dropdown/index.cjsx
@@ -90,16 +90,21 @@ module.exports = createReactClass
     @menus = []
 
   componentDidMount: ->
-    annotationValues = @props.annotation.value
-    unless annotationValues.length
-      @props.annotation.value = @props.task.selects.map -> {value: null, option: false}
+    @setDefaultAnnotations()
 
     if @props.autoFocus is true
       @handleFocus()
 
   componentDidUpdate: (prevProps) ->
-    if prevProps.task isnt @props.task and @props.autoFocus is true
-      @handleFocus()
+    if prevProps.task isnt @props.task
+      @setDefaultAnnotations()
+      if @props.autoFocus is true
+        @handleFocus()
+
+  setDefaultAnnotations: ->
+    unless @props.annotation.value.length
+      for select, i in @props.task.selects
+        @onChangeSelect(i)
 
   selectedOptions: () ->
     # return annotation values mapped to react-select option objects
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Should there be tests for these defaults in the dropdown task tests?

eatyourgreens:
Not a blocking comment but the naming here is a little bit confusing. The default annotation for a dropdown task is an empty array. Is this setting default initial values for the dropdown menus, then adding those to the empty annotation? If so, it might need to call the `onChange` handler to register the new values in classifier state.

mcbouslog:
Great point @eatyourgreens! I have a meeting this morning, but would then like to check and edit accordingly, please hold on merging.

eatyourgreens:
I'm not 100% sure about needing to call `onChange`, since the current component works without doing that. That might be because the current code changes `props.annotation.value` which then changes `state.annotations` in the classifier, indirectly.

mcbouslog:
Added tests.

@eatyourgreens I think your most recent comment is correct, it was working due to changing props. I refactored so on mount and on update it calls `setDefaultAnnotations` (also don't like name, open to change) which iterates through selects with their index and calls `onChangeSelect`. Which I think is preferable? Let me know what you think, happy to change back.

eatyourgreens:
I was thinking something like this to set up the initial values. I'm not sure if my coffeescript syntax is correct:

```coffeescript
value = @props.task.selects.map () -> { option: false, value: null }
newAnnotation = Object.assign {}, @props.annotation, { value }
@props.onChange newAnnotation
```

I'm not sure if it's better to call `onChange` once, with the new annotation, or once for each select. It looks like your current code works, so this isn't a blocking issue.

mcbouslog:
That's more elegant, only makes one `onChange` call - updated accordingly.

I think this is ready then, I'll assign to you to merge when comfortable with the recent change, thanks Jim!

eatyourgreens:
👍 Looks good to me. Calling `onChange` immediately on mount seems like an odd pattern, just to set some initial values, but I think fixing that would involve changes to how all the task components work, and is outside the scope of this PR.

app/classifier/tasks/dropdown/index.cjsx
```diff
@@ -90,16 +90,22 @@ module.exports = createReactClass
...
-      @handleFocus()
+    if prevProps.task isnt @props.task
+      @setDefaultValues()
+      if @props.autoFocus is true
+        @handleFocus()
+
+  setDefaultValues: ->
+    unless @props.annotation.value.length
+      value = @props.task.selects.map () -> { option: false, value: null }
+      newAnnotation = Object.assign @props.annotation, {value}
```
eatyourgreens: There's a missing {} here. This modifies `@props.annotation`, rather than copying it and creating a new annotation.

`newAnnotation = Object.assign {}, @props.annotation, {value}

mcbouslog:
Yes, makes sense. And I think I should change [the similar line in `onChangeSelect`](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/dropdown/index.cjsx#L205), to the same:
```
newAnnotation = Object.assign {}, @props.annotation, {value}
```

eatyourgreens:
Good idea. My reason for doing this is because `props.annotation` might be a reference to the original variable in the Classifier component.

mcbouslog:
Yep makes sense, good and important refactoring.

PR #4407, opened 2018-02-27
Add rotating rectangle as a drawing tool

CKrawczyk:
Staging branch URL: https://rotate-rect.pfe-preview.zooniverse.org/

Describe your changes.

This PR adds a rotating rectangle as a new drawing tool.

~~At the moment the drag handles do not work as expected when the rectangle is rotated. I think this has to do with the coordinate system the drag even is in being different than the one the `svg` object is in.~~

~~This is currently in coffee script since I started with the rectangle drawing tool's code.  I plan to convert it to ES6 once it is working.~~ The code is now in ES6.

Things left to do
- [x] fix drag handles when rectangle is rotated
- [x] convert to ES6

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


CKrawczyk:
Everything is now working correctly for this PR.

eatyourgreens:
Checking this out quickly, I find it weird that the rectangle rotates around the top left corner, rather than the centre. Maybe test this out with a few people to see if they feel the same?

CKrawczyk:
I tried that at first and had a hard time getting the resize handles work nicely (the point of rotation is constantly changing as it is being resized).   I will give it another try.

eatyourgreens:
Ask a couple of other people before you put a lot of effort into it. I might be the only person who finds this hard to use.

CKrawczyk:
I have updated the code to have the rectangle rotate about the center.

To make the drag handles work nicely the code has been updated so the center of the rectangle stays fixed during resize.  The staging link has been updated.

app/classifier/drawing-tools/index.coffee
```diff
@@ -14,4 +14,5 @@ module.exports =
   point: require './point'
   polygon: require './polygon'
   rectangle: require './rectangle'
+  rotateRectangle: require('./rotate-rectangle.jsx').default
```
eatyourgreens: Doesn't this need the `.jsx` extension? We don't use it anywhere else that we require JSX components into CJSX files.

app/classifier/drawing-tools/rotate-rectangle.jsx
```diff
@@ -0,0 +1,292 @@
...
+  constructor(props) {
+    super(props);
+    this.getDeletePosition = this.getDeletePosition.bind(this);
+    this.handleMainDrag = this.handleMainDrag.bind(this);
+    this.handleTopLeftDrag = this.handleTopLeftDrag.bind(this);
+    this.handleTopRightDrag = this.handleTopRightDrag.bind(this);
+    this.handleBottomLeftDrag = this.handleBottomLeftDrag.bind(this);
+    this.handleBottomRightDrag = this.handleBottomRightDrag.bind(this);
+    this.handleRotate = this.handleRotate.bind(this);
+    this.normalizeMark = this.normalizeMark.bind(this);
```
eatyourgreens: I'm not sure, but I don't think you need to bind `normalizeMark` here. Instance methods are bound to an instance of a class by default in javascript. Only event handlers need to be explicitly bound.

app/classifier/drawing-tools/index.coffee
CKrawczyk: Good catch, this is a hold over from when I was converting out of `cjsx`, I will remove it.

eatyourgreens: Cool. I'll hold off on merging until you've updated the code.

app/classifier/drawing-tools/rotate-rectangle.jsx
eatyourgreens: I missed that it's also an event handler, for `onEnd`, which does change the `this` context.

PR #4411, opened 2018-03-01
Update feedback after annotations

### LINKED ISSUE #4410: Failure feedback message displayed for dud subjects
https://master.pfe-preview.zooniverse.org/projects/rogerhutchings/feedback-test (point-based feedback workflow)

## Expected behavior
If I click Done on a dud subject, without drawing a mark, I should be told 'You found a dud.'

## Current behavior
If I click Done, I get the failure message 'It was actually a dud :(' or 'Actually it was a dud.'

If I draw a point, then delete it again, I get the success message.

eatyourgreens:
Staging branch URL: https://fix-4410.pfe-preview.zooniverse.org

Fixes #4410 .

Describe your changes.
Check the default annotation for feedback after loading a new subject.

Wait for `state.annotations` to change before updating feedback rules.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -147,11 +147,8 @@ class Classifier extends React.Component {
 
   updateAnnotations(annotations) {
     annotations = annotations || this.props.classification.annotations.slice();
-    this.setState({ annotations });
     this.props.classification.update({ annotations });
-    if (this.props.feedback.active) {
-      this.updateFeedback();
-    }
+    this.setState({ annotations }, this.updateFeedback);
   }
 
   updateFeedback() {
@@ -175,7 +172,7 @@ class Classifier extends React.Component {
       }, false);
     }
 
-    if (!isInProgress) {
+    if (this.props.feedback.active && !isInProgress) {
       this.props.actions.feedback.update(currentAnnotation);
     }
   }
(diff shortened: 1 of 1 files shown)


eatyourgreens:
This displays the correct feedback message if the first subject you see is a dud. Subsequent dud subjects display the wrong feedback message, however.

eatyourgreens:
This is ready for review now.

app/classifier/classifier.jsx
```diff
@@ -175,7 +172,7 @@ class Classifier extends React.Component {
       }, false);
     }
 
-    if (!isInProgress) {
+    if (this.props.feedback.active && !isInProgress) {
```
rogerhutchings: I'd recommend putting this check at the top of the function and returning early if false - it gets the 'no feedback' case out of the way quicker.

rogerhutchings:
Dud messages now correctly display, and other feedback messages display as normal 👍

app/classifier/classifier.jsx
eatyourgreens: Good point.

eatyourgreens: Fixed in 03ce256.

rogerhutchings:
It just occurred to me that another way to do this would be to use `redux-thunk` to dispatch the `update()` action as a side-effect of the `init()` action. It's probably a bit more complex than what's really required for this case though

PR #4413, opened 2018-03-01
Grid draw tool

miclaraia:
Added a drawing tool that takes advantage of a strict grid structure in images. Define the number of rows and columns, as well as the x and y offsets of the grid in pixels, in the project builder. In the classifier, clicking anywhere in the grid region of a minor subject marks the entire region. Coordinates are passed as the original clicked location, not the center of the grid region.

giphy

In the muon hunters reboot we have users classifying many subjects together in a single zooniverse subject image. Originally we were using the point tool, but that was too difficult to use for this purpose. Some of our users told us during our beta testing that it was very difficult to see the point tool marker and know which ones they had already classified. We got multiple complaints from users on tablets that the point tool was impossible to use because trying to re-select the point marker on a tablet was near impossible. 

A project using this tool is configured on staging, project 1814 (http://localhost:3735/projects/miclaraia/muon-hunters-2/classify).

# Required Manual Testing

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari? Tested on Chrome and Firefox
- [x] Does it work on mobile?
- [ ] Are the tests passing locally and on Travis? - Not sure how to do this

Open to suggestions and improvements.


app/classifier/drawing-tools/point-grid.jsx
```diff
@@ -0,0 +1,105 @@
...
+const isInBounds = require('../../lib/is-in-bounds');
+const DeleteButton = require('./delete-button');
+
+const DELETE_BUTTON_ANGLE = 45;
+const SELECTED_RADIUS = {
+  large: 20,
+  small: 10
+};
+
+module.exports = createReactClass({
```
eatyourgreens: This should extend `React.Component` rather than using `createReactClass`.

miclaraia: Ah, my mistake, I was mimicking how the other drawing tools were written, and then put it through decaffeinate. I think I converted it over properly now.

eatyourgreens:
Are you able to stage this using `npm run stage`? If not, you might have to talk to @astopy about setting up access to our S3 bucket.

miclaraia:
@srallen I added the tool to the dev classifier. It seems to behave together with the other drawing tools.
@eatyourgreens I don't have any AWS credentials for Zooniverse, I'll have a chat with Adam and see what he says.

miclaraia:
I was able to stage this branch to https://grid-draw-tool.pfe-preview.zooniverse.org/

A project with this tool is set up at https://grid-draw-tool.pfe-preview.zooniverse.org/projects/miclaraia/muon-hunters-2/classify

rogerhutchings:
The project isn't currently visible - can you mark it as public in the Project Builder's Visibility page?

screenshot 2018-03-06 21 40 12

miclaraia:
Ah yes, of course. Should be fixed now.

miclaraia:
@eatyourgreens, @srallen, I think I've addressed the concerns that have been raised. Let me know if there's anything else that I need to do.

app/classifier/tasks/drawing/grid-editor.jsx
```diff
@@ -0,0 +1,168 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import AutoSave from '../../../components/auto-save';
+
+class GridEditor extends React.Component {
+  static propTypes = {
```
srallen: Same comment about the static declaration here. This will have to be instead:
```
GridEditor.propTypes = {};
```

app/classifier/drawing-tools/point-grid.jsx
```diff
@@ -0,0 +1,101 @@
...
+    const offset_y = Number.parseInt(props.task.tools[mark.tool].y_offset)
+
+    const notBeyondWidth = mark.x < naturalWidth;
+    const notBeyondHeight = mark.y < naturalHeight;
+    const beyondOffset = mark.x > offset_x && mark.y > offset_y
+
+    return notBeyondWidth && notBeyondHeight && beyondOffset;
+  }
+
+  static options = ['grid']
```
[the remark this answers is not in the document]
eatyourgreens: I think we can remove that rule. I’ll double check tomorrow, but I remember the PR that removed `createClass` added a Babel plugin for `static`, because the codemod used it.

CKrawczyk:
I also noticed no icon has been created for this drawing task: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/tasks/drawing/icons.cjsx

app/classifier/drawing-tools/point-grid.jsx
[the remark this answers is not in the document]
srallen: You're right. It looks like this babel plugin was added: https://babeljs.io/docs/plugins/transform-class-properties/

We just need to get ESLint rule set updated because it's flagging this as an error.

app/classifier/tasks/drawing/grid-editor.jsx
srallen: You can ignore this, but the ESLint rule set needs updating to reflect that the babel transforms are supporting this now.

app/classifier/drawing-tools/index.coffee
```diff
@@ -15,3 +15,4 @@ module.exports =
   polygon: require './polygon'
   rectangle: require './rectangle'
   triangle: require './triangle'
+  pointGrid: require './point-grid.jsx'
```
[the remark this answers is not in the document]
miclaraia: I don't know about this one. Removing that line makes the tool disappear from the drop-down list in the project builder. I'm not too familiar with this codebase and tools but it seems like this file gets used to load the drawing tools, but I can't find any explicit references.

Is there another way I should be exporting the drawing tool, if not like this?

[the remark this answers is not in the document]
srallen: Ah, I missed this in my first review. JSX files should use the newer export spec (see here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/export)

`modules.export` is the older CommonJS spec which coffeescript uses. 

I'll leave a comment inline about what should change.

app/classifier/drawing-tools/point-grid.jsx
```diff
@@ -0,0 +1,101 @@
...
+  destroyTool = (e) => {
+    if (this.props.mark._inProgress == false && this.props.selected == true) {
+      return this.setState({destroying: true}, () => {
+        return setTimeout(this.props.onDestroy, 300);
+      });
+    }
+  }
+};
+
+module.exports = PointGridTool
```
srallen: Change this to: `export default PointGridTool`

miclaraia: I'm having trouble getting this to work. If I do this and remove the reference from the index.coffee file then this tool doesn't show up in the project builder
screenshot

miclaraia:
I think these changes address the concerns raised in the last review.

- Added an icon
- Cleaned up further, addressed eslint errors, and updated static definitions, and module exports
- Render the delete button in a different position for the top row and right column so it can't get rendered off-screen

miclaraia:
@srallen looks like the changes you requested earlier are still blocking the PR. Are you happy with the changes I've made, and is there anything else I still need to do?

srallen:
I'll do another look over today. Thanks for the reminder.

PR #4421, opened 2018-03-06
Make the freehand drawing tools public - BLOCKED

eatyourgreens:
Staging branch URL: https://freehand-tools.pfe-preview.zooniverse.org

Makes all four freehand line tools available in the workflow task editor.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


camallen:
The data that comes out of this tool can result in really large annotation payloads, it would be better if we modify how the annotation data in tool is stored to reduce duplication of nested objects (e.g. each node on the line is {x: 1, y: 1} but we could easily strip out the x/y, though we should probably self document the schema in the payload as well to avoid mistakes when interpreting.

eatyourgreens:
I thought @rogerhutchings had already reduced the large number of points being recorded by Etch-A-Cell in #3827?

rogerhutchings:
I did, points are rounded to 1 decimal place and then duplicates are removed. But the way the data are stored could still be optimised:

```
// Current annotation array
[
	{
		x: 123,
		y: 456
	},
	{
		x: 789,
		y: 123
	}
]

// Proposed
[
	[123, 456],
	[789, 123]
]

eatyourgreens:
Should we block this PR until that change is made?

rogerhutchings:
Yes, and I'd want to warn Etch-A-Cell that the annotation schema is going to be changed. Besides, I don't want any other projects to have to handle a change like this when we can do it before the tools are public.

rogerhutchings:
@hrspiers Can you get in touch with the EaC team about the above?

rogerhutchings:
One other thing - do we want to announce that we've made these tools public, and how / where if so?

tlalka:
I was looking at this build, and it seems like some of the free hand tools aren't completely working. Free hand segment shape, and free hand segment line cannot make concave lines/shapes, and they also cannot place a point that overlaps another line/shape. When you attempt to place a point within another segment line/shape, you automatically select the new shape for dragging, but you can't move the line/shape anyway.

eatyourgreens:
There’s been a request to use these tools in a project. Is this still blocked by #4423?

eatyourgreens:
Rebased to the latest master.

srallen:
I think we should build the updated version of the freehand tools in the rewrite and ask project using it to migrate to the new classifier once it is ready. Then we can have a clear delineation of the tools.

PR #4388, opened 2018-02-22
Neutral classifier background

srallen:
Staging branch URL: https://neutral-classifier-background.pfe-preview.zooniverse.org/

This is follow up to the new project nav. This removes the project background as the default background from the whole project to the background color in the new classifier redesign. This obviously, has a cascading effect, so I've done by best to update any styles where necessary. This includes a start of bringing in more of the styles from the redesign such as removing a lot of button border radiuses, changing the flex max-width on the task area, and some of the colors with the survey task. 

I've already checked with and got an ok from @beckyrother on the changes to the in project context collections, recents, and user profile as well as removing with margin from the project announcement banner.

TO DO:
- [x] Check all the workflow tasks to make sure there aren't any styles that need updating as a consequence of making the task-area background white
- [x] Add the dark background toggle button

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4426, opened 2018-03-07
Convert ProjectClassifyPage from coffeescript to jsx

marten:
Staging branch URL: https://convert-classify-to-js.pfe-preview.zooniverse.org/

I converted the classify.cjsx file to JS.  This was prompted by wanting to extract features from it into discrete objects, such that the Turk interface can reuse this.

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional


app/pages/project/classify.jsx
```diff
@@ -0,0 +1,434 @@
...
+
+        {this.renderClassifier()}
+      </div>
+    );
+  },
+
+  renderClassifier() {
+    if (this.state.classification) {
+      return (
+        <Classifier.default
```
eatyourgreens: Why is this `Classifier.default` and not `Classifier`?

marten: That's how I had it originally, but it didn't work (React fails to render, and says that the page is trying to render something that's not a React component). I haven't a clue why not.

app/pages/project/classify.cjsx
```diff
@@ -1,318 +0,0 @@
...
-# In the future user might be able to specify subject sets, which we'll record here similarly.
-currentClassifications =
-  forWorkflow: {}
-
-# Queue up subjects to classify here.
-upcomingSubjects =
-  forWorkflow: {}
-
-emptySubjectQueue = ->
-  console?.log 'Emptying upcoming subjects queue'
```
marten: A lot (but not all) `console` statements in the original had the question mark operator with it. I changed them to just `console.log` in JS. When and why would this not be present?

srallen: If you're not running the script in the browser. It was left over from when Brian published the entirety(!) of PFE as a npm package and it used to be imported into node scripts like the old panoptes-subject-uploader. Checking for it still could be useful if we're ever in the place to reuse code with the native app, but until we're at that place, I wouldn't worry about it.

app/pages/project/classify.jsx
srallen: The Classifier being imported, index.cjsx in the classifier folder, is still coffeescript, so it's using the CommonJS modules rather than the newer spec. You can switch the import to a CommonJS require instead and specify default there:
`const Classifier = require('../../classifier').default;`

eatyourgreens: My bad. I broke that export when I connected the classifier to the Redux store.

eatyourgreens: Should be fixed in #4429.

srallen: #4429 is now merged.

app/pages/project/classify.jsx
```diff
@@ -0,0 +1,434 @@
...
+// In the future user might be able to specify subject sets, which we'll record here similarly.
+const currentClassifications = { forWorkflow: {}};
+
+// Queue up subjects to classify here.
+const upcomingSubjects = { forWorkflow: {}};
+
+function emptySubjectQueue() {
+  console.log('Emptying upcoming subjects queue');
+
+  for (const workflowID in upcomingSubjects.forWorkflow) {
```
srallen: I think `Object.keys(upcomingSubjects.forWorkflow).forEach((workflowID) => {})` would preferable over `for...in` here. For loops in javascript will also iterate over non-enumerable properties the object may have inherited: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in

app/pages/project/classify.jsx
```diff
@@ -0,0 +1,434 @@
...
+
+const classificationQueue = new ClassificationQueue(window.localStorage, apiClient, onClassificationSaved);
+
+auth.listen('change', emptySubjectQueue);
+apiClient.type('subject_sets').listen('add-or-remove', emptySubjectQueue);
+
+// Store this externally to persist during the session.
+let sessionDemoMode = false;
+
+const ProjectClassifyPage = createReactClass({
```
srallen: I won't block this, but a native ES6 class would be preferable in future conversions.

app/pages/project/classify.spec.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import assert from 'assert';
+import { shallow } from 'enzyme';
+
+import { FakeApiClient } from '../../../test/fake-api-client';
+import ProjectClassifyPage from './classify';
+import FinishedBanner from './finished-banner';
+
+describe('ProjectClassifyPage', function () {
```
marten: Don't be too fooled by these, since all the interesting bits require API interaction to load workflows, subjects etc, and that's really not tested afaik?

srallen:
The rewrite is also missing the dark theme from #4388

marten:
Looks good. Who reviews the reviewer?

srallen:
@mcbouslog could you help review this? I'd like to get this reviewed sooner than later because this rewrite is going to keep getting out of sync with some of the design updates that are happening.

app/pages/project/classify.jsx
```diff
@@ -0,0 +1,447 @@
...
+
+const classificationQueue = new ClassificationQueue(window.localStorage, apiClient, onClassificationSaved);
+
+auth.listen('change', emptySubjectQueue);
+apiClient.type('subject_sets').listen('add-or-remove', emptySubjectQueue);
+
+// Store this externally to persist during the session.
+let sessionDemoMode = false;
+
+function shouldSimulateSaveFailure(location) {
```
eatyourgreens: Is this function used any more? We might be able to delete it.

eatyourgreens:
Changes look good to me. I've one minor question, but nothing blocking. I'm happy to merge this then iterate on it, so that it doesn't get in the way of the design updates.

PR #4437, opened 2018-03-09
Featured project toggle

### LINKED ISSUE #4394: Add featured project toggle to admin page
## Expected behavior
Featured project is now a property on the project resource. See https://github.com/zooniverse/Panoptes/pull/2602

## Current behavior
Admin page doesn't have this toggle yet.

simoneduca:
Staging branch URL: https://featured-project-toggle.pfe-preview.zooniverse.org/

Fixes #4394 .

Describe your changes.
Add a checkbox that wires up the project `featured` property.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/admin/featured-project-toggle.jsx
```diff
@@ -0,0 +1,43 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import AutoSave from '../../components/auto-save';
+
+class FeaturedProjectToggle extends Component {
+  constructor() {
+    super();
+    this.handleChange = this.handleChange.bind(this);
```
eatyourgreens: This does nothing, and can be removed.

eatyourgreens:
Not a blocking issue, but this continues the bad practice of modifying component properties, like other lab components. We should stop doing that, because we're just building more technical debt for ourselves to deal with eventually.

simoneduca:
Sorry I meant to change that and forgot. I'll move that local state.

eatyourgreens:
By the way, a controlled component would look something like this, passing in a callback that handles changes in the component's value.
`<FeaturedProjectToggle project={this.state.project} onChange={onProjectChange} />`

app/pages/admin/featured-project-toggle.jsx
```diff
@@ -0,0 +1,46 @@
+import PropTypes from 'prop-types';
+import React, { Component } from 'react';
+import AutoSave from '../../components/auto-save';
+
+class FeaturedProjectToggle extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      project: props.project
```
eatyourgreens: `Object.assign({}, props.project)` here, otherwise changes to `state.project` which actually change `props.project`.

This also means that `state.project` in this component will now be out of sync with `state.project` in the parent, so you'll need to pass the updated project back to the parent.

Updating something in state in two places might not be a great idea, but the current implementation might be limited by the `AutoSave` component, so this isn't a blocking issue.

simoneduca: I feel like we went full circle. Which one do you think would be the best option in this case?

eatyourgreens: Updating state in the parent seems like the ideal here, no, unless I'm missing something? Would that involve removing the `AutoSave`? If so, don't worry about it.

eatyourgreens:
```jsx
<FeaturedProjectToggle project={this.state.project} onChange={onProjectChange} />
```

simoneduca: OK, I've got a proper working controlled component of that form. However the test now is useless. I've also tried to test the handler is called in the parent component, but that unsurprising fails. Should I test just the change of props in the controlled? I'm thinking of deleting its spec file and just leave the rendering test in the parent.

eatyourgreens: Isn't the test that `props.onChange` is called when the box is toggled? I'm not sure I understand.

eatyourgreens:
I just tried to make a project featured, and autosave failed with an error. The error message was that there is already a featured project on staging, but I had to go into the network tools to find that out.

If I want to change the featured project, do I have to search out the current featured project, uncheck it, then find the new project and check it?

srallen:
Yeah, only one featured project at a time per the request: https://github.com/zooniverse/Panoptes/pull/2602

So first, a GET req for `featured: true` and if one comes back, it has to be switched to false before the new one is made true.

simoneduca:
Ah good catch. I've removed the autosave now and state is handled in the parent. Test is commented out, while I figure it out.

simoneduca:
I've updated the featured-project-toggle test but it still isn't passing. I keep getting `TypeError: Cannot read property 'bind' of undefined` no matter what I try, which makes me think that actually the problem is not the test. The new toggle seems to work fine though, so I'm not sure.

simoneduca:
Restaged https://featured-project-toggle.pfe-preview.zooniverse.org/

eatyourgreens:
`props.handleProjectChange` is missing from your test component, which is why you get that error.

app/pages/admin/project-status.jsx
```diff
@@ -52,6 +55,13 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
+  getFeaturedProject() {
+    return apiClient.type('projects').get({ featured: true })
+      .then((featuredProject) => {
```
eatyourgreens: This should be either `(featuredProjects)` to make it clear this is an array of projects, or `([featuredProject])` to get the first project from the array.

simoneduca:
Well, I'm glad that was a quick fix 🤦‍♂️ Jim++

app/pages/admin/project-status.jsx
```diff
@@ -113,6 +123,20 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
+  handleFeaturedProjectChange({ target }) {
+    const { featured, project } = this.state;
+    const featuredProject = featured[0];
+    if (featuredProject) {
+      featuredProject.update({ featured: false });
+      featuredProject.save()
+        .then(() => this.setState({ featured: featuredProject }))
```
eatyourgreens: A couple of points here. `state.featuredProject` was an array before this update but is a project now. This will crash this function the next time it runs. 

Second question, should this update `state.featuredProject` to be the new featured project, not the old one?

app/pages/admin/project-status.jsx
simoneduca: If I `setState` after getting the project with `([featuredProject])`, then the `state.featured` is always a project. So if I understand correctly, your other comment becomes irrelevant as long as I do `([featuredProject])`. I've tested this change and the function doesn't crash.

eatyourgreens: Yeah, if you’re consistent about it being a project then the comment about arrays is irrelevant.

app/pages/admin/featured-project-toggle.jsx
```diff
@@ -0,0 +1,28 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+const FeaturedProjectToggle = ({ handleProjectChange, project }) => {
```
eatyourgreens: You should use `function` instead of `const` to declare components. React uses `Function.name` to get the component display name, which doesn’t work with anonymous functions.

simoneduca:
Restaged https://featured-project-toggle.pfe-preview.zooniverse.org/
Note that the featured project is shown only if it's also `launch_approved`.

eatyourgreens:
The featured project image at https://featured-project-toggle.pfe-preview.zooniverse.org/?env=production has `src="//undefined"`.

app/pages/home-common/featured-project.jsx
```diff
@@ -1,32 +1,42 @@
...
+  }
+  return null;
+}
+
+FeaturedProject.propTypes = {
+  project: PropTypes.shape({
+    avatar_src: PropTypes.string,
+    description: PropTypes.string,
+    display_name: PropTypes.string,
+    links: {
```
eatyourgreens: React is warning that one of the PropType checkers in this component is not a function. I think it's referring to this line. All the other props have functions defined, but this one is an object.

simoneduca:
Restaged https://featured-project-toggle.pfe-preview.zooniverse.org/?env=production

app/pages/home-common/featured-project.jsx
```diff
@@ -1,32 +1,42 @@
...
-    </div>
-    <h2 className="secondary-headline">SCOTUS Notes</h2>
-    <p className="display-body">Behind the scenes at Supreme Court Conference</p>
-    <Link to="projects/zooniverse/scotus-notes-behind-the-scenes-at-supreme-court-conference">View Project!</Link>
-  </section>
-  )
-);
+function FeaturedProject({ project }) {
+  if (project) {
+    const featuredImage = (`//${project.avatar_src}` || (project.links.background && project.links.background.href));
```
eatyourgreens: It’s probably worth adding some tests to check this line performs as expected when `project.avatar_src` is defined and undefined.

eatyourgreens:
Looks good to me on staging. The error I was seeing earlier seems to be the component rendering while `project.avatar_stc` is undefined, so I’d recommend adding a couple of tests for the logic that sets the thumbnail src.

simoneduca:
I think I'm happy with this now. Any more comments?

eatyourgreens:
Why does the code add `//` to `avatar_src`? I'm still seeing the image fail to load when the component first mounts, on staging. There's a noticeable delay until it updates to use the correct URL.

eatyourgreens:
If it helps, the console shows that the component is requesting https://thumbnails.zooniverse.org/999x400/undefined before the image src changes to the avatar URL.

app/pages/home-common/featured-project.jsx
```diff
@@ -1,32 +1,42 @@
...
-    </div>
-    <h2 className="secondary-headline">SCOTUS Notes</h2>
-    <p className="display-body">Behind the scenes at Supreme Court Conference</p>
-    <Link to="projects/zooniverse/scotus-notes-behind-the-scenes-at-supreme-court-conference">View Project!</Link>
-  </section>
-  )
-);
+function FeaturedProject({ project }) {
+  if (project) {
+    const featuredImage = ((project.avatar_src && `//${project.avatar_src}`) || (project.links.background && project.links.background.href));
```
eatyourgreens: `project.links.background.href` isn't an image URL, it's the API URL for the linked `background` resource. If you want to use the background image, you have to request the background from Panoptes, then use `background.src`.

simoneduca: The project background was initially intended to be a fallback when avatar wasn't available. However, since now I'm filtering through `launch_approved` projects now, there no need for a fallback. I'll remove that bit.

simoneduca: Actually, having an avatar isn't a requirement for launching a project, is it? For The Plastic Tide doesn't have one. Should we make it a requirement? Now I'm not sure which image would be safe to use for the featured project. Any thoughts?

simoneduca:
>Why does the code add // to avatar_src? I'm still seeing the image fail to load when the component first mounts, on staging. There's a noticeable delay until it updates to use the correct URL.

If I remove `//` the url doesn't resolve correctly. PFE has two other examples where that seems to be case. [One is in the project-card](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/project-card.cjsx#L28).

app/pages/home-common/featured-project.jsx
eatyourgreens: The Plastic Tide does have an avatar. It's the image that loads on https://featured-project-toggle.pfe-preview.zooniverse.org/?env=production after the initial request to thumbnails.zooniverse.org fails.

eatyourgreens:
Ah right, got you. The `//` is adding the protocol to the avatar URL. It might be better to use `https://` on zooniverse.org. https://www.paulirish.com/2010/the-protocol-relative-url/

app/pages/home-common/featured-project.jsx
simoneduca: Sorry I got that wrong.

eatyourgreens:
Even with the latest changes, `featuredImage` is undefined when the `Thumbnail` component first renders. I'm not sure what's going on there. Do you have any ideas why the avatar isn't showing up immediately?

eatyourgreens:
Tests are passing, so this is a case that isn't caught by the tests.

eatyourgreens:
Debugging locally, `props.project` is defined when the component first renders, but `props.project.avatar_src` is undefined. I don't think there's a test for that case.

simoneduca:
That should fix the delay you were seeing @eatyourgreens

eatyourgreens:
👍

simoneduca:
Restaged https://featured-project-toggle.pfe-preview.zooniverse.org/?env=production

PR #4440, opened 2018-03-12
Use task key to update annotations

eatyourgreens:
Staging branch URL: https://update-annotations.pfe-preview.zooniverse.org

Use `annotation.task` to determine which annotation to update when changing a classification, rather than automatically using the last annotation in the annotations array.

First step towards allowing the classifier to operate on more than one annotation at a time, when making a classification. That would allow us to deprecate combo task annotations, in favour of each task being responsible for updating its own annotation in the annotations array.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -228,15 +228,14 @@ class Classifier extends React.Component {
   }
 
   completeClassification() {
-    const currentAnnotation = this.state.annotations[this.state.annotations.length - 1];
-      if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
-        this.props.onCompleteAndLoadAnotherSubject()
-          .catch(error => console.error(error));
-      } else {
-        this.props.onComplete()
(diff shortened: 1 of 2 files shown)


app/classifier/classifier.jsx
```diff
@@ -221,20 +222,20 @@ class Classifier extends React.Component {
...
-      }
-      this.setState({ annotations: [{}] });
+    if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
+      this.props.onCompleteAndLoadAnotherSubject()
+        .catch(error => console.error(error));
+    } else {
+      this.props.onComplete()
+        .catch(error => console.error(error));
+    }
+    this.setState({ annotations: [{}] });
```
rogerhutchings: The initial annotations state is an empty array - should we set it to that for consistency?

eatyourgreens: I can't remember, but I think I added the empty object to initialise the first annotation.

Issue #4434, opened 2018-03-09
Feedback: Missed classifications with Done&Talk


aprajita:
## Expected behavior
For training subjects (duds or lenses) I expect to see the feedback success or fail window whether I go to done or done&talk...
 
## Current behavior
![image](https://user-images.githubusercontent.com/1917970/37185580-e64cbc70-2338-11e8-93cf-a96e313467c8.png)

... but done&Talk takes me to the talk page without delivering the feedback message.
Returning from talk takes me to the next subject so i never see what the feedback message. 

In the older (non-modal) version of the feedback, this was a risk for duds but not for other training subjects. We thought the risk of losing some dud classifications wasn't ideal but we could live it. Indeed i think roger mentioned that this was a motivation to move to modal feedback that would solve this, but now the risk to miss classifications is for all training images and that seems to be a much larger. See discussion in closed issue [#4078](https://github.com/zooniverse/Panoptes-Front-End/issues/4078) 

given the done&talk button as fa as i understand doesn't register a done but is a link to talk, i am unsure how to remedy this without enabling the classification summary screen which we wanted to avoid.

## Steps to replicate
press done & talk in any training image

## Additional information
FAO @trouille @rogerhutchings 

- **Operating system:**
- **Browser:**

aprajita:
i believe classifications will be missed for any subject not just training subjects - is that right or are the marks recorded in the database when done&talk is pressed?

eatyourgreens:
Same bug as #4433 I think. Feedback should hijack the link to Talk and only follow it after you close the modal.

eatyourgreens:
This should have been fixed by #4450.

rogerhutchings:
@aprajita can you review to see if this is working for you now?

aprajita:
Yes,it is! thanks very much. 
There's just my own human error pressing done immediately after closing the modal automatically without checking that the image is actually different. But i think that's just going to be done once or twice for most users, and then you get used to not making the mistake again.

Issue #4433, opened 2018-03-09
Feedback: subject/feedback mismatch, possible solutions hold subject or change feedback button from 'ok' to 'next'?


aprajita:
## Expected behavior
Feedback message appears on top of the subject it refers to

## Current behavior
for any training image whether success or fail the done button loads the next image **before** the feedback message/modal for the subject it corresponds to is shown

e.g. for the subject here (training dud) 
![image](https://user-images.githubusercontent.com/1917970/37184574-e51c09aa-2333-11e8-8136-5f26a4e46ee0.png)

the corresponding feedback message for this training subject appears overlaid on the next subject
![image](https://user-images.githubusercontent.com/1917970/37184584-eef88e12-2333-11e8-82ce-e1859b0ca325.png)

Potential consequences after returning to the classification screen after pressing OK is that the user thinks the subject shown is the one they just did and proceeds to pressing done, thus missing a classification (I found a couple of times, i didn't recall the image i viewed and I automatically went to done after clicking ok without noticing the underlying image had changed).

This gets even more confusing if that next image is also a training image and no marker placed triggers another feedback message.

Possible remedies:

- For training images done on the screen produces the feedback message but the loading of the next image is triggered by pressing the ok button on the modal feedback message. From earlier issues where we discussed something similar, I realise this isn't something you are comfortable with since the done button then has a different action for training images (to feedback) than non-training (to next subject). I understand that perspective but to me this seems like a reasonable solution to hold the current subject or delay the next subject until the button in the feedback window is pressed  if possible.

- If this isn't possible, I wonder if changing 'ok' to 'next' in the feedback modal is more likely to avoid people skipping he subject - perhaps this is something we can test in the beta or with our smaller team of initial testers/collaborators.

## Steps to replicate
Classify any training image.

## Additional information
FAO: @trouille @rogerhutchings 

- **Operating system:**
- **Browser:**

eatyourgreens:
This is actually the same bug as #4434. Feedback should interrupt the current action (loading the next workflow task, navigating to Talk, loading a new subject) then resume that action when the modal closes. 

This might require me to finish off the work I started on the classification lifecycle, while @rogerhutchings was away.

Also, in workflows with more than one task, we should check what happens when you press Back. I think the current setup might fire off feedback whenever the task changes, going forwards or backwards through the workflow.

aprajita:
ok thanks @eatyourgreens 
[...]

Issue #3679, opened 2017-03-30
"Done & Talk" button loading the next-to-classify subject in Talk instead of the just-classified subject


vrooje:
## Expected behavior

When I click "Done & Talk" in the classification interface, the subject that loads in Talk is the subject I just classified.

## Current behavior

The subject that loads when I click "Done & Talk" seems to be the subject I'm *about to be served next* instead of the one I've just finished by clicking "Done & Talk". 

Not sure if this is always or only when I take these specific steps...

## Steps to replicate

1. Classify a subject (call it subject A) on [Planet 9 (SGL)](https://www.zooniverse.org/projects/skymap/planet-9/classify).
2. When finished classifying, click "Done & Talk" with ctrl/command (depending on OS) held down, so that Talk opens in a new tab (tab a), but the classification tab stays in focus.
3. Don't visit tab a right away; instead classify the new subject (subject B) and click "Done & Talk" to open in a new tab (tab b) again. Subject C will load in the classification interface, but don't classify it.
4. Click on tab a: it should be showing subject A, but for me it's showing subject B.
5. Click on tab b: it should be showing subject B, but for me it's showing subject C, which I haven't classified yet. 

## Additional information

- **Operating system:** Mac OS X El Capitan (10.11.6 (15G1108))
- **Browser:** Chrome Version 56.0.2924.87 (64-bit)

I'm also doing this from my home internet connection, which regularly loses connection for a minute or so at a time. (I need a new router.)

srallen:
Noting that Planet 9 is using the hide classification summaries configuration, so when done or done & talk is clicked, it is saving the classification and then calling to load the next subject. A good guess as to what is happening is that the subject id in the button link is updating with the new subject id before the tab is opened and loads the URL.

eatyourgreens:
Could be a browser bug, since the event handler on the link lets the browser handle the default event (opening the link URL.) Interesting clickjacking technique if the javascript is changing the URL before the browser loads the linked page.

It could also be a bug in `react-router`, since internal links should be intercepted by the router and pushed to location history, rather than reloading the page.

Linked to #3680 since fixing that issue will introduce this bug for the Talk link in the classification summary.

yshish:
It happens to me too on AmazonCam project but only when using chromium browser. It doesn't happen when using Firefox.

mschwamb:
[...]

Issue #4288, opened 2018-01-15
Tutorial is wrapped in a <form> element, but is not a form


eatyourgreens:
## Expected behavior
Tutorials should be marked as regular text.

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
Tutorials are actually wrapped in `<form>` tags. I think this might be a bug in the underlying `modal-form` library, so it may apply to other dialog boxes too.

## Steps to replicate
Inspect a Tutorial with browser dev tools.

srallen:
They're forms by design. It's written to hijack the native browser form submit behavior to control the display of the dialog: https://github.com/zooniverse/modal-form/blob/master/dialog.js#L23-L47

There is a prop to specify a different tag, but I'm not sure if that would have an effect on the open/close state being managed by hijacking the form submission behavior. 

What I'd like to do is rewrite this and the mini-course component in the ZRC repo to be proper modal dialogs with callback props to handle the minor differences between and a prop to specify the display size (mini-course vs standard tutorial).

eatyourgreens:
Ok, I'll leave this open as an accessibility bug since screenreaders like JAWS will probably try to read the tutorial contents as a form.

Sounds like there's some terrible code in that modal library: 😢

srallen:
It's a good example of when to not be clever in trying to solve a problem. Pretty much every other React modal library out there expects you to manage the open/close state in whatever app state you're using rather than being self contained like this is. I'd be happy to redo it with Grommet's Layer component.

eatyourgreens:
That would be excellent. Are you thinking of moving dialogs over to use ZRC instead of modal-form?

eatyourgreens:
Great! I've also flagged up #1953 as zooniverse/Zooniverse-React-Components#51. The solution for that might be to just rewrite the Tutorial component anyway.

srallen:
Just to clarify, I don't think we should write our own basic dialog. Grommet has a Layer, which in version 2 does modals and pull outs. Grommet version one has a tool tip which as of now, I'm expecting to still be available for version 2, but they haven't added it yet. I think that would cover the different modal styles that modal-form does?

wgranger:
I was unaware Grommet had a layer component. This seems like a pretty easy modal to work with, so I might try implementing it in some work for Scribes of the Cairo Geniza.

eatyourgreens:
[...]

PR #4450, opened 2018-03-13
Store workflow task history in Classifier state.

eatyourgreens:
https://update-annotations.pfe-preview.zooniverse.org

Get the current task, and corresponding annotation, from classifier state.
Add handlers for task navigation changes, so that feedback can hook into the Next and Done buttons.

Continuation of #4440, slowly getting rid of all code which would block us from deprecating combo task annotations.

~@rogerhutchings this PR can also add hooks to fire off feedback checks when navigating backward and forward through a workflow, rather than checking for prop changes and assuming that those correspond to navigation buttons being pressed.~

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -243,23 +243,17 @@ class Classifier extends React.Component {
       }
     });
 
+    let onComplete = this.props.onComplete;
     if (this.props.workflow.configuration.hide_classification_summaries && !this.subjectIsGravitySpyGoldStandard()) {
-      this.checkForFeedback(this.state.taskKey)
-        .then(() => {
-          this.props.classification.update({ completed: true });
-          this.setState({ taskKey: null });
-        })
-        .then(this.props.onCompleteAndLoadAnotherSubject)
-        .catch(error => console.error(error));
-    } else {
-      this.checkForFeedback(this.state.taskKey)
-        .then(() => {
-          this.props.classification.update({ completed: true });
-          this.setState({ taskKey: null });
-        })
-        .then(this.props.onComplete)
-        .catch(error => console.error(error));
+      onComplete = this.props.onCompleteAndLoadAnotherSubject;
     }
+    this.checkForFeedback(this.state.taskKey)
+      .then(() => {
+        this.props.classification.update({ completed: true });
+        this.setState({ taskKey: null });
+      })
+      .then(onComplete)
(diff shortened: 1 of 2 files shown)


rogerhutchings:
Okay, tracking the state through `i-fancy-cats`, everything works as expected, and the task key is correctly stored in the state. I did notice that the task ID doesn't change even when technically out of the workflow and on the classification summary, which might be a useful hook?

eatyourgreens:
Good point. I need to split the `onNewTask` handler up into different handlers for Back, Next and Done. Feedback for the current task should be called on next task, before state is updated with the new task key.

Should feedback use promises, so that it interrupts UI changes until the OK button is clicked? I'm thinking it needs to interrupt new subjects loading, or Talk pages loading.

eatyourgreens:
When those changes are made, I think this PR could close #3679, #4433 and #4434.

eatyourgreens:
Feedback checks removed from `componentDidUpdate` and checked instead when you move forward through the workflow, but not back.

TODO: fix focus for the modal's ok button being stolen by the task's Next button.

eatyourgreens:
I kind of hacked the summary step by calling `onNextTask` with no task key. 😕

rogerhutchings:
The feedback modal is created using [ModalFormDialog.alert()](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/features/feedback/classifier/helpers/open-feedback-modal.jsx#L58), which returns a promise 💥

eatyourgreens:
@rogerhutchings `ModalFormDialog` wraps everything in `<form>` so I’d recommend not using it, except for forms. See #4288.

eatyourgreens:
This now blocks the Next and Done buttons while feedback is open. The Done & Talk link still needs some work, and the whole thing needs checking with summaries turned off.

eatyourgreens:
Restaged to work with summaries turned off. Done & Talk still needs to be looked at, but then this should be done.

eatyourgreens:
Now fixes #4433.

eatyourgreens:
Fixes #4434 but hijacks the Done & Talk link, so cmd-click to open in a new tab no longer works.

eatyourgreens:
Cmd-click should work now. Also fixed pushState not being used correctly.

Closes #3679.

Issue #4428, opened 2018-03-07
Interactive User Promotion is broken in UI


amyrebecca:
## Expected behavior
When the user's assigned `workflow_id` is changed to one that doesn't match their `selected_workflow_id` in Panoptes, this change should eventually be reflected in the client-side state.

## Current behavior
The change is never reflected in the client-side state, and in fact the user isn't offered to change workflows until they either reload the page or start a new session.

## Steps to replicate
Use the python client, ruby client, or a caesar rule to change a user's assigned workflow while classifying. Watch the `props` of the `ProjectClassifyPage` component and note that `settings.workflow_id` is never updated, nor is `state.promptWorkflowAssignmentDialog` ever set to true. Note that reloading the page causes the prompt to be displayed immediately. Note that changing `preferences.settings.workflow_id` in a parent of this component DOES cause the correct state update to occur.

## Additional information
Sarah suggests that this update had been performed by one of the `ChangeListener`s in the app and that when it was removed, this side effect was no longer produced. If this is the case, then it's likely this bug has existed for several months. Users of Gravity Spy had reported that the initial promotion was not coming up properly any more but are all eventually getting promoted through new sessions, page reloads, etc. Since we built on this client-side functionality for Snapshot Wisconsin, this behavior is also broken there.

- **Operating system:**
Not an OS-related bug

- **Browser:**
Not a browser bug

amyrebecca:
@marten had at one point made a PR #4116 to replace Sugar, which would be a good candidate for this fix, but I believe it went through pusher which costs us money. The PR was eventually abandoned. We need some kind of mechanism for either polling periodically to get updates to the UPP or receiving notifications when a UPP has changed so that the UI can be updated.

amyrebecca:
To clarify, Sugar currently provides this kind of functionality for notifications and for project completion status but was also written in Coffeescript years ago by someone that is no longer on-staff so it isn't an ideal place to put new code.

marten:
No, Comms started as a hackday project. It's a Sugar-like thing written in Elixir, since Phoenix provides pretty much all we want out of the box, including a JS client library. We tried putting it up on Swarm, but didn't have enough time to get websockets working there. If it's a priority, I'm sure we can try deploying the backend again and get #4116 updated (or write a new one like it).

srallen:
Ok, I wanted to check in in case there was a plan that I was unaware of. Sounds like Sugar is best for the immediate fix.

[...]

PR #4451, opened 2018-03-13
Add UPP request after every 10 classifications

srallen:
This is an ugly hack, but I think it'll work for checking if the user has updated project preferences. This is related to #4428, but I won't tag for closing since I don't think this should be the long term solution.

@amyrebecca could you check to see if this works for Snapshot WI's workflow promotion project?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -25,6 +25,8 @@ auth.listen ->
 
 PROMPT_MINI_COURSE_EVERY = 5
 
+RELOAD_UPP_EVERY = 10
+
 ClassifierWrapper = createReactClass
   displayName: 'ClassifierWrapper'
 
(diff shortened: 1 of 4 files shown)


app/pages/project/classify.cjsx
```diff
@@ -132,6 +132,7 @@ module.exports = createReactClass
           currentClassifications.forWorkflow[props.workflow.id] = classification
           @setState { classification }
         .catch (error) =>
+          # console.error(error)
```
amyrebecca: might as well remove but nothing that blocks merging

app/pages/project/classify.cjsx
```diff
@@ -291,7 +295,7 @@ module.exports = createReactClass
         .then =>
           @setState { promptWorkflowAssignmentDialog: false }
         .then =>
-          if props.preferences.selected_workflow isnt props.preferences.settings.workflow_id
+          if props.preferences.preferences.selected_workflow isnt props.preferences.settings.workflow_id
```
amyrebecca: this was a good catch!

app/classifier/index.cjsx
```diff
@@ -25,6 +25,8 @@ auth.listen ->
 
 PROMPT_MINI_COURSE_EVERY = 5
 
+RELOAD_UPP_EVERY = 10
```
amyrebecca: I'd prefer if this were 5 instead of 10 because I think that Snapshot WI wanted to try 15 as the promotion threshhold, but I understand it's a balancing act wrt API calls

app/pages/project/classify.cjsx
srallen: I left it commented in because I uncovered a bug that was being swallowed by just setting it in the react component state and I was going to come back to it later.

app/classifier/index.cjsx
srallen: Ok, I can make the change.

PR #4463, opened 2018-03-16
Reinstate promoted projects

simoneduca:
Staging branch URL: https://reinstate-promoted-projects.pfe-preview.zooniverse.org/

Fixes # .

Describe your changes.
Add back in the promoted projects section code.
In the interest of getting this merged quickly, I will add tests and fix the featured project on a different PR.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/lib/promoted-projects.js
```diff
@@ -0,0 +1,42 @@
...
+      title: 'Shakespeare\'s World'
+    },
+    1734: {
+      image: '/assets/featured-projects/3098.jpg',
+      title: 'Muon Hunter'
+    }
+  }
+};
+
+const env = process.env.NODE_ENV === 'production' ? 'production' : 'other';
```
wgranger: The app is unable to load when I use the `?env=production` query at the end of the staging url, so maybe some sort of `envFromBrowser = locationMatch(/\W?env=(\w+)/);` variable is in order?

wgranger:
Overall, this looks good. There is just one item that I think should be addressed before a merge. Why was this removed originally? I probably missed that conversation.

app/lib/promoted-projects.js
simoneduca: I think it should work if you `npm start NODE_ENV=production`, but I'm not sure what this is supposed to do tbh. Do you remember what was your intention here?

simoneduca:
It was removed because I mistakenly thought it was old code for the featured project. My bad, sorry.

simoneduca:
This is the PR that removed it https://github.com/zooniverse/Panoptes-Front-End/pull/4437/files

app/lib/promoted-projects.js
wgranger: Yes, that does work. I just know that sometimes when testing staging branches or local work I will apply the `?env=production` ending to the url to see how things appear in production. In this case I did https://reinstate-promoted-projects.pfe-preview.zooniverse.org/?env=production and saw the loading spinner in place of promoted projects because the client is trying to fetch project ids for promoted projects in staging (1731, 1732, 1733, 1734) while pinging the production server.

simoneduca: I see what you, but why is that line need at all?

wgranger: I was thinking the locationMatch line would be used similarly to how it is used in the [split-config](https://github.com/zooniverse/Panoptes-Front-End/blob/595b686eb1a244e924f3df3ff317bba46f3fc551/app/lib/split-config.js#L5) or [pusher-env](https://github.com/zooniverse/Panoptes-Front-End/blob/595b686eb1a244e924f3df3ff317bba46f3fc551/app/lib/pusher-env.js)

simoneduca: I've added it in.

simoneduca:
@wgranger something else I've noticed: Promoted projects seem to show only when logged out. Is that right? If so, why?

wgranger:
I don't think we've ever had a discussion on that. I think the original reasoning was simply because that's how the design/wireframe was constructed.

simoneduca:
I see. I don't see any logical reason for not showing both to all users. Do you? If not, I'll change that in a different PR.

wgranger:
I don't see a glaring reason why it can't be revealed to a logged-in user. I think the original intent was to reduce the noise for logged-in users who were familiar with the site, but I can see a reason why a logged-in user would want to know about new or interesting projects (provided we update the list of promoted projects).

simoneduca:
>provided we update the list of promoted projects

You're right, I think that's an issue. How about only adding featured project for anonymous users instead?

simoneduca:
Either way, that's irrelevant for closing this PR, I think.

wgranger:
Yes, I think so. Let's move the conversation to a channel. I think this PR looks good and is ready to merge.

PR #4477, opened 2018-03-21
Wait for promises to resolve before setting new featured project.

### LINKED ISSUE #4466: Can't set featured project.
## Expected behavior
Being able to set a new feature ject, with no errors, even when one is already set.

## Current behavior
screen shot 2018-03-19 at 11 58 25

## Steps to replicate
Try to set a new feature ject.

simoneduca:
Staging branch URL: https://fix-4466.pfe-preview.zooniverse.org/

Fixes #4466 .

Describe your changes.
Wait for new state to update before setting the new feat project.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/admin/featured-project-toggle.jsx
+++ b/app/pages/admin/featured-project-toggle.jsx
@@ -2,20 +2,28 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 function FeaturedProjectToggle({ error, handleProjectChange, project }) {
+  if (project.launch_approved) {
+    return (
+      <div>
+        <h4>Featured project</h4>
+        <label className="">
+          <input
+            disabled={project.featured}
+            checked={project.featured}
+            name="featured"
+            onChange={handleProjectChange.bind(this)}
+            type="checkbox"
+          />
+          {'Is the project featured?'}
+        </label>
+        <span style={{ color: 'red' }}>{error ? error.message : null}</span>
+      </div>
+    );
+  }
   return (
     <div>
       <h4>Featured project</h4>
-      <label className="">
-        <input
-          disabled={project.featured}
-          checked={project.featured}
-          name="featured"
-          onChange={handleProjectChange.bind(this)}
-          type="checkbox"
-        />
-        {'Is the project featured?'}
-      </label>
-      <span style={{ color: 'red' }}>{error ? error.message : null}</span>
+      <p>{project.display_name} has not been approved for launch, so it can't be featured.</p>
     </div>
   );
 }
--- a/app/pages/admin/featured-project-toggle.jsx
+++ b/app/pages/admin/featured-project-toggle.jsx
@@ -18,7 +18,7 @@ function FeaturedProjectToggle({ error, handleProjectChange, project }) {
       <span style={{ color: 'red' }}>{error ? error.message : null}</span>
     </div>
   );
-};
+}
 
 FeaturedProjectToggle.propTypes = {
   error: PropTypes.shape({
--- a/app/pages/admin/featured-project-toggle.jsx
+++ b/app/pages/admin/featured-project-toggle.jsx
@@ -11,7 +11,7 @@ function FeaturedProjectToggle({ error, handleProjectChange, project }) {
             disabled={project.featured}
             checked={project.featured}
             name="featured"
-            onChange={handleProjectChange.bind(this)}
+            onChange={handleProjectChange}
             type="checkbox"
           />
           {'Is the project featured?'}
--- a/app/pages/admin/featured-project-toggle.spec.js
+++ b/app/pages/admin/featured-project-toggle.spec.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow, mount } from 'enzyme';
+import { shallow } from 'enzyme';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import FeaturedProjectToggle from './featured-project-toggle';
@@ -14,7 +14,7 @@ const project = {
 
 function setup() {
   featuredChangeSpy = sinon.spy();
-  wrapper = mount(
+  wrapper = shallow(
     <FeaturedProjectToggle
       handleProjectChange={featuredChangeSpy}
       project={project}
@@ -31,18 +31,18 @@ describe('Featured Project Toggle', function () {
   beforeEach(setup);
   afterEach(tearDown);
 
-  it('should render the featured checkbox when the project is lauch_approved', function () {
+  it('should render the featured checkbox when the project is launch_approved', function () {
     expect(checkbox).to.have.lengthOf(1);
   });
 
-  it('should render a checked featured check box when the project is lauch_approved and featured', function () {
+  it('should render a checked featured check box when the project is launch_approved and featured', function () {
     wrapper.setProps({
       project: {
         featured: true,
         launch_approved: true
       }
     });
-    expect(checkbox).to.have.lengthOf(1);
+    expect(wrapper.find('input[type="checkbox"][checked=true]')).to.have.lengthOf(1);
   });
 
   it('should call handleProjectChange on checkbox change', function () {
@@ -50,7 +50,7 @@ describe('Featured Project Toggle', function () {
     expect(featuredChangeSpy.called).to.be.true;
   });
 
-  it('should not render the featured checkbox when the project is not lauch_approved', function () {
+  it('should not render the featured checkbox when the project is not launch_approved', function () {
     wrapper.setProps({
       project: {
         featured: false,
--- a/app/pages/admin/featured-project-toggle.spec.js
+++ b/app/pages/admin/featured-project-toggle.spec.js
@@ -1,5 +1,5 @@
 import React from 'react';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import { expect } from 'chai';
 import sinon from 'sinon';
 import FeaturedProjectToggle from './featured-project-toggle';
@@ -8,12 +8,13 @@ let checkbox;
 let featuredChangeSpy;
 let wrapper;
 const project = {
-  featured: false
+  featured: false,
+  launch_approved: true
 };
 
 function setup() {
   featuredChangeSpy = sinon.spy();
-  wrapper = shallow(
+  wrapper = mount(
     <FeaturedProjectToggle
       handleProjectChange={featuredChangeSpy}
       project={project}
@@ -29,8 +30,33 @@ function tearDown() {
 describe('Featured Project Toggle', function () {
   beforeEach(setup);
   afterEach(tearDown);
+
+  it('should render the featured checkbox when the project is lauch_approved', function () {
+    expect(checkbox).to.have.lengthOf(1);
+  });
+
+  it('should render a checked featured check box when the project is lauch_approved and featured', function () {
+    wrapper.setProps({
+      project: {
+        featured: true,
+        launch_approved: true
+      }
+    });
+    expect(checkbox).to.have.lengthOf(1);
+  });
+
   it('should call handleProjectChange on checkbox change', function () {
     checkbox.simulate('change');
     expect(featuredChangeSpy.called).to.be.true;
+  });
+
+  it('should not render the featured checkbox when the project is not lauch_approved', function () {
+    wrapper.setProps({
+      project: {
+        featured: false,
+        launch_approved: false
+      }
+    });
+    expect(wrapper.find('input[name="featured"]')).to.have.lengthOf(0);
   });
 });
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -57,7 +57,7 @@ class ProjectStatus extends Component {
 
   getFeaturedProject() {
     return apiClient.type('projects')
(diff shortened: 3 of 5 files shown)


simoneduca:
Just to be clear, the current code still doesn't fix the issue. To test:
1. Set I Fancy Cat as featured.
2. Try to set martens-testproject-staging as new featured.
3. Sometimes that works, sometimes it throws `featured has already been taken`.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
 
+  setFeatured(project, value) {
+    project.update({ featured: value.checked });
+    return project.save()
+      .catch(error => this.setState({ error }));
+  }
+
   handleFeaturedProjectChange({ target }) {
     const { featured, project } = this.state;
     if (featured) {
```
eatyourgreens: This is going to be false while the initial API call is waiting to resolve, so while you don’t know yet if there’s a featured project.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
       featured.update({ featured: false });
-      featured.save()
+      return featured.save()
         .then(() => this.setState({ featured }))
-        .catch(error => this.setState({ error }));
+        .catch(error => this.setState({ error }))
+        .then(this.setFeatured(project, target));
+    } else {
+      return this.setFeatured(project, target);
```
eatyourgreens: There must always be a featured project in the database. If you get here, something is wrong.

simoneduca: Panoptes allows no featured projects being set, so technically it's not wrong, I suppose.

simoneduca:
I'm not sure what's going on. I still see the `featured has already been taken` error, even if I don't render the checkbox until the request that sets the featured project has finished. This makes me think the problem isn't with clicking the checkbox before the previous project is "un-featured". Also I just broke the tests.

simoneduca:
I've removed the `loading` flag as it doesn't help solving the problem.

simoneduca:
This seems to be working now. Restaged https://fix-4466.pfe-preview.zooniverse.org/admin/project_status?filterBy=launch_approved
Note for testing:  projects should be filtered by `launch_approved=true`.

app/pages/home-for-user/index.jsx
```diff
@@ -70,7 +70,7 @@ export default class HomePageForUser extends React.Component {
   }
 
   getFeaturedProject() {
-    const query = { featured: true, launch_approved: true, cards: true };
+    const query = { featured: true, launch_approved: true };
```
eatyourgreens: Won’t this remove the project avatar from the response, re-introducing the problem we had on the original PR, where the featured project image doesn’t load?

simoneduca: Sorry, I'll add it back in there.

app/pages/admin/featured-project-toggle.jsx
```diff
@@ -2,23 +2,31 @@ import PropTypes from 'prop-types';
...
+  if (project.launch_approved) {
+    return (
+      <div>
+        <h4>Featured project</h4>
+        <label className="">
+          <input
+            disabled={project.featured}
+            checked={project.featured}
+            name="featured"
+            onChange={handleProjectChange.bind(this)}
```
eatyourgreens: What is `.bind(this)` doing here?

simoneduca: I thought that since the function deals with events needed to be bound. But it looks like it isn't needed after all.

eatyourgreens: It would bind the function to `FeaturedProjectToggle` but that’s not where the function is invoked, so it confused me. I think it should be bound before it is passed down as a prop.

simoneduca: It is [bound in the parent component](https://github.com/zooniverse/Panoptes-Front-End/blob/8ed377c0288e2e62ea65251121de0b7980b95a3e/app/pages/admin/project-status.jsx#L29). That's why it's not needed in `FeaturedProjectToggle`, thanks for spotting it.

simoneduca:
Thanks to @camallen this works now. I'm fixing the tests.

simoneduca:
This should be ready to go now.

app/pages/admin/featured-project-toggle.spec.js
```diff
@@ -8,12 +8,13 @@ let checkbox;
...
 const project = {
-  featured: false
+  featured: false,
+  launch_approved: true
 };
 
 function setup() {
   featuredChangeSpy = sinon.spy();
-  wrapper = shallow(
+  wrapper = mount(
```
eatyourgreens: `FeaturedProjectToggle` is a stateless function, with no children, so why is it using `mount`?

app/pages/admin/featured-project-toggle.spec.js
```diff
@@ -29,8 +30,33 @@ function tearDown() {
...
+  });
+
+  it('should render a checked featured check box when the project is lauch_approved and featured', function () {
+    wrapper.setProps({
+      project: {
+        featured: true,
+        launch_approved: true
+      }
+    });
+    expect(checkbox).to.have.lengthOf(1);
```
eatyourgreens: This doesn't test if the checkbox is checked.

app/pages/admin/featured-project-toggle.spec.js
```diff
@@ -29,8 +30,33 @@ function tearDown() {
...
+  });
+
   it('should call handleProjectChange on checkbox change', function () {
     checkbox.simulate('change');
     expect(featuredChangeSpy.called).to.be.true;
   });
+
+  it('should not render the featured checkbox when the project is not lauch_approved', function () {
```
eatyourgreens: Typo in `launch_approved` here.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
+  saveProject(project) {
```
eatyourgreens: This function seems like overkill. All it does is call `project.save()`.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
   handleFeaturedProjectChange({ target }) {
-    const { featured, project } = this.state;
-    if (featured) {
-      featured.update({ featured: false });
-      featured.save()
-        .then(() => this.setState({ featured }))
+    const { featuredProject, project } = this.state;
+    project.update({ featured: target.checked });
+    if (featuredProject) {
+      featuredProject.update({ featured: false });
```
eatyourgreens: Maybe return `featuredProject.update({ featured: false }).save()`?

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
+      return this.saveProject(featuredProject)
```
eatyourgreens: You can chain the save onto the update here.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
+        .then(() => this.saveProject(project))
+        .then(() => this.setState({ featuredProject: project }))
```
eatyourgreens: `.save()` resolves with the updated resource from Panoptes so `.save().then( project => ...)`

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
+
   handleFeaturedProjectChange({ target }) {
-    const { featured, project } = this.state;
-    if (featured) {
-      featured.update({ featured: false });
-      featured.save()
-        .then(() => this.setState({ featured }))
+    const { featuredProject, project } = this.state;
+    project.update({ featured: target.checked });
+    if (featuredProject) {
```
eatyourgreens: This will be true even if `featuredProject.featured` is false.

app/pages/admin/project-status.jsx
```diff
@@ -124,17 +124,23 @@ class ProjectStatus extends Component {
...
+      featuredProject.update({ featured: false });
+      return this.saveProject(featuredProject)
+        .then(() => this.saveProject(project))
+        .then(() => this.setState({ featuredProject: project }))
         .catch(error => this.setState({ error }));
+    } else {
+      return this.saveProject(project);
```
eatyourgreens: Is this line still needed?

app/pages/admin/project-status.jsx
```diff
@@ -210,7 +216,7 @@ class ProjectStatus extends Component {
   }
 
   render() {
-    const { error, project } = this.state;
+    const { error, inProgress, project } = this.state;
```
eatyourgreens: Where does `inProgress` get set?

simoneduca: Just a sad leftover.

app/pages/admin/featured-project-toggle.spec.js
simoneduca: Oops sorry.

app/pages/admin/featured-project-toggle.spec.js
simoneduca: My bad, sorry

app/pages/admin/project-status.jsx
simoneduca: Yeas, but I'm checking if a featured project exists at all.

app/pages/admin/project-status.jsx
simoneduca: It's useful if somehow no featured project is set. The DB allows that.

eatyourgreens: You can get here if a featured project is set in Panoptes, but it hasn't been set in the component state yet. In that case, if it arises, the save here will fail with an error.

app/pages/admin/project-status.jsx
eatyourgreens: In that case, you will update the featured project even when `featuredProject.featured` is false.

simoneduca: I don't think so, because [`getFeaturedProject`](https://github.com/zooniverse/Panoptes-Front-End/blob/4c8da67db77815b9ce7900a80dbf5fbb093a1be5/app/pages/admin/project-status.jsx#L60) add a project to local state only if `featured: true`.

app/pages/admin/project-status.jsx
simoneduca: Sorry I'm not seeing how. This is only called when clicking on the checkbox. By then the `componentDidMount` will have set the state, no? Do you think there's a scenario where it's possible to click on it before that happens?

eatyourgreens: `featuredProject` is initially null, so that's the case I was thinking of. Is it also null after errors? Just reading through the code, it looked like the logic allows you to save here in circumstances where `featuredProject` isn't determined, and you might end up with an uncaught error.

app/pages/admin/project-status.jsx
```diff
@@ -125,16 +125,19 @@ class ProjectStatus extends Component {
...
-    const { featured, project } = this.state;
-    if (featured) {
-      featured.update({ featured: false });
-      featured.save()
-        .then(() => this.setState({ featured }))
+    const { featuredProject, project } = this.state;
+    project.update({ featured: target.checked });
+    if (featuredProject) {
+      return featuredProject.update({ featured: false }).save()
+        .then(() => project.save())
```
eatyourgreens: If this save (or the previous one) throws an error, then `state.featuredProject.featured` is false, and `project.featured` is true, but I think the checkbox might still be active. I'm not sure.

I'm not sure how big a deal error checking is, or how the UI is intended to react if Panoptes is having problems. I don't think any of the other admin form elements have great error handling either.

simoneduca: I'm going to catch the error before `setState`, which should help with your point. The UI just consists of the error message being displayed next to the checkbox. Also when Panoptes has problems, e.g. screenshot below
screen shot 2018-04-03 at 10 40 54

simoneduca: Just waiting for checks to pass, before merging.

eatyourgreens: Remember that a then block after a catch always runs. In the event of an error, it will receive the return value of the catch block.

eatyourgreens: The latest commit sets `state.featuredProject`, even after an error occurs. I'm not sure what it sets it to after an error, so the component could get into a weird state there. No big deal if refreshing the page clears the error.

simoneduca: It does.

simoneduca: But I'll remove the latest commit, just because it's confusing/has the wrong message.

eatyourgreens: Yeah, if the code is going to continue and set state after an error, then the error handler should probably reverse any unsaved changes.

PR #4481, opened 2018-03-22
Store text task value in component state

eatyourgreens:
https://fix-text-input.pfe-preview.zooniverse.org/

Update text task value in local state while typing. Update the annotation on blur, or when the component unmounts.

This fixes performance problems by updating the classifier less frequently. However, testing on the dev classifier shows that it also breaks the Next button for required tasks.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/text/index.cjsx
+++ b/app/classifier/tasks/text/index.cjsx
@@ -58,7 +58,7 @@ module.exports = createReactClass
   getInitialState: ->
     textareaHeight: undefined
     initOffsetHeight: undefined
-    value: ''
+    value: @props.annotation.value
 
   componentDidMount: ->
     @setState initOffsetHeight: @refs.textInput.offsetHeight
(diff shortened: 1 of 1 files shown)


rogerhutchings:
Also, if the component doesn't unmount, old values are left in state for the next workflow step :(

eatyourgreens:
The value isn't remembered if 'persist annotations' is turned on, either. 😢

eatyourgreens:
I've staged some changes that might fix those issues. Not sure about handling required tasks though.

mcbouslog:
Initial question from first pass, if there are back to back text tasks, where component will not unmount, do we need to add an additional lifecycle check and `updateAnnotation`? Maybe `componentWillUpdate`.

eatyourgreens:
2881765 takes care of initialising back-to-back text tasks. I'm not sure about updating the annotation. We didn't come up with any problems while doing some quick testing, but it's worth double-checking.

PR #4484, opened 2018-03-26
[WIP]: Classifier: Don't update classification prop directly

eatyourgreens:
Staging branch URL: https://update-annotations.pfe-preview.zooniverse.org/

Removes any task or drawing tool code that changed `classification.annotations`. Instead, tasks and drawing tools update `state.annotations` in the classifier. The classification is updated when complete and then submitted.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Tagging this WIP until I'm sure what it breaks.

eatyourgreens:
Feedback seems to work with these changes. 'Hide previous marks' still worked when I tried it. No problems on a quick pass through the dev classifier, but it was very quick.

eatyourgreens:
Hmmm, tagging text seems to have broken for the text input task, but also broken on master so probably not caused by this PR.

eatyourgreens:
I've gone through a few combo tasks at https://update-annotations.pfe-preview.zooniverse.org/projects/markb-panoptes/focus-testing-project Classifications seemed to work, and annotations persisted properly when moving backward and forward through a workflow.

eatyourgreens:
I'll double check the model renderer. That definitely sounds like it should have changed.

I skipped the summary because I think that should be displaying `classification.annotations`. Or at least, `classification.annotations` is no longer being updated at that point and can be trusted as a source of data.

This PR is mostly about catching the components that were changing annotations by reference, indirectly updating classifier state without explicitly calling `setState` anywhere.

eatyourgreens:
Oh, I also think that old components that change the classification indirectly will actually stop working in this PR, because I'm trying to copy the annotations to a new array and break the reference back too the classification.

eatyourgreens:
I haven't been able to find anything that's broken by this PR, which makes me nervous that I've missed something.

eatyourgreens:
@mcbouslog's recursive workflow also works on this branch: https://update-annotations.pfe-preview.zooniverse.org/projects/markb-panoptes/test-project-2-mb/classify

PR #4490, opened 2018-03-29
Remove object.values from subject-viewer

tingard:
Fixes incompatibility with internet explorer.

# Review Checklist

- [ ] Does it work on mobile?

# Optional


srallen:
Should we add explicitly this to the polyfill service so that this doesn't cause issues in the future? We might not remember that this isn't polyfilled by the default setting from polyfill.io and someone might try to use it.

eatyourgreens:
Good idea. I _think_ we polyfill `Object.keys` already, and it doesn't make sense to support that but not `Object.values`.

PR #4513, opened 2018-04-12
Add cached string to voice over

tingard:
I can't commit to the main repo, but here's the cached code.


tingard:
^^ you have two days 😂

PR #4528, opened 2018-04-18
Galaxy builder rendering changes

tingard:
Makes changes to galaxy-builder project following reviews from various collaborators.

- Fixed sizing issue if change frame when zoomed in
- Fixed zero-size issue when canvas-viewer is the first subject (needs to call `onLoad` with desired sizes)
- Enabled adding of custom messages to a text-box positioned at the bottom of the canvas
- Changed scaling of residuals  / models to better deal with noisy data
- Changed score scaling (mainly to enable a nonzero score for noisy data)


app/components/file-viewer/canvas-viewer.jsx
```diff
@@ -143,12 +147,12 @@ class CanvasViewer extends React.Component {
           style={Object.assign({}, this.props.style, this.state.canvasStyle)}
         />
         {
-          this.state.hasScore && this.state.score !== null &&
+          this.state.hasMessage && this.state.message !== null &&
           <span
-            ref={(r) => { this.scoreSpan = r; }}
-            className="canvas-renderer-score"
+            ref={(r) => { this.messageSpan = r; }}
```
eatyourgreens: Is `this.messageSpan` used? I don't see it anywhere else in this PR.

app/features/modelling/galaxy-builder/index.js
```diff
@@ -127,25 +134,26 @@ class GalaxyBuilderModel extends baseModel {
...
     }
+    let message = '';
     if (this.state.shouldCompareToImage) {
       this.calculateDifference({
         texture: this.state.pixels,
         imageTexture: this.imageData
       });
       this.state.pixels({ copy: true });
-      this.eventHandlers.setScore(this.getScore());
+      message += `Difference Image - Score: ${this.getScore()}`;
```
eatyourgreens: Now that we have translations, will this component need to be localised?

app/components/file-viewer/canvas-viewer.jsx
tingard: ah well spotted - removed

app/features/modelling/galaxy-builder/index.js
eatyourgreens: Here's Galaxy Builder in Dutch: https://www.zooniverse.org/projects/tingard/galaxy-builder?language=nl

Localised text is stored in `/app/locales` then translated using `counterpart` or `react-translate-component`.  Here's an example that uses a templated string (and allows HTML content): https://github.com/zooniverse/Panoptes-Front-End/blob/1171e88d228a2ffdca02747886da2b32d37a6e8b/app/pages/project/home/talk-status.jsx#L33-L40

If Galaxy Builder is only ever going to be in English, then there's no pressing need to translate it.

tingard: I think the lifetime of the project is such that the text can be left in English. If needed I can work more on translating it post-launch and after my placement.

tingard: How would you go about making a non-english project (i.e. about page etc...)? Thinking more for galaxy zoo than builder.

tingard:
I no longer have write access, so someone else will have to press the magical button.

app/features/modelling/galaxy-builder/index.js
eatyourgreens: We're testing out a translations editor at https://pandora.zooniverse.org. It shows you existing text (created by the project owner) and lets you create new versions for each resource (project, workflow, tutorial, field guide, project page.)

eatyourgreens:
Weird. You've definitely got read/write access to this repo.

PR #4543, opened 2018-04-24
React 16 upgrade

srallen:
Staging branch URL: https://react-16-upgrade.pfe-preview.zooniverse.org/

Branch on the main repo version of #4324. ~According to the original PR, the only thing left is updating react-swipe.~ React-swipe is on its latest version and it is using React 16 as a dependency. 

It would be helpful whenever someone has time to try out this branch and test a few projects. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
The project home page locks up, and eventually kills the browser tab that it's running in. I'm looking into this.

eatyourgreens:
If I remove the `ProjectNavBar` component from projects, they load fine. With that component, CPU usage shoots up to 100% (Chrome) and the page never renders.

Could there be a problem with styled components on this branch?

srallen:
It could be or it could be react-sizes which is used to dynamically determine the breakpoint between the mobile version and the desktop version of the navigation. I'll take a look.

eatyourgreens:
I've tracked it down to `setBreakPoint` in the navbar. Remove that function and everything is fine, but you lose the narrow navigation bar.

srallen:
Ok, it's likely `react-rendered-size` dependency. We might be able to figure out a better way to do what either of those do and remove the dependency. I did look into that a bit back in January, but didn't get very far. @rogerhutchings wrote this bit originally. @rogerhutchings, do you have any suggestions?

eatyourgreens:
Why are we using a library when there are standard DOM methods to get the rendered size of an element?

Also, why render the wide nav bar twice in the DOM? Surely it's faster to render it once (hidden),  measure it then either make it visible or replace it with the narrow version.

eatyourgreens:
I looked in the code for `react-rendered-size` and it's returning `offsetHeight` and `offsetWidth`.
```js
var size = {
    height: renderedElement.offsetHeight,
    width: renderedElement.offsetWidth
  };
```
No idea why it's using 100% CPU though.

srallen:
I spent a bit of time yesterday afternoon looking this over and reminding myself what exactly is happening with the dynamic setting of the breakpoint. Rendering each navbar hidden on mount, measuring the wide one, and then display the appropriate navbar isn't sufficient here because the width of the navbar contents can change depending on a few variables: sign in or sign out, if the user can toggle admin mode on or off, and if the project title changes (though I'm not totally understanding the purpose of this one since navigating to a new project would be an unmount and new mount of the navbar). So it makes sense to be able to render a separate hidden wide navbar to measure and then update the breakpoint as new props are received and then trigger a re-render. 

I'm still playing around with React 16. I'm wondering if between the new Fragments and Portals API if I can't come up with another way to do this.

eatyourgreens:
I don’t like the idea of the breakpoint varying between projects on the same device. For a given viewport width, I think the better UX is to always have the same navigation menu (dropdown or list) for every project. Maybe we just have a fixed breakpoint for the dropdown menu?

rogerhutchings:
> width of the navbar contents can change depending on a few variables: sign in or sign out, if the user can toggle admin mode on or off, and if the project title changes (though I'm not totally understanding the purpose of this one since navigating to a new project would be an unmount and new mount of the navbar)

I remember why I did this now - if you switch languages, the title translation could be a different length to the previous one but wouldn't trigger a remount

As it stands, I think the current design necessitates a dynamic breakpoint. Any changes required to permit a static breakpoint are outside the scope of this PR, and would need @beckyrother's input (things like changing nav item behaviour and truncating project titles)

eatyourgreens:
I need to investigate more, but `_.map` and `_.difference` in `havNavLinksChanged` might be causing the 100% CPU usage. Maybe there's a circular reference in there somewhere.

eatyourgreens:
#4567 needs to be checked on this branch. If my guess is right, then it has the same bug.

eatyourgreens:
`npm ls react` shows that Markdownz and SevenTen are still using React 15 on this branch.

beckyrother:
I agree with @eatyourgreens – it's confusing for people and bad UX to go between projects and experience different breakpoints. I'm in favor of setting a consistent break point across projects – maybe either at 1200 or 960 since those are both common screen sizes.

This does not solve the current problem, but maybe setting a character limit for project titles would be something to consider to get ahead of this in the future.

srallen:
I've restaged this now with the crashing bug fix.

eatyourgreens:
No problems here doing some drawing task classifications and commenting on Talk.

srallen:
`modal-form` is crashing the whole site on this branch when you click on the user dropdown menu in the header:
```
Uncaught TypeError: Cannot read property 'style' of undefined
    at StickyModalForm.reposition
```

eatyourgreens:
The drawing task details form also crashes the page (same error.)

eatyourgreens:
Tutorials and mini-courses are positioned off-screen when they first open. Probably the same `modal-form` bug.

eatyourgreens:
I've updated that modal form PR and it throws the same errors that we see here. Refs that were defined with React 15 are showing as undefined in React 16, immediately after a component mounts.

eatyourgreens:
Good news: I've installed zooniverse/modal-form#react-16 on a local copy of this branch and modal forms all work again (user account menu, drawing subtasks, tutorials and field guides.) 🎉 

Bad news: tutorial content is rendered with a height and width of 0px, resulting in blank tutorial slides.
Screenshot showing a blank tutorial for The American Soldier project.

Resizing the window, while the tutorial is open, renders the tutorial content. I think this is caused by `react-swipe`.

srallen:
Scrolling will get the tutorial to appear too.

eatyourgreens:
Hmmm. Could be these listeners, which force a re-render of the modal:
https://github.com/zooniverse/modal-form/blob/10f02cfc3ee17bea313395ad50adb47a15ca8ff5/base.js#L76-L77
🤔

srallen:
`this.form` might be undefined when it's first called so it never gets positioned when it first mounts? https://github.com/zooniverse/modal-form/blob/10f02cfc3ee17bea313395ad50adb47a15ca8ff5/base.js#L160

srallen:
I just realized I forgot to update my local copy to 2.6.0 before testing 🤦‍♀️ . The behavior is a bit different on 2.6.0, so it might not be what I was looking at.

eatyourgreens:
The tutorial and `StepThrough` components both use `this.refs`, which may behave differently in React 16.

srallen:
The mini-course component displays the content and it is not using StepThrough, though it's not rendering at the correct width. You can see this with my test project, http://localhost:3735/projects/srallen086/workflow-assignment-testing/classify

srallen:
I've merged the latest fixes for the tutorial. Hopefully that is it for the modal-form fixes.e

srallen:
I've rebase this and had to regenerate the package-lock.json since checking out the file from master and npm installing didn't work to resolve the conflict.

srallen:
I've tested adding a comment in talk as well as editing content in the project lab, and markdownz seems to be working fine. We should probably update it to React 16 at some point anyway, but it isn't a blocker. 

The lab is throwing warnings about `isMounted` being deprecated since it has the last few uses of the PromiseRenderer and ChangeListener.

eatyourgreens:
zooniverse/Seven-Ten-Client#3 and zooniverse/markdownz#67 should remove those last couple of dependencies on React 15. I don't think there's anything in either of those packages that breaks with React 16 but it's worth double-checking.

srallen:
I've ran npm dedupe and restaged this.

srallen:
Let's hold off on merging this until after SGL.

eatyourgreens:
Shall we merge this once SGL is done tomorrow?

srallen:
I think I'd prefer the Tuesday after the holiday. I'm out this Friday and Monday is a holiday for all of us.

PR #4550, opened 2018-04-25
Galaxy builder bugfixes

tingard:
Fixes #4542

Describe your changes.
Added error catching for canvas-viewer where text is displayed rather than any image. Howerver, volunteers could still hypothetically complete a classification.

Added a dynamic scaling to the maximum number of points in a spiral arm. This will not fix mobile classifying, as PSF convolution requires a minimum of 121 uniforms, which significantly eats into the 256 most mobile devices allow.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4565, opened 2018-04-27
Changed galaxy-builder webgl error message to be more specific and helpful

tingard:
Addresses concerns raised in  #4550 

Changed error message from 
> We’re afraid your browser doesn’t support the WebGL we need to render galaxies. You can comment on photos (the second image) in Talk, but won’t be able to classify.

To
> Sorry, this project requires features which are not supported by your device. Please try again on a desktop computer, and let us know in talk if you're still having problems. You can comment on photos (the second image) in Talk, but please do not classify!

### DIFF
--- a/app/features/modelling/baseReglModel.jsx
+++ b/app/features/modelling/baseReglModel.jsx
@@ -1,8 +1,9 @@
 import reglBase from 'regl';
 
 class BaseReglModel {
-  constructor(canvas, { frame, metadata, src, sizing }, eventHandlers) {
+  constructor(canvas, { frame, metadata, src, sizing }, eventHandlers, errMessage) {
     this.eventHandlers = eventHandlers;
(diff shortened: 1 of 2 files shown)


tingard:
Let me just make sure this works before merging

tingard:
Okay we're good

eatyourgreens:
Minor point but “your device” may well be a desktop computer. One of the people reporting problems on Talk gave Windows XP as their OS.

tingard:
@astopy as one of the messages is being called during the `super` function (inside `setModel`, which is called by the constructor of the parent) , I can't define a constant on `this` as I'd like. I could pass it as a param to super but that seems messy. Open to suggestions!

Good point re. device assumptions, maybe "try again on a different computer" instead?

astopy:
Yeah that sounds good.

PR #4567, opened 2018-04-27
Remove react-rendered-size

rogerhutchings:
Staging branch URL: https://react-measure.pfe-preview.zooniverse.org/

Uses `react-sizeme` instead of `react-rendered-size` to determine which nav bar to show on the project home page.

The current design specifies that the contents of the nav bar change, which necessitates a dynamic breakpoint - let's get that working for React 16 (#4543 ), and move the discussion of whether that's actually a good idea to a separate issue.

This still needs the tests finishing, but is enough to test in browser.

Still to do:

- [ ] Finish tests

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
+++ b/app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
@@ -1,76 +1,57 @@
 import _ from 'lodash';
-import getRenderedSize from 'react-rendered-size';
 import React, { Component } from 'react';
-import PropTypes from 'prop-types';
-import withSizes from 'react-sizes';
+import sizeMe from 'react-sizeme'
 import ProjectNavbarNarrow from './components/ProjectNavbarNarrow';
-import ProjectNavbarWide from './components/ProjectNavbarWide';
+import ProjectNavbarWide, { SizeAwareProjectNavbarWide } from './components/ProjectNavbarWide';
 
 function haveNavLinksChanged(oldProps, newProps) {
-  const oldNavLinks = _.map(oldProps.navLinks, 'label');
-  const newNavLinks = _.map(newProps.navLinks, 'label');
-  return _.difference(newNavLinks, oldNavLinks).length > 0 || oldNavLinks.length !== newNavLinks.length;
-}
-
-function hasTitleChanged(oldProps, newProps) {
-  return oldProps.projectTitle !== newProps.projectTitle;
+  const oldLinks = _.map(oldProps.navLinks, 'label');
+  const newLinks = _.map(newProps.navLinks, 'label');
+  return _.difference(oldLinks, newLinks).length > 0 ||
+    oldLinks.length !== newLinks.length;
 }
 
 export class ProjectNavbar extends Component {
   constructor(props) {
     super(props);
+    this.setBreakpoint = this.setBreakpoint.bind(this);
     this.state = {
-      breakpoint: 0
+      useWide: false
     };
   }
 
-  componentDidMount() {
-    this.setBreakpoint();
-  }
-
   componentDidUpdate(prevProps) {
-    if (haveNavLinksChanged(prevProps, this.props) ||
-      hasTitleChanged(prevProps, this.props)) {
+    if (haveNavLinksChanged(prevProps, this.props)) {
       this.setBreakpoint();
     }
   }
 
-  setBreakpoint() {
-    const sizes = getRenderedSize(
-      <ProjectNavbarWide
-        {...this.props}
-        style={{
-          position: 'absolute',
-          visibility: 'hidden'
-        }}
-      />
-    );
-
-    this.setState({
-      breakpoint: sizes.width
-    });
+  setBreakpoint(size) {
+    // `size` is undefined when the component is first mounted, as there hasn't
+    // been time for the callback to fire.
+    if (size) {
+      const useWide = size.width < document.body.clientWidth;
+      this.setState({ useWide });
+    }
   }
 
   render() {
-    const isWindowWide = this.props.width > this.state.breakpoint;
-    if (isWindowWide) {
-      return <ProjectNavbarWide {...this.props} />;
-    }
+    const NavBarComponent = (this.state.useWide) ? ProjectNavbarWide : ProjectNavbarNarrow;
 
-    return <ProjectNavbarNarrow {...this.props} />;
+    return (
+      <div>
+        <NavBarComponent {...this.props} />
+        <SizeAwareProjectNavbarWide
+          {...this.props}
+          onSize={this.setBreakpoint}
+          style={{
+            visibility: 'hidden',
+            position: 'absolute'
+          }}
+        />
+      </div>
+    );
   }
 }
 
-ProjectNavbar.propTypes = {
-  navLinks: PropTypes.arrayOf(PropTypes.shape({
-    label: PropTypes.string
-  })),
-  projectTitle: PropTypes.string,
-  width: PropTypes.number
-};
-
-const mapSizesToProps = ({ width }) => ({
-  width
-});
-
-export default withSizes(mapSizesToProps)(ProjectNavbar);
+export default ProjectNavbar;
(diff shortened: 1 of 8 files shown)


app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
```diff
@@ -1,76 +1,57 @@
...
+
+    return (
+      <div>
+        <NavBarComponent {...this.props} />
+        <SizeAwareProjectNavbarWide
+          {...this.props}
+          onSize={this.setBreakpoint}
+          style={{
+            visibility: 'hidden',
+            position: 'absolute'
```
eatyourgreens: If `useWide` is true, I think you should remove these styles and show this nav bar, rather than rendering it twice in the DOM.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
```diff
@@ -1,76 +1,57 @@
...
-    );
-
-    this.setState({
-      breakpoint: sizes.width
-    });
+  setBreakpoint(size) {
+    // `size` is undefined when the component is first mounted, as there hasn't
+    // been time for the callback to fire.
+    if (size) {
+      const useWide = size.width < document.body.clientWidth;
```
eatyourgreens: If you set up a ref to the wide navbar component's DOM node, I think this check can just be `const useWide = this.navbarElement.offsetWidth < document.body.clientWidth`.

eatyourgreens:
The document also uses `<header>` for the top-level zooniverse site navigation. I'm not sure if a HTML document is allowed to have two headers. Two `<nav>` elements should be fine though.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
eatyourgreens: If I remove these styles, the visible navbar goes into narrow mode, even if there's enough room to render the wide navbar. I think this happens because of a couple of things:
 - `position: absolute` roughly halves the rendered size of the project navbar (see screenshots).
 - `document.body.clientWidth` is smaller than the viewport, so will always be smaller than an element that complete fills the window.

Here's what the size-aware component looks like, if we remove `visibility: hidden`.
Without `position: absolute`:
Screenshot showing both project navbars rendered one above the other.

With `position: absolute`.
Screenshot showing how absolute positiion reduces the width of the navigation.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
```diff
@@ -1,76 +1,57 @@
...
 
 function haveNavLinksChanged(oldProps, newProps) {
-  const oldNavLinks = _.map(oldProps.navLinks, 'label');
-  const newNavLinks = _.map(newProps.navLinks, 'label');
-  return _.difference(newNavLinks, oldNavLinks).length > 0 || oldNavLinks.length !== newNavLinks.length;
-}
-
-function hasTitleChanged(oldProps, newProps) {
-  return oldProps.projectTitle !== newProps.projectTitle;
+  const oldLinks = _.map(oldProps.navLinks, 'label');
```
eatyourgreens: I think this would be much easier to read if it used `Array.map`.
`const oldLinks = oldProps.navLinks.map(link => link.label);`

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
```diff
@@ -1,76 +1,57 @@
...
+  const newLinks = _.map(newProps.navLinks, 'label');
```
eatyourgreens: Ditto regarding array maps.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
eatyourgreens: Should this check be `size.width <= document.body.clientWidth`? That might explain the weird behaviour in my screenshots when the wide navbar exactly fills the screen.

srallen:
@eatyourgreens, since @rogerhutchings is out until mid next week, I'll take over making changes in this PR so that this doesn't get stale. I'll first get the tests working.

srallen:
According to [MDN](https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines), there can be multiple headers, but the intention I think is a single page header, then section headers if needed:

> The HTML Header Element (<header>) defines a page header — typically containing the logo and name of the site and possibly a horizontal menu — or section header, containing perhaps the section's heading, author name, etc. <article>, <section>, <aside>, and <nav> can have their own <header>. Despite its name, it is not necessarily positioned at the beginning of the page or section.

eatyourgreens:
I was thinking of landmark navigation in browsers. I'm not sure anyone supports multiple headers, but I'd need to double check that. It's outside the scope of this PR, but I noticed it while debugging the navigation.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
srallen: Using a separate rendering of the navbar makes sense to me for the purposes of dynamically setting the breakpoint as new props are received. Absolute positioning it gives us the width of the children. I'm not sure why Space Warps was collapsing into the narrow version here. My guess is because the default state is `useWide: false` and the callback to get the size measurement hadn't updated it yet. I've updated this branch to have a loading state rather than show a default of the narrow navbar which seems to have fixed this. I was seeing a flash of it before, but I think the component needs to wait for the setBreakpoint callback to fire first.

srallen:
I think this is ready for final review. Deciding on dynamic breakpoints vs static will answer the question about the double rendering of the wide navbar variant and I think outside of scope?

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
eatyourgreens: When the narrow version was rendered, `document.body.clientWidth` was 1280 and the wide navbar width was also 1280, so the dropdown is being shown if the wide navbar exactly fills the screen. This is the missing `=` in the comparison that I picked up in a separate comment.

If I set `width: 99%` on the second navbar, the first example renders the wide navbar too.

eatyourgreens: It's not obvious from the code that we're measuring the minimum width of the wide navbar and comparing that with the document inner width. I thought that the code was measuring the total space that the wide navbar would take up if it was rendered as the project navigation. Better variable naming might make this clearer to someone coming to this code for the first time.

eatyourgreens:
👍 This fixes the 100% CPU bug which hangs the React 16 branch on any project.

I've got a couple of comments but nothing that should block merging.

- it adds a  new dependency when all the code does is compare the navbar's [`offsetWidth`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetWidth) with the inner width of the document body. We might be able to use `element.offsetWidth` directly in the code (or better, [`element.scrollWidth`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollWidth).)
- the code isn't measuring the full width that the navbar would take up when rendered and maybe this should be made explicit.

app/pages/project/components/ProjectNavbar/ProjectNavbar.jsx
```diff
@@ -1,76 +1,85 @@
...
-
-    this.setState({
-      breakpoint: sizes.width
-    });
+  setBreakpoint(size) {
+    // `size` is undefined when the component is first mounted, as there hasn't
+    // been time for the callback to fire.
+    
+    if (size) {
+      const useWide = size.width < document.body.clientWidth;
```
eatyourgreens: `useWide` is false here if the navigation exactly fills the document body.

srallen:
I agree that this could be refactored to not use a dependency, but I'd rather that be in a separate PR and merge this into the React 16 branch so we can keep testing it.

PR #4602, opened 2018-05-04
Workflow Selection: parse IDs as numbers

eatyourgreens:
Parse `workflow_id` as an integer before using it in comparisons, or storing it in user preferences. Fixes a bug where a malformed ID could cause an infinite loop of requests to `/api/workflows`.

Staging branch URL: https://workflow-selection.pfe-preview.zooniverse.org/

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -23,8 +23,9 @@ class WorkflowSelection extends React.Component {
       nextProps.preferences.preferences.selected_workflow &&
       this.state.workflow
     ) {
+      const userWorkflowID = parseInt(nextProps.preferences.preferences.selected_workflow);
       if (!nextState.loadingSelectedWorkflow &&
-        nextProps.preferences.preferences.selected_workflow !== this.state.workflow.id
+        userWorkflowID !== parseInt(this.state.workflow.id)
       ) {
         this.getSelectedWorkflow(nextProps);
       }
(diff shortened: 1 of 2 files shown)


app/pages/project/workflow-selection.jsx
```diff
@@ -23,8 +23,9 @@ class WorkflowSelection extends React.Component {
       nextProps.preferences.preferences.selected_workflow &&
       this.state.workflow
     ) {
+      const userWorkflowID = parseInt(nextProps.preferences.preferences.selected_workflow);
       if (!nextState.loadingSelectedWorkflow &&
-        nextProps.preferences.preferences.selected_workflow !== this.state.workflow.id
```
marten: Is this `parseInt` really necessary? Shouldn't it come back as an int already, since JSON has data types?

eatyourgreens: When I reproduced the bug, the stored workflow ID was coming back from the preferences as `6605env=production` so I think this is definitely needed.

marten: No I meant on the `parseInt(this.state.workflow.id)`. Isn't `state.workflow` filled with an object coming from the Panoptes API?

eatyourgreens: I think that compares `6605 !== ‘6605’`. Resource IDs are strings, for some reason.

eatyourgreens:
Hmmm, tests need fixing.

eatyourgreens:
Tests highlighted that I hadn't been very thorough with type conversion, so I've rebased and all workflow IDs should now be passed around the WorkflowSelection component as ints.

app/pages/project/workflow-selection.jsx
eatyourgreens: I've tested this now. If I take out that `parseInt`, the exact comparison continuously fails and the component spams the API with requests.

eatyourgreens: Also, looking at the network responses, resource IDs are sent back as strings from Panoptes.

marten:
I'm wondering if this should come with a new test example. Right now there isn't a specific example that documents why we need to convert these (what the problem was with the old way). Other than that, or if you think that's fine left out, LGTM

eatyourgreens:
Good point. I've added a couple of tests.

PR #4604, opened 2018-05-04
Split up the survey task chooser

eatyourgreens:
Staging branch URL: https://survey-task.pfe-preview.zooniverse.org/

Splits the survey task chooser up into a CharacteristicsFilter and Choices component.

This also fixes a bug where you can't tab into the filtered choice list if your currently selected choice isn't present in the filtered list (every item is set to tabindex -1.)

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/survey/chooser/CharacteristicsFilter.jsx
+++ b/app/classifier/tasks/survey/chooser/CharacteristicsFilter.jsx
@@ -1,0 +1,125 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import TriggeredModalForm from 'modal-form/triggered';
+import Translate from 'react-translate-component';
+import counterpart from 'counterpart';
+
+class CharacteristicsFilter extends React.Component {
+  handleFilter(characteristicId, valueId) {
+    this.props.onFilter(characteristicId, valueId);
+  }
+
+  render() {
+    const { task, translation, filters } = this.props;
+    return (
+      <div className="survey-task-chooser-characteristics">
+        {task.characteristicsOrder.map((characteristicId, i) => {
+          const characteristic = task.characteristics[characteristicId];
+          const selectedValue = characteristic.values[filters[characteristicId]];
+          const characteristicStrings = translation.characteristics[characteristicId];
+          const selectedValueStrings = characteristicStrings.values[filters[characteristicId]];
+          let hasBeenAutoFocused = false;
+          return (
+            <TriggeredModalForm
+              key={characteristicId}
+              ref={`${characteristicId}-dropdown`}
+              className="survey-task-chooser-characteristic-menu"
+              triggerProps={{ autoFocus: i === 0 && !this.props.focusedChoice }}
+              trigger={
+                <span className="survey-task-chooser-characteristic" data-is-active={!!selectedValue}>
+                  <span className="survey-task-chooser-characteristic-label">
+                    {selectedValue ? selectedValueStrings.label : characteristicStrings.label}
+                  </span>
+                </span>
+                }
+            >
+              <div className="survey-task-chooser-characteristic-menu-container">
+                {characteristic.valuesOrder.map((valueId) => {
+                  const value = characteristic.values[valueId];
+                  const valueStrings = characteristicStrings.values[valueId];
+                  const disabled = (valueId === filters[characteristicId]);
+                  const autoFocus = (!disabled && !hasBeenAutoFocused);
+                  const selected = (valueId === filters[characteristicId]);
+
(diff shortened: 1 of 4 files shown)


eatyourgreens:
I thought this might make the survey task chooser a little easier to deal with. I've left the original logic for determining the choices alone, just tried to split up the rendering method of the original component.

app/classifier/tasks/survey/chooser/Chooser.jsx
```diff
@@ -0,0 +1,123 @@
...
+          <Translate
+            content="tasks.survey.showing"
+            with={{ count: sortedFilteredChoices.length, max: task.choicesOrder.length }}
+          />
+          &ensp;
+          <button
+            type="button"
+            className="survey-task-chooser-characteristic-clear-button"
+            disabled={Object.keys(filters).length === 0}
+            onClick={this.handleClearFilters}
```
eatyourgreens: `this.handleClearFilters` isn't defined in this component any more.

srallen:
I'll take a look at this. This is a good first step for the redesign since the filter UI is what's going to change the most, so pulling that out into its own component is a good idea.

eatyourgreens:
I found a bug where all the choices might get set to tabindex -1, so I’ve fixed that here too. The tabindex code isn’t great, to be honest.

app/classifier/tasks/survey/chooser/Choices.jsx
```diff
@@ -0,0 +1,183 @@
...
+        this.choiceButtons[newIndex].focus();
+        e.preventDefault();
+        break;
+      default:
+    }
+  }
+
+  render() {
+    const { filteredChoices, task, translation } = this.props;
+    this.choiceButtons = [];
```
srallen: I know this was in the original code, but it's confusing to me to have this reset to an empty array on render. Not a blocker for this, but something to consider for future refactors.

app/classifier/tasks/survey/chooser/Choices.jsx
```diff
@@ -0,0 +1,183 @@
...
+          const choice = task.choices[choiceId];
+          const chosenAlready = selectedChoices.indexOf(choiceId) > -1;
+          let tabIndex = -1;
+          if (i === 0 && choiceNotPresent) {
+            tabIndex = 0;
+          }
+          if (choiceId === this.props.focusedChoice) {
+            tabIndex = 0;
+          }
+          const src = this.props.task.images[choice.images[0]];
```
srallen: Should this have a default value too?

app/classifier/tasks/survey/chooser/Chooser.jsx
```diff
@@ -0,0 +1,113 @@
...
+import Choices from './Choices';
+
+class Chooser extends React.Component {
+
+  getFilteredChoices() {
+    return this.props.task.choicesOrder.map((choiceId) => {
+      const choice = this.props.task.choices[choiceId];
+      let rejected = false;
+      Object.keys(this.props.filters).map((characteristicId) => {
+        const valueId = this.props.filters[characteristicId];
```
srallen: Same question about default value.

app/classifier/tasks/survey/chooser/Choices.jsx
eatyourgreens: If you don’t reset your refs on each render, the array will just grow and grow and grow. There’s probably a better way to do this.

app/classifier/tasks/survey/chooser/Choices.jsx
eatyourgreens: The default value here should be `undefined`, if you don’t want the `Thumbnail` component to generate an img tag.

app/classifier/tasks/survey/chooser/CharacteristicsFilter.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+  render() {
+    const { task, translation, filters } = this.props;
+    return (
+      <div className="survey-task-chooser-characteristics">
+        {task.characteristicsOrder.map((characteristicId, i) => {
+          const characteristic = task.characteristics[characteristicId];
+          const selectedValue = characteristic.values[filters[characteristicId]];
+          const characteristicStrings = translation.characteristics[characteristicId];
+          const selectedValueStrings = characteristicStrings.values[filters[characteristicId]];
```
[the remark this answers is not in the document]
eatyourgreens: It looks to me like this would break if `characteristic` was null or undefined.

PR #4580, opened 2018-05-01
Add deleteIfOutOfBounds to drawing tool root

eatyourgreens:
Staging branch URL: https://delete-out-of-bounds.pfe-preview.zooniverse.org

Fixes #4579 .

Check deleteIfOutOfBounds on every drawing tool update and delete the current tool if outside the subject.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4608, opened 2018-05-07
Revert "Add deleteIfOutOfBounds to drawing tool root"

### LINKED ISSUE #4579: Out-of-bounds marks can be submitted in a completed classification
## Expected behavior
Drawing tasks that lie outside the subject image area should be automatically detected and deleted.

## Current behavior
Marks that are created by a mouseup/touchend event outside the subject image are deleted. However, if the mouseup/touchend event that creates the mark is inside the subject, the mark is not deleted, even if its final (x,y) coordinates lie outside the subject area.

## Steps to replicate
Draw a point and drag it quickly upwards to fling it outside the subject. Lift your finger (or lift the mouse button) while still inside the subject. It's possible for the subject to end up outside the image area, but the mark is not deleted and remains in the submitted classification.

## Additional information

- **Operating system:** n/a
- **Browser:** n/a

### LINKED ISSUE #4607: New PW: Some marks disappear when that part of image is panned away from the screen
## Expected behavior
When I mark a group of penguins and pan away to mark another group, I expect to find the placed marks exactly where I left them. Also, I expect the mark count to stay on the same number and not start decreasing while panning.

## Current behavior
After marking a group of penguins while panning away to mark another group, I see the mark count decreasing and after panning back to the first group I've marked already, I find out some marks are gone. Especially when zoomed in, especially when the marks are near an image edge.

Screenshot made after panning the image away with 30 adult marks placed all over the image - you can see the count is now saying 19 adults only:
ysh_2018-05-05_13-44-28

## Steps to replicate
The best way to test is to just go along the left image edge and place marks in a row along the edge randomly. Then to pan away to the right side without zooming and watching the total mark count. If nothing happens, then zooming in usually makes them do so. And sometimes even all marks disappear.

## Additional information
Quite annoying to figure out that after you had marked several hundred of distant penguins and after panning back and finding some of them unmarked you marked them all again... and again... and then to find the total mark count showing much less than before. 
Also, I haven't found a way to mark penguins standing along the edge, when zoomed in.

- **Operating system:**
Linux x86_64
- **Browser:**
Chromium 63

shaunanoordin:
## PR Overview
Reverts zooniverse/Panoptes-Front-End#4580

This reversion removes the `deleteIfOutOfBounds` logic in the `markings-renderer.cjsx` component, which was introduced to fix Issue #4579. (i.e. users could place markings outside of the Subject bounds, noted on Space Warps.)

Unfortunately, the introduction of `deleteIfOutOfBounds` introduced a _new, more serious issue,_ where if a user zoomed into Subject image, any markings that _are outside the field of view_ start to disappear. (i.e. the problem is that `deleteIfOutOfBounds` checks for the bounds of the _subject viewer/viewport,_ not the _subject image itself._ )

This newly introduced issue has significant impact on projects like Penguin Watch (see issue #4607), so we need to undo these changes.

### Dev Notes
- This should fix #4607 
- This should re-open #4579 
- A proper solution would be to,
  1. fix the deleteIfOutOfBounds logic (after confirming that it's not doing what it's _supposed to do._ )
  2. reapply the code from PR #4580 to re-fix #4579.
- ...but that would require mucking about in old undocumented PFE logic, so let's do that later.
- For now, REVERT!

There is a separate, though more minor, issue where if a user zooms into a Subject image, then clicks on Reset Zoom Levels, the markings _appear_ to disappear, but are actually shrunk to something like 5% size. This is mostly a presentational/UI issue, and isn't as severe as what we're trying to solve right now.

### Status
@eatyourgreens @srallen please give this a check.

### DIFF
--- a/app/classifier/tasks/drawing/markings-renderer.cjsx
+++ b/app/classifier/tasks/drawing/markings-renderer.cjsx
@@ -1,13 +1,10 @@
 React = require 'react'
 createReactClass = require 'create-react-class'
 drawingTools = require '../../drawing-tools'
-deleteIfOutOfBounds = require '../../drawing-tools/delete-if-out-of-bounds'
 strategies = require('../../../features/feedback/shared/strategies').default
 
 module.exports = createReactClass
   displayName: 'MarkingsRenderer'
-
-  tools: []
 
   getDefaultProps: ->
     annotations: []
@@ -38,13 +35,8 @@ module.exports = createReactClass
     @setState oldSetOfMarks: newSetOfMarks
     # console.log 'Marks are now', newSetOfMarks
 
-  componentDidUpdate: () ->
-    for tool in this.tools
-      deleteIfOutOfBounds(tool)
-
   render: ->
     skippedMarks = 0
-    this.tools = []
     <g>
       {for annotation in @props.annotations
         annotation._key ?= Math.random()
@@ -120,13 +112,7 @@ module.exports = createReactClass
                 getScreenCurrentTransformationMatrix: @props.getScreenCurrentTransformationMatrix
 
               ToolComponent = drawingTools[toolDescription.type]
-              <ToolComponent
-                ref={(tool) => this.tools.push(tool) if tool}
-                key={mark._key}
-                {...toolProps}
-                {...toolEnv}
-                {...toolMethods}
-              />}
+              <ToolComponent key={mark._key} {...toolProps} {...toolEnv} {...toolMethods} />}
           </g>}
     </g>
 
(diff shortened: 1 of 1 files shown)


eatyourgreens:
The solution to this would be to change this line so that it gets the bounding client rectangle from the subject image, not the SVG rectangle that’s used to draw marks. 

That rectangle then gets passed to every tool as the `containerRect` prop. With that change, the changes in #4580 so be safe to add. 

https://github.com/zooniverse/Panoptes-Front-End/blob/d9834aa0b47b7f13053c59bf7dd133c162370923/app/classifier/annotation-renderer/svg.jsx#L19

PR #4623, opened 2018-05-15
Fix highlighter styling

srallen:
Staging branch URL: https://fix-highlighter-styling.pfe-preview.zooniverse.org/

Fixes the styling for the highlighter tool. I pulled the button code out into its own component so I could target it with styled-components and the classifier theme. I wasn't sure how the buttons worked and checked out the commit before the theming and saw that it previously didn't get active styles. I did a slight modification to add some component state to track which button is being actively selected. I hope this works with how the highlighter tool is intended to work, which I'm still not totally clear about.

I've left a to do comment that the focus and hover styles while the button has the active class for some reason aren't working. I haven't fixed it since I'm not aware of any project that is using this in production yet.

In a future PR, the highlighter should be added to the dev classifier.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
In `TaskInputField`, you define rules for `.active` after rules for `:hover` and `:focus`. Pseudoclasses have the same specificity as classes, so `.active` overrides the hover and focus styles. The solution is probably to define `:hover` and `:focus` rules after `.active`.

eatyourgreens:
Ignore that, I missed this block: https://github.com/zooniverse/Panoptes-Front-End/blob/67ffa3f973182e0669a0cf9d86722d6394438d62/app/classifier/tasks/components/TaskInputField/TaskInputField.jsx#L101-L110

I'm not sure why those styles aren't being applied.

eatyourgreens:
One minor comment. The highlighter buttons aren’t radio buttons so they don’t have a selected state. They should only be active when being clicked.

srallen:
You're right, the focus, hover, and active styles are being applied, but at least for me just hard to see with the browser outline. I've also removed the active class from being applied.

PR #4628, opened 2018-05-18
handle a deactivated workflow being found in user project preferences

camallen:
Updates to the workflow selection code to avoid continuous looping and data requests on the API,  specifically:
 + When attempting to select a workflow ensure we test the IDs using the same type by using a function.
 + Instead of using a search query to find the workflow resource, get the resource via the ID and manually test the active state on the returned resource if it exists.

Staging branch URL: https://get-workflow-resource-not-search.pfe-preview.zooniverse.org/

##### Steps to reproduce:
+ visit https://www.zooniverse.org/projects/tkotwim/the-american-soldier
+ update your user project preferences to have an inactive selected_workflow
```
const upp = zooAPI.type('project_preferences').get({ user_id: '1', project_id: '5551'})
upp.update({ 'preferences.selected_workflow': '6627' }).save();
```
+ reload the project page to see the workflow query loop https://www.zooniverse.org/projects/tkotwim/the-american-soldier

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -99,16 +99,11 @@ class WorkflowSelection extends React.Component {
       }
     })
     .then((workflow) => {
-      let activeWorkflow;
-      if (activeFilter && workflow && workflow.active) {
-        activeWorkflow = workflow;
-      } else {
-        activeWorkflow = null;
-      }
-
-      if (activeWorkflow) {
-        this.setState({ loadingSelectedWorkflow: false, activeWorkflow });
-        actions.translations.load('workflow', activeWorkflow.id, translations.locale);
+      const allowedActiveWorkflow = activeFilter && workflow && workflow.active
+      const ignoreActiveFilter = !activeFilter && workflow
+      if (allowedActiveWorkflow || ignoreActiveFilter) {
+        this.setState({ loadingSelectedWorkflow: false, workflow });
+        actions.translations.load('workflow', workflow.id, translations.locale);
       } else {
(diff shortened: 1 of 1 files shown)


eatyourgreens:
I think the code should also check that the workflow project matches `props.project.id`.

app/pages/project/workflow-selection.jsx
```diff
@@ -97,14 +98,19 @@ class WorkflowSelection extends React.Component {
...
     })
-    .then(([workflow]) => {
-      if (workflow) {
+    .then((workflow) => {
+      const workflowMatchesProjectID = this.idsMatch(workflow.links.project, this.props.project.id)
+      const allowedActiveWorkflow = activeFilter && workflow && workflow.active
+      const ignoreActiveFilter = !activeFilter && workflow
+      const loadableWorkflow = allowedActiveWorkflow || ignoreActiveFilter
+
+      if (workflowMatchesProjectID && loadableWorkflow) {
```
camallen: I'm not 100% sure about the consequence of not loading if the project doesn't match. From my understanding this will call itself again and try to pick a workflow that will match. It may get stuck in a loop as this happened when i was testing this, e.g. set the `workflowMatchesProjectID = false` but this is contrived and i expect that at least 1 workflow will be loaded correctly.

eatyourgreens: I'm not able to get workflows to load, since this change was committed.

camallen:
superseded by #4630

PR #4631, opened 2018-05-21
Rewrite workflow selection

eatyourgreens:
- [x] Sanitise workflow ID strings from URLs
- [x] Sanitise workflow ID strings from preferences
- [x] remove query params from the workflow API request
- [x] add tests for malformed query params
- [x] add tests to verify that preferences are cleared correctly
- [x] validate workflow IDs against project links

Staging branch URL: https://workflow-selection.pfe-preview.zooniverse.org

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -77,13 +77,9 @@ class WorkflowSelection extends React.Component {
 
   getWorkflow(selectedWorkflowID, activeFilter = true) {
     const { actions, translations } = this.props;
-    const query = {};
-    if (activeFilter) {
-      query.active = true;
-    }
     apiClient
     .type('workflows')
-    .get(selectedWorkflowID.toString(), query)
+    .get(selectedWorkflowID.toString(), {}) // the empty query here forces the client to bypass its internal cache
     .catch((error) => {
       if (error.status === 404) {
         this.clearInactiveWorkflow(selectedWorkflowID)
@@ -97,6 +93,16 @@ class WorkflowSelection extends React.Component {
       if (workflow) {
         const isWorkflowForProject = workflow.links.project === this.props.project.id;
         return isWorkflowForProject ? workflow : null;
+      }
+      return null;
+    })
+    .then ((workflow) => {
+      if (workflow) {
+        if (activeFilter) {
+          return workflow.active ? workflow : null;
+        } else {
+          return workflow;
+        }
       }
       return null;
     })
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -76,36 +76,33 @@ class WorkflowSelection extends React.Component {
   }
 
   getWorkflow(selectedWorkflowID, activeFilter = true) {
-    const { actions, translations } = this.props;
-    apiClient
-    .type('workflows')
-    .get(selectedWorkflowID.toString(), {}) // the empty query here forces the client to bypass its internal cache
-    .catch((error) => {
-      if (error.status === 404) {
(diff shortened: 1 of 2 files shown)


eatyourgreens:
This tries to avoid workflow selection loops by sanitising IDs that are read from the URL or ffrom preferences. Code is still pretty ugly though. 😞

eatyourgreens:
@camallen There's a couple of tests that pass strings like `4random` to `getSelectedWorkflow`, and check that `getWorkflow` is called with `4` as the workflow ID.
https://github.com/zooniverse/Panoptes-Front-End/blob/27e7e4e2e0bb698606a4d5c585405645e6827814/app/pages/project/workflow-selection.spec.js#L209-L214

`clearInactiveWorkflows` could probably do with tests, to catch bugs like the one that you found on Friday.

Maybe also some tests that check whether the code is robust against the ID being a number or a string, just in case.

camallen:
I'd go with what you have. I think we're fine if someone passes in garbage, we parse to int then string but the UI doesn't load the correct workflow is ok.

eatyourgreens:
Cool. I think NfN is the only project that passes in workflow IDs via the URL, so the only one where a malformed ID might be submitted, but that's also one where you really don't want workflow selection to spam the API either.

app/pages/project/workflow-selection.jsx
```diff
@@ -143,13 +153,13 @@ class WorkflowSelection extends React.Component {
 
   clearInactiveWorkflow(selectedWorkflowID) {
     const { preferences } = this.props;
-    const selectedWorkflow = preferences.preferences ? preferences.preferences.selected_workflow : undefined;
-    const projectSetWorkflow = preferences.settings ? preferences.settings.workflow_id : undefined;
+    const selectedWorkflow = (preferences && preferences.preferences) ? this.sanitiseID(preferences.preferences.selected_workflow) : undefined;
```
eatyourgreens: Should this be called `userSelectedWorkflow` for clarity, and for consistency with the same block earlier in the component?

eatyourgreens: Also, I think it would be better if values were sanitised in one place and the sanitised values stored in the instance for use by other component methods.

app/pages/project/workflow-selection.jsx
```diff
@@ -96,6 +89,23 @@ class WorkflowSelection extends React.Component {
...
+    .then ((workflow) => {
+      if (workflow) {
+        if (activeFilter) {
+          return workflow.active ? workflow : null;
+        } else {
+          return workflow;
+        }
+      }
+      return null;
+    })
```
srallen: The project resource's `links.active_workflows` could be used for the activeness check and the project link check before requesting the workflow. We could reduce unnecessary requests this way. I checked Notes from Nature and I see 139 id strings in that serialized property. The `links.workflows` property has 161, so I don't think we're running into max length limits.

eatyourgreens: Good idea. It doesn't make a lot of sense to download the entire workflow, then throw it away.

eatyourgreens: If the project owner is testing an inactive workflow, should we just skip the checks completely?

srallen: Yes, I think owners, collaborators, expert, tester, and admins should be able to skip activeness check. We'll probably still want to check the workflow belongs to the project though.

eatyourgreens: I've added a check that compares with either `project.links.workflows` or `project.links.active_workflows`, depending on `activeFilter`.

PR #4635, opened 2018-05-22
Convert ProjectPageController to ES6

eatyourgreens:
Staging branch URL: https://project-es6.pfe-preview.zooniverse.org

Runs `/app/pages/project/index.cjsx` through [decaffeinate](https://decaffeinate-project.org), then tidies up the resulting code and converts it from `createReactClass` to an ES6 class.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Tests are passing, and the code seems to be working on the staging branch, so I think this ready for review.

app/pages/project/index.jsx
```diff
@@ -0,0 +1,428 @@
...
+      Split.clear();
+      this.setState({ splits: null });
+      (this.context.geordi != null ? this.context.geordi.forget(['experiment', 'cohort']) : undefined);
+    }
+  }
+
+  getUserProjectPreferences(project, user) {
+    this.listenToPreferences(null);
+
+    const userPreferences = (user != null) ?
```
srallen: I'm seeing explicit conditional checks against null all throughout the code. We don't typically write existential checks this way when writing new components and the decaffeinate comment at the top of the file suggests changing this if it makes sense. Perhaps a find and replace for all the `!= null` to remove?

app/pages/project/index.jsx
```diff
@@ -0,0 +1,428 @@
...
+            this.setState({ background, organization, owner, pages, projectAvatar, projectIsComplete, projectRoles, preferences, ready });
+            this.loadFieldGuide(project.id);
+            this.props.actions.translations.loadTranslations('project_page', pages.map(page => page.id), this.props.translations.locale);
+          })
+          .catch((error) => {
+            this.setState({ error });
+            console.error(error);
+          });
+        } else {
+          this.setState({
```
srallen: Out of scope of a conversion PR, but does the code ever get to this else? Wouldn't a 404 be caught by the catch of the Promise?

app/pages/project/index.jsx
```diff
@@ -0,0 +1,428 @@
...
+  loadFieldGuide(projectId) {
+    return apiClient.type('field_guides').get({ project_id: projectId })
+    .then(([guide]) => {
+      const { actions, translations } = this.props;
+      this.setState({ guide });
+      actions.translations.load('field_guide', guide != null ? guide.id : undefined, translations.locale);
+      getAllLinked(guide, 'attached_images')
+      .then((images) => {
+        const guideIcons = {};
+        images.map(image => guideIcons[image.id] = image);
```
srallen: This could be a `forEach` instead of `map`

app/pages/project/index.jsx
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  render() {
+    const slug = `${this.props.params.owner}/${this.props.params.name}`;
+    const betaApproved = this.state.project != null ? this.state.project.beta_approved : undefined;
+    const launchApproved = this.state.project != null ? this.state.project.launch_approved : undefined;
+
+    return (
+      <div className="project-page-wrapper">
+        <Helmet title={`${(this.state.project != null ? this.state.project.display_name : undefined) != null ? (this.state.project != null ? this.state.project.display_name : undefined) : counterpart('loading')}`} />
```
srallen: The ternary in the title prop is hard to read and understand.

app/pages/project/index.jsx
eatyourgreens: Beats me, but the render method does check for the `NOT_FOUND` error message which implies the code does end up here if the project doesn't exist.

I think this might be using an API route that returns an array of projects, rather than a specific project, in which case it won't 404.

app/pages/project/index.jsx
```diff
@@ -0,0 +1,428 @@
...
+        if (!splits) {
+          return;
+        }
+        Object.keys(splits).forEach((split) => {
+          // log the first active split and break the loop
+          if (splits[split].state === 'active' && this.context.geordi) {
+            const experiment = splits[split].name;
+            const cohort = splits[split].variant ? splits[split].variant.name : undefined;
+            this.context.geordi.remember({ experiment, cohort });
+            return;
```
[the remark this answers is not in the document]
eatyourgreens: This return breaks the loop after the first active split. I added a comment on line 98 to explain this function but I could make that more explicit if necessary.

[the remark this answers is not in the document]
eatyourgreens: I should also add a test to make sure that this really does do that.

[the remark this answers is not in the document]
eatyourgreens: Looking back over it, this function totally does not do what I thought it did.

eatyourgreens:
Rebased and restaged, now that tests are working again.

eatyourgreens:
@srallen those `null` checks should all be gone now, and I've rewritten the `setupSplits` method so that it should be logging active splits properly.

eatyourgreens:
Should be fixed now.

PR #4640, opened 2018-05-23
Updating publications page

hrspiers:
Updating publications page to include SGL publication


hrspiers:
@srallen no I don't think so - Chris asked directly for this to be added to the publications page without asking to delay for an announcement, so I'm going to merge now. Thanks for reviewing!

PR #4646, opened 2018-05-24
Strip HTML from blog excerpts

astopy:
Staging branch URL: https://remove-blog-tags.pfe-preview.zooniverse.org/

Fixes https://www.zooniverse.org/talk/17/638676

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/home-common/social.jsx
+++ b/app/pages/home-common/social.jsx
@@ -79,7 +79,7 @@ export default class HomePageSocial extends React.Component {
         <div className="home-social__blog-post">
           <div style={background}></div>
           <div>
-            <span className="regular-body">{post.excerpt}</span>
+            <span className="regular-body">{post.excerpt.replace(/<.+?>/g, '')}</span>
             <br />
             <div>
               <a className="home-social__italic-link" href={post.link}> Read More... </a>
--- a/app/pages/home-common/social.jsx
+++ b/app/pages/home-common/social.jsx
@@ -6,6 +6,7 @@ import counterpart from 'counterpart';
 import moment from 'moment';
 import ProjectCard from '../../partials/project-card';
 import { getPublication, getRecentProjects, getBlogPosts, getNewestProject } from '../../lib/get-social-data';
+import striptags from 'striptags';
 
 counterpart.registerTranslations('en', {
   socialHomePage: {
@@ -79,7 +80,7 @@ export default class HomePageSocial extends React.Component {
         <div className="home-social__blog-post">
           <div style={background}></div>
           <div>
-            <span className="regular-body">{post.excerpt.replace(/<.+?>/g, '')}</span>
+            <span className="regular-body">{striptags(post.excerpt)}</span>
             <br />
             <div>
               <a className="home-social__italic-link" href={post.link}> Read More... </a>
--- a/package-lock.json
+++ b/package-lock.json
@@ -8980,6 +8980,11 @@
       "integrity": "sha1-HhX7ysl9Pumb8tc7TGVrCCu6+5E=",
       "dev": true
     },
+    "striptags": {
+      "version": "2.2.1",
+      "resolved": "https://registry.npmjs.org/striptags/-/striptags-2.2.1.tgz",
+      "integrity": "sha1-TEULcI1BuL85zyTEn/I0/Gqr/TI="
+    },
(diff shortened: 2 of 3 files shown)


shaunanoordin:
1. Question/confirmation: The intent of this PR is to strip all HTML tags in a Talk post, correct?

For example if I posted:
```
Doctors <b>hate</b> this one weird trick, <br> click <a href="http://totally-not-a-trap.com">here</a> to learn more.
```

...I should only see:
```
Doctors hate this one weird trick,  click here to learn more.
```

2. Observation: the Regex seems to be a little too greedy(? hungry? all-consuming?) when matching `<` and `>` pairs, and will capture even `<` and `>` that are, e.g., used for mathematical expressions instead of HTML tags.

For example, this text:
```
If n < 4 but n > 2, then X happens.
```

...will be trimmed to:
```
If n 2, then X happens.
```

...even though `< 4 but n >` is not a valid HTML element.

astopy:
> The intent of this PR is to strip all HTML tags in a Talk post, correct?

Nope, it's to strip HTML from the blog posts on the home page:

screen shot 2018-05-24 at 14 43 15

All that content is coming from Wordpress, which should already have replaced any greater/less than signs with HTML entities.

shaunanoordin:
👍 Thanks for the clarification on the PR intent.

Suggestion: here's a potential improvement on the Regex that captures HTML tags. (warning: I'm probably missing out on some edge cases)
```
/<\/?\w[\w\d_]*(?:\s*[\w\-]+(?:=".*"|='.*')?\s*)*\s*\/?>/ig

<\/?\w[\w\d_]* 
- catches tag starting points, e.g. '<div...', '</div...', '<h3...', '<weird_custom_tag_with_underscores_I_guess...'
- does not catch LT arrows standing on their own, e.g. '1 < 3'

(?:\s*[\w\-]+(?:=".*"|='.*')?\s*)*
- catches attributes, e.g. ' href="#" ', ' href='#' ', 'disabled'
- No idea what happens when you have invalid attribute forms (so please don't use this RegEx for security sensitive stuff.) e.g. '<span class="<script src='lol.js'></script>">'

\s*\/?>
- catches the tag ending points, e.g. '... >' or '... />'

Known limitations:
- Doesn't know how to deal with namespace, e.g. '<h:table>'
- Catches '</nonsense/>' when it shouldn't.
- No idea what happens when HTML tag or its attributes are malformed.
- Weak against the dark incantations of the Sorceror Supreme and Dormammu, the only beings who can fully comprehend the true mystical powers of Regex.
```

Regex matching seems lazy, which is good for our purpose.

Tested on the following text:
```
<div id="articles" class="articles">
  <div data-id="1">
    <h3>New Marvel Movie coming out</h3>
    <a href="#"><img src='avengers.jpg'></a> <br> <br/><br />
	 <span   class='spaced caption'  >Hulk is < Iron Man, but Hulk > Thor</span>
	 <!--This comment is unaffected-->
  </div>
  <input disabled><input type="">
</div>
```

Note: all this said, however, if we're not _too) worried about getting a perfect HTML-tag-stripping function for the Blog excerpts on the home page, (e.g., we can always argue that hey, we have control of what goes on there anyways) then this PR is actually good to go as-is.

eatyourgreens:
If we trust that feed 🤔 then React has [`dangerouslySetHTML`](https://reactjs.org/docs/dom-elements.html#dangerouslysetinnerhtml) which will parse the string content as HTML, rather than adding it to the DOM as a text node.

astopy:
@eatyourgreens I did think we might be able to do something like that, but I'm not sure we'd want to display any images, video, etc. that happens to be in the first paragraph of the blog posts.

@shaunanoordin I think if we're going to try to make it that sophisticated we might want to use this npm package: https://www.npmjs.com/package/striptags Want me to do that instead?

astopy:
I've added striptags. The latest version is incompatible with uglifyjs:

```
ERROR in vendor-b31245ac3abf1cc05f22.js from UglifyJs
Unexpected token: name (context) [./~/striptags/src/striptags.js:26,0][vendor-b31245ac3abf1cc05f22.js:187009,12]
```

So I went with version 2.2 instead.

https://remove-blog-tags.pfe-preview.zooniverse.org

PR #4677, opened 2018-06-06
Language picker

eatyourgreens:
Staging branch URL: https://language-picker.pfe-preview.zooniverse.org

https://language-picker.pfe-preview.zooniverse.org/projects/y-dot-liefting/snapshot-hoge-veluwe?env=production is available in English and Dutch

- [x] Add a simple language picker to projects.
- [x] Style the language picker.
- [x] Reload translations when language changes.
- [x] Let project owners add and remove languages from their projects.
- [x] Add tests.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/locales/nl.js
+++ b/app/locales/nl.js
@@ -54,6 +54,7 @@ export default {
     miniCourseButton: 'Herstart de minitour'
   },
   project: {
+    language: 'Taal',
     loading: 'Project wordt geladen',
     disclaimer: 'Dit project is gemaakt met de Zooniverse projectbouwer maar is not niet een officieel Zooniverseproject. Vragen en problemen met betrekking tot dit project die gestuurd worden aan het Zooniverseteam krijgen mogelijk geen reactie.',
     about: {
--- a/app/classifier/index.cjsx
+++ b/app/classifier/index.cjsx
@@ -75,6 +75,12 @@ ClassifierWrapper = createReactClass
       @setState {minicourse}
       this.props.actions.translations.load('minicourse', minicourse.id, this.props.translations.locale) if minicourse?
 
+  componentDidUpdate: (prevProps) ->
+    { tutorial, minicourse } = @state
+    if prevProps.translations.locale isnt this.props.translations.locale
+      this.props.actions.translations.load('tutorial', tutorial.id, this.props.translations.locale) if tutorial?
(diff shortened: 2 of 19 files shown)


beckyrother:
Design feedback:
- [ ] use background color #005d69 (dark teal);
- [ ] color: #fff;
- [ ] padding: .5em 1em;
- [ ] text-align: right
- [ ] margin-left between "Language" and select: 1em;

a_project_ _zooniverse

Overall I worry about the height this adds to the navigation. 
Will this only appear on projects with language options, or will it always be visible?
Another option might be to include it as a drop-down option in the nav bar, and on mobile create a nested dropdown.

eatyourgreens:
>Overall I worry about the height this adds to the navigation.
Would it be ok to move it down into the header section? The layout here is only to quickly show how it should work.

>Will this only appear on projects with language options, or will it always be visible?
It doesn't make sense to show it, to me, if there's only one language available for your project.

eatyourgreens:
I've updated the mock data to include almost all the languages that we had on the old Galaxy Zoo and Penguin Watch projects, so the language menu might be more realistic now.

beckyrother:
> Would it be ok to move it down into the header section?
Yes, definitely. 
And I agree, only show it when there are options. 

Thanks!

eatyourgreens:
I kept those separate because the layout might be different in the narrow and wide contexts, but I had the same thought about duplicating code.

app/locales/nl.js
```diff
@@ -54,6 +54,7 @@ export default {
     miniCourseButton: 'Herstart de minitour'
   },
   project: {
+    language: 'Language',
```
marten: "Taal"

eatyourgreens:
@srallen can you see why the single space between 'Language' and the select menu isn't rendering? I am baffled. https://github.com/zooniverse/Panoptes-Front-End/pull/4677/files#diff-f6bac5cf2f79d0acc13e4ab12016c1b1R26

app/locales/nl.js
eatyourgreens: 👍

eatyourgreens:
@srallen I ended up using `word-spacing: 2em` to set the spacing. Should be okay, I think.

PR #4680, opened 2018-06-07
Remove react swipe. Fix tutorial rendering.

### LINKED ISSUE #4659: Tutorial width doesn't match content
## Expected behavior
Tutorial widths should automatically adjust to wrap content

## Current behavior
Tutorials always render at 400px wide, even when the content is narrower. Hidden tutorial slides are partially revealed to the side of the tutorial popup.
The Galaxy Zoo tutorial, showing other sides just visible on the right.

## Steps to replicate
Open the tutorial on a project with no images on the first tutorial slide, such as Galaxy Zoo.

Further information available in #4652.

srallen:
Staging branch URL: https://remove-react-swipe.pfe-preview.zooniverse.org/

Fixes #4659. This refactors the Tutorial component into ES6, removes react-swipe and the StepThrough component. I realized that once you remove react-swipe from the StepThrough component, all that is left are the controls, so I moved that into the Tutorial itself. The Tutorial was the only component that used StepThrough.

Projects to test with:
https://remove-react-swipe.pfe-preview.zooniverse.org/projects/zooniverse/scotus-notes-behind-the-scenes-at-supreme-court-conference/classify?env=production
https://remove-react-swipe.pfe-preview.zooniverse.org/projects/zookeeper/galaxy-zoo/classify?env=production

Note: A side effect of removing react-swipe is that the touch swipe controls are now gone for the tutorial. Since we want to do further refactoring of the tutorial, including decoupling the data requests from the view component, a new approach to touch swipe could be reimplemented later. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
I think you can uninstall `react-swipe` too.

eatyourgreens:
Layout is broken if the slide content overflows the height of the tutorial.

Screenshot showing tutorial text overflowing the tutorial box and rendering against a transparent background.

eatyourgreens:
There's a jump in height between tutorial slides, which causes the top of the slide to jump up and down. This leads to a jarring effect where you're trying to follow the text from one slide to another and the top line jumps up and down. From a design point-of-view, I think it would be better if the top of the tutorial stayed in the same place. Maybe we need to calculate the slide heights in advance, and use the largest?

srallen:
Hm, seems like react-swipe was doing a bit more for styles than managing the width and which step the tutorial was on.

srallen:
I've fixed the keyboard input issue and the overflow issue. Still need to resolve the height variance between the slides and focus.

srallen:
I'm not sure why calling focus on the node after using `ReactDOM.findDOMNode` isn't working in the component. Works fine in the console using `document.querySelector`, so something may not be working as expected.

eatyourgreens:
Divs don’t interact with the keyboard. You can only focus elements like links, buttons and form inputs.

eatyourgreens:
The radio buttons for the step controls have autofocus already, so they should be handling the keyboard controls without requiring any additional code.

srallen:
@eatyourgreens Oh I've probably knew that at some point, but one of those things that I've forgotten. I was trying to get the overflow scroll to work by focusing the div that is overflowing. Since divs can't get focus, is there another way to do this?

eatyourgreens:
I’m not sure. Now that I think about it, the radio buttons will hijack those keys too. So up and down will change slide. 

You could add tabindex -1 to the container, which will allow it to be focused from scripts. The downside of that approach, since it’s a div, is that you have to write all the keyboard interactions yourself. However, it looks like all that code is already there. 

If you do focus the container, you might have to delete the autofocus from the radio buttons so that it doesn’t interfere.

app/classifier/tutorial.jsx
```diff
@@ -0,0 +1,301 @@
...
+  }
+
+  handlePositionAndFocus() {
+    this.handleFocus();
+    this.handleScroll();
+  }
+
+  handleFocus() {
+    if (this.div) {
+      const reactNode = ReactDOM.findDOMNode(this.div);
```
eatyourgreens: React refs return the DOM node, so you don’t need `findDOMNode` here.

app/classifier/tutorial.jsx
```diff
@@ -0,0 +1,301 @@
...
+      reactNode.focus(); // Why isnt this working?
+    }
+  }
+
+  handleScroll() {
+    const reactNode = ReactDOM.findDOMNode(this.div);
```
eatyourgreens: Ditto re. refs and DOM nodes.

eatyourgreens:
I think this is good to go now that the overflow has been fixed. I couldn't reproduce the problem where the cursor keys stopped working while navigating the tutorial slides. Scrolling long slides is hard from the keyboard with the existing tutorial too, so I don't think that should hold this up.

srallen:
I've cleaned up the refs. The last couple of remaining minor issues I'll tackle in the next refactor.

PR #4670, opened 2018-06-01
Reset classifier annotation state when new classifications load

eatyourgreens:
Sync classifier annotation state when a classification loads.
Hide summaries if classification summaries are disabled.
Add tests for annotation state when the classifier mounts, when classifications are received, when subjects load and when a classification completes.

Staging branch URL: https://classifier-state.pfe-preview.zooniverse.org

Fixes #4668.
Fixes #4634.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4687, opened 2018-06-08
Only call getWorkflows if selectedWorkflowID is defined and truthy

srallen:
Staging branch URL: https://fix-infinite-loop-bug.pfe-preview.zooniverse.org/

Fixes regression introduced in #4672. #4672 attempted to fix a workflow selection infinite loop by checking first if a project had a linked workflows array before calling `getSelectedWorkflow`. This stopped the workflow selection loop in a scenario when a project has no workflows, but it has a few issue:

- Empty arrays in javascript evaluate true, so you have to check that the array exists and the length of it (oops). So the previous conditional always evaluated true.

- Even if you check the length for `project.links.workflows`, it introduced a new infinite loop with projects that have a linked workflow that is not active like http://zooniverse.org/projects/rolando-dot-tokez/canna-if-i-juanna

- What the actual issue is that sometimes we're calling `getWorkflow` when `selectedWorkflowID` is undefined or falsey. This can be undefined or falsey in a couple of scenarios: 
  - A project that has no linked workflows
  - A project with an inactive workflow linked, but there is no workflow query param, and/or there's no logged in user (like an admin, owner, etc), so the `selectedWorkflowID` from 'selectRandomWorkflow` returns an empty string

This fixes both scenarios by removing the conditional check from the then block after calling `clearInactiveWorkflow` and instead only calls `getWorkflow` if `selectedWorkflowID` is defined and truthy.

I've renamed `selectRandomWorkflow` to `selectRandomActiveWorkflow` to make it clear what it is doing. There's a follow up to this, though: should this method be changed to allow admins, owners, collaborators, and testers to select a random workflow out of the inactive workflow links? That's out of scope of this PR, so I didn't add that.

Projects to test with:
A project with only inactive workflows linked: https://fix-infinite-loop-bug.pfe-preview.zooniverse.org/projects/rolando-dot-tokez/canna-if-i-juanna?env=production
A project with no workflows linked: https://fix-infinite-loop-bug.pfe-preview.zooniverse.org/projects/srallen086/untitled-project-5-10-2018-11-53-20-am

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -104,7 +104,7 @@ class WorkflowSelection extends React.Component {
       awaitWorkflow = Promise.resolve(null);
     }
 
-    awaitWorkflow
+    return awaitWorkflow
     .then((workflow) => {
       if (workflow) {
         this.setState({ loadingSelectedWorkflow: false, workflow });
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -69,10 +69,11 @@ class WorkflowSelection extends React.Component {
     } else if (project.configuration && project.configuration.default_workflow) {
       selectedWorkflowID = project.configuration.default_workflow;
     } else {
-      selectedWorkflowID = this.selectRandomWorkflow(project);
+      selectedWorkflowID = this.selectRandomActiveWorkflow(project);
     }
 
-    this.getWorkflow(selectedWorkflowID, activeFilter);
+    if (selectedWorkflowID) return this.getWorkflow(selectedWorkflowID, activeFilter);
+    if (process.env.BABEL_ENV !== 'test') console.warn('Cannot select a workflow.')
   }
 
   getWorkflow(selectedWorkflowID, activeFilter = true) {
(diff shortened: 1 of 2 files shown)


app/pages/project/workflow-selection.jsx
```diff
@@ -69,10 +69,11 @@ class WorkflowSelection extends React.Component {
     } else if (project.configuration && project.configuration.default_workflow) {
       selectedWorkflowID = project.configuration.default_workflow;
     } else {
-      selectedWorkflowID = this.selectRandomWorkflow(project);
+      selectedWorkflowID = this.selectRandomActiveWorkflow(project);
     }
 
-    this.getWorkflow(selectedWorkflowID, activeFilter);
+    if (selectedWorkflowID) return this.getWorkflow(selectedWorkflowID, activeFilter);
```
eatyourgreens: `this.getWorkflow` doesn't return a value, so I think this will return undefined. Is it intended to return a promise?

srallen: Not exactly. I just wanted to make sure the following line's console message only happened if `selectedWorkflowID` is falsey. But good point, I'll make sure `getWorkflow` returns something.

eatyourgreens: in #4670 I found that updating a method to return a promise made it easier to test, because I could look at how the promise resolved. Probably not relevant here because `getWorkflow` is stubbed in the tests, but it might come in useful.

PR #4703, opened 2018-06-18
Fix a heading in the readme

eatyourgreens:
Fixes the heading on the 'Developing with Docker' section.


eatyourgreens:
Weird that the markdown parser is fussy about unicode spaces too.

PR #4732, opened 2018-07-05
ITA: translation fix

SebastianoPistore:
Staging branch URL:

Describe your changes.  The italian translation is updated, some phrases are now translated.


eatyourgreens:
Thanks for taking the time to do this, but it looks like these changes have already been made in another PR. #4651

Issue #4666, opened 2018-05-31
Drawing subtask no longer receives focus when it opens


eatyourgreens:
## Expected behavior
Drawing subtasks should receive focus when they open, and retain focus until the popup closes.

## Current behavior
Tab navigation is broken when a drawing subtask opens. I can't continue classifying from the keyboard until I use a mouse to focus the subtask popup. After the subtask dialog opens, focus remains with the drawing tool that opened it.

## Steps to replicate
Open up the rating subtask for the catface drawing task in I Fancy Cats and try to choose a rating using only the keyboard.

I think this might have been broken by #4610 or by the upgrade to React 16.

srallen:
I can take a look at this now that I finally have tests working again. I had been planning on writing tests and adding the light and dark themes to this.

eatyourgreens:
I'm not sure, but it might have been broken by React 16. I think autofocus is working for the task component, but the drawing tool's focus method is being run again and stealing focus from the dialog.

The solution is probably to remove tabindex/focusable from the drawing tool while the details form is open.

eatyourgreens:
Tabbing through the drawing subtask modal, when it has focus, is also broken. If I tab into a set of radio buttons, the first button is focussed regardless of which button is actually selected.

PR #4743, opened 2018-07-16
Convert text task to ES6, refactor

### LINKED ISSUE #4095: the text subTask modal changes size when re-opened
## Expected behavior
The text input in a subTask modal should stay consistently sized each time the subTask is opened.

## Current behavior
The text input in a subTask modal is the correct size when initially opened. However, if a volunteer re-selects and opens a subTask modal, the size of the text input is greatly increased.

## Steps to replicate
On a project with a drawing task with a transcription subTask (e.g., decodingthecivilwar.org):
1. Draw a line
2. Provide a transcription for that line
3. Press "Ok" on the transcription modal
4. Re-open the modal by selecting the drawn line

## Additional information

- **Operating system:** MacOS
- **Browser:** Chrome

### LINKED ISSUE #4632: Shift+tab into text box should send carat to end of transcription
## Expected behavior
Shift+tab into text box should send carat to end of transcription (or where transcriber left off)
## Current behavior
Shift+tab into text box sends carat to the beginning of the transcription box
## Steps to replicate
- Go to a PFE transcription project: https://www.zooniverse.org/projects/tkotwim/the-american-soldier/classify
- Enter transcription, then use any of the text modifiers
- Use Shift+Tab to return to text box
- Carat will default to the start of the text, rather than the end
## Additional information
Original Talk post flagging this up: https://www.zooniverse.org/projects/tkotwim/the-american-soldier/talk/1671/633636?page=1&scrollToLastComment=true
- **Operating system:** Mac OSX High Sierra 10.13.3
- **Browser:** Chrome

### LINKED ISSUE #4751: Truncation in text tasks
## Expected behavior
- User enters text in text area and presses Done
- The entered text is shown in the classification summary

## Current behavior
- User enters text and _very quickly_ presses done: text missing or truncated in summary (and database)
- User enters text and, _after a second or longer_, presses done: full text is shown in summary

## Steps to replicate
- Enter text and without delay press done
- The timespan between entering and pressing done is the deciding factor. If entering each character slowly and pressing done quickly leads to truncation. Entering the whole text quickly and pressing done immediately after leads to empty result.
- Slower Internet connections might make the problem more pronounced

## Additional information
See screen capture:
[Screen capture](https://www.dropbox.com/s/2ikbvxoq3qqcla3/TruncationTest.mp4?dl=0)

We could also reproduce this reliably using the workflow classification export (tested on various systems in different text entry speeds).
Had a look at "Panoptes-Front-End-master\app\classifier\tasks\text\...", but am not familiar enough with React to track down this problem. Perhaps something to do with the onChange of the text area and synchronization with the Done button.

- **Operating system:** Tested on Windows and Android
- **Browser:** Tested with Chrome and Firefox

mcbouslog:
Staging branch URL: https://text-task-es6-refactor.pfe-preview.zooniverse.org/
- I used the NfN Combo 1st, Text Tasks and TestDCW workflows in [Focus Testing Project here](https://text-task-es6-refactor.pfe-preview.zooniverse.org/lab/1608/workflows) 

Fixes #4632 
Fixes #4751 
Towards #4095.

Describe your changes.
- add tests for text task index
- convert text task to ES6
- refactor lifecycle, handleChange, handleResize

Notes
- ~~intended to clean-up lifecycle and state vs props issues more, but due to tags and resizing couldn't figure out how to simplify more~~ thanks to help from Jim this has been addressed! [relevant React blog post](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key).
- when it's a subtask, it can't properly access the textarea properties onMount, for example, if you log `this.textInput.current.scrollHeight` after mount, it'll be 0, even if has value, but same log in regular text task will yield a scrollHeight - related to [focus issue for subtasks](https://github.com/zooniverse/Panoptes-Front-End/issues/4666), punting on for now

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/classifier/tasks/text/index.jsx
```diff
@@ -0,0 +1,189 @@
...
+    } else {
+      const textInBetween = textAreaValue.substring(selectionStart, selectionEnd);
+      textAfter = textAreaValue.substring(selectionEnd, textAreaValue.length);
+      value = textBefore + startTag + textInBetween + endTag + textAfter;
+    }
+    this.setState({ value });
+    this.debouncedUpdateAnnotation();
+  }
+
+  setValue() {
```
mcbouslog: should prob be renamed, maybe something like `setInitialValues` or open to suggestion

app/classifier/tasks/text/index.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+  componentDidMount() {
+    this.setValue();
+  }
+
+  componentDidUpdate(prevProps, prevState) {
+    if (prevProps.task && this.props.task && (prevProps.task !== this.props.task)) {
+      const value = prevState.value;
+      const newAnnotation = Object.assign(prevProps.annotation, { value });
+      prevProps.onChange(newAnnotation);
```
mcbouslog: Lines 37-39 and on line 119 the `onBlur={this.updateAnnotation}` are from when I was trying to avoid any updateAnnotation in `handleChange`, but `handleChange` needs some updateAnnotation so the container knows the task has been completed, otherwise I ran into the scenario with a required text task where I would have typed an answer, but not triggered the `onBlur` and it would think I had not typed anything and would still have "Next" or "Done" disabled when it shouldn't have been.

With `onBlur={this.updateAnnotation}` lines 37-39 may not be needed?

I've kept `onBlur={this.updateAnnotation}` for the scenario of back-to-back two text task combos. If the first text task is completed, then the second, then next task, neither text task is unmounted and with the debounced updateAnnotation I'm not sure if there's a chance the last few keypresses could be missed, but the `onBlur` would make sure those would be saved?

mcbouslog:
With the rows in state, it'll be easier to refactor the editor and task so someone building a text task can set a default row height other than 1.

eatyourgreens:
I haven't looked at the code, but the TestDCW project looks good. The line-height code also works (in Chrome) regardless of the text-size that I've got set in my browser (tested up to a zoom of around 200%.)

👍

app/classifier/tasks/text/index.jsx
```diff
@@ -0,0 +1,189 @@
...
+
+  setTagSelection(e) {
+    const textTag = e.target.value;
+    const startTag = `[${textTag}]`;
+    const endTag = `[/${textTag}]`;
+    const textArea = this.textInput.current;
+    const textAreaValue = textArea.value;
+    const selectionStart = textArea.selectionStart;
+    const selectionEnd = textArea.selectionEnd;
+    const textBefore = textAreaValue.substring(0, selectionStart);
```
eatyourgreens: I think you can test this by mocking `this.textInput`. Something like
```
wrapper.instance().textInput.current = {
  value: 'blah',
  selectionStart: 1,
  selectionEnd: 3
}
```
then looking at `state.value` after this method runs runs to see if it updates as expected.

app/classifier/tasks/text/index.jsx
eatyourgreens: If you're setting state from props, React 16 now has [`TextTask.getDerivedStateFromProps()`](https://reactjs.org/docs/react-component.html#static-getderivedstatefromprops), which might be what you're looking for.

eatyourgreens:
Is #4632 a small enough fix that it could be added to this PR too?

app/classifier/tasks/text/index.jsx
mcbouslog: I tried using `getDerivedStateFromProps`, and could set the initial state value as the provided props annotation value, but had issues when clicking a tag. I'll take another look today, I thought I tried it, but maybe if `setTagSelection` calls `updateAnnotation` `getDerivedStateFromProps` can work as intended.

mcbouslog:
^ ooh (#4632) the carat to the end would be nice, from a quick Google search looks possible, I'll give it a try this afternoon.

app/classifier/tasks/text/index.jsx
eatyourgreens: `getDerivedStateFromProps` runs after every update. Looking the alternatives mentioned in the docs, the case here might be this one:
>If you want to “reset” some state when a prop changes, consider either making a component fully controlled or fully uncontrolled with a key instead.

eatyourgreens: If you look at [the uncontrolled example](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#recommendation-fully-uncontrolled-component-with-a-key), it turns out you could use a key to force React to remount the task component when a new task loads. Maybe use the task key from the workflow?

I think that might simplify the case where the task changes but the component is already mounted.

eatyourgreens:
I think my suggested fix in the comments to #4632 might work, but it was based on a really quick reading of the spec. If it turns out to be a lot of work, then leave it for another PR.

app/classifier/tasks/text/index.jsx
mcbouslog: Initial attempt without luck, replaced constructor state with state declaration like in the uncontrolled example, added key in textarea of this.props.annotation.task, but with back to back text tasks the annotation would carryover to the second task. Will look again with fresh eyes on Wednesday.

mcbouslog:
I think I have the intended post tag selection functionality working, but please confirm.

eatyourgreens:
I've added a key to tasks in #4745, which ought to make setting state from props easier.

eatyourgreens:
👍 Excellent work with the text selection! I think @snblickhan (and the volunteers) noted that it was fiddly to continue typing immediately after adding a tag, and this fixes that for me in Chrome. I haven't checked Firefox, Safari or IE.

app/classifier/tasks/text/index.jsx
eatyourgreens: >added key in textarea of this.props.annotation.task
I think the key needs to be added to the task component itself. I've tried that in #4745.

mcbouslog:
The lifecycles and state look much better to me now, thanks @eatyourgreens for your help.

I still have one issue (and fixing the tests) though, which is when a drawing task text subtask is mounted, it doesn't get focus or resized. I'm not sure why this is only happening for the text task as a drawing subtask.

eatyourgreens:
Autofocus is broken in general for drawing subtasks. Radio buttons for single answer tasks have the same problem. See #4666.

app/classifier/tasks/text/index.jsx
mcbouslog: this worked, added [this test](https://github.com/zooniverse/Panoptes-Front-End/pull/4743/files#diff-6b77c8c82a092d3f72f871b21598b89eR77), thanks!

mcbouslog:
I think this is ready for final review. I'm working on the Required Manual Testing checklist at the moment.

Unfortunately punting on the row sizing issue when a text subtask is re-opened.

app/classifier/tasks/text/index.jsx
```diff
@@ -0,0 +1,184 @@
...
+    const oldRows = this.textInput.current.rows;
+    const newRows = Math.floor(this.textInput.current.scrollHeight / LINEHEIGHT);
+
+    if (newRows && (newRows !== oldRows)) {
+      this.setState({ rows: newRows });
+    }
+  }
+
+  updateAnnotation() {
+    if (this.textInput.current) {
```
eatyourgreens: Minor point but if `this.textInput.current` is undefined, then something has gone wrong here. Any pending calls to `updateAnnotation` should be canceled when the component unmounts, and this method should probably error if we're trying to sync annotation values after the text input has been removed from the DOM.

PR #4744, opened 2018-07-16
Add simple interventions to the classifier

eatyourgreens:
Staging branch URL: https://interventions.pfe-preview.zooniverse.org

First steps towards adding interventions/notifications to PFE projects.

- Add a redux store, and a few actions, for the Sugar client.
- Moves Sugar project channel subscriptions, and tests, over to Redux.
- Adds a simple notifications listener, which simply stores an array of received notifications.
- Adds a very simple Intervention component, which simply renders the latest Sugar message during the classification summary step. If summaries are hidden, it interrupts the 'Done' and 'Done & Talk' buttons in order to show a message.

TODO, either in this PR or later.

- [x] Display notifications to volunteers somewhere in the classifier.
- [x] Remove notifications from the store once they've been dismissed.
- [ ] Offer an option to unsubscribe from receiving notifications.
- [ ] Subscribe volunteers to channels by UserID, as well as by ProjectID.
- [ ] Style interventions and position them so that they won't be missed.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -4,11 +4,16 @@ ReactDOM = require 'react-dom'
 useScroll = require 'react-router-scroll/lib/useScroll'
 routes = require './router'
 style = require '../css/main.styl'
+{ sugarClient } = require 'panoptes-client/lib/sugar'
 
 # Redux
 `import { Provider } from 'react-redux';`
 `import configureStore from './redux/store';`
+`import { notify, injectSubjects } from './redux/ducks/interventions';`
 store = configureStore()
+
+sugarClient.on('notification', (message) => store.dispatch(notify(message)));
+sugarClient.on('subject-queue', (message) => store.dispatch(injectSubjects(message)));
 
 # Redirect any old `/#/foo`-style URLs to `/foo`.
 if location?.hash.charAt(1) is '/'
--- a/app/redux/ducks/interventions.js
+++ b/app/redux/ducks/interventions.js
@@ -1,0 +1,47 @@
+import { sugarClient } from 'panoptes-client/lib/sugar';
+
+const ADD_NOTIFICATION = 'pfe/interventions/ADD_NOTIFICATION';
+const SUBSCRIBE = 'pfe/interventions/SUBSCRIBE';
+const UNSUBSCRIBE = 'pfe/interventions/UNSUBSCRIBE';
+
+const initialState = {
+  notifications: []
+};
+
+export default function reducer(state = initialState, action = {}) {
+  const { notifications } = state;
+  switch (action.type) {
+    case ADD_NOTIFICATION:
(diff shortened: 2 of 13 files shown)


chrislintott:
Note on unsubscribing from notifications - this would have to be very modular, so that people can unsubscribe from specific notifications associated with particular campaigns. Ideally this would be done on the notification ('Never show me notifications like this')

eatyourgreens:
Previously, subscriptions were managed by experiments.zooniverse.org. Will there be an external service managing these campaigns, or do we manage subscriptions in Panoptes user preferences?
https://github.com/zooniverse/Panoptes-Front-End/blob/b1b0c36be22f66b61211a463a8a97093bddbaf24/app/lib/experiments-client.coffee#L229-L250

eatyourgreens:
If you'd like to try this out, https://interventions.pfe-preview.zooniverse.org/ now loads with a single test message, which should show after your first classification on any project.

eatyourgreens:
I broke the feedback tests. 😞

app/redux/ducks/interventions.js
```diff
@@ -0,0 +1,55 @@
...
+export function dismiss() {
+  return { type: DISMISS_NOTIFICATION };
+}
+
+export function injectSubjects() {
+  // Get new subjects and add them to the subject queue
+}
+
+sugarClient.on('notification', notify);
+sugarClient.on('subject-queue', injectSubjects);
```
eatyourgreens: Action creators shouldn’t be invoked directly like this. The Sugar client should dispatch an action when it handles an event.

srallen:
I might be able to take a look at this later in the week.

app/main.cjsx
```diff
@@ -4,12 +4,17 @@ ReactDOM = require 'react-dom'
...
 style = require '../css/main.styl'
+{ sugarClient } = require 'panoptes-client/lib/sugar'
 
 # Redux
 `import { Provider } from 'react-redux';`
 `import configureStore from './redux/store';`
+`import { notify, injectSubjects } from './redux/ducks/interventions';`
 store = configureStore()
 
+sugarClient.on('notification', message => store.dispatch(notify(message)));
```
mcbouslog: Getting ReferenceError on local and staged branch `message is not defined`?

eatyourgreens: That'll be because this should be coffeescript, not ES6.

eatyourgreens: I've added the missing brackets now.

eatyourgreens:
`fakeMessage` was just for debugging. I’ll remove that commit before this gets merged.

mcbouslog:
Sounds good. I'll assign to you then. It's ready for merge pending the removal of `fakeMessage`.

PR #4749, opened 2018-07-21
Convert account page to ES6

jarihaa:
Staging branch URL:

Fixes # .

* Convert account page to es6
* Update router
* Move change password form to its own file
* Add translations
* Add tests

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/settings/account.jsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import counterpart from 'counterpart';
+import AutoSave from '../../components/auto-save';
+import handleInputChange from '../../lib/handle-input-change';
+import ChangePasswordForm from './change-password-form';
+
+export default class AccountInformationPage extends React.Component {
```
srallen: Since this doesn't have state or class methods, this could just be [a functional component](https://reactjs.org/docs/components-and-props.html#functional-and-class-components). 

Could you match the file name to the component name, so `AccountInformationPage.jsx`? I know we have a fix of conventions with file names going on, but having the file name match the class name/function name is what we're going with for new components.

app/pages/settings/change-password-form.jsx
```diff
@@ -0,0 +1,147 @@
+import React from 'react';
+import auth from 'panoptes-client/lib/auth';
+import counterpart from 'counterpart';
+
+const MIN_PASSWORD_LENGTH = 8;
+
+export default class ChangePasswordForm extends React.Component {
+
+  constructor(props) {
+    super(props);
```
srallen: `props` only needs to be passed into the `super()` call if `this.props` is being used in the constructor, so this can just be `super()`

app/pages/settings/change-password-form.jsx
```diff
@@ -0,0 +1,147 @@
...
+    return status;
+  }
+
+  render() {
+    return (
+      <form ref={this.changePasswordForm} method="POST" onSubmit={this.handleSubmit}>
+        <p>
+          <strong>{counterpart('userSettings.account.changePassword.heading')}</strong>
+        </p>
+        <table className="standard-table">
```
srallen: This is a comment for a later PR since the original component did this. It's not great for accessibility to use a table for layout purposes rather than tabular data. You can use CSS to make the markup render like a table with `display: table;` instead and use good semantic HTML for a form. MDN has a good list of what elements should be used with a form: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/form. The `label` tags should be wrapping the `input`s.

PR #4754, opened 2018-07-24
Fix text task row issues in Windows/Firefox

mcbouslog:
Staging branch URL: https://fix-text-rows.pfe-preview.zooniverse.org/

Fixes https://www.zooniverse.org/talk/17/692672

Describe your changes.
- add `.length` to comparison of `value` strings before reseting rows to 1
- add Math.min and Math.max so rows never below 1 or above 10 (per `MAX_ROWS` constant)
- add `overflow: hidden` to fix issue on Windows 7, Firefox 61 where scrollHeight included scrollbar, creating row resizing loop and inconsistency with other browsers

Used BrowswerStack to test, see Passbolt for login info.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/text/index.jsx
+++ b/app/classifier/tasks/text/index.jsx
@@ -75,9 +75,9 @@ export default class TextTask extends React.Component {
   handleChange() {
     const value = this.textInput.current.value;
     if (value.length < this.state.value.length) {
-      this.setState({ rows: 1, value }, () => { this.handleResize(); });
+      this.setState({ rows: 1, value }, this.handleResize);
     } else {
-      this.setState({ value }, () => { this.handleResize(); });
+      this.setState({ value }, this.handleResize);
     }
 
     this.debouncedUpdateAnnotation(value);
@@ -85,7 +85,8 @@ export default class TextTask extends React.Component {
 
   handleResize() {
     const oldRows = this.textInput.current.rows;
-    const newRows = Math.max(Math.min(Math.floor(this.textInput.current.scrollHeight / LINEHEIGHT), MAX_ROWS), 1);
+    let newRows = Math.min(Math.floor(this.textInput.current.scrollHeight / LINEHEIGHT), MAX_ROWS);
+    newRows = Math.max(newRows, 1);
 
     if (newRows && (newRows !== oldRows)) {
       this.setState({ rows: newRows });
--- a/app/classifier/tasks/text/index.jsx
+++ b/app/classifier/tasks/text/index.jsx
@@ -7,6 +7,7 @@ import TextTaskEditor from './editor';
 import TextTaskSummary from './summary';
 
 const LINEHEIGHT = 22.5;
+const MAX_ROWS = 10;
 const NOOP = Function.prototype;
 
 export default class TextTask extends React.Component {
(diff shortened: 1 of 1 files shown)


app/classifier/tasks/text/index.jsx
```diff
@@ -84,7 +85,7 @@ export default class TextTask extends React.Component {
 
   handleResize() {
     const oldRows = this.textInput.current.rows;
-    const newRows = Math.floor(this.textInput.current.scrollHeight / LINEHEIGHT);
+    const newRows = Math.max(Math.min(Math.floor(this.textInput.current.scrollHeight / LINEHEIGHT), MAX_ROWS), 1);
```
eatyourgreens: I would split this across two lines for readibility.

app/classifier/tasks/text/index.jsx
```diff
@@ -73,7 +74,7 @@ export default class TextTask extends React.Component {
 
   handleChange() {
     const value = this.textInput.current.value;
-    if (value < this.state.value) {
+    if (value.length < this.state.value.length) {
       this.setState({ rows: 1, value }, () => { this.handleResize(); });
```
eatyourgreens: The callback can just be `this.handleResize` here. It doesn’t need to be wrapped in an anonymous function.

app/classifier/tasks/text/index.jsx
```diff
@@ -73,7 +74,7 @@ export default class TextTask extends React.Component {
...
     } else {
       this.setState({ value }, () => { this.handleResize(); });
```
eatyourgreens: This can also be simplified to `setState({ value }, this.handleResize)`.

app/classifier/tasks/text/index.jsx
```diff
@@ -110,7 +111,7 @@ export default class TextTask extends React.Component {
             onChange={this.handleChange}
             ref={this.textInput}
             rows={this.state.rows}
-            style={{ lineHeight: `${LINEHEIGHT}px` }}
+            style={{ lineHeight: `${LINEHEIGHT}px`, overflow: 'hidden' }}
```
eatyourgreens: If the input text is very long, will this prevent it from being scrolled to edit it?

mcbouslog: You can still use the cursor and arrow keys to scroll down, but there isn't a scroll bar. The following screen shot has several lines of text below what is visible:
screen shot 2018-07-25 at 2 24 22 am

I'm not sure if setting a max number of rows (could be something more obnoxious, like 20) and risk text cutoff without scroll bar (like above) or not setting any max and letting expand as intended (with this PR hopefully fixing noted related bug) preferable.

eatyourgreens: To be honest, a vertical scroll bar changes the width of an element, so I’m not really sure what was happening with `scrollHeight` here.

mcbouslog:
Updated with proper callbacks, improved `newRows` min/max.

The `overflow` edit is certainly hacky, I'm totally open to a diff approach, or I should at least add some comments in code?

eatyourgreens:
Let's merge this and see how it goes with the `overflow: hidden`. We can remove it quickly if it causes problems on any of the transcription projects.

PR #4781, opened 2018-08-08
Extract the classifier subject queue to the redux store

eatyourgreens:
Extract the subject queue code from the classifier and into redux actions for filling the queues, emptying them and getting the next subject to classify.
Removes the `currentClassifications` and `upcomingSubjects` global variables and stores those variables in the redux store instead.
Fixes a bug where the classifier state could preserve old annotations after a new classification had been created.
Fixes a small bug where the classifier annotations array could be accidentally updated at index -1.

Staging branch URL: https://subject-queue.pfe-preview.zooniverse.org/

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -21,10 +21,6 @@ import WorkflowAssignmentDialog from '../../components/workflow-assignment-dialo
 import ProjectThemeButton from './components/ProjectThemeButton';
 import { zooTheme } from '../../theme';
 
-// Map each project ID to a promise of its last randomly-selected workflow ID.
-// This is to maintain the same random workflow for each project when none is specified by the user.
-const currentWorkflowForProject = {};
-
 function onClassificationSaved(actualClassification) {
   Split.classificationCreated(actualClassification); // Metric log needs classification id
 }
@@ -41,13 +37,10 @@ let sessionDemoMode = false;
 export class ProjectClassifyPage extends React.Component {
   constructor(props) {
     super(props);
-    this.loadingSelectedWorkflow = false;
     this.project = null;
     this.workflow = null;
 
     this.state = {
-      classification: null,
-      subject: null,
       projectIsComplete: false,
       demoMode: sessionDemoMode,
       promptWorkflowAssignmentDialog: false,
@@ -58,7 +51,7 @@ export class ProjectClassifyPage extends React.Component {
 
   componentDidMount() {
     Split.classifierVisited();
-    if (this.props.workflow && !this.props.loadingSelectedWorkflow) {
+    if (this.props.workflow) {
       this.loadAppropriateClassification();
     }
 
@@ -66,7 +59,7 @@ export class ProjectClassifyPage extends React.Component {
   }
 
   componentWillReceiveProps(nextProps, nextContext) {
-    if (nextProps.loadingSelectedWorkflow === false && nextProps.user !== null) {
+    if (nextProps.user !== null) {
       this.shouldWorkflowAssignmentPrompt(nextProps, nextContext);
     }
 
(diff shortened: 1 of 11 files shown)


eatyourgreens:
The code is kind of ugly at the moment, but seems to work. It's basically the old classifier code copied into Redux actions.

eatyourgreens:
We'd like to be able to add subjects to the queue from Sugar notifications. The aim of the changes here is that a Sugar notification should be able to fetch specific subjects from Panoptes, then  dispatch `ADD_SUBJECTS` with those subjects in the action payload.

app/pages/project/classify.jsx
```diff
@@ -51,8 +41,8 @@ function isPresent(val) {
 
 const classificationQueue = new ClassificationQueue(window.localStorage, apiClient, onClassificationSaved);
 
-auth.listen('change', emptySubjectQueue);
-apiClient.type('subject_sets').listen('add-or-remove', emptySubjectQueue);
+auth.listen('change', classifierActions.emptySubjectQueue);
+apiClient.type('subject_sets').listen('add-or-remove', classifierActions.emptySubjectQueue);
```
eatyourgreens: These handlers should dispatch actions, rather than invoking action creators directly.

eatyourgreens:
Latest commit has broken this project, which uses 'Persist annotations'. I'm not sure why.
https://subject-queue.pfe-preview.zooniverse.org/projects/srallen086/srallen-testing
New subjects load, but the new classification is overwritten with the annotations from the previous subject.

eatyourgreens:
It looks like ‘Hide Previous Marks’ might be breaking this, because it restores the old annotations after the classification is complete. It shouldn’t really be updating annotations at all.

eatyourgreens:
If you switch between projects while classifying, you can get a subject from one project to show up in the workflow of another. Screenshot of the focus testing project with a subject loaded from a different project.

eatyourgreens:
This should be ready to review now.

app/redux/ducks/classify.js
```diff
@@ -0,0 +1,169 @@
...
+        const forWorkflow = Object.assign({}, currentClassifications.forWorkflow);
+        forWorkflow[workflow.id] = classification;
+        return Object.assign({}, state, { currentClassifications: { forWorkflow }});
+      }
+      return state;
+    }
+    case RESET_CLASSIFICATION: {
+      const { workflow } = action.payload;
+      currentClassifications.forWorkflow[workflow.id] = null;
+      upcomingSubjects.forWorkflow[workflow.id].shift();
```
eatyourgreens: This runs after submitting a classification, but also when switching workflows. Is there a case where we'd want to destroy a classification, but keep the subject in the queue?

app/pages/project/classify.jsx
```diff
@@ -128,34 +82,45 @@ export class ProjectClassifyPage extends React.Component {
...
+  componentWillUpdate(nextProps) {
+    const nextWorkflowID = (isPresent(nextProps) && isPresent(nextProps.workflow)) ? nextProps.workflow : null;
+    this.context.geordi.remember({ workflowID: nextWorkflowID });
+  }
+
+  componentDidUpdate(prevProps) {
+    const { actions, currentClassifications, project, upcomingSubjects, workflow } = this.props;
+
+    if (project !== prevProps.project) {
+      this.loadAppropriateClassification();
```
eatyourgreens: Classifications are keyed by workflow ID so why does the classifier load a new classification when the project changes?

app/pages/project/classify.jsx
```diff
@@ -128,34 +82,45 @@ export class ProjectClassifyPage extends React.Component {
...
 
-  loadAppropriateClassification(props) {
-    // Create a classification if it doesn't exist for the chosen workflow, then resolve our state with it.
-    if (this.state.rejected && this.state.rejected.classification) {
-      this.setState({ rejected: null });
+    if (workflow &&
+      currentClassifications.forWorkflow[workflow.id] !== prevProps.currentClassifications.forWorkflow[workflow.id]
+    ) {
+      const classification = currentClassifications.forWorkflow[workflow.id];
+      const subject = upcomingSubjects.forWorkflow[workflow.id][0];
```
eatyourgreens: We’re trusting here that this is the right subject for the classification.

app/redux/ducks/classify.js
```diff
@@ -0,0 +1,171 @@
...
+
+export default function reducer(state = initialState, action = {}) {
+  const { currentClassifications, upcomingSubjects } = state;
+  switch (action.type) {
+    case ADD_SUBJECTS: {
+      const { subjects, workflow_id } = action.payload;
+      if (!upcomingSubjects.forWorkflow[workflow_id]) {
+        upcomingSubjects.forWorkflow[workflow_id] = [];
+      }
+      upcomingSubjects.forWorkflow[workflow_id].push(...subjects);
```
eatyourgreens: Dedupe the array after adding new subjects?

eatyourgreens:
Rebased to bring in the interventions work from #4744.

rogerhutchings:
This looks good to me. I've tested classifying, switching projects between classifications, and everything seems to work. Tests are in a separate PR.

We talked about:

- Deduping the subject queue for that theoretical instance where a queue could be requested twice; I'll create a separate issue.
- Splitting up the destroy classification and next subject actions, which may come in the next PR

app/redux/ducks/classify.js
srallen: This appears to be the line that is crashing when I first loaded Gravity Spy's classify page.

srallen:
Just to note, I was only able to get Gravity Spy to crash the first time I visited it. The workflows loaded fine after subsequent loads. I'm not sure why.

app/redux/ducks/classify.js
srallen: I'll look it over. I didn't notice yesterday that #4790 is set to merge into the branch for this PR, so if I can't replicate using its staged branch, then I'll merge and do a final review for this one here.

PR #4784, opened 2018-08-14
Snapshot WI split setup

srallen:
Staging branch URL: https://sw-split-setup.pfe-preview.zooniverse.org/

For an upcoming split for Snapshot Wisconsin at the start of their new season. This is similar to a previous split run by Gravity Spy (#3254 and #3286) where depending on the split cohort, users will be assigned a workflow.

For SW's experiment, I tried to generalize this process a bit more than the GS experiment. The client is still calling the update to the UPP to set the workflow id, but the ids aren't hardcoded in PFE, but instead stored in the Seven-Ten split variants. 

This experiment also implies a visibility split for the home page workflow buttons. Users who will not be promoted should not see the workflow buttons and see instead the 'Get Started' link to the classify page. Users who will promote should see the workflow buttons and not the 'Get Started' link.

The ProjectPageController is refactored to include the request for the splits in the `Promise.all` that is written to wait for all the project associated async requests. Then, `ready: true` is not set until we know whether or not there are splits, and if there is this specific kind of split running, also wait for the workflow assignment to happen. 

This still might not be where it should be for generalized cases, so this should be reevaluated again once we have another project that wants to run an experiment like this. I do not think Seven-Ten should be making any calls to Panoptes to update UPPs, however.

This can be tested with this project: https://sw-split-setup.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing

If you are in the group that will not promote, you will see the UPP `settings.workflow_id` be set to workflow id `2758`. If you are in the group that will promote, `settings.workflow_id` is set to workflow id `2333` which is also the project default. Logged out users should see project default workflow. 

Note: This project isn't configured to _actually_ promote you. You can mimic it by updating your own UPP if you wish, but you shouldn't need to test that part since it's out of scope of this PR.

If you wish to use them, the zootester1 account has been assigned to the split group that does promote. The account zootester5 is assigned to the group that doesn't promote. (these accounts are in passbolt now)

As an aside, I renamed `preferences` and the associated class methods to update UPP to `projectPreferences` in the ProjectPageController, because `preferences.preferences` is hard to read.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/home/project-home.jsx
+++ b/app/pages/project/home/project-home.jsx
@@ -32,12 +32,7 @@ const ProjectHomePage = (props) => {
     };
   }
 
-  // variant div: true shows the workflow buttons below
-  // and is for users in a split variant group that allows workflow promotion
-  // so here, we show the get started link if div: false
-  // since users who are not being promoted will work on only one workflow.
-  const showGetStartedLink = (!props.showWorkflowButtons && projectIsNotRedirected) ||
-    (projectIsNotRedirected && props.splits && props.splits['workflow.assignment'] && !props.splits['workflow.assignment'].variant.value.div)
+  const showGetStartedLink = (!props.showWorkflowButtons && projectIsNotRedirected) || props.splits['workflow.assignment']
 
   return (
     <div className="project-home-page">
@@ -61,17 +56,19 @@ const ProjectHomePage = (props) => {
         </div>
 
         <div className="project-home-page__call-to-action">
-          {props.project && !props.project.redirect &&
+          {projectIsNotRedirected &&
             <Link to={`/projects/${props.project.slug}/about`} className="project-home-page__button call-to-action__button call-to-action__button--learn-more">
               <Translate content="project.home.learnMore" />
             </Link>}
           {showGetStartedLink &&
-            <Link
-              to={`/projects/${props.project.slug}/classify`}
-              className="project-home-page__button call-to-action__button call-to-action__button--get-started"
-            >
-              <Translate content="project.home.getStarted" />
-            </Link>}
+            <VisibilitySplit splits={props.splits} splitKey='workflow.assignment' elementKey='link'>
+              <Link
+                to={`/projects/${props.project.slug}/classify`}
+                className="project-home-page__button call-to-action__button call-to-action__button--get-started"
+              >
+                <Translate content="project.home.getStarted" />
+              </Link>
+            </VisibilitySplit>}
           {props.project && props.project.redirect &&
             <a href={props.project.redirect} className="project-home-page__button">
               <strong><Translate content="project.home.visitLink" /></strong>
--- a/app/pages/project/home/project-home.jsx
+++ b/app/pages/project/home/project-home.jsx
@@ -4,6 +4,8 @@ import { Link } from 'react-router';
 import classnames from 'classnames';
 import { Markdown } from 'markdownz';
 import Translate from 'react-translate-component';
+import { VisibilitySplit } from 'seven-ten';
+
 import getSubjectLocations from '../../../lib/get-subject-locations';
 import Thumbnail from '../../../components/thumbnail';
 import FinishedBanner from '../finished-banner';
@@ -13,6 +15,7 @@ import TalkStatus from './talk-status';
 import ExternalLinksBlock from '../../../components/ExternalLinksBlock';
 
 const ProjectHomePage = (props) => {
+  const projectIsNotRedirected = props.project && !props.project.redirect;
   const avatarSrc = props.researcherAvatar || '/assets/simple-avatar.png';
 
   const descriptionClass = classnames(
@@ -28,6 +31,13 @@ const ProjectHomePage = (props) => {
       backgroundImage: `radial-gradient(rgba(0, 0, 0, 0.2), rgba(0, 0, 0, 0.8)), url('${props.background.src}')`
     };
   }
+
+  // variant div: true shows the workflow buttons below
+  // and is for users in a split variant group that allows workflow promotion
+  // so here, we show the get started link if div: false
+  // since users who are not being promoted will work on only one workflow.
+  const showGetStartedLink = (!props.showWorkflowButtons && projectIsNotRedirected) ||
+    (projectIsNotRedirected && props.splits && props.splits['workflow.assignment'] && !props.splits['workflow.assignment'].variant.value.div)
 
   return (
     <div className="project-home-page">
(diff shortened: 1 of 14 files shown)


srallen:
Note: https://sw-split-setup.pfe-preview.zooniverse.org/projects/srallen086/srallen-testing is broken, but working on master branch of staging. I'm looking into it.

app/pages/project/home/project-home.jsx
```diff
@@ -29,6 +32,13 @@ const ProjectHomePage = (props) => {
     };
   }
 
+  // variant div: true shows the workflow buttons below
+  // and is for users in a split variant group that allows workflow promotion
+  // so here, we show the get started link if div: false
+  // since users who are not being promoted will work on only one workflow.
+  const showGetStartedLink = (!props.showWorkflowButtons && projectIsNotRedirected) ||
+    (projectIsNotRedirected && props.splits && props.splits['workflow.assignment'] && !props.splits['workflow.assignment'].variant.value.div)
```
eatyourgreens: The logic is a bit confusing here. Why isn't `props.showWorkflowButtons` set to true if the split variant shows the workflow buttons?

app/pages/project/index.jsx
```diff
@@ -86,23 +88,17 @@ class ProjectPageController extends React.Component {
...
   }
 
-  setupSplits(props) {
-    if (!props) {
-      ({ props } = this);
-    }
-    const { user } = props;
-    const { owner, name } = props.params;
-
+  setupSplits(slug, user) {
```
eatyourgreens: This returns a promise to return a splits object. Can it be renamed to `getSplits` or awaitSplits`?

app/pages/project/index.jsx
```diff
@@ -86,23 +88,17 @@ class ProjectPageController extends React.Component {
...
     if (user) {
-      Split.load(`${owner}/${name}`)
+      return Split.load(slug)
       .then((splits) => {
```
eatyourgreens: The name `splits` implies an array, but it's actually a single object. Is it worth changing that to `split` throughout the code to make it clearer that it's a single thing?

app/pages/project/index.jsx
```diff
@@ -86,23 +88,17 @@ class ProjectPageController extends React.Component {
...
-        this.setState({ splits });
         if (!splits) {
-          return;
+          return null;
```
eatyourgreens: If `splits` defaults to `{}` instead of `null`, that will simplify any logic that checks for `splits && splits[property]`. `ProjectHomeWorkflowButtons` defaults `splits` to `{}` so I think that change should be safe.

app/pages/project/index.jsx
```diff
@@ -303,15 +311,43 @@ class ProjectPageController extends React.Component {
...
       if (this.props.user) {
-        preferences.save();
+        projectPreferences.save();
+      }
+    }
+  }
+
+  handleSplitWorkflowAssignment(projectPreferences, splits) {
+    if (splits && splits['workflow.assignment']) {
+      const projectSetWorkflow = (splits['workflow.assignment'].variant
```
eatyourgreens: Can the `workflow.assignment` key be renamed to `workflowAssignment`? No big deal if it can't, it would just make the key easier to work with in JavaScript.

eatyourgreens:
All projects currently throw `Cannot read property 'forEach' of undefined` in `ProjectPageController.fetchProjectData`. I think the broken line might be `Object.keys(splits).forEach` which will error when `splits` is null.

app/pages/project/index.jsx
eatyourgreens: Feel free to ignore this if `splits` is actually a hash of splits keyed by split type/name.

eatyourgreens:
My bad, if you add a catch to `setUpSplits` then the error is being thrown by `Split.load(slug)`.
```javascript
return Split.load(slug)
      .catch((error) => {
        console.log(error);
        return {};
      })
      .then((splits) => {
```

app/pages/project/index.jsx
srallen: It can't since that is what the key is in the SplitVariant resource itself. Michael Parrish setup the split key naming convention that way and I'm not sure why. It may make more sense in ruby.

app/pages/project/index.jsx
srallen: Good idea.

app/pages/project/index.jsx
srallen: It is indeed a hash of keyed splits. In theory, a project can have many experiments running so many splits could be returned. In practice, projects usually a single experiment at a time.

app/pages/project/home/project-home.jsx
srallen: I'll have to think about this. `props.showWorkflowButtons` is set to true if the project is configured to use workflow assignment or if they have `user_chooses_workflow` enabled. I could check for the split there, however, I wanted to use the `VisibilitySplit` component from the Seven-Ten-Client and have the visibility logic separate from the logic that is independent of an experiment running. I can see how it's confusing though here. 

I think the solution I'll do is also add a boolean to the split variant for the visibility split of the get started link and wrap the `Link` component also with `VisibilitySplit` so it's more clear what is going on.

srallen:
Ok this is ready for review again. I've address the review comments and then also include a few updates to the text and added Snapshot Wisconsin to the projects that can get the workflow promotion dialog. I know these are hardcoded and need to move to being stored in the project contents, but that's out of scope here.

app/pages/project/home/project-home.spec.js
```diff
@@ -262,4 +262,95 @@ describe('ProjectHome', function() {
...
+      });
+
+      it('should render the ProjectHomeWorkflowButtons', function () {
+        expect(wrapper.find('ProjectHomeWorkflowButtons')).to.have.lengthOf(0);
+      });
+
+      it("should not render a 'Get Started' link to the classify page", function () {
+        expect(wrapper.find({ to: `/projects/${project.slug}/classify` })).to.have.lengthOf(1);
+      });
+    });
```
eatyourgreens: The wording of the last two tests doesn’t match the expectations. Should and should not are the wrong way round.

eatyourgreens:
Did you figure out why Seven-Ten started throwing errors? I didn’t see any thing wrong with the code, compared to what’s on master.

app/pages/project/home/project-home.spec.js
srallen: Good catch! Thanks!

srallen:
I'm not sure why it started to throw. Maybe it's always been broken in the client, but now that the loading is a part of the promises that are being waited on and the promise itself is being returned that now we're seeing it? I was going to take a look at that next and see if we can get a patch release soon to follow.

eatyourgreens:
Sounds good to me. I think this is good to merge now.

eatyourgreens:
That `TypeError: Cannot read property 'forEach' of undefined` is being thrown by `Split.load()` on master too. It's just that it didn't halt the classifier until you moved the call into the `Promise.all()` block.

PR #4790, opened 2018-08-16
Store a single subject queue and classification

eatyourgreens:
Builds on #4781.
Remove upcomingSubjects.forWorkflow and currentClassifications.forWorkflow.
Simplify redux actions to work with a single classification and queue.
Update ProjectClassifyPage to use props.classification and props.upcomingSubjects.
Modify WorkflowSelection so that its children don't render unless there is a workflow loaded.

Staging branch URL: https://subject-queue-workflow.pfe-preview.zooniverse.org

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4797, opened 2018-08-21
Add subject queue interventions

eatyourgreens:
Staging branch URL: https://subject-queue-intervention.pfe-preview.zooniverse.org/

Updates the sugar notifications listener to look for experiments messages of the following type:
```javascript
{"experiments": 
  [{
    "user_id": "1", 
    "message": {
       "type": "subject_queue", 
       "subject_ids": [1,2],
       "workflow_id": "1234"
    }, 
    "url": "test", 
    "delivered": "false"
  }]
}

```
The subjects are requested from the API, then added to the upcoming subject queue if the workflow ID matches the current workflow.

TODO: 
- [x] record whether or not a subject, and its classification, came from a Sugar message.

~~Requires #4781 and #4790.~~

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/redux/ducks/interventions.js
+++ b/app/redux/ducks/interventions.js
@@ -1,11 +1,16 @@
+import apiClient from 'panoptes-client/lib/api-client';
 import { sugarClient } from 'panoptes-client/lib/sugar';
+import { addSubjects } from './classify';
 
 const ADD_NOTIFICATION = 'pfe/interventions/ADD_NOTIFICATION';
 const DISMISS_NOTIFICATION = 'pfe/interventions/DISMISS_NOTIFICATION';
+const ERROR = 'pfe/interventions/ERROR';
+const FETCH_SUBJECTS = 'pfe/interventions/FETCH_SUBJECTS';
 const SUBSCRIBE = 'pfe/interventions/SUBSCRIBE';
 const UNSUBSCRIBE = 'pfe/interventions/UNSUBSCRIBE';
 
 const initialState = {
+  error: null,
   notifications: []
 };
 
@@ -18,6 +23,10 @@ export default function reducer(state = initialState, action = {}) {
     case DISMISS_NOTIFICATION:
       notifications.pop();
       return Object.assign({}, state, { notifications });
+    case ERROR: {
+      const error = action.payload;
+      return Object.assign({}, state, { error });
+    }
     case SUBSCRIBE:
       sugarClient.subscribeTo(action.payload);
       return state;
@@ -37,9 +46,32 @@ export function unsubscribe(channel) {
   return { type: UNSUBSCRIBE, payload: channel };
 }
 
-export function notify(message) {
-  // Do something with notifications
-  return { type: ADD_NOTIFICATION, payload: message };
+export function notify(notification) {
+  const { message } = notification.data;
+  const { type } = message;
+  if (type === 'subject_queue') {
+    const subjectIDs = message.subject_ids;
+    const workflowID = message.workflow_id;
+    return (dispatch) => {
+      dispatch({
+        type: FETCH_SUBJECTS,
+        payload: subjectIDs
+      });
+      apiClient.type('subjects').get(subjectIDs)
+      .catch((error) => {
+        dispatch({
+          type: ERROR,
+          payload: error
+        });
+        return [];
+      })
+      .then((subjects) => {
+        dispatch(addSubjects(subjects, workflowID));
+      });
+    };
+  } else {
+    return { type: ADD_NOTIFICATION, payload: notification };
+  }
 }
 
 export function dismiss() {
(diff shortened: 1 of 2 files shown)


srallen:
The PR for `subject-queue-workflow` merged yesterday so I've updated the base to merge to master. I can take a look at resolving the merge conflict.

shaunanoordin:
Much thanks to Cam for providing information on how to test notifications: https://github.com/zooniverse/sugar#manual-testing-of-notifications

eatyourgreens:
This needs to be rebased to master.

shaunanoordin:
### Feature Clarification

Thanks to Jim for the clarification on the intended experience of this feature. The 'subject_queue' _adds Subjects to a user's queue_ via the Notifications system. (Most likely, to the tail of the user's current subject queue.)

### Testing Update

Hence, the new test procedure is as follows:
1. User X starts working on classifications in workflow Y in Project Z.
  - let's assume Subject queue is as follows: `[SbjA, SbjB, SbjC, SbjD]`
2. Separately, an Admin User sends a 'notification' or an ['experiments message'](https://github.com/zooniverse/sugar#manual-testing-of-notifications) _specifically_ to User X.
  - The experiments message specifies new Subject IDs, (e.g. `[SbjFoo, SbjBar]`), the target User X, and the target workflow Y.
3. User X's Subject queue should (at runtime) now add the new Subjects to their queue.
  - So in our example, the queue should now look like `[SbjA, SbjB, SbjC, SbjD, SbjFoo, SbjBar]`

Failing conditions to check for:
- If User X doesn't receive the new Subjects in their queue when the message is received, this is an error. 
- If User X receives the new Subjects in their queue when they're _not_ in Workflow Y, this is an error.
- These changes to the Subject queue should only happen _once per experiment message._

Further things to test:
- what if the new subjects were added to the queue, but the user doesn't actually reach/finish classifying all the new subjects? Is this still marked as "user has received the notification, and we're done with that"?
- confirm that the new subjects are added to the tail of the queue, not the head.

shaunanoordin:
To be clear, this PR looks _OK_ in the sense that _it doesn't break anything that currently exists_ and the code looks clean. The only problem at the moment is that we still need to test the new feature this PR is trying to add - that's mostly a matter of being able to send someone a notification/experiment message.

eatyourgreens:
I've updated the example in the PR description to what I think the posted Sugar notification should look like. Still needs testing though.

marten:
Looks like this is going to be appending the subjects from the intervention message at the end of the existing subject queue. This sounds like it's probably not what researchers actually want, presumably the idea is that the subjects they send should appear as quickly as possible.

@chrislintott @mrniaboc any thoughts?

mrniaboc:
Yes, for this specific experiment with MSR/BGU they really want the
subjects to be added to the front of the queue so the volunteers see them
immediately once they finish their current classification.

I think that functionality will also be really important for use cases like
PRN and LSST.

eatyourgreens:
I've split the `ADD_SUBJECTS` action type into `APPEND_SUBJECTS` and `PREPEND_SUBJECTS`. Fetching from the API appends, but notifications from Sugar should prepend.

eatyourgreens:
Here's the script I'm using to test this:
```shell
curl -vv \
     -X "POST" "https://notifications-staging.zooniverse.org/experiment" \
     -H 'Content-Type: application/json' \
     -u 'USERNAME:PASSWORD' \
     -d $'{
  "experiments": [{
    "message": {
       "type": "subject_queue", 
       "subject_ids": ["1","2"],
       "workflow_id": "1483"
    },
    "user_id": "1755",
    "url": "http://test.net",
    "delivered": false
  }]
}'
```
where `USERNAME` and `PASSWORD` are the basic auth credentials for `notifications-staging`.

eatyourgreens:
The easiest way to test the front-end is probably to install the redux dev tools and track the actions and state changes when notifications are received.
Screenshot of the redux dev tools, highlighting the pfe/interventions/FETCH_SUBJECTS action

app/redux/ducks/interventions.js
```diff
@@ -1,11 +1,16 @@
+import { apiClient } from 'panoptes-client/lib/api-client';
```
eatyourgreens: This should be `import apiClient from…`.

eatyourgreens:
@mrniaboc @marten based on our Slack conversation yesterday, I've added `classification.metadata.source` to record where the subject came from. It should be `sugar` for intervention subjects, `api` otherwise.

mrniaboc:
Awesome!

rogerhutchings:
I've successfully tested this now by posting a notification and having the redux store update, and the submitted classification has the correct `source` property in `metadata` - LGTM.

Are you happy for this to merge @eatyourgreens?

eatyourgreens:
Sounds good to me. The message format hasn't been finalised yet but that can be changed later.

PR #4813, opened 2018-08-24
Fix linting

rogerhutchings:
Following on from #4806, this bumps the remaining linting packages so that `npm run lint` actually works.


package.json
```diff
@@ -85,14 +86,13 @@
     "ejs": "~2.5.7",
     "enzyme": "~3.3.0",
     "enzyme-adapter-react-16": "~1.2.0",
-    "eslint": "~3.9.1",
-    "eslint-config-airbnb": "~13.0.0",
+    "eslint": "~5.4.0",
+    "eslint-config-airbnb": "~17.1.0",
     "eslint-config-standard": "~5.3.1",
```
srallen: Should this be removed along with the plugin?

rogerhutchings: Weird, I thought I had. But yes

PR #4845, opened 2018-09-03
Add a saveAnnotations action to the classifier

eatyourgreens:
Replace `classification.update` with a saveAnnotations action in the classifier, to make it explicit that annotations are saved when the Classifier component unmounts.
Update the wording on a couple of the Classifier tests to make it less confusing.

Staging branch URL: https://save-annotations.pfe-preview.zooniverse.org/

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -12,6 +12,7 @@ import { getSessionID } from '../lib/session';
 import preloadSubject from '../lib/preload-subject';
 import workflowAllowsFlipbook from '../lib/workflow-allows-flipbook';
 import workflowAllowsSeparateFrames from '../lib/workflow-allows-separate-frames';
+import * as classifierActions from '../redux/ducks/classify';
 import * as feedbackActions from '../redux/ducks/feedback';
 import * as interventionActions from '../redux/ducks/interventions';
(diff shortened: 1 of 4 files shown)


eatyourgreens:
@srallen @rogerhutchings there are a few places in the classifier that call `classification.update`, mostly to update metadata. Should those calls also be moved into action creators now that the current classification is in the redux store?

eatyourgreens:
Rebased to latest master.

rogerhutchings:
> there are a few places in the classifier that call classification.update, mostly to update metadata. Should those calls also be moved into action creators now that the current classification is in the redux store?

If we can - would that mean we could stop passing around the classification as a prop? 😱

eatyourgreens:
I'm a bit confused by the endpoint comment. This doesn't send anything to Panoptes, it just preserves your annotations when you navigate away from the classify page. We could certainly update the classifier to save work-in-progress to Panoptes. That might be a bigger change than this PR.

rogerhutchings:
Ah right - I assumed it would be saving an incomplete classification back to the API (not sure why). Forget I mentioned it

PR #4799, opened 2018-08-21
Fix discipline and other tag inputs

mcbouslog:
Staging branch URL: https://fix-tags-inputs.pfe-preview.zooniverse.org/

Fixes broken discipline and other tag UI in project details lab section.

Updating on project resource instead of using `handleInputChange`.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #4914, opened 2018-09-21
Fix handleTagChange

### LINKED ISSUE #4801: handleInputChange silently converts array and object values to strings
https://github.com/zooniverse/Panoptes-Front-End/blob/09cc8f46d34b2a6dbf413b14b3fa935232ab5052/app/lib/handle-input-change.coffee#L21-L24

`DOMPurify.sanitize` silently converts `value` to a string, breaking `resource.update(changes)` when value is an array or an object.

This broke the project builder tag inputs in #4799.

### LINKED ISSUE #4907: Adding 'discipline' and 'other' tags is broken
Tags won't save, and I see the following error in the console:

image

noahmalmed:
Staging branch URL: https://fix_4801.pfe-preview.zooniverse.org/

Fixes #4801 Fixes #4907

Just moved this to be an inline function, so the tag array doesn't get converted to a string. @srallen is this the type of thing you had in mind as a solution?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/lib/sanitize-array-input.coffee
+++ b/app/lib/sanitize-array-input.coffee
@@ -1,0 +1,7 @@
+createDOMPurify = require('dompurify');
+
+DOMPurify = createDOMPurify(window);
+
+module.exports = (array) ->
+  sanitizedArray = array.map (string) -> DOMPurify.sanitize(string)
+  return sanitizedArray.filter (string) -> string != ''
--- a/app/lib/sanitize-array-input.coffee
+++ b/app/lib/sanitize-array-input.coffee
@@ -1,7 +1,0 @@
-createDOMPurify = require('dompurify');
-
-DOMPurify = createDOMPurify(window);
-
-module.exports = (array) ->
-  sanitizedArray = array.map (string) -> DOMPurify.sanitize(string)
-  return sanitizedArray.filter (string) -> string != ''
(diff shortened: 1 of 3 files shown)


eatyourgreens:
This is a duplicate of #4799, which bypassed the bug by bypassing DOMPurify. It's dangerous to pass unsanitised user input to Panoptes. I'd recommend sanitising the tags by passing something like `{tags: value.map(tag => DOMPurify.sanitise(tag))}` to the update method.

https://github.com/zooniverse/Panoptes-Front-End/pull/4799#issuecomment-414983273

noahmalmed:
👍 that makes sense, I'll update this

noahmalmed:
-Updated to sanitize each member of the tag array
- Updated to sanitize the array before we update state, so the sanitized values are displayed to the user

app/lib/sanitize-array-input.coffee
```diff
@@ -0,0 +1,7 @@
+createDOMPurify = require('dompurify');
```
eatyourgreens: Can this helper be written in JS? We're dropping coffeescript in favour of ES6.

app/lib/sanitize-array-input.coffee
```diff
@@ -0,0 +1,7 @@
...
+
+DOMPurify = createDOMPurify(window);
+
+module.exports = (array) ->
+  sanitizedArray = array.map (string) -> DOMPurify.sanitize(string)
+  return sanitizedArray.filter (string) -> string != ''
```
eatyourgreens: Use strict inequality (!==) here to clean up empty strings.

noahmalmed:
@eatyourgreens changes made per your request

PR #4916, opened 2018-09-24
Move the fan tool delete button

### LINKED ISSUE #4912: very difficult to draw fans that have same starting point but are in different directions with the fan tool
Feedback after further playing around with the new  fan tool for P4:

The new fan tool for P4 has an issue that will need a solution before we can launch the new P4 We have not too infrequent cases where fans have the same vent but are blown by the wind in different directions

On the Ouroboros P4 the x to remove the fan was in the center of the marker. The panoptes version has the x at the end of the fan. So now if you try and make a fan at the same starting point. You either need to move off the x or draw the fan somewhere else and move it into position. We want to encourage people to draw fans with the same origin point, since this is one of the science drivers for P4 to look at whether the wind is changing over a HiRISE monitoring season. 

Can the 'x' to delete a fan be moved in the fan tool be moved somewhere from the starting tip?

eatyourgreens:
Move the delete button from the end of the fan to just outside the curved top of the shape. 

Staging branch URL: https://fan-tool.pfe-preview.zooniverse.org/

Fixes #4912.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/drawing-tools/components/Fan/fan.jsx
+++ b/app/classifier/drawing-tools/components/Fan/fan.jsx
@@ -131,8 +131,8 @@ class Fan extends React.Component {
           <React.Fragment>
             <DeleteButton
               tool={this}
-              x={spreadX}
-              y={0}
+              x={radius + 25}
+              y={-25}
               getScreenCurrentTransformationMatrix={getScreenCurrentTransformationMatrix}
             />
             <DragHandle
--- a/app/classifier/drawing-tools/components/Fan/fan.jsx
+++ b/app/classifier/drawing-tools/components/Fan/fan.jsx
@@ -131,7 +131,7 @@ class Fan extends React.Component {
           <React.Fragment>
             <DeleteButton
               tool={this}
-              x={-20}
+              x={spreadX}
               y={0}
               getScreenCurrentTransformationMatrix={getScreenCurrentTransformationMatrix}
             />


eatyourgreens:
I've added @mschwamb as a reviewer so she can double check the position of the delete button.

There's a minimum size of 10px set in the code at the moment. Any drawn fans have to be larger than that size, otherwise they're automatically deleted when you finish drawing. We could also consider making those constants into options that are set in the project builder. I _think_ that's supported by the drawing tools, although I haven't actually looked to see how they work in the project builder.
https://github.com/zooniverse/Panoptes-Front-End/blob/44af03915b86cc2b4b66b38c72af161c18c91cba/app/classifier/drawing-tools/components/Fan/fan.jsx#L9-L10

mschwamb:
I can have a look on the minimum size used in the current P4. I think the 'x' outside the fan looks better and works better. The 'x' is bigger than in P4/filled in with a solid color that made it harder to see what one was outlining in the panoptes version.

I'm noticing that I'm having a hard time getting an intermediate fan width. It seems to easily go to 180 degrees, but nothing in between. Is there an issue with the calculation of the angle that's going beyond 180 degrees?

fan_tool

eatyourgreens:
I've opened #4925 for the spread angle bug.

PR #4923, opened 2018-09-25
Classification queue: save each queued classification once only

### LINKED ISSUE #4911: PFE potentially submits multiple copies of a Classification
## Bug Report

It is currently possible for PFE to submit multiple copies of the same Classification. There are multiple ways this bug can occur, but this report focuses on the confirmed, replicable issue that relies on **network/time out/latency errors.**

### Analysis

In PFE, when a Classification is submitted, it is added to a "submission queue". If a Classification isn't successfully submitted the first time, it's then _re-submitted_ when a _new Classification_ is submitted and the queue processed again.

However, the re-submission logic doesn't account for what happens when the initial Classification submission _doesn't fail,_ but just takes _a long time to process._

### Analysis - Example Scenario

- User submits Classification A (attempt 1) to PFE, then proceeds to the next Subject.
  - However, Classification A (attempt 1) takes a long time to resolve, possibly due to network latency or a slow API response.
- User submits Classification B to PFE.
  - PFE notes that Classification A (attempt 1) hasn't succeeded or failed (i.e. it's still being sent, with a status of 'pending')
  - Hence, PFE attempts to re-submit Classification A (attempt 2) along with Classification B.
- After a while, _all three_ submissions resolve successfully.
  - Classification A (attempt 1) resolves with a 201.
  - Classification B resolves with a 201.
  - Classification A (attempt 2) resolves with a 201.
- Now, PFE has _two copies_ of Classification A, and one copy of Classification B.

Note that, in this scenario, the two copies of Classification A have the same **metadata** and **annotations,** but different classification IDs.

### Steps to Replicate

This testing setup uses localhost (staging), OSX+Chrome, and an [example Project](https://pfe-preview.zooniverse.org/projects/darkeshard/transformers/classify?reload=0&workflow=3272) with a small subject set and simple text field workflow. This test uses Chrome's dev tools to simulate a slow network or a laggy response from the API.

Note: before I test, I run through the project and make one "filler" classification for each Subject. This ensures that each Subject image is cached in my browser, so Chrome's simulated network lag _only affects the API calls._

Testing steps:
- Open the example Project's Classification page.
- Open Chrome Dev Tools (Cmd+I on OSX) to simulate network lag/slow API response times.
  - Go to the Performance tab, click on Network dropdown, and select/create a custom Network Throttling Profile with some significant latency, e.g. 30 or 60 seconds.
- Switch the Chrome Dev Tool window to Network to monitor Classification submissions (i.e. POSTs to the `/classifications` API endpoint)
- Submit the first Classification with some easily identifiable annotation (e.g. "Hello this is Classification A"`)
  - You should see the first POST to `/classifications`, and it should be in progress (status: pending)
  - Quickly move to the next step while that POST is in progress.
- Submit the second Classification with easily identifiable annotations. (e.g. "G'day this is Classification B")
  - You should now see _three_ POSTs to `/classifications`, all in progress (status: pending)
  - Two of these POSTS should be for Classification A, and the other for Classification B.
- Wait a while (i.e. however long you set the latency to be) for all three POSTs to resolve.
  - You should now see _three_ 201 responses (resources created) for your _two_ Classification submissions.
- Download the Project's Classification exports to confirm that Classification A was received twice.

I'll post my test results separately to avoid cluttering this post further.

### Status
Issue identified, but solution is still pending.

Warning: this analysis does not exclude other scenarios where duplicate Classifications are found, esp. if those Classifications have slightly different metadata (e.g. the known case where `metadata.finished_at` of two potentially duplicate Classifications are off by a microsecond.)

Tagging @camallen for the initial problem report, and @rogerhutchings for sanity checking/counter thoughts on the PFE queue.

eatyourgreens:
Add a test for the case where the API save request is not resolved between calls to ClassificationQueue.add().

Fix a bug where pending classifications persisted in the queue, resulting in them being sent again each time a new classification was added to the queue.

Fixes some bugs in the fake API classes that break the tests.

Todo:
- [x] Fix the duplicate classification bug #4911
- [x] Fix broken async tests

Staging branch URL: https://classification-queue.pfe-preview.zooniverse.org

Fixes #4911.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -18,9 +18,9 @@ class ClassificationQueue {
 
     try {
       this._saveQueue(queue);
-      console.info('Queued classifications:', queue.length);
+      if (process.env.BABEL_ENV !== 'test') console.info('Queued classifications:', queue.length);
     } catch (error) {
-      console.error('Failed to queue classification:', error);
+      if (process.env.BABEL_ENV !== 'test') console.error('Failed to queue classification:', error);
     }
 
     this.flushToBackend();
@@ -40,11 +40,13 @@ class ClassificationQueue {
   }
 
   flushToBackend() {
-    const queue = this._loadQueue();
+    const pendingClassifications = this._loadQueue();
+    const failedClassifications = [];
+    this._saveQueue(failedClassifications);
 
-    if (queue.length > 0) {
-      console.log('Saving queued classifications:', queue.length);
-      queue.forEach((classificationData) => {
+    if (pendingClassifications.length > 0) {
+      if (process.env.BABEL_ENV !== 'test') console.log('Saving queued classifications:', pendingClassifications.length);
+      pendingClassifications.forEach((classificationData) => {
         this.apiClient.type('classifications').create(classificationData).save().then((actualClassification) => {
           console.log('Saved classification', actualClassification.id);
           this.onClassificationSaved(actualClassification);
@@ -65,11 +67,8 @@ class ClassificationQueue {
 
           if (error.status === 422) {
             console.error('Dropping malformed classification permanently', classificationData);
-            const indexInQueue = queue.indexOf(classificationData);
-            queue.splice(indexInQueue, 1);
-
             try {
-              this._saveQueue(queue);
+              this.add(classificationData);
             } catch (saveQueueError) {
               console.error('Failed to update classification queue:', saveQueueError);
             }
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -5,8 +5,8 @@ const FAILED_CLASSIFICATION_QUEUE_NAME = 'failed-classifications';
 const MAX_RECENTS = 10;
 
 class ClassificationQueue {
-  constructor(storage, api, onClassificationSaved) {
-    this.storage = storage || window.localStorage;
+  constructor(api, onClassificationSaved) {
+    this.storage = window.localStorage;
     this.apiClient = api || apiClient;
     this.recents = [];
     this.onClassificationSaved = onClassificationSaved;
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -51,16 +51,6 @@ class ClassificationQueue {
           console.log('Saved classification', actualClassification.id);
           this.onClassificationSaved(actualClassification);
           this.addRecent(actualClassification);
-
-          const indexInQueue = queue.indexOf(classificationData);
-          queue.splice(indexInQueue, 1);
-
-          try {
-            this._saveQueue(queue);
-            console.info('Saved a queued classification, remaining:', queue.length);
-          } catch (error) {
-            console.error('Failed to update classification queue:', error);
-          }
         })
         .catch((error) => {
           if (process.env.BABEL_ENV !== 'test') console.error('Failed to save a queued classification:', error);
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -13,6 +13,11 @@ class ClassificationQueue {
   }
 
   add(classification) {
+    this.store(classification);
+    return this.flushToBackend();
+  }
+
+  store(classification) {
     const queue = this._loadQueue();
     queue.push(classification);
 
@@ -22,8 +27,6 @@ class ClassificationQueue {
     } catch (error) {
       if (process.env.BABEL_ENV !== 'test') console.error('Failed to queue classification:', error);
     }
-
-    return this.flushToBackend();
   }
 
   length() {
@@ -58,7 +61,7 @@ class ClassificationQueue {
           if (error.status === 422) {
             console.error('Dropping malformed classification permanently', classificationData);
             try {
-              this.add(classificationData);
+              this.store(classificationData);
             } catch (saveQueueError) {
               console.error('Failed to update classification queue:', saveQueueError);
             }
--- a/app/lib/classification-queue.js
+++ b/app/lib/classification-queue.js
@@ -47,31 +47,28 @@ class ClassificationQueue {
     const failedClassifications = [];
     this._saveQueue(failedClassifications);
 
-    if (pendingClassifications.length > 0) {
-      if (process.env.BABEL_ENV !== 'test') console.log('Saving queued classifications:', pendingClassifications.length);
-      return Promise.all(pendingClassifications.map((classificationData) => {
-        return this.apiClient.type('classifications').create(classificationData).save()
-        .then((actualClassification) => {
-          console.log('Saved classification', actualClassification.id);
-          this.onClassificationSaved(actualClassification);
-          this.addRecent(actualClassification);
-        })
-        .catch((error) => {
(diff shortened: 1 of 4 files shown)


shaunanoordin:
To be clear, there are _two things_ to test for with these changes.

1. If a user submits multiple Classifications in rapid order, and there's network lag delaying the response of each submission, the system does **not submit duplicates**
2. If a user submits a Classification, and that submission fails for whatever reason, the system **tries again**

I'm emphasising point 2 because I've mostly been testing for point 1 up to now (i.e. the main point of this PR).

I now understand in what _specific scenario_ the previous _"add to queue first, then if success, remove from queue"_ makes a modicum of sense - e.g. if a user makes 10 Classifications while [Panoptes is spontaneously taking a break](https://giphy.com/gifs/fire-richard-ayoade-the-it-crowd-zyclIRxMwlY40), and then reloads the browser window (because they couldn't see any new Subject because the queue is empty and they think refreshing might fix it), then _all 10 Classifications will be lost_ because none of them were saved to local.

eatyourgreens:
If Panoptes is down, then your 10 classifications should be saved to local storage when the API request times out (although we need to test that.) If the time out is long, you'll lose them if you refresh the page before the browser gives up on the request.

If you reload the page when a request is in progress, you might lose that classification. Then again, that classification might be successfully saved too. There's no way of knowing because the response from Panoptes won't be received by the refreshed page.

shaunanoordin:
Yoinks, I just realise there's a major error where Classifications aren't being sent at all; I keep seeing `_this.apiClient.type is not a function` errors on Classification submission. I didn't realise this before because it was failing silently. Hold on, checking...

app/lib/classification-queue.js
```diff
@@ -40,43 +43,35 @@ class ClassificationQueue {
...
-      queue.forEach((classificationData) => {
-        this.apiClient.type('classifications').create(classificationData).save().then((actualClassification) => {
+    const pendingClassifications = this._loadQueue();
+    const failedClassifications = [];
+    this._saveQueue(failedClassifications);
+
+    if (pendingClassifications.length > 0) {
+      if (process.env.BABEL_ENV !== 'test') console.log('Saving queued classifications:', pendingClassifications.length);
+      return Promise.all(pendingClassifications.map((classificationData) => {
+        return this.apiClient.type('classifications').create(classificationData).save()
```
shaunanoordin: I think I'm seeing errors here - apparently `this.apiClient` doesn't exist? Debugging in process.

shaunanoordin: Wait, sorry, it's `this.apiClient.type` that doesn't exist.

It seems that `this.apiClient` is the following object:

```
Storage {failed-classifications: "[]", minilogSettings: "true", pusherTransportEncrypted: "{"timestamp":1538059649538,"transport":"ws","latency":628}", length: 3}
```

Huuhhh?

eatyourgreens: Dammit. Why is that not erroring when running the tests. `npm run test-one app/lib/classification-queue.spec.js`

shaunanoordin: It's the `constructor`

This PR's updated `constructor` call changes it from `constructor(storage, api, onClassificationSaved)` to `constructor(api, onClassificationSaved)`, but the defining call of `new QueueStorage` (see `pages/projects/classify.jsx`) still defines some sort of Storage object as the first argument.

eatyourgreens: D'oh!

eatyourgreens: By the way, I'm going to rebase this because a bunch of packages have been updated recently.

eatyourgreens: That extra storage argument should be gone now.

eatyourgreens:
Do service workers persist across page refreshes? We could look at moving the API requests into a service worker, if that's the case.

srallen:
@eatyourgreens yes they do. They can work offline and independent of session as well. The next iteration of this definitely should be a service worker.

shaunanoordin:
## PR Review (Update)

### Testing
Basic testing looks good:
- users can login to/logout from PFE, access the projects page, access a project page, and submit classifications.

Primary test looks good so far:
- users can submit multiple Classifications in rapid succession and not have duplicates
  - As explained in Issue #4911 , this is tested by sending, say, 6 Classifications with an artificial 60s lag on the network. The Network communication was monitored to ensure exactly 6 matching calls to the `/classifications` API endpoint were made and resolved.
- users can submit Classifications, and should they fail, they'll be added to the queue to be retried later.
  - This is a bit more complicated to test:
  - First, Classification C1 is submitted with normal network conditions.
  - Then, Classification C2 is submitted but with the network artificially disabled.
  - Finally, Classification C3 is submitted with normal network conditions restored.
  - As expected, we see exactly 4 calls to the `/classifications` endpoint: C1 success, C2 fail, C2 success (retried from queue when C3 was submitted), C3 success. 👍 

Tested using localhost OSX+Chrome68 on a simple project (`staging: /projects/darkeshard/transformers`)

### Status
Testing looks functional, and I have no comments to make about the code. I think this is looking good. 👍

app/lib/classification-queue.js
```diff
@@ -40,43 +43,35 @@ class ClassificationQueue {
...
-
-    if (queue.length > 0) {
-      console.log('Saving queued classifications:', queue.length);
-      queue.forEach((classificationData) => {
-        this.apiClient.type('classifications').create(classificationData).save().then((actualClassification) => {
+    const pendingClassifications = this._loadQueue();
+    const failedClassifications = [];
+    this._saveQueue(failedClassifications);
+
+    if (pendingClassifications.length > 0) {
```
eatyourgreens: This if can be removed, since the promise resolves with an empty array when the map is empty.

eatyourgreens:
I'm seeing uncaught test failures on this branch, so maybe hold off on merging just yet.
```
keeps classifications in localStorage if backend fails
      ✓ should not save failed classifications
(node:9497) UnhandledPromiseRejectionWarning: AssertionError: expected [ Array(1) ] to have a length of 0 but got 1
    at lengthOf (/Users/odonnell/zooniverse/Panoptes-Front-End/app/lib/classification-queue.spec.js:59:41)
    at <anonymous>
    at runMicrotasksCallback (internal/process/next_tick.js:121:5)
    at _combinedTickCallback (internal/process/next_tick.js:131:7)
    at process._tickCallback (internal/process/next_tick.js:180:9)
(node:9497) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 2)
(node:9497) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

eatyourgreens:
Tests should be fixed now. There were a couple of syntax errors in the test spec that caused passing tests to fail, but also prevented failures in the `then` blocks from being caught.

eatyourgreens:
By the way, I've running `test-one` to check the tests. `npm run test-one app/lib/classification-queue.spec.js`.

PR #4924, opened 2018-09-25
Update to Babel 7 and Webpack 4

srallen:
Staging branch URL: https://update-babel-webpack.pfe-preview.zooniverse.org/

Supersedes all the dependabot PRs for updating babel and webpack. I ran [babel-upgrade](https://github.com/babel/babel-upgrade) which nearly automates all of the updates to babel 7. I removed the type checking plugin because it hasn't been updated in two years and I'm not convinced it was doing anything.

I also updated to webpack 4, removed the old deprecated HMR setup, migrated to webpack-dev-server which has HMR built in and removed the dev-server.js file. 

TODO:
- [x] Migrate from `extract-text-webpack-plugin` to [mini-css-extract-plugin](https://github.com/webpack-contrib/mini-css-extract-plugin). Extract text plugin is deprecated for webpack 4.
-  [x] Testing production builds

In a separate PR update the chunks setup to eliminate the duplicate code reports we're seeing in the dashboard stats.  

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
`npm run serve-static` uses the production API. I'm not sure if that's intended.

`npm start` seems to start up a local server but visiting the site gives me `Invalid Host header` in the browser.

`npm run stage` successfully builds and deploys to https://update-babel-webpack.pfe-preview.zooniverse.org/ but also uses the production API, not staging.

eatyourgreens:
The Invalid Host error happens because I use http://localhost.pfe-preview.zooniverse.org:3735/ locally, so that I can sign in.

http://localhost:3735 works, but is blocked from making authenticated requests.

webpack.dev.js
```diff
@@ -7,6 +7,10 @@ var CopyWebpackPlugin = require('copy-webpack-plugin');
 var DashboardPlugin = require('webpack-dashboard/plugin');
 
 var config = {
+  mode: 'development',
+  devServer: {
+    port: 3735
```
eatyourgreens: Looking at the old dev server config, should this also set `host: process.env.HOST || "localhost"`?

srallen: Yeah I can add that. I'm not sure if anyone was using it before, but let's keep parity anyway.

srallen:
I think I've resolved all of the issues by adding a few options:
- `allowedHosts` is now added with `.zooniverse.org` which is a django style wildcard to allow any subdomains that may be setup in your `/etc/hosts` file.
- `host` is defined to have parity with the old dev server in setting an environment variable for host. 
- `https` is set to true which will generate a self-signed cert. Chrome will complain, but just allow it. This appears to be getting `https://localhost:3735` working for auth.
- `open` is set to true which will open `https://localhost:3735` in the browser as soon as the dev server starts. I like this convenience but if others find it annoying, then we can turn it off.
- `overlay` is true which will display compiling errors in the browser

In the `webpack.build.js` file, I've added back the custom configuration for setting the `NODE_ENV` which fixes the environment for staging deploys. The new config for `mode` if set to production will auto set the environment for production, but we don't always want that when using the build config. We still do want to use the production mode since this auto enables uglifying and a few other options.

webpack.dev.js
eatyourgreens: I'm not sure if anyone was using it either. At least, I don't see it in the start scripts anywhere.

eatyourgreens:
Looks good to me. I've got a local site up and running, and authenticating to Panoptes staging, on http://localhost.pfe-preview.zooniverse.org:3735/

Nice to have HTTPS on localhost now.

I'm not sure about the auto-opening behaviour. Some of our older projects did that (readymade?) and it could get very annoying if your default browser is not the browser you were working in.

eatyourgreens:
Not a blocking comment, but `npm start` opens https://localhost:3735 in Safari 11 for me, which then gives me a bunch of stern unsigned certificate warnings. Ignoring those, and browsing the site, signing in returns an invalid grant error from the staging API. I think you'll need to alias a zooniverse.org domain locally if you want to test anything that requires signing in.

PR #4934, opened 2018-09-26
New, slimmer image and updated Dockerfile

zwolf:
There were a few issues early this week with PFE deployment Jenkins jobs failing to complete. The issue was node crashing with an error that indicated it was out of memory. Taking a look at the Dockerfile, it seemed that it was building a larger than necessary image (full ubuntu) and loading a bunch of stuff it didn't need (i.e. phantomjs) just to run the npm build. 

This PR switches to a slim alpine image straight from npm and only installs git because npm needs it (and bash, but that's probably extraneous). The extra Dockerfile.deps was removed with a piece or two collapsed into the main one. This should keep total image size way down and eliminate the need for two separate images.

This is WIP because while I tested a few `docker run` commands, I'm not sure how to test short of straight up deploying master to staging and I'd like another pair of PFE-familiar and/or devops eyeballs looking when that happens.


Dockerfile
```diff
@@ -1,4 +1,8 @@
-FROM zooniverse/panoptes-front-end:deps
+FROM node:8.12-alpine
+
+WORKDIR /src
+
+RUN apk update && apk upgrade && apk add --no-cache bash git openssh
```
adammcmaster: It's a minor thing, but there's no need to update/upgrade here. The `--no-cache` option loads the package index automatically.

adammcmaster:
It should be safe to just run the normal stage command to test this.  `docker-compose run dev npm run stage`

Dockerfile
zwolf: Good to know. I'm gonna pare this down to just adding git since bash isn't needed and ssh is a dep of git.

zwolf:
Seems to work fine, I was able to successfully stage a branch here with `docker-compose run dev npm run stage`: https://dockerfile-update.pfe-preview.zooniverse.org/

PR #4978, opened 2018-10-10
Connect the dev classifier to the classify store

eatyourgreens:
Staging branch URL: https://dev-classifier.pfe-preview.zooniverse.org

- Connects the dev classifier up to the Redux store, so that it uses the stored classification, workflow and subject queue.
- Creates a new classification and queue when the component mounts and also whenever the classification is reset to null (login/logout resets the subject queue and classification.)
- Fixes the Next button so that it reloads the mock subject with a new classification.
- Adds a simple error handler that catches any errors and prints the component stack to the page when the classifier component crashes.
- Also fixes the RESUME_CLASSIFICATION action so that it uses the passed classification, rather than assuming that the correct classification is already stored in state.
- Updates the Webpack source map setting so that we get meaningful error messages in development, rather than cross-origin errors when components crash.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/dev-classifier/ClassificationViewer.jsx
+++ b/app/pages/dev-classifier/ClassificationViewer.jsx
@@ -21,7 +21,8 @@ export default class ClassificationViewer extends React.Component {
 
   static defaultProps = {
     annotations: [],
-    classification: null
+    classification: null,
+    workflow: {}
   }
 
   componentDidMount() {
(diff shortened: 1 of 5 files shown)


eatyourgreens:
The classifier still won't render if you load https://dev-classifier.pfe-preview.zooniverse.org/dev/classifier directly for an anonymous user, but I'm not sure how big of a deal that is. Anonymous users can see the dev classifier if they navigate to it from a different page.

I think this is something to do with the auth change in `main.cjsx` firing after the classifier mounts, instead of before. https://github.com/zooniverse/Panoptes-Front-End/blob/db3530ddc080d5e499be8712fd82f135adc3d5c0/app/main.cjsx#L17

eatyourgreens:
Anonymous use should be fixed by that last commit.

PR #4987, opened 2018-10-12
Fix the gold standard checkbox

### LINKED ISSUE #4564: Gold standard button not responsive
See video showing buggy behavior. 
In Chrome v65 on mac os high sierra

Initially reported by Chicago Wildlife Watch researcher 

https://drive.google.com/file/d/1YZnMwuIDqYaztxA8BAQgo4GrLYccPcuo/view?usp=sharing

This is lower priority, but perhaps an easy fix?

### LINKED ISSUE #4984: Gold-standard mode check-box doesn't get checked upon click.
## Expected behavior
To do a gold standard classification for a project, there is a check-box next to the Done button that pops-up after clicking on a cog button. When I click on the check-box, it is supposed to get 'checked' and display I am doing a gold standard classification by adding a star into the Done button, and add a star together with 'Gold Standard mode text' just below.
When I want to cancel the Gold standard mode, I click on the cog button to uncheck the box and the check-box changes to an empty check-box.

## Current behavior
When I click on the check-box, nothing seems to happen. No matter how many times I click or how long I wait. After escaping the pop-up with the check-box, it doesn't show a star for gold classification. But when I choose any category and go back to the list of categories, it suddenly shows the results of checking the check-box = I see the start together with 'Gold standard mode' text. 
When I decide to cancel the Gold standard mode, no matter how many times I click on the check-box, it stays checked. Just after choosing any category and getting back to the list of categories, it suddenly shows the results of canceling the Gold standard mode.

Why is that a problem? It seems I haven't succeeded in setting the gold standard mode when I first choose the category and then try to set the gold standard mode. I realize it was a Gold standard after finishing my classification. That may be too late. The same when I manage to set the gold standard and then realize there may be more animals in the bush and decide not to do a gold standard but cannot uncheck the box. Then it is safer to leave the classification instead of trying one's luck and clicking on the Done button to see the results. (Unless you know you can see the results when you go to any category and back. But that took me a while to figure out.)

## Steps to replicate
I guess trying to check the box and then uncheck it? I'll try to test that on another device later today.

## Additional information
Happens on Western Shield Camera Watch (new data released today). Unable to test it elsewhere.

Screenshot of the browser console errors:
ysh_screenshot_2018-10-11_15-46-43_gold_doesnt_check

- **Operating system:**
Ubuntu
- **Browser:**
Chromium 69

eatyourgreens:
Staging branch URL: https://gold-standard.pfe-preview.zooniverse.org/

Fixes #4984.
Fixes #4564.

Adds an action to toggle `classification.gold_standard` and binds the gold standard checkbox to that action.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -402,7 +402,6 @@ class Classifier extends React.Component {
             >
               {!!this.props.expertClassifier &&
                 <ExpertOptions
-                  classification={currentClassification}
                   userRoles={this.props.userRoles}
                   demoMode={this.props.demoMode}
                   onChangeDemoMode={this.props.onChangeDemoMode}
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -296,7 +296,7 @@ class Classifier extends React.Component {
   }
 
   render() {
-    const { interventions, user } = this.props;
+    const { actions, interventions, user } = this.props;
     const { showIntervention, showSummary, workflowHistory } = this.state;
     const currentTaskKey = workflowHistory.length > 0 ? workflowHistory[workflowHistory.length - 1] : null;
     const largeFormatImage = this.props.workflow.configuration.image_layout && this.props.workflow.configuration.image_layout.includes('no-max-height');
@@ -436,7 +436,7 @@ class Classifier extends React.Component {
                   <strong>Gold standard mode:</strong>
                   <br />
                   Please ensure this classification is completely accurate.{' '}
-                  <button type="button" className="secret-button" onClick={currentClassification.update.bind(currentClassification, { gold_standard: undefined })}>
+                  <button type="button" className="secret-button" onClick={() => actions.classify.toggleGoldStandard(undefined)}>
                     <u>Disable</u>
                   </button>
                 </small>
--- a/app/classifier/expert-options.jsx
+++ b/app/classifier/expert-options.jsx
@@ -1,11 +1,14 @@
 import PropTypes from 'prop-types';
 import React from 'react';
 import TriggeredModalForm from 'modal-form/triggered';
+import { bindActionCreators } from 'redux';
+import { connect } from 'react-redux';
 import isAdmin from '../lib/is-admin';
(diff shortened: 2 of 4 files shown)


eatyourgreens:
Gold standard mode gets turned off each time you start a new classification. Is that the correct behaviour?

wgranger:
I think this code looks sound and I can confirm it fixes the bug experienced. I find it a bit odd that gold standard mode would be turned off after each classification, but maybe there is a reason for this.

wgranger:
It looks like gold standard mode should reset after each classification so users don't forget they are in GS and make a ton of classifications https://github.com/zooniverse/Panoptes-Front-End/pull/1670#issuecomment-144447375

yshish:
> I think this code looks sound and I can confirm it fixes the bug experienced. I find it a bit odd that gold standard mode would be turned off after each classification, but maybe there is a reason for this.

Yes, doing the gold standards is pretty annoying, especially when I use it for most of my classification. as well as the need to first click on the 'cog button' to see a pop-up with the check-box. I did suggested some changes about that for the new design to Becky about two days ago.

Here is what I suggested her:

1. Make possible to have the Gold standard mode set ON for all classification in a row with an easy way to set it OFF for the few cases when we aren't confident enough to make it gold standard.

2. Make the process of setting the Gold standard mode easier by just a single click. For example, you could add the check-box just next to the Done button (without the necessity to click on the cog button first to see the pop-up).

3. Or you could make more 'Done' button choices: one regular and one with the 'star' symbol for Gold standard on it like we have somewhere 'Done' and 'Done&Talk' option next to each other. But that would mean having 4 Done buttons in those projects where you have also Done&Talk. (But that would be probably to much to have: "Done" and "Done&Talk" and "Done*" and "Done*&Talk" all at once).

(Or whatever you can think of that would make the gold standard classifications less painful and less time consuming would be very welcome.)

Any chance some of that could be realized? Thanks, guys!

wgranger:
I'm going to open a new issue mentioning this where we can have further discussion about the UX for GS mode.

PR #4999, opened 2018-10-17
Convert freehand shapes

wgranger:
Staging branch URL:
https://convert-freehand-shapes.pfe-preview.zooniverse.org/

Describe your changes.
This PR converts two freehand shape tools to ES6 and adds tests for both.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Something weird has happened with the package lock file here. The diff has a bunch of changes (846 lines?) but I think they're relative to a much older version. Some packages are changing version number in the diff, for example.

wgranger:
Ah right. I noticed something weird was happening with the package lock. I was trying to omit any changes to the package.lock with the most recent commit, but I don't think that worked correctly. I don't think this package should make any changes to the lock file.

app/classifier/drawing-tools/freehand-segment-shape.jsx
```diff
@@ -0,0 +1,250 @@
...
+      _currentlyDrawing: false
+    };
+  }
+
+  static initStart(coords, mark) {
+    mark.points.push(roundCoords(coords));
+    mark._inProgress = true;
+    mark._currentlyDrawing = true;
+    return mark;
+  }
```
eatyourgreens: I'm not sure, but I think this should return a new object, rather than mutating the mark directly. See the Fan tool, for example. https://github.com/zooniverse/Panoptes-Front-End/blob/07e975ef1f7a3918c7a966e3659974ea89d97327/app/classifier/drawing-tools/components/Fan/fan.jsx#L24-L26

app/classifier/drawing-tools/freehand-segment-shape.jsx
```diff
@@ -0,0 +1,250 @@
...
+
+  static initMove(coords, mark) {
+    mark.points.push(roundCoords(coords));
+    return mark;
+  }
```
eatyourgreens: Same comment here about returning a new object instead of mutating the mark directly.

eatyourgreens:
Probably best just leave the package lock file out of this PR, since it shouldn't change at all.

The drawing tools API isn't explained very clearly in the PFE Readme, but looking through the drawing task code, functions like `initStart` don't return a modified mark but instead should return a set of changes that will be applied to the current mark: https://github.com/zooniverse/Panoptes-Front-End/blob/88efefcc6ecfe276d0b8a146c0bf2cacb3721034/app/classifier/tasks/drawing/marking-initializer.cjsx#L64-L67

app/classifier/drawing-tools/freehand-segment-shape.jsx
wgranger: Ahh, this makes total sense. Thanks for pointing this out. Was a bit confused initially at the best way of returning the mark.

eatyourgreens: The Readme isn't very clear. I only figured it out by reading the drawing task code to see how these functions are used.

wgranger:
This should be good for a final review. I didn't refer to `FreehandShapeTool.defaultValues` in the `initStart` function because `mark` already possesses the default values at that point. However, I'm happy to make any changes if recommended.

PR #5011, opened 2018-10-23
Convert Draggable to ES6

eatyourgreens:
Staging branch URL: https://draggable.pfe-preview.zooniverse.org

Rewrites the Draggable decorator (used in the pan-zoom tools and all the drawing tools) as an ES6 functional component.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Drawing bugs where marks are accidentally placed outside the subject could be fixed if the Draggable component allowed you to restrict events to its children, rather than listening for click and drag across the entire document body.

eatyourgreens:
Selecting a previous mark is broken on this branch, and I can't see why.

app/lib/draggable.jsx
```diff
@@ -0,0 +1,102 @@
...
+        [moveEvent, endEvent] = ['mousemove', 'mouseup'];
+        break;
+      case 'touchstart':
+        [moveEvent, endEvent] = ['touchmove', 'touchend'];
+        break;
+      default:
+        [moveEvent, endEvent] = ['pointermove', 'pointerup'];
+    }
+
+    e = (e.touches && e.touches[0]) ? e.touches[0] : e;
```
eatyourgreens: Reassigning `e` here, to get the event (x,y) coords, is confusing later on. I keep thinking that the callback is called with the original event, but it isn't really.

app/lib/draggable.jsx
```diff
@@ -0,0 +1,102 @@
...
+
+    // If there's no `onStart`, `onDrag` will be called on start.
+    const startHandler = props.onStart || handleDrag;
+    if (startHandler) { // You can set it to `false` if you don't want anything to fire.
+      startHandler(e);
+    }
+  }
+
+  function handleDrag(e) {
+    e = (e.touches && e.touches[0]) ? e.touches[0] : e;
```
eatyourgreens: Same comment about reassigning `e` here.

app/lib/draggable.jsx
```diff
@@ -0,0 +1,102 @@
...
+    const d = {
+      x: e.pageX - _previousEventCoords.x,
+      y: e.pageY - _previousEventCoords.y
+    };
+    props.onDrag(e, d);
+    _rememberCoords(e);
+  }
+
+  function handleEnd(e) {
+    e = (e.touches && e.touches[0]) ? e.touches[0] : e;
```
eatyourgreens: Same comment about reassigning `e` here.

eatyourgreens:
Tested by zooming and drawing on the dev classifier in IE11, Safari 11, Firefox 62 and Chrome 70.

app/lib/draggable.spec.js
```diff
@@ -0,0 +1,256 @@
...
+      <Draggable
+        onStart={onStart}
+        onDrag={onDrag}
+        onEnd={onEnd}
+      >
+        <p>Hello</p>
+      </Draggable>
+    );
+  });
+  describe('on mousedown', function () {
```
eatyourgreens: These tests get repeated three times for mouse, touch and pointer. They could be rolled up into a function that's called three times with different event types.

shaunanoordin:
@eatyourgreens I think the next thing we'd need to do to figure out the Android/iOS drawing-dragging problem is to create a test-variant of the classifier that prints out (i.e. visible on a touch device) every touch/mouse/pointer event. I'd really like to know what the Classifer _thinks_ the user is doing when the user touches the screen.

I can sorta understand why the "touch-drag" action might simultaneously trigger the "move marker" and "pan window" events, but the insane lag and the multiple markers appearing from a simple touch-drag action (at least on Android, it seems the Milky Way Project user didn't see such on iOS) there's another layer to this madness.

shaunanoordin:
Anyway, this PR is good, but perhaps we'll only merge on Monday so we don't test our luck with the Friday Deploy Gods?

eatyourgreens:
@shaunanoordin maybe `preventDefault()` isn't working the way we think it should on Android, and `handleStart` is firing multiple drag start interactions inside nested Draggable elements? For instance, each point is draggable but the entire drawing SVG is also a draggable element (I think.)

I'm hoping this rewrite makes it easier to add logging and trace what's going on, though.

I've also noticed how insanely slow drawing can be. On my iPhone, there's sometimes a lag of several seconds between a touch move and a drawn mark actually moving.

PR #5034, opened 2018-10-31
allow admins to find users by email address

camallen:
add user-search-admin component with some slightly different filtering / lookup behaviour vs the normal user-search component

I had to relax the option filtering to handle the value being an email address vs the option label, https://github.com/JedWatson/react-select/tree/v1.0.0-rc.10#advanced-filters

I'm not super happy with this implementation but i've sunk a few hours into it already. Ideally i'd have some shared behaviour that i can override in each component (user-search vs user-search-admin) but i haven't got a decent idea to do that. thoughts / directions / suggestions appreciated. 

Staging branch URL:

Fixes the ability to find users by their emails for admin users - quite a common issue when dealing with unsubscription requests.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/components/user-search-admin.cjsx
+++ b/app/components/user-search-admin.cjsx
@@ -1,0 +1,76 @@
+React = require 'react'
+PropTypes = require 'prop-types'
+createReactClass = require 'create-react-class'
+Select = require('react-select').default
+apiClient = require 'panoptes-client/lib/api-client'
+
+delayBy = (timeout, fn) ->
+  setTimeout fn, timeout
+
+module.exports = createReactClass
+  displayName: 'UserSearchAdmin'
+
+  queryTimeout: NaN
+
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Looks good to me. If you wanted to render a UserSearch component, but decorate it with new behaviour, then you could wrap UserSearch in a [higher order component](https://reactjs.org/docs/higher-order-components.html) that adds the new search without duplicating code.

You could also import UserSearch into UserSearchAdmin, then render `<UserSearch />` from the admin component, with new functionality added by overriding props. For example, override the callback that's invoked when you type into the box and pass it to UserSearch as a prop. Then UserSearch Admin only needs to define the callback function and a render method.

PR #5035, opened 2018-10-31
Remove preferences change listener from workflow selection

eatyourgreens:
Staging branch URL: https://pr-5035.pfe-preview.zooniverse.org

Removes `listenToProjectPreferences` from the top-level project page component and the `onChangePreferences` callback from its children. Instead, the project workflow buttons are connected to the redux store and set a new workflow when clicked.

Selecting a new workflow clears the current workflow, loads the new workflow from Panoptes, then stores it.

User preferences are still updated on workflow change, but only so as to persist workflow selection across browser sessions. Updating user project preferences does not trigger a re-render of the project components and the fake preferences for anonymous users can be completely removed too.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/home/home-workflow-button.jsx
+++ b/app/pages/project/home/home-workflow-button.jsx
@@ -3,8 +3,12 @@ import React from 'react';
 import { Link } from 'react-router';
 import classnames from 'classnames';
 import Translate from 'react-translate-component';
+import apiClient from 'panoptes-client/lib/api-client';
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import * as classifierActions from '../../../redux/ducks/classify';
 
-export default class ProjectHomeWorkflowButton extends React.Component {
+class ProjectHomeWorkflowButton extends React.Component {
   constructor(props) {
     super(props);
 
@@ -12,10 +16,20 @@ export default class ProjectHomeWorkflowButton extends React.Component {
   }
 
   handleWorkflowSelection(e) {
-    if (this.props.disabled) {
+    const { actions, disabled, preferences, user, workflow } = this.props;
+    if (disabled) {
       e.preventDefault();
     } else {
-      this.props.preferences.update({ 'preferences.selected_workflow': this.props.workflow.id });
+      actions.classifier.setWorkflow(null);
+      return apiClient
+        .type('workflows')
+        .get(workflow.id, {})
+        .then((newWorkflow) => {
+          if (user) {
+            preferences.update({ 'preferences.selected_workflow': newWorkflow.id }).save();
+          }
+          actions.classifier.setWorkflow(newWorkflow);
+        });
     }
   }
 
@@ -64,6 +78,11 @@ ProjectHomeWorkflowButton.defaultProps = {
 };
 
 ProjectHomeWorkflowButton.propTypes = {
+  actions: PropTypes.shape({
+    classifier: PropTypes.shape({
+      setWorkflow: PropTypes.func
+    })
+  }),
   disabled: PropTypes.bool,
   preferences: PropTypes.shape({
     update: PropTypes.func
@@ -80,3 +99,15 @@ ProjectHomeWorkflowButton.propTypes = {
   }).isRequired,
   workflowAssignment: PropTypes.bool
 };
+
+const mapStateToProps = state => ({});
+
+const mapDispatchToProps = dispatch => ({
+  actions: {
+    classifier: bindActionCreators(classifierActions, dispatch)
+  }
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ProjectHomeWorkflowButton);
+export { ProjectHomeWorkflowButton };
+
--- a/app/pages/project/home/home-workflow-button.jsx
+++ b/app/pages/project/home/home-workflow-button.jsx
@@ -15,7 +15,7 @@ export default class ProjectHomeWorkflowButton extends React.Component {
     if (this.props.disabled) {
       e.preventDefault();
     } else {
-      this.props.onChangePreferences('preferences.selected_workflow', this.props.workflow.id);
+      this.props.preferences.update({ 'preferences.selected_workflow': this.props.workflow.id });
     }
   }
 
@@ -57,7 +57,7 @@ export default class ProjectHomeWorkflowButton extends React.Component {
 
 ProjectHomeWorkflowButton.defaultProps = {
   disabled: false,
-  onChangePreferences: () => {},
+  preferences: {},
   project: {},
   workflow: {},
   workflowAssignment: false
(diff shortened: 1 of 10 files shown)


eatyourgreens:
TODO: look at why workflow assignment tests are failing.

eatyourgreens:
Clearing the current workflow before a change should fix some weird behaviour noticed by @camallen. If the classifier mounts while a new workflow is loading, it fetches subjects for the workflow that was set on project load, then fetches again for the new workflow once that API call resolves.

app/pages/project/home/home-workflow-button.jsx
```diff
@@ -77,4 +98,16 @@ ProjectHomeWorkflowButton.propTypes = {
...
+const mapStateToProps = state => ({});
+
+const mapDispatchToProps = dispatch => ({
+  actions: {
+    classifier: bindActionCreators(classifierActions, dispatch)
+  }
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ProjectHomeWorkflowButton);
+export { ProjectHomeWorkflowButton };
```
wgranger: What is the purpose of having two exports here? (`export default` on line 111 and `export` on line 112)

app/pages/project/index.jsx
```diff
@@ -159,18 +158,9 @@ class ProjectPageController extends React.Component {
             }));
         })
     :
-      Promise.resolve(apiClient.type('project_preferences').create({
-        id: 'GUEST_PREFERENCES_DO_NOT_SAVE',
-        links: {
-          project: project.id
-        },
-        preferences: {}}));
+      Promise.resolve(null);
```
wgranger: I'm a bit worried this might cause some conflicts elsewhere in code. Beforehand, even non-logged-in users were given a `preferences` object. Curious if existence checks will need to be made elsewhere in code.

wgranger: Possibly here? https://github.com/zooniverse/Panoptes-Front-End/blob/e99fa0f4df6b4270f7d984e985656653431426ab/app/pages/project/classify.jsx#L192

wgranger: Or here? https://github.com/zooniverse/Panoptes-Front-End/blob/fe0c87a8ef589ef6c662bbb350ed35f8b1b8e895/app/classifier/drawing-tools/grid.cjsx#L105

wgranger: https://github.com/zooniverse/Panoptes-Front-End/blob/fe0c87a8ef589ef6c662bbb350ed35f8b1b8e895/app/classifier/tasks/drawing/marking-initializer.cjsx#L85

app/pages/project/home/home-workflow-button.jsx
eatyourgreens: The unwrapped component is exported for tests.

app/pages/project/index.jsx
eatyourgreens: I think this is worth checking with @srallen but my understanding is that the fake user preferences added here were just there to persist the selected workflow.

app/pages/project/home/home-workflow-button.spec.js
```diff
@@ -55,9 +77,29 @@ describe('ProjectHomeWorkflowButton', function() {
...
+    .then(done, done);
+  });
+  it('should clear the current workflow', function (done) {
+    wrapper.instance().handleWorkflowSelection()
+    .then(function () {
+      assert.equal(actions.classifier.setWorkflow.calledWith(null), true);
+    })
+    .then(done, done);
+  });
+  it('should select a new workflow', function (done) {
```
eatyourgreens: This test, and the preceding test, should really test that the old workflow is cleared before the new workflow is set. I wasn't quite sure how to do that with sinon.

eatyourgreens: Fixed now, thanks to @rogerhutchings advice.

app/pages/project/index.jsx
srallen: There are likely other side effects, possibly also for tracking whether a user has seen the tutorial too. Removing this might cause the tutorial to pop up every time an anonymous user loads the classify page during a session.

srallen: Will is right that some existential checks will need to be added to certain lines of code too.

eatyourgreens: The tutorial here only loads once for me when I'm not logged in. https://pr-5035.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground/classify

Now that we've got the redux store, we should try not to use the API client for storage.

eatyourgreens: Drawing tools with details tasks are broken on that project though. 😞

eatyourgreens: My preferred solution would be not to rely on the existence of user preferences for anything that can be done by an anonymous user, so set a sensible default here or set sensible defaults in components that take a preferences prop.

eatyourgreens: Tutorials use a global variable to store session data. That could be handled by the store too, if only so we know where to look for it.
`window.tutorialsCompletedThisSession = completedThisSession;`

srallen: Ah yes, I agree that is the preferred solution and the route we'll take with the new classifier.

eatyourgreens: The marking initialiser should be fixed now. I'm not sure if the grid tool was ever used in a project. Can that code can go (but maybe in a separate PR.)

eatyourgreens: I've checked Gravity Spy on this branch. The workflow assignment popup only shows up if I'm logged in. 'Take me to the next level' updates my preferences, but doesn't refresh the workflow so I guess that needs to call `actions.classifier.setWorkflow` too. I probably need to add a new action that takes a workflow ID and fetches that workflow from Panoptes.

eatyourgreens: Ok I've added a quick, and hacky (sorry!), fix for workflow assignment that requests the new workflow from Panoptes, then updates the store.

I haven't touched the grid tool because I'm not sure what the best approach is there. It seems like,the way it's written at the moment, that tool is intended for logged-in users anyway since the mark annotation is using user preferences (if I'm reading the code correctly.)

wgranger: I believe the grid tool can be deleted in a separate PR. I'll open an issue about it now and bring it up in a channel.

eatyourgreens: If it's still needed, I'd suggest rewriting it to store data using redux.

eatyourgreens: The grid tool appears to work in the dev classifier even if I'm not logged in. Not logged in, I can create a grid, give it a name and use the saved grid on subsequent classifications in the dev classifier.

Not logged in, I see API errors because creating a new grid tries to save user preferences, so I'm guessing that grids are saved so that they don't have to be created again on the next visit to the site. However, save won't work with the fake preferences that used to be created here, so I don't think this PR changes the grid tool save feature for anonymous users.

Ideally, the grid tool should check whether there's a user before attempting to save to Panoptes.

eatyourgreens: Confirmed on https://master.pfe-preview.zooniverse.org/dev/classifier that the preferences save error exists on master, and isn't introduced by this change.

eatyourgreens:
Drawing tools and workflow assignment should be working on this branch now. Checking the grid tool against master, there doesn't seem to be any difference in how it behaves, so I've left it alone.

PR #5054, opened 2018-11-07
Fix freehand line static methods and tests

eatyourgreens:
Updates the generic drawing tool tests to expect initMove to return an object.
Applies changes to the mock mark before rendering the tool component.
Fixes the freehand line tools to conform with the behaviour expected by the marking initialiser in the drawing task.

Staging branch URL: https://test-drawing-tools.pfe-preview.zooniverse.org/

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/drawing-tools/freehand-helpers.coffee
+++ b/app/classifier/drawing-tools/freehand-helpers.coffee
@@ -2,6 +2,7 @@ round = require 'lodash/round'
 
 createPathFromCoords = (coordsArray) ->
   [firstCoord, otherCoords...] = coordsArray
+  return '' unless firstCoord?
   path = "M #{firstCoord.x},#{firstCoord.y} "
   path += "L #{x},#{y} " for {x, y} in otherCoords
   path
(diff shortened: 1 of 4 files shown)


wgranger:
LGTM. Thanks for these changes.

eatyourgreens:
@wgranger it just occurred to me that these expectations will never fail, because `initialMark` and `mark` are always different copies of the same object. Maybe they should be deleted too?
`expect(initialMark).to.not.equal(mark);`

wgranger:
Ah! Of course. I suppose in this case the best we can do is check that mark is an object, unless there's a way to check for deep vs shallow equality with the exception library.

eatyourgreens:
I think you can `expect(a).to.not.deep.equal(b)` with chai. It's pretty powerful.
https://www.chaijs.com/api/bdd/

eatyourgreens:
In that case, I think the drawing tools API expects you to do something like:
`const mark = Object.assign({}, initialMark, TaskComponent.initStart(props, mark))`
in order to get the new mark from the initial mark.

Returning a new mark from those static methods is something to consider if we rewrite that drawing tools API in PFE. It might just end up getting thrown out for the new classifier, though.

PR #5076, opened 2018-11-14
Convert Survey Components

wgranger:
Staging branch URL: https://convert-survey-components-two.pfe-preview.zooniverse.org/

Describe your changes.
This branch converts remaining survey components (aside from the massive editor) while adding tests for the index (bringing it to 100%!). 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/classifier/tasks/survey/index.jsx
```diff
@@ -0,0 +1,182 @@
...
+      filters: {},
+      focusedChoice: '',
+      selectedChoiceID: ''
+    };
+  }
+
+  handleFilter(characteristicID, valueID) {
+    const { filters } = this.state;
+
+    setTimeout(() => {
```
eatyourgreens: Does the survey task break if you remove this timeout? It's hard to see why it's here, and it might not be needed.

app/classifier/tasks/survey/index.jsx
```diff
@@ -0,0 +1,182 @@
...
+    });
+    const newAnnotation = Object.assign({}, annotation, { _choiceInProgress: false });
+    onChange(newAnnotation);
+  }
+
+  handleAnnotation(choice, answers, e) {
+    const { filters } = this.state;
+    const { annotation, onChange } = this.props;
+
+    const parsedFilters = JSON.parse(JSON.stringify(filters));
```
eatyourgreens: Copying the filters object by converting it to a string and back might be slow.

app/classifier/tasks/survey/index.spec.js
```diff
@@ -1,72 +1,162 @@
...
+    after(function () {
+      clock.restore();
+    });
+  });
+
+  describe('static methods', function () {
+    it('should correctly return isAnnotationComplete', function() {
+      const isComplete = SurveyTask.isAnnotationComplete(task, annotation);
+      expect(isComplete).to.be.a('boolean');
+      expect(isComplete).to.be.true;
```
eatyourgreens: Should there also be a test for the case where isAnnotationComplete is false?

app/classifier/tasks/survey/index.spec.js
```diff
@@ -1,72 +1,162 @@
...
+    });
+
+    it('should correctly return a default annotation', function() {
```
eatyourgreens: Default annotation should be an object, I think. Optionally, you can also check the returned object to see if it has the properties of a survey task annotation.

app/classifier/tasks/survey/index.spec.js
```diff
@@ -1,72 +1,162 @@
...
+      const defaultAnnotation = SurveyTask.getDefaultAnnotation();
+      expect(defaultAnnotation).to.be.an('object');
+      expect(defaultAnnotation.value.length, 0).to.be.equal;
+    });
+
+    it('should return the correct value from getTaskText', function() {
```
eatyourgreens: getTaskText should be a string.

app/classifier/tasks/survey/index.spec.js
```diff
@@ -1,72 +1,162 @@
...
+      const taskText = SurveyTask.getTaskText(task);
+      expect(taskText).to.be.a('string');
+    });
+
+    it('should correctly return a default task', function() {
```
eatyourgreens: Default task should be an object. Optionally, you can also test the properties of the task object to make sure they're correct.

app/classifier/tasks/survey/utility.js
```diff
@@ -0,0 +1,12 @@
...
+    return [];
+  }
+  if (!(task.questionsMap && Object.keys(task.questionsMap).indexOf(choiceID) >= 0)) {
+    return task.questionsOrder;
+  }
+  return task.questionsMap[choiceID];
+}
+
+export default getQuestionIDs;
+export { getQuestionIDs };
```
eatyourgreens: You don't need to export getQuestionIds twice here. I think `export default { getQuestionIDs }` should work.

app/classifier/tasks/survey/index.jsx
wgranger: Currently, when you select a filter, the modal asking which filter you want closes after making a selection. If `setTimeout` is removed, the modal doesn't close after making a selection, you then have to click off of the modal.

app/classifier/tasks/survey/index.jsx
wgranger: I'm guessing the alternative is `Object.assign()`? Although, I'm not sure why this wasn't initially used.

app/classifier/tasks/survey/index.spec.js
```diff
@@ -1,72 +1,162 @@
...
+
+    before(function () {
+      const date = new Date();
+      clock = sinon.useFakeTimers(date.getTime());
+    });
+
+    it('should correctly call handleAnnotation', function() {
+      wrapper.instance().handleAnnotation(selection.choice, selection.answers);
+      expect(onChangeSpy.calledOnce).to.be.true;
+      clock.tick();
```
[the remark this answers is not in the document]
wgranger: Exactly. Without `clock.tick()` the test cannot be accurately executed because the function and test both finish before the items inside `setTimeout` execute.

wgranger:
PR updated per your requests. I have a couple remaining questions based on your comments.

eatyourgreens:
Changes look good but I can't submit a classification on the test project: the Done button is always disabled. https://convert-survey-components-two.pfe-preview.zooniverse.org/projects/srallen086/survey-test/classify

app/classifier/tasks/survey/index.jsx
eatyourgreens: [`Object.assign()` won't make a new copy of a deeply nested object](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#Deep_Clone). This code is ok, just slightly puzzling.

eatyourgreens:
By the way, I can still see `survey/index.cjsx` on this branch.

wgranger:
Agreed. There should be a test for this. Good catch. I'm currently following the stack trace to see where, and why, that would be changed.

eatyourgreens:
I'm trying this out:
```javascript
describe('handleAnnotation', function () {
      it('should call onChange with an annotation', function() {
        wrapper.instance().handleAnnotation(selection.choice, selection.answers);
        clock.tick();
        expect(onChangeSpy.calledOnce).to.be.true;
        const returnValues = onChangeSpy.returnValues[0];
        expect(returnValues, annotation).to.be.equal;
      });
      it('should complete an incomplete annotation', function () {
        wrapper.setProps({ annotation: incompleteAnnotation });
        wrapper.instance().handleAnnotation(selection.choice, selection.answers);
        clock.tick();
        const newAnnotation = onChangeSpy.returnValues[0];
        expect(newAnnotation._choiceInProgress).to.be.false;
      })
    })
```

eatyourgreens:
I've pushed those tests to this branch. The onChange test is failing, which I can't figure out.

wgranger:
I think when a new annotation is created it immediately replaces the completed annotation at the same index, but I'm still trying to figure out how that happens.

eatyourgreens:
Tests pass for me locally, so this might have something to do with the `setTimeout` behaving differently in the tests than in the browser. In the browser, I'm seeing `annotation._choiceInProgress` is still true after making an identification. I think it gets set to false by `clearSelection()` then reset to true by `onChange` at the end of the `handleAnnotation` function.

wgranger:
I think what's happening is that when `newAnnotation` is declared [here](https://github.com/zooniverse/Panoptes-Front-End/pull/5076/files#diff-dd673cd659d6631c84d690ac0581d556R114) it is using the destructured prop assignment from earlier `const { annotation, onChange } = this.props;` before the new annotation has been reset and loaded through `clearSelection`. Changing that line to `const newAnnotation = Object.assign({}, this.props.annotation, { value });` seems to fix the problem.

eatyourgreens:
I can fix it by removing the setTimeout too. I'll push what I've got.

wgranger:
Nice. This looks good by my testing. I can merge if everything checks out with Travis, etc, unless there are any other comments that need addressing.

eatyourgreens:
Looks okay to me, from some quick classifying on the test survey project.

eatyourgreens:
Oh, the original coffeescript file still needs to be deleted.

eatyourgreens:
Should be good to merge once Travis and Jenkins have done their thing.

eatyourgreens:
I'm not sure why that one test is failing. It passes when I run the tests in isolation with `docker-compose run dev npm run test-one app/pages/project/workflow-selection.spec.js` but fails if I run the entire test suite.

eatyourgreens:
Wait I'm running the wrong tests. D'oh!

Issue #5095, opened 2018-11-27
Fix long link styling in subject metadata


mcbouslog:
## Expected behavior
Link to appear readable, within modal.

## Current behavior
https://www.zooniverse.org/projects/marckuchner/backyard-worlds-planet-9/talk/subjects/13772456
Link doesn't wrap.

eatyourgreens:
This has been requested before by NfN. We can’t control how browsers wrap URLs but I pointed out to Michael that the solution would be to add link text to their Markdown links in metadata. I’ll dig up the original PFE discussion.

eatyourgreens:
See https://github.com/zooniverse/Panoptes-Front-End/issues/4691#issuecomment-396664168

That discussion was for tutorials, but the problem is the same: long URL strings that the browser won't wrap.

mcbouslog:
per https://github.com/zooniverse/Panoptes-Front-End/issues/4691#issuecomment-396664168:
"I’d recommend adding link text, instead of using the bare URLs for the links: [link text](URL)."

mcbouslog:
Could adding the `word-break: break-all` styling be an option?

eatyourgreens:
It’s definitely worth trying. The exact wrapping might vary depending on which characters are in a URL.

eatyourgreens:
That branch looks good on my phone, by the way.

PR #5103, opened 2018-11-28
Add break-word styling to anchor

### LINKED ISSUE #4691: styling issue of the anchor tag within the markdown (wrapping text in tutorial)
## Expected behavior
Be able to wrap URL text to fit within tutorial window

## Current behavior

text does not get wrapped (see link below)

## Steps to replicate
see: 
https://www.zooniverse.org/projects/zooniverse/notes-from-nature/classify?reload=1&workflow=6014&_ga=2.184423526.1576714224.1528666194-301989501.1485808297

See the last page of the Tutorial there.

## Additional information
Some further discussion here: #4659 

- **Operating system:**
OS X 10.11.6

- **Browser:**
Chrome

mcbouslog:
Staging branch URL: 
- (metadata info) https://metadata-link-styling.pfe-preview.zooniverse.org/projects/marckuchner/backyard-worlds-planet-9/talk/subjects/13772456?env=production
- (last page of tutorial) https://metadata-link-styling.pfe-preview.zooniverse.org/projects/zooniverse/notes-from-nature/classify?reload=1&workflow=6014&env=production

Fixes #4691, #5095 

Adds `word-break: break-word` and `overflow-wrap: break-word` styling to anchors.
Would a new/specific class for the above linked instances be preferred? 
The anchors are nested with `markdownz`.
I navigated around the site and anchors appear to look normal otherwise, but certainly could be missing something.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/css/typography.styl
+++ b/css/typography.styl
@@ -26,6 +26,8 @@ hr
 
 a
   color: MAIN_HIGHLIGHT
+  overflow-wrap: break-word
+  word-break: break-word
 
 .emoji
   max-height: 1em


eatyourgreens:
I think NfN should be reminded that it's not good to have bare URLs embedded in tutorials, like the second example above. Links should be easy to read!

eatyourgreens:
With this styling, links are being wrapped mid-word on the FAQ page, which is not great.
https://metadata-link-styling.pfe-preview.zooniverse.org/about/faq

mcbouslog:
Nice catch! Yes that looks bad. I changed it to `word-break: break-word` and I think it looks ok, better?
screen shot 2018-11-28 at 10 49 50 am

eatyourgreens:
Sounds good but [probably won't work in Firefox or IE](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#Browser_compatibility).

eatyourgreens:
Looks like [`overflow-wrap: break-word`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap) has better browser compatibility.

mcbouslog:
`overflow-wrap` isn't working for me, but I think the issue may lie with `[standard-table](https://github.com/zooniverse/Panoptes-Front-End/blob/master/css/tables.styl#L1)`, or at least something to do with the table width. I can limit the width when I set a specific pixel on the subsequent `td`, but attempts with %'s or similar don't limit the width.

mcbouslog:
It doesn't make sense to me per the MDN compatibility docs ([`overflow-wrap`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Browser_compatibility) and [`word-break`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break#Browser_compatibility)), but `overflow-wrap` doesn't get the desired result for Chrome or Safari while `word-break` does. Firefox works as desired with `overflow-wrap`. I couldn't get anything to work with IE or Edge (even using the old `word-wrap` as noted above and in https://caniuse.com/#feat=wordwrap), though it looks like Edge 18 should work.

Is it ok to cover what we can with `overflow-wrap` and `word-break`? I don't like it, but maybe it's ok.

eatyourgreens:
One of the nice things about CSS is that parsers skip past declarations they don't understand, without breaking, so you could include both rules in the stylesheet, in an order that catches most browsers.

Each browser will have its own implementation of word wrapping at the end of lines, so I don't think you can expect the same results in every browser anyway.

PR #5107, opened 2018-11-29
Improve workflow selection for anonymous users

eatyourgreens:
Staging branch URL: https://pr-5107.pfe-preview.zooniverse.org

- Changes the project home page workflow links so that they defer loading the classify page until the new workflow has loaded.
- Changes workflow selection so that a new workflow is not loaded if there is already an active workflow for the project on mount.
- Updates components that load workflows so that translations are loaded in parallel with workflows, rather than after workflows load.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -12,12 +12,17 @@ class WorkflowSelection extends React.Component {
     super();
     this.state = {
       error: null,
-      loadingSelectedWorkflow: true
+      loadingSelectedWorkflow: false
     };
   }
 
   componentDidMount() {
-    this.getSelectedWorkflow(this.props);
+    const { project, workflow } = this.props;
+    const linkedActiveWorkflows = project.links.active_workflows || [];
+    const workflowExistsForProject = workflow && linkedActiveWorkflows.indexOf(workflow.id) > -1;
+    if (!workflowExistsForProject) {
+      this.getSelectedWorkflow(this.props);
+    }
   }
 
   componentDidUpdate(prevProps, prevState) {
@@ -87,6 +92,7 @@ class WorkflowSelection extends React.Component {
     }
     let awaitWorkflow;
     if (isValidWorkflow) {
+      actions.translations.load('workflow', sanitisedWorkflowID, locale);
       awaitWorkflow = apiClient
         .type('workflows')
         .get(sanitisedWorkflowID, {}) // the empty query here forces the client to bypass its internal cache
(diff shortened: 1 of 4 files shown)


app/pages/project/workflow-selection.jsx
```diff
@@ -12,12 +12,17 @@ class WorkflowSelection extends React.Component {
     super();
     this.state = {
       error: null,
-      loadingSelectedWorkflow: true
+      loadingSelectedWorkflow: false
```
wgranger: Why was this changed to false? It seems with this change the component would load on initial render https://github.com/zooniverse/Panoptes-Front-End/blob/97f8bc24ba926fa24ab6756b82727cd0d8efbd16/app/pages/project/workflow-selection.jsx#L190

eatyourgreens: If you set this to true, you should notice that the classifier never loads.

wgranger: Ahh. You are correct.

eatyourgreens: If there’s already a valid workflow loaded, we set this to false and skip workflow loading completely.

eatyourgreens: That could be set explicitly in `componentDidMount` if that makes the code easier to understand. I’ll leave it to you to change if you think that’s better.

app/pages/project/workflow-selection.jsx
```diff
@@ -12,12 +12,17 @@ class WorkflowSelection extends React.Component {
...
     };
   }
 
   componentDidMount() {
-    this.getSelectedWorkflow(this.props);
+    const { project, workflow } = this.props;
+    const linkedActiveWorkflows = project.links.active_workflows || [];
+    const workflowExistsForProject = workflow && linkedActiveWorkflows.indexOf(workflow.id) > -1;
+    if (!workflowExistsForProject) {
+      this.getSelectedWorkflow(this.props);
```
wgranger: Ok, so this only gets called if it needs to (if there isn't currently a workflow present? And this is what eliminates the second workflow request?

app/pages/project/workflow-selection.jsx
wgranger: No, I think it's unnecessary to have an additional setState in `componentDidMount` if this is the property you want.

wgranger:
I've tested this PR on a majority of the active projects, signed in and not, and they all appear to handle workflow selection correctly.

app/pages/project/workflow-selection.jsx
eatyourgreens: Yes. Without this change, a workflow is always loaded on component mount, even if there’s already one in the Redux store.

Issue #5125, opened 2018-12-05
[RFC] Removing ambiguity from Workflow selection


shaunanoordin:
## Request for Comments: how to remove ambiguity in PFE Classifier's Workflow selection

This RFC follows the Oxford team's discussion today where we delved into the issue of the _ambiguous Workflow selection_ (see #5111) and proposed some solutions to fix it. what we want now is to share those plans and to seek feedback from the whole Zooniverse team, as any solutions to this Workflow ambiguity problem will affect projects across the whole platform.

**Problem:**
As described in #5111, the PFE Classifier has an issue where, when a Workflow is selected on a project with multiple Workflows, we can't be too sure _which_ workflow resources the user _sees._ This leads to problematic conditions where, for example, a User selects Workflow A, sees Subjects from Workflow A, but sees the translated questions from Workflow B.

**Proposed Solution**
We need to be **extremely clear** which Workflow is currently being worked by the PFE Classifier.

There are two parts to this solution:
1. when fetching a Workflow, we must get the ID **explicitly** from the URL - e.g. `https://www.zooniverse.org/projects/drewdeepsouth/southern-weather-discovery/classify?workflow=6442` - and not "guess" which Workflow the user wanted to work on based on click events or etc.
2. the Classification needs to record which resources were seen by the user when they were classifying - this usually takes the form of an additional `classification.links.translation_resource = [123]` or similar.

Solution Part 1 should eliminate any sort of ambiguity in workflow selection, while Solution Part 2 is a logical followthrough which allows us to double-check that users classifying Subjects from Workflow X were indeed looking at (translated) questions from Workflow X.

### Additional Notes

- If we're enforcing that the PFE Classifier needs to be able to explicitly identify the workflow it's supposed to work on via URL or query params, we also need to consider what happens when none is provided - e.g. if a user visits `https://www.zooniverse.org/projects/drewdeepsouth/southern-weather-discovery/classify`, do we _redirect_ to the **default workflow URL** (with workflow ID in the query param)? What if there's no default workflow?
- Additional problems solved by having a specific Workflow ID in the URL include:
  - user bookmarks consistently taking them to the same workflow
  - users opening a specific workflow _via "right-click->open in new tab"_ on a Project home page will now open the correct workflow.
- When storing Translation resources, we must store them by _Workflow ID._ The Workflow ID adds an extra level of identity when fetching Translations, and gets around the problem where two workflows might share the same Task ID (e.g. `"T0"`)

[...]

PR #5127, opened 2018-12-05
Add workflow translation id to classification metadata

mcbouslog:
Staging branch URL: https://pr-5127.pfe-preview.zooniverse.org/
- helpful test project: https://pr-5127.pfe-preview.zooniverse.org/projects/eatyourgreens/-project-testing-ground

Towards #5125.

- adds translation resource id to corresponding translations store strings type
- saves `workflow_translation_id` and `user_language` to classification metadata on completion
- renames `updateClassification` to `updateMetadata` to better reflect actual action

Questions
- [ ] **would this record the necessary info so we could search classifications and check for mismatched workflow translations?**
- [x] preferred default `null`? could be empty string or something else. some of my default/initial settings might not be necessary.

Apologies for the excessive commits, if merging please squash.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Should the translation ID be added to metadata or links? All the other resource IDs are recorded in `classification.links`?

What's your thinking behind adding a new action creator, rather than adding the translation ID as an extra argument to either classification creation or classification completion? My thinking would have been to add it to the creation function, along with the project, workflow and subject but I could be missing something.

app/pages/project/classify.jsx
```diff
@@ -82,7 +86,11 @@ export class ProjectClassifyPage extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { classification, upcomingSubjects, workflow } = this.props;
+    const { actions, classification, translations, upcomingSubjects, workflow } = this.props;
+
+    if (translations !== prevProps.translations) {
```
eatyourgreens: `translations` is a nested object, so I'm not sure that you can use direct comparison here to detect changes.

app/redux/ducks/classify.js
```diff
@@ -32,11 +32,12 @@ function awaitSubjectSet(workflow) {
   }
 }
 
-function createNewClassification(project, workflow, subject, goldStandardMode) {
+function createNewClassification(project, workflow, subject, goldStandardMode, workflowTranslationId) {
   const source = subject.metadata.intervention ? 'sugar' : 'api';
   const classification = apiClient.type('classifications').create({
     annotations: [],
     metadata: {
+      workflow_translation_id: workflowTranslationId,
```
eatyourgreens: Is this here because it won't work in the links object?

eatyourgreens:
I'm seeing the workflow translation action fire before the classification is created, which I think is a bug.
Redux dev tools showing the order in which the classifier actions fire.

If I change the project language from English to Italian, I see the workflow translation action fire multiple times, with a null payload.
Redux dev tools showing the workflow translation action firing multiple times on language change.

eatyourgreens:
I'm seeing the workflow translation action fire before the classification is created, which I think is a bug.
Redux dev tools showing the order in which the classifier actions fire.

If I change the project language from English to Italian, I see the workflow translation action fire multiple times, with a null payload.
Redux dev tools showing the workflow translation action firing multiple times on language change.

eatyourgreens:
I'm seeing the workflow translation action fire before the classification is created, which I think is a bug.
Redux dev tools showing the order in which the classifier actions fire.

If I change the project language from English to Italian, I see the workflow translation action fire multiple times, with a null payload.
Redux dev tools showing the workflow translation action firing multiple times on language change.

eatyourgreens:
It would help if I used a workflow that actually has an Italian version, right?

If I change the language, the translation ID changes but I don't see the classification change. Here, the classification is recording the English language translation.
Redux dev tools showing classifier state after a language change.

eatyourgreens:
Since I can change language during the course of classifying, it probably makes more sense to record the language and translation ID at the end of the classification, not the start.

eatyourgreens:
With the latest commit, do we still need the workflow translation action?

mcbouslog:
I think the action can be removed if saving on completion.

This is likely overkill, not necessary, but do we want to save the workflow translation id at creation and then separately at completion? I don't think so, but thought I'd ask.

mcbouslog:
I think the action can be removed if saving on completion.

This is likely overkill, not necessary, but do we want to save the workflow translation id at creation and then separately at completion? I don't think so, but thought I'd ask.

mcbouslog:
Originally went the link route, but per a brief Slack conversation with @srallen and @camallen we thought we'd go with saving it in the metadata.

Once this is in good shape it might be worth having @marten or @camallen confirm this will provide the necessary info so we can search all classifications for mismatches.

eatyourgreens:
I think quite a lot of code can go,  in favour of using `actions.classify.updateClassification()` to change classification metadata directly from `props.translations`.

eatyourgreens:
Incidentally, I noticed while working on the classifier that we only ever update classification metadata, which is why that function doesn't update any other classification properties.

mcbouslog:
I'll remove some code and update to store an array of translation ids.

eatyourgreens:
Maybe that `updateClassification` function would be better named `updateMetdata`? That would be consistent with the naming of the `updateAnnotations` function.

app/pages/project/classify.jsx
srallen: lodash has a deep equality function: https://lodash.com/docs/4.17.11#isEqual

eatyourgreens: In other components, I've listened for locale changes in order to detect when the language changes.
https://github.com/zooniverse/Panoptes-Front-End/blob/eb1a1556d22405adea855fbac1bd6c58d9ec2fac/app/pages/project/workflow-selection.jsx#L39-L41

eatyourgreens: Also note that you could dispatch an `updateClassification` action from within the translations ducks file, whenever a new workflow translation loads. Interventions does this in order to update the classifier store from Sugar messages:
https://github.com/zooniverse/Panoptes-Front-End/blob/eb1a1556d22405adea855fbac1bd6c58d9ec2fac/app/redux/ducks/interventions.js#L3

mcbouslog:
Ok I think this is looking better now. I like the array of workflow translation ids, captures user behavior better.

I'm indifferent between the current approach where on classifier translations prop change it updates the metadata vs updating the metadata from within the translations duck. I kind of like keeping the translations duck without importing other actions or checking the translation type then doing something unique, but I don't feel strongly about that at all, not really strong technical arguments.

mcbouslog:
While a classification metadata property `workflow_translation_ids` as an array of null or translation resource ids captures user behavior well, I worry it might make searching classifications for mismatched workflows and translations significantly harder or impossible?

If so, and if the immediate goal is to track if the workflow-translation bug is still occuring, we might want to just save the `initial_workflow_translation_id` as a string to metadata? As I think that initial classifier load with the mismatch is what we want to capture in metadata.

Roping in @camallen or @marten for their thoughts on this point.

eatyourgreens:
I’m not sure about capturing an array of translations either, since we only record the user language once, at the end of the classification.

camallen:
>While a classification metadata property workflow_translation_ids as an array of null or translation resource ids captures user behavior well, I worry it might make searching classifications for mismatched workflows and translations significantly harder or impossible?
>If so, and if the immediate goal is to track if the workflow-translation bug is still occuring, we might want to just save the initial_workflow_translation_id as a string to metadata? As I think that initial classifier load with the mismatch is what we want to capture in metadata.

Is it an array of translation ids to capture the different language strings? Correct me where i'm not getting it. 

As it stand in code now, I assume for say an default language project (let's say english from the lab) it'll record just eng language translation id. If however i change languages, do we want to record this change and track which strings were shown? 

If so then instead of an array of ids, I think i'd prefer the metadata object to be an object keyed by language code for review in exports, etc `{ 'en-au': 12345, 'en-gb': 12354 }` i think that'll help get a sense of what the user saw without having to lookup the translation record manually (and help us track the state of the classifier as it changes, thus what the user actually saw).

If you just want to track the translation id for the submitted language strings, then why does it need to be in an array? just plonk it into a `{ 'workflow_translation_id': 12345 }`  attribute in the metadata object.

mcbouslog:
Per Slack conversation we've decided to save the workflow translation id on completion.

No projects identified as possibly affected by the workflow-translation bug have multiple translations. Therefore, we believe it's not necessary to save multiple translation ids to the classification metadata.

app/pages/project/classify.jsx
```diff
@@ -182,7 +184,7 @@ export class ProjectClassifyPage extends React.Component {
   loadAnotherSubject() {
     const { actions, project, workflow } = this.props;
     if (workflow) {
-      actions.classifier.nextSubject(project, workflow);
+      actions.classifier.nextSubject(project);
```
eatyourgreens: Out of interest, why was the workflow removed here?

mcbouslog: Per https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/redux/ducks/classify.js#L315-L320 I don't think the workflow param is used? The classify duck tracks workflow in it's state (https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/redux/ducks/classify.js#L164)?

mcbouslog: This also applies to line 123.

app/classifier/classifier.jsx
```diff
@@ -253,15 +253,17 @@ class Classifier extends React.Component {
...
         width: innerWidth,
         height: innerHeight
       },
       interventions: {
         message: !!showIntervention,
         opt_in: !!user && user.intervention_notifications
-      }
+      },
+      workflow_translation_id: translations.strings.workflow.id,
+      user_language: translations.locale
```
eatyourgreens: Small point, but I think I forgot to add tests for this change.

mcbouslog: Just added.

app/redux/ducks/translations.js
```diff
@@ -40,7 +40,7 @@ const initialState = {
   rtl: false,
   strings: {
     project: {},
-    workflow: {},
+    workflow: { id: null },
```
wgranger: Why was this `id: null` added?

mcbouslog: If there's no translation for the workflow I wanted to provide a default or null value so when on classification completion `translations.strings.workflow.id` is saved we'll know there wasn't a translation, as opposed to just not saving anything and wondering if it worked at all. So I added that default id of null in the duck as well as the classifier component.

eatyourgreens: Does the API client record a value if you pass `null`? I thought it removed the key in that case. I could be wrong though.

eatyourgreens:
Tests are passing. 👍

app/redux/ducks/translations.js
wgranger: Looks like if no id is present, `workflow_translation_id: null` is recorded in the metadata.

mcbouslog: I just did a few classifications on a testing project without any workflow translations and ran a data export and I think we're ok:
screen shot 2018-12-07 at 11 00 45 am

eatyourgreens: Cool. Sounds like we didn't roll out translations on staging though. All projects should have them at this point, even if it's just the one English translation that you get by default.

PR #5140, opened 2018-12-11
Add simple URL checks to the external URL editor

eatyourgreens:
Make both text inputs required.
Use input[type=url] for URLs and add simple pattern matching for http[s]:// URLs.
Add some string validation on change for both inputs.

Staging branch URL: https://pr-5140.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/external-links-editor.jsx
+++ b/app/pages/lab/external-links-editor.jsx
@@ -64,6 +64,7 @@ export default class ExternalLinksEditor extends React.Component {
           <input
             type="text"
             name={`urls.${idx}.label`}
+            required
             value={link.label}
             onChange={handleInputChange.bind(this.props.project)}
             onMouseDown={this.handleDisableDrag}
@@ -72,8 +73,10 @@ export default class ExternalLinksEditor extends React.Component {
         </AutoSave>
         <AutoSave tag="td" resource={this.props.project}>
           <input
-            type="text"
+            type="url"
             name={`urls.${idx}.url`}
+            pattern="https?://.+"
+            required
             value={link.url}
             onChange={handleInputChange.bind(this.props.project)}
             onMouseDown={this.handleDisableDrag}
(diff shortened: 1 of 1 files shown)


rogerhutchings:
In Chrome, this doesn't appear to work:

screenshot 2018-12-11 16 04 08

eatyourgreens:
How odd. Browser compatibility includes Chrome. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/url#Browser_compatibility

eatyourgreens:
Probably not working because those forms aren't submitted. They're saved using the `AutoSave` component.

rogerhutchings:
Okay - if that's the case, we could add a `validateFn` prop to `AutoSave` which accepts a function and passes the value as an argument

rogerhutchings:
I'm happy to write the validator and merge it into this branch

eatyourgreens:
I've added really simple validation on change.

rogerhutchings:
This still autosaves on blur :(

eatyourgreens:
Can you elaborate? It should save on blur, but the URL value should be an empty string.

rogerhutchings:
Sure, sorry - if I create a new link, enter `Foo` for the label, and `Bar` as the value, the autosave kicks off. If I then refresh, the link with `Foo` and `Bar` is present.

eatyourgreens:
That's fine though. 'Bar' isn't an XSS URL.

rogerhutchings:
Ah, I misunderstood how it was going to work then; I thought it was going to strip out anything not starting with `http`

eatyourgreens:
Problem is, you can't type links at all if it deletes anything that starts with `ht` or `htt`.

A better validator would hold off on saving if the URL is too short, but not actually delete it from the text box.

PR #5171, opened 2018-12-20
Validate project assigned workflows

eatyourgreens:
Check that project assigned workflows are active before showing the workflow assignment popup. If they are not active, clear `preferences.settings.workflow_id`. Update workflow assignment to use `componentDidUpdate` rather than `componentWillReceiveProps`.

Staging branch URL: https://pr-5171.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -149,14 +149,21 @@ export class ProjectClassifyPage extends React.Component {
 
   shouldWorkflowAssignmentPrompt(nextProps) {
     // Only for Gravity Spy which is assigning workflows to logged in users
+    const { preferences, project, workflow } = this.props;
     if (nextProps.project.experimental_tools.indexOf('workflow assignment') > -1) {
       const assignedWorkflowID = nextProps.preferences &&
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Interventions could be used to decide whether or not to show a prompt, depending on whether or not there's a valid workflow ID set in preferences. I don't think it would be difficult to add an intervention action that responds to preference changes, rather than Sugar messages. We could also use Sugar for workflow assignment, similar to how it's used to assign subjects to volunteers.
https://github.com/zooniverse/Panoptes-Front-End/blob/45ad6c8e0eca74da5eb927bf365f70b10b6addad/app/redux/ducks/interventions.js#L102-L111

srallen:
Those are all good ideas, but I'd consider that to be new functionality for workflow selection and should go into the rewrite. Checking user roles to see if they can load an inactive workflow is something the old workflow selection code did and it would've happened in this case since the code was in a single place. My comment is about getting back to parity with previous functionality.

PR #5186, opened 2019-01-04
Reset the classifier on auth change

### LINKED ISSUE #5184: Workflow should be re-requested on sign out if the workflow should only be accessed by certain users
## Expected behavior
For projects that use workflow assignment, if a user signs out, then the workflow needs to be re-requested to fallback to the project default workflow, which typically should be "level 1". Additionally, if a workflow is not active and but it's being used by a project owner, collab, tester, or admin for testing or for submitting gold standard classifications and then the user signs out, then the workflow should be re-requested to load a workflow that an anonymous user has permission to load.

## Current behavior
The classifier state remains on the previously loaded workflow when signing out.

Discovered while using my workflow assignment testing project on staging for PR #5171: https://pr-5171.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing/classify

If I'm assigned to workflow 2334 (level 2) and have that workflow loaded in the classifier and then sign out, the classifier remains on workflow 2334 rather than switch to workflow 2333 (default workflow/level 1)

## Steps to replicate
- Load this project: https://pr-5171.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing
- Make sure your user is assigned to workflow `2334` in your UPP `settings.workflow_id` (can use console to set it)
- Load classifier with workflow `2334` aka the journey workflow.
- Sign out.

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Reset the classifier store to its initial state on log in or log out. This should clear the stored workflow, empty the subject queue and destroy any incomplete classifications.

Key workflow selection by workflow ID, so that workflow selection mounts and runs again when the stored workflow is reset to null.

Staging branch URL: https://pr-5186.pfe-preview.zooniverse.org

Fixes #5184.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -12,9 +12,9 @@ style = require '../css/main.styl'
 `import { Provider } from 'react-redux';`
 `import configureStore from './redux/store';`
 `import { processIntervention } from './redux/ducks/interventions';`
-`import { emptySubjectQueue } from './redux/ducks/classify';`
+`import { reset } from './redux/ducks/classify';`
 store = configureStore()
-auth.listen('change', () => store.dispatch(emptySubjectQueue()));
+auth.listen('change', () => store.dispatch(reset()));
 apiClient.type('subject_sets').listen('add-or-remove', () => store.dispatch(emptySubjectQueue()));
 sugarClient.on('experiment', (message) => store.dispatch(processIntervention(message)));
 
--- a/app/main.cjsx
+++ b/app/main.cjsx
@@ -1,7 +1,6 @@
 React = require 'react'
 ReactDOM = require 'react-dom'
 apiClient = require 'panoptes-client/lib/api-client'
-auth = require 'panoptes-client/lib/auth'
 { applyRouterMiddleware, Router, browserHistory } = require 'react-router'
 useScroll = require 'react-router-scroll/lib/useScroll'
 routes = require './router'
@@ -12,9 +11,7 @@ style = require '../css/main.styl'
 `import { Provider } from 'react-redux';`
 `import configureStore from './redux/store';`
 `import { processIntervention } from './redux/ducks/interventions';`
-`import { reset } from './redux/ducks/classify';`
 store = configureStore()
-auth.listen('change', () => store.dispatch(reset()));
 apiClient.type('subject_sets').listen('add-or-remove', () => store.dispatch(emptySubjectQueue()));
 sugarClient.on('experiment', (message) => store.dispatch(processIntervention(message)));
 
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -360,8 +360,10 @@ const mapDispatchToProps = dispatch => ({
 const ConnectedClassifyPage = connect(mapStateToProps, mapDispatchToProps)(ProjectClassifyPage);
 
 function ConnectedClassifyPageWithWorkflow(props) {
+  const workflowKey = props.workflow ? props.workflow.id : 'no-workflow';
   return (
     <WorkflowSelection
+      key={workflowKey}
       actions={props.actions}
       location={props.location}
       preferences={props.preferences}
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -26,7 +26,9 @@ class WorkflowSelection extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { locale, preferences, workflow } = this.props;
+    const { actions, locale, preferences, user, workflow } = this.props;
+    const prevUser = prevProps.user && prevProps.user.login;
+    const currentUser = user && user.login;
     const userSelectedWorkflow = (preferences && preferences.preferences) ? this.sanitiseID(preferences.preferences.selected_workflow) : undefined;
     if (userSelectedWorkflow && workflow) {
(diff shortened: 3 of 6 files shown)


eatyourgreens:
This PR changes `WorkflowSelection` to mount and run on each change to the stored workflow, so I've altered the `loadWorkflows` action to change state directly from old workflow -> new workflow. Before, it would go old workflow -> null -> new workflow and that null state might trigger workflow selection when it wasn't needed.

eatyourgreens:
@srallen I think `WorkflowSelection` could now be changed to use the `loadWorkflow` action here, but I'm not sure if that change should be made in this PR, or held off until later.
https://github.com/zooniverse/Panoptes-Front-End/blob/32a595bcc80ab3ca57052ff2a2e50c6dccba3493/app/pages/project/workflow-selection.jsx#L95-L111

srallen:
To fix #5188, I think the call to display the modal needs to only happen if props.workflow isn't null.

eatyourgreens:
That's weird. When you log out, the auth change handler should reset the workflow to null. Are there any errors in the console?

eatyourgreens:
I've just tried logging out on your test project and I see the stored workflow reset to null, then workflow selection mounts and loads a new workflow. I get the project default workflow when not logged in.

srallen:
Ok, let me try again. I'm thinking I might've forgotten to pull down the latest commits 🤦‍♀️

eatyourgreens:
I'm wondering if maybe project preferences aren't being reset to null when you log out?

I tried adding `?workflow=2334` to the classifier URL, and workflow 2334 was reloaded when I logged out. That shouldn't happen for an anonymous user if assignment is enabled, so I might have to trace what's happening inside the `getSelectedWorkflow` function.

srallen:
I double checked that I had the latest pulled down and tried both the staged branch and the local copy. 

What I'm doing is loading the classify page with the 'Journey' workflow (id 2334) while signed in, doing a single classification, then signing out. I looked in the React dev tools and preferences is set to null and passed down as such. The Redux tools show the classifier store with workflow id `2334` still in its state. I don't see any relevant console errors.

I tried the query param while signed out, `?workflow=2334`, and the default workflow `2333` loaded as expected.

`getSelectedWorkflow` will be called when preferences are null since `userSelectedWorkflow` is set to undefined when preferences is falsey in `componentDidUpdate` in the `WorkflowSelection` component. So it gets called with workflow id 2334 because that's what is in the store. I saw the request for 2334 in the network dev tools when signing out.

eatyourgreens:
If there’s a stored workflow then workflow selection won’t make a network request. It’ll just use the stored workflow.

I’m seeing the workflow being cleared, then a network request for the wrong workflow. I think that’s happening because `getSelectedWorkflow` doesn’t test whether a user exists when it gets the workflow ID from preferences.

eatyourgreens:
>getSelectedWorkflow will be called when preferences are null since userSelectedWorkflow is set to undefined when preferences is falsey in componentDidUpdate in the WorkflowSelection component.
 
`getSelectedWorkflow` is only called if `preferences` is truthy.
https://github.com/zooniverse/Panoptes-Front-End/blob/b0ad8e6f8e55f57e812e5ed22af2885754715fef/app/pages/project/workflow-selection.jsx#L30-L35

>I think we do want the reset action to reset the workflow to null in this case and then trigger the fallback to load

The reset action does set the stored workflow to null. It sets the store back to its initial state. The changes in this PR should then trigger workflow selection to mount with a null workflow prop, which should load the default workflow for an anonymous user.

eatyourgreens:
@srallen after adding some console logs to workflow selection, I think I see what's going on now. The auth change handler is fired before the user changes. On sign out, I'm seeing workflow selection being called with both my user and preferences as props, which then loads workflow 2334.

I see the 'signed out' message from the client appear in the console after the auth change handler has run workflow selection. 🤦‍♂️

eatyourgreens:
I gave up on the auth change listener and removed it. It seems to be really unreliable. Instead, I've updated workflow selection to reset the classifier when the user prop changes.

That seems to have actually fixed the issue.

srallen:
Ah, good catch! I think this is another good example of the panoptes javascript client doing event handling and data storage that is conflicting with React/Redux's way of doing things and we just can't have both. Looking over this again today and I'll post my updated review.

eatyourgreens:
@srallen I spent a couple of hours this morning playing with the auth change listener to see what it did. One thing I tried was resetting the classifier on auth change in the classifier component. That resulted in RESET firing six times when I logged out. 

Posting that here because I think it indicates that those PJC event handlers are more unreliable than we thought.

srallen:
@eatyourgreens yikes! Perhaps we should do an audit to see if they're being used for anything else?

eatyourgreens:
@srallen good idea! I think it's mostly used in the project builder and in the admin pages, to save resources on change. As far as I know, we've removed all the change listeners from the classification parts of PFE.

The PFE app updates the user context on auth change. I think that's why I was seeing weird behaviour in this PR: the two listeners were running at different times and the user context changed after the reset. I've no idea why I was seeing the handler run multiple times, though. https://github.com/zooniverse/Panoptes-Front-End/blob/a5f249bc934b17743967b5c6cc2cdcf2f0381aea/app/partials/app.cjsx#L49-L55

PR #5187, opened 2019-01-04
Add spec tests for workflow selection

### LINKED ISSUE #5165: Workflow selection can select workflows twice on error
https://github.com/zooniverse/Panoptes-Front-End/blob/8094513fa05210c331b4d55d5008e0366764d244/app/pages/project/workflow-selection.jsx#L94-L110
Workflow selection catches 404 errors from Panoptes, clears inactive workflows and runs selection again. Once the catch block has run, the following `then` block is run with the workflow undefined (since the catch block doesn't return anything.) If the passed workflow is undefined, the `then` block also clears inactive workflows and runs selection again.
https://github.com/zooniverse/Panoptes-Front-End/blob/8094513fa05210c331b4d55d5008e0366764d244/app/pages/project/workflow-selection.jsx#L113-L136

This could lead to a situation, as we've seen recently, where two different workflows are requested almost simultaneously, leading to problems.

In practice, I don't think the first catch block ever runs since we can't get a 404 error from a valid workflow, and `isValidWorkflow` must be true before we make the API request.

However, I think this code is confusing and easy to misunderstand, and should be fixed at some point.

### LINKED ISSUE #5173: Workflow selection can accidentally reset workflow assignment
https://github.com/zooniverse/Panoptes-Front-End/blob/8094513fa05210c331b4d55d5008e0366764d244/app/pages/project/workflow-selection.jsx#L161-L175

`clearInactiveWorkflows` doesn't check if a workflow is actually inactive. It's possible for an active workflow ID to be passed down to it after an API client error. User preferences would then be cleared when they contain a valid workflow ID. This then prevents those volunteers from being promoted on Gravity Spy, since the project team will have them down as having been promoted and further classifications will not set `preferences.settings.workflow_id` again.

I think this would be a rare bug. It would happen in the case where `isValidWorkflow` is true but `workflow` is undefined. Maybe for API errors where the status is something other than 404, or perhaps if the API client rejects the workflow request promise after a timeout. The underlying cause is that we have a `then` block which runs after request errors have been caught and handled, but that block also contains error-handling code that assumes the workflow ID is not valid.

Based on my experience, the symptoms would be volunteers finding themselves stuck on level 1 and unable to advance to level 2. They would probably have seen the promotion dialog, but when they clicked it to advance, a server error caused PFE to delete the project set workflow by mistake.

_Originally posted by @eatyourgreens in https://github.com/zooniverse/Panoptes-Front-End/issues/5165#issuecomment-448955518_

eatyourgreens:
Spec out how loading a workflow from the API should work in the Workflow Selection component, based on the workflow tests in #5124. Adds some tests for successful workflow requests and a couple of failing tests for API errors, reproducing #5173 and #5165.

Update workflow error handling so that it only runs if the workflow ID is invalid, rather than for all workflow IDs.

Fixes #5173.
Fixes #5165.

Staging branch URL: https://pr-5187.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -99,7 +99,8 @@ class WorkflowSelection extends React.Component {
         .type('workflows')
         .get(sanitisedWorkflowID, {}) // the empty query here forces the client to bypass its internal cache
         .catch((error) => {
-          if (error.status === 404) {
+          const errorType = error.status && parseInt(error.status / 100, 10);
+          if (errorType === 4) {
             this.clearInactiveWorkflow(sanitisedWorkflowID)
             .then(this.getSelectedWorkflow(this.props));
           } else {
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -103,39 +103,37 @@ class WorkflowSelection extends React.Component {
             this.clearInactiveWorkflow(sanitisedWorkflowID)
             .then(this.getSelectedWorkflow(this.props));
           } else {
-            console.error(error);
+            console.error(error.message);
             this.setState({ error, loadingSelectedWorkflow: false });
           }
+          return null;
         });
     } else {
       awaitWorkflow = Promise.resolve(null);
       awaitTranslation = Promise.resolve(null);
+      if (process.env.BABEL_ENV !== 'test') console.log(`No workflow ${selectedWorkflowID} for project ${this.props.project.id}`);
+      if (this.props.project.configuration &&
+        selectedWorkflowID === this.props.project.configuration.default_workflow
+      ) {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Moving the invalid workflow error-handling code to only run if `isValidWorkflow` is false fixes the failing tests. I think that's the correct behaviour for requesting inactive workflows.

This PR would be a good place to update workflow error handling to cover all 400 errors, including authorisation errors, as per the discussion on #5124. https://github.com/zooniverse/Panoptes-Front-End/pull/5124#pullrequestreview-189055680

eatyourgreens:
Coveralls shows that the `getWorkflows` function is pretty much completely covered too. 👍

Issue #5162, opened 2018-12-17
Hide previous marks crashing when signing out on classify page


srallen:
## Expected behavior
The hide previous marks toggle shouldn't throw and crash the app. Does this really need to make this call to update annotations on unmount or at all?

## Current behavior
_Please include any error messages from the browser console and/or screenshots_

I signed out while on the classify page and got:
`Uncaught TypeError: Cannot read property 'update' of null`

It's throwing on line 221 in the classify duck. The hide previous marks toggle is calling to update the annotations on unmount.

## Steps to replicate
You can replicate on my test project: https://master.pfe-preview.zooniverse.org/projects/srallen086/sgl-tester

1. Sign in if not already. 
2. Select "freeline tool test not required" workflow and load the classify page. 
3. Sign out

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Hide Previous Marks was originally stored as an underscored property on the classificiation. It would be great if someone could find the time to rewrite it so that it doesn't use either the classification or annotations to persist across components.

eatyourgreens:
I'm a little confused by the bug report, though. That line toggles gold standard classifications.
https://github.com/zooniverse/Panoptes-Front-End/blob/7eb2125d7f966f87cc84be913d7706975817a5d7/app/redux/ducks/classify.js#L218-L223

srallen:
Whoops! That is a typo. It's line 211, not 221. I should not be lazy and use the permalinks. This is what the stack trace was leading to:

https://github.com/zooniverse/Panoptes-Front-End/blob/8094513fa05210c331b4d55d5008e0366764d244/app/redux/ducks/classify.js#L209-L213

Being called from line 23 here:

https://github.com/zooniverse/Panoptes-Front-End/blob/21cf42485929a62938112d5e2d4bca1a6702b00e/app/classifier/tasks/drawing/hide-previous-marks-toggle.jsx#L19-L24

It's throwing because `classification` is undefined?

eatyourgreens:
Probably. I guess that's preserving the component state when you leave the classifier and come back, but logging out (I assume) clears the classification anyway.

From what I remember, Hide Previous Marks preserves state between the subject viewer and the task components by saving its current state on the classification. I think that dates back to when the classification was a global var on the window object, so it was used as a permanent store.

PR #5198, opened 2019-01-10
Workflow selection updates and bug fixes

### LINKED ISSUE #5188: Several clicks of 'take me to new workflow' to close modal
## Expected behavior
Clicking 'take me to new workflow' should only be needed once. It should load the workflow that is stored in `settings.workflow_id` in the user's project preferences.

## Current behavior
If I'm classifying on level 1, clear the subject queue, and get prompted to try the level I'm promoted to, then it takes several clicks of 'take me to the new workflow' to get the modal to clear. The classifier appears to be loading slowly behind the modal. The only issue in the console I see is:

`Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

Which traces to: 
https://github.com/zooniverse/Panoptes-Front-End/blob/860e7624fb19c711dcd47d0bbbea490b8ce3164a/app/components/subject-viewer.cjsx#L229

And it's not being properly cleared on timeout. 

I'm also seeing the modal open several times, so it's not the same modal that I'm clicking to proceed on multiple times:

screen shot 2019-01-07 at 11 24 08 am

I didn't see this in previous testing of the recent changes to workflow selection, so I'm not sure why this has started up. 

## Steps to replicate
Using Gravity Spy to test: https://master.pfe-preview.zooniverse.org/projects/zooniverse/gravity-spy?env=production with a user that is promoted.

Or my test project: 
https://master.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing
Start with the beginner workflow and promote your user to workflow id `2334` (journey) and when the subject queue is next clear while classifying on the beginner workflow, you'll be prompted to try the new workflow. 

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Remove the API client from workflow selection. Instead, use the loadWorkflows redux action once a workflow ID has been selected.

Remove user preference code from workflow selection, since the loadWorkflows action also updates preferences with the selected workflow, if successful.

Update the home workflow buttons to check the stored workflow and only load the classifier if the stored workflow matches the workflow for that button.

Check workflow assignment prompt state on the classify page before checking whether to show a new prompt.

Validate the stored workflow against the project before rendering the classifier.

Staging branch URL: https://pr-5198.pfe-preview.zooniverse.org

Fixes #5188.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/home/home-workflow-button.spec.js
+++ b/app/pages/project/home/home-workflow-button.spec.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
+import { browserHistory } from 'react-router';
 import sinon from 'sinon';
 import apiClient from 'panoptes-client/lib/api-client';
 import { ProjectHomeWorkflowButton } from './home-workflow-button';
@@ -98,8 +99,36 @@ describe('ProjectHomeWorkflowButton', function () {
     it('calls handleWorkflowSelection onClick', function() {
       expect(handleWorkflowSelectionSpy).to.have.been.calledOnce;
     });
-    it('should load a new workflow', function () {
-      expect(actions.classifier.loadWorkflow).to.have.been.calledWith(testWorkflowWithoutLevel.id, translations.locale, preferences);
+
+    describe('for a new workflow', function () {
+      after(function () {
+        actions.classifier.loadWorkflow.resetHistory();
+      });
+
+      it('should load the workflow', function () {
+        expect(actions.classifier.loadWorkflow).to.have.been.calledWith(testWorkflowWithoutLevel.id, translations.locale, preferences);
+      });
+    });
+
+    describe('for the current workflow', function () {
+      before(function () {
+        sinon.stub(browserHistory, 'push');
+        wrapper.setProps({ classifierWorkflow: testWorkflowWithoutLevel });
+      });
+
+      after(function () {
+        browserHistory.push.restore();
+        wrapper.setProps({ classifierWorkflow: undefined });
+      });
+
+      it('should not load the workflow', function () {
+        expect(actions.classifier.loadWorkflow).to.have.not.been.called;
+      });
+
+      it('should load the classify page', function () {
+        expect(browserHistory.push).to.have.been.calledOnce;
+        expect(browserHistory.push).to.have.been.calledWith(`/projects/${testProject.slug}/classify`);
+      });
     });
   });
 
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -148,8 +148,9 @@ class WorkflowSelection extends React.Component {
   }
 
   render() {
-    const { children, workflow } = this.props;
-    return workflow ? children : <p>Loading workflow</p>;
+    const { children, project, workflow } = this.props;
+    const validWorkflowForProject = workflow && workflow.links.project === project.id;
+    return validWorkflowForProject ? children : <p>Loading workflow</p>;
   }
 }
 
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -149,7 +149,7 @@ class WorkflowSelection extends React.Component {
 
   render() {
     const { children, workflow } = this.props;
-    return workflow ? children: <p>Loading workflow</p>;
+    return workflow ? children : <p>Loading workflow</p>;
   }
 }
 
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -24,24 +24,17 @@ class WorkflowSelection extends React.Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { actions, locale, preferences, user, workflow } = this.props;
-    const prevUser = prevProps.user && prevProps.user.login;
-    const currentUser = user && user.login;
-    const userSelectedWorkflow = (preferences && preferences.preferences) ? this.sanitiseID(preferences.preferences.selected_workflow) : undefined;
-    if (userSelectedWorkflow && workflow) {
-      const prevWorkflowID = prevProps.workflow && prevProps.workflow.id;
-      const workflowChanged = workflow.id !== prevWorkflowID;
-      if (workflowChanged && userSelectedWorkflow !== workflow.id) {
-        this.getSelectedWorkflow(this.props);
-      }
-    }
+    const { actions, locale, preferences, workflow } = this.props;
+    const prevPrefs = prevProps.preferences && prevProps.preferences.id;
+    const currentPrefs = preferences && preferences.id;
+    
     if (prevProps.project.id !== this.props.project.id) {
       this.getSelectedWorkflow(this.props);
     }
     if (workflow && prevProps.locale !== locale) {
       actions.translations.load('workflow', workflow.id, locale);
     }
-    if (currentUser !== prevUser) {
+    if (currentPrefs !== prevPrefs) {
       actions.classifier.reset();
     }
   }
--- a/app/pages/project/workflow-selection.jsx
+++ b/app/pages/project/workflow-selection.jsx
@@ -59,17 +59,11 @@ class WorkflowSelection extends React.Component {
     ) {
       selectedWorkflowID = workflowFromURL;
       activeFilter = false;
-      if (userSelectedWorkflow && userSelectedWorkflow !== selectedWorkflowID) {
-        this.handlePreferencesChange('preferences.selected_workflow', selectedWorkflowID);
-      }
     } else if (workflowFromURL &&
       project.experimental_tools &&
       project.experimental_tools.indexOf('allow workflow query') > -1
     ) {
       selectedWorkflowID = workflowFromURL;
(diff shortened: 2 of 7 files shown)


eatyourgreens:
This PR updates the project workflow buttons to behave more like links, which made me feel more strongly that those buttons should just be links, and we should allow all projects to link to workflows by ID in the classify page URL, as per #5125.

shaunanoordin:
Quick notes:
- Basic testing (login/logout, Projects page, Project pages, ability to submit Classifications on a test project) look good on OSX+Chrome
- code changes LGTM with no glaring issues, though I'll need a second read through to make sure I'm following the logic flow correctly.
- Manual testing is taking a bit longer since I have no idea how to perform promotions on myself.

eatyourgreens:
NB. the fix for #5188 is the changes in 70de137: check whether we're already showing the prompt before showing it again. The other changes here are tidying up workflow selection, so we don't have the same API code in two places, and making sure the home page buttons don't load a workflow if we're using it already.

app/pages/project/workflow-selection.jsx
```diff
@@ -133,16 +104,9 @@ class WorkflowSelection extends React.Component {
             this.getSelectedWorkflow(this.props);
```
eatyourgreens: Should this be replaced with `actions.classifier.loadWorkflow(default_workflow, locale, preferences)`? Loading the default workflow on error seems more sensible to me (as long as we indicate to the user that the workflow has been changed from the one that they thought was loading.)

app/pages/project/home/home-workflow-button.spec.js
```diff
@@ -98,8 +99,36 @@ describe('ProjectHomeWorkflowButton', function () {
...
+    });
+
+    describe('for the current workflow', function () {
+      before(function () {
+        sinon.stub(browserHistory, 'push');
+        wrapper.setProps({ classifierWorkflow: testWorkflowWithoutLevel });
+      });
+
+      after(function () {
+        browserHistory.push.restore;
```
eatyourgreens: Typo. Should be `restore()`.

app/pages/project/workflow-selection.jsx
srallen: That's a change in behavior that I think we should discuss and if we go forward with it, implement in a separate PR. Even though a lot of projects have default workflows set, not all do, so for what it's worth, I think just running through the strategies again as it does now is overall safer. 

Informing the user is something that @shaunanoordin has brought up and I think better fits as part of the rewrite.

eatyourgreens:
That's annoying. I don't get the type errors when running the tests (using `docker-compose run dev npm run test`.)

Loading state's been removed because the store is updating now, when a workflow loads, so the component reacts to changes in the workflow prop instead.

I'll take a look at why the wrong workflow is being loaded on login. It sounds like workflow selection is mounting while the user prop is still null.

eatyourgreens:
I can reproduce that sign-out/sign-in problem on the classify page, with my user account, on this branch, but not on https://master.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing/classify so it's definitely a problem with this PR.

app/pages/project/workflow-selection.jsx
```diff
@@ -31,7 +29,9 @@ class WorkflowSelection extends React.Component {
     const currentUser = user && user.login;
     const userSelectedWorkflow = (preferences && preferences.preferences) ? this.sanitiseID(preferences.preferences.selected_workflow) : undefined;
     if (userSelectedWorkflow && workflow) {
-      if (!this.state.loadingSelectedWorkflow && userSelectedWorkflow !== workflow.id) {
+      const prevWorkflowID = prevProps.workflow && prevProps.workflow.id;
+      const workflowChanged = workflow.id !== prevWorkflowID;
+      if (workflowChanged && userSelectedWorkflow !== workflow.id) {
```
eatyourgreens: This change isn't loading a new workflow, from preferences, when a user logins in. Possibly because `userSelectedWorkflow` doesn't change at the same time as the stored workflow. This should maybe listen for a change in user ID, or a change in user preferences ID, rather than a change in workflow ID.

eatyourgreens:
>I found a regression testing https://pr-5198.pfe-preview.zooniverse.org/projects/srallen086/workflow-assignment-testing/classify. If visiting the classify page signed out, the default workflow, id 2333, loads. I sign in, then the classifier resets and loads the default workflow again instead of my UPP selected workflow which is id 2334. What should happen is the workflow selection shouldn't run until the the UPP request finishes and the state is set on the ProjectPageController.

Changing the component to reset the classifier on preferences change, rather than user change, fixes this.

app/pages/project/workflow-selection.jsx
```diff
@@ -60,17 +52,11 @@ class WorkflowSelection extends React.Component {
     ) {
       selectedWorkflowID = workflowFromURL;
       activeFilter = false;
```
eatyourgreens: The active filter isn't being used by Gravity Spy, only by Notes from Nature. Is that correct?

app/pages/project/workflow-selection.jsx
eatyourgreens: Looking at the code, this line only runs if `project.configuration` exists but the selected workflow ID is not the project default. There's no check to see if a project default actually does exist so changing this would introduce bugs.

If there is a workflow loaded at this point, then `actions.classifier.reset()` later in this block will also force selection to run again, since changing the stored workflow unmounts and remounts workflow selection (and the classifier.)

eatyourgreens: These nested if blocks are also very confusing.

app/pages/project/workflow-selection.jsx
eatyourgreens: Resolved by deleting this block of code and checking for a change of preferences instead.

eatyourgreens:
@srallen I don't see any type errors in [the tests on Travis](https://travis-ci.org/zooniverse/Panoptes-Front-End/builds/479795673), so I think this is ready now.

app/pages/project/workflow-selection.jsx
srallen: The active filter is true for most projects' workflows, but changed to false if the user is an admin, project owner, tester, or expert that may have a reason to load an inactive workflow. `this.checkUserRoles(project, user)` this returns true for these user roles and then the active filter is changed to false.

For Notes from Nature, the next set of conditions is for them and projects like them. 
```
else if (workflowFromURL &&
      project.experimental_tools &&
      project.experimental_tools.indexOf('allow workflow query') > -1
    ) {
      selectedWorkflowID = workflowFromURL;
```

The active filter isn't changed and stays true.

eatyourgreens: I forgot about previewing workflows in development. 😳 

I remember we talked about workflow assignment for inactive workflows on #5171 where you mentioned owners being able to load inactive workflows, so thought I'd check because the workflow selection code does not allow that.
https://github.com/zooniverse/Panoptes-Front-End/pull/5171#pullrequestreview-189074992

eatyourgreens: Easy enough to change though, by setting the filter from the user roles earlier in the code.
`let activeFilter = this.checkUserRoles(project, user);`

eatyourgreens:
I can reproduce that too. The crash is happening in the `HidePreviousMarksToggle`, which I think you've come across before? I can't reproduce the crash on staging, switching between I Fancy Cats and my own test project. My test project has a default workflow, but I think American Soldier might not have one.

On staging and production, I'm also seeing the tutorial for the previous project open eg. on production, American Soldier launches the tutorial for TESS.

I'm packing up for the day but I'll look into this tomorrow.

eatyourgreens:
I think that crash is the same bug as #5162.

srallen:
I was using The American Soldier to test random workflow selection. I can't get this issue to replicate on master using the same steps I was using before, so I'm not sure if it is the same as #5162

eatyourgreens:
I’m seeing the same error message and stack trace as #5162. `HidePreviousMarksToggle` is attempting to update a null classification.

eatyourgreens:
@srallen I've updated the redux action and `HidePreviousMarksToggle` to guard against updating null classifications or undefined annotations. I'm not going to mark #5162 as fixed, though, because that component should not be updating the classification, as noted in the comments on #5162.

eatyourgreens:
I've also updated the workflow selection render method to validate workflows against the project before rendering the classifier. That should prevent American Soldier from rendering the TESS workflow while the American Soldier workflow is loading.

app/pages/project/workflow-selection.jsx
```diff
@@ -197,8 +148,10 @@ class WorkflowSelection extends React.Component {
   }
 
   render() {
-    const { loadingSelectedWorkflow } = this.state;
-    return loadingSelectedWorkflow ? <p>Loading workflow</p> : this.props.children;
+    const { children, project, workflow } = this.props;
+    const linkedWorkflows = project.links.workflows || [];
+    const workflowExistsForProject = workflow && linkedWorkflows.indexOf(workflow.id) > -1;
```
eatyourgreens: Should this be inverted and check `workflow.links.project` against the project ID?

PR #5233, opened 2019-01-25
Fix hero navs on JSX pages

lancesnider:
1. On small screens the `hero-nav` splits into multiple lines in strange places. 

screen shot 2019-01-25 at 3 59 14 pm

This was because each button was just a span so the browser was treating them as one long word. It would break it into multiple lines where there were word breaks within a button. 

screen shot 2019-01-25 at 2 57 19 pm

I wanted to make a `HeroNav` component, but each set of links was built in such a unique way that I don't think `HeroNav` would have added much value. Maybe a component would be nice in the future, but for now wrapping menus in `<ul>`'s does the trick. 

screen shot 2019-01-25 at 4 25 39 pm
screen shot 2019-01-25 at 4 24 39 pm
screen shot 2019-01-25 at 4 24 30 pm
screen shot 2019-01-25 at 4 24 22 pm

I also added a `focus` state so that you can see which button or `<a>` is selecting when tabbing through the page. 

screen capture on 2019-01-25 at 15-26-49

## Out of scope

I didn't fix the collections `hero-nav` menu. That page is currently built in .cjsx. I went to convert it to .jsx, but that would have required rewriting at least a couple other .cjsx files. I'm happy to do that, but it should probably be its own PR. Alternatively, I could just update the .cjsx to include `<ul>`'s. 

# Required Manual Testing

# Review Checklist

- [x] Does it work on mobile?

# Optional


css/secondary-page.styl
```diff
@@ -55,11 +55,25 @@
...
+        li
+          margin: 0
+          padding: 0
+
         a
           @extends .main-nav-item
           font-size: 14px
           line-height: 3em
           text-decoration: none
+          &:focus
```
srallen: Could you also apply these styles for hover? `&:focus, &:hover`

srallen:
@lancesnider I agree updating the collections pages are out of scope of this PR

css/secondary-page.styl
lancesnider: Done. :) Sorry for the late reply. I've been trekking around the Colombian jungle. ;)

srallen: No worries! I was on holiday last week too!

PR #5236, opened 2019-01-28
Workflow selection: default preferences to null

eatyourgreens:
Set preferences to null by default, otherwise the loadWorkflows action will fail with 'preferences.update is not a function'.

Staging branch URL: https://pr-5236.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/index.jsx
+++ b/app/pages/project/index.jsx
@@ -139,22 +139,20 @@ class ProjectPageController extends React.Component {
   }
 
   getUserProjectPreferences(project, user) {
-
     const userPreferences = user ?
       user.get('project_preferences', { project_id: project.id })
         .then(([projectPreferences]) => {
-          let newPreferences;
-          return projectPreferences ||
-            (newPreferences = apiClient.type('project_preferences').create({
+          if (projectPreferences) {
+            return projectPreferences;
+          } else {
+            return apiClient.type('project_preferences').create({
               links: {
                 project: project.id
               },
               preferences: {}
             })
-            .save()
-            .catch((error) => {
-              console.warn(error.message);
-            }));
+            .save();
+          }
         })
     :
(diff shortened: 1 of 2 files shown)


srallen:
@eatyourgreens Which project initially reported the bug that this is fixing?

eatyourgreens:
The project page container returns preferences as undefined from the catch block here, which might be the source of their problem.
https://github.com/zooniverse/Panoptes-Front-End/blob/792aaa5816f3518498fbd4501fe5e3f2a3f9a20b/app/pages/project/index.jsx#L155-L157

eatyourgreens:
@srallen to reproduce this bug, throw an error from the API request in [`ProjectPageController.getUserPreferences`](https://github.com/zooniverse/Panoptes-Front-End/blob/326f933aed727063de7ba4e87ab2e94946bd7036/app/pages/project/index.jsx#L148-L173).

Doing that crashes the React app for me on master.
Screenshot showing 'preferences.update is not a function' error in WorkflowSelection.

srallen:
FWIW, this branch is in a detached head state for me when I check it out. Not sure what happened, but maybe in reusing it, this is happening. I typically delete branches that have been remotely deleted. In the future for specific bugs, perhaps we can have clean branches and issues to make it clear what's being fixed?

eatyourgreens:
Weird. It was built from a clean branch locally. Just those two commits on top of master.

srallen:
Huh, I haven't figured out exactly what causes the detached head state. I thought the branch was being reused because it has the same name as an old branch? It also might be something I'm doing.

eatyourgreens:
I reused the name of an old, deleted branch so it could be that.

PR #5259, opened 2019-02-06
Add a workflow selection strategy for classrooms

### LINKED ISSUE #5256: For Zooniverse Classrooms, experimental "allow workflow query" should allow inactive workflows
## Discussion

Related to: [Zooniverse Classrooms](https://classroom.zooniverse.org/), WildCam Darien, WildCam Darien Lab, WildCam Gorongosa

Situation: in WildCam Darien Lab (and eventually WildCam Gorongosa Lab), we want Teachers to create Assignments (new EduAPI assignment = new Zooniverse workflow) which Students can work on in PFE.

Example: Teacher X creates Assignment Y in Zooniverse Classrooms. Student Z goes to PFE to work on the Assignment by **selecting the workflow via query** - e.g. at the URL `https://www.zooniverse.org/projects/wildcam/wildcam-darien/classify?workflow_id=8752`.

Problem:
- The ability to _select a workflow_ (i.e. "Assignment") via query works by setting `project.experimental_tools = ["allow workflow query"]`
- However, this only works **if the selected workflow is active**
- The **workflows created by WildCam Assignments are inactive**, to prevent Assignments-based Classifications from accidentally retiring Subjects too early.

### Code Dive

In the `workflow-selection` [code](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/project/workflow-selection.jsx), we see this:

```
  getSelectedWorkflow({ project, preferences, user }) {
    ...
    let activeFilter = true;  // <-- by default, only active workflows can be selected
    ...
    if (workflowFromURL &&
      this.checkUserRoles(project, user)
    ) {
      selectedWorkflowID = workflowFromURL;
      activeFilter = false;  //  <-- the exception to this rule is if the user is an admin, owner, collaborator, etc
    } else if (workflowFromURL &&
      project.experimental_tools &&
      project.experimental_tools.indexOf('allow workflow query') > -1
    ) {
      selectedWorkflowID = workflowFromURL;  // <-- so even if we enable the experimental tool, students from WildCam Lab can't select their workflows
    } ...
  }
```

### Proposed Solutions

**Solution A:**  modify `"allow workflow query"` to query even inactive workflows, i.e. add `activeFilter = false;` to that if-else up there. ☝️ 

- Problem: this may have unintended side effects on projects that use this. e.g. with Notes from Nature, perhaps it will allow users with bookmarks to select old, inactive workflows?
- Anybody knows of any other flaw?

**Solution B:** create a new experimental tool that says `"allow workflow query even if inactive"` and does exactly that

- How To Zooniverse would need to be updated if this is the case.

### Status

I'm currently creating a PR to implement Solution B, but I wanted to bring this up for discussion first in case adding a new experimental tool might have some drawbacks.

Tagging @mcbouslog and @srallen for comments, since I believe you were both involved in the original "allow workflow query" design.

eatyourgreens:
Add a new workflow strategy for classrooms, which always loads the workflow from a workflow query parameter.

ToDo:

- [x] Conditional logic to choose the selection strategy
- [x] Tests

Staging branch URL: https://pr-5259.pfe-preview.zooniverse.org

Fixes #5256.

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -365,9 +365,10 @@ function ConnectedClassifyPageWithWorkflow(props) {
   const workflowKey = props.workflow ? props.workflow.id : 'no-workflow';
   
   //Check for WildCam Lab classrooms (see https://github.com/zooniverse/edu-api-front-end)
+  const workflowFromUrl = props.location.query && props.location.query.workflow;
   const isProjectForClassrooms = (props.project && props.project.experimental_tools && props.project.experimental_tools.indexOf('wildcam classroom') > -1);
   const isUrlForClassrooms = props.location.query && props.location.query.classroom;
-  const isClassroom = isProjectForClassrooms && isUrlForClassrooms;
+  const isClassroom = isProjectForClassrooms && isUrlForClassrooms && workflowFromUrl;
   
   const WorkflowStrategy = isClassroom ? ClassroomWorkflowSelection : WorkflowSelection;
   return (
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -363,7 +363,12 @@ const ConnectedClassifyPage = connect(mapStateToProps, mapDispatchToProps)(Proje
 
 function ConnectedClassifyPageWithWorkflow(props) {
   const workflowKey = props.workflow ? props.workflow.id : 'no-workflow';
-  const isClassroom = true; // TODO: add a conditional check for classrooms here.
+  
+  //Check for WildCam Lab classrooms (see https://github.com/zooniverse/edu-api-front-end)
+  const isProjectForClassrooms = (props.project && props.project.experimental_tools && props.project.experimental_tools.indexOf('wildcam classroom') > -1);
+  const isUrlForClassrooms = props.location.query && props.location.query.classroom;
+  const isClassroom = isProjectForClassrooms && isUrlForClassrooms;
+  
   const WorkflowStrategy = isClassroom ? ClassroomWorkflowSelection : WorkflowSelection;
   return (
     <WorkflowStrategy
--- a/app/pages/project/classify.jsx
+++ b/app/pages/project/classify.jsx
@@ -20,6 +20,7 @@ import FinishedBanner from './finished-banner';
 import WorkflowAssignmentDialog from '../../components/workflow-assignment-dialog';
 import ProjectThemeButton from './components/ProjectThemeButton';
 import WorkflowSelection from './workflow-selection';
+import ClassroomWorkflowSelection from './workflow-selection-classroom';
 import { zooTheme } from '../../theme';
 
 function onClassificationSaved(actualClassification) {
@@ -362,8 +363,10 @@ const ConnectedClassifyPage = connect(mapStateToProps, mapDispatchToProps)(Proje
 
 function ConnectedClassifyPageWithWorkflow(props) {
   const workflowKey = props.workflow ? props.workflow.id : 'no-workflow';
+  const isClassroom = true; // TODO: add a conditional check for classrooms here.
+  const WorkflowStrategy = isClassroom ? ClassroomWorkflowSelection : WorkflowSelection;
   return (
-    <WorkflowSelection
+    <WorkflowStrategy
       key={workflowKey}
       actions={props.actions}
       location={props.location}
@@ -373,7 +376,7 @@ function ConnectedClassifyPageWithWorkflow(props) {
       user={props.user}
     >
       <ConnectedClassifyPage {...props} />
-    </WorkflowSelection>
+    </WorkflowStrategy>
   );
 }
 export default ConnectedClassifyPageWithWorkflow;
--- a/app/pages/project/workflow-selection-classroom.jsx
+++ b/app/pages/project/workflow-selection-classroom.jsx
@@ -1,0 +1,29 @@
+import { connect } from 'react-redux';
+import { bindActionCreators } from 'redux';
+import * as classifierActions from '../../redux/ducks/classify';
+import * as translationActions from '../../redux/ducks/translations';
+import { WorkflowSelection } from './workflow-selection';
+
+export class ClassroomWorkflowSelection extends WorkflowSelection {
+  getSelectedWorkflow({ location }) {
+    const workflowFromURL = this.sanitiseID(location.query.workflow);
+    if (workflowFromURL) return this.getWorkflow(workflowFromURL, false);
+    if (process.env.BABEL_ENV !== 'test') console.warn('Cannot select a workflow.');
+    return Promise.resolve(null);
+  }
+}
+
+const mapStateToProps = state => ({
+  locale: state.translations.locale,
+  workflow: state.classify.workflow
+});
+
+const mapDispatchToProps = dispatch => ({
+  actions: {
+    classifier: bindActionCreators(classifierActions, dispatch),
+    translations: bindActionCreators(translationActions, dispatch)
+  }
+});
+
+export default connect(mapStateToProps, mapDispatchToProps)(ClassroomWorkflowSelection);
+
--- a/app/pages/project/workflow-selection-classroom.jsx
+++ b/app/pages/project/workflow-selection-classroom.jsx
@@ -5,9 +5,10 @@ import * as translationActions from '../../redux/ducks/translations';
 import { WorkflowSelection } from './workflow-selection';
 
 export class ClassroomWorkflowSelection extends WorkflowSelection {
-  getSelectedWorkflow({ location }) {
+  getSelectedWorkflow(props) {
+    const { actions, locale, location, preferences } = props;
     const workflowFromURL = this.sanitiseID(location.query.workflow);
-    if (workflowFromURL) return this.getWorkflow(workflowFromURL, false);
+    if (workflowFromURL) return actions.classifier.loadWorkflow(workflowFromURL, locale, preferences);
     if (process.env.BABEL_ENV !== 'test') console.warn('Cannot select a workflow.');
     return Promise.resolve(null);
   }
--- a/app/pages/project/workflow-selection-classroom.spec.js
+++ b/app/pages/project/workflow-selection-classroom.spec.js
@@ -1,0 +1,170 @@
+import React from 'react';
+import { expect } from 'chai';
+import { mount } from 'enzyme';
+import sinon from 'sinon';
+import { ClassroomWorkflowSelection } from './workflow-selection-classroom';
+import mockPanoptesResource from '../../../test/mock-panoptes-resource';
+
+function StubPage(props) {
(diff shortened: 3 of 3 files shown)


eatyourgreens:
@shaunanoordin all this should need now is a conditional to decide when to use the new strategy.

shaunanoordin:
## Discussion Notes

Some quick documentation of the real-life discussion Jim & me had.

(Context for people jumping in: read #5256, basically WildCam Lab needs the ability to choose workflows (i.e. classroom assignments) that are inactive, which requires modifying the PFE workflow selection logic, which would be best served by creating a Worklow Selection Strategy _Selector._ )

- Point: the conditional logic to check if we should use the "Classroom's workflow selection strategy" could just check if `?classroom=yes` is in the query.
- Counterpoint: wait, no, bad idea, this means any Project will let user bypass standard workflow selection by passing in a query.

- Point: instead of specifying `?workflow=1234&classroom=yup`, we could just use `?classroom=1234`
- Counterpoint: see above - if we're not checking the query to see if we're in a classroom or not, then this is a bit pointless.

- Point: we should be wary about letting users access workflows they're not supposed to, in PFE.
- Counterpoint: do we though? @marten pointed out that if we wanted to be really strict about workflow access, security needs to be placed on the backend, not the front end, where anyone with a modicum of JS skills can hack things.

- Point: it'll make more sense for the conditional logic to check if a Project is _configured_ to use the "Classroom Workflow Selection Strategy"
- Counterpoint: better figure out whether you want to put that config in `project.config` or `project.experimental_tools`
- Answer: I say `experimental_tools` for now.

- Point: this new Workflow Selection Strategy philosophy should be implemented to reduce overloading an already fragile Workflow Selection (see `workflow-selection.jsx`) logic.
- Counterpoint: in that case, better check that [front-end-monorepo](https://github.com/zooniverse/front-end-monorepo) is already thinking about this.

- Point: 👉 
- Counterpoint: 👈

eatyourgreens:
@shaunanoordin will the classroom projects also allow volunteers to choose a workflow from the home page? Things get a bit more complicated if the new component is also used by regular volunteers.

In an ideal world, the `ClassroomWorkflowSelection` component would only be used by students doing classroom work.

shaunanoordin:
> will the classroom projects also allow volunteers to choose a workflow from the home page? Things get a bit more complicated if the new component is also used by regular volunteers

Fortunately for this case, WildCam Lab's assignments create workflows that are _inactive,_ so 1. they won't appear on the home page even if allowed the standard "Users Can Choose Workflow" option (tested yesterday), and 2. Classifications on the workflow won't accelerate retirement.

eatyourgreens:
Thanks @shaunanoordin. Will there ever be a case where there are two or more active workflows on the home page, for regular volunteers, plus a bunch of inactive workflows for students?

In that case, we'd have to write logic to use regular selection for the home page choices but classroom selection for the students.

shaunanoordin:
Note: at the moment, I'm already looking at some code logic for the conditional check.

Solution 1A: check if the Project has the `wildcam classroom` Experimental Tool.

`const isClassroom = (props.project && props.project.experimental_tools && props.project.experimental_tools.indexOf('wildcam classroom') > -1);` works for [WildCam Darien](https://www.zooniverse.org/projects/wildcam/wildcam-darien/) (which has `'wildcam classroom'` in its experimental_tools config)... but only if a `?workflow` is specified. Solution 1A breaks for standard volunteers attempting to work on the default workflow, because it passes standard voluneers to Classroom Workflow Selection strategy even if they're not students.

So...

Solution 1B: check if the Project has the `wildcam classroom` Experimental Tool AND there's `classroom=1` in the query

Working on this.

eatyourgreens:
@shaunanoordin I guess the inverse of that would also be that classrooms can't be added to projects that have more than one active workflow.

eatyourgreens:
NB. [ES6 super() syntax](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/super) would allow us to fall back by calling `super.getSelectedWorkflow(props)` but I'm not sure if that's the desired behaviour for classroom students. It would definitely work for regular volunteers.

eatyourgreens:
@shaunanoordin I'm leaning towards thinking that the logic should be: if the project has classrooms AND the URL looks like a classroom URL (has a workflow param, maybe has a classroom param too?)

We could update the PFE routes to map `/projects/{project_slug}/classroom` to the classifier, then use the new strategy on that route. It depends whether changing the classroom links is too big of a change.

shaunanoordin:
@eatyourgreens Changing the classroom links, at Zooniverse Classrooms (aka Education API Front End) ain't a problem whatsoever: https://github.com/zooniverse/edu-api-front-end/blob/master/src/programs/darien/wildcam-darien.classroom-config.js

eatyourgreens:
By the way, I set up the fallback here to return null, rather than retrying and loading the default workflow, because I assumed that a student would not want to start doing their classroom assignment only to realise they were on the active workflow.

shaunanoordin:
I'm about to push some changes to this PR that will address the _conditional check._

```
const isClassroom = true; 
---- change to -->
//Check for WildCam Lab classrooms (see https://github.com/zooniverse/edu-api-front-end)
const isProjectForClassrooms = (props.project && props.project.experimental_tools && props.project.experimental_tools.indexOf('wildcam classroom') > -1);
const isUrlForClassrooms = props.location.query && props.location.query.classroom;
const isClassroom = isProjectForClassrooms && isUrlForClassrooms;

```

- a Classroom URL is only valid if 1. the Project is configured to accept classrooms and 2. the URL has the `?classrooms=yup` in its query
- if the Classroom URL is invalid, the Classifier uses the standard workflow selection logic.
- if the Classroom URL is invalid but the worklow ID is invalid (e.g. deleted workflow), then the Classifier displays blank content - THIS IS INTENDED. For Classrooms, it's better to error out than to direct students to an incorrect workflow (i.e. the default workflow)

External changes required:
- WildCam Darien Lab on Education API Front End needs to changes its Assignment URL template (Shaun to make changes)
- WildCam Darien needs to have `'wildcam classroom'` added to its Experimental Tools (done ✔️)

eatyourgreens:
@shaunanoordin will the changes here affect the astronomy classrooms too?

eatyourgreens:
Looks good to me but I will give it a closer look tomorrow.

eatyourgreens:
@shaunanoordin I'm going to make one small change, which is to add a check that there is a workflow present in the URL.

eatyourgreens:
@shaunanoordin in the last example, `workflow=11111111` the page is blank because it enters an infinite loop of constantly requesting the invalid workflow from Panoptes. I thought returning null would override the retry but obviously not. I'll look into this.

eatyourgreens:
@shaunanoordin this could do with another look. I've rewritten the code to avoid the infinite loop, which happened because `getWorkflow` will call `getSelectedWorkflow` if it can't find the workflow ID in the linked workflows for the project. `getSelectedWorkflow` will then call `getWorkflow` with the invalid workflow from the URL, so `getWorkflow` will call `getSelectedWorkflow` etc. etc. etc. I've updated the new component to call the `loadWorkflow` action just once when the component mounts.

I've also updated the tests and removed unused code that was leftover from the original tests.

eatyourgreens:
I think this demonstrates that small components, and simple functions that do one thing well, are a lot easier to work with than a complex component with functions that try to handle every possible case for workflow selection.

shaunanoordin:
👍 Cool, thanks Jim - give me a few minutes to run through my tests

eatyourgreens:
I changed the rule for classrooms to look for `classroom` and `workflow` in the URL, but that can be changed back.

srallen:
Sorry can someone please explain to me why the workflows couldn't just be active to begin with? None of the discussion above clearly states why they were inactive. My suggestion would have been to have the workflows be set to `active: true` when created by Edu API (by POSTING to Panoptes) and set them to be hidden from the stats page (a workflow config). It is my guess that they are inactive because they didn't want to be seen on the stats page of the project. This would have been a more simple fix that didn't require a new workflow selection strategy.

marten:
Wouldn't that mean they'd get randomly assigned to normal volunteers?

eatyourgreens:
@srallen my understanding was classrooms don't retire subjects, so the classifications aren't counted.

marten:
Oh, also if you make them active, they'll all appear in the `workflow.links.active_workflows` field when requested from the API. And it's a loooot of workflows.

srallen:
The Wildcam project has a default workflow? Retirement can be set to "never_retire" independently of activeness, which is what the I2A classrooms do. The serialization is a good point, though.

shaunanoordin:
@srallen Sorry, the original rationale was buried here in the issue #5256 :

> The workflows created by WildCam Assignments are inactive, to prevent Assignments-based Classifications from accidentally retiring Subjects too early. 

The original WildCam system was designed to keep the 'official' workflow separate from the school-created ones, since setting teacher-created workflows to 'active' may have a number of unwanted side effects to the official research.

Also - Darien's official workflow: 3033, Gorongosa's official workflow: 338.

eatyourgreens:
Darién has a default workflow. One other thing we wanted to do was override default error handling, so that students don't classify on the default workflow by mistake.

srallen:
The assignment workflows retirement can be set to `never_retire`. It's possibly Wildcam was setup before that retirement strategy was supported by Panoptes, but it can be used now. 

There's a banner that shows up for the query param for `group` used by I2A which validates that the user is a member of that user group. All classrooms are just Panoptes user groups. I had suggested that Wildcam use this for classroom membership validation as well and so that that UI component shows up. 

We now have two different ways that classroom workflows are rendered.

marten:
(Retirement is set to that for all workflows EduAPI makes, and has been that way from basically the start, so has always applied to WG:OG as well as Darien)

srallen:
Ok, so if retirement is set to `never_retire` then they could be set to be active? That isn't a reason to keep them inactive as far as I can tell. The serialization on the project is the only potential blocker I can see.

camallen:
We currently serialize all the workflows anyway via links, changing them to `active` would result in another probably expensive serializer query though. 
screenshot 2019-02-08 at 15 41 38

Personally i prefer the strategy pattern to simplify the logic of the workflow selection. The default selection code is burdened with logical flows that are hard to understand and debug.

srallen:
Yeah, I understand the serialization concern. The strategy approach is what we were planning on for the rewrite. We need to minimize changes here and we had a potential solution that did not involve modifying PFE code.

We now have two different ways that the workflows for classroom associated projects are being rendered and it's not clear which does what. Does Wildcam projects validate that the logged in user is a member of the group aka classroom? Does it POST classifications with the user group id in the metadata like I2A too (and is this necessary)? 

For the Edu API, we do have a distinction between "Wildcam like" and "Intro 2 Astro like" education programs, if there are truly differing requirements for the rendering of the workflow, then it should be documented. If not, then there should be consistency for projects that are associated with the Edu API.

PR #5270, opened 2019-02-12
Preview video formats other than mp4

### LINKED ISSUE #5269: Recent classification video thumbnail fail
## Expected behavior
The Recent page should show my latest classifications done in the project together with image/video thumbnails.
The video thumbnails are supposed to be playable on the Recents page, and they should display the speed option buttons.

## Current behavior
Here's a screenshot of what I get on the new C&S we're just 'building' where we classify videos:
https://panoptes-uploads.zooniverse.org/production/project_attached_image/f9b5b3cc-2222-4d10-8d80-02bfdf815b9b.png
The only thumbnail I see is an older classification. The others that failed loading come from today.
All four video files are m4a file types despite of the 'unknown file type' label over the three of them.
I have no troubles displaying my Recents on Battling Birds (video ones) or in other projects but here.
Also, you can notice the only loaded thumbnail in my screenshot is displayed as an image and not as a video thumbnail. Why is that so?
This is what my Recents look like on Battling birds:
https://panoptes-uploads.zooniverse.org/production/project_attached_image/811eb66e-62ba-421b-8567-764512eafd3a.png
These are playable videos - you can even choose what speed to play it with (out of the box though).

- **Operating system:**  Ubuntu
- **Browser:** Chromium 71

eatyourgreens:
Staging branch URL: https://pr-5270.pfe-preview.zooniverse.org

Fixes #5269 

Remove explicit  references to `.mp4` files from the code.
Pass a type prop to thumbnail previews, with value `image|audio|video`.
Add `x-m4v` to the list of allowed subject video formats when rendering subjects.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/components/thumbnail.jsx
+++ b/app/components/thumbnail.jsx
@@ -42,21 +42,21 @@ export default class Thumbnail extends React.Component {
       maxHeight: this.props.height
     };
 
-    if (this.props.format === 'mp4') {
+    if (this.props.type === 'video') {
       return (
         <div>
           <video style={style} controls={this.props.controls} onClick={this.playVideo}>
-            <source src={this.props.src} type="video/mp4" />
+            <source src={this.props.src} type={`video/${this.props.format}`} />
           </video>
         </div>
       );
     }
 
-    if (this.props.format === 'audio') {
+    if (this.props.type === 'audio') {
       return (
         <div>
           <audio style={style} controls={this.props.controls} onClick={this.playVideo}>
-            <source src={this.props.src} />
+            <source src={this.props.src} type={`audio/${this.props.format}`} />
           </audio>
         </div>
       );
@@ -82,10 +82,11 @@ Thumbnail.getThumbnailSrc = function getThumbnailSrc({ origin, width, height, sr
 
 Thumbnail.defaultProps = {
   controls: true,
-  format: 'image',
+  format: '',
   height: MAX_THUMBNAIL_DIMENSION,
   origin: 'https://thumbnails.zooniverse.org',
   src: '',
+  type: 'image',
   width: MAX_THUMBNAIL_DIMENSION
 };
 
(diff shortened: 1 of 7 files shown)


eatyourgreens:
See https://pr-5270.pfe-preview.zooniverse.org//projects/sassydumbledore/chimp-and-see-updated?env=production for the project that's affected by this bug.

Check other video projects to make sure that they still work too.

eatyourgreens:
Should we just drop MIME type sniffing when rendering subjects?

srallen:
I don't think so, but we could expand it maybe using the approach I did with the new Media component?

srallen:
FWIW, there are potential security issues with that as we discussed, but subject viewers still are designed around rendering based on mimetype, so we do have to know it at some point.

eatyourgreens:
Sorry, I meant should we be still filtering on file format, not type. One of the reasons that C&S are having problems is the `READABLE_FORMATS` filters in the code, so maybe we should just drop those?

shaunanoordin:
> ...should we be still filtering on file format, not type. One of the reasons that C&S are having problems is the `READABLE_FORMATS` filters in the code, so maybe we should just drop those?

I'm still trying to think if there is any reason why we'd need to keep a media-format whitelist in READABLE_FORMATS. AFAIK it's not as if our Thumbnails or Subject renderers can be coaxed into presenting malicious file types/file formats to users. 🤷‍♂️  (note to self: check on this)

shaunanoordin:
Wait, edit: I can think of one reason why we'd want to have the READABLE_FORMATS media whitelist, and that's if we want to enforce a 'media standard' across all projects. (Assumption: we clearly communicate this standard to the project owners.)

This media standard might be applicable if we're worried about project owners uploading, e.g. `.ogg` video/audio files that work fine in Chrome and FF, [but go kaput in Safari.](https://developer.mozilla.org/en-US/docs/Web/HTML/Supported_media_formats#Ogg_Theora_Vorbis)

PR #5274, opened 2019-02-15
Add subject frame duration slider

### LINKED ISSUE #5092: Control multi-image iteration speed
## Requested behavior
User can control speed that multi-image subject iterates through frames.
Maybe set options (similar to video playback), maybe slider.

## Current behavior
I believe [this code](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/subject-viewer.cjsx#L52-L53):
```
playFrameDuration: 667
playIterations: 3
```

mcbouslog:
Staging branch URL: https://pr-5274.pfe-preview.zooniverse.org/projects/markb-panoptes/backyard-worlds-test-project-mb/classify

Closes #5092.

Describe your changes.
- adds slider to control multi-image frame rate duration

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/components/subject-viewer.cjsx
+++ b/app/components/subject-viewer.cjsx
@@ -65,9 +65,10 @@ module.exports = createReactClass
   getInitialState: ->
     loading: true
     playing: false
-    playFrameDurationRate: 2
     frame: @getInitialFrame()
     frameDimensions: {}
+    frameDurationRate: 2
+    frameDurationIntervalId: null
     inFlipbookMode: @props.allowFlipbook
     promptingToSignIn: false
 
@@ -164,6 +165,8 @@ module.exports = createReactClass
                     className="secret-button"
                     onKeyDown={@keyDownFrameDuration.bind this, -0.25}
                     onMouseDown={@mouseDownFrameDuration.bind this, -0.25}
+                    onMouseUp={@clearFrameDurationInterval}
+                    onMouseOut={@clearFrameDurationInterval}
                     title="Slower"
                     type="button"
                   >
@@ -179,13 +182,15 @@ module.exports = createReactClass
                     onChange={@handleFrameDurationChange}
                     step="0.25"
                     type="range"
-                    value={@state.playFrameDurationRate}
+                    value={@state.frameDurationRate}
                   />
                   <button
                     aria-label="Faster"
                     className="secret-button"
                     onKeyDown={@keyDownFrameDuration.bind this, 0.25}
                     onMouseDown={@mouseDownFrameDuration.bind this, 0.25}
+                    onMouseUp={@clearFrameDurationInterval}
+                    onMouseOut={@clearFrameDurationInterval}
                     title="Faster"
                     type="button"
                   >
@@ -308,7 +313,7 @@ module.exports = createReactClass
       if @state.playing is on and (counter < flips or infiniteLoop is on)
         counter++
         @handleFrameChange (@state.frame + 1) %% totalFrames
-        setTimeout flip, (1000 / @state.playFrameDurationRate)
+        setTimeout flip, (1000 / @state.frameDurationRate)
         if counter is flips and infiniteLoop is off
           @setPlaying false
       else @setPlaying false
@@ -317,16 +322,23 @@ module.exports = createReactClass
       setTimeout flip, 0
 
   handleFrameDurationChange: (event) ->
-    @setState playFrameDurationRate: parseFloat(event.target.value)
+    @setState frameDurationRate: parseFloat(event.target.value)
 
   keyDownFrameDuration: (step, event) ->
     if event.which is 13 or event.which is 32
-      if 0.25 <= (@state.playFrameDurationRate + step) <= 10
-        @setState playFrameDurationRate: @state.playFrameDurationRate + step
+      @changeFrameDuration(step)
 
   mouseDownFrameDuration: (step) ->
-    if 0.25 <= (@state.playFrameDurationRate + step) <= 10
-      @setState playFrameDurationRate: @state.playFrameDurationRate + step
+    @clearFrameDurationInterval()
+    @setState frameDurationIntervalId: setInterval (=> @changeFrameDuration step), 100
+
+  changeFrameDuration: (step) ->
+    if 0.25 <= (@state.frameDurationRate + step) <= 10
+      @setState frameDurationRate: @state.frameDurationRate + step
+
+  clearFrameDurationInterval: () ->
+    if @state.frameDurationIntervalId
+      clearTimeout @state.frameDurationIntervalId
 
   handleFrameChange: (frame) ->
     @setState {frame}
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Looks good to me after a quick look but:
- it's odd that video and slideshows have different controls for changing the playback speed. I think the speed controls should really be the same.
- there's no labelling in the UI here to tell a volunteer that the range control speeds up or slows down the slideshow.

mcbouslog:
- agreed the different controls between video and image feels off, but the x.25, x.5 etc. we see on video, for multiple images I think would be based on an arbitrary frame duration we originally set that may or may not work well with a broad range of projects that use multiple images, where the slider gives a decent input range, and the Backyard World users would really like the more precise control
- maybe tooltips on hover for the > and >> noting "slower" and "faster" would help with labeling?
- I plan to do a new PR to save the duration selected for the next classification, maybe to user project preferences, but would like to get this in with the recent relaunch and media related to Backyard Worlds

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,24 @@ module.exports = createReactClass
                   <button aria-label="Play" title="Play" type="button" className="secret-button subject-tools__play" onClick={@setPlaying.bind this, true}>
                     <i className="fa fa-play fa-lg fa-fw"></i>
                   </button>}
+                <div className="subject-frame-duration">
+                  <i className="fa fa-angle-right fa-lg fa-fw subject-frame-duration--slower"></i>
+                  <input
```
eatyourgreens: The linter should be warning about this anyway, but this input is unlabelled. It should be something like `<label>Playback Speed <input /></label>` or `<input arial-label="Playback Speed" />`.

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,24 @@ module.exports = createReactClass
...
+                    aria-valuemin="-2000"
```
eatyourgreens: `aria-valuemin` is only needed on elements that don't support the `min` attribute.

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,24 @@ module.exports = createReactClass
...
+                    aria-valuemax="-100"
```
eatyourgreens: This is duplicating the `max` attribute and can be removed.

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,24 @@ module.exports = createReactClass
...
+                    aria-valuenow={@state.playFrameDuration * -1}
```
eatyourgreens: This is duplicating the `value` attribute and can be removed.

The value of this ranfe input doesn't make a great deal of sense to me. VoiceOver, for example, announces 'range: -600', 'range: -500' etc. when I change it and I'm thinking '-500 what?'.

Can it be changed to generate a value that makes sense in terms of playback speed? I'd suggest a range from 0.25 to 1.0, for consistency with our other speed controls, and then divide the frame duration by that fraction.

mcbouslog: Yes! This originally made my head spin a bit, but I like your suggestion a lot. I've updated accordingly while leaving increased options at higher speeds as requested by users. I think this range makes a lot more sense. Thanks!

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,26 @@ module.exports = createReactClass
                   <button aria-label="Play" title="Play" type="button" className="secret-button subject-tools__play" onClick={@setPlaying.bind this, true}>
                     <i className="fa fa-play fa-lg fa-fw"></i>
                   </button>}
+                <div className="subject-frame-duration">
+                  <button aria-label="Slower" title="Slower" type="button" className="secret-button" onClick={@setFrameDuration.bind this, -0.25}>
```
eatyourgreens: `onMouseDown`/`onKeyDown` might be better than `onClick` here, so that I can lower the speed by holding the button down or holding enter down. The keyboard handler would have to filter keys for enter/space.

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,26 @@ module.exports = createReactClass
...
+                    id="frame-duration"
+                    min="0.25"
+                    max="10"
+                    name="frame-duration"
+                    onChange={@handleFrameDurationChange}
+                    step="0.25"
+                    type="range"
+                    value={@state.playFrameDurationRate}
+                  />
+                  <button aria-label="Faster" title="Faster" type="button" className="secret-button" onClick={@setFrameDuration.bind this, 0.25}>
```
eatyourgreens: Ditto replacing `onClick` with `onMouseDown` and `onKeyDown`.

app/components/subject-viewer.cjsx
```diff
@@ -158,6 +158,26 @@ module.exports = createReactClass
                   <button aria-label="Play" title="Play" type="button" className="secret-button subject-tools__play" onClick={@setPlaying.bind this, true}>
                     <i className="fa fa-play fa-lg fa-fw"></i>
                   </button>}
+                <div className="subject-frame-duration">
+                  <button aria-label="Slower" title="Slower" type="button" className="secret-button" onClick={@setFrameDuration.bind this, -0.25}>
+                    <i className="fa fa-angle-right fa-lg fa-fw subject-frame-duration--slower"></i>
+                  </button>
+                  <input
+                    arial-label="Playback Speed Rate Adjustment"
```
eatyourgreens: Typo: aria-label

mcbouslog:
I'm not sure `onMouseDown` yields the expected behavior if click and hold, it just changes the value once, but with the slider and other options I think that's ok? Alternatively I could create a timeout that'd clear `onMouseUp` and `onMouseOut`.

eatyourgreens:
I'm fairly sure `onMouseDown` fires repeatedly as long as the mouse is down. Have a look at the code for the zoom buttons, they work while they are held down. At the moment, you have to click constantly on the buttons to change the slider, which is going to give our volunteers RSI.

mcbouslog:
No prob, I'll add the timeout and similar clears of the timeout as https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/pan-zoom.jsx#L81.

eatyourgreens:
How annoying (and inconsistent) that `onKeyDown` fires continuously while a key is down but `onMouseDown` does not. 😠

eatyourgreens:
It also occurred to me that if you need really fine control, the range control responds to the cursor keys when it has focus.

mcbouslog:
I was very surprised as well, as were many of the Stackoverflow or similar posts that came up as I was googling around about it! So at least we're not alone.

Yeah the slider itself gives a good amount of control, and a user can click the slider on the opposite end of the current value to quickly change it via mouse.

mcbouslog:
That feels better, thanks for your patience and suggestions @eatyourgreens!

eatyourgreens:
Feels a lot better to me too. Here's what I'm seeing in VoiceOver for the range control now. Previously it said '-600, slider' at the end.
Screenshot showing VoiceOver announcement for a labelled range control in Chrome.

PR #5278, opened 2019-02-18
hide subject intervention metadata

### LINKED ISSUE #5277: Interventions subject metadata is visible in UI
https://github.com/zooniverse/Panoptes-Front-End/blob/adebee3e88554afe006cf329f522ff6da6d7078b/app/redux/ducks/interventions.js#L57

Playing around with subject task allocation interventions and I'm wondering why we add the visible `intervention: true` metadata field to subjects? 

screenshot 2019-02-18 at 06 03 49

camallen:
avoid showing the intervention state of a subject in the UI

Staging branch URL: https://pr-5278.pfe-preview.zooniverse.org

Fixes #5277.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/redux/ducks/classify.spec.js
+++ b/app/redux/ducks/classify.spec.js
@@ -11,6 +11,16 @@ global.innerHeight = 1000;
 global.sessionStorage = new FakeLocalStorage();
 sessionStorage.setItem('session_id', JSON.stringify({ id: 0, ttl: 0 }));
 
+function mockSubject(id) {
+  return mockPanoptesResource('subjects', {
+    id,
+    locations: [],
+    metadata: {
+      intervention: true
+    },
+  });
+}
+
 describe('Classifier actions', function () {
   describe('add intervention', function () {
     const action = {
@@ -42,10 +52,16 @@ describe('Classifier actions', function () {
     });
   });
   describe('append subjects', function () {
+    const subjects = [
+        mockSubject('3'),
+        mockSubject('4'),
+        mockSubject('1'),
+        mockSubject('2')
+      ]
     const action = {
       type: 'pfe/classify/APPEND_SUBJECTS',
       payload: {
-        subjects: [1, 2],
+        subjects: [subjects[2], subjects[3]],
         workflowID: '1'
       }
     };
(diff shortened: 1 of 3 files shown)


eatyourgreens:
Failing tests because changing the field name breaks classifications.

I think a better change would be to delete that metadata key after the classification source has been set.

eatyourgreens:
I can't remember the exact client syntax for deleting keys but something like `subject.update('metadata.intervention', undefined)` might work after these two lines.
https://github.com/zooniverse/Panoptes-Front-End/blob/61217589c83861c3e5e372dffdc1b93c23838049/app/redux/ducks/classify.js#L49-L50

eatyourgreens:
@camallen I've fixed this and added some tests.

srallen:
Why aren't we using the `#` to hide this metadata?

eatyourgreens:
Because that breaks the new classification, so I'm just deleting that key instead.

srallen:
Hm, could we have a comment inline about it? I just can see coming across this in the future and not remembering why we're doing this/it seems project specific, etc.

eatyourgreens:
Should that be a comment on the `createNewClassification` function? Or on the `prependSubjectQueue` action?

srallen:
`createNewClassification` makes sense to me. Thanks!

eatyourgreens:
Good idea to record this by the way. The first conversation @camallen and I had about #5277 was 'Why is that there anyway?'.

app/redux/ducks/classify.spec.js
```diff
@@ -11,7 +11,17 @@ global.innerHeight = 1000;
...
+  return mockPanoptesResource('subjects', {
+    id,
+    locations: [],
+    metadata: {
+      intervention: true
+    },
+  });
+}
+
+describe.only('Classifier actions', function () {
```
eatyourgreens: Stray `.only`

eatyourgreens:
@camallen does this fix the issue?

camallen:
> @camallen does this fix the issue?

Code looks good and manual testing on the deployed branch works 👍 

screenshot 2019-03-08 at 10 51 54

camallen:
Sadly i can't approve my own PR even though you've done all the work :(

Issue #5279, opened 2019-02-19
workflow test button page broken


lfortson:
## Expected behavior
Clicked on test workflow at end of workflow set up 

## Current behavior
_Please include any error messages from the browser console and/or screenshots_
There was an error retrieving project lucyitswd/cssitest.

undefined is not an object (evaluating 'n.type')

    in t
    in div
    in t
    in styled.div
    in t
 ....

## Steps to replicate

I did not add any tasks to the workflow, but in my defense, I got to the bottom of the screen and the next obvious thing to do (as a newbie) was to click "test workflow"

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Was this fixed by #5280?

srallen:
Sort of. It's disabling the 'test this workflow' button if there are **no** tasks, but doesn't fix the classifier for some kinds of tasks if they are partially setup. For example, if a survey task is added, but has no choices, the test button will be enabled and work, but the loaded classifier will error. This is because we don't have defaults for several expected attributes still. This is I think only an issue for survey task and drawing tasks.

I think for survey task and drawing task we might want to display a more useful error like "No choices setup" or something on the classifier page as well as make improvements to the lab UI, but that's long term.

Perhaps this issue can be closed and a new one be opened for improved error handling for partially setup workflow tasks?

eatyourgreens:
Maybe tasks should have a static validation function that validates the task object against some simple validation rules? Then the editor could check `TaskComponent.isValid(taskDescription)` or similar. The function could return a helpful error message specific to that task if the task is invalid.

srallen:
This would be easier to do with the models we're defining with the MST stores.

PR #5280, opened 2019-02-19
Disable test workflow button if no tasks are defined

srallen:
Staging branch URL:

Toward #5279.

Describe your changes.
This disables the test workflow button if no tasks are defined. The drawing task and survey task still break the classifier if no drawing tools or none of the survey task items are defined, but doing validations/default values for those are a bit more involved and should be fixed individually.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -442,17 +442,20 @@ EditWorkflowPage = createReactClass
             </div>}
 
           <div className={if @isThereNotADefinedTask() then 'disabled-section' else ''}>
-            <a
-              disabled={@isThereNotADefinedTask()}
-              href={@workflowLink()}
-              className="standard-button"
-              target="from-lab"
-              onClick={@handleViewClick}
-            >
-              Test this workflow
-            </a>
+            {if not @isThereNotADefinedTask()
+              <a
+                href={@workflowLink()}
+                className="standard-button"
+                target="from-lab"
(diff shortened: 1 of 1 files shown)


app/pages/lab/workflow.cjsx
```diff
@@ -437,8 +441,18 @@ EditWorkflowPage = createReactClass
 
             </div>}
 
-          <div>
-            <a href={@workflowLink()} className="standard-button" target="from-lab" onClick={@handleViewClick}>Test this workflow</a>
+          <div className={if @isThereNotADefinedTask() then 'disabled-section' else ''}>
+            <a
+              disabled={@isThereNotADefinedTask()}
```
eatyourgreens: Links can't be disabled by adding a `disabled` attribute. Render the link's content as plain text if you don't want it to be clickable, or just don't render the link at all if you don't want project owners to have access to it.

srallen: Right --- confusion ensues when we style links link buttons. I do want them to know it's possible to do this and why they can't.

eatyourgreens: Wrap the text in a span or div that's styled to look like a disabled button? One massive bug in the project builder at the moment is that you can tab to, and activate, inputs and links that are styled as disabled but not actually disabled in the DOM.

PR #5287, opened 2019-03-01
Add form link to publications page (Mk II)

shaunanoordin:
## PR Overview

This PR adds a small bit of text to the [Publications](https://www.zooniverse.org/about/publications) page, providing a small bit of extra info to users. 

As described in the original PR, the intention of this update is:
> Short text & form link to the top of publications page for easy access.

Replaces #5285, by fixing the layout and hooking up the intended text to our language translatomatics.

screen shot 2019-03-01 at 16 56 44
_The "Submit New Publication" text appears above the list of publications._
_Note: the tops of the sidebar and main section don't align no more because the new text is in a Markdown block that adds unnecessary margins. I am ashamed_

**Collateral Updates:**
- All the locale files were updated so the structure of the `pages.about.publications` section of every language matches that of the English's.
  - Basically, don't worry that a lot of locale files were updated with mundane fixes unrelated to the "Submit New Publication" text. _Fuhgeddaboudit._

### Status
This is ready for review. Tagging @snblickhan as the OG PR originator.

### DIFF
--- a/app/locales/ar.js
+++ b/app/locales/ar.js
@@ -268,10 +268,11 @@ export default {
         header: {
           showAll: 'All Publications'
         },
-        publication: {
-          viewPublication: 'View publication.',
-          viewOpenAccess: 'View open access version.'
-        }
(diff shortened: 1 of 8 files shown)


eatyourgreens:
Not a blocking comment, but having the initial paragraph after the page heading, rather than before it, would be a more logical document structure.

shaunanoordin:
## PR Update

> Not a blocking comment, but having the initial paragraph after the page heading, rather than before it, would be a more logical document structure.

Dangit, that makes way more sense!

screen shot 2019-03-01 at 17 20 32
_Updated text position_

eatyourgreens:
How did a text change increase test coverage?
🤔

Issue #4727, opened 2018-07-04
Notifications don't update when you have unread messages


eatyourgreens:
## Expected behavior
When I have unread messages in a project I participate in, I should be notified in the top-level navigation bar.

## Current behavior
From Talk:
>Hello Zooniverse team: is anyone looking into this problem? Beginning Friday 29 June, when you have unread notifications, the "notifications" link/icon in the Zooniverse menu bar is no longer highlighted an no longer shows the number of unread messages when you are on Zooniverse Talk or Zooniverse Projects, although these indicators of unread messages still appear when you are on in-Project pages. Is this an intentional change that we just have to live with, or did a bug creep in? I would really like to have the former consistent behavior across all pages on the site reinstated.
>** or maybe beginning thursday 28th

## Additional information

- **Operating system:**
- **Browser:**

eatyourgreens:
Noting that the cause of this seems to be a null query parameter `section=` in the API request to get unread messages. #4728 removes the section parameter from the query if it isn't defined.

yshish:
I'm still experiencing that, no notification when someone replies to my posts on a Talk. 
Not sure it will help but here's a screenshot of my browser console made at the moment when I came to Zooniverse, after receiving an email notification about a mentioning of my user name in a post reply on Zooniverse Talk:
![screenshot_2018-07-07_16-55-59](https://user-images.githubusercontent.com/7211888/42412177-6bdd30ae-8207-11e8-94f1-6cf39483f8b4.png)
(Using Chromium on latest Ubuntu)

yshish:
(I have just received a personal message via Zooniverse and I did receive the notification for messages, but not for the project Talk replies.)

yshish:
Hah, after 25 minutes, I suddenly did get 2 notifications displayed in the top-nav bar. (I think it is the first time in ages that happened.) Anyway, I reloaded the Zooniverse main page without either clicking on the notifications or visiting the project Talk via email, but when the page reloaded, the number of notifications displayed 1 (instead of 2). It doesn't make sense, does it? And after another reload, the notifications display no number at all.

eatyourgreens:
I think this is still a problem. There’s a discussion on Talk about problems with notifications. https://www.zooniverse.org/talk/17/735435

camallen:
What was the reasoning to mark notifications as read when the component unmounts / pages through the notifications? 
[...]

PR #5294, opened 2019-03-13
Refactor notifications, keep unread

mcbouslog:
Staging branch URL: https://pr-5294.pfe-preview.zooniverse.org/

Towards #4727, specifically [this comment](https://github.com/zooniverse/Panoptes-Front-End/issues/4727#issuecomment-430651987).

Let me know if you'd like a few notifications on staging, I have multiple testing accounts, I can create notifications while you're testing to see if/how count updates.

Describe your changes.
- remove project context (`section`) of notification counter
    - counter is in global header nav, shouldn't be project specific, looks strange when changes as navigate around site
- remove marking notifications as read if shown
    - the notification property `delivered` is meant for if related email sent, marking as true when shown on notifications page (without user clicking anything) has caused some confusion, leaving as `false` might help

related Talk discussions [735435](https://www.zooniverse.org/talk/17/735435) and [917872](https://www.zooniverse.org/talk/17/917872)

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -53,12 +53,12 @@ PanoptesApp = createReactClass
     auth.listen 'change', @handleAuthChange
     @handleAuthChange()
     generateSessionID()
+  
+  componentDidUpdate: ->
+    @updateNotificationsCount()
 
   componentWillUnmount: ->
     auth.stopListening 'change', @handleAuthChange
-
-  componentWillReceiveProps: (nextProps) ->
-    @updateNotificationsCount params: nextProps.params
 
   handleAuthChange: ->
     @geordiLogger.forget ['userID']
@@ -66,14 +66,12 @@ PanoptesApp = createReactClass
       @setState
         initialLoadComplete: true
         user: user
-      @updateNotificationsCount {user}
+      @updateNotificationsCount user
       @geordiLogger.remember userID: user.id if user?
 
-  updateNotificationsCount: ({user, params}) ->
+  updateNotificationsCount: (user) ->
     user or= @state.user
-    params or= @props.params
-    {owner, name} = params
-    @props.notificationsCounter.update(user, owner, name)
+    @props.notificationsCounter.update(user)
 
   render: ->
     <div className="panoptes-main">
--- a/app/partials/app.cjsx
+++ b/app/partials/app.cjsx
@@ -54,8 +54,9 @@ PanoptesApp = createReactClass
     @handleAuthChange()
     generateSessionID()
   
-  componentDidUpdate: ->
-    @updateNotificationsCount()
+  componentDidUpdate: (prevProps) ->
+    if prevProps.params isnt @props.params 
+      @updateNotificationsCount()
 
   componentWillUnmount: ->
     auth.stopListening 'change', @handleAuthChange
--- a/app/lib/notifications-counter.js
+++ b/app/lib/notifications-counter.js
@@ -1,19 +1,11 @@
-import apiClient from 'panoptes-client/lib/api-client';
 import talkClient from 'panoptes-client/lib/talk-client';
 
 class NotificationsCounter {
-  static getProject(slug) {
-    return apiClient.type('projects').get({ slug });
-  }
-
-  static getNotifications(section) {
+  static getNotifications() {
     const query = {
       delivered: false,
       page_size: 1
     };
-    if (section) {
-      query.section = section;
-    }
     return talkClient.type('notifications').get(query);
   }
 
@@ -21,7 +13,7 @@ class NotificationsCounter {
     this.callbacks = [];
   }
 
-  update(user, owner, name) {
+  update(user) {
     let unreadCount = 0;
     if (!user) {
       return Promise.resolve(unreadCount);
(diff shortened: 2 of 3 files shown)


app/partials/app.cjsx
```diff
@@ -53,27 +53,25 @@ PanoptesApp = createReactClass
     auth.listen 'change', @handleAuthChange
     @handleAuthChange()
     generateSessionID()
+  
+  componentDidUpdate: ->
```
mcbouslog: This might be a bad idea.

camallen: Did you test how often this callback is run when using the system? I'm wondering especially during classifying and how that might increase the talk api traffic? 

TBH getting notifications updates more often seems really useful, i'm just wondering about API calls and their frequency...

eatyourgreens: Just noting that this code is currently in `componentWillReceiveProps`, which fires more often than `componentDidUpdate`. [`componentWillReceiveProps` is being removed in React 17](https://reactjs.org/docs/react-component.html#unsafe_componentwillreceiveprops), so this looks good to me.

eatyourgreens: On second thoughts, why aren't these notifications using the sugar client?

wgranger: If this is the correct staging branch https://pr-5294.pfe-preview.zooniverse.org/ I think this line has caused an infinite loop. Check out the networks tab. I'm guessing the call to `notificationsCounter.update(user)` causes the component to update and make the call again. 

If that is the issue, you could add a condition to only make the call when the route changes, or something similar.

eatyourgreens: Good point. `componentDidUpdate` will run after changes are batched and executed, including calls to `setState`. Any code inside `componentDidUpdate` should be guarded to avoid running on every update to state or props.

My preferred solution would be to use the sugar client to listen for notifications. The code is pretty much set up in the redux directory.

Second best solution would be to listen for notification changes on a time. Maybe [setInterval](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setInterval) to update the counter on a fixed interval.

mcbouslog: I'd like to try using the sugar client, that just makes sense. I'll make some time, but until then I'm not sure what to do with this PR. I'm leaning towards closing for now, then will reopen once utilize the sugar client or setInterval. I'll prob do that (close for now) towards the end of today, unless I hear otherwise. 

A big thanks for comments though! We're a little closer to some much requested functionality.

mcbouslog:
Closing pending refactor to utilize sugar or setInterval.

mcbouslog:
Reopening per Talk/Troubleshooting activity (https://www.zooniverse.org/talk/17/735435 and https://www.zooniverse.org/talk/17/917872).
- added conditional in `componentDidUpdate` to only update notifications on params change

@wgranger - please confirm no longer infinite loop
@eatyourgreens - apologies as not ideal sugar or interval implementation, which I actually was planning to try this Wednesday (scheduled hack day), but I thought maybe we can get this in asap then I can implement one of those (sugar or interval) on Wednesday? When you mentioned the sugar code is pretty much set up in the redux directory you're referring to the interventions code?

eatyourgreens:
`sugarClient.on('notification', callback)` should run `callback` whenever a notification is received. The inteventions code has a couple of actions defined for subscribing/unsubscribing to channels too. Channels is the bit of Sugar that I'm still unclear about but I _think_ you want notifications on your personal user channel. I also _think_ that the client subscribes to that channel when you log in but I'm guessing so don't take my word for it.

wgranger:
Yep, that params check solved the infinite loop bug. The notifications endpoint is being hit on route changes. We aren't hitting the endpoint while classifying, so I don't think we should be worried about increased traffic in the classifier.

camallen:
What's left on this except integration with Sugar? 

The API looking calls have been fixed, the code looks good. I've asked am.zooni and yshish for thoughts if they have time on the new behavior. @wgranger If there is nothing blocking here can we approve?

mcbouslog:
I think we're all on the same page, but just to be clear I think this PR should be merged as is (pending Will's approval and user feedback) and the refactor to use the sugar client will be a different PR.

wgranger:
Yes. This is good to go. Merged.

camallen:
Am.zooni tested the staged branch and verified the new behaviour. https://www.zooniverse.org/talk/17/735435?comment=1555209

> the changes appear to be working as intended. Details:
>
>I am using my main (am.zooni) account; created a Talk thread in NfN. I checked for the various types of "notification signals" with both standard Zooniverse URLs and https://pr-5294.pfe-preview URLs.
>
> When I submitted comments from my test account to that NfN thread, my main account's Notifications link in the nav bar is highlighted with number, and the number and yellow star appears in the notifications list - these signals appear on both NfN Talk and Zooniverse Talk when I use the standard zooniverse URL and the https://pr-5294.pfe-preview URL. The nav bar signals do not appear when I open Mapping Change Talk with the standard URL BUT the nav bar signal DOES appear in Mapping Change Talk (for an NfN notification) when I use the https://pr-5294.pfe-preview URL.

PR #5301, opened 2019-03-28
Update notification to delivered onClick

mcbouslog:
Staging branch URL: https://pr-5301.pfe-preview.zooniverse.org

Fixes issue noted in Talk/Troubleshooting - https://www.zooniverse.org/talk/17/735435?comment=1555431.

Updates notification delivered attribute to true when click link to related resource.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/notifications/notification-section.jsx
+++ b/app/pages/notifications/notification-section.jsx
@@ -32,30 +32,29 @@ export default class NotificationSection extends Component {
       });
     } else {
       apiClient.type('projects').get(this.props.projectID, { include: 'avatar' })
-      .catch((error) => {
-        this.setState({ error });
-      })
-      .then((project) => {
-        if (project.links.avatar) {
-          apiClient.type('avatars').get(project.links.avatar.id)
-          .then((avatar) => {
-            this.setState({
-              name: project.display_name,
-              avatar: avatar.src
-            })
-            .catch(() => {
-              this.setState({
-                name: project.display_name
+        .catch((error) => {
+          this.setState({ error });
+        })
+        .then((project) => {
+          if (project.links.avatar) {
+            apiClient.type('avatars').get(project.links.avatar.id)
+              .catch(() => {
+                this.setState({
+                  name: project.display_name
+                });
+              })
+              .then((avatar) => {
+                this.setState({
+                  name: project.display_name,
+                  avatar: avatar.src
+                });
               });
+          } else {
+            this.setState({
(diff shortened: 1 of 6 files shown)


app/pages/notifications/notification-section.jsx
```diff
@@ -32,30 +32,29 @@ export default class NotificationSection extends Component {
...
+          this.setState({ error });
+        })
+        .then((project) => {
+          if (project.links.avatar) {
+            apiClient.type('avatars').get(project.links.avatar.id)
+              .catch(() => {
+                this.setState({
+                  name: project.display_name
+                });
+              })
```
eatyourgreens: Won't this catch block crash the following then block, because it doesn't pass an avatar into it?

eatyourgreens: I'm not really sure what this block is doing because it's setting state, but the then block sets state again.

If it's supposed to catch errors in the API request, it should return a default avatar for the setState call in the then block.

mcbouslog: I moved the catch below the then block, similar to before, though before I was getting linter/console warnings.

mcbouslog:
Refactored onClick to update counter in nav and mark related notifications as read, which is much better UX.

app/pages/notifications/notification-section.jsx
```diff
@@ -140,6 +139,19 @@ export default class NotificationSection extends Component {
     });
   }
 
+  markAsRead(readNotification) {
+    const { notificationsCounter } = this.context;
+    const { user } = this.props;
+    const { notifications } = this.state;
+    const relatedNotifications = notifications.filter(
```
camallen: I think i'm just missing something but why are we updating the related notifications here? 

How is the notification i just read linked to another and why do i want to mark that as 'read' as well?

eatyourgreens: Is this the case where I click on one message in a discussion so I've read all the preceding new messages too, without having to go back up the page and individually click each one.

I often get a series of notifications for the same discussion, but only click through to one of them in order to read them all.

mcbouslog: Good question and I'm not tied to current code, but intention is if you have 3 notifications related to the same discussion, clock notification to go to that discussion, it'd mark all the related notifications as read as they'd have likely been shown in the discussion page you're now on.

PR #5304, opened 2019-03-29
Fix notification-section markAsRead

mcbouslog:
Staging branch URL:

I hope fixes #5301, as currently notifications still not marked as "read" (`delivered: true`) after clicking to related source.

Describe your changes.
- binds `markAsRead`
- refactor `markAsRead`, including filtering on `subscription_id`, though imperfect, I think best option that applies to all sources (i.e. comment, moderation, data request, etc.)
- [x] tests

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/notifications/notification-section.jsx
+++ b/app/pages/notifications/notification-section.jsx
@@ -23,6 +23,7 @@ export default class NotificationSection extends Component {
       page: 1
     };
     this.markAllRead = this.markAllRead.bind(this);
+    this.markAsRead = this.markAsRead.bind(this);
   }
 
   componentWillMount() {
@@ -143,13 +144,22 @@ export default class NotificationSection extends Component {
     const { notificationsCounter } = this.context;
     const { user } = this.props;
     const { notifications } = this.state;
-    const relatedNotifications = notifications.filter(
-      notification => notification.source_id === readNotification.source_id
-    );
-    relatedNotifications.push(readNotification);
-    const readRequests = relatedNotifications
-      .map(relatedNotification => relatedNotification.update({ delivered: true }).save());
-    Promise.all(readRequests).then(() => notificationsCounter.update(user));
+    const relatedNotifications = notifications.filter(notification => (
+      notification.delivered === false
+        && (notification.id === readNotification.id
+          || notification.source.discussion_id === readNotification.source.discussion_id)
+    ));
+    const notificationPromises = relatedNotifications
+      .map((relatedNotification) => {
+        return relatedNotification
+          .update({ delivered: true })
+          .save()
+          .catch((error) => {
+            console.warn(error);
+            return {};
+          });
+      });
+    Promise.all(notificationPromises).then(() => notificationsCounter.update(user));
   }
 
   avatarFor() {
--- a/app/pages/notifications/notification-section.spec.js
+++ b/app/pages/notifications/notification-section.spec.js
@@ -93,4 +93,77 @@ describe('Notification Section', function() {
       assert.equal(wrapper.find('.fa-chevron-up').length, 1);
     });
   });
+
+  describe('will update notifications as read', function () {
+    const newNotifications = [
+      {
+        id: '123',
+        delivered: false,
+        source: {
+          discussion_id: '456'
+        },
+        source_type: 'Comment',
+        update: sinon.stub().returnsThis(),
+        save: sinon.stub().resolves({})
+      },
+      {
+        id: '124',
+        delivered: false,
+        source: {
+          discussion_id: '456'
+        },
+        source_type: 'Comment',
+        update: sinon.stub().returnsThis(),
+        save: sinon.stub().rejects()
+      },
+      {
+        id: '125',
+        delivered: false,
+        source_type: 'Moderation',
+        source: {},
+        update: sinon.stub().returnsThis(),
+        save: sinon.stub().resolves({})
+      },
+      {
+        id: '126',
+        delivered: false,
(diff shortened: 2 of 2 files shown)


app/pages/notifications/notification-section.jsx
```diff
@@ -143,13 +144,13 @@ export default class NotificationSection extends Component {
...
-    relatedNotifications.push(readNotification);
-    const readRequests = relatedNotifications
+    const relatedNotifications = notifications.filter(notification => (
+      notification.subscription_id === readNotification.subscription_id
+        && notification.delivered === false
+    ));
+    const promises = relatedNotifications
       .map(relatedNotification => relatedNotification.update({ delivered: true }).save());
-    Promise.all(readRequests).then(() => notificationsCounter.update(user));
+    Promise.all(promises).then(() => notificationsCounter.update(user));
```
eatyourgreens: Are the promises in `promises` guaranteed to resolve? `Promise.all` will reject if any of them reject and there's no catch block to handle that case. I'm not familiar with the notifications code, but my guess is you want to update the notifications counter after all the API requests complete, even if some failed.

app/pages/notifications/notification-section.jsx
```diff
@@ -143,13 +144,13 @@ export default class NotificationSection extends Component {
...
-    const relatedNotifications = notifications.filter(
-      notification => notification.source_id === readNotification.source_id
-    );
-    relatedNotifications.push(readNotification);
-    const readRequests = relatedNotifications
+    const relatedNotifications = notifications.filter(notification => (
+      notification.subscription_id === readNotification.subscription_id
+        && notification.delivered === false
+    ));
+    const promises = relatedNotifications
```
eatyourgreens: `promises` doesn't tell us what this variable is doing. Is there a better name that could be used here?

app/pages/notifications/notification-section.jsx
eatyourgreens: The [MDN example of fail-fast behaviour](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/all#Promise.all_fail-fast_behaviour) suggests you could add catch blocks to the individual requests too.

app/pages/notifications/notification-section.jsx
```diff
@@ -143,13 +144,13 @@ export default class NotificationSection extends Component {
     const { notificationsCounter } = this.context;
     const { user } = this.props;
     const { notifications } = this.state;
-    const relatedNotifications = notifications.filter(
-      notification => notification.source_id === readNotification.source_id
-    );
-    relatedNotifications.push(readNotification);
-    const readRequests = relatedNotifications
+    const relatedNotifications = notifications.filter(notification => (
+      notification.subscription_id === readNotification.subscription_id
```
camallen: I think this needs to incorporate the `source_id` and `source_type` as well. I'm not sure the `subscription_id` matches across different users. 

I'm not entirely sure which attributes you need to 'relate' the notifications together but my testing resulted in the above attributes. I think you need to handle @ mentions, comments and other forms of notification events.

eatyourgreens:
I think `NotificationsCounter` does update the counts, it's just hard to follow because it uses context.
https://github.com/zooniverse/Panoptes-Front-End/blob/cf0e6264aa067a53f9cd326ceba4e73a933e4e16/app/talk/lib/notifications-link.cjsx#L21-L26

camallen:
Thanks @eatyourgreens context + callback on the `notifications-counter` was easy to miss.

eatyourgreens:
React is designed specifically to avoid using event listeners, so it's easy to miss how these counts work. Depending on how much effort we want to put into this, the unread count and associated actions could move out of app state and into a redux store.

mcbouslog:
Sharing [this gist](https://gist.github.com/mcbouslog/37f9c33b1ba12497e7d0b8a184f7765b) with a comment, reply, mention, moderation report and subject note (also a comment) used to attempt to determine "related notifications."

app/pages/notifications/notification-section.jsx
mcbouslog: Based on [this gist](https://gist.github.com/mcbouslog/37f9c33b1ba12497e7d0b8a184f7765b) I think comments, replies and mentions in the same discussion should be marked as read if the user goes to said discussion, otherwise I think other notifications shouldn't be grouped (moderation reports, data requests) but marked as read individually. This is just my opinion though, of course open to input.

app/pages/notifications/notification-section.jsx
```diff
@@ -143,13 +144,22 @@ export default class NotificationSection extends Component {
...
-    );
-    relatedNotifications.push(readNotification);
-    const readRequests = relatedNotifications
-      .map(relatedNotification => relatedNotification.update({ delivered: true }).save());
-    Promise.all(readRequests).then(() => notificationsCounter.update(user));
+    const relatedNotifications = notifications.filter(notification => (
+      notification.delivered === false
+        && (notification.id === readNotification.id
+          || notification.source.discussion_id === readNotification.source.discussion_id)
+    ));
```
mcbouslog: `notification.id === readNotification.id` captures notifications for moderation, data requests, whereas `notification.source.discussion_id === readNotification.source.discussion_id` captures comments (which includes replies and mentions) related to a clicked comment notification

camallen: LGTM

mcbouslog:
Related to #4727

PR #5318, opened 2019-04-04
Add a workflow selection strategy for the workflow query param

eatyourgreens:
Staging branch URL: https://pr-5318.pfe-preview.zooniverse.org

Prompted by #5125, this adds a new workflow selection strategy to use when a workflow query param is present, and removes those rules from the default selection strategy component. I've also added a note to `WorkflowSelection` saying that it shouldn't be overloaded with new rules any more. Rather, add new strategies using this one or classroom selection as examples.

I've also updated the classify page component to add a simple strategy selector, which uses the query param strategy if a query param is present. If workflow ID validation fails, the query param is removed from the location, which was the existing behaviour and, I think, will fall back to loading the project default workflow.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
Testing this on Gravity Spy, I can't upgrade myself to a new level by adding a workflow param to the URL, so that's good.

PR #5338, opened 2019-05-09
[RFC] Add active workflow filter on completeness

mcbouslog:
Staging branch URL: https://pr-5338.pfe-preview.zooniverse.org
- [staging project with 4 active workflows, 2 incomplete, 2 complete](https://pr-5338.pfe-preview.zooniverse.org/projects/markb-panoptes/completed-workflows)
- [Notes from Nature project](https://www.zooniverse.org/projects/zooniverse/notes-from-nature) (many workflows have been deactivated, linking not for testing code, but for reference)

Ideally all of the NfN workflows would be active as is intended, so classification, retired and subject stats are accurate. But there are 100+ workflows, which causes backend issues and having 100+ workflow buttons to choose from on the now being used project landing page is not a preferable option (formerly custom frontend, now part of [org](https://www.zooniverse.org/organizations/md68135/notes-from-nature)), especially with 95% of the workflows complete.

One possible initial solution is this PR, until a project page redesign and maybe a completed projects page can be implemented. With this PR, if a project wants to hide completed workflows, but still link to them, we can enable the "allow workflow query" feature in addition to this new "hide completed workflows" feature and they can list and link to the completed workflows somewhere in the project's About pages.

@camallen - is "manually updating the workflows to not serialize as links" an option for these issues with NfN? Or would the not serialized active workflows not count towards stats (and therefore wouldn't entirely help here)?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/admin/project-status/experimental-features.jsx
+++ b/app/pages/admin/project-status/experimental-features.jsx
@@ -25,7 +25,8 @@ const experimentalFeatures = [
   'general feedback',
   'slider',
   'highlighter',
-  'translator-role'
+  'translator-role',
+  'hide completed workflows'
 ];
 
 class ExperimentalFeatures extends Component {
--- a/app/pages/admin/project-status/experimental-features.jsx
+++ b/app/pages/admin/project-status/experimental-features.jsx
@@ -25,8 +25,7 @@ const experimentalFeatures = [
   'general feedback',
   'slider',
   'highlighter',
-  'translator-role',
-  'hide completed workflows'
+  'translator-role'
 ];
 
 class ExperimentalFeatures extends Component {
--- a/app/pages/project/home/index.jsx
+++ b/app/pages/project/home/index.jsx
@@ -91,9 +91,16 @@ export default class ProjectHomeContainer extends React.Component {
   }
 
   render() {
+    let shownWorkflows = [];
+    if (this.props.project.experimental_tools.includes('hide completed workflows')) {
+      shownWorkflows = this.state.activeWorkflows.filter(workflow => workflow.completeness !== 1);
+    } else {
+      shownWorkflows = this.state.activeWorkflows;
+    }
+
     return (
       <ProjectHomePage
-        activeWorkflows={this.state.activeWorkflows}
+        activeWorkflows={shownWorkflows}
         background={this.props.background}
         organization={this.props.organization}
         preferences={this.props.preferences}
--- a/app/pages/project/home/index.jsx
+++ b/app/pages/project/home/index.jsx
@@ -36,11 +36,12 @@ export default class ProjectHomeContainer extends React.Component {
     }
   }
 
-  fetchAllWorkflows(props, query) {
+  fetchWorkflows(props, query) {
     if (this.state.activeWorkflows.length === 0) {
       getWorkflowsInOrder(props.project, query)
-        .then((activeWorkflows) => {
-          this.setState({ activeWorkflows });
+        .then((allActiveWorkflows) => {
+          const activeIncompleteWorkflows = allActiveWorkflows.filter(workflow => workflow.completeness !== 1);
+          this.setState({ activeWorkflows: activeIncompleteWorkflows });
         });
     }
   }
@@ -84,23 +85,16 @@ export default class ProjectHomeContainer extends React.Component {
 
     if ((props.project.configuration && props.project.configuration.user_chooses_workflow && !workflowAssignment) ||
       (workflowAssignment && props.user)) {
-      this.setState({ showWorkflowButtons: true }, this.fetchAllWorkflows.bind(this, this.props, { active: true, fields: 'active,completeness,configuration,display_name' }));
+      this.setState({ showWorkflowButtons: true }, this.fetchWorkflows.bind(this, this.props, { active: true, fields: 'active,completeness,configuration,display_name' }));
     } else {
       this.setState({ showWorkflowButtons: false });
     }
   }
 
   render() {
-    let shownWorkflows = [];
-    if (this.props.project.experimental_tools.includes('hide completed workflows')) {
-      shownWorkflows = this.state.activeWorkflows.filter(workflow => workflow.completeness !== 1);
-    } else {
-      shownWorkflows = this.state.activeWorkflows;
-    }
-
     return (
       <ProjectHomePage
-        activeWorkflows={shownWorkflows}
+        activeWorkflows={this.state.activeWorkflows}
         background={this.props.background}
         organization={this.props.organization}
         preferences={this.props.preferences}


app/pages/admin/project-status/experimental-features.jsx
```diff
@@ -25,7 +25,8 @@ const experimentalFeatures = [
   'general feedback',
   'slider',
   'highlighter',
-  'translator-role'
+  'translator-role',
+  'hide completed workflows'
```
eatyourgreens: I'm inclined to make this the default behaviour for all projects, and not hide it behind an experimental feature flag.

app/pages/project/home/index.jsx
```diff
@@ -91,9 +91,16 @@ export default class ProjectHomeContainer extends React.Component {
   }
 
   render() {
+    let shownWorkflows = [];
+    if (this.props.project.experimental_tools.includes('hide completed workflows')) {
+      shownWorkflows = this.state.activeWorkflows.filter(workflow => workflow.completeness !== 1);
```
eatyourgreens: Filtering this array on every call to render is very inefficient. I think it would be much faster to filter the array once, and store the filtered array in component state.

camallen:
> @camallen - is "manually updating the workflows to not serialize as links" an option for these issues with NfN? 

That depends, marking those workflows that way will basically hide them from the project resource. If you want all the finished workflows to be available in the lab then this setting will break that. We can make a new relation just for the lab that can provide these workflows though that requires lab work etc (rebuild effort maybe?)

> Or would the not serialized active workflows not count towards stats (and therefore wouldn't entirely help here)?

They would not count for the project stats once they are marked this way. basically they don't exist on the project anymore, i.e. `education-api` use case the workflow is spawned from a project but is accessed independent of the project resource. Happy to help / clarify any of this.

srallen:
We could make this the default behavior for all projects. For NfN since they allow loading via query param, they could link to the completed projects somewhere on a page or on talk f they want them still accessible? We should come up with a quick solution for rendering links for completed workflows somewhere... maybe on the results page?

mcbouslog:
You're absolutely right @eatyourgreens:

> The problem with this approach is that you're filtering the API response, so you're requesting 100+ workflows in order to only show 5 or 6, if I'm reading the code correctly. It would be better if the filter on completed state could happen as part of the initial request query.

Trying to think about next steps, maybe this is where we're at - so when a project allows users to choose from workflows,
1. is showing completed workflows as options an issue we're going to address in the short-term?
2. if yes, 
  a. do we try to solve it in Panoptes with a filter on completed state (I'm unsure of difficulty, though I think preferable)?
  b. do we solve it with this PR's approach in PFE? (not preferable, but if big enough issue and Panoptes solution not tenable, maybe this)

Mentioning @trouille

mcbouslog:
Workflows are continuing to complete on the old NfN project, which still has a handful of workflows in progress, and workflows will begin to complete on the new NfN projects soon. I think NfN will reluctantly continue de-activating complete workflows unless we implement one of the options noted above. I think if a backend solution isn’t viable that’s ok, this fronted solution is less preferable, but ok and still more preferable than showing complete workflows?

srallen:
This would be a really ugly hack, but I think would be fast: we could write a conditional just for NfN checking for its project id, then write an API call that only calls for the workflows by id for the ones that they want to show still. I see right now that is 15 workflows. Then when those are done, we can rip this code out and fall back to the default render behavior which is just to show all the workflows as options to classify one. Will they be adding any new workflows to the NfN project or will new workflows only go into new projects under the NfN org?

mcbouslog:
New workflows will be going in new projects, but the new workflows [will complete soon](https://www.zooniverse.org/projects/md68135/notes-from-nature-calbug/stats) and somewhat soon the new projects will have the same issue the old project has, which is differentiating between complete and incomplete workflows on the project landing.

srallen:
Yes that's true, but this has been true for all projects that allow volunteers to choose the workflows to work on since PFE launched. The design update for addressing this will come in the rewrite; and I think that's an adequate answer to give them for now. The more immediate issue I see is just specific to NfN because they still have 100+ workflows linked to the original project, which the home page design never accounted for.

mcbouslog:
More so for the new projects there are other options, like once complete editing the workflow name with "Done" or similar, then repositioning to the bottom of the workflow list, but that'd still look less than desirable for the old NfN project.

mcbouslog:
I don't know if actually happening, but I wonder if even projects with only a few workflows deactivate their complete workflows, so as not to show the completed workflows on their project landing page and confuse volunteers.

trouille:
Yes to Mark's question -- there are projects that start with multiple workflows, complete all subjects for a subset of their workflows, and so deactivate workflows as they complete so their volunteers know not to spend their time on them. As Sarah mentions, instead, better would be if they just update the workflow name to 'Done: X'. That way their volunteers can still see the old workflow tasks for context, etc. This seems fine in the context of a project with not too many workflows and will work well in Becky's future project landing page design.

trouille:
Here though, it would be good to focus specifically on the Notes from Nature situation in which both old and new NfN projects will all eventually have many workflows because of their small subject sets and approach with workflow generation. 

It'd be good to get @camallen perspective -- Should we provide a short-term front-end-based solution like Mark's PR here or Sarah's hack suggested above. Or should we find a panoptes-based solution, knowing that that might take longer to implement. If the latter, any thoughts on what might be best and what timeline might work? 

Note - I'm comfortable following up with NfN to explain why this isn't possible to fix now and that it will be folded into our 'Organizations' enhancements work in the fall.

camallen:
We can add filtering on a workflow's `completeness` attribute API side. I assume from the PR code that the main use case is to filter out any completed project workflows? Would the inverse apply as well? 

I assume adding a query param like `completed=true|false` would suffice on the workflow fetch calls?

eatyourgreens:
I think we should merge this, and apply it by default to all projects There are projects like Weather Rescue which have many completed workflows, but only a few active, incomplete workflows that are shown on the home page. They've also had complaints from volunteers about clicking through to a workflow only to find all the subjects are finished.

srallen:
@camallen do you have a time estimate on how long it would take to add that to Panoptes? I think using a query param would be the preferred solution, but I want to be considerate of dev time and that asking to do this would bump other work.

srallen:
If we use the query param to request for not completed workflows and all the workflows are completed, would the API return workflows that are completed or nothing? If nothing, then the front end should fallback to make another request as we do now so we still show something on finished projects.

camallen:
> do you have a time estimate on how long it would take to add that to Panoptes? I think using a query param would be the preferred solution, but I want to be considerate of dev time and that asking to do this would bump other work.

A day or two to get it deployed, it's a fairly simple change to the WorkflowSerializer to modify the page scope based on the incoming params https://github.com/zooniverse/Panoptes/blob/1cf49c6779346578551d40216f6a4eb083f283f5/app/serializers/workflow_serializer.rb

I think the API side filtering is more optimal for many reasons, less db work serializing possibly large workflows, less transfer traffix $$, less client network usage $, etc

> If we use the query param to request for not completed workflows and all the workflows are completed, would the API return workflows that are completed or nothing? If nothing, then the front end should fallback to make another request as we do now so we still show something on finished projects.

You'd get an empty resource response if all the workflows are completed and you asked for `completed=false` say.

camallen:
I've added https://github.com/zooniverse/Panoptes/pull/3138 to allow API completeness filtering. Please check the API filter param names and values to make sure they make sense and do what you want. Happy to change them now before they make it into the wild.

camallen:
API changes for workflow completeness filtering are now deployed and available on staging and production

mcbouslog:
Awesome, thanks Cam! 

I'll close this and open a different PR to add the completeness filter for the project home, but if response empty request workflows as is currently. I'll reference this PR in the new PR.

PR #5341, opened 2019-05-15
Clear interventions after they've been displayed

### LINKED ISSUE #5337: Intervention messages may not present in the UI
It seems there is a race condition with message events and the redux action, specifically if:

1. Intervention message is sent / received after a classification has been sent and the UI is on the summary screen
0. Press Next subject button and the intervention message is wiped. 

Could the intervention message be wiped in another classifier event or after display? 
https://github.com/zooniverse/Panoptes-Front-End/blob/6169d2b3fd5936b7d53e8818a2026f4d01ffef70/app/redux/ducks/classify.js#L193

Redux dev tools helped massively debugging this.

eatyourgreens:
Staging branch URL: https://pr-5341.pfe-preview.zooniverse.org

Fixes #5337.

Adds a `clearIntervention` action to the classifier and calls it from the `Intervention` component on component cleanup. Changes the `nextSubject` action so that it doesn't clear any stored interventions by default.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/components/Intervention/Intervention.jsx
+++ b/app/classifier/components/Intervention/Intervention.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { memo, useEffect } from 'react';
 import PropTypes from 'prop-types';
 import counterpart from 'counterpart';
 import styled from 'styled-components';
@@ -14,9 +14,14 @@ const StyledInterventionMessage = styled.div`
   }
 `;
 
-function Intervention({ intervention, user }) {
+function Intervention({ onUnmount, intervention, user }) {
   const { message } = intervention;
   const checkbox = React.createRef();
+
+  useEffect(() => {
+    // the return value of an effect will be called to clean up after the component
+    return onUnmount;
+  });
 
   function onChange() {
     // Invert the checked value because true means do not send me messages.
@@ -40,13 +45,19 @@ function Intervention({ intervention, user }) {
   );
 }
 
+Intervention.defaultProps = {
+  onUnmount: () => true
+};
+
 Intervention.propTypes = {
   intervention: PropTypes.shape({
       message: PropTypes.string
     }).isRequired,
+    onUnmount: PropTypes.func,
   user: PropTypes.shape({
     intervention_notifications: PropTypes.bool
   }).isRequired
 };
 
-export default Intervention;
+export default memo(Intervention);
+export { Intervention }
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -370,6 +370,7 @@ class Classifier extends React.Component {
             {showIntervention &&
               <Intervention
                 intervention={intervention}
+                onUnmount={actions.classify.clearIntervention}
                 user={user}
               />
             }
--- a/app/classifier/components/Intervention/Intervention.spec.js
+++ b/app/classifier/components/Intervention/Intervention.spec.js
@@ -2,7 +2,7 @@ import React from 'react';
 import { mount } from 'enzyme';
 import { expect } from 'chai';
 import sinon from 'sinon';
-import Intervention from './Intervention';
+import { Intervention } from './Intervention';
 import { Markdown } from 'markdownz';
 
 describe('Intervention', function () {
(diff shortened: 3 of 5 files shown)


app/classifier/components/Intervention/Intervention.jsx
```diff
@@ -14,10 +14,15 @@ const StyledInterventionMessage = styled.div`
...
 `;
 
-function Intervention({ intervention, user }) {
+function Intervention({ onUnmount, intervention, user }) {
   const { message } = intervention;
   const checkbox = React.createRef();
 
+  useEffect(() => {
+    // the return value of an effect will be called to clean up after the component
+    return onUnmount;
```
camallen: My reading suggests this will only run on unmount but it looks like effects will cleanup up if props change as well - https://reactjs.org/docs/hooks-effect.html#explanation-why-effects-run-on-each-update

I don't think this is an issue as we don't pass props to this function, thoughts?

app/classifier/components/Intervention/Intervention.jsx
```diff
@@ -40,13 +45,19 @@ function Intervention({ intervention, user }) {
...
       message: PropTypes.string
     }).isRequired,
+    onUnmount: PropTypes.func,
   user: PropTypes.shape({
     intervention_notifications: PropTypes.bool
   }).isRequired
 };
 
-export default Intervention;
+export default memo(Intervention);
```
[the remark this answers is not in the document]
eatyourgreens: The default will be used anywhere that we use `import Intervention from '/path/to/Intervention'.

Enzyme doesn't support `memo()` so the unwrapped component is exported for tests.

app/classifier/components/Intervention/Intervention.jsx
eatyourgreens: Good catch. The props here are `onUnmount, intervention, user` and we probably do want cleanup to run if any of those change.

eatyourgreens: This might be useful.
>If you want to run an effect and clean it up only once (on mount and unmount), you can pass an empty array ([]) as a second argument. This tells React that your effect doesn’t depend on any values from props or state, so it never needs to re-run. This isn’t handled as a special case — it follows directly from how the dependencies array always works.
https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects

app/classifier/components/Intervention/Intervention.jsx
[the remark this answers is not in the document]
camallen: Thanks - makes sense.

app/classifier/components/Intervention/Intervention.jsx
eatyourgreens: I've been playing around with this a bit and I think the only time it might happen is if the `intervention` prop changes while the component is being displayed. Is there a case where I might get a new intervention message while I'm already reading one?

camallen: Yes - though in practice it would be slim. They don't send many messages but I could receive one while i'm viewing one, especially if the network / gateway has issues, i.e. received message at gateway, send to sugar but timeout to client who retries. 

Would the current code unmount and then update with the newly received message? Should we drop incoming messages until there are no interventions in the store?

I'd say no to switching the store to a list of messages, unless the incoming message has a validity period attached to it, https://github.com/zooniverse/interventions-gateway/issues/22

eatyourgreens: My reading of the code is the classifier would receive a new `intervention` prop, which it would pass down to the `Intervention` component, which would then run `useEffect()` because of the props change? I'm not sure what would happen then: maybe the new intervention would render briefly, then be cleared by the cleanup code.

camallen: I'm happy to add the validity period to the incoming messages via the gateway API if it helps simplify this issue.

eatyourgreens: We could use the second argument to `useEffect` to ignore changes in the `intervention` prop. I'm not sure what the implications of that are, but it sounds like a case that shouldn't occur anyway.

At the moment, if a second message were received while you're currently reading one, it would overwrite it and the screen would update to show the new message.

PR #5343, opened 2019-05-16
add links to the opt out message

camallen:
use markdown to render the optOut string correctly with a link to study. ~Also test the default translation fall back works~

Staging branch URL:

Fixes # .

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/components/Intervention/Intervention.spec.js
+++ b/app/classifier/components/Intervention/Intervention.spec.js
@@ -4,7 +4,6 @@ import { expect } from 'chai';
 import sinon from 'sinon';
 import { Intervention } from './Intervention';
 import { Markdown } from 'markdownz';
-import counterpart from 'counterpart';
 
 describe('Intervention', function () {
   let wrapper;
--- a/app/classifier/components/Intervention/Intervention.spec.js
+++ b/app/classifier/components/Intervention/Intervention.spec.js
@@ -7,6 +7,7 @@ import { Markdown } from 'markdownz';
 import counterpart from 'counterpart';
 
 describe('Intervention', function () {
+  let wrapper;
   const intervention = { message: 'Hello!' };
   const { message } = intervention;
   const user = {
@@ -17,12 +18,15 @@ describe('Intervention', function () {
   };
   const onUnmount = sinon.stub()
 
-  let wrapper = mount(
+  before(function () {
+    wrapper = mount(
     <Intervention
       intervention={intervention}
       onUnmount={onUnmount}
       user={user}
-    />);
+      />
+    );
+  });
 
   it('should render', function () {
     expect(wrapper).to.be.ok;
--- a/app/classifier/components/Intervention/Intervention.spec.js
+++ b/app/classifier/components/Intervention/Intervention.spec.js
@@ -4,9 +4,9 @@ import { expect } from 'chai';
 import sinon from 'sinon';
 import { Intervention } from './Intervention';
 import { Markdown } from 'markdownz';
+import counterpart from 'counterpart';
 
 describe('Intervention', function () {
-  let wrapper;
   const intervention = { message: 'Hello!' };
   const { message } = intervention;
   const user = {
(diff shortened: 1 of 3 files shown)


app/classifier/components/Intervention/Intervention.spec.js
```diff
@@ -17,20 +17,31 @@ describe('Intervention', function () {
...
   it('should render', function () {
     expect(wrapper).to.be.ok;
   });
   it('should show a notification message', () => {
     const newlineMsg = intervention.message + "\n"
-    expect(wrapper.find(Markdown).text()).to.equal(newlineMsg);
+    expect(wrapper.find(Markdown).first().text()).to.equal(newlineMsg);
+  });
+  it('should show an opt-out message', () => {
+    const optOutHTML = '<div class="markdown "><p>I do not want to take part in this messaging ' +
```
camallen: I'm not super happy testing the HTML here, but i wasn't sure what to do, i could check for a link `<a href=` tag vs the component html but lose some granularity over the test. 

Also for some reason there is a new line before the last div 🤷‍♂

app/classifier/components/Intervention/Intervention.spec.js
```diff
@@ -17,20 +17,31 @@ describe('Intervention', function () {
...
 
-  before(function () {
-    wrapper = mount(
-      <Intervention
-        intervention={intervention}
-        onUnmount={onUnmount}
-        user={user}
-      />);
-  });
+  let wrapper = mount(
```
eatyourgreens: Why did you delete the `before` hook?

app/classifier/components/Intervention/Intervention.spec.js
eatyourgreens: Yeah, this is a really brittle test.

app/classifier/components/Intervention/Intervention.spec.js
```diff
@@ -17,20 +17,31 @@ describe('Intervention', function () {
...
+    expect(wrapper.find(Markdown).first().text()).to.equal(newlineMsg);
+  });
+  it('should show an opt-out message', () => {
+    const optOutHTML = '<div class="markdown "><p>I do not want to take part in this messaging ' +
+      '<a href="https://docs.google.com/document/d/1gLyN6Dgff8dOCOC88f47OD6TtFrfSJltsLgJMKkYMso/preview" ' +
+      'target="_blank" ref="noopener nofollow">study</a>. Do not show me further messages.</p>' +
+      '\n' + '</div>'
+    expect(wrapper.find(Markdown).last().html()).to.equal(optOutHTML);
+  });
+  it('should show a fallback to english opt-out message', () => {
```
eatyourgreens: This is testing counterpart, not the component.

app/classifier/components/Intervention/Intervention.spec.js
srallen: Could try [`contains`](https://airbnb.io/enzyme/docs/api/ShallowWrapper/contains.html) instead? I sometimes find that a bit more readable than the html render method for shallow.

eatyourgreens: Examining the props of the Markdown component might allow you to test the markdown string that’s passed into it as `props.children`.

app/classifier/components/Intervention/Intervention.spec.js
camallen: The test was on the wrapper component output, which relies on counterpart. Sure i had to modify counterpart in setup...anyway i can't reliably get the test to fail as counterpart seems to have fallbackLocales setup by default. 

I couldn't see where this config was being set, default? So I'm going to assume that this is normal, expected behaviour and remove the test.

```
{ locale: 'en',
     interpolate: true,
     fallbackLocales: [ 'en' ],
     scope: null,
     translations: 
      { en: [Object],
        cs: [Object],
        it: [Object],
        es: [Object],
        nl: [Object],
        pt: [Object] },
     interpolations: {},
     normalizedKeys: { '.': [Object] },
     separator: '.',
     keepTrailingDot: false,
     keyTransformer: [Function: keyTransformer],
     generateMissingEntry: [Function: generateMissingEntry] },
```

app/classifier/components/Intervention/Intervention.spec.js
camallen: Thanks that is a much nicer solution :)

app/classifier/components/Intervention/Intervention.spec.js
```diff
@@ -30,7 +30,14 @@ describe('Intervention', function () {
   });
   it('should show a notification message', () => {
     const newlineMsg = intervention.message + "\n"
-    expect(wrapper.find(Markdown).text()).to.equal(newlineMsg);
+    expect(wrapper.find(Markdown).first().text()).to.equal(newlineMsg);
+  });
+  it('should show an opt-out message', () => {
+    const expectedMarkdown = 'I do not want to take part in this messaging [study]' +
```
eatyourgreens: You can also get this string by loading the English language file, but that might be a lot of work for a single string.

camallen: I really want to make sure that the test has the study link setup, if we use the translation strings themselves and someone removes the link in the translation string the tests will pass but without the correct study link.

eatyourgreens: That makes sense, but we're now maintaining the text in two places, which is slightly annoying.

app/classifier/components/Intervention/Intervention.spec.js
eatyourgreens: From memory, [counterpart falls back to English by default](https://www.npmjs.com/package/counterpart#locales). We should probably update that to set the default locale to the project primary language when a project loads, if we don't do that already. It just so happens that counterpart's default is the same as Panoptes' default, so it magically works.

eatyourgreens: Ignore that. We set the fallback when the JS first loads.
https://github.com/zooniverse/Panoptes-Front-End/blob/9461eb8f83b4b12e0b9bee501cbe8d535783e1a7/app/redux/ducks/translations.js#L5-L11

camallen: I commented out the two `counterpart.setFallbackLocale();` functions and it still had the fallback set. I was a bit lost at that point and decided `countepart` was magical

app/classifier/components/Intervention/Intervention.spec.js
eatyourgreens: By the way, is this split into multiple lines so that the linter doesn't complain? The concatenation is a bit strange.

PR #5342, opened 2019-05-15
Force reload of publications page route

rogerhutchings:
Uses the RELOAD component to force a reload of `/about/publications` and ensure it's fetched from the new content pages app

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #5345, opened 2019-05-16
Change `RELOAD` component to redirect when not in production

rogerhutchings:
The RELOAD component will now redirect to a new URL (passed as the `newUrl` prop). This means that in staging, going to a route which returns this component will redirect to `newUrl`, rather than getting trapped in a refresh loop.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/router.cjsx
+++ b/app/router.cjsx
@@ -75,7 +75,10 @@ ONE_UP_REDIRECT = createReactClass
 # Usage: <Route path="path/to/location" component={RELOAD} />
 RELOAD = createReactClass
   componentDidMount: ->
-    window.location.reload()
+    if window.location.hostname is 'www.zooniverse.org'
+      window.location.reload()
+    else
+      window.location = @props.newUrl
   render: ->
     null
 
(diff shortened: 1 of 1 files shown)


srallen:
Does #5342 need to come before this or does this supersede it?

rogerhutchings:
This supersedes #5342

eatyourgreens:
I'm missing something here. Both of these lines force the browser to issue a new GET request for the current URL, so why do we need `location.reload()`?

rogerhutchings:
On production, it'll reload the same URL (https://www.zooniverse.org/about/publications), so it hits the pod serving up `app-content-pages` through CloudFront. Aside from a full-page refresh, nothing changes.

On staging, it'll redirect you to `newUrl`, which for this route is set to `app-content-pages` on its own subdomain. So when you click on the link in staging, it a) goes somewhere without breaking, and b) goes to somewhere that isn't obviously production

PR #5346, opened 2019-05-17
Only show intervention messages if workflow matches payload

camallen:
Ensure we only show intervention messages for specified project, workflow combinations, if the intervention payload has a `workflow_id`. If there is no specified payload `workflow_id` default to project wide messages but still check the `project_id` matches the current store state.

Also ensure we compare project and workflow ids as strings as as sugar messages have int link ID values.

Staging branch URL:

Fixes # .

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -150,7 +150,7 @@ export default function reducer(state = initialState, action = {}) {
       // to allow project level as well as targetted workflow messages
       let matchesWorkflow = true;
       if (intervention.hasOwnProperty('workflow_id')) {
-        matchesWorkflow = workflow === String(intervention.workflow_id);
+        matchesWorkflow = workflow === intervention.workflow_id.toString();
       }
       if (classification && matchesProject && matchesWorkflow ) {
         return Object.assign({}, state, { intervention });
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -146,11 +146,11 @@ export default function reducer(state = initialState, action = {}) {
       const { classification } = state;
       const { project, workflow } = classification.links;
       const matchesProject = project === intervention.project_id;
-      // only test workflow matches known state
-      // if the incoming payload has this property
+      // only test workflow matches known state if the incoming payload has this property
+      // to allow project level as well as targetted workflow messages
       let matchesWorkflow = true;
       if (intervention.hasOwnProperty('workflow_id')) {
-        let matchesWorkflow = workflow === intervention.workflow_id;
+        matchesWorkflow = workflow === intervention.workflow_id;
       }
       if (classification && matchesProject && matchesWorkflow ) {
         return Object.assign({}, state, { intervention });
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -141,11 +141,17 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
 export default function reducer(state = initialState, action = {}) {
   switch (action.type) {
     case ADD_INTERVENTION: {
+      // this action payload is the json intervention object from sugar
       const intervention = action.payload;
       const { classification } = state;
       const { project, workflow } = classification.links;
       const matchesProject = project === intervention.project_id;
-      const matchesWorkflow = workflow === intervention.workflow_id;
+      // only test workflow matches known state
+      // if the incoming payload has this property
+      let matchesWorkflow = true;
+      if (intervention.hasOwnProperty('workflow_id')) {
+        let matchesWorkflow = workflow === intervention.workflow_id;
+      }
       if (classification && matchesProject && matchesWorkflow ) {
         return Object.assign({}, state, { intervention });
       }
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -144,6 +144,9 @@ export default function reducer(state = initialState, action = {}) {
       // this action payload is the json intervention object from sugar
       const intervention = action.payload;
       const { classification } = state;
+      if (!classification) {
+        return state;
+      }
       const { project, workflow } = classification && classification.links;
       const matchesProject = project === intervention.project_id.toString();
       // only test workflow matches known state if the incoming payload has this property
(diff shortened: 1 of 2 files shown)


app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
 export default function reducer(state = initialState, action = {}) {
   switch (action.type) {
     case ADD_INTERVENTION: {
+      // this action payload is the json intervention object from sugar
       const intervention = action.payload;
       const { classification } = state;
-      if (classification && classification.links.project === intervention.project_id) {
+      const { project, workflow } = classification.links;
+      const matchesProject = project === String(intervention.project_id);
```
srallen: `String()` will cast any type to string, so while unlikely, if for some reason whatever we get back from Sugar is not an integer here, do we still want to code to convert that to a string or do we want it to throw? `Number.prototype.toString()` will throw.

srallen: Also, this will fail if `intervention` is undefined.

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
...
+      // only test workflow matches known state if the incoming payload has this property
+      // to allow project level as well as targetted workflow messages
+      let matchesWorkflow = true;
+      if (intervention.hasOwnProperty('workflow_id')) {
```
srallen: Is there any chance `intervention` will be undefined here?

srallen: `hasOwnProperty` will evaluate whether or not that object has that property, but not if the property has a value, so: 
```
const intervention = { workflow_id: undefined }
intervention.hasOwnProperty('workflow_id') // true
```

If you mean to check that the property exists and has a value, I think you'll have to use dot notation:
```
if (intervention && intervention.workflow_id) {
 // do something
}
```

app/redux/ducks/classify.js
eatyourgreens:
```suggestion
      const matchesProject = project === intervention.project_id.toString();
```

eatyourgreens: The deleted line would have failed if `intervention` was undefined too, so I think this is ok.

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
...
+        matchesWorkflow = workflow === String(intervention.workflow_id);
```
eatyourgreens:
```suggestion
        matchesWorkflow = workflow === intervention.workflow_id.toString();
```

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
...
+      }
+      if (classification && matchesProject && matchesWorkflow ) {
```
eatyourgreens: if `classification` was falsy then line 147 would error, so I think it will always evaluate to `true`.

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
 export default function reducer(state = initialState, action = {}) {
   switch (action.type) {
     case ADD_INTERVENTION: {
+      // this action payload is the json intervention object from sugar
       const intervention = action.payload;
       const { classification } = state;
-      if (classification && classification.links.project === intervention.project_id) {
+      const { project, workflow } = classification.links;
```
eatyourgreens:
```suggestion
      const { project, workflow } = classification && classification.links;
```

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
 export default function reducer(state = initialState, action = {}) {
   switch (action.type) {
     case ADD_INTERVENTION: {
+      // this action payload is the json intervention object from sugar
       const intervention = action.payload;
       const { classification } = state;
```
eatyourgreens: If the classification is null, we should probably return the current state at this point and not take any further action.

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
...
-      if (classification && classification.links.project === intervention.project_id) {
+      const { project, workflow } = classification && classification.links;
+      const matchesProject = project === intervention.project_id.toString();
+      // only test workflow matches known state if the incoming payload has this property
+      // to allow project level as well as targetted workflow messages
+      let matchesWorkflow = true;
+      if (intervention.hasOwnProperty('workflow_id')) {
```
eatyourgreens:
```suggestion
      if (intervention && intervention.workflow_id) {
```

app/redux/ducks/classify.js
```diff
@@ -141,9 +141,18 @@ const TOGGLE_GOLD_STANDARD = 'pfe/classify/TOGGLE_GOLD_STANDARD';
...
+        matchesWorkflow = workflow === intervention.workflow_id.toString();
+      }
+      if (classification && matchesProject && matchesWorkflow ) {
```
eatyourgreens:
```suggestion
      if (matchesProject && matchesWorkflow ) {
```

app/redux/ducks/classify.js
eatyourgreens:
```suggestion
      const { classification } = state;
      if (!classification) {
        return state;
      }
```

app/redux/ducks/classify.js
camallen: avoid `hasOwnProperty` directly access the attribute

PR #5347, opened 2019-05-17
Track intervention UUID in classification metadata for downstream clients

camallen:
Track the intervention payload UUID in the next classification metadata, this links the any classification after an intervention message was seen.

linked to https://github.com/zooniverse/interventions-gateway/issues/28

Staging branch URL:

Fixes # .

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -152,7 +152,6 @@ export default function reducer(state = initialState, action = {}) {
     case STORE_INTERVENTION_UUID: {
       const { intervention } = state;
       let lastInterventionUUID = null
-      console.log(intervention);
       if (intervention && intervention.hasOwnProperty('uuid')) {
         lastInterventionUUID = intervention.uuid;
       }
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -122,6 +122,7 @@ const initialState = {
   classification: null,
   goldStandardMode: false,
   intervention: null,
+  lastInterventionUUID: null,
   upcomingSubjects: [],
   workflow: null
 };
--- a/app/redux/ducks/classify.js
+++ b/app/redux/ducks/classify.js
@@ -45,22 +45,27 @@ function awaitWorkflow(workflowId) {
     });
 }
 
-function createNewClassification(project, workflow, subject, goldStandardMode) {
+function createNewClassification(project, workflow, subject, goldStandardMode, lastInterventionUUID) {
   // Record whether this subject was received from Sugar or from the Panoptes API.
   const source = subject.metadata.intervention ? 'sugar' : 'api';
   // Delete the metadata key because we don't want volunteers to see it.
   subject.update({ 'metadata.intervention': undefined });
+  let newMetadata = {
+    workflow_version: workflow.version,
+    started_at: (new Date()).toISOString(),
+    user_agent: navigator.userAgent,
+    user_language: counterpart.getLocale(),
+    utc_offset: ((new Date()).getTimezoneOffset() * 60).toString(), // In seconds
+    subject_dimensions: (subject.locations.map(() => null)),
+    source
+  }
+  // record if this classification had an intervention payload directly before it
+  if (lastInterventionUUID) {
+    newMetadata.intervention_uuid = lastInterventionUUID
+  }
   const classification = apiClient.type('classifications').create({
     annotations: [],
-    metadata: {
-      workflow_version: workflow.version,
-      started_at: (new Date()).toISOString(),
-      user_agent: navigator.userAgent,
-      user_language: counterpart.getLocale(),
-      utc_offset: ((new Date()).getTimezoneOffset() * 60).toString(), // In seconds
-      subject_dimensions: (subject.locations.map(() => null)),
-      source
-    },
+    metadata: newMetadata,
     links: {
       project: project.id,
       workflow: workflow.id,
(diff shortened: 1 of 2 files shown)


app/redux/ducks/classify.js
```diff
@@ -264,8 +286,9 @@ export function addIntervention(data) {
 }
 
 export function clearIntervention() {
-  return {
-    type: CLEAR_INTERVENTION
+  return (dispatch) => {
+    dispatch({type: STORE_INTERVENTION_UUID});
```
camallen: I wasn't sure that this was the way to do it, this code path isn't tested in specs by the look. Thoughts appreciated.

app/redux/ducks/classify.js
```diff
@@ -173,11 +187,19 @@ export default function reducer(state = initialState, action = {}) {
       const { goldStandardMode } = state;
       const { project } = action.payload;
       const { workflow } = state;
+      let { lastInterventionUUID} = state;
```
camallen: I didn't take into account if the user is actually opted into the interventions and thus saw the message. I think this variable needs to be reflect this state before propagating to the `createNewClassification` method. Thoughts?

app/redux/ducks/classify.spec.js
```diff
@@ -278,6 +307,26 @@ describe('Classifier actions', function () {
...
+      expect(newState.classification.metadata.hasOwnProperty('intervention_uuid')).to.be.false;
+    });
+    describe('with lastInteventionUUID set', function () {
+      const interventionUUIDState = {
+        classification: { id: '1' },
+        workflow: { id: '1' },
+        upcomingSubjects: [mockSubject('1')],
+        lastInterventionUUID: '2d931510-d99f-494a-8c67-87feb05e1594'
+      };
+      it('should record the lastInteventionUUID as metadata.intervention_uuid', function () {
```
srallen: Typo in the it statement. It should be `should record the lastInterventionUUID as metadata.intervention_uuid`

app/redux/ducks/classify.js
srallen: If I'm understanding this question correctly, there are two pieces of information you need to know? Am I also correct in understanding this to mean that there can be a last intervention UUID, but the user may not have actually seen the intervention? If that's the case, then perhaps the state can be an object and the UUID and seen state is explicitly stored. Something like?: `{ lastIntervention: { UUID: '', seen: boolean } }` and this also gets recorded in the classification metadata?

app/redux/ducks/classify.js
srallen: As someone who isn't familiar with the implementation of interventions in general, reading this is confusing and hard to follow: A clear intervention function is called but then is calling an action to store the id in the state, then in the classification metadata, then the id gets cleared from the state? I would suggest at the very least adding a separate redux action for clearing the UUID if there's going to be an action for storing it. Calling to store the UUID here doesn't seem quite like the right place to me either. The naming of `clearIntervention` would not convey to me that actions are called in the function that do store data.

eatyourgreens: Firing two state changes together should be ok, but I don’t know if Redux guarantees that the new state after saving the UUID will be the initial state for clearing the intervention. 

I’d be inclined to return a single action which remembers the UUID then clears the intervention, so state is only updated once.

app/redux/ducks/classify.js
eatyourgreens: If you call actions from the `Intervention` component, I think that guarantees they will only be called for volunteers who have opted to see interventions.

app/redux/ducks/classify.js
eatyourgreens: The specs for these actions are the first tests I wrote for redux in PFE and I'm not sure any more that testing the actions directly is right.

You should be able to test the `clearIntervention()` action creator if you import it and the reducer into the tests, and pass it a mock `dispatch`. Have a look at the specs for the translations action creators, where I used that approach. The mock dispatch function I came up with is this one:
https://github.com/zooniverse/Panoptes-Front-End/blob/9461eb8f83b4b12e0b9bee501cbe8d535783e1a7/app/redux/ducks/translations.spec.js#L26-L31

app/redux/ducks/classify.js
eatyourgreens: Reading through this some more, `lastInterventionUUID` is set when the `Intervention` component unmounts, so I think it can only be truthy here if an intervention was shown to the volunteer.

If I'm understanding these changes correctly:

1. I see an intervention and continue.
2. The intervention UUID is stored and the seen intervention is cleared.
3. When the next subject loads, `createNewClassification` is called, the UUID is added to the new classification and the stored UUID is cleared.

app/redux/ducks/classify.js
```diff
@@ -148,6 +154,14 @@ export default function reducer(state = initialState, action = {}) {
       }
       return state;
     }
+    case STORE_INTERVENTION_UUID: {
+      const { intervention } = state;
+      let lastInterventionUUID = null
+      if (intervention && intervention.hasOwnProperty('uuid')) {
+        lastInterventionUUID = intervention.uuid;
+      }
+      return Object.assign({}, state, { lastInterventionUUID });
```
eatyourgreens: This returns a new state object even if there's no intervention. Is that intentional? I think it should return the current state in that case, but I'm not sure.

app/redux/ducks/classify.js
```diff
@@ -148,6 +154,14 @@ export default function reducer(state = initialState, action = {}) {
       }
       return state;
     }
+    case STORE_INTERVENTION_UUID: {
+      const { intervention } = state;
+      let lastInterventionUUID = null
+      if (intervention && intervention.hasOwnProperty('uuid')) {
```
eatyourgreens:
```suggestion
      if (intervention && intervention.uuid) {
```

app/redux/ducks/classify.spec.js
camallen:
```suggestion
      it('should record the lastInterventionUUID as metadata.intervention_uuid', function () {
```

app/redux/ducks/classify.js
eatyourgreens: I've pushed a test for the `clearIntervention` action creator.

eatyourgreens:
I think that would require changing the behaviour of the classifier, so couldn't be done in a refactor. At the moment, the intervention is cleared when you close the Intervention component and finish a subject, then the new classification is created when the next subject loads.

srallen:
I've realized that my interpretation of what's going on was incorrect because of how github is previewing the diff of that file, so you can ignore my suggestion. I'll read it over again, but in VSCode instead.

PR #5349, opened 2019-05-21
Only clear interventions on Intervention unmount

eatyourgreens:
Staging branch URL: https://pr-5349.pfe-preview.zooniverse.org

Expect that the Intervention component will not call `props.unMount` on props change. Update `useEffect` to pass that test.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/components/Intervention/Intervention.jsx
+++ b/app/classifier/components/Intervention/Intervention.jsx
@@ -20,7 +20,9 @@ function Intervention(props) {
   const checkbox = React.createRef();
 
   useEffect(() => {
-    // the return value of an effect will be called to clean up after the component
+     /* the return value of an effect will be called to clean up after the component.
+     Passing an empty array ([]) as a second argument tells React that your effect doesn’t depend on any values from props or state
+     so it never needs to re-run, https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects */
     return onUnmount;
   }, []);
 
--- a/app/classifier/components/Intervention/Intervention.jsx
+++ b/app/classifier/components/Intervention/Intervention.jsx
@@ -14,14 +14,15 @@ const StyledInterventionMessage = styled.div`
   }
 `;
 
-function Intervention({ onUnmount, intervention, user }) {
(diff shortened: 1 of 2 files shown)


app/classifier/components/Intervention/Intervention.jsx
```diff
@@ -14,14 +14,15 @@ const StyledInterventionMessage = styled.div`
...
+function Intervention(props) {
+  const { onUnmount, intervention, user } = props;
   const { message } = intervention;
   const checkbox = React.createRef();
 
   useEffect(() => {
     // the return value of an effect will be called to clean up after the component
     return onUnmount;
-  });
+  }, []);
```
camallen: OOI - what does this empty array do, signal what props are used to determine when the `onUnmount` function is called vs when the `useEffect` hook is run?

Should we add a comment explaining that?

eatyourgreens: https://github.com/zooniverse/Panoptes-Front-End/pull/5341#discussion_r284258363

app/classifier/components/Intervention/Intervention.jsx
```diff
@@ -14,14 +14,15 @@ const StyledInterventionMessage = styled.div`
...
 `;
 
-function Intervention({ onUnmount, intervention, user }) {
+function Intervention(props) {
+  const { onUnmount, intervention, user } = props;
   const { message } = intervention;
   const checkbox = React.createRef();
 
   useEffect(() => {
     // the return value of an effect will be called to clean up after the component
```
camallen:
```suggestion
    // the return value of an effect will be called to clean up after the component
    // passing an empty array ([]) as a second argument tells React that your effect doesn’t depend on any values from props or state
    // so it never needs to re-run, https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
```

eatyourgreens:
I'm ambivalent about the comment duplicating the Hooks documentation, which is still new-ish and might change. Can't hurt though.

PR #5353, opened 2019-05-23
record the last intervention UUID on the interventions metadata object

camallen:
Avoid storing intervention metadata in different places instead consolidate all interventions metadata to the one object store.

Staging branch URL:

Fixes # .

Describe your changes.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -262,7 +262,7 @@ class Classifier extends React.Component {
       {},
       classification.metadata.interventions,
       {
-        message: !!showIntervention,
+        messageShown: !!showIntervention,
         opt_in: !!user && user.intervention_notifications
       }
     );
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -258,15 +258,20 @@ class Classifier extends React.Component {
     }
     const { already_seen, finished_workflow, retired, selection_state, user_has_finished_workflow, selected_at } = subject;
     const workflowTranslation = translations.strings.workflow[workflow.id];
+    const mergedInterventions = Object.assign(
+      {},
+      classification.metadata.interventions,
+      {
+        message: !!showIntervention,
+        opt_in: !!user && user.intervention_notifications
+      }
+    );
     actions.classify.updateMetadata({
       viewport: {
         width: innerWidth,
         height: innerHeight
       },
-      interventions: {
-        message: !!showIntervention,
-        opt_in: !!user && user.intervention_notifications
-      },
+      interventions: mergedInterventions,
       subject_selection_state: {
         already_seen,
         finished_workflow,
--- a/app/redux/ducks/classify.spec.js
+++ b/app/redux/ducks/classify.spec.js
@@ -459,8 +459,9 @@ describe('Classifier actions', function () {
         }
       }
     };
-    // ensure state is reset after each test
+    // setup state to be reset in beforeEach function
     let state = null;
+
     beforeEach(function () {
       state = {
         classification: mockPanoptesResource('classifications', {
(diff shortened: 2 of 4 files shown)


app/classifier/classifier.jsx
```diff
@@ -258,15 +258,20 @@ class Classifier extends React.Component {
     }
     const { already_seen, finished_workflow, retired, selection_state, user_has_finished_workflow, selected_at } = subject;
     const workflowTranslation = translations.strings.workflow[workflow.id];
+    const mergedInterventions = Object.assign(
+      {},
+      classification.metadata.interventions,
+      {
+        message: !!showIntervention,
```
eatyourgreens: Not related to this PR, but should this be the message that was shown, rather than a boolean? I'm just wondering how the experiment tracks which messages are successful if we don't include the message text in the classification.

eatyourgreens: I'm also trying to remember why I set this as a boolean in the first place, but coming up blank.

camallen: I don't think we need to track the actual message text right now. The UUID is a proxy for this from our gateway system to the downstream data, i.e. for external users to track who saw what message.

Longer term i think we need to ensure we control the message content is from an internally approved list and then we can record the ID / TYPE of the message for lookup.

camallen: > I'm also trying to remember why I set this as a boolean in the first place, but coming up blank.

I think maybe to record that there was a message for the user to see vs the actual message, to be combined with the `opt_in` attribute to know if they saw it.

eatyourgreens: Sure, I'm just missing how an experiment tests which interventions were successful without either an ID/type or the message text in the classification.

srallen: I would say that if this is recording whether or not the message was shown, then this should be `messageShown: !!showIntervention` for clarity.

camallen: happy to make the change but noting that it is unrelated to this pr. I assume this won't break any downstream client scripts as they aren't using this attribute.

@eatyourgreens you ok with this requested change?

camallen:
Thanks for the update, i think this is good to got but needs another review.

app/redux/ducks/classify.spec.js
```diff
@@ -459,22 +459,26 @@ describe('Classifier actions', function () {
...
-      workflow: {
-        id: '1',
-        tasks: {
-          a: {}
-        }
-      },
-      upcomingSubjects: [mockSubject('1'), mockSubject('2')]
-    };
+    // ensure state is reset after each test
+    let state = null;
```
srallen: If you want to ensure this is happening after each test, then I think the safest way to do this is to use `afterEach(function () { state = null }`  within the describe block too.

camallen: This was working as expected, the comment was the problem as it was slightly misleading. 

`let state = null;` sets up the state variable for the describe block, the state itself is now reset in the `beforeEach` function on line 465 instead of having state carry over from one test to the other. 

I.e. the test setup is now clean whereas before state modifications through redux actions polluted the next test.

eatyourgreens: In that case, can you declare the state at the top of the block, so it's clear that it's set at the same time as the test functions are declared.

camallen: Can do - personally i think this is an issue with how the GH UI displays the change set as when you look at the code it's pretty obvious it's a variable setup declaration.
Screenshot 2019-06-12 at 10 25 21

eatyourgreens: Yeah, I misread it as being after the `it()` functions. My bad.

srallen: Oh same. I've been bitten by how github displays diffs in the past too. Gotta remember to view it as the full file.

camallen:
Still blocked as changes requested.....

PR #5372, opened 2019-06-07
Edit Visibility Page Text

lcjohnso:
Edited post-beta instructions on visibility page to have teams properly request launch via contact email.

While editing this page, I also removed the temporary "workflow controls have moved" notice, as that change is now over a year old (Feb 2018; PR https://github.com/zooniverse/Panoptes-Front-End/pull/4331).

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -90,7 +90,7 @@ module.exports = createReactClass
             <p>
               Review status for this project has been approved. To end the review and make changes, switch back to <em>development</em> mode.
               {unless @props.project.launch_requested or @props.project.launch_approved
-                <span>If you’re ready to launch this project, see the next section.</span>}
+                <span>If you’re ready to launch this project, please email contact@zooniverse.org to submit your request.</span>}
             </p>
           </span>
         else if @props.project.beta_requested
--- a/app/pages/lab/visibility.cjsx
+++ b/app/pages/lab/visibility.cjsx
@@ -104,10 +104,5 @@ module.exports = createReactClass
         }
 
       </div>
-      <hr />
-      <div>
-        <p className="form-label">Temporary notice</p>
-        <p>The workflow settings have moved from the visibility page to the <Link to="/lab/#{@props.project.id}/workflows">workflows page</Link>. For improved discoverability, we have moved these settings so all relevant information and configuration of workflows is in one place.</p>
-      </div>
-
+  
     </div>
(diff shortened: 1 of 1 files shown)


app/pages/lab/visibility.cjsx
```diff
@@ -90,7 +90,7 @@ module.exports = createReactClass
             <p>
               Review status for this project has been approved. To end the review and make changes, switch back to <em>development</em> mode.
               {unless @props.project.launch_requested or @props.project.launch_approved
-                <span>If you’re ready to launch this project, see the next section.</span>}
+                <span>If you’re ready to launch this project, please email contact@zooniverse.org to submit your request.</span>}
```
srallen: Do you want this to be a working mailto link?

lcjohnso: I was following Brooke's lead (in PR https://github.com/zooniverse/help/pull/3) to not encode it due to spam harvesting, but seeing that there is a mailto link for this same email on the [About/Contact](https://www.zooniverse.org/about/contact) page, might as well add it.

app/pages/lab/visibility.cjsx
```diff
@@ -90,7 +90,7 @@ module.exports = createReactClass
             <p>
               Review status for this project has been approved. To end the review and make changes, switch back to <em>development</em> mode.
               {unless @props.project.launch_requested or @props.project.launch_approved
-                <span>If you’re ready to launch this project, see the next section.</span>}
+                <span>If you’re ready to launch this project, please email <a href="mailto:contact@zooniverse.org">contact@zooniverse.org</a> to submit your request.</span>}
```
lcjohnso: @srallen I'm supposed to use an html tag for the mailto link here, correct?

app/pages/lab/visibility.cjsx
srallen: I think bots can scrap email addresses regardless of the presence of an anchor tag using a mailto. I could be wrong so let me do a bit of research of what the current state of this is.

srallen: Having the address be plaintext in the html whether it is wrapped by an anchor is just as harvestable. Several of the obfuscation techniques break accessibility, so we shouldn't do those. If we're worried about it, then the best solution is probably to setup a webform that has a captcha rather than place the address in the html. At least with the anchor link, it's more convenient for use by the user.

app/pages/lab/visibility.cjsx
srallen: Yep!

app/pages/lab/visibility.cjsx
lcjohnso: Personally, I'm not worried. Contact doesn't seem to get much spam, so might as well make it as easy + accessible as possible.  Thanks for the info!

PR #5377, opened 2019-06-11
add retirement config selectors to admin page

### LINKED ISSUE #5367: Enable never_retire configuration for a workflow via project builder
A number of workflows have been set-up with retirement disabled via the `never_retire` criteria option.  For reference, this option is stored by Panoptes in the workflow table's retirement field as a json blob.  An example, as used by Space Warps (project_id=5596) primary workflow (workflow_id=6605): `{"options": {"count": 10000}, "criteria": "never_retire"}`

**PFE Feature Request:** allow project owners to enable this never_retire option.  I propose the option be an experimental feature, requiring permission for a project owner to enable.

Use Cases
1) The primary use case where this option is currently being used: to disable normal classification count-based retirement when Caesar-based retirement is enabled (e.g., Caesar is running SWAP and responsible for all retirement in case of Space Warps).

2) A new and perhaps growing use case is practice or training workflows (e.g., as used by Earthquake Detective), where a small number of subjects are made available in a stand-alone workflow.  These subjects are constant and unchanging and therefore should not retire.

Relevant file: app/components/retirement-rules-editor.cjsx

Admittedly this is a low-priority enhancement where the current workaround is to use a large N retirement threshold (e.g., 5000 or 10000).  But since the option already exists I wanted to see whether we could expose this configuration option to more projects.

camallen:
Fixes #5367 - allow admins to change a workflow's retirement config between `classification_count` and `never_retire`

Staging branch URL: http://pr-5377.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -59,7 +59,6 @@ class ProjectStatus extends Component {
   onChangeWorkflowRetirement(workflow, event) {
     this.setState({ error: null });
     let selected = event.target.value;
-    console.log(selected);
     return workflow.update({ 'retirement.criteria': selected }).save()
       .then(() => this.getWorkflows())
       .catch(error => this.setState({ error }));
@@ -209,7 +208,7 @@ class ProjectStatus extends Component {
                 Level:{' '}
                 <select
                   id='promotionLevels'
-                  onChange={this.onChangeWorkflowLevel.bind(null, workflow)}
+                  onChange={(event) => this.onChangeWorkflowLevel(workflow, event)}
                   value={workflow.configuration.level && workflow.configuration.level}
                 >
                   <option value="none">none</option>
@@ -232,7 +231,7 @@ class ProjectStatus extends Component {
                 Retirement:{' '}
                 <select
                   id='retirementConfig'
-                  onChange={this.onChangeWorkflowRetirement.bind(null, workflow)}
+                  onChange={(event) => this.onChangeWorkflowRetirement(workflow, event)}
                   value={workflow.retirement.criteria}
                 >
                   <option value="never_retire">Never Retire</option>
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -18,6 +18,7 @@ class ProjectStatus extends Component {
   constructor(props) {
     super(props);
     this.onChangeWorkflowLevel = this.onChangeWorkflowLevel.bind(this);
+    this.onChangeWorkflowRetirement = this.onChangeWorkflowRetirement.bind(this);
     this.getWorkflows = this.getWorkflows.bind(this);
     this.forceUpdate = this.forceUpdate.bind(this);
     this.renderWorkflows = this.renderWorkflows.bind(this);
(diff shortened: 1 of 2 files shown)


app/pages/admin/project-status.jsx
```diff
@@ -55,6 +56,15 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
+  onChangeWorkflowRetirement(workflow, event) {
+    this.setState({ error: null });
+    let selected = event.target.value;
+    console.log(selected);
```
eatyourgreens: Stray console log?

app/pages/admin/project-status.jsx
```diff
@@ -216,6 +227,18 @@ class ProjectStatus extends Component {
                   })}
                 </select>
               </label>
+              {' | '}
+              <label>
+                Retirement:{' '}
+                <select
+                  id='retirementConfig'
+                  onChange={this.onChangeWorkflowRetirement.bind(null, workflow)}
```
eatyourgreens: `onChangeWorkflowRetirement` was already bound in the constructor. If you're binding it on every render here, then the binding in the constructor is redundant.

Out of interest, why does this set the `this` context to null?

eatyourgreens: Alternatively, bind it in the constructor and change this line to call the bound function.
```suggestion
                  onChange={(event) => this.onChangeWorkflowRetirement(workflow, event)}
```

camallen: I blindly copied it from the current implementation for levels tbh, https://github.com/zooniverse/Panoptes-Front-End/blob/772d5dd2a0307919e072fa9e1cca72458b94d6d5/app/pages/admin/project-status.jsx#L201

app/pages/admin/project-status.jsx
camallen: blurgh - thanks

app/pages/admin/project-status.jsx
camallen: Shall i update the existing change bind setting as well?

eatyourgreens: Which line are you referring to? My suggested change is to only bind the handler once, in the constructor, then call that bound function whenever the selected option changes.

It needs to be bound to the component because the `this` context of a DOM event is the event target, not the React component instance. If the handler isn't bound, then `this` in the `onChange` handler will reference the `select` element.

camallen: this one, https://github.com/zooniverse/Panoptes-Front-End/blob/772d5dd2a0307919e072fa9e1cca72458b94d6d5/app/pages/admin/project-status.jsx#L201

eatyourgreens: Yeah, that should probably change too. Otherwise, it's creating a new function on each render, and the old functions don't get properly disposed of.

camallen: i've updated them both and tested locally with success. Thanks for the suggested change as well.

PR #5412, opened 2019-06-19
Add browser based validation for the real name field

### LINKED ISSUE #4553: Add validation to credited name field
## Expected behavior
Credited name field should check for and disallow email addresses being submitted.

## Current behavior
Currently we allow email addresses in this field which is a PII leak risk since project owners would like to use the credited name in publications, authors page, etc. Volunteers may not have understood what the field might be used for when they put in their email addresses.

## Additional information
This is a safety check just to make sure we don't accidentally leak PII as per GDPR requirements.

srallen:
Staging branch URL: https://pr-5412.pfe-preview.zooniverse.org/

Fixes #4553 

Adds a regex browser validation to prevent users from submitting email addresses in the real name field. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/partials/register-form.cjsx
+++ b/app/partials/register-form.cjsx
@@ -28,6 +28,7 @@ counterpart.registerTranslations 'en',
     email: 'Email address'
     emailConflict: 'An account with this address already exists'
     realName: 'Real name'
+    realNamePatternHelp: "Enter a real name, not an email address", 
     whyRealName: 'We’ll use this to give you credit in scientific papers, posters, etc'
     agreeToPrivacyPolicy: 'You agree to our %(link)s (required)'
     privacyPolicy: 'privacy policy'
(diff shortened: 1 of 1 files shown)


app/partials/register-form.cjsx
```diff
@@ -28,6 +28,7 @@ counterpart.registerTranslations 'en',
     email: 'Email address'
     emailConflict: 'An account with this address already exists'
     realName: 'Real name'
+    realNamePatternHelp: "Enter a real name, not an email address",
```
eatyourgreens:
```suggestion
    realNamePatternHelp: "Enter a name, not an email address", 
```

srallen:
Yep, built in browser validation only happens on the submit event. You should see something like this on submit (assuming the other validations pass because the validations run in order, so you have to fill in the form):

Screen Shot 2019-06-19 at 12 17 18 PM

app/partials/register-form.cjsx
srallen: I copied over what I used in the new app, so I'll have to make this change there too.

eatyourgreens: It's not a big deal. It occurred to me because I sometimes get told by web forms that O'Donnell is 'not a real name.'

srallen: It's probably best to avoid this language since folks might want to use a pseudonym or a nickname or something. 👍

PR #5461, opened 2019-07-17
Add front-end for the volunteer list page

rogerhutchings:
Adds a new about page for projects called `volunteers`, hooks up the route, and auto-creates this page when building a new project. This page will be auto-populated by the API, which @camallen is working on.

This was done during the July 17 Oxford hackday, so reviews welcome as it may be shonky

### DIFF
--- a/app/pages/lab/actions/project.js
+++ b/app/pages/lab/actions/project.js
@@ -10,7 +10,22 @@ var projectActions = {
       private: true
     }, projectData);
 
-    return projects.create(allProjectData).save();
+    return projects.create(allProjectData)
+      .save()
+      .then(function (projectResource) {
+        apiClient.post(projectResource._getURL('pages'), {
+          project_pages: {
+            language: projectResource.primary_language,
+            title: 'Volunteers',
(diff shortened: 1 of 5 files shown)


app/pages/lab/actions/project.js
```diff
@@ -10,7 +10,22 @@ var projectActions = {
       private: true
     }, projectData);
 
-    return projects.create(allProjectData).save();
+    return projects.create(allProjectData)
+      .save()
+      .then(function (projectResource) {
+        apiClient.post(projectResource._getURL('pages'), {
```
eatyourgreens: Out of interest, why don't the other project pages get set up here?

rogerhutchings: They get set up at https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/partials/project-page-editor.cjsx#L17. This page is different in that it'll be entirely autogenerated - the others have their resources created on first save.

camallen:
linked to https://github.com/zooniverse/Panoptes/pull/3155/files

PR #5473, opened 2019-07-24
Don't display private messages from deleted users

adammcmaster:
Staging branch URL: https://skip-deleted-user-messages.pfe-preview.zooniverse.org/

Messages from deleted users are broken, and often this comes up with spam messages that we don't want cluttering up the message list.

This PR simply doesn't display those messages in the inbox – this is obviously not ideal (this results in less than 10 messages being listed on the page, and the empty `div` is ugly) but it's the quickest workaround I can think of and imho it'll do for now.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/home-for-user/recent-messages.jsx
+++ b/app/pages/home-for-user/recent-messages.jsx
@@ -83,19 +83,23 @@ class RecentCollectionsSection extends React.Component {
       return [];
     })
     .then((users) => {
-      let partner = users.find((potentialPartner) => {
-        return potentialPartner !== currentUser;
-      });
-      if (partner === undefined) {
-        // Why're you talking to yourself?
-        partner = currentUser;
+      if (users.length > 0) {
+        let partner = users.find((potentialPartner) => {
+          return potentialPartner !== currentUser;
+        });
+        if (partner === undefined) {
+          // Why're you talking to yourself?
+          partner = currentUser;
+        }
+        const newState = Object.assign({}, this.state.conversationPartners);
+        newState[conversation.id] = partner;
+        this.setState({
+          conversationPartners: newState,
+        });
+        return partner;
+      } else {
+        return null;
       }
-      const newState = Object.assign({}, this.state.conversationPartners);
-      newState[conversation.id] = partner;
-      this.setState({
-        conversationPartners: newState,
-      });
-      return partner;
     });
   };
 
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Cheers! There's an inbox listing on the home page too, which might also need to be changed.
https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/home-for-user/recent-messages.jsx

Direct link to that list is https://www.zooniverse.org/#messages

adammcmaster:
@eatyourgreens I've updated that one too now.

app/pages/home-for-user/recent-messages.jsx
```diff
@@ -83,19 +83,23 @@ class RecentCollectionsSection extends React.Component {
...
-      if (partner === undefined) {
-        // Why're you talking to yourself?
-        partner = currentUser;
+      if (users.length > 0) {
+        let partner = users.find((potentialPartner) => {
+          return potentialPartner !== currentUser;
+        });
+        if (partner === undefined) {
+          // Why're you talking to yourself?
+          partner = currentUser;
```
eatyourgreens: Ha! I'd been wondering why my name shows up on messages from deleted users.

app/pages/home-for-user/recent-messages.jsx
```diff
@@ -131,24 +135,31 @@ class RecentCollectionsSection extends React.Component {
...
+          avatarState[author.id] = avatar;
+          this.setState({
+            avatars: avatarState,
+          });
+          return author;
         });
-        return author;
-      });
+      } else {
+        return [];
```
eatyourgreens: I think `author` is an object so this should return `{}` or null. The catch block returns an array because that's the default for a list of avatars.

app/pages/home-for-user/recent-messages.jsx
```diff
@@ -131,24 +135,31 @@ class RecentCollectionsSection extends React.Component {
...
+      }
     });
   };
 
   renderConversation = (conversation, index, allConversations) => {
     const partner = this.state.conversationPartners[conversation.id];
+    if (!partner) {
+      return (<div></div>);
```
eatyourgreens:
```suggestion
      return null;
```
I'm pretty sure JSX allows you to return null if you don't want to render an element at all.

eatyourgreens:
I have a message from a deleted user in my current inbox, and this PR does hide it.

One question: how would I go about deleting that message?

adammcmaster:
There isn't a way to delete that message. The delete button is broken for those messages anyway, so that's actually the case even without this PR.

Issue #5332, opened 2019-04-24
Colors of the Feedback tool


noraeisner:
## Expected behavior

It would be great if a different combination of colors could be used for the feedback tool. At the moment it is red for false and green for true, however, red/green colour blindness is quite common. Can this be changed? @rogerhutchings @eatyourgreens ?

eatyourgreens:
Looks like the colours are currently hardcoded in the feedback strategy code. It would be easy enough to change the values here: https://github.com/zooniverse/Panoptes-Front-End/blob/2f7c55797bc9837ae5f0e6407d7c28d78b7bbe9b/app/features/feedback/shared/strategies/drawing/column/feedback-mark.jsx#L4-L5

srallen:
@beckyrother any thoughts on the colors to use? I think we should include a small guide in the feedback for whatever colors get used, red and green have cultural meaning for fail and success, so if we use something else, we'll need to communicate that.

beckyrother:
From the survey it seems like it's just not clear what the feedback is indicating; maybe it would help to also include icons over the markings, similar to where the `X` appears while doing the task. As Sarah mentioned the red and green is nearly universal (with some exceptions) in meaning fail and success. You're right about the colorblindness issue but adding additional information rather than relying solely on color is a more effective way to ensure legibility.

eatyourgreens:
@beckyrother I think this issue is referring to the PFE version of Planet Hunters: https://www.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/classify

Here's a screenshot of the feedback in PFE: 
![A screenshot of feedback for a simulated transit in Planet Hunters Tess.](https://user-images.githubusercontent.com/59547/56680122-84d6cd80-66be-11e9-95e8-99015858cd61.png)
For that one, I wasn't really sure if it's telling me that I spotted a transit or that I missed one. I think I got the right place in the light curve, but the mark I drew was too broad, so red for a miss?

beckyrother:
I think **icons and maybe a key** will help this. In doing this myself I also didn't draw exactly the right size so I got a red. What does the blue mean?

eatyourgreens:
Icons would definitely make the feedback easier to understand. Something like a tick for success, cross for failure?

[...]

PR #5475, opened 2019-07-24
Improved feedback

hughdickinson:
Staging branch URL: https://pr-5475.pfe-preview.zooniverse.org

Enhances experimental realtime feedback. New functionality:

* Identical feedback messages corresponding to multiple feedback marks can be pluralized using the special `$count}` placeholder.  This is configurable in the project builder. See `app/features/feedback/README.md`.
* For the radial tool, the feedback mark colors are customizable (independently for success and failure cases). This is configurable in the project builder. 
* For the radial tool, the feedback mark opacity is customizable. This is configurable in the project builder. 
* For the radial tool different shapes can be used to mark successes and failures. This is configurable in the project builder. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


hughdickinson:
> Enhances experimental realtime feedback. New functionality:
> 
> > Identical feedback messages corresponding to multiple feedback marks can be pluralized using the special $count} placeholder. This is configurable in the project builder. See app/features/feedback/README.md.
> 
> I'm not sure if it is the right approach to add another configurable message using subject metadata. This is adding more complexity to the configuration that some project builders have already struggled with. In the rewrite, we add a reduced count of the number of times a success or fail happens next to its corresponding message. See: [zooniverse/front-end-monorepo#892](https://github.com/zooniverse/front-end-monorepo/pull/892)
> 
> I think it is a good idea to have a reduced count, but I think we should consistent with what we're doing in the rewrite. Do you have a reasoning as to why this might be a better approach? Does this approach outweigh the difficulty project owners may have in configuration?

My motivation for implementing it this way (with a plural message and a singular message) was to allow for cases when there can be one or more successes or failures. In that case, providing a meaningful message may involve pluralization of nouns as well as just adding a number e.g. 

"You found an object" becomes "You found 10 objects". 

This allows more natural sounding feedback and I imagine it would also be more flexible for different languages although I'm not certain about that.
 
> > For the radial tool, the feedback mark colors are customizable (independently for success and failure cases). This is configurable in the project builder. For the radial tool, the feedback mark opacity is customizable. This is configurable in the project builder.
>
> We generally don't allow style customizations by project owners. Could you explain the rational for allowing this?

There are two reasons.
1. Hard-coded colours may be invisible on certain subjects.
2. The default color scheme in this case was red/green which is not great for anyone with red-green colour blindness.

> > For the radial tool different shapes can be used to mark successes and failures. This is configurable in the project builder.
> 
> Again this is adding configuration complexity that may be difficult for project builders to understand and configure. This is a good idea, however, I think we can just have a non-configurable general solution for this.

At the moment this is the least configurable of all the changes. It's just a simple on/off toggle. I can imagine situations for which users may not want to show squares for failures on the radial feedback tool, but for my purposes I'd be happy to hard code it to be on.

> Is there a project we can manually test these enhancements with?

I had expected the branch to be automatically staged when I submitted the PR, but it seems not to have been. Will this happen? I can manually stage it if not. Once it is staged, this project/workflow demonstrates the functionality

`/hughdickinson/galaxy-zoo-clump-scout/classify?workflow=11223`

app/features/feedback/README.md
```diff
@@ -14,15 +15,15 @@
 ### Workflow
 
 1. In the project builder, navigate to your workflow, and find the feedback section in the bottom right.
-1. Start creating a new rule by clicking the button.
-1. Define a unique id for the rule, check the boxes for whether you want feedback to be shown on success, failure or both. Where enabled, a default message must be defined.
-1. Select a strategy for your rule. See the Strategies section below for more information.
-1. Define any additional options required by the chosen strategy, and save the new rule.
+2. Start creating a new rule by clicking the button.
```
eatyourgreens: These numbers are calculated by the Markdown parser. The original code was the markdown for a numbered list.

app/features/feedback/lab/components/feedback-section.jsx
```diff
@@ -12,61 +17,147 @@ counterpart.registerTranslations('en', {
...
+
+        <CheckboxInput title={fieldText('pluralFailureMessagesEnabled.title')} help={fieldText('pluralFailureMessagesEnabled.help')} name="pluralFailureMessagesEnabled" checked={this.state.form.pluralFailureMessagesEnabled} onChange={this.handleInputChange.bind(this)}/>
+
+        <CheckboxInput title={fieldText('colorizeUniqueMessagesEnabled.title')} help={fieldText('colorizeUniqueMessagesEnabled.help')} name="colorizeUniqueMessagesEnabled" checked={this.state.form.colorizeUniqueMessagesEnabled} onChange={this.handleInputChange.bind(this)}/> {this.state.form.colorizeUniqueMessagesEnabled && <SelectInput title={fieldText('allowedSuccessFeedbackMarkerColors.title')} help={fieldText('allowedSuccessFeedbackMarkerColors.help')} options={colorOptions} name="allowedSuccessFeedbackMarkerColors" multi={true} value={this.state.allowedSuccessFeedbackMarkerColors} placeholder={fieldText('allowedSuccessFeedbackMarkerColors.placeholder')} onChange={this.handleAllowedSuccessFeedbackMarkerColors.bind(this)}/>}
```
eatyourgreens: Long lines like this should be flagged by eslint. Can you lint your changes, please?

app/features/feedback/lab/containers/feedback-section-container.jsx
```diff
@@ -42,8 +44,22 @@ class FeedbackSectionContainer extends Component {
         rules: []
       }
     });
-
+    console.log(editedTask.rules);
```
eatyourgreens: Stray console log.

app/features/feedback/lab/containers/feedback-section-container.jsx
```diff
@@ -42,8 +44,22 @@ class FeedbackSectionContainer extends Component {
...
     editedTask.feedback.rules = _.unionBy([rule], editedTask.feedback.rules, 'id');
+    console.log(editedTask.feedback.rules);
```
eatyourgreens: Stray console log.

app/features/feedback/shared/helpers/get-global-feedback-options-from-rules.jsx
```diff
@@ -0,0 +1,24 @@
...
+  }).value();
+  const colorizeUniqueMessagesEnabled = _.chain(rules).some(rule => {
+    return rule.colorizeUniqueMessagesEnabled;
+  }).value();
+  const successFailureShapesEnabled = _.chain(rules).some(rule => {
+    return rule.successFailureShapesEnabled;
+  }).value();
+  const allowedSuccessFeedbackMarkerColors = _.chain(rules).some().value() ? rules[0].allowedSuccessFeedbackMarkerColors : getPfeMarkerColors();
+  const allowedFailureFeedbackMarkerColors = _.chain(rules).some().value() ? rules[0].allowedFailureFeedbackMarkerColors : getPfeMarkerColors();
+  return {pluralSuccessMessagesEnabled: pluralSuccessMessagesEnabled, pluralFailureMessagesEnabled: pluralFailureMessagesEnabled, colorizeUniqueMessagesEnabled: colorizeUniqueMessagesEnabled, successFailureShapesEnabled: successFailureShapesEnabled, allowedSuccessFeedbackMarkerColors: allowedSuccessFeedbackMarkerColors, allowedFailureFeedbackMarkerColors: allowedFailureFeedbackMarkerColors};
```
eatyourgreens: This needs linting too. Redundant property declarations in objects should be caught by the linter.

package-lock.json
```diff
@@ -9,16 +9,16 @@
...
-        "commander": "^2.8.1",
-        "convert-source-map": "^1.1.0",
-        "fs-readdir-recursive": "^1.1.0",
-        "glob": "^7.0.0",
-        "lodash": "^4.17.10",
-        "mkdirp": "^0.5.1",
-        "output-file-sync": "^2.0.0",
-        "slash": "^2.0.0",
-        "source-map": "^0.5.0"
+        "chokidar": "2.0.4",
```
eatyourgreens: This PR doesn't install or update any packages, so it shouldn't be changing the package lock file.

eatyourgreens:
As a volunteer on the test workflow, how do I know which types of clump I've missed? I'm seeing red squares and a message along the lines of `you missed 3 clumps` but I don't know which of the two types of drawing tool this refers to.

eatyourgreens:
For anyone else who'd like to test this, the link is https://pr-5475.pfe-preview.zooniverse.org/projects//hughdickinson/galaxy-zoo-clump-scout/classify?workflow=11223&env=production and you must be in admin mode.

eatyourgreens:
Yellow feedback is impossible to read on a white background.
Screenshot showing a feedback modal with yellow text on a white background.

eatyourgreens:
My initial reaction is that this is an interesting idea but it needs more work before it's ready to be used on a project. Ideally, it should be added to feedback in [the new classifier](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier), since we are moving projects across to using that code.

srallen:
> My motivation for implementing it this way (with a plural message and a singular message) was to allow for cases when there can be one or more successes or failures. In that case, providing a meaningful message may involve pluralization of nouns as well as just adding a number

I think the idea to support interpolation is a good one, but I don't think we should add yet another subject metadata field just for pluralization support. We can support this in the default messaging fields for rules that can be setup. Since the message will be the same, the default message can be used rather than use the subject metadata field messaging. Pretty much all the projects that have used feedback so far have misunderstood the subject metadata field for the message as being required when it's not. I'm afraid this would happen with this new field as well.

> Hard-coded colours may be invisible on certain subjects.
> The default color scheme in this case was red/green which is not great for anyone with red-green colour blindness.

The onus for good accessible design is on us, not the project builders. Allowing the colors to be customizable doesn't actually solve the accessibility problem because we're solely using color to communicate information. What the change actually should be is using color plus something else to indicate success or failure like icons or keys. This was discussed in https://github.com/zooniverse/Panoptes-Front-End/issues/5332

eatyourgreens:
I think I've got an issue (or a comment on an issue?) somewhere where I outline how string interpolation might work for feedback. I can't find it, though. It might have been a suggestion for the new classifier.

srallen:
Yep, we talked about it here: https://github.com/zooniverse/front-end-monorepo/issues/774

hughdickinson:
Thanks all for the feedback. I agree that doing this properly requires a lot of careful work. Ultimately, since the MVP for our project is that multiple identical messages get reduced, I'll close this PR and open a new one that *only* performs the reduction and renders the messages in the same way as the current monorepo example (with the repeat count in parentheses after the message).

PR #5485, opened 2019-08-05
Translate `hide previous marks` tool

### LINKED ISSUE #5484: `Hide previous marks` not available for translation
## Expected behavior

`Hide previous marks` should be available for translation.

## Current behavior

It's not available in Pandora (expected, as it's not related to a project-specific resource), nor in the app-wide set of locale strings (https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/locales/en.js)

## Notes

Looks like this one just got missed out when adding in counterpart: https://github.com/zooniverse/Panoptes-Front-End/blob/da2aa6f913d88f00d457f7f159d81e5c82ca4bf7/app/classifier/tasks/drawing/hide-previous-marks-toggle.jsx#L72

rogerhutchings:
- Add an EN and CS translation for `hide previous marks`, with count interpolation
- Remove text string from tool and use `Translate` component instead

Fixes #5484

### DIFF
--- a/app/classifier/tasks/drawing/hide-previous-marks-toggle.jsx
+++ b/app/classifier/tasks/drawing/hide-previous-marks-toggle.jsx
@@ -1,5 +1,6 @@
-import React from 'react';
 import PropTypes from 'prop-types';
+import React from 'react';
+import Translate from 'react-translate-component';
 
 class HidePreviousMarksToggle extends React.Component {
   constructor() {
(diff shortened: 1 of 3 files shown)


yshish:
@shaunanoordin 
> but I couldn't find any real project that uses this? Not even on `/dev/classifier`? 

Hey, We definitely use that in Penguin Watch and also in SeabirdWatch - I'm working on the translations now. Not sure if more projects have it though.

shaunanoordin:
👍 Thanks, @yshish - I've tested this on [localhost for Seabirdwatch](https://localhost:3735/projects/penguintom79/seabirdwatch/classify?env=production)

I can now confirm that the changes maintain feature parity with English and correctly feature the additions for Czech.

Screen Shot 2019-08-05 at 16 24 58

PR #5518, opened 2019-10-01
Fix volunteer count via constant offset

### LINKED ISSUE #4043: Registered volunteer count on home page is low
According to @mrniaboc we have 1,678,753 registered volunteers; the home page shows a lower number. 

Also, please add more space between the number and "REGISTERED VOLUNTEERS" 

zooniverse

lcjohnso:
Staging branch URL: https://pr-5518.pfe-preview.zooniverse.org

Fixes #4043 following style of #4511 solution.

Offset between true DB user count and the current/old displayed value is because the old count excluded Ouroboros users (where `USERS.OUROBOROS_CREATED` is true).  That population is currently 124921, so that is the offset used in this fix.  I'll note that this number appears to fluctuate with time (?): it was 126571 on 17 May 2019 and ~118600 in Sep 2017.

Edit to initialize counter to 0 in app/pages/home-not-logged-in.jsx follows fix in #4526.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/home-not-logged-in.jsx
+++ b/app/pages/home-not-logged-in.jsx
@@ -31,7 +31,7 @@ export default class HomePage extends React.Component {
       featuredProject: null,
       promotedProjects: [],
       screenWidth: 0,
-      volunteerCount: 1500000
+      volunteerCount: 0
     };
 
     this.showDialog = this.showDialog.bind(this);
--- a/app/pages/home-not-logged-in/research.jsx
+++ b/app/pages/home-not-logged-in/research.jsx
@@ -41,6 +41,7 @@ counterpart.registerTranslations('en', {
 const GZ123_COUNT = 98989226
 const OUROBOROS_COUNT = 142800311
 const OTHERS_COUNT = 8680290
+const OUROBOROS_USER_COUNT = 124921
 
 
 const HomePageResearch = (({ count, screenWidth, showDialog, volunteerCount }) =>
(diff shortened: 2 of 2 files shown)


app/pages/home-not-logged-in/research.jsx
```diff
@@ -41,6 +41,7 @@ counterpart.registerTranslations('en', {
 const GZ123_COUNT = 98989226
 const OUROBOROS_COUNT = 142800311
 const OTHERS_COUNT = 8680290
+const OUROBOROS_USER_COUNT = 124921
```
camallen: Thanks for naming this well :)

Pity, past us didn't name the other counts with the context of what they are counting :(

app/pages/home-not-logged-in.jsx
```diff
@@ -31,7 +31,7 @@ export default class HomePage extends React.Component {
       featuredProject: null,
       promotedProjects: [],
       screenWidth: 0,
-      volunteerCount: 1500000
+      volunteerCount: 0
```
camallen: So this now means the counter starts at 0 on the home page until the API request for users completes?

lcjohnso: I think that may be true.  But as we found out during the classification count work (see #4526), this initialization value was being added to the other constant values -and- the returned value, so I mimicked that fix to be sure such an offset didn't occur here.  Not exactly sure how/when the update from initialized value versus API requested value occurs, but know this is the safe route.

PR #5528, opened 2019-10-17
Fix 5527: prevent double annotations on touchscreen marking tasks

### LINKED ISSUE #5527: On touchscreen, "click to add annotation" creates two annotations
## UI Bug

On touchscreen devices, the "click to add annotation" action creates two annotations instead of one. Additionally, it is VERY difficult to select an existing annotation.

Tested on OSX+Chrome (emulating touchscreens) on https://www.zooniverse.org/projects/roxannebeltran/ano-nuevo-island-animal-count/classify

Original report: https://www.zooniverse.org/talk/17/1146166

### Expected behavior

- on touch screen devices, on projects where clicking on the Subject creates an annotation mark,
  - clicking on an empty space should create one (1) mark
  - clicking on an existing mark should select that mark

### Current behavior

- on touch screen devices,
  - clicking on an empty space should creates two (2) marks
  - clicking on an existing mark _sometimes_ creates two (2) marks

### Steps to replicate

1. open Chrome, enable touchscreen emulation - or use a native touchscreen device
2. go to https://www.zooniverse.org/projects/roxannebeltran/ano-nuevo-island-animal-count/classify and begin the main Task that uses Drawing Tool, point marker.
3. start clicking.

### Status

Issue is confirmed to exist. Investigation into the cause is ongoing.

Investigating:
- Is this occurring on other projects?
- What's causing the action to trigger twice with a touch interface?
- What's the difference between a touch interface and a mouse interface?
- Why is it difficult to select an existing annotation?
- When did this start happening?

shaunanoordin:
## PR Overview

Fixes #5527 

Staging branch URL: https://pr-5528.pfe-preview.zooniverse.org

Currently, if we attempt to perform a marking task on a touchscreen device, _two_ annotation marks are created instead of one. This is caused by the `onPointerDown` and `onTouchStart` handlers (in the `Draggable` class) firing simultaneously when a touchscreen device detects touch input.

This PR fixes that issue by checking if Pointer Events exist on the browser, and if available, using it instead of Mouse Events or Touch Events. This prevents redundant events from firing.

(Note: we shouldn't remove Mouse Events or Touch Events altogether, since they're good fallbacks for less modern browsers.)

### Required Manual Testing

### Review Checklist

- [x] Does it work on mobile?

### Optional Checks

### Status

Ready for review, and this might be of interest to @eatyourgreens and @simensta

### DIFF
--- a/app/lib/draggable.jsx
+++ b/app/lib/draggable.jsx
@@ -18,8 +18,6 @@ function Draggable(props) {
     const multiTouch = e.touches && e.touches.length > 1;
 
     if (!multiTouch) {
-      e.preventDefault();
-
       switch (e.type) {
         case 'mousedown':
           [moveEvent, endEvent] = ['mousemove', 'mouseup'];
@@ -46,6 +44,8 @@ function Draggable(props) {
         startHandler(eventCoords);
       }
     }
+    
+    e.preventDefault();
   }
 
   function handleDrag(e) {
@@ -76,6 +76,7 @@ function Draggable(props) {
   }
 
   const { children, disabled } = props;
+  const usePointer = !!window.PointerEvent;
   const className = classnames({
     [children.props.className]: true,
     draggable: true
@@ -83,9 +84,9 @@ function Draggable(props) {
   const childProps = {
     className,
     'data-disabled': disabled ? true : undefined,
-    onMouseDown: disabled ? undefined : handleStart,
-    onTouchStart: disabled ? undefined : handleStart,
-    onPointerDown: disabled ? undefined : handleStart
+    onMouseDown: (!disabled && !usePointer) ? handleStart : undefined,
+    onTouchStart: (!disabled && !usePointer) ? handleStart : undefined,
+    onPointerDown: (!disabled && usePointer) ? handleStart : undefined
   };
   return React.cloneElement(children, childProps);
 }
--- a/app/lib/draggable.jsx
+++ b/app/lib/draggable.jsx
@@ -75,8 +75,7 @@ function Draggable(props) {
     }
   }
 
-  const { children, disabled } = props;
-  const usePointer = !!window.PointerEvent;
+  const { children, disabled, usePointer } = props;
   const className = classnames({
     [children.props.className]: true,
     draggable: true
@@ -98,14 +97,16 @@ Draggable.propTypes = {
   ]),
   onDrag: PropTypes.func,
   onEnd: PropTypes.func,
-  disabled: PropTypes.bool
+  disabled: PropTypes.bool,
+  usePointer: PropTypes.bool
 };
 
 Draggable.defaultProps = {
   onStart: false,
   onDrag: () => false,
   onEnd: () => false,
-  disabled: false
+  disabled: false,
+  usePointer: !!window.PointerEvent
 };
 
 export default React.memo(Draggable);
--- a/app/lib/draggable.jsx
+++ b/app/lib/draggable.jsx
@@ -18,6 +18,8 @@ function Draggable(props) {
     const multiTouch = e.touches && e.touches.length > 1;
 
     if (!multiTouch) {
+      e.preventDefault();
+      
       switch (e.type) {
         case 'mousedown':
           [moveEvent, endEvent] = ['mousemove', 'mouseup'];
@@ -44,8 +46,6 @@ function Draggable(props) {
         startHandler(eventCoords);
       }
     }
-    
-    e.preventDefault();
   }
 
   function handleDrag(e) {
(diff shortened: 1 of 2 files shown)


shaunanoordin:
Wait, I still need to update the tests!

shaunanoordin:
## Resources / Further Reading

W3C gives a good example of how to handle Pointer Events in an environment where Mouse Events and Touch Events might still trigger.
https://w3c.github.io/pointerevents/#examples

patrickhlauke's tests on touch/pointer events gives a very thorough breakdown of how these events interact with/prioritise over each other.
https://patrickhlauke.github.io/touch/tests/results/

eatyourgreens:
The idea looks good but I think this needs tests.

shaunanoordin:
The challenge I have with updating `draggable.spec.js` is that the whole `describe('Draggable')` block operates assuming that there's only one kind of browser window (with no PointerEvent), whereas the new `draggable.jsx` checks for two different possibilities (a window with PointerEvent, and one without).

The _rough_ solution to this is to have `describe('Draggable (old browser compatibility)')` and `describe('Draggable (browser with Pointer Events)')`.

The former runs `before(() => delete global.window.PointerEvent; ) testDrag('mousedown', 'mousemove', 'mouseup'); testDrag('touchstart', 'touchmove', 'touchend');` and the latter runs `before(() => global.window.PointerEvent = {}; ) testDrag('pointerdown', 'pointermove', 'pointerup');`, so we're basically testing for two different environments.

Problem is, the _rough_ solution will create a lot of code duplication, which I'd rather avoid.

eatyourgreens:
Can you pass the a boolean into `testDrag()` as an additional argument that sets `window.pointerEvent`? It's been a while since I wrote those tests.

shaunanoordin:
> Can you pass the a boolean into testDrag() as an additional argument that sets window.pointerEvent? It's been a while since I wrote those tests.

Unfortunately, no - I tried that at first. Turns out I need to set the window variable before the `wrapper` is declared, i.e. ...

```
before(function() {
  ...
  window.PointerEvent = {};
  wrapper = shallow(
    <Draggable
      onStart={onStart}
      onDrag={onDrag}
      onEnd={onEnd}
    >
      <p>Hello</p>
    </Draggable>
  );
  ...
});
```

I'm not sure how I can change the 'before()' code in between the `testDrag()` calls.
```
// use a before() version that has window.PointerEvent = undefined
testDrag('mousedown', 'mousemove', 'mouseup');
testDrag('touchstart', 'touchmove', 'touchend');
// change to a before() version that has window.PointerEvent = {}
testDrag('pointerdown', 'pointermove', 'pointerup');
// Have some code to reset global.window.PointerEvent to its original state in case other tests need it.
```

eatyourgreens:
Good point. I'll have a look at those tests tomorrow. Mouse and touch are tested now when pointer events aren't available, but the tests don't check that mouse and touch are ignored when pointer events are enabled. In my head, the spec looks something like this:

- set `window.pointerEvent`
  - fire mock `pointerdown`, `mousedown` and `touchstart` events.
  - expect `pointerdown` to be handled but the others to be ignored.
- remove `window.pointerEvent`
  - fire mock `pointerdown`, `mousedown` and `touchstart` events.
  - expect `pointerdown` to be ignored but the others to be handled.

Does that mean splitting the tests for setting the event handlers (on start) out of the `testDrag` function and testing those separately? The tests for `move` and `end` don't change because those call the handlers directly.

shaunanoordin:
## PR Update

I've added a very rough update to `draggable.spec.js` that splits `describe('Draggable')` to two versions - one version where the window is a "classic browser" without PointerEvent support, and another version is a "modern browser" with support.

Tests now pass, but I welcome any suggestions on how to make the test code more streamlined.

As pointed out in the above posts, the key challenge here is that the _"this window has/doesn't have PointerEvent support"_ logic needs to run _before_ the `wrapper = <Draggable>` is created.)

shaunanoordin:
### Testing

To test this on touchscreen mobile devices, you can...
- go to https://pr-5528.pfe-preview.zooniverse.org/projects/deleted-1845/i-fancy-cats/classify
- go to the "Draw Cattail" Task
- each click should only create one annotation mark. (The Task panel counts the number of existing annotations.)
- for comparison, go to https://master.pfe-preview.zooniverse.org/projects/deleted-1845/i-fancy-cats/classify and note that a single click creates two annotation marks.

eatyourgreens:
My phone doesn’t support pointer events so I get the same results on both branches. Anyone testing this on iOS might need iOS 13.

eatyourgreens:
I've updated the tests in #5529.

eatyourgreens:
This looks good. Should we deploy it today or wait until Monday?

shaunanoordin:
Good point - let's not risk a Friday deploy. This PR is now scheduled for a Monday merge. In the meantime, I'm going to take a look at how existing Classifications might have been affected by the bug.

A big 👍 for Jim and Andrea for help with this PR

Issue #5531, opened 2019-10-18
Confusing UX with the tutorial and screen readers


srallen:
## Current behavior
_Please include any error messages from the browser console and/or screenshots_

Original report:
https://www.zooniverse.org/projects/anaelisa24/sounds-of-new-york-city-sonyc/talk/1795/1143269?comment=1867713

https://www.zooniverse.org/talk/18/1145904

> Hi, I'm totally blind and thought I'd try your project. I'm using Windows 10 and the NVDA screen reader. First I tried Firefox, that was a disaster, it froze my screen reader when I pressed the done and talk button, so I had to bring up Narrator and close Firefox completely. Also many things were hidden. I then tried Chrome and discovered I was missing important things with Firefox like the near and far questions entirely. They just weren't showing. So with Chrome I can play a sound, read through the choices, and pick one and click "Identify" and go on to the next sound I heard in the recording. After I've identified all the sounds I can press done and listen to another sound. But there's one problem. There's something on the page that seems to do absolutely nothing, and that is the 7 radio buttons for steps 1 through 7. I can't even guess what these are for because when I select #2 nothing changes. Also there is a button labelled task, and one labelled tutorial. These buttons also seem to do nothing. I thought when I pressed the tutorial button I would find a tutorial about how the site works, and if there's anything I need to do that I am missing, but I found no such thing. The one time I pressed done and talk and went to the talk page I thought that was what you were supposed to do each time you were done with a sound. One more thing. When using Chrome I identified a reverse beep, and it said something about the sound graph and how a reverse beep should look, and of course I can't see that, but it was obviously a truck backing up that I heard. But the curious thing was that there were 2 radio buttons for the reverse beep with no explanation or alt-text explaining what those were supposed to be. They just said 1 and 2. I assume there are supposed to be 2 kinds of reverse beeps and I'm supposed to identify which type it is, but I had no idea what to do there so I just left the 1 radio button checked.
Thanks.

## Additional information

- **Operating system:**
Windows 10
- **Browser:**
Firefox and Chrome with NVDA (https://www.nvaccess.org/)

eatyourgreens:
There’s a few different things going on here. 
- NVDA + Firefox misses a bunch of stuff, including the near and far questions for each sound. 
- Task and Tutorial aren’t marked up as tabs.

eatyourgreens:
[...]

PR #5535, opened 2019-10-25
Improve tutorial accessibility

eatyourgreens:
Render tutorial as a section, not a form.
Add a live region to tutorial step content.
Add more descriptive labels to the step controls.

Towards #5531.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/classifier/tutorial.jsx
```diff
@@ -290,19 +306,20 @@ export default class Tutorial extends React.Component {
...
             <span className="step-through-pips">
-              {allSteps.map(thisStep =>
+              {allSteps.map(thisStep => (
                 <label key={thisStep} className="step-through-pip" title={`Step ${thisStep + 1}`}>
                   <input
                     type="radio"
+                    name="tutorial-step"
                     className="step-through-pip-input"
-                    aria-label={`Step ${thisStep + 1} of ${totalSteps}`}
+                    aria-label={`Tutorial step ${thisStep + 1} of ${totalSteps}`}
```
eatyourgreens: Ideally, this text would be localised but we don't have any multi-lingual audio projects at the moment.

app/classifier/tutorial.jsx
```diff
@@ -263,10 +272,17 @@ export default class Tutorial extends React.Component {
...
+          <MediaCard
+            src={mediaCardSrc}
+          >
+            <Markdown>{this.props.translation.steps[this.state.stepIndex].content}</Markdown>
+            <hr />
+          </MediaCard>
+        </div>
         <p style={{ textAlign: 'center' }}>
           {(this.state.stepIndex === this.props.tutorial.steps.length - 1) ?
             <button type="submit" className="major-button">
```
simensta: I can press the "Let's go!" button on VoiceOver (i.e., control + option + space), but the action does not close the tutorial.

simensta:
On Chrome and Safari using VoiceOver, I've tested the new branch on the Sounds of New York City (SONYC) project. Here is my experience so far:
1. tab through the Classify page to the _Tutorial_ button
2. press the _Tutorial_ button with control+option+space
3. The tutorial modal renders and VoiceOver announces that the focus is on 1 of 7 radio buttons. However, the text content of the slide is not read.
4. Use the arrow keys to advance a slide (I was also able to use the _Continue_ button to navigate). Once the slide is advanced, VoiceOver will start reading the contents of the slides. To hear the content from the first slide, I have to advance to slide 2 and then back to slide 1. Is there a way to prompt VoiceOver to read the first slide once it is loaded?
5. Once I reach the last radio button, it is difficult to figure out how to exit the tutorial. The "Let's go!" button does not respond to control+option+space. I was able to use control+option+space to press the _X_ button to close the Tutorial. This dovetails with a volunteer's report that placing text above the radio buttons is not intuitive.

@eatyourgreens, do you think that the issues raised in 4 and 5 should be fixed in this PR? Or should we address 4 and 5 in a different PR?

eatyourgreens:
Let’s Go! is a submit button but I’ve removed the form, because reading the tutorial as a form would be really confusing. That submit button needs to be fixed in this PR. 

Fixing the reading focus so that it starts at the beginning of each step would be great too.

app/classifier/tutorial.jsx
eatyourgreens: It turns out `Dialog.alert()` only works if the modal content is a form. It hijacks the form's `submit` event to close the tutorial. Now that I've replaced `<form>` with `<section>`, that doesn't work.

The solution is to remove `Dialog.alert()` completely.

eatyourgreens:
The Let's Go! button is fixed now but focus still skips past the tutorial content straight to the radio buttons.

Wrapping the tutorial in `<div role='dialog'>` reads out the content when the dialog opens. That role is already declared in the modal-form code, but is never added to the rendered HTML. I'm not going to fix that modal-form bug here, as I believe it's already fixed in the new tutorial.

eatyourgreens:
I think this PR is now about as good as it's going to get. Big changes, like not using radio buttons for navigation, should go into the new tutorial and then, as @srallen suggests on #5531, we can look at importing the new tutorial into PFE and ditching the old modal-form component.

PR #5537, opened 2019-10-25
Adjust drawing tool drag handle size

### LINKED ISSUE #5536: Drawing tools are difficult to can't be resized or adjusted at all on mobile devices
Reviewing the feedback from the Planet Four beta we find a pattern with tablet and mobile devices. The feedback is that the volunteers couldn't complete the task or found it difficult or not possible at all to resize the drawing tools. 
Here's the comments received from those that said they were using Android or Apple mobile device:

> Make more mobile friendly.; Very difficult to place markers on my phone. Much better on desktop with mouse amd full screen. Unable to effectively use on Android mobile.

> Using a mouse or stylus to interface with the images. With this sort of task, one needs to be able to deal with the features quickly: at a pace more-compatible with the cerebral response.' My finger hid what I was trying to define. Also, the size of the finger rendered fine adjustment of the relatively-tiny images almost impossible. I only attempted two photographs before becoming tense and fatigued. Using a stylus would have made the task MUCH easier/ working on a larger screen, with more-enlarged images, would have rendered the task more-manageable

> Didnt seem to work

I used my iphone and ipad to classify and took videos of my interactions/movements. You can find them [here](https://www.dropbox.com/s/gn74k4ti1mgk7eh/img_6209.mov?dl=0) and [here](https://www.dropbox.com/s/35raxpnnri0fm4i/video%20oct%2025%2C%209%2018%2019%20am.mov?dl=0). I had the same issues as reported above and couldn't resize the marking shapes most of the time I tried to. 

The old Planet Four had a mobile version and the adjustment handles for the marking tools were as big as the (x) is is on the shape now to allow people to grab them and adjust the size. I think the handles need to be made bigger on the mobile versions to make drawing at least possible on mobile.

eatyourgreens:
Multiply the drawing tool drag handle size by 2.5 on screens smaller than 900px.

Staging branch URL: https://pr-5537.pfe-preview.zooniverse.org

Fixes #5536.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/drawing-tools/drag-handle.cjsx
+++ b/app/classifier/drawing-tools/drag-handle.cjsx
@@ -2,8 +2,8 @@ React = require 'react'
 createReactClass = require 'create-react-class'
 Draggable = require('../../lib/draggable').default
 
-RADIUS = 4
-OVERSHOOT = 4
+RADIUS = if screen.width > 900 then 4 else 10
+OVERSHOOT = if screen.width > 900 then 4 else 10
 
 module.exports = createReactClass
   displayName: 'DragHandle'


eatyourgreens:
I took the 900px from somewhere else in the drawing code. Maybe the size of the delete button? You’re right, the breakpoint shouldn’t be duplicated throughout the code like this.

PR #5553, opened 2019-11-19
Ellipse region feedback

hughdickinson:
Staging branch URL: https://pr-5553.pfe-preview.zooniverse.org

Describe your changes.

Implemented an elliptical region hit tester for the experimental feedback tool. Use context is training for the Radio Galaxy Zoo: LOFAR project (8190). Volunteers are asked to click within proposed elliptical regions on subject images to validate or invalidate them. A subset of such images will be classified by experts and used as training. The intention is that volunteers will be informed whether they have clicked on the correct ellipses and/or missed any that should have been clicked.

The first task of workflow [12686](https://pr-5553.pfe-preview.zooniverse.org/projects/chrismrp/radio-galaxy-zoo-lofar/classify?reload=0&workflow=12686?env=production) has been configured to demonstrate the new functionality using simple dummy subjects. Examples of the real subjects that require classification can be seen in the project's active workflow (11973).

The new tool is largely based on the radial hit-testing tool **but** the radial tool itself is not fit for this purpose. Also, the functionality will be required in January 2020 i.e. likely before the new monorepo classifier launches (and somewhat earlier beta review of the project would be nice), hence PFE is the target of this PR.

Tests based on those developed for the radial feedback tool have been implemented and all pass or expfail.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


hughdickinson:
Hi @srallen,

Thanks for the comments. Responses inline. 

> > The new tool is largely based on the radial hit-testing tool but the radial tool itself is not fit for this purpose.
> 
> Could you provide some details about why this is the case? As far as I can tell, the differences are:
> 
> * `createRule` function: ellipse includes theta. radial includes tolerance.
> * reducer: ellipse only checks if the annotation is within the ellipse. radial within the tolerance
> * UI for the lab: ellipse UI does not have the text input to define a tolerance
> * mark: SVG ellipse vs SVG circle
> 

The primary reason that the radial tool is not sufficient for our project is that the target region is explicitly elliptical and the ellipse has precise dimensions and orientation. It is not good enough to know whether a volunteer clicked within a circle centred on a point.

An important difference that you missed is that the ellipse is parameterised by two "radii", namely "a" and "b" defining the semimajor and semi minor axis lengths of the target ellipse.

> It makes sense to me to have a rule, reducer, and mark UI for an ellipse, however, I'm concerned that removing the tolerance configuration that this new rule and reducer is essentially the LOFAR project ellipse rule and reducer and not reusable by others as having a tolerance _would_ be useful for projects. Couldn't you just make the tolerance 0 which would eliminate some of the customization here? Or am I missing something else here?
> 

Note that the tolerance parameter in the radial tool effectively defines a circle with a single radius around a point i.e. tolerance does not specify a range of acceptable radii. For the elliptical case you could view "a" and "b" as two separate tolerance values (can change the names to tolerance_a and tolerance_b if you feel this would make the parallels clearer). It seems more likely to me that it would be a very special case project that required the an elliptical hit tester and was able to specify a default setting for either of "a" and "b". I think it would be much more common that each ellipse to be tested agains would have different parameters,  hence the removal of the tolerance field from the Lab. 

> Also I'm not totally sure on the naming here since this is like radial in that it's testing a point within a shape, specifically an ellipse, and at some point we may want a rule that tests an actual ellipse mark against an expected ellipse shape. Maybe this should be called a `pointInEllipse` or something like that.
> 

Happy to change the name. Please confirm that you are requesting this change explicitly.

> Last, this is out of scope of this PR, but I think there is an issue with the feedback popping up immediately after the task rather than after classification submission. I don't think you can trust this feedback for use with user weighting because I'm absolutely certain a volunteer will go back and change their annotations once they know what the correct answer is. I _think_ the other projects that have use feedback have had only one task, so that's why this hasn't come up before. I'll confirm what the expected behavior is supposed to be here, but this may need to be changed before your project launches as well.

I would note that in this particular case, the feedback will only be used for training subjects for which the ground truth is known. For that reason, it doesn't matter how many times volunteers change their annotations (indeed, it might be helpful for them to be able to retry until they get it right) since their classifications on training subjects will not supersede the known ground truth. In terms of weighting, if they have learned via training what the correct response is, then the weighting derived from their final response would be the appropriate one. Note that training subjects will be a very small fraction of the total subject set, shown only to logged-in users either when they first join the project or via a dedicated training workflow. Since annotating these images has a genuine utility in terms of training and improving the classifications on real subjects, it should not be considered as wasting volunteer effort.

eatyourgreens:
From what I can see, the radial strategy is a specialised case of this strategy when both radii are equal (and theta is 0?). +1 from me to renaming to make that clear. A future development might be to have one strategy for measuring closeness to a point within an elliptical hit area, and have the circular radial strategy being a specialised case where the tolerance input sets both a and b.

eatyourgreens:
>I think it would be much more common that each ellipse to be tested agains would have different parameters, hence the removal of the tolerance field from the Lab.

This is also true for circular hit areas (eg. Space Warps subjects), but we still prefer to set a default rule in the lab for the radial strategy on a workflow. I'd be happier if this PR followed established behaviour rather than changing how the lab works for feedback.

If you forget to add radial tolerances to a subject, feedback will fallback to the default rule and I'm not sure what happens if the tolerances are completely undefined.

app/features/feedback/shared/strategies/drawing/ellipse/reducer.js
```diff
@@ -0,0 +1,34 @@
...
+  /* eslint-disable no-restricted-properties */
+  const condition = Math.pow((2.0*projectedX/feedbackA), 2) + Math.pow((2.0*projectedY/feedbackB), 2);
+  /* eslint-enable no-restricted-properties */
+
+  return condition < 1.0;
+}
+
+// Determines whether there are any annotations falling within tolerance for a
+// rule, and appends all successful annotations if so.
+function radialReducer(rule, annotations = []) {
```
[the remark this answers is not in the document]
eatyourgreens: `radialEllipseReducer` maybe? I can't figure out whether this strategy extends the radial strategy, or the radial strategy extends this strategy to add a special case where both radial tolerances are equal.

srallen:
>  I'd be happier if this PR followed established behaviour rather than changing how the lab works for feedback.

I completely agree. I think that this rule should support tolerance and it is correctly identified that it'd have to support a tolerance for both radii. This is a bit more math, but keeps the API consistent in its behavior for feedback.

I think for us to accept this PR, tolerance should be added back in and this particular project configured to have zero tolerance instead.

> Happy to change the name. Please confirm that you are requesting this change explicitly.

Yes, @hughdickinson could you use the more explicit naming? Thanks.

eatyourgreens:
>This is a bit more math, but keeps the API consistent in its behavior for feedback.

I think it's just a case of renaming both A and B radii to make it clear that they are tolerances, but I'd have to go back and double check the radial strategy. My memory is that the radial strategy checks whether point A is within a certain radial tolerance of point B. This PR changes the tolerant area to be elliptical rather than circular.

eatyourgreens:
On the other hand, if you want to check whether a drawn ellipse is within a certain tolerance of a specified ellipse, then you'll need extra feedback params to specify those tolerances.

app/features/feedback/shared/strategies/drawing/pointInEllipse/reducer.js
```diff
@@ -0,0 +1,34 @@
...
+  const feedbackA = rule.tolerance_a;
+  const feedbackB = rule.tolerance_b;
+  const feedbackTheta = rule.theta;
+
+  const feedbackThetaRad = Math.PI * (rule.theta / 180.0)
+  const projectedX = (annotationX - feedbackX) * Math.cos(feedbackThetaRad) + (annotationY - feedbackY) * Math.sin(feedbackThetaRad)
+  const projectedY = (annotationY - feedbackY) * Math.cos(feedbackThetaRad) - (annotationX - feedbackX) * Math.sin(feedbackThetaRad)
+
+  // Math.pow is a restricted property, but using the exponential operator (**)
+  // breaks the build :(
```
srallen: Not a blocker, but I'm curious about why using ** was crashing the app?

hughdickinson:
@srallen Hope that's done the trick!

app/features/feedback/shared/strategies/drawing/pointInEllipse/reducer.js
hughdickinson: Not sure. I think either Mark or Roger wrote the original code and left that comment.

app/features/feedback/shared/strategies/drawing/ellipse/reducer.js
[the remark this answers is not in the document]
srallen: I think this is slightly out of scope, but something we should consider when porting this feedback strategy over to the new classifier.

app/features/feedback/shared/strategies/drawing/pointInEllipse/create-rule.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    y: "300",
+  }
+
+  const workflowRule = {
+    defaultFailureMessage: "=( Fail",
+    defaultSuccessMessage: "Success!",
+    defaultTolerance: "20",
+    failureEnabled: true,
+    id: "1234",
+    strategy: "ellipse",
```
srallen:
```suggestion
    strategy: "pointInEllipse",
```

app/features/feedback/shared/strategies/drawing/pointInEllipse/create-rule.spec.js
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  it('should return a valid rule without custom subject-specifics', function () {
+    const rule = createRule(subjectRule, workflowRule);
+    expect(rule).to.deep.equal({
+      failureEnabled: true,
+      failureMessage: "=( Fail",
+      hideSubjectViewer: false,
+      id: "1234",
+      strategy: "ellipse",
```
srallen:
```suggestion
      strategy: "pointInEllipse",
```

app/features/feedback/shared/strategies/drawing/pointInEllipse/create-rule.spec.js
```diff
@@ -0,0 +1,62 @@
...
+    subjectRule.tolerance_b = "60";
+    subjectRule.theta = "60";
+
+    const rule = createRule(subjectRule, workflowRule);
+    expect(rule).to.deep.equal({
+      failureEnabled: true,
+      failureMessage: "Subject-specific failure message",
+      hideSubjectViewer: false,
+      id: "1234",
+      strategy: "ellipse",
```
srallen:
```suggestion
      strategy: "pointInEllipse",
```

app/features/feedback/shared/strategies/drawing/pointInEllipse/reducer.js
```diff
@@ -0,0 +1,34 @@
+function isAnnotationWithinEllipse(rule, annotation) {
+  const annotationX = annotation.x;
+  const annotationY = annotation.y;
+  const feedbackX = rule.x;
+  const feedbackY = rule.y;
+  const feedbackA = rule.tolerance_a;
+  const feedbackB = rule.tolerance_b;
+  const feedbackTheta = rule.theta;
```
srallen:
```suggestion
```
(it's unused)

eatyourgreens:
For the rules, I think you want something like `feedback_1_toleranceA` which would parse as `rule[1].toleranceA`. Other projects using feedback have used camel-cased names for their rule properties.

Otherwise, I think these rules parse as `rule[1].tolerance.a` but I'm not sure.

eatyourgreens:
Just double-checking, but the default rule set in the project builder is now a circular region ie a radial strategy. Is that right, or did you want to enter two tolerances in the workflow editor?

srallen:
> the default rule set in the project builder is now a circular region ie a radial strategy. Is that right, or did you want to enter two tolerances in the workflow editor?

I did that just now just to get it to work in the app without crashing. The UI and code doesn't currently support entering multiple default tolerances. I think that's slightly out of scope here since the tolerance will have to be defined for each subject anyway for this specific project. I think we can update how multiple defaults can be supported in a follow up PR.

hughdickinson:
> Just double-checking, but the default rule set in the project builder is now a circular region ie a radial strategy. Is that right, or did you want to enter two tolerances in the workflow editor?

This was intended. Mainly because it simplified the lab code. If the subject-specific metadata aren't set, then it's effectively the radial tool.

hughdickinson:
> I have few minor suggested changes that I used the suggest change feature for, so you just need to approve them and it'll automatically commit to your fork.
> 
> For the manual review of workflow 12686, I updated the feedback rules on the workflow to use the `'pointInEllipse'` strategy since the rename, however, the subjects linked need their metadata updated too. I see `feedback_1_a`, `feedback_1_b`, etc. I would guess these are supposed to map to the now defined tolerances, `tolerance_a`, `tolerance_b`, so I think the metadata should now be `feedback_1_tolerance_a`, `feedback_1_tolerance_b`, etc. Could you confirm this? Let me know if you'd like me to make that change.
> 
> And as a heads up, I'm pretty sure after looking over the code that the `metadata_to_rules.js` helper function will need updating to handle the new tolerance metadata. It splits the string on `_` and we haven't had any rules yet that have more than one tolerance axis, so the code isn't written to handle it. A good approach would be to write a spec using a mock similar to the new metadata then update the code to be able to pass it while making sure the other specs continue to pass.
> 
> Thanks!

@srallen Implemented the attribute name changes and tested locally and on staging server. Updated the subject metadata for workflow 12686. Seems to be working well.

hughdickinson:
@srallen Can this be merged now?

PR #5558, opened 2019-11-20
Remove grid tool from list of experimental features

rogerhutchings:
Disables the grid experimental tool by removing it from the list of available experimental features.

I haven't removed the code yet, as there are two projects still using it - the prototype OCR project for Old Weather it was originally built for, and a test project owned by @snblickhan. My plan for this is to:

- Inform the OW team
- Remove the tool from the OCR project workflow
- Delete the code from PFE

### DIFF
--- a/app/pages/admin/project-status/experimental-features.jsx
+++ b/app/pages/admin/project-status/experimental-features.jsx
@@ -9,7 +9,6 @@ const experimentalFeatures = [
   'fan',
   'mini-course',
   'worldwide telescope',
-  'grid',
   'workflow assignment',
   'Gravity Spy Gold Standard',
   'allow workflow query',


snblickhan:
Definitely all good from my end -- the test project was just to get a sense of what that tool was, etc. No longer used in any way.

wgranger:
Yes. I built this a while back and am not using it for anything. The code also relies on putting a ton of information in a user's preferences, which is probably not a good idea. Deprecating this is preferable. I have code on my local machine for a revised grid tool, which we could always take a look at should we decide to bring in the grid tool at some point in the future.

srallen:
Could you update its status in this list after merge? https://github.com/zooniverse/how-to-zooniverse/blob/master/ProjectLifecycle/experimental_tools.md

rogerhutchings:
cc'ing @philip-brohan - I think the plan to use OCR for Old Weather using this tool has been shelved, but please create an issue if a need for this tool comes up again.

PR #5576, opened 2019-12-03
Remove https and open config options

srallen:

I've been seeing https://github.com/webpack/webpack-dev-server/issues/2313, so I'm removing the https dev server config for now until it is fixed. I'm also removing the auto open config because it'll open to `localhost` which is not useful (since you should be using `local.zooniverse.org` or one of the other allowed sub-domains)

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -13,9 +13,7 @@ var config = {
       '.zooniverse.org'
     ],
     historyApiFallback: true,
-    https: true,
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Noting also that you should also be testing locally with `docker-compose up`, which won't open a browser.

PR #5605, opened 2020-01-06
Fan tool: make default spread 10º

### LINKED ISSUE #5604: Changing the default spread angle of the fan tool for Planet Four
The default spread for the fan tool on panoptes is 30 degrees, We found in the P4 beta test that this is causing problems in the sense that people are adjusting it the spread of the fan tool.  This is because 30 degrees spread is  near the peak of the fan distribution observed in the P4 catalog. The original default value was 5 degrees spread on the original Ouroboros P4.  Can the default spread be changed to 5 degrees. 

P4 catalog spread:

catalog_spread

There is a peak at 5 degrees the default value but that's very tiny features the wider distribution is the one to look at. 

The disribution of fan spreads from P4 panotpes beta test from the two interfaces:

spread

We think the spike at 30 is saying people aren't adjusting the fan tool spread because it looks close enough the average value. We think we're losing information about areal coverage and having a small value like 5 degrees would help get people to fit the fan shape more accurately. 

Would you be able to make the change or if other groups are using the fan tool can a P4 clone version be generated with a 5 degree default spread? We think once this is done, P4 will very close to launching on panoptes 

cc @eatyourgreens @lcjohnso

eatyourgreens:
Change the default fan tool spread angle to 10º.

Staging branch URL: https://pr-5605.pfe-preview.zooniverse.org

Fixes #5604.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/drawing-tools/components/Fan/fan.jsx
+++ b/app/classifier/drawing-tools/components/Fan/fan.jsx
@@ -17,7 +17,7 @@ class Fan extends React.Component {
       y,
       radius: 0,
       rotation: 0,
-      spread: 30
+      spread: 10
     };
   }
 


eatyourgreens:
>I'm guessing the configuration is being added to support backward compatibility in case it is needed? 

If we don't have an initial estimate for the spread angle, the tool simply doesn't work. You'll get `NaN` for the drag handle coordinates if `spread` is 0.

srallen:
I'd prefer to have a discussion if this should be configurable or if there is no default, if the interaction of the tool must change so the volunteer can set the angle. Since this is a new change, I'd like this to be part of the ADR of the rewritten tools rather than paint ourselves into a corner prematurely here.  

Making this default to 5 here is fine for now as it enables P4 to launch and we can query the database in case we need to communicate this change to anyone else.

eatyourgreens:
>if the interaction of the tool must change so the volunteer can set the angle.

The tool must render with some guess at the initial angle before the volunteer can set the angle. Similarly, ellipses must render with an initial guess at the minor axis before a volunteer can adjust it.

srallen:
Can we scope this PR to just make the default 5 for the current fan tool, put off the configuration addition, and move the conversation on if we should do this to https://github.com/zooniverse/front-end-monorepo/pull/1390? I'd like researcher input into this one.

eatyourgreens:
I'd like Meg to be able to adjust the default until she finds a value that she's happy with.

srallen:
It's harder to take features away than to give them. It's not clear to me if more experimentation is needed since the request is directly to make it 5, so I'd rather not add the ability to configure and then take away if we don't need to.

eatyourgreens:
Tool options aren't a new feature. Point tools are already configurable in the project builder, so I reused that idea.

eatyourgreens:
Sorry, I should have added that my concern here is that the tool is almost unusable with an angle of 5, but I'd like Meg's opinion.

eatyourgreens:
Here's Planet Four on this branch. https://pr-5605.pfe-preview.zooniverse.org/projects/mschwamb/planet-four/classify?env=production

I'm finding that when the fan angle is initially very small, the drag handles are all so close together that they merge on the screen.

srallen:
I suggested default values that make the tool unusable in my comment [here](https://github.com/zooniverse/front-end-monorepo/pull/1390#issuecomment-571187469), but I'm not sure that's a good solution.

As for configurability in general, yes we have some tools that have some configurable values, but in this case I think we should be really thoughtful in how we approach this to avoid creating new issues for ourselves.

mschwamb:
I think 10 degrees spread is working okay I think. It's larger than 5  and less than the previous default value of 30.  I was able to adjust the fan in mobile. How about the spread be change to 10, and we do the final beta test for P4 and I can look at the spread of values?

eatyourgreens:
Sounds good to me.

eatyourgreens:
I'm inclined to close this and just set the default value to 10, as per https://github.com/zooniverse/Panoptes-Front-End/issues/5604#issuecomment-573685213.

srallen:
Ok, I'm fine with delaying the updates for the lab for when the fan tool gets added to the new classifier.

eatyourgreens:
@mschwamb Sorry for the delay in getting to this. I've set the default spread angle to 10º.

PR #5606, opened 2020-01-06
Implemented false-positive mode for drawing strategy feedback

hughdickinson:
Staging branch URL: https://pr-5606.pfe-preview.zooniverse.org

Implemented functionality that allows clicks within the geometrical regions defined by the "drawing" feedback strategies to be interpreted as false positives. This is useful in difficult cases when a mistaken click is anticipated and should be highlighted as such. Specifying that a defined region should be interpreted as a false positive results in three behavioural changes relative to the default feedback mode.

1. If the false-positive region is **not** clicked on, then no failure message is displayed.
2. If the false-positive region **is** clicked on then it is rendered using the failure colour (normally red) instead of the success colour (normally green) on the modal feedback subject view. Note that for consistency with the non-false-positive mode, the **success** message is rendered, even though clicking on a false-positive region could plausibly be interpreted as a failure.
3.  If the region is **not** clicked on, then it is **not** rendered on the modal feedback subject view.

False positive mode `false` by default and this can only be modified using subject metadata on a per-region basis.

[Workflow 12686 on project 8190](https://pr-5606.pfe-preview.zooniverse.org/projects/chrismrp/radio-galaxy-zoo-lofar/classify?workflow=12686&env=production) demonstrates the new functionality for the `pointInEllipse` feedback strategy.

Unit tests have been implemented for all the changes. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


srallen:
I think I should be able to have a look at this tomorrow.

hughdickinson:
> > If the false-positive region is clicked on then it is rendered using the failure colour (normally red) instead of the success colour (normally green) on the modal feedback subject view. Note that for consistency with the non-false-positive mode, the success message is rendered, even though clicking on a false-positive region could plausibly be interpreted as a failure.
> 
> If I'm understanding this correctly, then this means that the failure color is displayed with a success message. If I was a volunteer, I would interpret this as an error, not an indication of a false positive. Is this intended to be communicated to the volunteer in some way? If so, I think this needs to be reconsidered.

First, note that by default, false-positive mode is `false` and (at the moment) it cannot be activated or controlled from the lab. This means that in order to even use the facility, project owners must be able to modify the metadata of subjects using e.g. the Python API. If this is not the case then the behaviour reverts to standard. Accordingly, project owners should be able to provide a **specific** "failure" or "success" message for any **specific** region defined for a particular subject. It is not intended that the **default** (workflow/task) success message is shown if a false positive region is clicked but rather a **specific** "success" message that pertains to the false positive nature of the region. This is how the demonstration workflow is configured. A **user** should never encounter the situation you describe, when clicking a false positive results in a message that indicates they have performed the correct action.

I implemented the code as is because the alternative (explicitly swapping the semantics of the success and failure messages in false positive mode) would have required more logic and several extra lines of code and it is not clear where the best place to make this swap would be (`create-rule.js` perhaps?). The implemented behaviour can be more explicitly documented in the READMEs for each strategy. 

> With that said, I'm not sure if the deployed branch with the test workflow is demonstrating the intended behavior. The reason is that most of the files I see as being edited in the diff are in the `column` strategy folder, not `pointInEllipse`. Did you meant to edit these files?

Yes. The edits are intentional. In fact, exactly four files have been modified for each strategy:
* `create-rule.js`
* `create-rule.spec.js`
* `feedback-mark.jsx`
* `README.md`

There are a few more lines of code that changed for the `column` strategy because its rendering code is slightly more complicated. We require false positive mode for the `radial` and `pointInEllipse` strategies for our project, and since the logic is so similar, I implemented it for the remaining drawing strategy as well. In principle this would make it easier to add a generic lab component to set a workflow/task default for the false-positive mode (modulo the earlier discussion about "success"/"failure" messages).

The example project does illustrate false positive mode for the pointInEllipse and radial strategies but may be a little complex. I can create a project with more trivial subjects a a demonstrator if required.

srallen:
> Yes. The edits are intentional. In fact, exactly four files have been modified for each strategy

Ah ok, I missed the edits for the other strategies in the diff UI... I honestly think it only displayed edits for column yesterday, so I had a bit of confusion what was happening in your branch. I'll have another look again today. Thanks.

hughdickinson:
> > First, note that by default, false-positive mode is false and (at the moment) it cannot be activated or controlled from the lab. This means that in order to even use the facility, project owners must be able to modify the metadata of subjects using e.g. the Python API. If this is not the case then the behaviour reverts to standard. Accordingly, project owners should be able to provide a specific "failure" or "success" message for any specific region defined for a particular subject. It is not intended that the default (workflow/task) success message is shown if a false positive region is clicked but rather a specific "success" message that pertains to the false positive nature of the region. This is how the demonstration workflow is configured. A user should never encounter the situation you describe, when clicking a false positive results in a message that indicates they have performed the correct action.
> 
> Correct, so with all of the required configurations, it is possible for the red failure color to be shown with a subject specific success message. As I said previously, I think there's potential for confusion. Perhaps in this scenario, the false positive success messages can be positioned under a header that is labeled "False positives"?
> 
That's a good idea and I can implement it. I was just trying to minimise the number of changes in this PR.
> Just to be clear in the demonstration project, is the message "You don't have to click on the solid ellipse" the false positive message here?
> 
> Screen Shot 2020-01-15 at 2 03 52 PM
> 
> If not, then could you provide a screenshot of specifically what I should be looking for?
No, you're right. We knew about this case whereby correctly clicking on a true positive that overlaps with a false positive will fire the false positive response and were willing to accept it because we're using the feedback in a very controlled environment of a specifically designed tutorial workflow. I'm not sure that switching the ordering of the success/failure messages in false positive mode would resolve this pathology. I think it needs extra logic that addresses this edge (?) case but opted again for the simplest approach initially. I can implement the logic (i.e. check for try pos before firing false pos) if required but I can imagine situations when this might not be the desired behaviour, so I would recommend making it configurable with a flag like `truePosOverridesFalsePos`.
> 
> I'd like to invite @beckyrother to review this because I'm having concerns with UI aspect of this feature and it'd be good to find out if others do or do not.

To be honest, I don't love the feedback UI either, but it would be inconvenient for our project if it was changed now as a tutorial video that features it has been made.

beckyrother:
Thanks for tagging me @srallen. 
Showing both the positive and negative feedback in the same box without a way to know which mark is which isn't great UX, but I understand the limitations of the tool. This is probably ok for now.
That's quite a long tutorial video, love that there is one timestamp link, perhaps consider adding more 'chapters' so it's less daunting to watch?

srallen:
@hughdickinson I understand and appreciate the motivation to keep the changes minimal. Aside from the inline recommendations I've made, could we see what it looks like to have the header in the feedback modal? I think we can hold off on trying to add the more complicated logic for overlapping shapes for now. If it is an issue that you find while beta testing with the volunteers, then we can address it then. Thanks.

hughdickinson:
All changes implemented (apologies for getting messed up and having to do a force push to my fork!). The solution now looks like this:
Screenshot 2020-01-22 at 09 55 31
@beckyrother Any tips on how to make it **not** look hideous would be much appreciated! Also, I'm aware that the the rather generic titles (particularly "Failures") could probably be improved. Again, all suggestions welcome.

eatyourgreens:
Sorry for wading into this late but is False Positive a new type of feedback, in addition to Success and Failure, rather than a mode applied on top of Success/Failure? Just trying to get my head around what this PR is trying to do. The code is a bit confusing on first read through.

hughdickinson:
> Sorry for wading into this late but is False Positive a new type of feedback, in addition to Success and Failure, rather than a mode applied on top of Success/Failure? Just trying to get my head around what this PR is trying to do. The code is a bit confusing on first read through.

Hi @eatyourgreens. It was desirable for a project I was working on to be able to define a region that we might reasonably expect a confused volunteer to click on incorrectly. The way that the drawing strategies were set up before was such that clicking on the region was interpreted as success (and importantly, rendered in green) and a success message was shown,  whereas not clicking on it was interpreted as failure (and the "missed" region would be rendered in red) and a failure message was shown.

Now, if a region is defined to be a false positive, a different logic applies. Not clicking is really a success, but it doesn't make sense to highlight it by rendering or displaying a message "Well done! You didn't click on the thing you were not supposed to click on." For this reason, no message is displayed when the false positive region is not clicked on. On the other hand, clicking on the region shouldn't really result in it being rendered in green since this would indicate that a correct action has been performed. Accordingly, if a volunteer clicks on a false positive region, it is rendered in red. 

What is perhaps confusing is that I did not choose to swap the **messages** that are displayed for a false positive region. This was just to avoid adding even more extra logic. This means that clicking will display the "success" message (whereas not clicking will not display any message as outlined above). The idea being that the false positive mode is intended to be enabled on a subject-and-region-specific (i.e. not workflow-wide) basis and an appropriate "success" message (in fact indicating a false positive "failure") could be defined by the project owner using appropriate subject metadata.

As far as I was able to discern, it was not possible to implement this false-positive flagging behaviour using any "ingenious" combination of messages and regions with the standard success and failure options.

eatyourgreens:
Ok, so it's a new class of rule that inverts the success/failure logic?

app/features/feedback/shared/strategies/drawing/column/feedback-mark.jsx
```diff
@@ -2,7 +2,9 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 function FeedbackMark({ rule }) {
-  const color = (rule.success) ? 'green' : 'red';
+  const color = (rule.success) ? (rule.falsePosMode ? 'red' : 'green') : 'red';
```
eatyourgreens: So in false positive mode, both success and failure are red? Shouldn't failure be green?

Also, the linter should warn about nesting ternary operators like this. It's confusing to read and understand.

Declaring the colours separately might be easier to understand.
```js
const [ successColour, failureColour ] = rule.falsePosMode ? [ 'red', 'green' ] : [ 'green', 'red' ];
```

hughdickinson: > So in false positive mode, both success and failure are red? Shouldn't failure be green?
> 
No, not really, but only because your earlier statement:
> Ok, so it's a new class of rule that inverts the success/failure logic?

Is not correct. In false positive mode, the region doesn't render (the opacity gets set to zero), so the colour doesn't matter.

> Also, the linter should warn about nesting ternary operators like this. It's confusing to read and understand.
> 
> Declaring the colours separately might be easier to understand.
> 
> ```js
> const [ successColour, failureColour ] = rule.falsePosMode ? [ 'red', 'green' ] : [ 'green', 'red' ];
> ```

Will fix.

hughdickinson:
> Ok, so it's a new class of rule that inverts the success/failure logic?

Not really. If that were the case then not clicking in the region would render it green and display the "success" message, while clicking would render it red and display the "failure" message. What actually happens is that **not** clicking does **not** render the region and no message is displayed, whereas clicking renders in red and the "success" message (which should be individually specified using metadata to indicate a false positive click) is displayed.

eatyourgreens:
OK. The logic in this PR has become quite involved, quite quickly, for what seems to be a fairly straightforward decision. Right now there are a bunch of `if (!something && !somethingElse)` checks, which could be a problem for the next person who has to come in and change this code.

To me, it sounds like you want to check 'is this a false positive rule` once, when you process rules. False positive rules should set the success message to the failure message, and set the success colour to red (plus any other styling). Failure style could be `opacity: 0` but `display: none` is much better. Once those are set, then further processing and rendering proceeds as normal, right?

hughdickinson:
> OK. The logic in this PR has become quite involved, quite quickly, for what seems to be a fairly straightforward decision. Right now there are a bunch of `if (!something && !somethingElse)` checks, which could be a problem for the next person who has to come in and change this code.
> 
> To me, it sounds like you want to check 'is this a false positive rule`once, when you process rules. False positive rules should set the success message to the failure message, and set the success colour to red (plus any other styling). Failure style could be`opacity: 0`but`display: none` is much better. Once those are set, then further processing and rendering proceeds as normal, right?

The main reason that the logic has become so complicated now is the there was a request to group the messages into categories so that they can be displayed with appropriate headings. 

The initial implementation was approved in a previous PR and for that situation, your suggestion would basically have involved moving some lines of code to a different file. At the time it seemed sensible to place logic that was deciding how to render objects in the rendering code itself. Nonetheless, I can move it seems cleaner.

Even then I think I would still need to employ the logic to group messages in the `getFeedbackMessages` function, and it's not obvious to me that any processing at the rule creation stage would make this simpler.

I did try using `display:none` when first developing but for some reason it didn't work like I expected (or at all in fact) so I fell back to the non-optimal opacity based solution.

app/features/feedback/classifier/helpers/open-feedback-modal.jsx
```diff
@@ -67,7 +88,9 @@ function openFeedbackModal({ feedback, subjectViewerProps, taskId }) {
     subjectViewerProps: getSubjectViewerProps(feedback, subjectViewerProps, taskId)
   };
 
-  if (props.messages.length > 0) {
+  // console.log(props.messages.length);
```
eatyourgreens: Stray console log comment?

eatyourgreens:
>Even then I think I would still need to employ the logic to group messages in the getFeedbackMessages function, and it's not obvious to me that any processing at the rule creation stage would make this simpler.

I was thinking of adding the category during rule processing, then group on `item.category` where category is 'success', 'failure' or 'falsePositive'.

>I did try using display:none when first developing but for some reason it didn't work like I expected (or at all in fact) so I fell back to the non-optimal opacity based solution.

Weird. I can't think of any reason why `style = { display: 'none' }` wouldn't work either. This isn't a blocker to merging this. I was just interested to find out why it was using transparency to hide markers.

srallen:
@eatyourgreens just an FYI, this project won't likely migrate to the new classifier and might finish before we begin migration anyway, so I'm not even sure if we'll port this functionality. Keeping that in mind while reviewing this, the code might not need to be optimal...

eatyourgreens:
That's fine, but might make feedback harder to modify for the next project that wants to modify it. The solution to that is to launch those projects in the new classifier, I guess.

srallen:
Feedback still needs a lot of work generally speaking. We need to be more mindful of enabling it on PFE projects and any new projects / migrating projects to the new classifier will use new code. Ideally this code isn't going to be modified again after this. If and when it is ported, we can address some of the implementation issues there.

eatyourgreens:
Sounds good to me.

beckyrother:
@hughdickinson visually, adding the headings helps a lot in terms of quick reads and legibility. If there is a way to indicate which markings are successes, failures, or false positives, that would help a ton; right now there's not really a way to know which markings are which (unless I'm missing something). 
Agreed with Sarah, feedback needs a lot of work and this is ok for this project but we should definitely consider thinking this through in more detail before we roll it out to more projects.

hughdickinson:
Just to semi close the loop on this, I'm happy to implement the refactor suggested by @eatyourgreens. It shouldn't take too long, but perhaps it would be better if I submitted that as a new PR and closed this one? Also @beckyrother, I can render false positives in a different colour and indicate this in the headings e.g. "False Positives (Yellow)", but that's another level of complexity that we may wish to avoid. Finally, are there any thoughts on some more encouraging or neutral category titles? "Failures" sounds pretty demoralising to me.

eatyourgreens:
Hits and Misses maybe? I'm not sure.

Whatever text you use, it should probably be localised so that it can be translated by volunteers. That depends if this text is going to be shown on projects which use more than one language.

If feedback text isn't localised at the moment, then that can be a separate PR.

app/features/feedback/classifier/components/feedback-modal.jsx
```diff
@@ -27,17 +28,24 @@ class FeedbackModal extends React.Component {
...
-          {messages.map(message =>
-            <li key={Math.random()}>
-              {message}
-            </li>
-          )}
-        </ul>
+          {_.chain(messages).map((catMessages, cat) => {
+            return (<div key={Math.random()}>
+              <h3 key={Math.random()}>
+              {cat}
```
eatyourgreens: This should probably be a `<Translate />` component, like `FeedbackModal.title` on line 34. Then store the actual category text in `app/locales/en.js`. That'll allow it to be translated.

hughdickinson: Good advice. Thanks.

srallen:
> It shouldn't take too long, but perhaps it would be better if I submitted that as a new PR and closed this one?

You don't need to do that. If the additional change are small, then we can review in this one.

beckyrother:
>  I can render false positives in a different colour and indicate this in the headings e.g. "False Positives (Yellow)", but that's another level of complexity that we may wish to avoid.

At the moment I don't see a lot of use in knowing that I made a mistake somewhere but I'm not told where. PH:TESS is able to show marks in red and green, is that possible to port over here? 


Agreed with Jim's 'hits and misses' labels – feels much friendlier than 'failure'. Good call to rethink that wording 👍

hughdickinson:
@beckyrother The red=fail, green=success rendering is already implemented (see some screenshots above). I thought you meant another colour to indicate the false positives.

eatyourgreens:
I think Becky might mean the colours of the point marks themselves in that screenshot. They're all light blue at the moment, so it's not clear which message refers to which mark.

hughdickinson:
> I think Becky might mean the colours of the point marks themselves in that screenshot. They're all light blue at the moment, so it's not clear which message refers to which mark.

@beckyrother @eatyourgreens Do you want me to implement that functionality? Doing so on PFE would probably involve altering code somewhat disconnected from the feedback "module" itself.

eatyourgreens:
I'd vote against modifying the subject viewer.

eatyourgreens:
One other thought, sorry. Has this been checked for any side effects on any other projects that use feedback?

hughdickinson:
> One other thought, sorry. Has this been checked for any side effects on any other projects that use feedback?

To unexpectedly see a behaviour that deviates in any sense from the current one, such a project would have to have inadvertently defined a subject metadatum of the the form `#feedback_<feeback_id>_falsePosMode`. That seems to me to be extremely unlikely.

eatyourgreens:
I'd expect that too, but we're changing code here that's used by all projects with feedback so it needs to be checked.

srallen:
Ok, let's rein this in a bit. I agree this should be encapsulated as much as can be.  Changing styles for the marks in feedback itself will impact other projects and we don't necessarily have a) the dev time and b) this change has already happened in the new classifier.  

The only other drawing mark related strategies that exist are radial and column. Space Warps HSC uses radial which is currently out of data and therefore can't demo feedback directly since it's now in fallback subject selection mode. The early version of TESS used column, but the project is migrated to the new classifier, so it's no longer using this code. We, however, have a test project on staging that can be leveraged for testing potential impacts.

We can consider more additional style changes but I would prefer that to be in subsequent follow up PRs and only done if absolutely necessary (i.e. this project gets beta tester feedback that indicates more work is needed). Other improvements should be scheduled for the new classifier.

@eatyourgreens Secondly as far as testing goes, please trust that the folks, me and @beckyrother,  who've been reviewing this are doing our due diligence and are handling it.

@hughdickinson I'll give this PR a final review and if there are any other additional changes requested, I'll let you know if it should happen in this PR or in a new one. Thanks.

app/features/feedback/classifier/components/feedback-modal.jsx
srallen: This change should still be incorporated.

app/features/feedback/classifier/components/feedback-modal.jsx
```diff
@@ -27,17 +28,24 @@ class FeedbackModal extends React.Component {
...
-        <ul>
-          {messages.map(message =>
-            <li key={Math.random()}>
-              {message}
-            </li>
-          )}
-        </ul>
+          {_.chain(messages).map((catMessages, cat) => {
+            return (<div key={Math.random()}>
+              <h3 key={Math.random()}>
```
srallen: The key prop only needs to be included on the parent element.

app/features/feedback/classifier/components/feedback-modal.jsx
```diff
@@ -27,17 +28,24 @@ class FeedbackModal extends React.Component {
...
+              {cat}
+              </h3>
+              <ul key={Math.random()}>
```
srallen: `key` here is not necessary

app/features/feedback/classifier/components/feedback-modal.jsx
srallen: Let's still incorporate translation into this PR.

eatyourgreens:
>Let's hold off from changing the mark styles themselves in the feedback until we get beta test volunteer feedback for this.

I'd recommend not changing the subject viewer (and drawing task mark renderer), unless the changes turn out to be trivial.

hughdickinson:
> It looks like maybe the change to use categories might have broken some logic somewhere. With the same subject as I tested 9 days ago:

I think I tracked this down to a minus sign that should have been present but wasn't in the hit testing code. My trials indicate that everything is working as expected now.

hughdickinson:
> I believe the "Unlucky" message is meant to be in the miss category?
> 
No. The project owner defines those as false positives i.e. a click that is in a region that was offered but should not have been chosen. The solid ellipse is a distinct subset (with one member) within this category. A miss would be a dashed ellipse that should have been clicked in, but wasn't.

> Question about the next task's feedback, should the radius of the missed area be just that small circle or should it be matching the ellipse?
> 
It should be the small circle. That task uses the radial strategy, since the volunteer is supposed to indicate a pseudo-point-like source in the background optical image.

beckyrother:
Imo, "unlucky" kind of delegitimizes the classification process here, indicating that luck rather than skill is needed to make contributions. If I were a volunteer on this project it would be a bit of a turnoff; I'd recommend just removing that word.

hughdickinson:
> Imo, "unlucky" kind of delegitimizes the classification process here, indicating that luck rather than skill is needed to make contributions. If I were a volunteer on this project it would be a bit of a turnoff; I'd recommend just removing that word.

That was the choice of the project owners, and is set using subject metadata i.e. not hard-coded into the PR. I will pass on the advice.

PR #5608, opened 2020-01-08
Add "Museum" role to PFE Lab Collaborators (experimental tool)

### LINKED ISSUE #5607: Lab: Add 'museum' role checkbox in user roles
Exhibits in museums are using a new user role to specify certain app behavior in the mobile app. The process to add a user to a project with this role requires using the Rails console to add a new AccessControlList with the 'museum' role. This could be an extra role checkbox in the Collaborators tab to allow it to be set manually by our team and by researchers for testing.

shaunanoordin:
## PR Overview

Fixes #5607

Staging branch URL: https://pr-5608.pfe-preview.zooniverse.org

This PR adds 'museum' role option, to the PFE Lab's Collaborators page.

- The new `museum` role is designed, presumably, for use with museum exhibits or similar events. (See #5607)
- ❗️ To see the `museum` role on the PFE Lab's Collaborators page, EITHER the `experimental_tools: ['museum-role']` value must be enabled for the project, OR the user viewing the page must be in admin mode.
- ❗️ To actually add/remove museum roles to certain users, the Panoptes API needs to have a `museum` role first.

Screen Shot 2020-01-08 at 14 16 53

### Notes

I'm actually not aware if there was ever any discussion whether there _should_ be a `museum` role or not, but AFAIC that's irrelevant to this PR and the requirements set out by 5607. Any discussion on whether there should be a new role, and what that role needs to perform, etc, should be discussed on #5607 or on the Panoptes API repo.

### Status

⚠️  **DO NOT MERGE** until...
- [ ] The requirements for the `museum` role is confirmed in #5607 
- [x] Panoptes API has the `museum` role

**Required follow up:**
- [ ] Experimental Tools documentation needs to be updated if/when this PR gets merged.

### Required Manual Testing

### Review Checklist

- [x] Does it work on mobile?

### DIFF
--- a/app/pages/lab/collaborators.cjsx
+++ b/app/pages/lab/collaborators.cjsx
@@ -38,6 +38,9 @@ ROLES_INFO =
   translator:
     label: 'Translator'
     description: 'Translators will have access to the translation site.'
+  museum:
+    label: 'Museum'
+    description: 'Museum visitors will experience a different user interface/user experience in the mobile app. This role is designed for curated events, such as a museum exhibit.'
 
 CollaboratorCreator = createReactClass
   displayName: 'CollaboratorCreator'
@@ -51,6 +54,7 @@ CollaboratorCreator = createReactClass
 
   render: ->
     @showTranslatorRole()
+    @showMuseumRole()
 
     style = if @state.creating
       opacity: 0.5
@@ -84,6 +88,10 @@ CollaboratorCreator = createReactClass
   showTranslatorRole: ->
     if (@props.project.experimental_tools?.indexOf('translator-role') > -1) or isAdmin()
       POSSIBLE_ROLES = Object.assign({}, POSSIBLE_ROLES, {translator: 'translator'});
+  
+  showMuseumRole: ->
+    if (@props.project.experimental_tools?.indexOf('museum-role') > -1) or isAdmin()
+      POSSIBLE_ROLES = Object.assign({}, POSSIBLE_ROLES, {museum: 'museum'});
 
   handleSubmit: (e) ->
     e.preventDefault()
--- a/app/pages/lab/collaborators.cjsx
+++ b/app/pages/lab/collaborators.cjsx
@@ -19,6 +19,9 @@ POSSIBLE_ROLES = {
   tester: 'team'
 }
 
+# NOTE: Panoptes API and Talk API keep track of user roles separately; Panoptes can accept arbitrary role values, but Talk API can't.
+# The key-value pairs of POSSIBLE_ROLES maps the roles on Panoptes API (key) to the 'comparable' roles on Talk API.
+
 ROLES_INFO =
   collaborator:
     label: 'Collaborator'
@@ -41,6 +44,10 @@ ROLES_INFO =
   museum:
     label: 'Museum'
     description: 'Enables a custom interface for the project on the Zooniverse iPad app, specifically designed to be used in a museum or exhibit space.'
+    
+ROLES_NOT_IN_TALK_API = [
+  'museum'
+]
 
 CollaboratorCreator = createReactClass
   displayName: 'CollaboratorCreator'
@@ -101,7 +108,7 @@ CollaboratorCreator = createReactClass
     roles = for checkbox in checkboxes when checkbox.checked
       checkbox.value
 
-    talkRoles = for role, talkRole of POSSIBLE_ROLES when role in roles and role isnt 'museum'  # 'museum' role should be added to Panoptes API's project_roles, but NOT Talk API's 
+    talkRoles = for role, talkRole of POSSIBLE_ROLES when role in roles and role not in ROLES_NOT_IN_TALK_API 
       talkRole
 
     talkRoles = talkRoles.reduce(((memo, role) ->
@@ -233,7 +240,7 @@ module.exports = createReactClass
         user_id: parseInt(projectRoleSet.links.owner.id)
         section: @talkSection()
         name: POSSIBLE_ROLES[role]
-      ).save() unless role is 'museum'  # 'museum' role should be added to Panoptes API's project_roles, but NOT Talk API's 
+      ).save() unless role in ROLES_NOT_IN_TALK_API 
     else
       projectRoleSet.roles.splice index, 1
       filteredRoles = projectRoleSet.talk_roles.filter (talkRole) ->
--- a/app/pages/lab/collaborators.cjsx
+++ b/app/pages/lab/collaborators.cjsx
@@ -101,7 +101,7 @@ CollaboratorCreator = createReactClass
     roles = for checkbox in checkboxes when checkbox.checked
       checkbox.value
 
-    talkRoles = for role, talkRole of POSSIBLE_ROLES when role in roles
+    talkRoles = for role, talkRole of POSSIBLE_ROLES when role in roles and role isnt 'museum'  # 'museum' role should be added to Panoptes API's project_roles, but NOT Talk API's 
       talkRole
 
     talkRoles = talkRoles.reduce(((memo, role) ->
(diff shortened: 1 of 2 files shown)


shaunanoordin:
## PR Update

This PR has been updated with two items:

- The text for the 'museum' role has been changed
- When a user has the 'museum' role  added or removed, these changes are NOT propagated to the Talk API. (i.e. the 'museum' role is an exception that can only exist on the Panoptes API, but not the Talk API.)

Tech details:

- there are TWO APIs which care about user roles, and the Lab page actually makes POST/PUT updates to these two APIs: the Panoptes API and the Talk API. (Surprise!)
- Unfortunately, these APIs aren't marching in tandem. The Panoptes API can accept any value as a user role, but the Talk API has as specific whitelist of allowed user roles.

Rationale for this change:

- After a discussion with @camallen it seems prudent not to change the Talk API any further, i.e. there's no point adding 'museum' to the Talk API's list of acceptable roles. Instead, the front end's lab page will cater for the 'museum' exception.

### Status

Ready for review - and I'd appreciate it if anyone can take a good look at the code functionality to catch any gotchas I might have missed with this new 'museum' exception.

FYI these are the scenarios I've considered and tested for:

- Adding a new user with the 'museum' role.
- Adding a new user without the 'museum' role. (e.g. 'tester' role)
- Adding a new user with the 'museum' role and another role.
- Toggling an existing user's 'museum' role.
- Toggling an exiting user's non-'museum' role. (e.g. 'tester role)

Condition of success is that any changes to the roles must be reflected correctly by both Panoptes API and Talk API.

shaunanoordin:
UPDATE: after further thought, I realise that another, possibly simpler solution would be to use how `POSSIBLE_ROLES` maps Panoptes API role-values to Talk API role-values. For example, changing...

```
showMuseumRole: ->
    if (@props.project.experimental_tools?.indexOf('museum-role') > -1) or isAdmin()
      POSSIBLE_ROLES = Object.assign({}, POSSIBLE_ROLES, {museum: 'team'});  # <-- this bit here
```

While this is a much _cleaner_ implementation for the front end, (i.e. no `unless role is 'museum'` conditions here and there,) it might also lead to its own weirdness on Talk - i.e. the Talk API will think anyone with the 'museum' role is basically a member of the Team. I won't go down this alternate path unless someone else can convince me it's a good idea.

shaunanoordin:
## PR Update

The code has been updated with some documentation, and the more flexible+self-explanatory `ROLES_NOT_IN_TALK_API` array.

app/pages/lab/collaborators.cjsx
```diff
@@ -225,7 +240,7 @@ module.exports = createReactClass
         user_id: parseInt(projectRoleSet.links.owner.id)
         section: @talkSection()
         name: POSSIBLE_ROLES[role]
-      ).save()
+      ).save() unless role in ROLES_NOT_IN_TALK_API
```
camallen: does this have to return a role for use in `talkRoleAction` later on?

app/pages/lab/collaborators.cjsx
```diff
@@ -84,6 +95,10 @@ CollaboratorCreator = createReactClass
   showTranslatorRole: ->
     if (@props.project.experimental_tools?.indexOf('translator-role') > -1) or isAdmin()
       POSSIBLE_ROLES = Object.assign({}, POSSIBLE_ROLES, {translator: 'translator'});
+  
+  showMuseumRole: ->
+    if (@props.project.experimental_tools?.indexOf('museum-role') > -1) or isAdmin()
+      POSSIBLE_ROLES = Object.assign({}, POSSIBLE_ROLES, {museum: 'museum'});
```
camallen: Should the `POSSIBLE_ROLES` have the museum mode setup always in code? 

Instead we could rely on presentation in the UI to only show this `museum` role checkbox if the experimental tool is enabled?

app/pages/lab/collaborators.cjsx
shaunanoordin: In this bit, if the toggled `role` is `'museum'`, then `talkRoleAction` just becomes `undefined`, so no interaction with Talk occurs whatsoever.

Sorry, CoffeeScript is a wonk to read... this basically translates to...

```
if (index === -1) {
  projectRoleSet.roles.push(role)
  if (!ROLES_NO_IN_TALK_API.contains(role)) {
    talkRoleAction = talkClient.type('roles').create(...).save()
  } else {
    talkRoleAction = undefined
  }
}  ...
```

app/pages/lab/collaborators.cjsx
shaunanoordin: > Should the POSSIBLE_ROLES have the museum mode setup always in code?

Sorry Cam, I need a bit more clarification on what you mean?

Currently, the `'museum'` role is only added to `POSSIBLE_ROLES` - and therefore only visible on the UI - when the `museum-role` experimental tool is added to the Project.

Viewing the relevant code...

```
POSSIBLE_ROLES = {
  collaborator: 'admin',
  expert: 'team',
  scientist: 'scientist',
  moderator: 'moderator',
  tester: 'team'
}
```

...the Collaborators page starts off knowing 5 possible roles, then adds `translator` and `museum` only depending on the Project's experimental-tools config.

camallen: I was thinking it might be easier to have the full set of possible roles already setup instead of modifying that object conditionally later in the code. If that's the way the translator role is setup then this is fine.

app/pages/lab/collaborators.cjsx
camallen: it's the `undefined` that i was wondering about where it'll be used later via `talkRoleAction` https://github.com/zooniverse/Panoptes-Front-End/blob/3d1a78f8012c5401b92d6b5dc357b096be6ea8bb/app/pages/lab/collaborators.cjsx#L250

I'm not really sure how that later code handles undefined...

eatyourgreens: The coffeescript is hard to read but I don't see the later code doing anything with the return values from those promises, so I think it ignores `talkRoleAction` (unless the promise rejects.)

wgranger:
I tested this PR and all looks well aside from the point @camallen made above. I'd feel better approving and merging if we add "museum" to the experimental tools if the conditional is looking for it. Otherwise, I feel like it may fall off the radar.

app/pages/lab/collaborators.cjsx
shaunanoordin: AH! Okay, big thanks to @wgranger for clarifying this. I understand now that the comment was to _"add the 'museum role' checkbox to the Experimental Features list, on the **Project Admin** page"_

Gotcha! That's a legit change to add to this PR, as I _completely blanked out_ on the fact that the Project Admin page is how normal people (who don't dick around with the zooAPI as their default method of doing things) usually add experimental features.

👌 Give me a sec!

shaunanoordin:
Many thanks to Cam, Will, and Jim for recent comments. Here's what's happening next:

- I've added a [documentation PR on How To Zooniverse](https://github.com/zooniverse/how-to-zooniverse/pull/50)
- I'm currently adding the 'museum-role' checkbox to the Project Admin page.

Back in a few minutes.

shaunanoordin:
## PR Update

The `museum-role` has now been added to the list of Experimental Tools on the Project Admin page. 👌 

Screen Shot 2020-01-20 at 17 55 39

wgranger:
I think you forgot to push that most recent commit?

shaunanoordin:
:picard_face_palm: Goddammit, Shaun.

Thanks @wgranger, you should now see commit dc6010a with the previously mentioned changes. 👌

wgranger:
Travis is pending? Seems like it has been stuck for a couple days.

PR #5627, opened 2020-01-20
Show Background Color Over External Links

wgranger:
Staging branch URL: https://pr-5627.pfe-preview.zooniverse.org

Fixes an issue appearing [on Talk](https://www.zooniverse.org/talk/17/1232948) 

Describe your changes.
Gives a background to an otherwise transparent div that was not receiving styling from the theme. I think this issue was only appearing on Windows in particular browsers? It appears to only appear when images are quite tall. However, I went ahead and made the change as I can see this occurring for any project with tall subjects.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/components/ExternalLinksBlock/ExternalLinksBlock.jsx
+++ b/app/components/ExternalLinksBlock/ExternalLinksBlock.jsx
@@ -7,7 +7,7 @@ import { zooTheme } from '../../theme';
 
 export const StyledExternalLinksBlock = styled.div`
   background-color: ${({ isItAProject }) => {
-    return (isItAProject) ? zooTheme.colors.background : 'white';
+    return (isItAProject) ? '#eff2f5' : 'white';
   }};
   box-sizing: border-box;
   color: ${theme('mode', {
--- a/app/components/ExternalLinksBlock/ExternalLinksBlock.jsx
+++ b/app/components/ExternalLinksBlock/ExternalLinksBlock.jsx
@@ -7,7 +7,7 @@ import { zooTheme } from '../../theme';
 
 export const StyledExternalLinksBlock = styled.div`
   background-color: ${({ isItAProject }) => {
-    return (isItAProject) ? '#eff2f5' : 'white';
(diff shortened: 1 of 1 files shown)


srallen:
Sort of here: https://github.com/zooniverse/Panoptes-Front-End/blob/master/css/common.styl

How styles are defined have never been consistent on PFE.

eatyourgreens:
I realised after I'd posted that the only change here is to change `zooTheme.colors.background` to a hex code, which answers my own question. Not a blocking change at all, but using the theme would be cleaner.

srallen:
I forgot I had done this: https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/theme/colors.js 

It's an early version of the theme concept being used by the styled components we have in this code base. From that file using `lightTheme.background.default` will work, since that's where that color is stored. I'd refrain adding the new zoo theme to the codebase because that'll just add another way of doing things here and it's hard to keep track of it all.

app/components/ExternalLinksBlock/ExternalLinksBlock.jsx
```diff
@@ -7,7 +7,7 @@ import { zooTheme } from '../../theme';
 
 export const StyledExternalLinksBlock = styled.div`
   background-color: ${({ isItAProject }) => {
-    return (isItAProject) ? zooTheme.colors.background : 'white';
+    return (isItAProject) ? '#eff2f5' : 'white';
```
eatyourgreens: Should this PR change `zooTheme.colors.background` to `zooTheme.colors.background.default`?

wgranger: `zooTheme.colors.background.default` doesn't exist, but I can do `zooTheme.colors.background.default`

srallen: Yes, please do

wgranger:
Updated using the light theme background default instead of a hex.

PR #5630, opened 2020-01-30
Remove Mobile Check for One Image on Workflow Tasks

wgranger:
Staging branch URL: https://pr-5630.pfe-preview.zooniverse.org

Describe your changes.
A volunteer recently brought up [on Talk](https://www.zooniverse.org/talk/18/1238480) that we still have validations in the lab for a mobile task to only have one image tasks (no multi-image workflow). This was originally a validation in early iterations of the app; however, the drawing tool allowed multi-image workflows. 

The [mobile validations](https://github.com/zooniverse/mobile/blob/a8ca98fa3e567b0ab2e733352d5afe66f9272d07/src/utils/workflow-utils.js#L27) also make no reference to a workflow having one image tasks. The `SwipeClassifier` is also capable of showing the mobile app's [version of the filmstrip viewer](https://github.com/zooniverse/mobile/blob/282603301f9260d956e6771fcb02bf3c7c46d02e/src/components/classifier/SwipeCardSubjectsView.js#L56) when multiple subjects are present.

You can test this out on this PR by navigating to a question workflow with multiple subjects and marking that subject mobile friendly. You should then be able to open the staging version of the app and enter that workflow.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/mobile/mobile-section-container.jsx
+++ b/app/pages/lab/mobile/mobile-section-container.jsx
@@ -39,16 +39,6 @@ function workflowHasNoMoreThanXShortcuts(shortcutsLimit) {
   };
 }
 
-function workflowQuestionHasOneOrLessImages({ task }) {
-  let validation = convertBooleanToValidation(false);
-  if (task.question) {
-    const matchArray = task.question.match(MARKDOWN_IMAGE);
-    validation = convertBooleanToValidation(matchArray ? matchArray.length < 2 : true, true);
-  }
-
-  return validation;
-}
-
 function drawingToolTypeIsValid({ task }) {
   let validationBool = false;
   if (task.tools[0]) {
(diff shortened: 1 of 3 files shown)


eatyourgreens:
If I'm reading the code correctly, this removes the validation that checks if the task question has more than one image in the markdown. Is that right, because that's a different check from validating whether the subject has more than one image in the locations array?

shaunanoordin:
I'm going to echo what Jim said - this PR removes the check that warns if multiple images are being embedded in the _task question._

If I set this as my single-answer question...

Screen Shot 2020-03-03 at 13 49 04

Screen Shot 2020-03-03 at 13 49 53

...then I'll see these warnings in `master`...

Screen Shot 2020-03-03 at 13 49 16

...and this PR removes the warnings above.

Judging from the how you phrased _"mobile apps now support multi-image workflows",_ I assume you're talking about Subjects with multiple images? Is there actually a check for that, because logically you'd need to go through every Subject in every Subject Set in the workflow to see if any Subject has >= 1 image? 🤷‍♂ 

More to the point: when the user said _"I notice that in the project builder it says the criteria for being an app project is to only have one image..."_ where is this criteria stated? (Is it this `workflowQuestionHasOneOrLessImages()` warning message, which was misinterpreted? Or is there some other line of text that need to be removed instead?)

wgranger:
My mistake on this one. I was a bit quick to adjust the code without thinking about the talk question first.

I think the user was confused by the phrasing on the project builder, which states "Task question has no more than one image," thinking it was referring to multi-frame subjects when if does refer to the markdown question. 

You are correct, we do need to keep this check in place. I'm closing this PR for now and will respond to the Talk post accordingly, although my response will be a bit belated.

Issue #3611, opened 2017-03-16
Add ability to configure frequency of mini-course prompt


mcbouslog:
## Expected behavior
Add ability to configure frequency of mini-course prompt in mini-course editor.

## Current behavior
Mini-course is prompted every 5 classifications, as it has been hard coded to.

lcjohnso:
Reopening this issue as there's been some renewed interest in minicourse use, and would benefit from this added config option.

Suggestions:
- Keep default as 5 unless config option says otherwise.  Currently configured here (see #2651): https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/classifier/index.cjsx#L26
- Need to add configuration on Project Builder workflow config page
- Perhaps store `minicourse_prompt_frequency` (or similar) config param in workflow configuration?

cc @srallen

camallen:
We are looking at combining another use case where we want a message shown at set classification count intervals, e.g 
> As a project owner I want to present my volunteers with a congratulations message when they hit the target of 10, 25 and 50 classifications on my project.

eatyourgreens:
One option would be a comma-delimited list of values in the project builder, saved as an array. Then the logic for showing a message is either every N classifications for a single integer, or when the current count is in the array for an array.

srallen:
We could certainly use the workflow configuration object, but I'm thinking a config object would be a better fit on the tutorial resource. It'd be a small migration for the database on Panoptes.

wgranger:
Configuration blobs should now be available on tutorial resources, so we can start moving on this in the frontend.

camallen:
ON staging only until we deploy to production next week (17th August 2020).

wgranger:
Another question before building this out. I'm assuming a project owner will set the `minicourse_prompt_frequency` tag on the project builder, so there will need to be some sort of editing option there.

PR #5672, opened 2020-04-03
Toggle feedback group titles

hughdickinson:
Staging branch URL: https://pr-5672.pfe-preview.zooniverse.org

Describe your changes.

Added functionality that allows the feedback category titles i.e. "Hits", "Misses", "False Positives" to be suppressed on a per-subject basis using subject metadata. If the corresponding metadatum is not present, or set equal to `false` then the behaviour falls back to default i.e. feedback category titles are displayed.

The ability to suppress these titles is highly desirable for project [10969](https://zooniverse.org/projects/sandorkruk/istrox), which is using the feedback system in conjunction with the "dud" strategy and a fixed subject order to provide periodic messages to users that cannot be classified as "Hits", "Misses" or "False Positives".

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
I'm not sure why it isn't being used here, but we do have the mini-course feature to pop up a message to volunteers every N subjects (where N is currently 5.)

srallen:
Agreed. I think the mini-course is a better fit with some modification. @lcjohnso could you update us on what the team needs or wants for this feature?

lcjohnso:
@srallen See here for conversation via Sandor's recent ticket: https://zooniverse.freshdesk.com/a/tickets/2510

Briefly: feedback option seemed best from an implementation perspective: no new coding required (at least originally; use of mini-course would require addition of new frequency param) and suited to use of sequential subjects (can dictate exactly when the messages are sent, not just at regular interval).

Relevant issue for mini-course frequency param: https://github.com/zooniverse/Panoptes-Front-End/issues/3611

lcjohnso:
Regarding this PR specifically: I'm OK with the principle of this enhancement.  However, I'm unconvinced the setting should be passed on a per-subject basis.

@hughdickinson Why not add this as a feedback strategy param, similar to "Hide Subject Viewer" option?  It is very messy adding this as yet another metadata field when it seems to pertain to a strategy-wide decision.

hughdickinson:
@lcjohnso @srallen I can certainly implement a workflow level option for this feature if there is general consensus that this is the right way to proceed. The reason for implementing it as a subject-metadata-based feature was to try to avoid making this an expected capability for Zooniverse feedback that might need to be transferred to the new classifier. 

This is an orthogonal discussion but I note here that @sandorkruk (project owner) has also asked whether it would also be possible to implement optional suppression of the reduction count i.e. remove the strings like "(1 match)" and enable markdown for the feedback message. The latter is trivial to implement, but there may be reasons to disfavour it that I hadn't considered. The former would involve similar changes to this PR. Could I check whether there would be reticence to implement these features?

sandorkruk:
Thanks a lot @hughdickinson @lcjohnso @srallen for looking into this! Indeed the feedback method seems the best and only viable option in this case, since the subjects are ordered (it is a audio-translation project, in which volunteers have to translate short clips in a given order), and the feedback should appear at a non-regular interval, so the mini-course does not work. The reason for adding markdown to the feedback message, is to be able to embed a link to an audio file in it. If this is possible, it would be great!

eatyourgreens:
@sandorkruk I think we support embedding the browser's audio player directly in markdown, using the same markdown syntax as images. If it works, that might be better than linking away from the classifier to an audio file.

srallen:
Adding a frequency configuration parameter to the mini-course I'm pretty sure would involve less code changes than this PR. I still think the mini-course is the best fit for the use case.

sandorkruk:
@eatyourgreens - the feedback message does not seem to support embedding.
Screen Shot 2020-04-07 at 17 37 03 @srallen - does the mini-course support embedding or markdown?

sandorkruk:
@srallen - the issue is that the frequency with which the message should appear is irregular - sometimes after 10, sometimes after 20 subjects etc. Would that be an issue for the mini-course? does the mini-course support markdown?

srallen:
> Would that be an issue for the mini-course? does the mini-course support markdown?

We would need to make a modification to allow customization of the frequency. We're currently discussing this in the associated issue where the configuration of the frequency could be a single value where the slide would pop up every N classifications or when the classification count equals one of the values in an array list, supporting your irregular frequency request. See https://github.com/zooniverse/Panoptes-Front-End/issues/3611

Another project is now also asking for this, so we'll likely do this anyway. Also, we support markdown for the mini-course content.

srallen:
@sandorkruk One limitation to note about the mini-course is that it is only displayed to logged in volunteers right now because that is how we track the last slide they've seen. Would that be an issue for your use case?

sandorkruk:
@srallen that is not an issue for this project, users will be logged in. "When the classification count equals one of the values in an array list" means that the feedback message can appear after subjects e.g. 15, 27, 31 etc. (the subjects are already ordered)? Can there be multiple feedback messages in a project? When do you expect this feature to be implemented? The researchers would like to roll out the project soon after Easter.

srallen:
@sandorkruk 
> "When the classification count equals one of the values in an array list" means that the feedback message can appear after subjects e.g. 15, 27, 31 etc. (in this case the subjects are already ordered)?

Yes, this is what this means.

> Can there be multiple feedback messages in a project?

Could you clarify what you mean here? I'm happy to enable the mini-course feature now on your project so you can see how it functions. The mini-course is similar to a workflow tutorial. It's a modal pop-up that displays media (usually an image) and then some markdown content. After N classifications, currently 5, the first slide displays, then after the next 5, the next slide displays. We don't have a limit on number of slides. You can also visit the project Gravity Spy to see it in action.

So just to be additionally clear, we have a few different features that provide different kinds of messaging to volunteers. A project can have both subject metadata based feedback, used for messaging to the volunteer training information about that specific subject. A project could also have the mini-course setup, which is intended to message non-tutorial additional information about the project, educational information, or other meta information, etc.e Both can be enabled and configured. 

I've been starting the conversation of fitting this work in. Could you be more specific about  "soon after Easter" so we can best plan? Thanks.

sandorkruk:
@srallen what I meant is that there would be two different feedback messages to be shown, unrelated to the task or subject metadata. Something like "Thanks for your effort, please continue!" and second message like "Thanks for completing this batch. You can listen to a song by ... here". I checked Gravity Spy and it looks like what we want. Is the mini-course formatable from the lab or from the python client?

When will the frequency of the mini-course be adjustable? They were initially planning to release it this week, but it slipped...The messages and everything else is ready though for launch.

srallen:
@sandorkruk I've enabled the mini-course for project 10969. The lab now has an editor for it available in the left sidebar. It's very similar to the tutorial editor, so setting it up should have familiarity. I'll get back to you on ETA as soon as I can.

sandorkruk:
Thank you @srallen!

trouille:
@sandorkruk I'm sorry for the delay in responding here. It's mostly been held up on my end as we plan the next month of effort, juggling existing commitments and deadlines, grant deadlines, new needs with the increased traffic, etc. All good things, but means that it's tricky to fit in even relatively 'quick' efforts into this load. Our team will be able to tackle what's needed here in this upcoming month of May effort, but I'd appreciate being able to prioritize a few existing efforts in early May and do this in the second half. Sarah mentioned that you had had a hoped-for Easter timeline for this. Could you share a bit more about your time constraints (if there are serious pressing time constraints), so I can take that into account? Thanks.

sandorkruk:
Hi @trouille, thanks for the reply! of course I understand that there are more pressing issues than this one, being a relatively minor and specific functionality. The original aim to release the project 20 April, so Hugh implemented the feedback functionality, which worked, but needed some changes for which the PR was not accepted. The mini-course is a good option, if the custom frequency can be added. The urgency is more that they have a group of volunteers who they've already engaged through private conversation to join the project during this time (this audio translation project is aimed for a small community who speak a specific language). If it can't be implemented until the end of the month, I will talk to the team to see whether they want to release it as it is, and add the mini-course later (if it's possible).

trouille:
Helpful to have this context. Yes, we'll need to stick with the end of the month timeline. Post here how the conversation goes w/ the group and what you decide on launching now vs waiting for when the mini-course is integrated. That's more so a question for the researchers about how you'll analyze the results, if something changes partway through the project.

sandorkruk:
Thanks @trouille! The project has not been sent out to the volunteers yet, but their planning to launch at the end of the week. They've been waiting for the integration of the Croatian version of Zooniverse platform content (sent to @lcjohnso). Any updates on whether the custom frequency mini-course would be done this week?

trouille:
Hi Sandor - we had planned to work on this this week/next. However, we have also been working over the past several months on our migration to a new host (from AWS to Azure). The timing has worked out where this week we will be migrating our staging database. We'll then test and set things up for the week after to migrate our production database. This migration effort takes priority over other efforts. Because the solution in this thread is to make the mini-course configurable and includes a change to a database resource, we must wait until after the migration is complete. This means an implementation timeline of 3rd week in June. I'm really sorry for the additional delay.

sandorkruk:
Thanks for updating me on this, I will let the research team know and they'll decide whether to go forward without the mini-course. If we switch it on later, will it be enabled for the volunteers *already* registered? For some reason, since I've been using the website a lot, the mini-course feedback no longer appears for me. 

Hugh's implementation of the feedback message (with some minor changes, removing some text and allowing for urls, this PR request) would have been a faster way forward for this project.

trouille:
https://github.com/zooniverse/operations/projects/1 provides a helpful overview of where we are in the migration. https://github.com/zooniverse/panoptes/projects/9 is a detailed view of one element. Final push over the next two weeks. 

Yes, volunteers already registered will see the mini-course when it's added to the project.

srallen:
Closing in favor of #5786 and #5789

PR #5784, opened 2020-08-19
Add overflow-anchor style to task nav container

### LINKED ISSUE #5781: Classify page scrolls to top after user clicks 'Done' in Chrome
This [issue](https://www.zooniverse.org/talk/17/1533848) has been raised on the Zooniverse Troubleshooting forum.

I think that this issue is related to element focus as suggested by @srallen, but not due to the browser scrolling to a newly focused element. Instead, it seems that Chrome, or at least the latest version, scrolls to keep the *currently* focused element in view if the position of that element changes. Firefox (for example) doesn't seem to do this.

When the user clicks 'Done', the app updates the page for the next classification, resulting in a number of separate modifications to the DOM. One modification in particular removes all of the questions (the div with class `coverable-task-container` becomes empty). This results in the 'Done' button, which still has focus, jumping up to the top of the task container. If necessary, Chrome scrolls up to keep the button in view.

I've tested this in Rainfall Rescue and American WWI Burial Cards, and both show this behaviour.

## Steps to reproduce

- Navigate to classify page.
- In the developer tools set a DOM change breakpoint (subtree modifications) on the body element.
- Click 'Done', and step through the DOM modifications in the debugger.

srallen:
Staging branch URL: https://pr-5784.pfe-preview.zooniverse.org

Fixes #5781 

The underlying scroll anchor behavior has changed with Chrome since v84. We're disabling scroll anchoring on the task nav container by adding `overflow-anchor: none` style to it. This seems to fix the unusual scroll behavior originally reported on Talk: https://www.zooniverse.org/talk/17/1533848

No changelog report exist for this change, so this is still a guess, but #5781 got us on the right track.

This PR also adds back in `https: true` to the webpack dev server configuration. It was removed a few months ago because of a bug in webpack, but webpack has been updated since then. We need this to be able to run the site locally.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/css/classify.styl
+++ b/css/classify.styl
@@ -201,6 +201,7 @@
 
   .task-nav
     display: flex
+    overflow-anchor: none
     padding: 0 2em 0 2em
 
     > :not(:first-child)
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -14,6 +14,7 @@ var config = {
     ],
     historyApiFallback: true,
     host: process.env.HOST || "localhost",
+    https: true,
     overlay: true,
     port: 3735
   },


eatyourgreens (approved):

eatyourgreens:
Looks good to me. Tested on https://pr-5784.pfe-preview.zooniverse.org/projects/welldone/american-wwi-burial-cards/classify? env=production by scrolling down to put the subject at the top of the window, then classifying in demo mode.

PR #5789, opened 2020-08-27
Irregular frequency

srallen:
Staging branch URL: https://pr-5789.pfe-preview.zooniverse.org

I made a mistake of not updating the related issue to include the requirement that the frequency can be irregular. This implements that requirement. I've branched off of the original `prompt-frequency` branch and have requested to merge into it. Since we're missing specs for this, I've left in a couple of console logs to help with manual testing. Those can be removed once you've reviewed and approved its working.

The input now allows single digits, a comma separated list, or no value.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/partials/project-modal-editor.cjsx
```diff
@@ -78,15 +78,33 @@ ProjectModalEditor = createReactClass
...
+    allowedString = /^(\d(,\d)*)$/
+    if allowedString.test(@state.frequency)
+      splitString = @state.frequency.split(',')
+      frequency = splitString.map((digit) => parseInt(digit))
+      @props.projectModal.update
+        'configuration.minicourse_frequency': frequency
+      @props.projectModal.save()
+    else if !@state.frequency
+      @props.projectModal.update
+        'configuration.minicourse_frequency': undefined
```
wgranger, line 104: Any reason for `undefined` rather than `null` here? To me, 'null' seems more explicit when setting a value and it's a bit odd to set something to `undefined`.

srallen: The underlying json-api-client deletes the object property when it's set to undefined:

https://github.com/zooniverse/json-api-client/blob/master/src/model.coffee#L49

which is what I want to do. We do it a few other places:

https://github.com/zooniverse/Panoptes-Front-End/blob/04f4fda0efbc05d0177f6e2c3d38b2b6da22ca09/app/classifier/classifier.jsx#L76

In addition, depending on what you're trying to represent in the state of something, I think `undefined` can be safer than `null` since `null` is still an object.

srallen:
Good point about the wording. I'll update to use `whole number` rather than `digit` which is what I meant.

PR #5815, opened 2020-09-17
Expand user moderation features for admins.

eatyourgreens:
Staging branch URL: https://pr-5815.pfe-preview.zooniverse.org

Add global, beta and NASA email preference checkboxes to the user admin page.
Add `valid_email` and `created_at` to the listed details.
Allow `valid_email` to be reset by admins.
Convert `handleInputChange` to JavaScript.
Require `handleInputChange.default` in old coffeescript files.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/lib/handle-input-change.js
```diff
@@ -0,0 +1,35 @@
...
+  /*
+  Using this module is a little odd.
+  Ensure that it's called in the context of a JSON-API Model instance.
+  */
+  if (typeof this.update !== 'function' || typeof this.emit !== 'function') {
+    throw new Error('Bind the handleInputChange function to a json-api-client Model instance')
+  }
+
+  let valueProperty = 'value';
+  switch (e.target.type) {
```
mcbouslog, line 13: just confirming that I think with `let valueProperty = 'value'`, then setting a `default` within the `switch` isn't necessary?

app/lib/handle-input-change.js
```diff
@@ -0,0 +1,35 @@
...
+      valueProperty = 'checked';
+      break;
+    }
+    case 'file': {
+      valueProperty = 'files';
+      break;
+    }
+  }
+
+  const value = e.target[valueProperty];
```
mcbouslog, line 24: if possibly redefined on line 27 should this be `let`?

app/pages/admin/user-settings/details.jsx
```diff
@@ -14,8 +20,50 @@ const UserDetails = (props) => {
...
+              <label>
+                <input
+                  type="checkbox"
+                  name="beta_email_communication"
+                  checked={props.user.beta_email_communication}
+                  onChange={onChange}
+                />{' '}
+                Beta test emails
+              </label>
+            <li>
```
mcbouslog, line 61: should go on line 52 to wrap the `<label>`?

also, I'm not sure if we should add `<li>` and label/input for NASA email communication?

mcbouslog (changes requested):
Very sorry it took me so long to get back to reviewing this. Nothing major, but a few minor comments. Only one of note - I'm not sure if we should add input for NASA email communications?

app/pages/admin/user-settings/details.jsx
eatyourgreens: Sorry, I'm not sure what Nasa email communication is. Is it something that admins need to set/reset on user accounts?

app/lib/handle-input-change.js
eatyourgreens: A default might be a good idea. The `let` is here so that `valueProperty` is scoped and available outside the switch.

app/lib/handle-input-change.js
eatyourgreens: Good catch. I'm not sure what `dataset.jsonValue` is doing (other than marking the input value as JSON.) I couldn't find a reference to it anywhere else in the code.

app/pages/admin/user-settings/details.jsx
eatyourgreens: I've added it to the form as 'NASA emails.'

app/lib/handle-input-change.js
eatyourgreens: Looks like `target.dataset.jsonValue` is set by adding `data-json-value` to an input element, and it tells the change handler to parse the value as JSON.
https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/dataset

eatyourgreens: I think it's used to parse strings to booleans and null values.
https://github.com/zooniverse/Panoptes-Front-End/search?q=data-json-value

eatyourgreens: I've left it in the code, since `data-json-value` is still in use.

mcbouslog (approved):
Looks great, thank you!

PR #5830, opened 2020-10-23
Admin interface for subject viewer settings

srallen:
Staging branch URL: https://pr-5830.pfe-preview.zooniverse.org

I added a select dropdown to the admin interface for projects to set the subject viewer to use per workflow. I also reorganized the layout because it was really inconsistent and didn't clearly label what different settings were for.

Here we see the staging Anti-Slavery Test project with its default workflow setup for the new classifier using the `multiFrame` viewer. Most workflows won't have this set yet because they haven't migrated over to the new classifier.
Screen Shot 2020-10-23 at 11 40 53 AM

Specs are updated.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -291,7 +291,7 @@ class ProjectStatus extends Component {
                   >
                     <option value="none">None Selected</option>
                     <option value="dataImage">Data Image</option>
-                    <option value="lightCurve">(D3/TESS) Light Curve</option>
+                    <option value="lightcurve">(D3/TESS) Light Curve</option>
                     <option value="multiFrame">Multi-Frame</option>
                     <option value="singleImage">Single Image</option>
                     <option value="subjectGroup">Subject Group</option>
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -62,6 +62,18 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
+  onChangeSubjectViewer(workflow, event) {
+    this.setState({ error: null })
+    let selected = event.target.value
+    selected = selected === 'none' ? undefined : selected;
+    if (!workflow.configuration.classifier_version) {
+      workflow.update({ 'configuration.classifier_version': '2.0' });
+    }
+    return workflow.update({ 'configuration.subject_viewer': selected }).save()
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
   getProject() {
     const { owner, name } = this.props.params;
     const slug = `${owner}/${name}`;
@@ -175,7 +187,7 @@ class ProjectStatus extends Component {
     }
 
     return (
-      <ul className="project-status__section-list">
+      <div>
         {this.state.dialogIsOpen &&
           <WorkflowDefaultDialog
             onCancel={this.handleDialogCancel}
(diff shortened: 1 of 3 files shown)


srallen:
ASM test can be found here: https://local.zooniverse.org:3735/admin/project_status/wgranger-test/anti-slavery-testing

app/pages/admin/project-status.jsx
```diff
@@ -241,20 +265,44 @@ class ProjectStatus extends Component {
...
+                <label>
+                  Subject Viewer:{' '}
+                  <select
+                    id="subject-viewers"
+                    onChange={(event) => this.onChangeSubjectViewer(workflow, event)}
+                    value={workflow.configuration.subject_viewer}
+                  >
+                    <option value="none">None Selected</option>
+                    <option value="dataImage">Data Image</option>
+                    <option value="lightCurve">(D3/TESS) Light Curve</option>
```
eatyourgreens, line 294: This is `lightcurve` in the monorepo code but maybe we should change it there?
https://github.com/zooniverse/front-end-monorepo/blob/80ce4f31ce3ab3c1f5efd24c3f71e0d70f755198/packages/lib-classifier/src/store/Workflow/WorkflowConfiguration/WorkflowConfiguration.js#L8-L15

eatyourgreens (approved):
Looks good to me. We should decide whether the TESS light curve viewer should be `lightCurve` or `lightcurve`.

eatyourgreens:
One accessibility enhancement would be to link up the headings to the controls, so that they're read by NVDA or JAWS. I can't remember if VoiceOver ignores non-form elements inside forms.
https://www.w3.org/WAI/tutorials/forms/instructions/

app/pages/admin/project-status.jsx
srallen: It's also `lightCurve`: https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/helpers/getViewer/getViewer.js

This is the file I used as reference, so presumably it's working as `lightCurve` or else the viewer component itself wouldn't render? I'm not sure.

eatyourgreens: Weird. Maybe the workflow config model is wrong?

eatyourgreens: I did a bit of a digging and `workflow.configuration.viewerType` is used to set `subject.viewer`. I'm not sure what the difference is between that and the `getViewer` helper function.
https://github.com/zooniverse/front-end-monorepo/blob/80ce4f31ce3ab3c1f5efd24c3f71e0d70f755198/packages/lib-classifier/src/store/Subject/Subject.js#L81-L109

eatyourgreens: Oh, this is very convoluted. The workflow config uses the enumeration from `@helpers/subjectViewers`, which translates `lightcurve` to `lightCurve`.
https://github.com/zooniverse/front-end-monorepo/blob/80ce4f31ce3ab3c1f5efd24c3f71e0d70f755198/packages/lib-classifier/src/helpers/subjectViewers/subjectViewers.js#L8-L11

So, if I'm reading the code correctly, the workflow config stored in Panoptes is `lightcurve`, which the classifier converts internally to `lightCurve`, which is then used to lookup the viewer component in `getViewer`.

srallen: Thanks for looking. That is really convoluted. I think I'll update this here to use `lightcurve`, then also update `lightCurve` to be `lightcurve` where applicable in the code there so it's all consistent. The inconsistency makes for easy mistakes... like I almost did here.

srallen:
```suggestion
                    <option value="lightcurve">(D3/TESS) Light Curve</option>
```

eatyourgreens: Sounds like a good plan. I haven't tried this in the code, but now that `workflow.configuration.subject_viewer` is an enumeration in the model, I think the `@helpers/subjectViewers` enumeration is redundant and could be removed, along with the `workflow.configuration.viewerType` view.

srallen: Yep it's redundant, so it could be simplified.

PR #5834, opened 2020-11-02
Project Builder: Align workflow completeness radio buttons

eatyourgreens:
Align the workflow completeness radio buttons, in the lab, so that they are above one another.

Staging branch URL: https://pr-5834.pfe-preview.zooniverse.org

**before**
Screenshot showing the workflows table in the lab. The retirement count radio button is easy to miss because it isn't beneath the classification count radio button.

**after**
Screenshot showing the same table, but now the two radio buttons are vertically aligned.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflows-table.jsx
+++ b/app/pages/lab/workflows-table.jsx
@@ -57,8 +57,8 @@ const WorkflowsTable = ({
                   Active
                 </label>
               </td>
-              <td class="stats_completeness_type">
-                <label>
+              <td>
+                <label class="standard-table__label__completeness">
                   <input
                     checked={statsCompletenessType === 'classification'}
                     name={`stats_completeness_type.${workflow.id}`}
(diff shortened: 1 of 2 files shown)


beckyrother (approved):
So nice! Thanks for doing this Jim.

eatyourgreens:
Github doesn't recognise Becky's approval as valid. 😞

app/pages/lab/workflows-table.jsx
```diff
@@ -57,7 +57,7 @@ const WorkflowsTable = ({
                   Active
                 </label>
               </td>
-              <td>
+              <td class="stats_completeness_type">
```
srallen, line 60: Non-blocking, but we were using BEM convention last on PFE, so that'd still be useful I think.

srallen (approved):

app/pages/lab/workflows-table.jsx
eatyourgreens: I took a guess at how PFE is using BEM and went for `standard-table__label__completeness` as the class name. Unfortunately, I don't think we documented how we were using BEM.

srallen: I'm not sure what would help regarding documentation, however, using the BEM standard, the block would be `standard-table`, the element is `label`, and `completeness` would be the modifier, so `standard-table__label--completeness` is how I'd do it.

PR #5847, opened 2020-11-20
Lab data export error

eatyourgreens:
Staging branch URL: https://pr-5847.pfe-preview.zooniverse.org

Closes: #5848. 

Fixes an error where the project builder crashes with `exported.metadata is null` if you try to view data exports on a project that has no exports, like I Fancy Cats (staging project 335.)

This PR adds an error boundary to lab projects, which prints out the error message and stack trace for page crashes.
The bug is fixed by adding an existence check to `exported.metadata` in the data download link component.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/project.cjsx
+++ b/app/pages/lab/project.cjsx
@@ -25,6 +25,10 @@ EditProjectPage = createReactClass
   contextTypes:
     router: PropTypes.object.isRequired
 
+  componentDidCatch: (error, info) ->
+    console.log(error, info);
+    this.setState({ error, info });
+
   getDefaultProps: ->
     project: id: '2'
     workflowActions: workflowActions
@@ -36,7 +40,16 @@ EditProjectPage = createReactClass
   labPath: (postFix = '') ->
     "/lab/#{@props.project.id}#{postFix}"
 
+  renderError: ->
+    <div>
+      <h1><code>{this.state.error.message}</code></h1>
+      <p>{!!this.state.info && <pre>{this.state.info.componentStack}</pre>}</p>
+    </div>
+
   render: ->
+    if @state.error
+      return @renderError()
+
     linkParams =
       projectID: @props.project.id
 
--- a/app/partials/data-export-download-link.jsx
+++ b/app/partials/data-export-download-link.jsx
@@ -20,7 +20,7 @@ export default class DataExportDownloadLink extends React.Component {
     }
 
     recentAndReady(exported) {
-        return exported && (exported.metadata.state === 'ready' || !exported.metadata.state);
+        return exported?.metadata && (exported.metadata.state === 'ready' || !exported.metadata.state);
     }
 
     pending(exported) {


eatyourgreens:
https://pr-5847.pfe-preview.zooniverse.org/lab/335/data-exports loads without errors.

shaunanoordin:
Code changes look legit, `exported?.metadata` is straightforward and sensible check to add. Let me run some quick tests.

srallen:
We haven't dropped IE support for PFE, only the new project app, so do be careful when using new syntax in that case.

shaunanoordin (approved):
## PR Review

This PR adds a simple existential check for a Data Export resource's metadata field. This prevents the Project Builder's Data Export page from crashing if there's a DE resource that's not in the shape it's supposed to be.

Tested on macOS + Chrome on [localhost](https://local.zoonverse.org:3735/lab/335/data-exports)

- Data Export page crashes immediately on load on `master`, but
- Data Export page renders correctly on `lab-data-export-error`

### ADDENDUM

Sarah added a comment in what I assume is in reference to the `?.` operator in `exported?.metadata`

> We haven't dropped IE support for PFE, only the new project app, so do be careful when using new syntax in that case.

This is a good point, but I thought that the Babel transpiler would have converted the operator into something safer for IE11.

Unfortunately, I have not been able to test this - I just spent the better part of the last hour struggling with VirtualBox (Win 7 + IE11) to get the localhost to run, but with no luck. (Checking simulated IE11 vs `master` and `lab-data-export-error` results in the localhost crashing when it encounters a `=>` operator (what?); IE11 plus live zooniverse.org seems fine though.) I have theories why this borked but that's irrelevant to this PR review.

### Status

LGTM 👍  I'm happy to approve this since the tests that I could run looked good, but if we were to err on the side of caution, I'm more than happy to see that `exported?.metadata` check change into `exported && exported.metadata`

srallen:
@shaunanoordin when newer syntax goes stage 4, then yes, it's included with babel's core transpiler. If PFE is on the latest version of babel, then we should be ok here. 

I think Browserstack has IE and Windows combinations you can try for testing these things? I think Microsoft stopped supporting the VMs for dev testing some time ago in favor of services like Browserstack.

shaunanoordin:
Thanks @srallen - I'm updating my testing method for IE11 to use Browser Stack instead of a local VM. 👌 

(❔  EDIT: following isn't related to PR review, but is a tangent about testing methods. Feel free to disregard for this PR.)

Frustratingly, I'm encountering the same problems on both BStack _and_ VBox (Win 7 + IE11): 1. trying to view PFE running on localhost causes a crash at the `=>` operator (both master and this PR branches), and 2. `https://pr-5847.pfe-preview.zooniverse.org` refuses to connect on IE11 (but other browsers work fine??). This is of some concern to me as it means I can't run thorough tests in certain scenarios (i.e. PFE). If someone else can take a stab at IE11, that'd be ace, (with any luck it's just some bad network config on my machine and not a failure of the testing method itself with IE11) but I don't recommend spending too much time trying to figure things out if you have other things to do.

eatyourgreens:
Browserstack also gives me an error when I'm trying to browse either this PR or https://master.pfe-preview.zooniverse.org in IE11.

eatyourgreens:
https://www.zooniverse.org works in IE11 on Browserstack, so I'm going to merge this then check it after it goes live.

shaunanoordin:
Jim, I assume you're using Browserstack to test right now, so I went ahead and tested `zooniverse.org` on Win7+IE11 VirtualBox as a supplement.

- No apparent issues with baseline `zooniverse.org` functionality
- No issues loading the Data Export page (production project 1291, at various stages of Classifications exporting), though I've unfortunately been unable to find a project that accurately replicates the export state of staging project 335... so I can't check for that specific error.

No issues detected on my end. 👌

eatyourgreens:
Browserstack won't let me check staging projects in IE11, but the lab data exports page still works for production projects. Checked against the latest commit at https://www.zooniverse.org/lab in IE11 on browserstack.com.

PR #5857, opened 2020-12-09
Allow admins to edit Training data in Workflow config

shaunanoordin:
## PR Overview

Context (Why): a Workflow's Training data can currently only be edited by manually editing the Workflow config object, e.g. by the Rails console. It would be much easier to make these minor changes via a UI. As a Hackday project, @nciemniak wants to add such a functionality to the Admin page.

Details (What): This PR allows admins to edit the Training data values in a Workflow's configuration.

- This can be accessed by going to the Admin page, "Set Project Status", selecting a Project, then scrolling down to the individual workflows.
- The Training data fields are: `training_chances`, `training_default_chance`, `training_set_ids`
- Example:
  ```
  workflow.configuration = {
    training_chances: [1, 0.5, 0.3],
    training_default_chance: 5,
    training_set_ids: [4808,4807],
  }
  ```

Screen Shot 2020-12-09 at 3 15 49 PM

_Example of the Configure Training Data section.

### Status

Ready for review!

### DIFF
--- a/app/pages/admin/project-status.jsx
+++ b/app/pages/admin/project-status.jsx
@@ -19,7 +19,7 @@ class ProjectStatus extends Component {
     super(props);
     this.onChangeWorkflowLevel = this.onChangeWorkflowLevel.bind(this);
     this.onChangeWorkflowRetirement = this.onChangeWorkflowRetirement.bind(this);
-    this.onChangeTrainingDefaultChance = this.onChangeTrainingDefaultChance.bind(this);
+    this.onBlurTrainingDefaultChance = this.onBlurTrainingDefaultChance.bind(this);
     this.onBlurTrainingSetIds = this.onBlurTrainingSetIds.bind(this);
     this.onBlurTrainingChances = this.onBlurTrainingChances.bind(this);
     this.getWorkflows = this.getWorkflows.bind(this);
@@ -39,8 +39,9 @@ class ProjectStatus extends Component {
       workflows: [],
 
       // transitional input values for text input fields, in the format of { workflow_id_1: 'string', ... }
+      valTrainingSetIds: {},
       valTrainingChances: {},
-      valTrainingSetIds: {}
+      valTrainingDefaultChance: {}
     };
   }
 
@@ -69,16 +70,6 @@ class ProjectStatus extends Component {
       .catch(error => this.setState({ error }));
   }
 
-  onChangeTrainingDefaultChance (workflow, event) {
-    this.setState({ error: null });
-    const val = event.target.value ? parseInt(event.target.value) : '';
-    if (val && isNaN(val)) return;  // Ignore invalid values (note: empty string is considered valid here)
-    // TODO: if the value is an empty string, the field should be deleted altogether.
-    return workflow.update({ 'configuration.training_default_chance': val }).save()
-      .then(() => this.getWorkflows())
-      .catch(error => this.setState({ error }));
-  }
-
   onBlurTrainingSetIds(workflow, event) {
     this.setState({ error: null });
     let val = this.state.valTrainingSetIds[workflow.id] || '';
@@ -97,6 +88,18 @@ class ProjectStatus extends Component {
       .filter((num) => { return !isNaN(num) });
     // TODO: if the value is an empty string, the field should be deleted altogether.
     return workflow.update({ 'configuration.training_chances': val }).save()
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
+  onBlurTrainingDefaultChance (workflow, event) {
+    this.setState({ error: null });
+    const val =
+      this.state.valTrainingDefaultChance[workflow.id]
+      ? parseFloat(this.state.valTrainingDefaultChance[workflow.id]) : '';
+    if (val && isNaN(val)) return;  // Ignore invalid values (note: empty string is considered valid here)
+    // TODO: if the value is an empty string, the field should be deleted altogether.
+    return workflow.update({ 'configuration.training_default_chance': val }).save()
       .then(() => this.getWorkflows())
       .catch(error => this.setState({ error }));
   }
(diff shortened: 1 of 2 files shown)


shaunanoordin:
## PR Update

Here's the plan for this PR

- [x] Add `training_default_chance` field
- [x] Add `training_set_ids`, `training_chances` fields (note: these are of type ARRAY)
  - [ ] solve the issue of array data types (Solution is to add React states to record the values in the fields while the text inputs are still being edited (as strings), and then only fire an update request to the API when the text inputs lose focus.)
- Update the tests in `project-status.spec.jsx`

### Status

WIP. We plan to continue this during the next Hackday in January 2021.

app/pages/admin/project-status.jsx
```diff
@@ -257,6 +324,52 @@ class ProjectStatus extends Component {
...
+              <div>
+                <h4>Configure Training Data</h4>
+                <label>
+                  Training Set IDs:{' '}
+                  <input
+                    id="training-set-ids"
+                    type="text"
+                    onBlur={(event) => this.onBlurTrainingSetIds(workflow, event)}
+                    onChange={(event) => {
+                      const updatedTrainingSetIds = Object.apply({}, this.state.valTrainingSetIds)
```
eatyourgreens, line 336: Should this be `Object.assign` in each of the change handlers? [`apply`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/apply) is used to call a function with an explicit `this` context.

```suggestion
                      const updatedTrainingSetIds = Object.assign({}, this.state.valTrainingSetIds)
```

app/pages/admin/project-status.jsx
```diff
@@ -257,6 +324,52 @@ class ProjectStatus extends Component {
...
+                  />
+                </label>
+                <label>
+                  Training Chances:{' '}
+                  <input
+                    id="training-chances"
+                    type="text"
+                    onBlur={(event) => this.onBlurTrainingChances(workflow, event)}
+                    onChange={(event) => {
+                      const updatedTrainingChances = Object.apply({}, this.state.valTrainingChances)
```
eatyourgreens, line 350:
```suggestion
                      const updatedTrainingChances = Object.assign({}, this.state.valTrainingChances)
```

app/pages/admin/project-status.jsx
```diff
@@ -257,6 +324,52 @@ class ProjectStatus extends Component {
...
+                  />
+                </label>
+                <label>
+                  Training Default Chance:{' '}
+                  <input
+                    id="training-default-chance"
+                    type="text"
+                    onBlur={(event) => this.onBlurTrainingDefaultChance(workflow, event)}
+                    onChange={(event) => {
+                      const updatedTrainingDefaultChance = Object.apply({}, this.state.valTrainingDefaultChance)
```
eatyourgreens, line 364:
```suggestion
                      const updatedTrainingDefaultChance = Object.assign({}, this.state.valTrainingDefaultChance)
```

eatyourgreens (commented):
I wasn't sure about `Object.apply` in the change handlers, so I've left some suggestions to use `Object.assign`.

shaunanoordin:
## PR Update

omfg, thank you @eatyourgreens for spotting the fatal mistake between `Object.apply` and `Object.assign`. This PR has been updated with the fixes

- I made a big typo back in commit 7a860d6 , so this one's on me.
- The bug in detail: 🐞 
  - When we run `const updatedTrainingChances = Object.assign({}, this.state.valTrainingChances)`, what we want to do is make a copy of `valTrainingChances` that we can safely modify.
  - For example, if the project has three workflows, each with their own training chances, we expect something like `updatedTrainingChances = { 'wf_1': '100', 'wf_2': '200', 'wf_3': '300'  }`
  - Then, if we change `updatedTrainingChances['wf_1']`, the value of workflow 1 will change while the values of workflows 2 and 3 will be preserved.
  - Using the incorrect `Object.apply()`, on the other hand, will always return `{}`. _(Instead of crashing, for some reason!)_
  - This means that updating workflow 1's training chances would have DELETED workflow 2 and 3's training chances, Whoops!
  - ❗ This wasn't caught in functionality testing because we were testing on _a single workflow_ project.

Again, big ol' thanks to Jim for spotting that massive error!

PR is once again ready for review.

app/pages/admin/project-status.jsx
```diff
@@ -62,6 +70,40 @@ class ProjectStatus extends Component {
...
+    return workflow.update({ 'configuration.training_chances': val }).save()
+      .then(() => this.getWorkflows())
+      .catch(error => this.setState({ error }));
+  }
+
+  onBlurTrainingDefaultChance (workflow, event) {
+    this.setState({ error: null });
+    const val =
+      this.state.valTrainingDefaultChance[workflow.id]
+      ? parseFloat(this.state.valTrainingDefaultChance[workflow.id]) : '';
```
eatyourgreens, line 99: This sets the value to either a number or an empty string. Just checking that the mixed types are intended? I don't see it causing any problems.

app/pages/admin/project-status.jsx
```diff
@@ -90,8 +132,33 @@ class ProjectStatus extends Component {
...
+          (wf.configuration.training_set_ids && wf.configuration.training_set_ids.join)
+          ? wf.configuration.training_set_ids.join(',')
+          : '';
+        valTrainingChances[wf.id] =
+          (wf.configuration.training_chances && wf.configuration.training_chances.join)
+          ? wf.configuration.training_chances.join(',')
+          : '';
+        // Note: please allow default chance to be 0. This is a valid edge case.
+        valTrainingDefaultChance[wf.id] =
+          (wf.configuration.training_default_chance >= 0 && wf.configuration.training_default_chance !== null)
```
eatyourgreens, line 149: I think that `>= 0` means that the default chance is never null for the second test? So I'm not sure what the second test is testing?

eatyourgreens (approved):
I've got a couple of questions where the code confused me.

app/pages/admin/project-status.jsx
shaunanoordin: Yup, that's correct! Accepted values are either numbers, or an empty string.

app/pages/admin/project-status.jsx
shaunanoordin: omfg you won't believe this one...

```
undefined >= 0 (OR) undefined <= 0 (OR) undefined == 0
result: false

"" >= 0 (OR) "" <= 0 (OR) "" == 0
result: true

null >= 0 (OR) null <= 0
result true

null == 0 (OR) null > 0 (OR) null < 0
result false
```

🤦

shaunanoordin:
Thanks again for the review Jim!

To anybody else reading this, please pay special attention to Jim's comments here - https://github.com/zooniverse/Panoptes-Front-End/pull/5857/files#r573603622 - and my response.

Because _JavaScript gotta JavaScript._

app/pages/admin/project-status.jsx
eatyourgreens: Oh no!

Oh no, facepalm GIF

eatyourgreens: In happier news, the [`??` operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) might work here.

eatyourgreens: `wf.configuration.training_default_chance ?? ''` will be `0` if the default chance is zero, but an empty string if the default chance is `null`.

PR #5897, opened 2021-02-15
Redirect for P4

srallen:
Retrying what I meant to do in #5894. I reused `RELOAD` when I shouldn't have. It does what we expect when the hostname is not `www.zooniverse.org`, but not when it is (aka production) because what its intended behavior is to reload the page to then proxy redirect to the new apps. All we want here is a simple redirect, so I added code to do just that.

Can test with:  https://pr-5897.pfe-preview.zooniverse.org/projects/mschwamb/planet-four/authors?env=production

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/router.cjsx
+++ b/app/router.cjsx
@@ -86,6 +86,13 @@ RELOAD = createReactClass
   render: ->
     null
 
+ExternalRedirect = createReactClass
+  componentDidMount: ->
+    if @props.newUrl
+      window.location = @props.newUrl
+  render: ->
+    null
+
 module.exports =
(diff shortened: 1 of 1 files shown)


mcbouslog (approved):
LGTM! Thanks for quick fix.

I see now how originally the fix was getting stuck within `RELOAD`, line 83, on production, but not staging. I'm surprised a reroute like this hasn't come up before, but scanning `router.cjsx` I don't see any other similar situation.

PR #5906, opened 2021-03-03
Add functionality to add or delete transcription task and steps

srallen:
Staging branch URL: https://pr-5906.pfe-preview.zooniverse.org

For the upcoming NEH workshop and Davy Notebooks use, the lab now has the ability to add and remove the transcription task and the associated question task for caesar retirement grouped together in a single step. If transcription task is enabled, any task that is added or removed will be also added or removed in the workflow steps. This is because some of the projects may want additional tasks. The order of the steps determines the order of the workflow, so the first task designation and setting what comes next does not matter.

Here's what you should be able to do (recommend inspecting the network response for workflow PUTs):

- Add `transcription-task` as an experimental tool for a project in the admin pages
- Then add any number of tasks preceding a transcription task, with correctly formatted `steps` and `tasks` on the workflow resource 
- Add a transcription task, which adds the task and the associated question task as a single step
- Add any number of tasks after the transcription task
- Delete any task preceding or after the transcription task and the associated step is correctly deleted without removing other steps.
- Delete the transcription task which deletes the associated question task

Whatever workflow you create, activate it, and then you should be able to view the project workflow as intended at https://frontend.preview.zooniverse.org/projects/:owner/:name/classify

TODO (in future PRs): 
- The sticky modal form component that opens when adding a task sometimes causes and invariant violation react error which sometimes prevents the PUT from succeeding. As far as I can tell, this is random. Sometimes it prevents the PUT, sometimes it doesn't. If you refresh the page after happens, you might have to try to add the task again, sometimes not. If this is happening _a lot_ we can look to remove it, but after this PR.
- If logic for branching single choice tasks needs to be supported within the steps paradigm, I'll add it separately.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -626,13 +626,15 @@ EditWorkflowPage = createReactClass
     }</PromiseRenderer>
 
   getNextTaskID: ->
+    nextTaskNumber = -1
     # This could be random, but we might as well make it sorta meaningful.
     taskCount = Object.keys(@props.workflow.tasks).length
     taskIDNumber = -1
     until nextTaskID? and nextTaskID not of @props.workflow.tasks
       taskIDNumber += 1
-      nextTaskID = "T#{taskCount + taskIDNumber}"
-    { nextTaskID, taskCount }
+      nextTaskNumber = taskCount + taskIDNumber
+      nextTaskID = "T#{nextTaskNumber}"
+    { nextTaskID, nextTaskNumber }
 
   getNextStepID: ->
     stepCount = @props.workflow.steps.length
@@ -697,8 +699,8 @@ EditWorkflowPage = createReactClass
 
   addNewTranscriptionTask: () ->
     nextStepID = @getNextStepID()
-    { nextTaskID, taskCount } = @getNextTaskID()
-    questionTaskID = "T#{taskCount + 1}"
+    { nextTaskID, nextTaskNumber } = @getNextTaskID()
+    questionTaskID = "T#{nextTaskNumber + 1}"
     changes = {}
 
     # Steps changes
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -625,11 +625,15 @@ EditWorkflowPage = createReactClass
         <span>This project has no mini-courses.</span>
     }</PromiseRenderer>
 
-  getNextTaskID: ->
+  getNextTaskID: (lastTaskNumber) ->
+    # The task ID could be random, but we might as well make it sorta meaningful.
     nextTaskNumber = -1
-    # This could be random, but we might as well make it sorta meaningful.
-    taskCount = Object.keys(@props.workflow.tasks).length
     taskIDNumber = -1
+    taskCount = Object.keys(@props.workflow.tasks).length
+    # We want to allow 0 despite it being falsey
+    unless lastTaskNumber is null or lastTaskNumber is undefined
+      taskIDNumber = lastTaskNumber
+
     until nextTaskID? and nextTaskID not of @props.workflow.tasks
       taskIDNumber += 1
       nextTaskNumber = taskCount + taskIDNumber
@@ -700,18 +704,18 @@ EditWorkflowPage = createReactClass
   addNewTranscriptionTask: () ->
     nextStepID = @getNextStepID()
     { nextTaskID, nextTaskNumber } = @getNextTaskID()
-    questionTaskID = "T#{nextTaskNumber + 1}"
+    questionTaskID = @getNextTaskID(nextTaskNumber)
     changes = {}
 
     # Steps changes
-    newStep = [nextStepID, { taskKeys: [nextTaskID, questionTaskID] }]
+    newStep = [nextStepID, { taskKeys: [nextTaskID, questionTaskID.nextTaskID] }]
     newSteps = [...@props.workflow.steps, newStep]
     changes["steps"] = newSteps
 
     # Tasks changes
-    tasks = @buildTranscriptionTask(nextTaskID, questionTaskID)
+    tasks = @buildTranscriptionTask(nextTaskID, questionTaskID.nextTaskID)
     changes["tasks.#{nextTaskID}"] = tasks[nextTaskID]
-    changes["tasks.#{questionTaskID}"] = tasks[questionTaskID]
+    changes["tasks.#{questionTaskID.nextTaskID}"] = tasks[questionTaskID.nextTaskID]
 
     configuration = {}
     # Configure version 2.0 in the workflow if missing
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -242,7 +242,7 @@ EditWorkflowPage = createReactClass
                       </AutoSave>}{' '}
                     {if @canUseTask(@props.project, "transcription-task")
                       <AutoSave resource={@props.workflow}>
-                        <button type="button" className="minor-button" onClick={() => console.log 'add transcription task'} title="Transcription tasks: the volunteer marks a line under text and transcribes the text into a text box. If caesar is configured, then text suggestions if available from other volunteers are options.">
+                        <button type="submit" className="minor-button" onClick={@addNewTranscriptionTask} title="Transcription tasks: the volunteer marks a line under text and transcribes the text into a text box. If caesar is configured, then text suggestions if available from other volunteers are options.">
                           <i className="fa fa-font fa-2x"></i>
                           <br />
                           <small><strong>Transcription</strong></small>
@@ -494,7 +494,7 @@ EditWorkflowPage = createReactClass
             TaskEditorComponent = tasks[@props.workflow.tasks[@state.selectedTaskKey].type]?.Editor
             <div>
               {if 'shortcut' in @props.project.experimental_tools
-                <ShortcutEditor workflow={@props.workflow} task={@props.workflow.tasks[@state.selectedTaskKey]} >
+                <ShortcutEditor workflow={@props.workflow} task={@props.workflow.tasks[@state.selectedTaskKey]}>
                   <TaskEditorComponent
                     workflow={@props.workflow}
                     task={@props.workflow.tasks[@state.selectedTaskKey]}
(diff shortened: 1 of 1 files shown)


snblickhan:
> Whatever workflow you create, activate it, and then you should be able to view the project workflow as intended at https://frontend.preview.zooniverse.org/projects/:owner/:name/classify

I created a test project on staging but am unable to view via this link (404 Project snblickhan/tt-staging-test was not found).

Additional questions:
- What, if anything, will the remaining fields in the editor do if used with the transcription task? e.g. retirement (I'm guessing nothing, as we're using Caesar for retirement), persist annotations (similarly nothing?), classification summaries
- Will the task instruction + help text remain the same as for the beta test? I may want to edit those instructions slightly based on beta feedback, but I can open an issue for that separately.

srallen:
@snblickhan try this: https://frontend.preview.zooniverse.org/projects/snblickhan/tt-staging-test/classify/workflow/3466?env=staging

This should load the project, but there's still a TypeError, so something still going wrong.

srallen:
> What, if anything, will the remaining fields in the editor do if used with the transcription task? e.g. retirement (I'm guessing nothing, as we're using Caesar for retirement), persist annotations (similarly nothing?), classification summaries

Depends on the feature. Linking subject sets, tutorials should work. Retirement settings are used by Panoptes and so will have to be set to never retire if you want caesar to do it instead.

Persist annotations aren't supported yet in the new classifier. Summaries aren't either and are slated for potential deprecation since most projects turn it off. 

> Will the task instruction + help text remain the same as for the beta test? I may want to edit those instructions slightly based on beta feedback, but I can open an issue for that separately.

I have the transcription task setup to use that text when it's added, but I can update it to be whatever you like or change it to be an empty string. Once added, project builders can also modify the instruction and help text at will.

snblickhan:
> This should load the project, but there's still a TypeError, so something still going wrong.

Same error on my end.

> Once added, project builders can also modify the instruction and help text at will.

Oh that's helpful to know -- I wasn't sure if the production version would look exactly like this, or if those will be editable.

Re: retirement, etc. I'll make sure to touch base with you about all the features on this page so I can accurately communicate necessary settings to the NEH institute cohort!

shaunanoordin:
## Quick Review

- Code read looks good based on, uh, my shaky understanding of CoffeeScript.
- Functionality tests are a bit all over the place though. 😵 
  - When I add a new Transcription "Task", it adds two tasks (the transcription-drawing task, and the transcription-question task, as expected)
  - However, the added tasks display on the Workflow editor with the text "Task editor is unavailable" and a bunch of console error messages _until I refresh the page)
  - Adding tasks works... except when it doesn't. (i.e. refreshing the page results in no tasks being added)
    - Not sure if this is related to the comment, because I can't replicate my error consistently.
      > The sticky modal form component that opens when adding a task sometimes causes and invariant violation react error which sometimes prevents the PUT from succeeding. As far as I can tell, this is random.
  - Deleting _either_ the transcription-drawing task _or_ the transcription-question task of a Transcription "Task" correctly deletes _both_ the transcription-drawing task and the transcription-question task, as expected.
    - This is true whether I start by deleting the transcription-drawing task or the transcription-question task.
    - Looks good, though I'm going to give the CJSX a second pass to properly grok how this works.
- Attempting to preview the workflow with the transcription task results in a "Cannot read property 'type' of undefined" error that crashes the classifier.

I think I know where to start looking to debug this, so let me take a stab at a troubleshoot before I update my notes. brb!

shaunanoordin:
I'm right - even Will's baseline **Anti-Slavery Testing** workflow is broken. ❗ 

https://master.pfe-preview.zooniverse.org/projects/wgranger-test/anti-slavery-testing/classify?reload=0&workflow=3392

So it's not just Sam's WF and my WF that's broken, it's something that's borked with the Transcription system altogether. This should be easy enough to solve:

- Error message is "a.default[o.type].getDefaultAnnotation is not a function" on the pfe-preview site
- So my guess is... er... at some point... we added `getDefaultAnnotation` to the various PFE tasks in the Classifier, but the Transcription Task was  forgotten.
- I'm going to do some time travel to see when precisely this borked.

brb!

srallen:
I'm looking into the issue this morning why my test project worked but all others seem broken.

srallen:
@shaunanoordin the classifier on PFE will be broken for any project using the transcription task. Be sure to use https://frontend.preview.zooniverse.org/projects/:owner/:name/classify

shaunanoordin:
Ah, thanks Sarah! I had the mistaken impression that we had transcriptions working for classic PFE. I could have sworn I tested Will's Anti-Slavery Testing on PFE before - I must be losing my mind.

srallen:
@shaunanoordin The project has several other workflows that presumably work in the PFE classifier.

srallen:
It seems like what may be the issue with the modal form is really interfering with people's abilities to test this out. I'll look into that first.

srallen:
@shaunanoordin I think #5911 will help some of the issues we all have been seeing with this new feature. Could you review that and we can get it merged and rebased against this one to make this one easier to review?

shaunanoordin:
On it, let me reset my workspace and I'll get 5911 tested.

shaunanoordin (changes requested):
## PR Review

This PR adds the ability to add and remove "Transcription Tasks" via the PFE project builder.

- A "Transcription Task" is actually a combination of 1 transcription-drawing task, and 1 transcription-question task. These are added and deleted simultaneously.
- A new "Transcription Task" can be **added** via the standard method
  - Build A Project -> Workflow -> (choose a WF) -> Add A Task -> Transcription Task.
- "Transcription Tasks" can be **deleted** via the standard method
  - Build A Project -> Workflow -> (choose a WF) -> (choose Task) -> Delete Task

### Testing

Tested on macOS 10 + Chrome 88, on localhost.

**Basic functionality tests:**

- ✅ I can add and delete standard Question, Drawing tasks

LGTM 👍 

**Main functionality tests:**

- ✅ I can add "Transcription Tasks".
  - "Transcription Tasks" come in pairs of one transcription-drawing task and one transcription-question task.
  - I can add a "Transcription Task" as the **first task(s)** of the workflow. (Resulting in new tasks T0, T1.)
  - I can add a "Transcription Task" **after an existing** Question task, T0. (Resulting in new tasks T1+T2)
  - I can add a Question task **after an existing** Transcription Task T1+T2. (Resulting in new task T3)
- ✅ I can delete "Transcription Tasks".
  - If I delete the transcription-drawing task, the transcription-question task will also be delete.
  - Same thing, vice versa.
- ✅ I can view the workflow that uses "Transcription Tasks" on the monorepo Classifier

Main code checks out fine! Now let's get dangerous.

**Additional functionality tests:**

- ❌ I cannot properly add a new "Transcription Task" if there is a "hole" in the series of task numbers. 🤷‍♂️ 

### Issues Detected

**Functionality Bug:** adding a new "Transcription Task" gets wonky if the sequence of Task IDs isn't "perfect". (i.e. `[T0, T1, T2, ...]` with no missing skipped T-numbers. )

Replication:
- Create a new empty workflow
- Create TWO new _Text Tasks._ These will result in a sequence of Tasks `[T0, T1]`
- Delete T0. The sequence of Tasks is now `[T1]` 
- Attempt to add a new "Transcription Task"
  - Expected: "Transcription Task" should add to T2 and T3, resulting in `[T1, T2, T3]` (text, transcription-drawing, transcription-question)
  - Actual: "Transcription Task" attempts to add _T1_ and _T2_, resulting in `[T1 (original), T2]` (text, transcription-question). The transcription-drawing task vanishes into the ether.

Alternate replication: 
- Create a new empty workflow
- Create THREE new _Text Tasks._ These will result in a sequence of Tasks `[T0, T1, T2]`
- Delete T0 and T1. The sequence of Tasks is now `[T2]` 
- Attempt to add a new "Transcription Task"
  - Expected: "Transcription Task" should add to T3 and T4, resulting in `[T2, T3, T4]` (text, transcription-drawing, transcription-question)
  - Actual: "Transcription Task" attempts to add _T1_ and _T2_, resulting in `[T1, T2 (transcription)]` (transcription-drawing, transcription-question). The text tasks gets overwritten. (I wonder why it's different this time, and the transcription-question didn't vanish instead??)

⚠️ Warning: these replication steps WILL screw up the `workflow.steps` value, leaving a bunch of orphaned steps all over the place. You need to fix `workflow.steps` via CLI or ZooAPI, or deleting the workflow and starting again. If you don't, you're going to encounter very weird behaviour in the classifier.

Analysis: 
- OK, I think I've figured out where this is going wonky. It's this line of code in `addNewTranscriptionTask()`:
  ```
  { nextTaskID, taskCount } = @getNextTaskID()
  questionTaskID = "T#{taskCount + 1}"
  ```
- questionTaskID seems to _assume there's always a free space_ at taskCount+1, which causes all sorts of shenanigans when you don't have a perfect sequence of Task IDs.
 
- Side note: `getNextTaskID()` has a weird quirk ~~and I'm not sure how important this is.~~
  - The function always assumes that existing tasks are in a _perfect sequence._ Hence, it doesn't assume it should start with T0, but with T(taskCount). So if the sequence of tasks is [T7, T8], the next generated ID is T2. (Placing it before the initial two questions. This annoys my Lawful OCD alignment.)
  - ~~This "not starting at T0" quirk is harmless for every non-Transcription Task, which only needs one unique ID. For the Transcription Task, however... does it become a bug? Have you seen my replication notes?? Even I'm not sure WTF is going on.~~
  - UPDATE: `getNextTaskID()` is harmless and doesn't need to be changed/optimised any further for this PR. 👍 
    - I just figured out WTF happened in my replication steps. (The first one, not the alternative one.) When we have [T1] and then add a Transcription Task, the _transcription-drawing task_ first calls getNextID() and gets an ID of T2. However, the _transcription-question task_ then stumbles in drunk and late to the party, immediately declares _"My tashk ID is T#{tashkCount + 1} (hic) sho that makesh me... T2!"_ and the _transcription-drawing task,_ despite claiming the first legit slot, gets bumped off.

### Status

I think the functionality bug is the last thing that needs to be fixed before this is good to go! All other functionality checks out and the code read looks good.

Changes requested:
- Find a better way to generate `questionTaskID`
- EDIT: Wait, can't it just be `questionTaskID = "T#{nextStepID}.1"` or `questionTaskID = "T#{nextStepID}_question"` or `questionTaskID = "T#{nextStepID}meepmorpbleepblorp"` ? It doesn't have to be an integer number, just a unique string, innit?

srallen:
@shaunanoordin using the `taskCount` number to increment was a mistake. I've updated the `nextTaskID` method to return the number used by the `nextTaskID` and incremented that instead. I think the scenario you identified should work now.

shaunanoordin:
Thanks for making the changes @srallen ! The update fixes one scenario, but unfortunately I can still trigger the bug in _another scenario_ under specific circumstances:

**Scenario 1:** 👍 
- Create two tasks, `[T0, T1]`
- Delete T0, leaving `[T1]`
- Add a Transcription Task.
  - Success! Transcription Task added as T2, T3.

**Scenario 2:** ❌ 
- Create _three_ tasks, `[T0, T1, T2]`
- Delete T0 and T1, leaving `[T2]`
- Add a Transcription Task
  - Failure. Transcription Task added as T1, T2 - this overwrites the existing T2 and scrambles workflow.steps

Scenario 2 fails because `getNextTaskID()` will start with _nextTaskNumber = 1,_ and since T1 **is** an available "slot" for the drawing task ID, the question task ID is automatically T2 because that doesn't check for slot conflict.

### Possible Solution?

Nuts, apologies for my earlier review comments - it seems that the problem _might be getNextTaskID()_ after all. Maybe the best way to solve this bug is to address that function.

Here's my idea: instead of the current logic (find the _first available "slot",_ starting from the _lowest_ reasonable ID number), why not go through _all_ the Tasks to find the _highest current ID_ and _increment_ from there?

Proposed JS code + testing examples (run in the console!):
```
// Go through all the existing task IDs, find the highest ID number, then provide the next available one.
function getNextTaskID (arrayOfTaskIds) {
  return arrayOfTaskIds
    .filter(id=>/^T\d+$/.test(id))  // Pick only valid IDs
    .map(id=>parseInt(id.substring(1)) + 1)  // Transform to integers. Plus 1 because we want the NEXT available task ID. 
    .reduce((largest,id)=>Math.max(largest,id), 0)  // Pick the largest next possible ID, defaulting to 0
}

array1 = ["T0", "T1"]
array2 = []
array3 = ["T0", "T1", "T3", "nonsense", "T999"]
array4 = ["utter", "nonsense", "beepboop"]

console.log(
  'Here are some test results: ',
  getNextTaskID(array1),
  getNextTaskID(array2),
  getNextTaskID(array3),
  getNextTaskID(array4),
)
// Returns 2, 0, 1000, and 0
```

Proposed Coffeescript version:
```
getNextTaskID: ->
  nextTaskNumber = Object.keys(
    @props.workflow.tasks
  ).filter((id) ->
    /^T\d+$/.test id
  ).map((id) ->
    parseInt(id.substring(1)) + 1
  ).reduce ((largest, id) ->
    Math.max largest, id
  ), 0

  nextTaskID = "T#{nextTaskNumber}"    
  { nextTaskID, nextTaskNumber }

# Since by definition nextTaskID (used by drawing task) is the highest available number, nextTaskID + 1 (used by question task) will never have a conflict.
```

It's your call if you think this version is fit for purpose, or if there's a cleaner way of generating a transcription-question Task ID that avoids conflict. 👌

srallen:
@shaunanoordin good catch. The way the `nextTaskID` works fine if you aren't trying to add more than one task at a time like  I am. Your proposed solution would absolutely work, however, I want to avoid using a regex if we can. I think I came up with a simpler version since I'm returning the ID number that was last used already, so let's reuse this for generating the question task ID for the transcription task:

```
  getNextTaskID: (lastTaskNumber) ->
    # The task ID could be random, but we might as well make it sorta meaningful.
    nextTaskNumber = -1
    taskIDNumber = -1
    taskCount = Object.keys(@props.workflow.tasks).length
    # We want to allow 0 despite it being falsey
    unless lastTaskNumber is null or lastTaskNumber is undefined
      taskIDNumber = lastTaskNumber

    until nextTaskID? and nextTaskID not of @props.workflow.tasks
      taskIDNumber += 1
      nextTaskNumber = taskCount + taskIDNumber
      nextTaskID = "T#{nextTaskNumber}"
    { nextTaskID, nextTaskNumber }
```

For all other cases, the task id number starts as -1 and increments up to the next available, however, for the transcription task, we pass in the new parameter to start where we last left off in `addNewTranscriptionTask`:

```
    { nextTaskID, nextTaskNumber } = @getNextTaskID()
    questionTaskID = @getNextTaskID(nextTaskNumber)
```

This fixes the scenario where T2 is already taken and the new set of tasks for transcription end up being T1 and T3. Adding any additional tasks continue to increment up from T3. (It also generates it as T1 and T3 if T0 already exists, but they're unique and aren't clashing, so I'm gonna let that ride!)

shaunanoordin (approved):
## PR Review (Update)

Follows [first PR Review](https://github.com/zooniverse/Panoptes-Front-End/pull/5906#pullrequestreview-610410907) and [follow up](https://github.com/zooniverse/Panoptes-Front-End/pull/5906#issuecomment-797831198)

LGTM, @srallen! 👍

- New change addresses the issue where Task IDs could conflict with one another.
- getNextTaskID() now takes an optional "previously generated task ID" parameter so it can start incrementing from that ID, instead of always starting from T0.

### Testing

Same tests as before, except now everything is fine since getNextTaskID() never generates conflicting IDs. 👍

- I can add/delete Transcription Tasks
  - I can add Transcription Tasks even if there are existing tasks...
  - ...in pretty much every _potentially problematic permutation_ of existing tasks IDs I could think of. (`[T0]`, `[T1]`, `[T1, T2]`, etc etc)
- Other tasks work fine. 
  - I can add a Text Task before adding Transcription Tasks. This Text Task can be deleted.
  - I can add a Text Task after a Transcription Task. This Text Task can be deleted.
- Every Task will have unique Tasks IDs.
- Drawing Tasks can be worked with on the FEM preview URL.

Quirks (minor):
- The Transcription-Drawing and Transcription-Question tasks might "skip" an ID in a sequence in some cases (e.g. they're generated as T1 and T3, instead of T1 and T2) but this is not a breaking issue. All IDs are legit.

### Dev Notes

After staring at the `getNextTaskID()` logic for so long, my spidey senses tell me that I should _also_ run extra checks on `getNextStepID()`. The Task ID logic is now very thorough in checking for conflicting IDs, but the Step ID logic seems too simple to prevent similar issues. Hmm!

That said, I can't think of how I might break it, and my recent Drawing Task-focused tests didn't run into any issues, so I'll save a deep dive into getNextStepID as a mid-week mini assignment.

### Status

Noice! 👍 This PR is looking great and it's exciting that project owners will soon be able to add/edit/remove their own Transcription Tasks. Let's goooo!

PR #5908, opened 2021-03-05
remove hidden project prefs from the logged in user ribbon

camallen:
filter any user project preferences that have the settings object marked as `preference.hidden=true`

Currently relies on manually setting the value in the API db manually, however changes in https://github.com/zooniverse/panoptes/pull/3583 will expose this `hidden` attribute for update via the API.

I'm happy to modify any testing user project preferences to have this value. 

Staging branch URL: https://pr-5908.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/home-for-user/index.jsx
+++ b/app/pages/home-for-user/index.jsx
@@ -126,8 +126,12 @@ export default class HomePageForUser extends React.Component {
         if (meta.page !== meta.page_count) {
           getRibbonData(user, meta.page + 1);
         }
-        // filter out projects you haven't classified on.
-        let activePreferences = projectPreferences.filter((preference) => { return preference.activity_count > 0; });
+        // filter out projects you haven't classified on AND that are not marked as hidden.
+        let activePreferences = projectPreferences.filter((preference) => {
+          let userHasClassified = preference.activity_count > 0;
+          let isVisiblePreference = !preference.settings['hidden'];
+          return userHasClassified && isVisiblePreference;
+        });
         // get the projects that you have classified on, if any.
         if (activePreferences.length > 0) {
           activePreferences = activePreferences.map((preference, i) => {


app/pages/home-for-user/index.jsx
```diff
@@ -126,8 +126,12 @@ export default class HomePageForUser extends React.Component {
         if (meta.page !== meta.page_count) {
           getRibbonData(user, meta.page + 1);
         }
-        // filter out projects you haven't classified on.
-        let activePreferences = projectPreferences.filter((preference) => { return preference.activity_count > 0; });
+        // filter out projects you haven't classified on AND that are not marked as hidden.
+        let activePreferences = projectPreferences.filter((preference) => {
+          let userHasClassified = preference.activity_count > 0;
+          let isVisiblePreference = !preference.settings['hidden'];
```
camallen, line 132: I'm wondering what the data submitted by the UI would look like here? I assume it'll be similar to the data submitted for a json object via a checkbox, e.g. value will be true via https://github.com/zooniverse/Panoptes-Front-End/blob/35b3e059410509a4134e7cf213eae03d1a1a6e95/app/lib/handle-input-change.js#L14

but if that's not true then this boolen value test could perhaps this could check for the presence of a hidden attribute on the settings object, e.g. `settings.hasOwnProperty('hidden')`

thoughts?

eatyourgreens: I think it depends if the API uses a boolean in the JSON:
```json
"preference": { "settings": { "hidden": true } } } 
```

I know that workflow tasks sometimes send back `task.required` as `"true"` instead of `true`, but that's down to a bug in `handleInputChange`, long since fixed, where all values were converted to strings before saving. The result of that is that we can check for a boolean from anything created recently in the project builder, but older projects might still use string values for boolean or number properties.

Over time, that will presumably fix itself as project teams update and save their workflows with the fixed code.

I'd assume the same goes for user prefs. TLDR: I think you're safe checking for a boolean here, as this is a recent addition to the API.

eatyourgreens (approved):
LGTM! I say let's ship it and try it out.

PR #5911, opened 2021-03-08
Remove TriggeredModalForm from task add UI

srallen:

Related to #5906. The new transcription task lab UI revealed some buggy behavior having to do with adding tasks. Randomly the PUT request when clicking to add a task type wouldn't happen. My guess is because there is an uncaught React invariant violation happening with the `TriggeredModalForm` wrapping the UI. I think there may be a race condition with it closing and unmounting which throws the error and the workflow resource PUT. Try adding tasks on the master branch, and you inconsistently get successful PUT requests and always this invariant violation at roughly the same time. 

To hopefully solve this problem, I've removed `TriggeredModalForm` and just have the buttons render inline kind of like an according. Here's what it looks like with a project workflow with all the task types enabled:

add-task-area

How it will look for most project builders:

Screen Shot 2021-03-08 at 2 56 02 PM
 
With this change, you can inspect the network requests and the javascript console to see the invariant error is no longer thrown and the PUT succeeds every time now.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -41,6 +41,7 @@ EditWorkflowPage = createReactClass
     deletionInProgress: false
     deletionError: null
     workflowCreationInProgress: false
+    showTaskAddButtons: false
 
   workflowLink: ->
     [owner, name] = @props.project.slug.split('/')
@@ -73,6 +74,9 @@ EditWorkflowPage = createReactClass
   isThereNotADefinedTask: () ->
     workflowTasks = Object.keys(@props.workflow.tasks)
     workflowTasks.length is 0
+
+  showTaskAddButtons: () ->
+    @setState((prevState) => { showTaskAddButtons: !prevState.showTaskAddButtons })
 
   render: ->
     window.editingWorkflow = @props.workflow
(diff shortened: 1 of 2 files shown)


shaunanoordin:
## Quick Review

⚠️ 😞 Nyaargh, my tests are currently showing that I can _still_ replicate the problem, despite this PR's attempts to fix. Please see the MOV file: https://drive.google.com/file/d/1cCskjcHQPwHXXLr3-G9v469VK1mscz41/view?usp=sharing (16MB, requires Zooniverse account)

Tested on localhost on a [test workflow](https://local.zooniverse.org:3735/lab/1929/workflows/3470) (branch `remove-modal-form-task-editor`, I had to check this a few times) with macOS 10 + Chrome 88

OK, let's break it down.

Issue:
- As demonstrated in the video, it's _possible_ that when creating a new Task for a workflow (e.g. in the Workflow editor, click Add Task -> Question), the PUT request (to update the workflow resource on Panoptes) _doesn't_ trigger.
- It still only happens randomly though, which is dang annoying. (I'm getting it about once every 10 attempts)
- Replication steps:
  - Workflow Editor -> Add Task -> Question
  - Monitor network traffic and see when there's a missing PUT request.
  - Repeat about a dozen times for a single instance of the issue. (Protip: hire an infinite number of monkeys)
- There doesn't seem to be any errors in the console specific to the PUT request not triggering. (WTF?)

Thoughts:
- I'm going to look into this - I'll start spamming `console.log()`s to track every step from "click on Add Question Task button" to "send PUT request" and see what might be up. I do think you're right that it's a race condition somewhere, but it may be separate from the TriggeredModalForm. (Or maybe it's the TriggeredModalForm AND something else, dangit)
- ❗ That said, I suspect that this PFE Lab problem _might_ be a can of worms that will take a lot of time to debug. _If_ this is the case, we might just need to mark this as a "to fix later" issue and just try to get 5906 running with a bunch of warnings to the users. Here be dragons.exe !
- ❗ The reason why this might be a "to fix later" issue is that I think the problem has been going on for a long time... but you won't catch it, because **if you edit a Task after it's created** (e.g. by editing the question in a Question Task), **that Task will become legit** because there are follow up PUT requests to update the Workflow.

shaunanoordin:
Notes:
- OK, so there _is_ a warning when the "Add Question Task" button is clicked, but it's the same whether the PUT request triggers or not: `Warning: ChangeListener: isMounted is deprecated. Instead, make sure to clean up subscriptions and pending requests in componentWillUnmount to prevent memory leaks.` This seems irrelevant, but I might swing back and check this later.
- I'm trying to figure out if it might be my network, but it can't be, because I'd _at least see the request go through and fail, right?_

Time for them console logs.

shaunanoordin:
Me: adds console.log() to `EditWorkflowPage.addNewTask` to trace path of error

Code: "your 'add new Question Task' action will work perfectly for your next 23 tries."

Me: reverts changes

Code: "your very next attempt will trigger the error you're testing for, tee hee!"

Me: "Why??"

shaunanoordin:
```
// EditWorkflowPage, app/pages/lab/workflow.cjsx
addNewTask: (type) ->
  ...
  @props.workflow.update changes
  // 'changes' is a legit value. So that means something inside props.workflow.update() is randomly breaking, somehow.
  @setState { selectedTaskKey: nextTaskID, showTaskAddButtons: false }
  // setState here definitely works since changes are visible on the workflow editor.
```

Wait, could this be a `panoptes-client` issue? brb, going to wipe and reinstall.

shaunanoordin:
Update: wiped, reinstalled, but bug can still be replicated (randomly). PFE is using `panoptes-client` 3.0.0, but I think we're up to 3.3.0 right now. Gonna try to see if 3.3.0 magically fixes things. brb.

shaunanoordin:
Update update: 3.3.0 doesn't magically fix things, (bug can still be replicated randomly) though PFE should probably get updated anyway. Going to follow up with this tomorrow. Night!

srallen:
@shaunanoordin it's disappointing to hear we still have random failures. I'm going to guess now that it is a bug in PJC that is causing the problem. Let's scope this back to just removing the `TriggeredModalForm` which fixes the invariant error and in my opinion, improves the UI when you have access to all the experimental task types. I think the tooltip like pop up with the long list is a bit awkward, to be honest.

shaunanoordin (approved):
> I'm going to guess now that it is a bug in PJC that is causing the problem. Let's scope this back to just removing the TriggeredModalForm which fixes the invariant error and in my opinion, improves the UI when you have access to all the experimental task types.

Sounds good to me! 👍 

## PR Review

This PR updates the UI of the "Add Task" in the Workflow editor.

- Example URL: https://pr-5911.pfe-preview.zooniverse.org/lab/1929/workflows/3470
- When clicking on "Add Task", the list of possible Task Types no longer appears as a modal/popup.
- Instead, the list of possible Task Types to choose from appears as a standard panel/div element, which disappears only when either 1. the "Add Task" is clicked again, or 2. a Task is chosen from the list.
- Accessibility: keyboard nav works fine 👍 

Code read looks good, and the UI functionality (we're ignoring the Workflow PUT shenanigans, which seems to be a PJC issue) tests fine on localhost and on the [staging branch](https://pr-5911.pfe-preview.zooniverse.org/lab/1929/workflows/3470), with mac OS 10 + Chrome 88 using a simple test project.

### Status

LGTM! 👍

PR #5918, opened 2021-03-22
Replace Travis CI with GitHub Actions

eatyourgreens:
Remove travis.yml. Add a GH workflow which runs the CI tests, generates a coverage report and posts the results to Coveralls.

Staging branch URL: https://pr-5918.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -5,7 +5,7 @@ on:
   push:
     branches:    
       - master
-
+  workflow_dispatch:
 jobs:
   build:
    runs-on: ubuntu-latest
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,0 +1,37 @@
+name: CI Tests
+
+on:
+  pull_request:
+  push:
+    branches:    
+      - master
+
+jobs:
+  build:
+   runs-on: ubuntu-latest
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,37 @@
+name: CI Tests
+
+on:
+  pull_request:
+  push:
+    branches:    
+      - master
```
srallen, line 7: Why would we need to run tests on push to master too? Master branch is protected from direct pushes and we can set branches to be required to be up to date.

.github/workflows/ci-tests.yml
```diff
@@ -0,0 +1,37 @@
...
+
```
srallen, line 8:
```suggestion
  workflow_dispatch:
```
This allows us to run the action ad hoc from the actions tab.

srallen (approved):
We'll need to remove the travis ci requirement from the repo settings.

.github/workflows/ci-tests.yml
eatyourgreens: I think that's to catch regressions but, yeah, we could require branches to be rebased.

The monorepo runs CI on merges to master too. Maybe we should turn that off?

PR #5920, opened 2021-03-23
Transcription Task: add 3 new text tags for subtask

shaunanoordin:
## PR Overview

The _Davey Notebooks_ project wants to add three new tags to the Transcription task's Text subtask. This PR makes those tags available to any Transcription-type project.

- Added new tags to the Transcription -> Text subtask: `underline`, `superscript`, `&`
  - These tags are fairly generic ; I see no issue if other transcription projects want to use them.

Screen Shot 2021-03-23 at 13 29 44

### Testing

Go ahead and play with the workflow settings:

- Project builder: https://local.zooniverse.org:3735/lab/1929/workflows/3459
- Classifier: https://fe-project.preview.zooniverse.org/projects/darkeshard/davy-notebooks-2021/classify/workflow/3459?env=staging

### Status

Ready for review

### DIFF
--- a/app/classifier/tasks/transcription/SubtaskEditor.jsx
+++ b/app/classifier/tasks/transcription/SubtaskEditor.jsx
@@ -62,6 +62,21 @@ export default function SubTaskEditor({ subtask, subtaskPrefix, workflow }) {
           <input type="checkbox" value="unclear" checked={tagExists('unclear')} onChange={updateTags} />
           {"Unclear"}
         </label>
+        <br/>
+        <label>
+          <input type="checkbox" value="underline" checked={tagExists('underline')} onChange={updateTags} />
+          {"Underline"}
+        </label>
+        <br/>
+        <label>
+          <input type="checkbox" value="superscript" checked={tagExists('superscript')} onChange={updateTags} />
+          {"Superscript"}
+        </label>
+        <br/>
+        <label>
+          <input type="checkbox" value="&amp;" checked={tagExists('&')} onChange={updateTags} />
+          {"&"}
+        </label>
       </section>
     </div>
   )


app/classifier/tasks/transcription/SubtaskEditor.jsx
```diff
@@ -62,6 +62,21 @@ export default function SubTaskEditor({ subtask, subtaskPrefix, workflow }) {
...
+        </label>
+        <br/>
+        <label>
+          <input type="checkbox" value="superscript" checked={tagExists('superscript')} onChange={updateTags} />
+          {"Superscript"}
+        </label>
+        <br/>
+        <label>
+          <input type="checkbox" value="&amp;" checked={tagExists('&')} onChange={updateTags} />
+          {"&"}
```
eatyourgreens, line 78: Does JSX automatically escape entities in strings but not in attributes? That's confusing.

eatyourgreens (approved):

app/classifier/tasks/transcription/SubtaskEditor.jsx
shaunanoordin: Yep, that's JSX for ya - if the code ain't in `{javascript_escape_brackets}`, you're writing the value in ol' HTML. I could have made it less confusing by writing `value={'&'}` but that's still a bit chonky.

eatyourgreens: Yeah, I had to stop and think about that for a second. I did see that the code sets the input's ID to something like `T0.0.0-&` in the DOM, but the browser has no problems parsing that and it didn't break the accessible label so 🤷‍♂️ .

PR #5925, opened 2021-03-31
Prevent tasks linking to themselves

### LINKED ISSUE #2512: Prevent the project builder from making a recursive decision tree if they add in a new question before the previous one entered into the workflow
I've now done this for all 3 workflows I've developed with the project builder.I write the first question and then we end up adding in one before. In the workflow builder, the drop down menu lets you select the current task to be the next task . I seem to always select it accidentally the first go around. it seems pretty straigh forward to not include dropdown menu for the next task.

eatyourgreens:
Add a task prop to the `NextTaskSelector` in the lab and pass down the current task. Filter the current task out of the list of next task options.

This turned out to be more complicated than I thought because each task (`app/classifier/tasks`) implements its own 'Next task' menu in its editor component. I think I got them all. You can check it by editing tasks in the project builder. You shouldn't be able to link any task back to itself any more.

Staging branch URL: https://pr-5925.pfe-preview.zooniverse.org

Fixes #2512.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/combo/editor.cjsx
+++ b/app/classifier/tasks/combo/editor.cjsx
@@ -77,7 +77,7 @@ ComboTaskEditor = createReactClass
       <p>
         <label>
           Next task:{' '}
-          <NextTaskSelector workflow={@props.workflow} value={@props.task.next} onChange={@setNextTask} />
+          <NextTaskSelector task={@props.task} workflow={@props.workflow} value={@props.task.next} onChange={@setNextTask} />
         </label>
         <br />
         <span className="form-help">This overrides anything set by a sub-task.</span>
--- a/app/classifier/tasks/dropdown/editor.jsx
+++ b/app/classifier/tasks/dropdown/editor.jsx
@@ -216,6 +216,7 @@ export default class DropdownEditor extends React.Component {
           <span className="form-label">Next task</span>
           <br />
           <NextTaskSelector
+            task={this.props.task}
             workflow={this.props.workflow}
             name={`${this.props.taskPrefix}.next`}
             value={this.props.task.next || ''}
--- a/app/classifier/tasks/generic-editor.cjsx
+++ b/app/classifier/tasks/generic-editor.cjsx
@@ -115,7 +115,7 @@ module.exports = createReactClass
                       <div className="workflow-choice-setting">
                         <AutoSave resource={@props.workflow}>
                           Next task{' '}
-                          <NextTaskSelector workflow={@props.workflow} name="#{@props.taskPrefix}.#{choicesKey}.#{index}.next" value={choice.next ? ''} onChange={handleChange} />
+                          <NextTaskSelector task={@props.task} workflow={@props.workflow} name="#{@props.taskPrefix}.#{choicesKey}.#{index}.next" value={choice.next ? ''} onChange={handleChange} />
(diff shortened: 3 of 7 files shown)


shaunanoordin:
I'm trying to think of scenarios where a Task WOULD like to link to themselves. Uh...

T0: Are we there yet? Are we there yet? Are we there yet?
Answer 0: No -> Go to T0
Answer 1: Yes -> Submit and move to next classification

That doesn't make sense, so it the "let's not allow tasks to link to themselves" rule is valid. Let me think of Combo Tasks...

shaunanoordin:
I suppose, in theory, it's possible somebody might want to create a combo task like...

T0: Do you see one or more muppets in this photo?
Yes -> go to T1 (combo task)
No -> submit

T1.0: Describe one of the muppets.
(Text)

T1.1: Is there another muppet you want to talk about?
Yes -> go to T1 (combo task)
No -> submit

--------

EDIT: I've just tested this on PFE, and it's NOT possible to branch from a Combo task. A Combo Task on the PFE classifier has only ONE "Next Task", and ignores any "Next Task" logic of its children Tasks.

(Maybe I'm thinking of the front-end-monorepo... does that have more complicated branching logic?)

In any case, I see no logical reason for a Task to ever loop in on itself so immediately, so the goal of this PR is legit. I'll proceed with functionality tests next.

--------

NOTE: It will _still_ be possible to lock yourself in a two-step loop, e.g. T0 leads to T1, leads to T0, leads to T1, etc but [this PR can only fix so much insanity.](https://tvtropes.org/pmwiki/pmwiki.php/Main/UnwinnableByInsanity)

eatyourgreens:
Ha! Every now and again we get a report of a broken workflow where it turns out the solution is that Next Task has been accidentally set to something other than ‘submit the classification’. 

We could maybe have a check for that too. Might be complicated for branching workflows.

shaunanoordin (approved):
## PR Review

This PR changes the Workflow editor page, preventing any given Task from setting its "next task" to be _itself._

- i.e. this PR prevents the workflow logic from looping endlessly on one task.
- ⚠️ It is STILL possible to create a two-step loop, e.g. T0's next task is T1, and T1's next task is T0. This PR will not fix such a scenario, and trying to detect infinite loops in the logic is a gateway to madness.

Code read looks good, and functionality checks out.

### Testing

Tested on localhost + macOS 10 + Chrome 89, using a random test project.

Replication steps
- create 3 tasks, T0, T1, T2
- for T0, look at the options available for "Next Task"
  - The list should include T1, T2, and "submit"
  - The list shouldn't include T0
- Repeat for T0
- Repeat for different Task types

Tests look good for the Task types: Question, Drawing, Text, Survey, Combo. I'm happy to give this a blanket 👍 for all task types.

Additional notes on testing further Task types:
- Main logic change is in <NextTaskSelector>, which now requires a "current task" prop to check against.
- A quick look through every use of <NextTaskSelector> in the repo indicates that this "current" task is being passed down properly.
- This propagated logic seems legit, so a thorough test of e.g. Dropdown Task seems like overkill.

### Status

LGTM! As noted in earlier comments, I don't see any logical reason for a Task to link to itself, so this PR goals are legit to me.

PR #5943, opened 2021-05-11
[Security] Upgrade chart.js to version 3

eatyourgreens:
Upgrade `chart.js` to fix a security vulnerability in version 2.7.

Staging branch URL: https://pr-5943.pfe-preview.zooniverse.org

I think `chart.js` is only used on project stats pages
eg. https://pr-5943.pfe-preview.zooniverse.org/projects/stephenresearch/beluga-bits/stats?env=production

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7101,49 +7101,14 @@
       "dev": true
     },
     "chart.js": {
-      "version": "2.7.3",
-      "resolved": "https://registry.npmjs.org/chart.js/-/chart.js-2.7.3.tgz",
-      "integrity": "sha512-3+7k/DbR92m6BsMUYP6M0dMsMVZpMnwkUyNSAbqolHKsbIzH2Q4LWVEHHYq7v0fmEV8whXE0DrjANulw9j2K5g==",
-      "requires": {
-        "chartjs-color": "^2.1.0",
-        "moment": "^2.10.2"
-      },
-      "dependencies": {
-        "moment": {
-          "version": "2.23.0",
-          "resolved": "https://registry.npmjs.org/moment/-/moment-2.23.0.tgz",
-          "integrity": "sha512-3IE39bHVqFbWWaPOMHZF98Q9c3LDKGTmypMiTM2QygGXXElkFWIH7GxfmlwmY2vwa+wmNsoYZmG2iusf1ZjJoA=="
-        }
-      }
+      "version": "3.2.1",
+      "resolved": "https://registry.npmjs.org/chart.js/-/chart.js-3.2.1.tgz",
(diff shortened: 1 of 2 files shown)


shaunanoordin (approved):
## PR Overview

This is a dependency update, but, er, I really can't confirm _on which project_ the dependency is used?

- Testing is... OK? I mean, on localhost + macOS + Chrome, I can compile and test basic functions (login, view home page, view projects page, pick a random project, classify & submit) quite fine.
- I just can't test the _specific functionality,_ because it requires figuring out a _specific project_ that uses said functionality

Look, I'm giving this an APPROVED! 👍 because I can't confirm that it breaks anything, but stay tuned for additional info that I dug up.

...

### Dev Notes

@eatyourgreens in your comments, you mentioned...
> I think chart.js is only used on project stats pages
> eg. https://pr-5943.pfe-preview.zooniverse.org/projects/stephenresearch/beluga-bits/stats?env=production

...but I don't think this is the case. That page maps to `app/pages/project/stat/stats.jsx`, which imports `./charts.jsx`, which uses dependencies like `react-chartist`, `moment`, etc but not chart.js

Here's my best trace for chart.js:

- `import (whatever) from 'chart.js'` is ONLY found in `app/features/modeling/line-plot/index.js` aka LinePlotModel.
- LinePlotModel is only used in `app/features/modelling/index.js` as part of a "modelSelector"
- That "modelSelector" is used in the `<CanvasViewer />` component
- The `<CanvasViewer />` component is used by `app/components/file-viewer/index.js` ...
- ...which is the "which viewer should I use?" logic of the `<FrameViewer />` component...
- ...which is part of the `<SubjectViewer />` component

So from what I can tell, the `chart.js` is used in the SubjectViewer, when the Subject in question is of the type `application` (`application/json` maybe?), AND the Subject has the `metadata['#models']['model']` value of "LINE_PLOT"

Basically chart.js is part of a SUPER NICHE Subject Viewer for a project with a VERY specific Subject setup. If I get a change to dive into the database, I'd love to figure out _which_ Project could possibly be using these line plots, but I suspect the answer is going to be like "something that was created in 1950 and retired at end of the reign of King George VI".

eatyourgreens:
How weird. I think that might be @tingard’s PhD project that you’re referring to. 

I thought `chart.js` was imported for the stats page. If it isn’t in use, we could consider removing it.

tingard:
My project uses Regl to render to canvases using webgl, so chartjs isn't needed for it.

I think I may have added the dependency though, as there was a thought about using it + canvas-viewer for generic "plot" subjects, which has ofc subsequently been implemented a different way for planet hunters.

tingard:
Okay so I can confirm that chart.js was introduced by me when developing the canvas-viewer, and can be removed

https://github.com/zooniverse/Panoptes-Front-End/pull/4270/files#diff-7ae45ad102eab3b6d7e7896acd08c427a9b25b346470d7bc6507b6481575d519

PR #5947, opened 2021-05-14
add owner display name to collection search

camallen:
Staging branch URL: https://pr-5947.pfe-preview.zooniverse.org

Fixes https://www.zooniverse.org/talk/17/1990968?comment=3246122

This PR disambiguates the collection names by providing the owner display name as well. 

Possible improvements to this PR: 
1. Since the data is already available for the component, i was thinking this could be a checkbox behaviour toggle as well as most of the time this will be redundant information (as folks normally don't collaborate on collections). 
2. Alternatively - this could be automated and only add the owner suffix when the logged in user and the collection owner differ (we have this information). 

I'd be happy to iterate on this solution, thoughts?

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/collections/collection-search.cjsx
+++ b/app/collections/collection-search.cjsx
@@ -10,6 +10,7 @@ module.exports = createReactClass
   propTypes:
     multi: PropTypes.bool.isRequired
     onChange: PropTypes.func
+    user: PropTypes.object
 
   componentDidMount: ->
     if @props.autoFocus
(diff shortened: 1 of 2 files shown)


eatyourgreens:
Looks good to me. I agree that it might be better if the owner's name isn't shown for my own collections. We could ask a few Gravity spy volunteers to try this branch and get their opinions?

eatyourgreens (commented):
Forgot to add my original comment as a review. After thinking about this a bit more, I'd be in favour of changing this so that the name is only added when the collection isn't yours. That way, the UX won't change for the majority of our users, but should be improved for cooperative collections on Gravity Spy.

camallen:
> Forgot to add my original comment as a review. After thinking about this a bit more, I'd be in favour of changing this so that the name is only added when the collection isn't yours. That way, the UX won't change for the majority of our users, but should be improved for cooperative collections on Gravity Spy.

ok - I think that's better UX for the requested use case 👍🏼. However how do i get the logged in user state in this component? Is it globally available by a context in PFE? Pointers welcome 🙏🏼

eatyourgreens:
Good question. I'll take a look. I think PFE uses global context to store the logged-in user.

eatyourgreens:
The user is passed in to parent components as a prop:
https://github.com/zooniverse/Panoptes-Front-End/blob/819875b286fbda8398d103b788766905cc070ab2/app/collections/show-list.cjsx#L48-L50

eatyourgreens (approved):
Looks good to me. I don't have any shared collections but the code changes look good.

camallen:
> Looks good to me. I don't have any shared collections but the code changes look good.

I tested on staging API and it works well

PR #5951, opened 2021-05-18
Publish commit_id.txt on deploy

eatyourgreens:
Add HEAD_COMMIT to the build environment (it's referenced in `package.json` in a couple of places.) Publish HEAD_COMMIT as `dist/commit_id.txt` and upload that to Azure with a cache lifetime of 60s.

Staging branch URL: https://pr-5951.pfe-preview.zooniverse.org/commit_id.txt

This is https://github.com/zooniverse/pandora/pull/189 adapted for PFE.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/.github/workflows/deploy_production.yml
+++ b/.github/workflows/deploy_production.yml
@@ -12,6 +12,8 @@ on:
 jobs:
   deploy_production:
     runs-on: ubuntu-latest
+    env:
+      HEAD_COMMIT: ${{ github.sha }}
     steps:
     - uses: actions/checkout@v2
     - uses: azure/login@v1
@@ -25,6 +27,9 @@ jobs:
         node-version: '14.x'
     - run: npm install
     - run: npm run _build-production
+
+    - name: Write commit_id.txt
+      run: echo ${HEAD_COMMIT} > ./dist/commit_id.txt
 
     - name: Upload to blob storage
       id: upload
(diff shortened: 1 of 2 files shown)


eatyourgreens:
The published commit hash is wrong for branches 😞 . Should be ok on master, though.

eatyourgreens:
I think I've fixed that by using `pull_request_target` for branches.

>This event runs in the context of the base of the pull request, rather than in the merge commit as the pull_request event does.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

camallen:
> I think I've fixed that by using `pull_request_target` for branches.
> 
> > This event runs in the context of the base of the pull request, rather than in the merge commit as the pull_request event does.
> 
> https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
> 
> EDIT: maybe not. The staging branch isn't deployed when I push to this PR. I'm going to try removing that commit.

pretty sure we don't want to change this as it allows forks to access secrets and the build context

> Warning: The pull_request_target event is granted a read/write repository token and can access secrets, even when it is triggered from a fork. Although the workflow runs in the context of the base of the pull request, you should make sure that you do not check out, build, or run untrusted code from the pull request with this event. Additionally, any caches share the same scope as the base branch, and to help prevent cache poisoning, you should not save the cache if there is a possibility that the cache contents were altered. For more information, see "Keeping your GitHub Actions and workflows secure: Preventing pwn requests" on the GitHub Security Lab website.

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

camallen:
this post looks like it might explain why the sha values differ on pull requests https://github.community/t/github-sha-not-the-same-as-the-triggering-commit/18286/2

camallen (approved):
LGTM - maybe remove the commit_id.txt from the deploy_branch workflow if the sha value isn't useful?

eatyourgreens:
Yeah, `pull_request` uses a merge commit. `pull_request_target` would run in the context of the original branch, but opens up a whole bunch of security concerns for PRs from forks.
https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

PR #5944, opened 2021-05-11
[Security] Upgrade css-loader to version 5

eatyourgreens:
Upgrade `css-loader` from 2.1 to 5.2, to fix a security vulnerability in older versions of `postcss`. I've also upgraded the related packages: `mini-css-extract-plugin`, `style-loader`, `stylus-loader` and `stylus`.

`css-loader@4` dropped support for absolute paths in URLs in CSS, which breaks a couple of our background images. I've updated those to use local file paths, which webpack will rewrite during the build. See https://github.com/webpack-contrib/css-loader/issues/1136

Staging branch URL: https://pr-5944.pfe-preview.zooniverse.org

This is a big version jump, so might break the webpack build.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #5953, opened 2021-05-19
Update to Webpack 5? Fix ssri?

shaunanoordin:
## PR Overview

Fixes https://github.com/zooniverse/how-to-zooniverse/issues/102 ?
Preview: https://pr-5953.pfe-preview.zooniverse.org/

This PR is an attempt to update the **ssri** sub-sub-sub dependency but is transforming into a large update from webpack 4 to **webpack 5.**

- ⚠️ Updating to Webpack 5!
- Since this fix is cascading into something much larger for PFE, I'm keeping track of why I'm performing every action, and what problems I'm facing. Please see dev notes for deets

### Dev Notes

- The problematic version of _ssri_ has been traced mostly to _copy-webpack-plugin 4.6.0_
- The recommended upgrade is to jump 4(!!!) major versions to _copy-webpack-plugin 8.1.1_
- _copy-webpack-plugin 8.1.1_ only works with Webpack 5 so now I'm installing _webpack 5.37.1,_
  - ⚠️ unfortunately, this is a major change
  - 📖 see migration documentation: https://webpack.js.org/migrate/5/
- (Fix build error) Webpack config changed: `node.fs: "empty"` -> `resolve.fallback.fs`
- (Fix build error) _html-webpack-plugin 3.2.0_ updated to _html-webpack-plugin 5.3.1_
- (Fix build error) _node-polyfill-webpack-plugin_ installed
  - ⚠️  This is one of the big ones: Webpack 5 stopped supporting Node.js polyfills natively, so one solution is to add a 3rd party polyfill. [(Documentaion)](https://github.com/webpack/changelog-v5#automatic-nodejs-polyfills-removed) [(Possible solutions)](https://stackoverflow.com/questions/64557638/how-to-polyfill-node-core-modules-in-webpack-5)
- (Fix build error) `webpack.EnvironmentPlugin` updated to new format, which has default values.
  - ❓ Default values are all empty strings, which uhhh... let's see if this causes problems. I could also set this to false.
- 🎉  HOORAY build issues all fixed!
- 😢 ~~...but now https://local.zooniverse.org:3735/ won't render.~~
  - ~~This maps to Level [5: Runtime Errors](https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises) in the migration guide.~~
- UPDATE: _node-polyfill-webpack-plugin_ was a false hope! This has now been removed.
  - This polyfill doesn't properly fix the missing requirements.
- Better solution is to add fallbacks/plugins to the webpack config for EACH missing requirement.
  - Added to the config: `plugins: { new webpack.ProvidePlugin({ process })` , `resolve.fallback = { fs, path, util }`
  - Added dependencies: _path-browserify, process, util_
- 🎉 HOORAY runtime issues all fixed! 

### Status

~~WIP, currently a nightmare, helpful insights welcome. Sanity checks also very welcome, in case someone can see a simpler way of doing this.~~

Woo hoo! This is now running on localhost, and I'm running extra tests.


shaunanoordin:
So the runtime error I'm getting when loading `localhost:3735` is... _Uncaught TypeError: Cannot read property 'custom' of undefined,_ and this traces to the _counterpart_ dependency.

I'd upgrade that dependency... but I just realised that _counterpart 0.18.6_ IS the latest version, despite being 3 years old at this point. Uh oh.

Note that the [documentation has a section for "What to do if 'process' is not defined"](https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises) but this just boils down to "don't use 'process'." Thanks, webpack 5!

I suppose I need to find a way to slam that `process` into the code. I'm looking at the [FEM lib-classifier](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/webpack.dev.js
) and I'm noticing this line...

```
plugins: [
  new webpack.ProvidePlugin({      process: 'process/browser',    }),
  ...
]
```
...which might be something to try. Gotta read up on ProvidePlugin!

EDIT: wait, this should be a draft PR. Gotta change that status.

shaunanoordin:
## PR Update

Runtime issues now all fixed! Dev Notes in the initial post updated.

Here's what changed:
- _node-polyfill-webpack-plugin_ was found to be a problem, not a fix. The polyfill didn't fill all the polys.
- Correct solution was to fill all missing Node.js polyfills on a _per polyfill/dependency_ basis.
  - Added to the config: plugins: { new webpack.ProvidePlugin({ process }) , resolve.fallback = { fs, path, util }
  - Added dependencies: path-browserify, process, util

### Status

Ready for review? I think so, I really need to get more eyes on this to test. If you're testing, please perform a clean NPM install to ensure everything works.

shaunanoordin:
@srallen I've requested a review from you because I believe you have the most experience in upgrading from Webpack 4 to 5, and I'm hoping you might spot any gotchas that I missed.

Note: @eatyourgreens made a good argument that if the SSRI issue is only a problem on the _build_ side, we can argue for a "don't fix" action. This might be a viable option if we thnk upgrading to Webpack 5 is a wee bit too dangerous for the creaky ol' PFE code.

At the moment I'm trying to figure out how to test the build for _production values._ I've been informed that I need to keep an eye on the `process.env` values not registering properly with Webpack 5? I'll look into this.

Additonal notes re: getting EnvironmentPlugin working?
- https://github.com/zooniverse/front-end-monorepo/pull/2160#discussion_r628355921

shaunanoordin:
### Ongoing Tests

package: `lib-classifier`

Issue 1: **can't go back to a previous step and delete mark.**
Replication:
- Open a workflow with Task 1: drawing with a Point tool, Task 2: text task. (e.g. https://localhost:8080/?env=staging&project=1873&workflow=3493)
- In Task 1, draw a single Point, Annotation X.
- Proceed to Task 2.
- Go Back to Task 1.
- Delete Annotation X
- Proceed to Task 2 again.
  - Expected: Task 2 should load
  - Actual: page crashes to white.
  - Error message:
    - _Uncaught Error: [mobx-state-tree] Failed to resolve reference 'ckoywfhfg00043h6792lv6jiz' to type '(Ellipse | Line | Mark | Point | Rectangle | TemporalPoint | TranscriptionLine)' (from node: /classifications/resources/ckoywfc8j00003h675ki5guw6/annotations/ckoywfc8n00013h67fxusda3f/value/0)_ ; _The above error occurred in the <PreviousMarks> component:_

❗ That said, I don't think this issue is caused by this branch, but is extant on `master` as of a09284a37e06666ac1285fa12179783fc2442c88 . I'm making a note of this in case I need to spin up a new issue. (TODO: check if this has been documented already.)

webpack.build.js
```diff
@@ -54,10 +56,18 @@ module.exports = {
     new MiniCssExtractPlugin({
       filename: '[name]-[contenthash].css'
     }),
+    new webpack.ProvidePlugin({  // Required for Webpack 5, since it removes Node.js polyfills
+      process: 'process/browser',
+    }),
```
eatyourgreens, line 61: Do you need  `process/browser` if `process.env` is defined by the EnvironmentPlugin? I interpreted this as saying that EnvironmentPlugin on its own should be fine to define environment variables for a build:
>Want to use environment variables with `process.env.VARIABLE`? You need to use the DefinePlugin or EnvironmentPlugin to define these variables in the configuration.

https://webpack.js.org/migrate/5/#run-a-single-build-and-follow-advises

shaunanoordin: This is required, yes. We need `new webpack.ProvidePlugin({ process: 'process/browser' })` to solve a runtime issue: _"process is not defined"_ error from `util`, which is used by `counterpart`

Side note though, after some additional testing, it seems that `resolve.fallback.util: require.resolve('util')` might not actually be needed! Going to try a clean reinstall to confirm that.

srallen: Yes, polyfilling the whole thing is safer since dependencies may be using it, but this can be tested and confirmed by just using the define plugin if you wish.

shaunanoordin: Update: `resolve.fallback.util: require.resolve('util')` (and the corresponding dependency in package.json) is ALSO required. Otherwise, we get a _build_ error with the message: _ERROR in ./node_modules/counterpart/index.js_ , _Module not found: Error: Can't resolve 'util' in ..._

My earlier testing _looked like_ it wasn't needed because I disabled the resolve.fallback, but didn't actually take out the dependency from package.json + do a clean install. I wasn't 100% sure that'd make a difference, but it does. Always test with a clean node_modules directory!

eatyourgreens: `npm ci` is your friend!

shaunanoordin:
Thanks for your comments, @srallen . I'll defer to your opinion since you've had more experience with Webpack 5, so it looks like the best path forward is to plan a full v5 migration instead of just trying to solve for the SSRI issue.

My only worry is with the increased scope of work is that I'll be hitting a moving target (what with dependabot constantly throwing new updates at us) but worse comes to worse I'll spin up a fresh new branch with a cleaner approach based on all lessons learnt.

### Status

**Plan A**
- Revert this PR back to draft mode (pls feel free to continue commenting though!)
- Spin up a new branch _off & targeting this PR_ to address...
  - Upgrading webpack-cli
  - Upgrading the loaders
  - + updating the file-loader configs(? will need to double check notes on this)
- Merge order: that branch -> this branch -> master 
- Figure out merge conflicts as we go along, natch.
- Then solve for optional upgrades, if/when we have dev time. (Hackday?)

**Plan B** (if the moving target moves too much)
- Repeat the full Webpack v5 migration in a new branch instead of trying to rebase multiple branches

**Plan C** (if they release webpack v6 while I'm doing the v5 migration)
- Shut down this laptop, lie on the floor, and cry.

srallen:
@shaunanoordin unfortunately I don't think we can rely on or guarantee a partial upgrade will continue to work, particularly because of some of the migration guide's warnings about some future deprecations. I appreciate you taking this on and do ask for help if you need it.

eatyourgreens:
#5957 upgraded the lockfile from[ version 1 to version 2](https://nitayneeman.com/posts/catching-up-with-package-lockfile-changes-in-npm-v7/), because I've upgraded to node 14 with npm 7. Hopefully that won't cause massive conflicts here.

eatyourgreens:
I updated a few loaders to latest in #5944 so hopefully that saves some work. `coffee-loader` is very, very out-of-date but `babel-loader` looks like it only needs a minor version bump.

srallen:
An upgraded package-lock file from an upgraded npm should be documented. The required npm version should be upgraded to version 7.

eatyourgreens:
The minimum npm version is still 6. There won't be any breaking changes until lockfile version 3. Otherwise, npm 7 would break CI for a lot of people, I guess. There's more detail about the changes in [the npm docs for `lockfileVersion`](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion). Commiting package changes should still work with node 14/npm 6, and CI with 6 should still run.

I've rebased this branch to latest master, to bring in yesterday's security update.

srallen:
Just because it still technically works shouldn't mean that it isn't documented. This app still technically works with Node 12, but we still upgraded to 14.

eatyourgreens:
The latest master says `ssri` is no longer installed, so the security fix might not be needed any more.
```
Panoptes-Front-End odonnell$ git st
## master...origin/master
Panoptes-Front-End odonnell$ npm why ssri
npm ERR! No dependencies found matching ssri
```

eatyourgreens:
Re. npm version, I'd suggest continuing to support 6 while we're using it for CI. We could try out node 15/npm 7 for builds, now that node 16 is out.

shaunanoordin:
Converting this to Draft while I spin up resources to get back to this.

shaunanoordin:
This PR has gotten pretty darn stale, so I'm closing this for now. We'll refer back to this PR if/when we come back to upgrade our Webpack.

PR #5962, opened 2021-06-01
ensure the workflow has a pimary language set

camallen:
Staging branch URL: https://pr-5962.pfe-preview.zooniverse.org

Ensure we have a primary language specified on workflow creation. Use the workflow to clone by default and fallback to the current locale.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/workflow-create-form.cjsx
+++ b/app/pages/lab/workflow-create-form.cjsx
@@ -7,7 +7,7 @@ WorkflowCreateForm = createReactClass
     onCancel: ->
     onSubmit: ->
     onSuccess: ->
-    projectID: ''
+    project: null
     workflowToClone: null
     workflowActiveStatus: false
 
@@ -26,7 +26,7 @@ WorkflowCreateForm = createReactClass
 
     newWorkflow =
       display_name: @refs.newDisplayName.value
-      primary_language: workflowToClone?.primary_language || counterpart.getLocale()
+      primary_language: workflowToClone?.primary_language || @props.project.primary_language
       steps: workflowToClone?.steps ? undefined
       tasks: workflowToClone?.tasks ? {}
       first_task: workflowToClone?.first_task ? ''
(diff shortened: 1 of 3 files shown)


app/pages/lab/workflow-create-form.cjsx
```diff
@@ -26,7 +26,7 @@ WorkflowCreateForm = createReactClass
 
     newWorkflow =
       display_name: @refs.newDisplayName.value
-      primary_language: workflowToClone?.primary_language
+      primary_language: workflowToClone?.primary_language || counterpart.getLocale()
```
eatyourgreens, line 29: This is going to re-introduce the bug where workflows are created in the user's current locale eg. German or Arabic, rather than the project's primary language.

eatyourgreens (changes requested):

app/pages/lab/workflow-create-form.cjsx
camallen: Confirming that behaviour - i've tested changing a project to a translated language (locale is set away from 'en' default) and then back to the lab - the workflow is then created with the translated locale. 

I think the best solution here is to align the default to the project's `primary_language` vs using counterpart - I'll update accordingly.

app/pages/lab/workflow.cjsx
```diff
@@ -105,7 +105,7 @@ EditWorkflowPage = createReactClass
       </h3>
       {if @state.workflowCreationInProgress
         <ModalFormDialog tag="div">
-          <WorkflowCreateForm onSubmit={@props.workflowActions.createWorkflowForProject} onCancel={@hideCreateWorkflow} onSuccess={@handleWorkflowCreation}  projectID={@props.project.id} workflowToClone={@props.workflow} workflowActiveStatus={not @props.project.live} />
+          <WorkflowCreateForm onSubmit={@props.workflowActions.createWorkflowForProject} onCancel={@hideCreateWorkflow} onSuccess={@handleWorkflowCreation}  project={@props.project} workflowToClone={@props.workflow} workflowActiveStatus={not @props.project.live} />
```
camallen, line 108: I wasn't sure if the csjx form of the jsx form is used - happy to remove one of these if it's not needed.

camallen: Aha - this is still needed. Reading into this more this is the workflow edit form and this links to the copy workflow button which launches the `WorkflowCreateForm` modal. so all this is needed.

eatyourgreens (approved):
Looks good to me. I can create new workflows, or clone existing workflows, on the staged branch.

PR #5965, opened 2021-06-03
Delete index.html and commit_id.txt after upload

eatyourgreens:
Upload `index.html` and `commit_id.txt` with a cache lifetime of 60s. Delete them after upload then upload the rest of the build with a cache lifetime of 60,4800s (1 week.)

This should fix any scenarios where `index.html` could be served out to the CDN with a lifetime of 1 week, before it's overwritten with a cache lifetime of 60s.

Staging branch URL: https://pr-5965.pfe-preview.zooniverse.org

Screenshot of the Zooniverse staging home page on this branch, showing the cache-control header in the Network panel, with a max-age of 60s.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/.github/workflows/deploy_branch.yml
+++ b/.github/workflows/deploy_branch.yml
@@ -35,17 +35,18 @@ jobs:
       with:
         creds: ${{ secrets.AZURE_STATIC_SITES }}
         inlineScript: |
-            az storage blob upload-batch \
-              --account-name zooniversestatic \
-              --content-cache-control 'public, immutable, max-age=604800' \
-              --destination '$web/preview.zooniverse.org/panoptes-front-end/pr-${{ github.event.number }}' \
-              --source ./dist
             az storage blob upload \
               --account-name zooniversestatic \
               --content-cache-control 'public, max-age=60' \
               --container-name '$web' \
               --name 'preview.zooniverse.org/panoptes-front-end/pr-${{ github.event.number }}/index.html' \
               --file ./dist/index.html
+            rm ./dist/index.html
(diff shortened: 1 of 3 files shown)


zwolf:
The CacheControl attributes on the files in blob storage are not wrong: 

blobprops

This only started happening yesterday. Overwriting index.html with the different CC header has worked since we started using Actions on every site that this has been adopted on, so there's still a mystery here. Even assuming a race condition situation where a request comes in after the first upload (week long header) but before the new one and the CDN caches the file for that long, why did a purge of both the FD cache and the nginx cache not cause the file to be reacquired from the origin, regardless of the headers? Please read the whole conversation in slack for more context.

All that said, this reordering and a quick rm isn't going to hurt anything and probably is better for higher traffic sites for the racy reason above.

eatyourgreens:
I'm baffled as to why purging the FrontDoor cache didn't work either, but I can't see any other way that files would be served from the origin with a max-age of 1 week.

zwolf:
I'm seeing a CORS error on the main site for the file zooniverse.org-status.txt. Is this related to getting chatops deployment working on PFE? Is it different somehow from commit_id.txt?

eatyourgreens:
`status.txt` is the popup we use when we're having technical problems. I'm not sure, off the top of my head, where that file is stored.

zwolf (approved):
This LGTM and at the minimum is a good excuse to see if another deploy makes AFD have any embarrassing realizations.

eatyourgreens:
Let’s see what happens!

PR #5987, opened 2021-07-12
Subject viewer: add isGroupSubject

eatyourgreens:
Add `isGroupSubject` to the subject viewer, which checks for undefined subject metadata and only displays group subjects if `subject.metadata` exists.

Staging branch URL: https://pr-5987.pfe-preview.zooniverse.org

Fixes #5986.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #5984, opened 2021-07-09
Talk: support special Subject Group metadata on Subject discussion pages

shaunanoordin:
## PR Overview

Context: for projects using the new "Subject Group" subject type (e.g. Galaxy Zoo: Weird and Wonderful, 2021), we want users on Talk to see + interact with a lot more of the (hitherto) hidden metadata.

This PR adds an additional data display for "Subject Groups" type of Subjects, in Talk's single Subject discussion pages.

- Talk's single Subject discussion pages will now have the `SubjectMetadata` sub-component.
  - This sub-component is responsible for surfacing certain kinds of data for certain kinds of Subject types.
  - For now, it only applies to the the "Subject Group" subject type. This sub-component will display all the _constituent Subjects_ of the group, in a list.
  - For every other Subject type, this component should render null. (i.e. no functional changes when discussing non-Subject Groups.)

Note: Subject Groups are identified by the presence of two metadata fields: `metadata.['#group_subject_ids']` and `metadata.['#subject_group_id']`. The former is a _string_ of the constituent _subjects' IDs,_ delineated by dash `-`. The latter is the a unique number ID.

### Status

WIP

- Current progress: POC is complete.
  image
- [x] need to add styling.
- [x] items in list should link to respective Subject pages on Talk.

Local testing URL: https://local.zooniverse.org:3735/projects/zookeeper/galaxy-zoo-weird-and-wonderful/talk/subjects/63395872?env=production

Staging test URL 1: https://pr-5984.pfe-preview.zooniverse.org/projects/zookeeper/galaxy-zoo-weird-and-wonderful/talk/subjects/63395872?env=production (Galaxy Zoo: Weird & Wonderful)
- ⚠️ NOTE: when clicking on the individual Subject links, the page might not load properly because the ?env=production isn't carried over. To test that more smoothly, try using the next URL instead...

Staging test URL 2: https://pr-5984.pfe-preview.zooniverse.org/projects/darkeshard/survos-testing-2020-subject-group-viewer/talk/subjects/136083 (Shaun's SURVOS Testing 2020)

PR #5983, opened 2021-07-08
Talk: turn off flipbook mode for group subjects

eatyourgreens:
When the subject viewer mounts, check subjects for `subject.metadata.#subject_group_id`. If that's present, turn off flipbook mode so that all the subject images are shown in a grid. Also set a new viewer class, `subject-viewer--layout-grid4`, when that metadata key is present.

Staging branch URL: https://pr-5983.pfe-preview.zooniverse.org/projects/zookeeper/galaxy-zoo-weird-and-wonderful/talk/subjects/63395872?env=production

Screenshot of a Galaxy Zoo: Weird and Wonderful subject, displayed as a grid in Talk.

I've styled images with `max-width: 100px` so they fill the available width without appearing too large.
https://pr-5983.pfe-preview.zooniverse.org/projects/darkeshard/survos-testing-2020-subject-group-viewer/talk/228/345
Screenshot of a test subject displayed with five images per row in a Talk discussion thread.

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #5989, opened 2021-07-13
Talk + SubjectViewer: make Subject Group's images link to component Subjects

shaunanoordin:
## PR Overview

Alternative to: #5984 
Follows #5983 (allows Subject Talk pages to display SubjectGroup images) and #5987

Context: for projects using the new "Subject Group" subject type (e.g. Galaxy Zoo: Weird and Wonderful, 2021), we want users on Talk to be able to go to the the individual Subjects that form the group.

This PR makes individual "frames" (i.e. component Subjects) of a SubjectGroup-type subject clickable, and link to their respective component subject's Talk pages.

- ⚠️ This adds new behaviour to the very widely used `subject-viewer.cjsx` component
  - While this behaviour is exclusive to only a specific type of subject (i.e. SubjectGroups), we need to confirm the functionality of the SubjectViewer component in multiple cases.
  - Known use cases: Subject Talk page, a project's Recents list, a user's Collections.
  - Cases that can be safely ignored: a project's Classify page (SubjectGroups are only classified on FEM)

### Status

WIP

- Functionality done
- [x] Tweaking styles. Need to communicate that these images (component Subjects) are clickable in a less obnoxious way.
- Check use cases for SubjectViewer + SubjectGroup subjects:
  - [x] Subject Talk page
  - [x] a project's Recents list
  - [x] a user's collections

### DIFF
--- a/app/components/frame-viewer.jsx
+++ b/app/components/frame-viewer.jsx
@@ -104,9 +104,13 @@ export default class FrameViewer extends React.Component {
           </FrameWrapper>
         </PanZoom>
       );
-    } else if (this.props.isGroupSubject && this.props.linkToSubject) {
+    } else if (this.props.isGroupSubject) {
       return (
-        <a href={this.props.linkToSubject} style={{ display: 'block', border: '1px solid cyan'}}>
+        <a
+          href={this.props.groupSubjectLink}
+          title={`Subject ${this.props.groupSubjectId}`}
+          style={{ display: 'block', border: '1px solid cyan'}}
+        >
           <FileViewer
             src={src}
             type={type}
@@ -143,8 +147,9 @@ FrameViewer.propTypes = {
   annotations: PropTypes.arrayOf(PropTypes.object),
   frame: PropTypes.number,
   frameWrapper: PropTypes.func,
+  groupSubjectId: PropTypes.string,
+  groupSubjectLink: PropTypes.string,
   isGroupSubject: PropTypes.bool,
-  linkToSubject: PropTypes.string,
   modification: PropTypes.object,
   onChange: PropTypes.func,
   onLoad: PropTypes.func,
@@ -163,8 +168,9 @@ FrameViewer.propTypes = {
 FrameViewer.defaultProps = {
   annotations: [],
   frame: 0,
+  groupSubjectId: undefined,
+  groupSubjectLink: undefined,  // If a Subject is a "Subject Group", each frame can link to its constituent Subject's Talk page.
   isGroupSubject: false,  // A "Subject Group" is a type of Subject that's composed of many (single image) Subjects
-  linkToSubject: undefined,  // If a Subject is a "Subject Group", each frame can link to its constituent Subject's Talk page.
   onChange: () => {},
   preferences: { },
   subject: {
--- a/app/components/frame-viewer.jsx
+++ b/app/components/frame-viewer.jsx
@@ -108,9 +108,9 @@ export default class FrameViewer extends React.Component {
     } else if (this.props.isGroupSubject) {
       return (
         <Link
-          to={this.props.groupSubjectLink}
+          className="linked-image"
           title={`Subject ${this.props.groupSubjectId}`}
-          style={{ display: 'block', border: '1px solid cyan'}}
+          to={this.props.groupSubjectLink}
         >
           <FileViewer
             src={src}
--- a/app/components/frame-viewer.jsx
+++ b/app/components/frame-viewer.jsx
@@ -1,5 +1,6 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import { Link } from 'react-router';
 import getSubjectLocation from '../lib/get-subject-location';
 import getSubjectLocations from '../lib/get-subject-locations';
 import PanZoom from './pan-zoom';
@@ -106,8 +107,8 @@ export default class FrameViewer extends React.Component {
       );
     } else if (this.props.isGroupSubject) {
       return (
-        <a
-          href={this.props.groupSubjectLink}
+        <Link
+          to={this.props.groupSubjectLink}
           title={`Subject ${this.props.groupSubjectId}`}
           style={{ display: 'block', border: '1px solid cyan'}}
         >
@@ -121,7 +122,7 @@ export default class FrameViewer extends React.Component {
             registerProgressObject={this.props.registerProgressObject}
             {...modellingProps}
           />
-        </a>
+        </Link>
       )
     } else {
       return (
--- a/app/components/frame-viewer.jsx
+++ b/app/components/frame-viewer.jsx
@@ -104,6 +104,21 @@ export default class FrameViewer extends React.Component {
           </FrameWrapper>
         </PanZoom>
       );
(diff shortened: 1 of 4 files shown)


app/components/frame-viewer.jsx
```diff
@@ -104,6 +104,21 @@ export default class FrameViewer extends React.Component {
           </FrameWrapper>
         </PanZoom>
       );
+    } else if (this.props.isGroupSubject && this.props.linkToSubject) {
+      return (
+        <a href={this.props.linkToSubject} style={{ display: 'block', border: '1px solid cyan'}}>
```
eatyourgreens, line 109: Using `a` here means the whole page gets reloaded, and PFE gets run again from scratch. This should use client-side routing instead.

eatyourgreens: This link also has no link text, so there’s no indication as to what clicking it does. Assistive technology might fall back to using the `src` of the wrapped image.

camallen:
Staging test URL 1: https://pr-5989.pfe-preview.zooniverse.org/projects/zookeeper/galaxy-zoo-weird-and-wonderful/talk/subjects/63395872?env=production (Galaxy Zoo: Weird & Wonderful)

    ⚠️ NOTE: when clicking on the individual Subject links, the page might not load properly because the ?env=production isn't carried over. To test that more smoothly, try using the next URL instead...

Staging test URL 2: https://pr-5989.pfe-preview.zooniverse.org/projects/darkeshard/survos-testing-2020-subject-group-viewer/talk/subjects/136083 (Shaun's SURVOS Testing 2020)

app/components/frame-viewer.jsx
eatyourgreens: In Firefox, you can right-click and inspect the accessibility API for the link. It has no label and I think is falling back to the target URL as the value.
Accessibility inspector for a subject image in Firefox, showing no accessible label.

shaunanoordin: 🆗 on adding a text label for accessibility. The solution I'm implementing is adding attributes like `title="Subject 1234"`.

❓ on changing `<a>` to `<Link>`, though. Sorry, I should have documented this here:
- a `<Link>` in the Subject Talk page that links to another Subject Talk page (within the same project) doesn't seem to work correctly (Chrome 91, FF 89)
- Clicking on the link once will change the URL on the browser's navigation bar, but the page won't refresh. 
  - (Clicking several times, on different links, results in different shenanigans based on the router history.)
- This behaviour was noted in the other branch as well: https://github.com/zooniverse/Panoptes-Front-End/pull/5984#issuecomment-878633203
- I've done some additional reading on this and the "correct solution" is to wrap the... Subject Talk page?... with `withRouter` but this was a rabbit hole. https://stackoverflow.com/questions/44823590/react-router-4-doesnt-update-ui-when-clicking-link

shaunanoordin: I'm going to try `withRouter` as a solution tomorrow, but if that doesn't work out, I'll need to fall back on using `<a>` for this stage of the project.

eatyourgreens: Oh that’s weird. The component is definitely set up to fetch a new subject when the ID in the URL changes.

eatyourgreens: The subject page is still using the deprecated `componentWillReceiveProps` so it might be worth upgrading it to `componentDidUpdate`.

shaunanoordin: Wait, no, the issue is with `SubjectPageContainer`'s `setSubject()` using `this.props`, when it might actually need `nextProps`. Classic React shenanigans!

shaunanoordin: HAH solved. `<a>` is now replaced with `<Link>` 👍

shaunanoordin:
## PR Update

- SubjectPageContainer has been tweaked (fixed, actually) so when a Subject Talk page is refreshed, it correctly fetches the ID of the _new Subject._ This fixes the issue where a Subject Talk Page couldn't properly link to another Subject Talk Page.
- Subject Viewer + Subject Group: `<a>` now replaced by `<Link>`
- Subject Viewer + Subject Group: linked images now have a `title="Subject 1234"` attribute for accessibility concerns.

TODO: styling for when a component subject has focus.

app/components/frame-viewer.jsx
eatyourgreens: D'oh! That mistake with the props might explain some other Talk bugs that have come up over the years. Could you change the function from `componentWillReceiveProps(nextProps)` to `componentDidUpdate(prevProps)`? That seems like a small change to me, and `componentWillReceiveProps` isn't part of React any more. Removed in React 15 somewhere, I think, because it's unpredictable if it fires multiple times during a render cycle.

eatyourgreens (approved):
Looks good to me. I'd replace `componentWillReceiveProps` with `componentDidUpdate`, so that it passes `this.props` instead of `nextProps`.

shaunanoordin:
Thanks for the review, @eatyourgreens !

## PR Update

- SubjectPageContainer's `componentWillReceiveProps` has been updated to `componentDidUpdate` 👍 
- Basic styles have been added to indicate that the images in a Subject Group grid can be clicked (a mild teal outline)

image

I'm just running some final scenario tests now, but otherwise, this update is ready to roll!

shaunanoordin:
## Update

> tabbing through the list of links, each of the images is also focusable

I'm noting that behaviour as well! You're right, Jim - I've traced it down to `app/components/file-viewer/image-viewer.jsx`, which renders an `` with `tabIndex={0}`. I won't touch this for now, but I can think of a solution where an optional `canFocus={false}` value is passed to FileViewer then passed to ImageViewer (other viewers too?) then change that tabindex to `tabIndex={(canFocus) ? 0 : undefined}`

Additional scenario tests look good 👌 

- **Subject Talk page:** 👍 
  - https://pr-5989.pfe-preview.zooniverse.org/projects/darkeshard/survos-testing-2020-subject-group-viewer/talk/subjects/136083
- **Classify pages:** SubjectGroup subject types will not be used on PFE, so the new additions to `subject-viewer` will not render on PFE's Classify pages.
- **Collections pages:** SubjectGroup subject types have that odd "tall" display, but this is extant behaviour.
  - _New behaviour:_ even when displayed as an item in a Collection, you can still click on individual images to open the Subject talk page.
  - This new behaviour does not interfere with standard interaction, however. e.g. to select/deselect an item in a Collection, you need to click on the "checkbox". Clicking on the checkbox does not also trigger a navigation to a Subject talk page, if the checkbox appears on top of an image. 👌  
  - Example: https://pr-5989.pfe-preview.zooniverse.org/projects/zookeeper/galaxy-zoo-weird-and-wonderful/collections/eatyourgreens/gz-weird-and-wonderful?env=production
- **Recents pages:** SubjectGroup subjects look wonky on the Recents pages, but this is extant behaviour.

Side note: I think I have an idea for a fix for Recents, by setting the offending `width: 100%` to `max-width: 100%` instead, but this needs testing in its own separate dev branch.

Screen Shot 2021-07-15 at 16 45 54

Screen Shot 2021-07-15 at 16 45 44

I'm going ahead and merging the PR. Thanks all for the eyes on this update!

PR #5992, opened 2021-07-14
Upgrade to React 17

eatyourgreens:
Staging branch URL: https://pr-5992.pfe-preview.zooniverse.org

Upgrade React to 17.0.2. Tests throw a lot of warnings about lifecycle methods that end support here (eg. `componentWillReceiveProps`.) Nothing seems to be broken, though.

React 17 [removes global event handlers](https://reactjs.org/blog/2020/08/10/react-v17-rc.html#changes-to-event-delegation) and fixes `event.stopPropagation`, which has been a source of bugs for us in the SVG and drawing tools (particularly on touch devices.)

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/components/pan-zoom.spec.js
```diff
@@ -39,11 +39,11 @@ describe('PanZoom', function () {
       let wrapper;
       let zoomInButton;
 
-      beforeEach(function () {
+      before(function () {
```
eatyourgreens, line 42: I think these tests were broken because `beforeEach` attached a new copy of the component to the document body for each test.

I'm not really sure how that worked before.

srallen (commented):
My concerns with the unofficial enzyme adapter remain here too and again, I don't see a reason to do this right at the moment. We still have some time to see if an official adapter gets released, so I'm going to recommend that this upgrade waits. We can, however, discuss how long it is reasonable to wait.

eatyourgreens:
I think the risk is pretty low here: tests pass and we're not going to write new tests for PFE, so we don't have to worry about future-proofing enzyme. As long as the site works, this should be a safe upgrade.

We might have to freeze PFE at React 17, which might mean freezing other dependencies too. Looking at the test warnings, React 18 might finally remove `componentWillReceiveProps` and `componentWillMount`, which are used throughout the lab and Talk code.

I am interested to see if the changes to event delegation, and `stopPropagation` finally being fixed, make the drawing and dragging components less buggy on touch devices.

shaunanoordin (approved):
## PR Review

This PR is a major dependency update to React 17.

Manual tests look good, and code read looks fine. The change from `coffee-script` to `coffeescript` was a bit of a "huh??" moment, but fine.

### Testing

Tested on localhost with macOS + Chrome 95

- I can view the home page
  - Home page show the correct details while logged in (user stats), and also while logged out (default)
- I can log in and log out
- I can view the projects page
- I can select a random project (I Fancy Cats) and view the project page
- I can make classifications and submit them (tested Question Tasks, Drawing Task)
- I can perform most standard/common Tasks on the Dev Classifier
  - Note: a number of the weirder Drawing Tasks (e.g. Anchored Ellipse and Triangle) appear to be borked, but that appears irrelevant to this PR.

### Status

LGTM 👍 Let's get this merged as soon as possible.

PR #6014, opened 2021-09-20
Add about pages for FEM redirect routes.

srallen:
Staging branch URL: https://pr-6014.pfe-preview.zooniverse.org

Also adds Peoples Contest Digital Archive project to the set of FEM projects with redirect. 

**This should not merge until zooniverse/operations#523, zooniverse/operations#528, and zooniverse/operations#529 are done and the about pages are included there for each project. Then this should merge shortly after we confirm the existing routes work with the NGINX http frontend and the about page routes are added**

# Required Manual Testing

- [ ] Does the non-logged in home page render correctly?
- [ ] Does the logged in home page render correctly?
- [ ] Does the projects page render correctly?
- [ ] Can you load project home pages?
- [ ] Can you load the classification page?
- [ ] Can you submit a classification?
- [ ] Does talk load correctly?
- [ ] Can you post a talk comment?

# Review Checklist

- [ ] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [ ] Does it work on mobile?
- [ ] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [ ] If the component is in coffeescript, is it converted to ES6? Is it free of eslint errors? Is the conversion its own commit?
- [ ] Are the tests passing locally and on Travis?

# Optional

- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you [resized and compressed](https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/image-optimization) any image you've added?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?
- [ ] Have you followed the [Springer guidelines for commit messages](https://github.com/springernature/frontend-playbook/blob/master/git/git.md#commit-messages)?

PR #6016, opened 2021-09-23
Add custom MonorepoRoute to the router

eatyourgreens:
`MonorepoRoute` is a [custom `Route` component](https://marmelab.com/blog/2016/09/20/custom-react-router-component.html) which takes a project path and redirects it to our NextJS project app. It handles both the home page and the Classify and About page URLs, which reduces router config down to one line of code for new projects:
```jsx
<MonorepoRoute path="/projects/nora-dot-eisner/planet-hunters-tess" />
```
Staging branch URL: https://pr-6016.pfe-preview.zooniverse.org/projects/nora-dot-eisner/planet-hunters-tess/talk?env=production
Try navigating to the Collect page, the About page, the Classify page and the project home page. The last three should redirect to NextJS. You can also try navigating to the Projects page, where HMS NHS should redirect too. 

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/MonorepoRoute.jsx
+++ b/app/MonorepoRoute.jsx
@@ -1,0 +1,50 @@
+import React from 'react';
+import { Route } from 'react-router';
+import { createRoutesFromReactChildren } from 'react-router/lib//RouteUtils';
+
+/*
+ Used to force a refresh on entering a route, causing it to be fetched from
+ the server - and allowing us to reverse proxy other apps to routes in PFE.
+
+ Usage: <Route path="path/to/location" component={RELOAD} />
+*/
+function RELOAD({ newUrl }) {
+  if (window.location.hostname === 'www.zooniverse.org') {
+    window.location.reload();
+  } else {
+    window.location = newUrl;
+  }
+
+  // we should never get here.
+  return (
+    <div>
+      Loading…
+    </div>
+  );
+}
+
+/*
+  This is a very useful article explaining how to write custom Route components for react-router v3.
+  https://marmelab.com/blog/2016/09/20/custom-react-router-component.html
+*/
+function MonorepoRoute() {
+  return (
+    <div>
+      &lt;MonorepoRoute&gt; elements are for configuration only and should not be
+      rendered
+    </div>
+  );
+}
+
+MonorepoRoute.createRouteFromReactElement = (element, parentRoute) => {
+  const { path } = element.props;
+
+  function Reload() {
+    return <RELOAD newUrl={`https://fe-project.zooniverse.org/${path}`} />;
+  }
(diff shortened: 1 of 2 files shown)


srallen (commented):
This should wait till after #6014 and then include the new project and the about page routes.

eatyourgreens:
Sounds good.

eatyourgreens:
I've added a project route for `about` and an entry for `kmc35/peoples-contest-digital-archive`.

eatyourgreens:
Rebased to include #6014.

lcjohnso:
@eatyourgreens -- This is helpful, and a great idea!  Sorry I missed that this PR was in the queue; I can adapt https://github.com/zooniverse/Panoptes-Front-End/pull/6017 accordingly.

Before this merges, I'm wondering if there's a way we might further abstract the creation of routing rules to streamline PFE-to-FEM migrations.  Would it be possible to create a file that lists all project slugs for which `MonorepoRoute`s should be created?  That would be even better than needing to add individual `MonorepoRoute` entries in the routing file, and might even be something we could automate?

Longer term I'm curious about how to streamline / automate the creation of these routing rules as much as possible.  See this RFC issue for some initial discussion: https://github.com/zooniverse/operations/issues/524.

What do you think?

eatyourgreens:
Dynamically generating routes from a list of project slugs might be a good hack day project. There’s a quote from the blog post I linked to above, about hacking routes:

> It turns out it's surprisingly difficult, and it's due to the nature of the `<Route>` component. It's not a real React component, it doesn't use the `render()` method, or any of the lifecycle callbacks.

I think it could be done, but I don’t know if I would hold up this PR while figuring out. Or at least, if I can’t add any working changes tomorrow then let’s review this as-is? I can try adding a loop to `MonorepoRoute` that iterates over a list of projects.

lcjohnso:
@eatyourgreens Totally fine to keep this scoped as-is.  If you figure out something clever, we can always change and add it at a later date.  But as this relates to the larger problem of PFE-to-FEM migration management, I wanted to raise the issue in case there was an immediate solution that came to mind.

Given this, tag me tomorrow to review when you decide you're happy.  I'll review and make the relevant changes to #6017 to follow-up, and I can take charge of deploying on Monday with both PRs merged (in sync with Zach re: https://github.com/zooniverse/static/pull/244).

eatyourgreens:
@lcjohnson I thought that the export from `router.cjsx` was a completely static structure, but that might be wrong. I'll push up a commit that interpolates an array of routes into the output. That would be a good starting point for automation, I think.

lcjohnso (approved):
LGTM!

PR #6023, opened 2021-10-04
Remove "Recent Publications" from PFE home page

shaunanoordin:
## PR Overview

This PR removes the "Recent Publications" from the home page.

- The "most recent Publication" blurb, and the logic used to fetch the data, has been removed.
- The "Recent Publications" renamed to "Publications". Only the link to "See all Publications" remains.

Context: our [About > Publications page](https://www.zooniverse.org/about/publications) is now hosted on FEM (not PFE!) and pulls data from the Contentful CMS. This means that the PFE method of managing a list of publications (i.e. `app/lib/publications.js` ) is no longer the "correct source of truth" for the data, and is outdated. We have two options: 1. hook up the PFE home page to pull the latest data from Contentful, or 2. remove the recent publication altogether, to avoid stale data appearing on the home page. This PR is option 2.

_Before: the "Recent Publications" shows a 2019 article._

image

_After: no more stale "Recent Publications"_
image

Staging Preview: https://pr-6023.pfe-preview.zooniverse.org

### Status

Ready for review


shaunanoordin:
I'm going ahead and merging this, since it's been sitting approved for a while. As mentioned, blurb looks fine, we can always revise later. Let's gooo!

PR #6030, opened 2021-10-08
Classrooms Workflow Selection Strategy - allow choosing default workflow

shaunanoordin:
## PR Overview

Context: Wildwatch Kenya Lab is introducing the idea of "non-custom" WildCam-type classrooms (see https://github.com/zooniverse/classroom/pull/251) . This means that, _instead of_ every Assignment having its own unique WF cloned from an "OG" workflow, we ask every Student to classify on the OG workflow.

This PR allows us to direct users to the _default workflow_ of a Classrooms-enabled Project by specifying `?workflow=default`

Expected behaviour of the Classifier:
- `?workflow=1234&classroom=1` - if a project allows for Classrooms (experimental tool: classrooms), and if a valid workflow is set in the URL, load that workflow.
- 🆕 `?workflow=default&classroom=1` - if a project allows for Classrooms, and if the string "default" is specified, load the default workflow (assuming it exists!)
- `?workflow=BEEPBOOPNONSENSE&classroom=1` - if a project allows for Classrooms, and if an invalid workflow is set in the URL, proceed to load NOTHING. ("Invalid workflow" applies to absurd workflow IDs, and workflow that are inaccessible to the user)

### Dev Notes

- As Jim noted, **default workflows are no longer used in FEM.** Basically, `project.default_workflow` will no longer be used the way they are in PFE.
- As for this PR, we're only leveraging the `project.default_workflow` value in this context to allow project owners to specify what the "OG workflow" they want to direct students towards.
- If we do retire `project.default_workflow` completely, then the onus of specifying what's the "OG workflow" for any project will fall to the [Classroom code](https://github.com/zooniverse/classroom) and the EduAPI. (This does remove power from the project owners though, so keep that in mind.)

### Testing

Example: https://local.zooniverse.org:3735/projects/wildcam/wildcam-darien/classify?env=production&workflow=8752&classroom=1

### Status

Code ready, needs testing.

### DIFF
--- a/app/pages/project/workflow-selection-classroom.jsx
+++ b/app/pages/project/workflow-selection-classroom.jsx
@@ -6,9 +6,22 @@ import { WorkflowSelection } from './workflow-selection';
 
 export class ClassroomWorkflowSelection extends WorkflowSelection {
   getSelectedWorkflow(props) {
-    const { actions, locale, location, preferences } = props;
+    const { actions, locale, location, project, preferences } = props;
+
+    // Normally, WildCam Classrooms are "custom programs" with a different
+    // (cloned) workflow for each Assignment. We expect a numerical Workflow ID.
     const workflowFromURL = this.sanitiseID(location.query.workflow);
     if (workflowFromURL) return actions.classifier.loadWorkflow(workflowFromURL, locale, preferences);
+
+    // Some WildCam Classrooms are now "non-custom programs", meaning every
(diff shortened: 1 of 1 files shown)


eatyourgreens:
Any idea why the branch deploy failed? It’d be good to test this out.

eatyourgreens:
Staging deploy should be up at https://pr-6030.pfe-preview.zooniverse.org

Issue #6051, opened 2021-11-19
FEM Lab: standardise all preview links 


shaunanoordin:
## "Front-End-Monorepo Migration" Decision Required

Back in March (#5919), Sarah modified the Project Builder's "Test this workflow" button to take people previewing Transcription Task workflows to https://fe-project.zooniverse.org/ (because Transcription Tasks only work on FEM) Now in November (#6049), I'm adding preview links for _every workflow_ that's built with the [FEM Lab,](https://github.com/zooniverse/Panoptes-Front-End/pull/6037) and these are going to https://frontend.preview.zooniverse.org/

**Question:** does anyone mind if I standardise ALL preview links for FEM projects to https://frontend.preview.zooniverse.org/ ?

Reason: frontend.preview seems to be our current go-to for previewing FEM projects
How I feel about this: strongly in favour of standardising to frontend.preview

### Additional info

Pros/cons of frontend.preview
- ➕ lets you preview the WHOLE zooniverse experience, including going from FEM Classify pages to PFE Talk pages
- ➕ most project owners are already using this as their FEM preview/testing link, as pointed out by @mrniaboc 
- ➖ is technically the 'canary version', and includes merged but not-yet-deployed-to-production changes

Pros/cons of fe-project.zooniverse
- ➕ has the production code
- ➖ doesn't have anything else, e.g. Talk links don't work

### Status

This is just the Github issue-ed version of a recent [Slack conversation.](https://zooniverse.slack.com/archives/C02HVGW0JAY/p1637247914058600) A decision is to be made by @lcjohnso

For the moment, for the sake of PR #(SHAUN, UPDATE THIS BIT ONCE YOU PUSH THE BRANCH THAT'S ADDRESSING #6049), the following _interim decision_ will be made:
- If WF has transcription task, the "Test This WF" preview link goes to `fe-project.zooniverse` (unchanged)
- (new) otherwise, if workflow is being edited in FEM Lab, the preview link goes to `frontend.preview`
- Otherwise, the preview link goes to `zooniverse.org`

I'm very much in favour of removing that first clause in FEM Lab _and PFE Lab._ (For consistency. Every project that uses the Transcription Task experimental tool _should also_ have the femLab experimental tool, in my opinion. That said, Cliff, this one's ultimately your call.)

PR #6053, opened 2021-11-24
FEM Lab: use FEM preview links, etc

### LINKED ISSUE #6049: FEM Lab: Nov 2021 misc TODO
## Misc ToDo

New behaviour:
- [ ] on FEM-Lab, `workflow.configuration.persist_annotations === undefined` should imply **true**.
  - On PFE-Lab, undefined implies false
  - See zooniverse/front-end-monorepo#2524
- [ ] NEW: on FEM-Lab, `workflow.configuration.hide_classification_summaries ===undefined` should also imply **true**.
- [ ] on FEM-lab, "View Project" and "Test this Workflow" should both point to the FEM preview URL.
  - ~~Note that we want to use `fe-project.zooniverse.org` points to the _production-release_ version of FEM...~~
  - ~~...whereas `frontend.preview.zooniverse.org` points to the _master_ version of FEM, which has not-yet-deployed changes.~~
  - EDIT: use `frontend.preview.zooniverse.org`
- [ ] FEM-Lab should have some visual indicators to separate it from PFE-Lab

shaunanoordin:
## PR Overview

Closes #6049

This PR updates the FEM Lab with...
- [x] preview links that point to `frontend.preview.zooniverse.org`
- [x] `persist_annotations` or `hide_classification_summaries` checkboxes that are ticked by default (i.e. undefined implies true)

### Status

WIP

Defers #6051 to another PR.


shaunanoordin:
## PR Update

This PR is now ready for review

### Testing

Setup:
- Test URL: `https://pr-6053.pfe-preview.zooniverse.org/lab/(ANY-PROJECT)/workflows/(ANY-WORKFLOW)` Project Builder -> Workflow Editor page
- You can force the FEM or PFE mode of the Project Builder by using `?pfeLab=true` or `femLab=true`
  - e.g. https://pr-6053.pfe-preview.zooniverse.org/lab/1873/workflows/3532?femLab=true

What you're testing for:
- In FEM Lab mode...
  - The "View Project" and "Test this Workflow" links should link to `frontend.preview.zooniverse.org/...`
  - "Persist Annotations" and "Hide classification summaries" checkboxes should default to ticked (assuming you've never changed these settings for this workflow before) AND disabled (i.e. you can't change it)
- In PFE mode...
  - The "View Project" and "Test this Workflow" links should link to `pr-6053.pfe-preview.zooniverse.org/...`

app/pages/lab/project.cjsx
```diff
@@ -53,13 +54,25 @@ EditProjectPage = createReactClass
     linkParams =
       projectID: @props.project.id
 
+    thisProjectUsesFEM = isThisProjectUsingFEMLab @props.project, @props.location
+
+    projectLink = "/projects/#{@props.project.slug}"
+    if thisProjectUsesFEM then projectLink = "#{FEM_LAB_PREVIEW_HOST}#{projectLink}"
```
eatyourgreens, line 60: This is going to need to add `?env=staging` on staging.
eg. here the project preview link gives a 404 error.
https://pr-6053.pfe-preview.zooniverse.org/lab/908?femLab=true

app/pages/lab-fem/workflow.cjsx
```diff
@@ -47,9 +48,14 @@ EditWorkflowPage = createReactClass
   workflowLink: ->
     [owner, name] = @props.project.slug.split('/')
     usingTranscriptionTask = Object.keys(@props.workflow.tasks).some((taskKey) => @props.workflow.tasks[taskKey].type is 'transcription')
-    if @canUseTask(@props.project, "transcription-task") and usingTranscriptionTask
+    if isThisProjectUsingFEMLab(@props.project, @props.location)
+      return "#{FEM_LAB_PREVIEW_HOST}/projects/#{owner}/#{name}/classify/workflow/#{@props.workflow.id}"
```
eatyourgreens, line 52: Same comment about adding `?env=staging` for staging projects here too.

eatyourgreens (commented):
A couple of comments about preview links being broken on the staging API. Otherwise, this looks good for production projects.

app/pages/lab/project.cjsx
eatyourgreens: I guess this is complicated by the fact that `*.pfe-preview.zooniverse.org` uses the staging API, but `frontend.preview.zooniverse.org` uses the production API.

eatyourgreens (approved):
Looks good to me. Just a couple of comments about how this works with the staging API, and I'm not sure about `location.query`. It works, but that's not a standard property of the Location interface.
https://developer.mozilla.org/en-US/docs/Web/API/Location

shaunanoordin:
Thanks Jim!

- noted that the preview links on the Workflow Editor don't work on PFE staging.
  - I initially opted not use Sarah's `env = process.env.NODE_ENV` pattern because I felt it made the links on production look a bit worse (i.e. explicitly stating ?env=production), but you're right that this shouldn't break the staging links. Gonna get in a patch for this. 🛠️ 
- re: `location`, I was passing in the 'location' data from the react-router since it was there. I can change this to e.g. `(new URL(window.location)).searchParams.get('femLab')` if anyone prefers, I don't feel strongly either way. 👌

eatyourgreens:
`NODE_ENV` controls the webpack build mode, so 'staging' isn't really a valid value. Webpack build modes are `none | development | production`.
https://webpack.js.org/configuration/mode/

In the monorepo, we've used `APP_ENV` and `PANOPTES_ENV` to indicate the deployment environment and the API but that would probably be a big change for PFE, when we want to build a production build that uses the staging API.

PR #6056, opened 2021-11-26
Fix: allow deletion of Tasks whether or not they exist in Steps

### LINKED ISSUE #6055: Project Builder: can't delete Tasks if Project has Transcription Tasks
## Project Builder - Functionality Bug

Original report: https://zooniverse.freshdesk.com/a/tickets/9726
Affects: Project Builder -> Workflow Editor

Simple description: if a Project has the "Transcription Tasks" experimental tool, then you can't delete SOME tasks.

Full description:
- If a Project has the "Transcription Tasks" experimental tool enabled, then the Workflow Editor's "add Tasks" and "delete Tasks" logic will modify **workflow.steps** (this is new) in addition to workflow.tasks (this is the usual)
  - You can only delete Tasks if they're registered in _both_ workflow.steps and workflow.tasks
  - Tasks created _after the experimental tool is enabled_ will be registered in both workflow.steps and workflow.tasks
  - Tasks created _before the experimental tool is enabled_ will only be registered in workflow.tasks
- i.e. once you've enabled Transcription Tasks, then you can't delete any Task created before you enabled the experimental tool.
  - The WF doesn't need to have Transcription Tasks, only have it enabled in the admin.

❗ Larger scope problem:
- Project Builder doesn't really handle **workflow.steps** really well.
- The code (as of 39d6257 25 Nov 2021) has several hacks that add workflow.steps logic, but only in a semi-piecemeal manner, and it isn't backwards compatible.
- ☠️ In fact, I have no idea how these Workflows will work on the FEM Classifier because they will have workflow.steps that will only contain SOME of their Tasks.

### Testing

1. In the Project Builder, open a project that _does not have_ Transcription Tasks enabled.
2. Create a new workflow
3. Create a new Task, T0.
4. Check the data structure sent to/from Panoptes. The WF resource should look like this:
    ```
    workflow:
      steps: []
      tasks:
        T0:
          type: "text",
          instruction: "Who is your daddy?"
    ```
5. Go to Admin -> Set Project Status -> your project. Enable the "Transcription Tasks" experimental tool.
6. Go back to the Project Builder and edit the workflow.
7. Create a new Task, T1.
8. Check the data structure sent to/from Panoptes. The WF resource should look like this:
    ```
    workflow:
      steps: [[ "S0", { taskKeys: ["T1"] } ]]
      tasks:
        T0:
          instruction: "Who is your daddy?"
          type: "text"
        T1:
          instruction: "And what does he do?"
          type: "text"
    ```
9. Observe that `workflow.steps` contains data for T1 but not T0.
10. In the Workflow Editor, attempt to delete Task T0
  - Expected: task T0 should be deleted
  - ❌ Actual: deletion attempt _fails silently_
11. Attempt to delete Task T1. This will work fine.

### Status

I'm marking this is a pretty serious Project Builder bug for any project with Transcription Tasks.

A solution is in the works.

Tagging @mrniaboc and @snblickhan since this involves Transcription projects.

shaunanoordin:
## PR Overview

Fixes #6055 
Affects: Project Builder -> Workflow Editor

This PR standardises the "Delete Task" action for projects that have "Steps".

The problem:
- If you have a "Transcription Task-enabled Workflow" (i.e. _the project has the experimental tool_ ), then the "Delete Task" action uses `deleteStepAndTask()`. That function assumes/REQUIRES the Task (workflow.tasks) to also exist in a Step (workflow.steps)
- If you created a Workflow BEFORE enabling the experimental tool, then every Task in that WF _won't be registered in steps_
- It's backwards-compatibility issue, basically

The solution:
- Modify `deleteStepAndTask()` to have a branch: one for when the Task exists in a Step, and one for when it doesn't
- Add comments to explain how the logic actually works

The caveats:
- The WF will STILL probably bork in the FEM Classifier, since the older pre-"TranscriptionTask-enabled" Task won't be registered in the workflow.steps. So you might end up in a scenario where a WF's tasks are visible on the Project Builder, but "disappear" on the FEM Classifier. 
- There's a separate issue where if you add a Transcription Task, it adds `workflow.configuration.subject_viewer = 'multiFrame'`, but there's nothing to remove this when the Transcription Task is deleted.

### Testing

Please see #6055 for the expected behaviour. Basically:

- You need to test deleting a Task that was created _before_ Transcription Tasks were enabled.
- You need to test deleting a Task that was created _after_ Transcription Tasks were enabled.
- You need to test deleting a Transcription Task (i.e. test that, when you delete a Task in a Step, then all Tasks in the Step gets delete)

Testing URL: https://pr-6056.pfe-preview.zooniverse.org/lab/1873/workflows/3545
- For convenience, I've set up a project with all these Tasks set up in advance. Feel free to delete them!
- You need to look at the Workflow resource before/after each delete to confirm everything's looking correct

### Status

Ready for review. Priority: **medium,** since this affects at least one active project (Every Name Counts)


eatyourgreens:
Code changes look good. I was confused to see changes  in `/lab` and `/lab-fem`. I expected the transcription task to only be available in `/lab-fem`.

eatyourgreens (approved):
Changes look good to me. If you look in `handleTaskDelete`, there's some additional logic for shortcuts that isn't present in `deleteStepAndTask`, so deleting a shortcut may not work in a workflow that has enabled the transcription task.

TBH, I think `handleTaskDelete` should check whether the task is included in a step, rather than checking that the transcription task is enabled for the entire project. That would be much safer, and not turn on this feature for all workflows.

shaunanoordin:
Thanks Jim!

Good point with the `handleTaskDelete` - I've opened #6061 as an ongoing TODO list for cleaning up the Transcription Task exceptions. This probably counts as extra credit for FEM-Lab work, to be honest.

I'm going ahead and merging this first so that the ENC folks can continue their work.

PR #6060, opened 2021-11-26
Add a reusable azure_deploy workflow

eatyourgreens:
Move the shared Azure deploy code into a [reusable workflow](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#creating-a-reusable-workflow). Call that workflow from the individual deployment workflows.

~~I wasn't sure if the shared workflow should be responsible for sending Slack notifications. Happy to make those reusable too. I'd have to look some more into how repo secrets, like the Slack webhook URL, are shared with called workflows.~~ 

~~Slack notifications for all three deploy workflows are handled by `slack_deploy_status.yml`.~~

~~I gave up on running the Slack notification from a called workflow, as it kept giving 'Job is not found' errors when calculating the job run time.~~ Figured out how job names work in called workflows. It's `Calling job name / Called job`.

Staging branch URL: https://pr-6060.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


eatyourgreens:
@zwolf the only bug I'm seeing with this is that `action-slack` reports the run time for the Slack notification (3s), rather than the Azure job time. I haven't figured out a way to get the time for both jobs. As per our Slack conversation, passing `job_name` to `action-slack` seems like it should work but I haven't been able to figure out the name that it expects.

Workflow run status showing 2 min 6 sec to deploy PFE then 3 sec to send the Slack notification.

eatyourgreens:
Update: the build time is reported correctly if you pass `job_name: Deploy staging branch / deploy`. I figured out the job name by going into the Actions tab and looking at the job names for the workflow runs on this PR.

eatyourgreens:
I think this means that the slack notification code can now be rolled up into a reusable job, but I might leave this now that the PR is actually working.

eatyourgreens:
@zwolf I think this is ready for review now that I've figured out how to pass the job names around.

.github/workflows/azure_deploy.yml
```diff
@@ -0,0 +1,155 @@
...
+          type: string
+      secrets:
+        creds:
+          required: true
+        slack_webhook_url:
+          required: true
+      outputs:
+        status:
+          description: 'Deploy status'
+          value: ${{ jobs.deploy.outputs.status }}
```
eatyourgreens, line 44: I think this is left over from a time before I found out that `job_id.result` exists, to get the status of a completed job. It's not used anywhere.

.github/workflows/azure_deploy.yml
```diff
@@ -0,0 +1,151 @@
...
+              --name '${{ inputs.target }}/commit_id.txt' \
+              --file '${{ inputs.source }}/commit_id.txt'
+            rm ${{ inputs.source }}/commit_id.txt
+            az storage blob upload-batch \
+              --account-name zooniversestatic \
+              --content-cache-control 'public, immutable, max-age=604800' \
+              --destination '$web/${{ inputs.target }}' \
+              --source ${{ inputs.source }}
+
+# Azure logout
```
zwolf, line 97: Logging out is only required on self hosted runners, per https://github.com/Azure/login#az-logout-and-security-hardening

zwolf (approved):
LGTM. I'm working on the Tove version of same now and I hope to be able to break out the individual operations (docker build/push, k8s deploy, slack notification, etc) into logical blocks that can have their own files. Putting these together like legos will let us centrally manage that code, calling individual workflows from a central repo.

That's the new goal, anyway, now that the new reusable wfs seem to be working.

.github/workflows/azure_deploy.yml
eatyourgreens: 👍🏻

eatyourgreens:
Chaining jobs together via `inputs/outputs` and `if` looks nifty.

zwolf:
I'm thinking that, say, a "Deploy Staging" job, such as the new one you've got here, could call out to some set of other workflows in a new, separate repo. Here, that might be "Boilerplate" (checkout, cache deps, node build, commit id), then "Deploy" (upload to blob storage), and finally "Slack notification". In that case, we could share the boilerplate workflow between every app that builds with node, the "Deploy" with every static site, and the Slack notification workflow between nearly everything, if I can stick enough conditionals in there.

Brainstorming for now but this is the direction I wanted to go back when I thought it would mean writing my own actions--reusable workflows seems to be the new lighter lift option.

PR #6067, opened 2021-12-03
Throw a meaningful error when workflow.first_task is missing

eatyourgreens:
Throw `First task has not been set for workflow`, when `workflow.first_task` is undefined or empty.

I Fancy Cats showing 'First task has not been set for workflow' instead of the Classify page.

There's a project that's broken because `workflow.first_task` is an empty string. This PR would show them a meaningful error, although it's not clear how they created a workflow with no first task.
https://www.zooniverse.org/projects/matildagrahn/korrekturlas-svenska-skillingtryck/classify?reload=0&workflow=20341

See also this Talk post:
https://www.zooniverse.org/talk/17/2244862?comment=3681061&page=1

TODO: is there a bug in the project builder that allows you to create invalid workflows?

Staging branch URL: https://pr-6067.pfe-preview.zooniverse.org/projects/matildagrahn/korrekturlas-svenska-skillingtryck/classify?reload=0&workflow=20341&env=production

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/classifier.jsx
+++ b/app/classifier/classifier.jsx
@@ -315,7 +315,7 @@ class Classifier extends React.Component {
   }
 
   render() {
-    const { actions, goldStandardMode, intervention, user } = this.props;
+    const { actions, goldStandardMode, intervention, user, workflow } = this.props;
     const { showIntervention, showSummary, workflowHistory } = this.state;
     const currentTaskKey = workflowHistory.length > 0 ? workflowHistory[workflowHistory.length - 1] : null;
     const taskAreaVariant = goldStandardMode ? 'goldStandardMode' : 'default';
@@ -325,6 +325,17 @@ class Classifier extends React.Component {
       'large-image': largeFormatImage,
       [this.props.className]: !!this.props.className
     });
+
+    /*
+      Throw an error if the workflow has tasks but a first task isn't set,
+      otherwise the classifier will crash deeper down the component tree.
+    */
+    if (!workflow.first_task) {
+      const hasTasks = Object.keys(workflow.tasks).length
+      if (hasTasks) {
+        throw new Error('First task has not been set for workflow.')
+      }
+    }
 
     let currentClassification;
     let currentTask;
--- a/app/classifier/classifier.spec.js
+++ b/app/classifier/classifier.spec.js
@@ -53,6 +53,7 @@ const workflow = mockPanoptesResource('workflow', {
   configuration: {
     hide_classification_summaries: false
   },
+  first_task: 'T0',
   tasks: {
     T0: {},
     T1: {}


shaunanoordin:
Unfortunately (for testing purposes), Workflow 20341 has now been fixed with a proper `first_task`. Give me a minute to spin up a WF with no first_task. (note: I'm only able to do this via console commands, so I'm still not sure how a WF without a `first_task` would be created on the Project Builder)

Also, I'm unsure if we can rely on a project having a `T0` task.

shaunanoordin:
I've created two test workflows:

- Staging Workflow 3549, which has task **T0** but no `first_task`
    https://pr-6067.pfe-preview.zooniverse.org/projects/darkeshard/test-project-2021/classify?reload=0&workflow=3549
- Staging Workflow 3551, which has task **T1** but no `first_task`
    https://pr-6067.pfe-preview.zooniverse.org/projects/darkeshard/test-project-2021/classify?reload=0&workflow=3549

As of commit 1045950, this PR's update allows 3549 to work, but 3551 still crashes, of course.

I'd normally feel strongly having a more resilient fallback for first_task, but since I'm unsure _how_ the empty first_task issue even cropped up on Project Builder, I'm OK with assuming T0 is the default for now. 🤷

(If we do want to be more resilient, I assume something like `if (!first_task) { first_task = Object.keys(workflow.tasks)[0] }` would work... right until we have a WF with 0 Tasks, which crashes anyway.)

mcbouslog:
> Unfortunately (for testing purposes), Workflow 20341 has now been fixed with a proper `first_task`. Give me a minute to spin up a WF with no first_task. (note: I'm only able to do this via console commands, so I'm still not sure how a WF without a `first_task` would be created on the Project Builder)
> 
> Also, I'm unsure if we can rely on a project having a `T0` task.

Sorry about that! I fixed the workflow per a contact ticket before realizing it was linked here.

mcbouslog:
I can't figure out how `first_task` became an empty string and couldn't be updated. I can't remember if before I fixed it in the originally broken workflow if the first task select input showed T0 was the first task, if the select was disabled, or something else? I see in https://pr-6067.pfe-preview.zooniverse.org/lab/1873/workflows/3549 if you select the task it doesn't update `first_task`, but I'm not sure why, maybe an AutoSave issue.

In this line - https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/workflow-tasks-editor.cjsx#L280:
`@props.workflow.update first_task: @props.workflow.tasks[taskKey].next ? Object.keys(@props.workflow.tasks)[0]`

I'm rusty on my coffeesscript, but should it be:
`@props.workflow.tasks[taskKey].next ? @props.workflow.tasks[taskKey].next : Object.keys(@props.workflow.tasks)[0]`

Prob not even be related to this bug.

I don't know, I don't have a strong opinion on what to do, but we could create an issue and revisit if occurs again? This PR will likely fix most of similar issues, but like Shaun notes if there isn't a T0 there'd still be an issue and with uncertainty around why maybe ok to punt until further occurrences?

mcbouslog (approved):
This fixes the issue noted. How the issue arose is unclear, and a similar issue could arise if there's no tasks or no `T0`, but this will likely solve the vast majority of related issues.

eatyourgreens:
@shaunanoordin @mcbouslog You're right that we can't really guess which task key should be first for a give workflow. I'm altering this PR to show a more meaningful error instead.

PR #6127, opened 2022-04-05
Experiment: Free Rotation for Subject Viewer

shaunanoordin:
## PR Overview - EXPERIMENTAL

Affects: Subject Viewer

This PR adds an experimental feature to the Subject Viewer, which allows the Subject image to be rotated freely (instead of being locked into 90-degree rotations).

image

_In the image above, there's a new vertical slider/range input (below the old "Rotate" button and above the "Reset" button) on the image toolbar, which allows users to set an arbitrary rotation for the Subject images._

Preview: https://pr-6127.pfe-preview.zooniverse.org/projects/ellenjj/rosetta-zoo/classify?env=production

### Status

Experimental.
- To be properly reviewed on a hackday.
- Feature needs to be hidden behind an experimental flag

### DIFF
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -346,7 +346,7 @@ class PanZoom extends Component {
               <button title="rotate" className={'rotate fa fa-repeat'} onClick={this.rotateClockwise} />
             </div>
             {canUseFreeRotation && (
-              <div class="experimental-free-rotation">
+              <div className="experimental-free-rotation">
                 <input type="range" min={0} max={360} value={this.state.rotation % 360} onChange={this.rotateFreely} />
               </div>
             )}
--- a/app/components/pan-zoom.jsx
+++ b/app/components/pan-zoom.jsx
@@ -346,8 +346,8 @@ class PanZoom extends Component {
               <button title="rotate" className={'rotate fa fa-repeat'} onClick={this.rotateClockwise} />
             </div>
             {canUseFreeRotation && (
-              <div style={{ display: 'flex', height: '80px' }}>
-                <input style={{ width: '60px', transform: 'rotate(270deg)' }} type="range" min={0} max={360} value={this.state.rotation % 360} onChange={this.rotateFreely} />
+              <div class="experimental-free-rotation">
+                <input type="range" min={0} max={360} value={this.state.rotation % 360} onChange={this.rotateFreely} />
               </div>
             )}
             <div>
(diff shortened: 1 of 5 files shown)


shaunanoordin:
Some important questions were asked about the free rotation feature:

Q1: Will the rotation affect how the any drawing marks (ellipses, rectangles, etc) are recorded?

A1: The _viewer rotation_ does **not** affect annotation data. This is the same principle how the viewer pan and zoom doesn't affect annotation data.

(You can see this by drawing a mark, and then zooming/rotating/panning the image, and the mark will be on the correct position relative to the zoomed/rotated/panned Subject image)

Q2: Do we record any rotation that was made?

A2: The _viewer rotation_ is **not** recorded in the annotation data, alas. I think(?) this is a more complex question, since our PFE classifications currently don't collect the zoom/pan values of the viewer either.

shaunanoordin:
Follow up to [Q1:](https://github.com/zooniverse/Panoptes-Front-End/pull/6127#issuecomment-1090255946)

- if you want to confirm that the rotation doesn't affect the annotation values, you can try placing marks on a Subject Image that's a 1000x1000 pixel grid, which is super easy to visually verify.
    image 
- Go to https://pr-6127.pfe-preview.zooniverse.org/projects/darkeshard/test-project-2022 and choose the "Drawing WF" workflow
- Place a point/rectangle/etc at (say) x=200 y=500 on the grid image.
- Rotate the image to an arbitrary angle
- Place a point/rectangle/etc at (say) x=300 y=400
- Rotate, place, repeat, etc
- Observe that the submitted Classification has marks placed at the expected x-y values.

app/components/pan-zoom.jsx
```diff
@@ -261,14 +261,28 @@ class PanZoom extends Component {
...
+    const newRotation = (this.state.rotation + 90) % 360;
     this.setState({
       rotation: newRotation,
       transform: `rotate(${newRotation} ${this.props.frameDimensions.width / 2} ${this.props.frameDimensions.height / 2})`
     });
   }
 
+  /*
+  TODO: hide this behind an experimental flag
+   */
```
goplayoutside3, line 273:
```suggestion
```

Non-blocking suggestion, but I think this comment can be deleted since pan-zoom's render method looks for `subjectViewer-freeRotation` in `experimental_tools` right?

goplayoutside3 (approved):
Looks good! I'm able to use the free rotation tool in all major browsers and mobile Safari (works best when phone is in landscape). Confirmed that a drawing annotation in Test Project 2022 is not affected by rotating the subject freely. Tests are updated and passing.

Non-blocking note: The styling of the free rotation slider is different between desktop Safari and desktop Chrome. Functionality is the same though.

app/components/pan-zoom.jsx
shaunanoordin: Ah, good catch! 👍

PR #6160, opened 2022-07-01
Add Turkish (tr) Locale

lcjohnso:
Adds Turkish (tr) locale file and platform-level translated content.

Staging branch URL: https://pr-6160.pfe-preview.zooniverse.org

Project-specific test URLs: TBD

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?
- [ ] Are the tests passing via GitHub Actions?

### DIFF
--- a/app/locales/tr.js
+++ b/app/locales/tr.js
@@ -1,0 +1,471 @@
+export default {
+  loading: '(Yükleniyor)',
+  aboutPages: {
+    missingContent: {
+      education: 'Bu projede henüz eğitim kaynağı yok.',
+      faq: 'Bu projede henüz sıkça sorulan sorular yok.',
+      research: 'Bu projenin henüz bir bilimsel çalışması yok.',
+      results: 'Bu projenin henüz rapor edilecek bir sonucu yok.',
+      team: 'Bu projede ekip bilgisi yok.'
+    }
+  },
+  projectRoles: {
+    title: 'Proje Ekibi',
+    owner: 'Proje Yürütücüsü',
+    collaborator: 'Proje Ortakları',
+    translator: 'Çevirmen',
+    scientist: 'Araştırmacı',
+    moderator: 'Moderatör',
+    tester: 'Test edici',
+    expert: 'Uzman',
+    museum: 'Müze'
+  },
+  classifier: {
+    back: 'Geri dön',
+    backButtonWarning: 'Geri gitmek, mevcut çalışmanızı silecektir.',
+    close: 'Kapat',
+    continue: 'Devam et',
+    detailsSubTaskFormSubmitButton: 'Tamam',
+    done: 'Bitti',
+    doneAndTalk: 'Bitti & Sohbet Et',
+    dontShowMinicourse: 'Do not show mini-course in the future',
+    letsGo: 'Hadi Başla!',
+    next: 'Sonraki',
+    optOut: 'Vazgeç',
+    taskTabs: {
+      taskTab: 'Transkripsiyon  işlemi',
+      tutorialTab: 'Tutorial'
+    },
+    recents: 'En son çalışmalarınız',
+    talk: 'Sohbet',
+    taskHelpButton: 'Bu işlemle ilgili yardıma mı ihtiyacınız var?',
+    miniCourseButton: 'Restart the project mini-course',
+    workflowAssignmentDialog: {
+      promotionMessage: 'Tebrikler! Bir sonraki iş akışının kilidini açtınız. Eğer istersenizö bu iş akışında kalmayı tercih edebilirsiniz.',
+      acceptButton: 'Beni bir sonraki seviyeye ilerlet!',
+      declineButton: 'Hayır, teşekkürler'
+    },
+    interventions: {
+      optOut: 'Bana daha fazla mesaj göstermeç',
+      studyInfo: 'Bu mesajlaşmada yer almak istemiyorum [study](+tab+https://docs.google.com/document/d/1gLyN6Dgff8dOCOC88f47OD6TtFrfSJltsLgJMKkYMso/preview).'
+    }
+  },
+  projects: {
+    welcome: {
+      heading: 'Hoş geldinız! Burada olduğunuz için çok mutluyuz',
+      thanks: 'Gerçek bir araştırmaya yardımcı olmak için gösterdiğiniz ilgiye teşekkür ederiz. Burada, yardımınızı kullanabileceğimiz birkaç projeyi bir araya getirdik. Daha fazla seçenek ve tüm aktif projelerimize göz atmak için aşağı kaydırmanız yeterlidir.',
+      talk: 'Diğer gönüllülerle sohbet edebileceğiniz [Talk](/talk) sayfasına da göz atmayı unutmayın.',
+      scrollDown: 'Daha fazlası için aşağı kaydırın'
+    }
+  },
+  project: {
+    language: 'Dil',
+    loading: 'Proje yükleniyor',
+    disclaimer: 'Bu proje Zooniverse Project Builder kullanılarak oluşturulmuştur ancak henüz resmi bir Zooniverse projesi değildir. Zooniverse Ekibine, bu proje ile ilgili yöneltilen sorular ve sorunlar herhangi bir yanıt almayabilir.',
+    fieldGuide: 'Kullanım Kılavuzu',
+    about: {
+      header: 'Hakkımızda',
+      nav: {
+        research: 'Araştırma',
+        results: 'Sonuçlar',
+        faq: 'SSS',
+        education: 'Eğitim',
+        team: 'Ekip'
+      }
+    },
+    nav: {
+      about: 'Hakkımızda',
+      adminPage: 'Admin sayfası',
+      classify: 'Transkripsiyon/Sınıflandırma',
+      collections: 'Toplamak',
+      exploreProject: 'Projeyi Keşfet',
+      lab: 'Lab',
+      recents: 'Son bilgiler',
+      talk: 'Sohbet',
+      underReview: 'İnceleme altında',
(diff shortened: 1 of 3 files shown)


lcjohnso:
**Significant Issue:** PFE uses `text-transform: uppercase` for multiple components, leading to the "infamous Turkish locale bug" where an uppercased version of "i" is not the correct replacement.  See [this blog entry](https://mattryall.net/blog/the-infamous-turkish-locale-bug) for a short write up.

Not sure how to proceed with this -- is there a way to hack the CSS for this particular locale?  Also acknowledging that this might require too much intervention to fix on PFE, but regardless thinking through a solution strategy would be helpful for FEM.

eatyourgreens:
That's an interesting bug. Is it affecting the project navigation here: [Seabird Watch in Turkish](https://pr-6160.pfe-preview.zooniverse.org/projects/tr/penguintom79/seabirdwatch).

eatyourgreens:
Also, 'Transcription' doesn't seem quite right for the Classify page there.

lcjohnso:
Re: "transcription" for "classify" -- good to remember that our translators are often working with a particular project context in mind, and it would be good to have them review their translations on two very different project types (i.e., transcription / humanities vs. astro) for accuracy of platform-level content.

eatyourgreens:
I see both dotted and non-dotted uppercase 'i' characters in the project navigation, but I don't know Turkish, so I can't say whether these are the correct characters or not.

Project navigation bar, showing uppercase navigation links in Turkish.

lcjohnso:
^^ I believe the dotted uppercase letters are incorrect.

eatyourgreens:
I found this, but it's eleven years old. [CSS text-transform not working properly for Turkish characters](https://stackoverflow.com/questions/3777443/css-text-transform-not-working-properly-for-turkish-characters).

eatyourgreens:
iOS looks ok (mobile Safari), comparing uppercase words in the project navigation with lowercase versions of the same words further down the home page. 

Desktop Firefox looks ok too but I think it would be useful to know what the project navigation should say, so that we can fix it.

eatyourgreens:
It might also be useful to know if this is just a problem in Chrome (and presumably Edge.)

eatyourgreens:
@lcjohnso we already add `lang=tr` to the project page. That might be why I’m not seeing any problems in mobile Safari or Firefox. 
https://github.com/zooniverse/Panoptes-Front-End/blob/45399f3267f45fa1d67d5bcd162b4050ed11e5ac/app/pages/project/project-page.jsx#L32

If you look in the CSS, we have styles to work around bugs in Czech too. 
https://github.com/zooniverse/Panoptes-Front-End/blob/45399f3267f45fa1d67d5bcd162b4050ed11e5ac/css/typography.styl#L5-L6

lcjohnso:
Mea culpa -- I was confused about what "correct" was supposed to look like.  I can confirm that I'm seeing consistent behavior between desktop Firefox and desktop Chrome, and in both cases I believe the navigation links are rendering correctly when upper cased: "i" -> "İ" AND "ı" -> "I".  

What confused me: seeing English words appear strangely in instances of mixed english+turkish -- example = the `<project name> Statistics` link on the project landing page.  For Seabirdwatch, it looks like this:
Screen Shot 2022-07-01 at 12 45 14 PM

Here, the issue is due to non-Turkish text mixed in ("Seabirdwatch") -- NOT incorrect capitalization transformations.

Maybe not such a big problem after all due to use of `lang=tr`!

eatyourgreens:
Project names can be translated too eg. Every Name Counts/Jeder Name Zählt so hopefully that wouldn’t be a problem on a real Turkish project.

eatyourgreens:
One problem we have run into in the past is the Karla font not containing the full range of Unicode characters in another language. That’s why we show Czech projects in Arial.

lcjohnso:
Yes, Sam and I were discussing the possible need for other scripts.  But for now, I'll reach out and get the translator / relevant project team to take a look and review the content.  They are best positioned to tell us if addition work is necessary.

Also noting here: the ultimate goal is to enable Turkish translation for an FEM project (using transcription task).  Nevertheless, this PFE PR will be a first step in populating the FEM dictionaries, so worth finishing.

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+        results: 'Sonuçlar',
+        faq: 'SSS',
+        education: 'Eğitim',
+        team: 'Ekip'
+      }
+    },
+    nav: {
+      about: 'Hakkımızda',
+      adminPage: 'Admin sayfası',
+      classify: 'Transkripsiyon/Sınıflandırma',
```
goplayoutside3, line 79:
```suggestion
      classify: 'Sınıflandırma',
```

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+      title: 'Transkripsiyon.Sınıflandırma',
+      themeToggle: '%(theme)s temasına geç'
+    },
+    home: {
+      organization: 'Organizasyon',
+      researcher: 'Araştırmacıdan sözler',
+      about: '%(title)s hakkında',
+      metadata: {
+        statistics: '%(title)s İstatistikler',
+        classifications: 'Transkripsiyonlar/Sınıflandırmalar',
```
goplayoutside3, line 100:
```suggestion
        classifications: 'Sınıflandırmalar',
```

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+    },
+    loading: 'Organizasyon yükleniyor...',
+    notFound: 'Organizasyon bulunamadı',
+    notPermission: "URL'nin doğru olduğundan eminseniz, bu organizasyonu görüntüleme izniniz olmayabilir.",
+    pleaseWait: 'Lütfen bekleyiniz...',
+    stats: {
+      adjustParameters: 'Parametreleri Ayarlayın',
+      byTheNumbers: 'Sayılara göre',
+      byTheNumbersContent: {
+        classifications: 'Transkripsiyonlar/Sınıflandırmalar',
```
goplayoutside3, line 156:
```suggestion
        classifications: 'Sınıflandırmalar',
```

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+        classifications: 'Sınıflandırmalar',
+        firstProject: 'İlk Proje',
+        firstProjectLaunch: 'İlk Proje Lansman Tarihi',
+        liveProjects: 'Devam Eden Projeler',
+        pausedProjects: 'Duraklatılmış Projeler',
+        retiredProjects: 'Tamamlanmış Projeler',
+        retiredSubjects: 'Tamamlanan Temalar',
+        subjects: 'Temalar'
+      },
+      classification: 'Transkripsiyonlar/Sınıflandırmalar',
```
goplayoutside3, line 165:
```suggestion
      classification: 'Sınıflandırmalar',
```

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+      },
+      classification: 'Sınıflandırmalar',
+      comment: 'Sohbet Yorumları',
+      dateRange: 'Tarih aralığı',
+      expandWorkflowStats: 'İş Akışı İstatistiklerini Genişlet',
+      for: 'için',
+      hidden: 'İstatistikler gizlendi',
+      hourly: 'Yalnızca en son 2 haftaya ait saatlik veriler mevcuttur.',
+      organizationStatistics: '%(title)s İstatistikler',
+      perclassification: 'Transkripsiyonlar/Sınıflandırmalar ',
```
goplayoutside3, line 173:
```suggestion
      perclassification: 'Sınıflandırmalar',
```

app/locales/tr.js
```diff
@@ -0,0 +1,471 @@
...
+      private: 'Özel',
+      submit: 'Koleksiyon Ekle'
+    }
+  },
+  emailSettings: {
+    email: 'E-posta adresi',
+    general: {
+      section: 'Zooniverse e-posta tercihleri',
+      updates: 'Genel Zooniverse e-posta güncellemelerini alın',
+      classify: 'Bir projede ilk transkripsiyon/sınıflandırma yaptığınızda e-posta güncellemesi alın',
```
goplayoutside3, line 255:
```suggestion
      classify: 'Bir projede ilk sınıflandırma yaptığınızda e-posta güncellemesi alın',
```

goplayoutside3 (approved):

goplayoutside3:
> **Significant Issue:** PFE uses `text-transform: uppercase` for multiple components, leading to the "infamous Turkish locale bug" where an uppercased version of "i" is not the correct replacement. See [this blog entry](https://mattryall.net/blog/the-infamous-turkish-locale-bug) for a short write up.
> 
> Not sure how to proceed with this -- is there a way to hack the CSS for this particular locale? Also acknowledging that this might require too much intervention to fix on PFE, but regardless thinking through a solution strategy would be helpful for FEM.
> 
> EDIT: it might be possible to add a `lang` attribute that would resolve this gotcha -- see [this SO entry](https://stackoverflow.com/questions/3777443/css-text-transform-not-working-properly-for-turkish-characters).

This bug is very apparent when adding Turkish to FEM. The localization feature in FEM translates the entirety of a page, including elements in the ZooHeader not built to be translated like my username.  For example, on a Turkish FEM page my username is transformed to uppercase and looks like this: DELİLAH / NOTİFİCATİONS.

PFE doesn't have as many spots where that's an issue simply because translations weren't built to cover all components of a page, so I think it's okay for this PR to stay merged. I'm not very familiar with Turkish, but can spot one obvious incorrect word --> Any PFE project displayed in Turkish labels the classifier Tutorial tab as "TUTORİAL" because the team who submitted the PFE Turkish dictionary left that key as English "Tutorial". 

I'll look for a solution for FEM before officially enabling Turkish there, and then assess if the same solution needs to be applied to PFE.

PR #6198, opened 2022-09-27
Update LinePlotModel (for canvas viewer) to use Chart.js v3

shaunanoordin:
## PR Overview

Follows #5943
❗ Affects experimental / no-longer-supported feature

This PR updates the LinePlotModel (for the canvas viewer, and it's actually a _scatter plot model,_ because why not) to properly use Chart.js v3

- Previously, the scatter plots on the LinePlotModel wouldn't render, since PR 5943 bumped the `chart.js` dependency to v3
- This is because chart.js v3 introduces tree-shaking, so the way the library is imported has to change. (See: [migration guide](https://www.chartjs.org/docs/latest/getting-started/v3-migration.html), [v3 integration guide](https://www.chartjs.org/docs/3.9.1/getting-started/integration.html))

### Dev Notes

For info on how the data should look like, see https://www.chartjs.org/docs/3.9.1/charts/scatter.html

Example [JSON](https://panoptes-uploads.zooniverse.org/subject_location/3abacdd5-d872-4045-b2d8-29ed7ee68e27.json?=):
```
{
   "datasets":[
      {
         "label":"Scatte Data!",
         "data":[
            {
               "x":-10,
               "y":0
            },
            {
               "x":0,
               "y":10
            },
            {
               "x":10,
               "y":5
            },
            {
               "x":0.5,
               "y":5.5
            }
         ],
         "backgroundColor":"rgb(255,99,132)"
      }
   ]
}
```

Associated Subject: [80709319](https://www.zooniverse.org/projects/camallen/talk-test-project/talk/subjects/80709319)

### Testing:

- Open https://pr-6198.pfe-preview.zooniverse.org/projects/camallen/talk-test-project/talk/subjects/80709319
- You should see a scatter plot chart rendered on your browser

Screenshot 2022-09-27 at 15 15 15

### Status

Ready for review. However, do note some _operational_ caveats:

- As discussed [on 5943](https://github.com/zooniverse/Panoptes-Front-End/pull/5943#issuecomment-843021367) and on [Slack](https://zooniverse.slack.com/archives/C06LHUC6R/p1664262051229589), **the canvas viewer is no longer supported.**
- Future project teams should be using FEM's [Scatter Plot Viewer](https://github.com/zooniverse/front-end-monorepo/tree/master/packages/lib-classifier/src/components/Classifier/components/SubjectViewer/components/ScatterPlotViewer#json-file) instead.
  - The only thing to note here is that the FEM Scatter Plot Viewer and its Subjects/data won't display properly on Talk, as that feature still needs to be integrated.

Long term, we may want to consider simply removing canvas-viewer from our code, since it's no longer officially supported on any project. Short term, this might still have some use for displaying data on Talk.

### DIFF
--- a/app/features/modelling/line-plot/index.js
+++ b/app/features/modelling/line-plot/index.js
@@ -1,4 +1,4 @@
-import { Chart } from 'chart.js';
+import Chart from 'chart.js/auto';
 
 // Help at http://www.chartjs.org/docs/latest
 class LinePlotModel {
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
         "@sentry/browser": "~6.1.0",
         "@sentry/tracing": "~6.1.0",
         "animated-scrollto": "~1.1.0",
-        "chart.js": "~3.2.1",
+        "chart.js": "~3.9.1",
         "chartist": "~0.11.0",
         "classnames": "~2.2.6",
         "counterpart": "~0.18.6",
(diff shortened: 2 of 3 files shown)


camallen (approved):
LGTM

PR #6212, opened 2022-10-12
Copy workflows via the Panoptes API

eatyourgreens:
- Copy workflows via the [`/project/{projectID}/links/workflows`](https://panoptes.docs.apiary.io/#reference/projects/project-create-links/add-a-link) API endpoint.
- Get the new workflow ID from the updated project links.
- Update the new workflow with the new title, then redirect.

Staging branch URL: https://pr-6212.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/pages/lab/actions/workflow.js
+++ b/app/pages/lab/actions/workflow.js
@@ -1,14 +1,27 @@
 var apiClient = require('panoptes-client/lib/api-client');
 
+var projects = apiClient.type('projects');
 var workflows = apiClient.type('workflows');
 
 var workflowActions = {
-  createWorkflowForProject: function(projectID, workflowData) {
+  createWorkflowForProject: function(project, workflowData) {
     var allWorkflowData = Object.assign({
-      links: {project: projectID}
+      links: {project: project.id}
     }, workflowData);
 
     return workflows.create(allWorkflowData).save();
+  },
+  copyWorkflowForProject: async function(project, workflowData) {
+    const oldLinks = new Set(project.links.workflows);
+    const [newProject] = await project.addLink('workflows', [workflowData.id]);
+    const newLinks = newProject?.links
+    const newWorkflowID = newLinks?.workflows.find(workflowID => !oldLinks.has(workflowID));
+    if (!newWorkflowID) {
+      return Promise.reject(new Error('no new workflow links on project.'))
+    }
+    const newWorkflow = await workflows.get(newWorkflowID)
+    await newWorkflow.update({ display_name: workflowData.display_name }).save();
+    return newWorkflow;
   }
 }
 
--- a/app/pages/lab/actions/workflow.js
+++ b/app/pages/lab/actions/workflow.js
@@ -14,8 +14,8 @@ var workflowActions = {
   copyWorkflowForProject: async function(project, workflowData) {
     const oldLinks = new Set(project.links.workflows);
     const [newProject] = await project.addLink('workflows', [workflowData.id]);
-    const newLinks = newProject?.links
-    const newWorkflowID = newLinks?.workflows.find(workflowID => !oldLinks.has(workflowID));
+    const newLinks = newProject?.links.workflows
+    const newWorkflowID = newLinks?.find(workflowID => !oldLinks.has(workflowID));
     if (!newWorkflowID) {
       return Promise.reject(new Error('no new workflow links on project.'))
     }
--- a/app/pages/lab/workflow-create-form.cjsx
+++ b/app/pages/lab/workflow-create-form.cjsx
@@ -25,18 +25,10 @@ WorkflowCreateForm = createReactClass
     workflowToClone = @props.workflowToClone
 
     newWorkflow =
+      id: workflowToClone?.id
       display_name: @refs.newDisplayName.value
-      primary_language: workflowToClone?.primary_language || @props.project.primary_language
-      steps: workflowToClone?.steps ? undefined
-      tasks: workflowToClone?.tasks ? {}
-      first_task: workflowToClone?.first_task ? ''
-      configuration: workflowToClone?.configuration ? {}
-      retirement: workflowToClone?.retirement ? {}
-      active: @props.workflowActiveStatus ? false
-      grouped: workflowToClone?.grouped ? false
-      prioritized: workflowToClone?.prioritized ? false
 
-    awaitSubmission = @props.onSubmit(@props.project.id, newWorkflow)
+    awaitSubmission = @props.onSubmit(@props.project, newWorkflow)
 
     Promise.resolve(awaitSubmission)
       .then (result) =>
--- a/app/pages/lab/workflow.cjsx
+++ b/app/pages/lab/workflow.cjsx
@@ -106,7 +106,7 @@ EditWorkflowPage = createReactClass
       </h3>
       {if @state.workflowCreationInProgress
         <ModalFormDialog tag="div">
-          <WorkflowCreateForm onSubmit={@props.workflowActions.createWorkflowForProject} onCancel={@hideCreateWorkflow} onSuccess={@handleWorkflowCreation}  project={@props.project} workflowToClone={@props.workflow} workflowActiveStatus={not @props.project.live} />
+          <WorkflowCreateForm onSubmit={@props.workflowActions.copyWorkflowForProject} onCancel={@hideCreateWorkflow} onSuccess={@handleWorkflowCreation}  project={@props.project} workflowToClone={@props.workflow} workflowActiveStatus={not @props.project.live} />
         </ModalFormDialog>}
       <p className="form-help">A workflow is the sequence of tasks that you’re asking volunteers to perform. For example, you might want to ask volunteers to answer questions about your images, or to mark features in your images, or both.</p>
       {if @props.project.live and @props.workflow.active


eatyourgreens:
Marking this as draft until we've had a chance to properly test it on staging.

eatyourgreens:
I think there might be a bug in the project builder where newly created workflows don't immediately show up in a project's links.


https://user-images.githubusercontent.com/59547/195413180-db46bf89-7fce-4750-a789-732c75028961.mov

eatyourgreens:
I think that might be a Panoptes bug, in that I'm not seeing the new workflow ID in the response that's sent back from the project link POST request.

eatyourgreens:
I've added a workaround, which is a bit of a hack. Request project workflows after updating links, then find the workflow with the largest ID.


https://user-images.githubusercontent.com/59547/195425682-05c19aaa-90d7-463e-827f-9d3950a3a72a.mov

app/pages/lab/actions/workflow.js
```diff
@@ -3,12 +3,23 @@ var apiClient = require('panoptes-client/lib/api-client');
...
+      links: {project: project.id}
     }, workflowData);
 
     return workflows.create(allWorkflowData).save();
+  },
+  copyWorkflowForProject: async function(project, workflowData) {
+    await project.addLink('workflows', [workflowData.id]);
+    project.uncacheLink('workflows');
+    // TODO: does this sort work for this query?
+    const newWorkflows = await project.get('workflows', { sort: '-updated_at' });
```
camallen, line 17: Currently this doesn't work, we can look to add it if it's useful, https://github.com/zooniverse/panoptes/blob/5b0cbf938e1e33830ec4204778aa0c6c9f5030bc/app/serializers/workflow_serializer.rb#L21

Alternatively we can modify the project response object to include some data on the copied project if that's easier,
i.e. something like
```
"projects": [{
"id": 1,
....
"copied_workflow_id": $id-of-newly-copied-workflow
}
# or in the links object of the resource response etc.
```
What do you think?

app/pages/lab/actions/workflow.js
```diff
@@ -3,12 +3,23 @@ var apiClient = require('panoptes-client/lib/api-client');
...
+    // TODO: is there a better way to get the ID of the new workflow?
+    const newWorkflowID = Math.max(...newWorkflows.map(workflow => workflow.id));
```
camallen, line 19: This is terrible and we can make this explicit in the create links response, see above.

app/pages/lab/actions/workflow.js
```diff
@@ -3,12 +3,23 @@ var apiClient = require('panoptes-client/lib/api-client');
...
+  createWorkflowForProject: function(project, workflowData) {
     var allWorkflowData = Object.assign({
-      links: {project: projectID}
+      links: {project: project.id}
     }, workflowData);
 
     return workflows.create(allWorkflowData).save();
+  },
+  copyWorkflowForProject: async function(project, workflowData) {
+    await project.addLink('workflows', [workflowData.id]);
```
camallen, line 14: i think the API should ensure we don't lose the workflow links if folks 'forget' to wrap the params in an array...that's terrible on our behalf and i can't see a usecase for overwriting the project's worklfow links. 

That said it runs the risk of breaking some clients expectations on API behaviours, https://www.hyrumslaw.com/

That said - I will change the API to ensure we never delete those links.

app/pages/lab/workflow-create-form.cjsx
```diff
@@ -25,6 +25,7 @@ WorkflowCreateForm = createReactClass
     workflowToClone = @props.workflowToClone
 
     newWorkflow =
+      id: workflowToClone?.id
       display_name: @refs.newDisplayName.value
       primary_language: workflowToClone?.primary_language || @props.project.primary_language
```
camallen, line 30: these `primary_language`, `steps`, `grouped` and `prioritized` attributes won't be needed anymore as the API copy function uses the source workflow data vs anything the client passes. I think the only attribute we need here is the `display_name` to update the workflow name after copying.

camallen (commented):
Looks good - a couple of questions about how we want to update the API response to make the clients life easier to figure out which workflow was copied.

app/pages/lab/actions/workflow.js
eatyourgreens: That sounds useful. According to the API docs, the new link should be in the links object returned from the POST request, but I couldn’t see it in `links.workflows`. 

https://panoptes.docs.apiary.io/#reference/projects/project-create-links/add-a-link

When I refreshed the page, to force a new load of the project from the API, the link was there but at a random position in the array. If possible, it would be useful if the serialiser sent the links back in creation order, or ordered by workflow ID (which might be the same thing.) In that case, I could reliably pluck the last ID from the array each time. 

Subject sets have the same problem so I’ll have a look at subject set creation to see how that works. I remember `addLink` being tricky when I worked on IIIF manifests.

app/pages/lab/actions/workflow.js
eatyourgreens: The API client has a very old TODO saying ‘should this always be an array?’ That can be enforced in JavaScript too. 
https://github.com/zooniverse/json-api-client/blob/f4cdb5498df385b38b3bd51a11ed80d44f14ea31/src/resource.coffee#L187

So old that it’s written in CoffeeScript.

app/pages/lab/actions/workflow.js
eatyourgreens: This query is also expensive on projects with large workflows. My test project has 25 workflows. I think Davy Notebooks was planning on having up to 80 workflows (though they might have switched to using a single workflow with many subject sets.)

eatyourgreens: Now that I write that, in many cases a project will want to adds new subject set rather than cloning a workflow. Worth checking with Sam as to exactly what this button is used for.

app/pages/lab/actions/workflow.js
eatyourgreens: If `project.links.workflows` was in a predictable order, I could pluck the last workflow from the array. 

I could also compare the new links array with the old one and take the difference.

app/pages/lab/workflow-create-form.cjsx
eatyourgreens: 👍🏻

app/pages/lab/actions/workflow.js
camallen: there are times when the link can be a singular resource, i.e. a project has one avatar etc so the array form doesn't make sense there. I think pushing the responsibility API side is a better approach than forcing the clients to know what kind of relation they are updating and which form is correct. 

> So old that it’s written in CoffeeScript.
It's a little worrying that a lot of our front end code relies on this client code, even the FEM code includes it some resources.

app/pages/lab/actions/workflow.js
camallen: I can intervene on specific resources in the serializer to say sort the workflow links by the id (i.e. in desc order) so you can quickly identify the 'latest' one. However with large payloads that does add a cost (most likely small) on each serialiser query.

For the copy workflow use case, i like your idea of doing a set diff of the links array before and after. That should uniquely identify the copied workflow. Failing that I can even add a new linked resource on this one end point like i added above, `{ copied_workflow: '$id' }` if that is easier. 

Let me know what you find regarding the `subject sets` creation behaviours and if we can generalize a pattern or just deal with this one use case.

camallen: I think this will fix the issue of the update links response not including the newly linked resources, https://github.com/zooniverse/panoptes/pull/3968

once that lands the relevant resource links response can be diffed with the previous links before the update to determine the newly added resource.

eatyourgreens: Cool. Right now, when I difference the new and old project links, they're both the same. This is after making a fresh project request too. Weirdly, If I go to the project builder workflows page and refresh the page, the new workflow is loaded then.

Screenshot of the workflow copy button showing a 'no new workflow links on project.' error.

eatyourgreens: I'll push that change up, so that it's available on staging.

app/pages/lab/actions/workflow.js
eatyourgreens: This code is gone now. 🙏

app/pages/lab/actions/workflow.js
eatyourgreens: Zoo Notes uses the new API client. I think that's a question of getting people out of the ingrained habit of using the old client. Or finding out why they're avoiding the new client and fixing it.

app/pages/lab/actions/workflow.js
camallen: I'll get that fix out to staging as soon as it's been reviewed

app/pages/lab/actions/workflow.js
```diff
@@ -1,14 +1,28 @@
...
+    project.uncache()
+    const newProject = await projects.get(project.id);
+    const oldLinks = new Set(project.links.workflows);
```
camallen, line 18: you'll need to get the old links before running the update as i believe the json api client will take the resource response from the `project.addLink('workflows', [workflowData.id]);` and update it's cached copy with the new copied workflow id in it.  

So if you do it in this order the two links arrays will match and we won't get the newly copied workflow id out of the response.

eatyourgreens: Good point. `addLink` makes a GET request to refresh the project too. Hidden side effects like this make the old API client a bit difficult to work with.

https://github.com/zooniverse/json-api-client/blob/f4cdb5498df385b38b3bd51a11ed80d44f14ea31/src/resource.coffee#L183-L191

eatyourgreens: Should be fixed now. I've updated the variable naming to use `oldLinks` and `newLinks` too. The new workflow ID is showing up in the API response too.

eatyourgreens:
@mcbouslog @lcjohnso @snblickhan you should be able to use this on staging without any errors now.
https://pr-6212.pfe-preview.zooniverse.org/

app/pages/lab/actions/workflow.js
```diff
@@ -1,14 +1,27 @@
...
-      links: {project: projectID}
+      links: {project: project.id}
     }, workflowData);
 
     return workflows.create(allWorkflowData).save();
+  },
+  copyWorkflowForProject: async function(project, workflowData) {
+    const oldLinks = new Set(project.links.workflows);
+    const [newProject] = await project.addLink('workflows', [workflowData.id]);
+    const newLinks = newProject?.links
```
camallen, line 17: not blocking but maybe for consistency of the links lookup and variable naming
```suggestion
    const newLinks = newProject?.links.workflows
```

app/pages/lab/actions/workflow.js
```diff
@@ -1,14 +1,27 @@
...
+    const newWorkflowID = newLinks?.workflows.find(workflowID => !oldLinks.has(workflowID));
```
camallen, line 18: if you adopt the proposed `links.workflows` change above
```suggestion
    const newWorkflowID = newLinks?.find(workflowID => !oldLinks.has(workflowID));
```

camallen (approved):
LGTM - minor non-blocking comment about consistent variable naming/use for `links.workflows`

eatyourgreens:
The copy button still throws errors when I use it with the production API, so I’m going to hold off on merging this. 
https://pr-6212.pfe-preview.zooniverse.org/

camallen:
> The copy button still throws errors when I use it with the production API, so I’m going to hold off on merging this. https://pr-6212.pfe-preview.zooniverse.org/

👍🏼 good point. I'll get the API changes deployed to prod asap next week.

camallen:
> > The copy button still throws errors when I use it with the production API, so I’m going to hold off on merging this. https://pr-6212.pfe-preview.zooniverse.org/
> 
> 👍🏼 good point. I'll get the API changes deployed to prod asap next week.

Those changes are now in production - this can merge/deploy safely

eatyourgreens:
@lcjohnso @snblickhan @mcbouslog this is now available on staging:
https://master.pfe-preview.zooniverse.org

PR #6215, opened 2022-10-14
Project builder: Fix undefined required tasks

### LINKED ISSUE #6214: Workflow Editor - Task's 'Required' checkbox may incorrectly appear checked, when required=undefined
## Project Builder Bug

Affects: Workflow Editor -> Task Editor
Freshdesk Ticket: [13976](https://zooniverse.freshdesk.com/a/tickets/13976)

When a WF has multiple (single question) Tasks - and some Tasks are required but others not - the Task editor may incorrectly display an _un-required Task as required..._ **IF** the user had previously looked at a required Task.

(i.e. the "Required" checkbox becomes visually "locked" to true/checked/enabled, after you click on a  required Task)

...but then again, this may **only** affect older Workflows, which still have data set like `myTask.required = "true"` (string) instead of `myTask.required = true` (boolean)

### Analysis

The problem only occurs one way; you HAVE to select the required Task first, then select an un-required Task. This triggers two errors/warnings on the console:

_Received the string `true` for the boolean attribute `checked`. Although this works, it will not work as expected if you pass the string "false". Did you mean checked={true}_

and

_A component is changing a controlled input to be uncontrolled. This is likely caused by the value changing from a defined to undefined, which should not happen._

The first error is more pertinent, and kudos to Jim for spotting that discrepancy when I shared the Workflow resource JSON earlier.

I do think the "incorrect data type" problem is just causing the "locked Required checkbox" state, because I can't replicate the problem with newly created Workflows (new WFs use `task.required=true` (boolean) instead of a string)

Attached videos: in **pt1,** I select the T4 task. The 'required' checkbox is false.
https://user-images.githubusercontent.com/13952701/195670312-23ac3a7f-5418-44dd-9831-998a6cee8dc3.mov

In **pt2,** I select the T1 (required) task first, then the T4 task. The 'required' checkbox incorrectly shows as true.
https://user-images.githubusercontent.com/13952701/195670371-5fb27c59-aa8c-4943-917b-97d86a8e1ce5.mov

### Status

This is a bug affecting live projects, but only older WFs. Priority level = ⁉️ 

Quick fix for project owners (untested): go into your old Workflows. For each Required Task, click on the checkbox to un-check it, then click it again to enable it. The task should now use `task.required=true` (boolean)

Project Builder fix: maybe see if I can add a filter/adapter that turns old "true" values to true when the project owner looks at the Task 🤔 

I'm not sure if we need to go through the DB and find all older Worflows that use `task.required="true"` (string) and manually change them to the correct value

eatyourgreens:
- Set a default for `task.required` in the workflow editor.
- Convert string values to booleans in the workflow editor.
- Add `required: false` to default tasks when new tasks are created.

Staging branch URL: https://pr-6215.pfe-preview.zooniverse.org
SquirrelMapper workflow from the original bug report: https://pr-6215.pfe-preview.zooniverse.org/lab/8149/workflows/20808

- Fixes #6214.
- This should fix the project builder showing tasks as required when `task.required` is undefined ~~but won't fix any tasks with `required: 'false'`, since it doesn't overwrite existing values~~. I'll add a couple of lines to convert string values to booleans.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/classifier/tasks/dropdown/index.jsx
+++ b/app/classifier/tasks/dropdown/index.jsx
@@ -263,6 +263,7 @@ DropdownTask.getDefaultTask = () => ({
   type: 'dropdown',
   instruction: 'Select or type an option',
   help: '',
+  required: false,
   selects: [
     {
       id: Math.random().toString(16).split('.')[1],
--- a/app/classifier/tasks/multiple/index.jsx
+++ b/app/classifier/tasks/multiple/index.jsx
@@ -66,6 +66,7 @@ MultipleChoiceTask.getDefaultTask = () => {
     type: 'multiple',
     question: 'Enter a question.',
     help: '',
+    required: false,
     answers: []
   };
 };
--- a/app/classifier/tasks/single/index.jsx
+++ b/app/classifier/tasks/single/index.jsx
@@ -67,6 +67,7 @@ SingleChoiceTask.getDefaultTask = () => {
     type: 'single',
     question: 'Enter a question.',
     help: '',
+    required: false,
     answers: []
   };
 };
--- a/app/classifier/tasks/subjectGroupComparison/index.jsx
+++ b/app/classifier/tasks/subjectGroupComparison/index.jsx
@@ -39,6 +39,7 @@ SubjectGroupComparisonTask.getDefaultTask = () => {
     type: 'subjectGroupComparison',
     question: 'Enter a question.',
     help: '',
+    required: false
   };
 };
 SubjectGroupComparisonTask.getTaskText = (task) => {
--- a/app/classifier/tasks/survey/index.jsx
+++ b/app/classifier/tasks/survey/index.jsx
@@ -16,6 +16,7 @@ export default class SurveyTask extends React.Component {
   static getDefaultTask() {
     return {
       type: 'survey',
+      required: false,
       characteristicsOrder: [],
       characteristics: {},
       choicesOrder: [],
--- a/app/classifier/tasks/textFromSubject/index.jsx
+++ b/app/classifier/tasks/textFromSubject/index.jsx
@@ -25,7 +25,8 @@ TextFromSubjectTask.getDefaultAnnotation = () => ({
 TextFromSubjectTask.getDefaultTask = () => ({
   help: '',
   instruction: 'Correct the text',
-  type: 'textFromSubject'
+  type: 'textFromSubject',
+  required: false
 });
 TextFromSubjectTask.getTaskText = task => (task.instruction);
 // isAnnotationComplete will return false within PFE to prevent classifications from being submitted within PFE, as the TextFromSubject task is exclusively for the front-end-monorepo.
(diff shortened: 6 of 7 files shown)


eatyourgreens:
This needs testing on staging, but I think it should work.

eatyourgreens:
SquirrelMapper task T4 (use tab to navigate the page and use the editor when it's blocked by an overlay like this.)

Workflow editor for SquirrelMapper, showing that question T4 is not actually required.

Side note: when the workflow editor is 'disabled' visually like this, the controls are still active and can be used from the keyboard. Active workflows are not really disabled from editing.

shaunanoordin:
Code read looks good, and I was able to replicate the Squirrel Mappers WF 20808 to my local for testing: https://local.zooniverse.org:3735/lab/1966/workflows/3675
- This would also effectively test if `task.required="true"` (string) triggers any additional issues.

I also have a more basic 3-question workflow to test: https://local.zooniverse.org:3735/lab/1966/workflows/3676
- `firstTask.required=true ; secondTask.required=undefined ; thirdTask.required=false`

Testing begins shortly, brb.

shaunanoordin (approved):
## PR Review

Fixes #6214 
Affects: Project Builder -> Workflow Editor -> Task Editor

This PR fixes the issue where (1) some Task types are created `task.required=undefined` by default, and (2) the "Required" checkbox showing incorrect states where there's a Task with `task.required=undefined`

- New Behaviour: most tasks (notable single and multiple question Tasks, but not drawing Tasks, since that's a different beast) have `task.required=false` by default
- Bug fix: if a previous Task has required=true, then a Task with required=undefined will no longer incorrectly display the "Required" checkbox as checked.

Code read looks very good, and testing showed that the fix works as intended, but there's one but missing: the changes to `app/pages/lab/workflow.cjsx` (PFE Lab) should also be applied to `app/pages/lab-fem/workflow.cjsx` **(FEM Lab)** ❗ 

### Testing

Tested on localhost with macOS + Chrome on localhost, on a Project Builder -> Workflow Editor page.

- Test URL: https://local.zooniverse.org:3735/lab/1966/workflows/3676?pfeLab=true
  - Workflow was created on `master` and has 3 single question Tasks, with this configuration: `T0.required=true , T1.required=undefined, T2.required=false`
  - _Note the `?pfeLab=true` ! This Workflow uses FEM Lab by default, so during my initial testing I was scratching my head for a long while before realising the fix currently only applies to the PFE Lab version._
- Testing for correct checkbox rendering:
  - Tasks Editor starts by automatically selecting T0
  - I'm able to switch to T1, and T1's "Required" checkbox correctly shows as un-checked.
  - I'm able to switch randomly between T0, T1, and/or T2 and the checkbox will always display the correct value: checked for _true,_ and un-checked for _false or undefined._ 👍 
- Testing for new Tasks:
  - When I create a new Single Question Task, I can check the network/WF resource to confirm that the new Task is created with `task.required=false` instead of undefined. 👍 

### Status

LGTM, but please apply changes for lab/workflow.cjsx to lab-fem/workflow.cjsx as well.

(Alternatively, just go ahead and merge this, I can copy the changes to lab-fem/workflow.cjsx in a separate PR - all good with me. 👍 )

PR #6216, opened 2022-10-14
Project builder: Fix undefined required tasks (FEM Lab ver)

### LINKED ISSUE #6214: Workflow Editor - Task's 'Required' checkbox may incorrectly appear checked, when required=undefined

shaunanoordin:
## PR Overview

Follows #6215 , which applies the same changes for PFE Lab's Workflow Editor.
Also fixes #6214, for FEM projects

This PR updates the FEM Lab's Workflow Editor so that "Required" checkbox for Tasks is checked when task.required is _true or "true",_ and un-checked when task.required is _false, "false", or undefined_

Please see #6215 for testing procedures. Note that the "Shortcut Editor" that exists on PFE Lab does not exist on FEM Lab.

Ready for review.

### DIFF
--- a/app/pages/lab-fem/workflow.cjsx
+++ b/app/pages/lab-fem/workflow.cjsx
@@ -536,18 +536,28 @@ EditWorkflowPage = createReactClass
 
         <div className={taskEditorClasses}>
           {if @state.selectedTaskKey? and @props.workflow.tasks[@state.selectedTaskKey]?
-            TaskEditorComponent = taskComponents[@props.workflow.tasks[@state.selectedTaskKey].type]?.Editor
+            task = @props.workflow.tasks[@state.selectedTaskKey]
+            if task.required is 'true'
+              task.required = true
+            if task.required is 'false'
+              task.required = false
+            TaskEditorComponent = taskComponents[task.type]?.Editor
+            taskWithDefaults = {
+              required: false,
+              ...task
+            }
             <div>
               {if TaskEditorComponent  # Note: ShortcutEditor deprecated (and removed entirely) for FEM-lab
                 <TaskEditorComponent
                   workflow={@props.workflow}
-                  task={@props.workflow.tasks[@state.selectedTaskKey]}
+                  task={taskWithDefaults}
                   taskPrefix="tasks.#{@state.selectedTaskKey}"
                   project={@props.project}
                   onChange={@handleTaskChange.bind this, @state.selectedTaskKey}
                 />
               else
-                <div>Editor is not available.</div>}
+                <div>Editor is not available.</div>
+              }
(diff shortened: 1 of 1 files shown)


shaunanoordin:
Once the Staging deploy is done, you can test the changes at https://pr-6216.pfe-preview.zooniverse.org:3735/lab/1966/workflows/3676?femLab=true&env=staging (FEM Lab ver)

eatyourgreens (approved):

PR #6219, opened 2022-10-19
Remove coffeescript from Project Editor

eatyourgreens:
- rewrite `project.cjsx` as functional JSX components.
- Remove `ChangeHandler` and `PromiseRenderer`.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/lab/project.jsx
```diff
@@ -0,0 +1,359 @@
...
+    setDeleteState(oldState => ({ ...oldState, deletionError: null }));
+    const confirmPrompt = `You are about to delete this project and all its data!
+    Enter ${DELETE_CONFIRMATION_PHRASE} to confirm.`;
+    const confirmed = prompt(confirmPrompt) === DELETE_CONFIRMATION_PHRASE;
+
+    if (confirmed) {
+      setDeleteState(oldState => ({ ...oldState, deletionInProgress: true }));
+      project.delete()
+        .then(() => {
+          context.router.push('/lab')
```
eatyourgreens, line 59: I think legacy context has been removed in React 17, so this probably won't work any more.

app/pages/lab/project.jsx
```diff
@@ -0,0 +1,355 @@
...
+                className="standard-button view-project-button"
+                target="_blank"
+                title="Open the current project in a new tab."
+              >
+                View project
+              </Link>
+            </li>
+            <li>
+              <IndexLink
+                aria-current='page'
```
eatyourgreens, line 102: This should be dynamically added to the link for the current page.

app/pages/lab/project.jsx
```diff
@@ -0,0 +1,367 @@
...
+import LoadingIndicator from '../../components/loading-indicator.jsx';
+import { Helmet } from 'react-helmet';
+import apiClient from 'panoptes-client/lib/api-client';
+import counterpart from 'counterpart';
+import workflowActions from './actions/workflow.js';
+import isAdmin from '../../lib/is-admin.coffee';
+import LabStatus from '../../partials/lab-status.jsx';
+import { isThisProjectUsingFEMLab, FEM_LAB_PREVIEW_HOST } from '../lab-fem/fem-lab-utilities.js';
+
+const DEFAULT_SUBJECT_SET_NAME = 'Untitled subject set'
```
shaunanoordin, line 14: [minor] Is this still being used?

shaunanoordin (changes requested):
## PR Review

Affects: Project Builder

This PR refactors the EditProjectPage, changing CoffeeScript to JavaScript. Functionality/features should remain the same.

The EditProjectPage is the main "container"/"body"/"frame" for editing a single project for the Project Builder.

Example URL: https://local.zooniverse.org:3735/lab/1982 (editing a testing project on staging) ; since this is a "container", changes here also tangentially affects all **sub-paths**

Main Features (parity):

- Checks for user auth: (includes sub-paths)
  - If user is not logged in, display a "you need to login" error message.
  - If user is logged in but doesn't have access to the selected project, display a "you don't have permission" error message.
  - If user is logged in and has access to the selected project, display the relevant sub-editors/sub-pages: (e.g. Project Details editor, Show Workflows page, etc)
- Allows users to navigate to sub-pages using links (on left-hand of page):
  - "View project" should link to different front ends (PFE or FEM), depending on whether user is using PFE Lab or FEM Lab (debugged with `?pfeLab=true` or `?femLab=true`)
  - Other links should work as expected, taking users to the appropriate sub-editor/sub-page.
- Gives user option to delete the project:
  - The "Delete this Project" button will prompt the user to type in a confirmation phrase.
  - If the user types in the confirmation phrase exactly AND (presumably) has permissions, the project will be deleted and the user taken to the main Project Builder page which lists all projects.
- Displays sub-pages, given the correct sub-path:
  - (Again, a result of this being the main "container"/"body"/"frame" for editing a single project)

Code read looks good, but testing has revealed some weird wonk.

### Testing

Tested on localhost with macOS + Chrome, using a series of test projects, as user `darkeshard`

Test URLs:
- Random test project owned by darkeshard: https://local.zooniverse.org:3735/lab/1982 and various others
- Test project NOT owned by darkeshard: https://local.zooniverse.org:3735/lab/335 (ONLY used to test login state, no modifications or deletions made.)

Features tested:
- ✅ Checks for user auth:
  - I tried viewing the test URLs (and a random sampling of sub-paths) as darkeshard, and then as a non-logged-in user. As expected, I can only view the project editor if I have permissions. 👍 
- ✅ Allows users to navigate to sub-pages using links
  - Links work as expected, tested by clicking on 'em. Confirmed that "View Project" is different between FEM Lab and PFE Lab.
- ✅ Gives user option to delete project:
  - Works as expected. Tested by following delete process, to successfully delete at least 2 newly-created test projects.
- ❓ Display sub-pages:
  - Sub-pages such as "About" and "Collaborators" work fine
  - ❌ However, at least 2 (probably related) issues have been in "Project Details" and "Workflows"

**Issue 1:** Can't edit "Name", "Description", etc under the "Project Details" sub-page

I'm unable to change the project's name, intro, or description text, _among other things._ Keyboard input appears to be ignored when typing into the respective text boxes. (More specifically, the input.value seems locked and isn't being updated on the 'onChange' event.)

Similarly, dragging/dropping an avatar or background image seems to make no changes until the page is refreshed.

Example URL: https://local.zooniverse.org:3735/lab/1982

_In the video below, I click on various text boxes and then randomly jam "asd" on my keyboard. Only the "Workflow Description" textbox registers any of my keyboard input._
https://user-images.githubusercontent.com/13952701/197251582-b1184d0e-c39b-45fc-90bf-1f6c2f19b36b.mov

Example URL: https://local.zooniverse.org:3735/lab/1982/workflows

- ❗ **UPDATE"** Which inputs are "locked" can change. After recording the video above and refreshing the page, I realise I couldn't edit the "workflow description" text box any further. Similarly, I could initially edit the "Announcement Banner" input text, but after reloading, the value gets "locked".
- **UPDATE 2:** I suspect that anything that is `undefined` in the project resource isn't locked, but once it gets any value (including `''` an empty string), it gets locked, as per React's "controlled values" rule.

**Issue 2:** Create new Workflows

On the "Workflows" sub-page, after I click on "New Workflow" and type in a name for my new workflow in the modal popup, nothing happens. The expectation here is that I should be taken to the WF Editor page with that newly created WF selected.

_Video below: exactly what I said_
https://user-images.githubusercontent.com/13952701/197253930-3a9b33d0-a3c3-4cad-83e4-2168376e0f72.mov

- Notably, there's no network call. So whatever's failing, it's happening before the POST/PUT request has a chance to be triggered.

### Status

Fixes required; I'm assuming there's something going on here with project resources not being cloned properly or something?

I've checked to ensure that Issues 1 and Issues 2 are _not_ extant on master.

eatyourgreens:
#6220 has the same bug: inputs are locked. The `ChangeListener` that I removed must be doing something to rerender the page when the watched resource changes, similar to `observer` in MobX.

eatyourgreens:
Unfortunately, the PFE Readme doesn’t offer much practical advice beyond ‘don’t use the `ChangeListener` component.’

app/pages/lab/project.jsx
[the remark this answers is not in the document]
eatyourgreens: When you create a new subject set, maybe?

eatyourgreens:
I read through the `ChangeListener`. It does something like this:

```js
  project.listen(‘change’, setState);
```

The workflow editor has similar code listening to the workflow. 

So the API client is being used as a state store, via `resource.update`, and needs to notify the React component when it changes. 

https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/components/change-listener.cjsx

eatyourgreens:
That code can be replaced by an effect that adds/removes the listener when the project/workflow changes.

app/pages/lab/project.jsx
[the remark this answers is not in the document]
eatyourgreens: I’m not sure where this is used, either.

eatyourgreens:
@shaunanoordin I've added a `useEffect` hook that replaces the old `ChangeHandler` component. That should fix the bugs that you saw.

#6220 will need a similar hook that watches the workflow for changes.

eatyourgreens:
I've also fixed `workflowActions`, which wasn't being imported into the Workflows page. That explains why the 'New workflow' button didn't work: the default form action was an empty function.

shaunanoordin (approved):
## PR Review (Update)

Follows: [earlier review](https://github.com/zooniverse/Panoptes-Front-End/pull/6219#pullrequestreview-1151254680)

Latest changes fix the 2 issues noted in the earlier review. All 4 features looked OK on tests. Approving.

PR #6227, opened 2022-10-23
Replace workflow change listener with a hook

eatyourgreens:
Replace the workflow editor's `ChangeHandler` with a hook, which subscribes the page to workflow changes, rerendering when `workflow.update()` runs.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/lab-fem/workflow.jsx
```diff
@@ -1082,6 +1083,16 @@ export default function EditWorkflowPageWrapper({
       .then(setWorkflow);
   }, [params.workflowID]);
 
+  useEffect(function subscribeToWorkflow() {
+    function onWorkflowChange(...payload) {
+      setPayload(payload);
```
eatyourgreens, line 1088: The API client doesn’t send a payload with `change` events, but the `ChangeListener` needs this state change to rerender when the workflow updates. 

https://github.com/zooniverse/json-api-client/blob/f4cdb5498df385b38b3bd51a11ed80d44f14ea31/src/model.coffee#L54

eatyourgreens: So this line could be rewritten with an empty array, and it would still work. 

```js
      setPayload([]);
```

eatyourgreens: Setting workflow state to the updated workflow might work, but might also be just as confusing to read as the code here. The `this` context of an API client event is the resource that fired the event. 

```js
function onWorkflowChange() {
  const updatedWorkflow = this;
  setWorkflow(updatedWorkflow);
}
```

PR #6247, opened 2022-11-07
Workflow editor: extract TaskPicker and TaskIcon

eatyourgreens:
Replace a couple of the auto-running anonymous closures, generated by decaffeinate, with functional React components:
- `TaskIcon` returns a font-awesome icon for a given task type.
- `TaskPicker` returns an array of `TaskButton` buttons for each workflow task, labelled with the icon, task name and task key. Clicking a button selects that task for editing.

Staging branch URL: https://pr-6247.pfe-preview.zooniverse.org

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/pages/lab/workflow-components/task-picker.jsx
```diff
@@ -0,0 +1,49 @@
+import taskComponents from '../../../classifier/tasks';
+import TaskIcon from './task-icon.jsx';
+
+function TaskButton({ active, definition, firstTask, onClick, value }) {
+  const taskDefinition = taskComponents[definition.type]?.getTaskText(definition);
+  const classNames = ['secret-button', 'nav-list-item'];
+  if (active) {
+    classNames.push('active');
+  }
```
eatyourgreens, line 9: We can get rid of this and use an attribute selector instead.
```css
button[aria-pressed=true] {
…
}
```

app/pages/lab/workflow-components/task-picker.jsx
```diff
@@ -0,0 +1,49 @@
...
+        aria-pressed={ active ? 'true' : 'false' }
+        type="button"
+        className={classNames.join(' ')}
+        name="selectedTask"
+        value={value}
+        onClick={onClick}
+      >
+        <TaskIcon type={definition.type} />
+        {' '}
+        {taskDefinition || 'Task editor is unavailable'}
```
eatyourgreens, line 22: Should the button be clickable when the label is "Task editor is unavailable."?

app/pages/lab/workflow.jsx
```diff
@@ -165,49 +166,12 @@ class EditWorkflowPage extends Component {
...
-                            </button>
-                          </div>);
-                        } else {
-                          result.push(undefined);
-                        }
-                      }
-                      return result;
-                    })()}
+                    <TaskPicker
+                      onClick={(event) => this.setState({ selectedTaskKey: event.target.value })}
```
eatyourgreens, line 170: Double check whether `event.target.value` is set when you click the label icon.

app/pages/lab/workflow-components/task-picker.jsx
shaunanoordin: I think it's perfectly OK for a Task with no Task Editor to be clickable.

_Screenshot: Planet Hunters TESSting with a Light Curve Task, which has no editor._
image

(Note: some jiggering was required to display this; see full review notes.)

shaunanoordin (approved):
## PR Review

Affects: Project Builder -> Workflow Editor

This PR refactors the "Tasks for a Workflow" list (i.e. TaskPicker).

_Screenshot: the list of Tasks (TaskPicker) on a test workflow, as viewed on the Workflow Editor. NOTE: Task icons aren't appearing because FontAwesome's CDN is inaccessible on my Mac in Malaysia for some bizarre reason. It probably got blocked by the national firewall for obscenity or something equally ridiculous._
image

Expected behaviour: (no changes introduced in this PR)
- The list should show all Tasks for a WF
  - If the WF has no Tasks, this list should show "No tasks yet" (this option can't be clicked)
- For each Task, its "main text" should appear as an item on the list.
  - Clicking on a Task on the list should load the corresponding Task Editor on the right-hand column of the page.
  - If the Task is a first task of the WF, it should be marked with the text "first".

Code read looks good: the inline-rendering bits have been extracted into their own components. There's a small flaw with TaskIcon (see Testing & Status), and we need to copy the PFE Lab's WF editor changes to FEM Lab (see Status), but otherwise everything is good. 👍 

### Testing

Tested on localhost with macOS + Chrome on [a random test project with PFE Lab enabled.](https://local.zooniverse.org:3735/lab/1986/workflows/3723?pfeLab=true)

Basic behaviour tests:
- Create a new WF. Observe that with no Tasks, the TaskPicker (list) says "no tasks yet"
- Create Task T1, then Task T2, etc. Observe that these Tasks now appear in the TaskPicker, with T1 being the first task by default.
- Change the first task to Task T2. Observe the change to the list.
- Delete all Tasks. Observe the return of "no tasks yet".

Additional tests:
- Test for various Task types: Question, Text, Drawing, and **Light Curve Task**

❗ Issue discovered: page crashes when viewing Planet Hunters TESSting
- Test URL: https://local.zooniverse.org:3735/lab/18190/workflows/20991?env=production&pfeLab=true
- When the TaskPicker encounters an _unaccounted-for Task,_ such as the Light Curve Task, the whole page crashes.
- Error message is `Error: TaskIcon(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.`
- This is due to TaskIcon returning... undefined? Which apparently causes the React rendering to choke.

### Status

Changes look mostly good, with only two issues:
- TaskIcon needs to have a `return null` to ensure it doesn't crash the page when an _unaccounted-for Task_ (i.e. Light Curve Task) is rendered.
- The changes to `lab/workflow.jsx` also needs to be copied to `lab-fem/workflow.jsx`

These changes are pretty straightforward, so I've opened #6269 to merge into this PR, before this PR gets merged. Anyway, LGTM 👍

PR #6389, opened 2023-01-25
Project Builder: remove CoffeeScript from Multi Image Subject Options Editor

shaunanoordin:
## PR Overview

This PR converts the Multi Image Subjects Options Editor from CoffeeScript to JavaScript. _No behaviour changes are expected._

Context: FEM projects will start supporting the Flipbook feature on the FEM Classifier, and as a result, the FEM Lab now needs to support configuring those options. At the moment, all "Multi Image Options" on FEM Lab are deprecated, so in preparation for future development, we need to decaffeinate the component.

### Testing

Testing steps:
- Open the Project Builder, choose a project, create a workflow. Ensure the WF Editor is in PFE Lab mode. [(Example)](https://master.pfe-preview.zooniverse.org/lab/1986/workflows/3714?pfeLab=true&env=staging)
- Default values: when the workflow is created, observe that the "Multi-Image Options" section has default values.
- Changing values: start modifying values on each option.
  - Reload the page to ensure the changed values stick.
  - Note that "iterations" defaults to 3, but deleting the value (setting it to blank) will set an implicit value of infinity
  - Note that setting layout to "Show separate frames" will display additional options for layout.

Additional Classifier checks:
- Try viewing the newly-created workflow on the PFE Classifier. All options should work as expected (Assumes you have multi-image Subjects.)

Additional FEM Lab checks:
- The newly-created workflow should be viewable in FEM Lab mode. [(Example)](https://master.pfe-preview.zooniverse.org/lab/1986/workflows/3714?femLab=true&env=staging)
  - The "Multi-Image Options" but should be deprecated (not editable/non-interactive), but the values should still be visible

### Status

Ready for review.

A follow up will be required to add FEM Lab-specific options to the Multi Image Subject Options Editor.


app/components/multi-image-subject-options-editor.jsx
```diff
@@ -0,0 +1,143 @@
...
+        </div>
+      </div>
+    )
+  }
+
+  return (
+    <ChangeListener target={workflow}>
+      {render}
+    </ChangeListener>
+  )
```
eatyourgreens, line 142: `ChangeListener` shouldn't be used in production code (see the Readme.)

app/components/multi-image-subject-options-editor.jsx
```diff
@@ -0,0 +1,143 @@
...
+            <label>
+              <input
+                type='radio'
+                value='grid3'
+                checked={layout === 'grid3'}
+                onChange={handleSelectLayout}
+              />
+              Grid (3col)
+            </label>
+          </div>
```
eatyourgreens, line 106: Accessibility error: These radio buttons are all missing a name attribute. That should have been caught by ESLint. 🤔

eatyourgreens (changes requested):
The radio buttons need a name to group them, and the `ChangeListener` should be replaced with local component state.

app/components/multi-image-subject-options-editor.jsx
```diff
@@ -0,0 +1,143 @@
...
+        {workflow?.configuration?.multi_image_mode === 'separate' ?
+          <div>
+            <span>
+              Show separate frames as:
+            </span>
+            <br/>
+            <label>
+              <input
+                type='radio'
+                value='row'
```
eatyourgreens, line 80:
```suggestion
                name='multi_image_layout'
                value='row'
```

app/components/multi-image-subject-options-editor.jsx
```diff
@@ -0,0 +1,143 @@
...
+                checked={layout === 'row'}
+                onChange={handleSelectLayout}
+              />
+              Single row (side by side)
+            </label>
+            <br/>
+            <label>
+              <input
+                type='radio'
+                value='grid2'
```
eatyourgreens, line 90:
```suggestion
                name='multi_image_layout'
                value='grid2'
```

app/components/multi-image-subject-options-editor.jsx
```diff
@@ -0,0 +1,143 @@
...
+                checked={layout === 'grid2'}
+                onChange={handleSelectLayout}
+              />
+              Grid (2col)
+            </label>
+            <br/>
+            <label>
+              <input
+                type='radio'
+                value='grid3'
```
eatyourgreens, line 100:
```suggestion
                name='multi_image_layout'
                value='grid3'
```

app/components/multi-image-subject-options-editor.jsx
eatyourgreens: I think the lab generally expects the name of a form control to be the name of the thing that it's setting, so I've changed these to `multi_image_layout`.

eatyourgreens (changes requested):

shaunanoordin:
Thanks Jim!

- [ ] All 3 Radio buttons will have names set to `multi_image_layout`
- [ ] ChangeListener will be removed

app/components/multi-image-subject-options-editor.jsx
shaunanoordin: 👌

app/components/multi-image-subject-options-editor.jsx
shaunanoordin: 👌

app/components/multi-image-subject-options-editor.jsx
shaunanoordin: 👌

shaunanoordin:
## PR Update

- ChangeListener removed.
- Accessibility fix: radio buttons now have name='multi_image_layout'

app/components/multi-image-subject-options-editor.jsx
shaunanoordin: 👌

eatyourgreens (approved):

eatyourgreens:
Looks good on this test workflow: https://pr-6389.pfe-preview.zooniverse.org/lab/12754/workflows/15651

I expected that the `ChangeListener` would have to be replaced with a `useEffect` hook, but the form seems to work without it at all. I wasn't expecting that.

PR #6732, opened 2023-07-20
Revise ZA relevant language entries

lcjohnso:
Staging branch URL: https://pr-6732.pfe-preview.zooniverse.org

Matching https://github.com/zooniverse/pandora/pull/661, update language entries relevant to South Africa, taking the following actions
- Remove duplicate entries.
- Make all `-za` entries general by removing country specification (non-specific versions were missing before).
- Edit labels for ZA-relevant entries to include standard/reference language name (matching ISO language name) and local names separated by "/" character.  e.g., for `xh`, "Xhosa / isiXhosa"

Surpasses and replaces https://github.com/zooniverse/Panoptes-Front-End/pull/6731.

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional


app/constants/languages.js
```diff
@@ -77,6 +75,7 @@ const languages = [
   { value: 'moh-ca', label: 'Mohawk (Mohawk)' },
   { value: 'mn', label: 'Mongolian' },
   { value: 'ne-np', label: 'Nepali (Nepal)' },
+  { value: 'nso', label: 'Northern Sotho / Sesotho sa Leboa' },
```
eatyourgreens, line 78: Can we please not use colonial names?

eatyourgreens (changes requested):
Can we use the language names that we’ve been given by our colleagues in South Africa?

app/constants/languages.js
lcjohnso: As I wrote in the PR text, I'm opting to include _both_ the official ISO Language Name and the local/native name. Including both names is useful -- it resolves possible confusion in the case of "Sesotho" -- Sesotho = Southern Sotho = `st` vs. Sesotho sa Leboa = Northern Sotho = `nso`.

eatyourgreens: The ISO name is not official. Can we use the names that we’ve been asked to use by people who speak these languages.

eatyourgreens: Also, ‘native name’????

eatyourgreens:
Here's the list of languages that we've been given:
- isiXhosa
- Sesotho
-Tshivenḓa
- Xitsonga
- isiZulu

If we could stick with those names, that would be great. If we must add the European names too, then we could maybe do something like 'Xitsonga / Tsonga' but I'm not sure what value the European names add to the project builder.

PR #6787, opened 2023-08-29
Pages Editor: initial setup. Add 'editor' route. Fetch Workflow.

shaunanoordin:
## PR Overview

Staging branch URL: https://pr-6787.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711

The Pages Editor is a variant of the Workflow editor in the Project Builder, except it's designed to _edit FEM workflows, in the PFE code base._ Check the `#fem-lab` channel on Slack for more details on the super-project.

Some core goals:
- The Pages Editor should allow a project owner to replicate I Fancy Cats or other simple project.
- The Pages Editor MVP needs to be completed fast (1-2 months)
- The Pages Editor code should NOT affect any other PFE code. (Try no to change existing files, don't add new dependencies if possible, etc)
- _Stretch goal:_ Pages Editor code should be portable(-ish) to a new home on a new Front-End-Monorepo package.

Added in this PR:
- New `/lab/{projectId}/workflows/editor/{workflowId}` route (`router.jsx` is the only existing file edited)
  - Example: https://local.zooniverse.org:3735/lab/1982/workflows/editor/3711
- New `/app/pages/lab-pages-editor` folder for all Pages Editor code
  - `<DataManager>` component, which manages fetching & updating the Workflow resource.
    - DataManager will fetch the Workflow resource **once,** on page load, and after that the local copy of the WF will act as the _source of truth_ that is _synced back up_ to Panoptes.
    - All the children wrapped by DataManager will have access to 1. the Panoptes Workflow resource (for reading data) and 2. a placeholder `update()` function (when complete, this function will update the local WF resource _and_ sync it to Panoptes.)
  - `<Tester>` component 

### Dev Notes

Here are some (questionable?) decisions made in the course of this PR.

**Aborted useApiData() hook:** At one point, I extracted all the action in `whenComponentMounts_doInitialise()` in `<DataManager>` into its own `useApiData()` hook, to make the code more declarative.

Unfortunately, I then realised that the `apiData` state might need to be managed by the DataManager itself, since the apiData.state (and maybe apiData.workflow, to prompt _observable_ changes) will also need to be jostled when the `update()` function _posts_ to Panoptes.

**No Stores For You:** there's no "global store" or "app-level store", so data is managed entirely by context (for sharing `workflow` and `update`) or local component states. This is as simple as I can make it, though this might bite me in the butt later if my sub-components somehow can't observe changes to the workflow.

### Status

Ready for review

Next up:
- Write to Workflow resource (i.e. update Panoptes).

❓ Unanswered questions (which I just thought of 5 minutes ago):
- Does the DataManager need to handle CREATING a new workflow? wait, that's a no, right? Right? Cripes, I need to double-check how users create workflows and are redirected(?) to the Workflow editor.


shaunanoordin:
> One small thing: All JSX files should be .jsx, not .js, in this repo.

👍 Thanks, this has been updated!

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,82 @@
...
+
+  // Fetch workflow when the component loads for the first time.
+  // See notes about 'key' prop, to ensure states are reset:
+  // https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
+  // Also see general pattern notes:
+  // https://react.dev/reference/react/useEffect#fetching-data-with-effects
+  useEffect(async function whenComponentMounts_doInitialise () {
+    console.log('+++ DataManager: fetchWorkflow', workflowId)
+    if (initialised) return 
+    if (!workflowId) return
```
eatyourgreens, line 30: You should return your cleanup function here, as well as later on. Having different return values from the effect, depending on external variables, might lead to behaviour that's difficult to debug later on.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,82 @@
...
+      console.error('DataManager: ', err)
+      setApiData({
+        workflow: null,
+        status: 'error'
+      })
+      initialised = true  // Don't re-run the initialisation function again
+    }
+
+    return () => { initialised = false }
+  }, [ workflowId ])
```
eatyourgreens, line 57:
```suggestion
  }, [ initialised, workflowId ])
```

This hook also depends on `initialised`. All of an effect hook's reactive dependencies must be declared here. You can't choose which values a `useEffect` hook will react to.
https://react.dev/reference/react/useEffect#specifying-reactive-dependencies

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,82 @@
...
+    workflow: apiData.workflow,
+    update
+  }
+
+  if (!workflowId) return (<div>ERROR: no Workflow ID specified</div>)
+  // if (!workflow) return null
+
+  return (
+    <WorkflowContext.Provider
+      value={contextData}
```
eatyourgreens, line 76: The API client is also a state store, so be aware that your context here might differ from the workflow stored in the client, once you start to edit and update values. You can debug this by logging the client to the console and inspecting its internal caches.

```js
// get workflow 1234 from Panoptes.
let workflow = await apiClient.type('workflows').get('1234')

// get the stored workflow from the client's internal caches.
workflow = await apiClient.type('workflows').get('1234')

// make some changes.
workflow.update(changeSet)

// saving fetches a fresh copy of workflow 1234 from Panoptes, via workflow.refresh(), before saving.
workflow.save()
```

https://github.com/zooniverse/panoptes-javascript-client/blob/b2f2b8aebc3ca43321619d29c5191fd8d536a4e2/lib/json-api-client/resource.js#L56-L58

app/pages/lab-pages-editor/context.jsx
```diff
@@ -0,0 +1,9 @@
+import { createContext, useContext } from 'react'
+
+const defaultWorkflowData = {}
+export const WorkflowContext = createContext(defaultWorkflowData)
+
+export function useWorkflowContext () {
+  const workflowData = useContext(WorkflowContext)
+  return workflowData
+}
```
eatyourgreens, line 9: This is plain JS, not JSX, so the file should be named `context.js`.

eatyourgreens (changes requested):
Code changes look ok, but your effect hook is breaking the rules for `useEffect`. The React linter should catch that.

Small linting tip: PFE uses semicolons, but you should be able to add them with `eslint --fix`.

I'd also recommend moving your data-fetching code into a custom hook. It can make components easier to read. Something like:
```js
const { workflow, error, loadingState } = usePanoptesWorkflow(workflowId);
```

I think the IIIF code in the lab has examples. Also the new classifier in the monorepo.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,82 @@
...
+    if (initialised) return 
+    if (!workflowId) return
+
+    try {
+      setApiData({
+        workflow: null,
+        status: 'fetching'
+      })
+
+      const wf = await apiClient.type('workflows').get(workflowId)
```
eatyourgreens, line 38: If this is going to be used in the monorepo, it needs to use `@zooniverse/panoptes-js` as its API client.

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: Thanks, I missed this major error - I didn't consider that the cleanup function wouldn't be returned in certain cases and would mess things up. 👌 I've completely rewritten the effect.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,97 @@
...
+  // Fetch workflow when the component loads for the first time.
+  // See notes about 'key' prop, to ensure states are reset:
+  // https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
+  // Also see general pattern notes:
+  // https://react.dev/reference/react/useEffect#fetching-data-with-effects
+  useEffect(() => {
+    if (!initialised) fetchWorkflow();
+    initialised = true;
+    return () => { initialised = false; };
+  }, [workflowId]);
```
eatyourgreens, line 55:
```suggestion
  }, [fetchWorkflow, initialised]);
```

The hook's external dependencies are now `initialised` and `fetchWorkflow`. `fetchWorkflow` will be a new function on every render, so be careful about that, as it will lead to your effect hook possibly triggering on each render.

The React docs recommend declaring your data-fetching function inside the effect hook, so that a new function is only created each time that the effect runs.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,97 @@
...
+
+  // Fetch workflow when the component loads for the first time.
+  // See notes about 'key' prop, to ensure states are reset:
+  // https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
+  // Also see general pattern notes:
+  // https://react.dev/reference/react/useEffect#fetching-data-with-effects
+  useEffect(() => {
+    if (!initialised) fetchWorkflow();
+    initialised = true;
+    return () => { initialised = false; };
```
eatyourgreens, line 54: `initialised` here should be declared in this function's scope. It won't change in the outer scope. See `ignored` in the [React data-fetching example](https://react.dev/reference/react/useEffect#fetching-data-with-effects).

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,97 @@
...
+  children,
+  workflowId
+}) {
+  let initialised = false;
+  const [apiData, setApiData] = useState({
+    workflow: null,
+    status: 'ready'
+  });
+
+  async function fetchWorkflow() {
```
eatyourgreens, line 23: This line declares a new `fetchWorkflow` function each time that the component renders. Be careful about that.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,97 @@
...
+  function update(data) {
+    console.log('+++ TODO');
+  }
+
+  // Wrap contextData in a memo so it doesn't re-create a new object on every render.
+  // See https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions
+  const contextData = useMemo(() => ({
+    workflow: apiData.workflow,
+    update
+  }), [apiData.workflow]);
```
eatyourgreens, line 69:
```suggestion
  }), [apiData.workflow, update]);
```

`update` is also an external dependency here. `update` is also declared on every render, which means your context data is a new object on every render.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -0,0 +1,97 @@
...
+    PropTypes.arrayOf(PropTypes.node),
+    PropTypes.node
+  ]),
+  workflowId: PropTypes.string
+};
+
+DataManager.defaultProps = {
+  children: null,
+  workflowId: ''
+};
```
eatyourgreens, line 95: `defaultProps` is deprecated for functional components. Use default args in the function declaration instead.

app/pages/lab-pages-editor/context.jsx
shaunanoordin: I'll revert this change

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: Noted, but for now this will use the classic PJC apiClient to keep in line with PFE. I'm still trying to isolate the the fetch functionality so the PJC can be replaced with zooniverse/panoptes-js if/when we eventually move this code to FEM

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: I purposely removed `initialised` in case it caused the useEffect hook to go into an infinite update loop. I'll try re-adding and test it out.

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: Noted, but some inefficiency is acceptable for now. I'm still considering how to organise the data fetch & _update_ functions. See Dev Notes above about "aborted useApiData() hook".

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: That's what I tried doing at the start, but ESLint seems to demand `DataManager.defaultProps`:

`propType "workflowId" is not required, but has no corresponding defaultProps declaration.eslint[react/require-default-props](https://github.com/jsx-eslint/eslint-plugin-react/tree/master/docs/rules/require-default-props.md)`

shaunanoordin: I've given up and wrapped `DataManager.propTypes` (which was giving that linting error) in `/* eslint-disable */`

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: I purposely removed `initialised` in case it caused the useEffect hook to go into an infinite update loop. I'll try re-adding and test it out.

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: Thanks - I'm aware that the Workflow resource has its own internal data store, which is why I'm experimenting with where this data is stored and the the associated fetch & update functions. 

That empty query argument is new, though - what on earth?

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: I've removed the `initialised` var altogether, to avoid additional complications at this stage.

My tests seem to confirm that the workflow fetch triggers only once, so that flag seems unnecessary. I'm not worried about a race condition on _fetching,_ but I will re-evaluate that [intialised/active/ignore](https://maxrozen.com/race-conditions-fetching-data-react-with-useeffect) variable when I'm working on _updating_ the workflow.

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: I've removed `initialised` altogether.

app/pages/lab-pages-editor/DataManager.jsx
shaunanoordin: Noted, I'll need to figure out where to house update() once I find the final home for the data fetch functionality.

app/pages/lab-pages-editor/DataManager.jsx
eatyourgreens: You’ve keyed this component by workflow ID in the pages editor, so it’s impossible for `workflowId` to change from one render to another. Each new workflow creates a new component.

shaunanoordin:
## PR Update

- `context.jsx` should have remained `context.js`
- Ran ESLint and fixed as many issues as possible. This was a small headache.
  - Clashing `defaultProps` lint has been ignored. Default props are now declared in the function arguments, not as `DataManager.defaultProps`.
- Workflow Fetch action (notably the Effect) has been heavily rejiggered, but maintains the same functionality
  - `initialised` flag now removed. 
  - Fetching action has been contained into its own `fetchWorkflow()` function. This is a candidate for extraction into its own helper function, assuming I'm OK with passing the 'saveApiData()' func as an arg. Hmm, thoughts for the next PR.
- Several lifecycle issues flagged for future work, notably that the update() function is being re-declared.

Other notes:

> I'd also recommend moving your data-fetching code into a custom hook. It can make components easier to read [(comment)](https://github.com/zooniverse/Panoptes-Front-End/pull/6787#pullrequestreview-1602396284)

Noted. This was attempted, but then aborted (see [Dev Notes](https://github.com/zooniverse/Panoptes-Front-End/pull/6787#issue-1871968175)) as I'm still experimenting with the data store for the Workflow resource. If I can compartmentalise the logic and abstract them into their own hooks or - at least - helper functions, I will.

app/pages/lab-pages-editor/DataManager.jsx
eatyourgreens: If it’s useful, take a look at the IIIF editor to see how it separates data-fetching state from editing state. You’ll probably run into the same problems that I did, so I suspect the solutions will be the same.

app/pages/lab-pages-editor/Tester.jsx
```diff
@@ -0,0 +1,30 @@
+/*
+eslint-disable no-console
+ */
+
+import React from 'react';
```
eatyourgreens, line 5:
```suggestion
```

app/pages/lab-pages-editor/DataManager.jsx
eatyourgreens: By the way, you can listen for changes to a resource too:
```js
workflow.listen('change', HANDLER)
```
https://github.com/zooniverse/panoptes-javascript-client/tree/master/lib/json-api-client#watching-a-resource-for-changes

That will fire whenever `workflow.update()` runs.

app/pages/lab-pages-editor/DataManager.jsx
eatyourgreens: Your code here will work, without React errors, if you declare update inside the `contextData` object:

```js
const contextData = useMemo(() => ({
    workflow: apiData.workflow,
    update: () => {
      console.log('update the workflow here')
    }
  }), [apiData.workflow]);
```

Call `contextData.update()`, rather than `update()`, in order to invoke it.

Otherwise, you will have to declare `update` as a dependency here, and likely also wrap `update` in `useCallback`. Like `useEffect`, you can't choose dependencies for `useMemo` or `useCallback`. You must declare all external reactive values used inside the hook. See the `login` function in the [useContext example code](https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions).

app/pages/lab-pages-editor/Tester.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+  return (
+    <div>
+      <h6>Tester</h6>
+      <div>
+        Workflow is called
+        {workflow?.display_name}
+      </div>
+      <form onSubmit={() => false}>
+        <input defaultValue={workflow?.display_name} onBlur={doUpdate} />
```
eatyourgreens, line 26: Accessibility error: this input has no label.

eatyourgreens:
Sorry, one other thought: wrapping this in `StrictMode` might help avoid data-fetching bugs by deliberately triggering them in development.

shaunanoordin:
👍 Sure thing. I've removed the React global (and disabled the corresponding eslint warnings), and I'll now wrap things in StrictMode.

PR #6795, opened 2023-09-04
Pages Editor: Save/Update Workflow

shaunanoordin:
## PR Overview

Follows #6787 

The aim of this PR is to allow the Pages Editor to update Workflow resources and save those changes to Panoptes.

The plan:
- The `DataManager` component should have an `update()` function 
  - [ ] The update() function must be passed to child components via context
  - [ ] Calling the update() function should update the values in the Workflow resource and save changes to Panoptes.
  - [ ] Child components should be notified of changes to the Workflow resource. (⚠️ the Panoptes Workflow object has its own 'memory', so keep that in mind.)
  - [ ] The DataManager should try to keep track of the update progress, to prevent multiple updates from overwriting each other. (see [PFE AutoSave's timed delay to saving data](https://github.com/zooniverse/Panoptes-Front-End/blob/1366d5c4900a7e69188687bc77bb0ea386075d57/app/components/auto-save.coffee#L65-L67))
  - ⚠️ OR better yet, updates should be queued _(Shaun's note: Not sure if this needs to be a stretch goal, or if this is already handled by [PJC's "save()"](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/json-api-client/resource.js#L40-L75) functionality)_
- For this PR the `Tester` UI component serves as a rudimentary read/write interface.

### Status

WIP

### DIFF
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -57,19 +57,36 @@ function DataManager({
     fetchWorkflow();
   }, [workflowId]);
 
-  /*
-  Updates the workflow with new data.
-   */
-  function update(data) {
-    console.log('+++ TODO');
-  }
-
   // Wrap contextData in a memo so it doesn't re-create a new object on every render.
   // See https://react.dev/reference/react/useContext#optimizing-re-renders-when-passing-objects-and-functions
-  const contextData = useMemo(() => ({
-    workflow: apiData.workflow,
-    update
-  }), [apiData.workflow]);
+  const contextData = useMemo(() => {
+    console.log('+++ DataManager.useMemo');
+
+    /*
+    Updates the workflow with new data.
+    */
+    async function update(data) {
+      console.log('+++ DataManager.update()');
+
+      setApiData({
+        workflow: apiData.workflow,
+        status: 'updating'
+      });
+
+      await new Promise((resolve) => setTimeout(resolve, 1000));
+
+      // Note to self: hang on, will setting setApiData() cause the useMemo to update perpetually?
+      setApiData({
+        workflow: apiData.workflow,
+        status: 'ready'
+      });
+    }
+
+    return {
+      workflow: apiData.workflow,
+      update
+    };
+  }, [apiData.workflow]); // Note to self: change this to workflowId?
 
   if (!workflowId) return (<div>ERROR: no Workflow ID specified</div>);
   // if (!workflow) return null
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -67,13 +67,15 @@ function DataManager({
     */
     async function update(data) {
       console.log('+++ DataManager.update()');
+      const wf = apiData.workflow;
+      if (!wf) return;
 
       setApiData({
-        workflow: apiData.workflow,
+        workflow: wf,
         status: 'updating'
       });
 
-      await new Promise((resolve) => setTimeout(resolve, 1000));
+      await wf.update(data).save();
 
       // Note to self: hang on, will setting setApiData() cause the useMemo to update perpetually?
       setApiData({
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -24,6 +24,8 @@ function DataManager({
     status: 'ready'
   });
 
+  const [randomFlagToPromptMemoUpdate, setRandomFlagToPromptMemoUpdate] = useState(0);
+
   // Fetch workflow when the component loads for the first time.
   // See notes about 'key' prop, to ensure states are reset:
   // https://react.dev/learn/you-might-not-need-an-effect#resetting-all-state-when-a-prop-changes
@@ -82,13 +84,23 @@ function DataManager({
         workflow: apiData.workflow,
         status: 'ready'
       });
+
+      // Hmmm... this seems to be the only thing that's updating the context properly.
+      // Without it, it seems that setApiData() isn't changing the value
+      // (object reference) of`apiData.workflow`, which means the useMemo
+      // doesn't realise that oohhh wait I think I get it now.
+      // TODO: how to indicate that the value of the apiData.workflow object
+      // has changed when the change is local to the resource itself?
+      // Does the workflow object have an internal version tracker?
+      // @shaunanoordin 20230902
+      setRandomFlagToPromptMemoUpdate(Math.floor(Math.random() * 10000));
     }
 
     return {
       workflow: apiData.workflow,
       update
     };
-  }, [apiData.workflow]); // Note to self: change this to workflowId?
+  }, [apiData.workflow, randomFlagToPromptMemoUpdate]);
 
   if (!workflowId) return (<div>ERROR: no Workflow ID specified</div>);
   // if (!workflow) return null
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -32,7 +32,6 @@ function DataManager({
   // https://react.dev/reference/react/useEffect#fetching-data-with-effects
   useEffect(() => {
     async function fetchWorkflow() {
-      console.log('+++ fetchWorkflow');
       try {
         setApiData({
           workflow: null,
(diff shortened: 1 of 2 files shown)


eatyourgreens:
>So a quick workaround to trigger useMemo() was to create a placeholder state (randomFlagToPromptMemoUpdate) that gets bumped after update() does its job. This works! But is it functional, or just hacky?

Total hack, I'm afraid. If you're working with the old Panoptes client, you have to listen to 'change' events in order to observe a resource for changes. Have a look at `.listen` and `.stopListening` here:
https://github.com/zooniverse/front-end-monorepo/blob/b3670c5a347f9c6e529b78afda8a13a4f30018fd/packages/lib-classifier/src/hooks/usePanoptesUser.js#L30-L48

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -57,19 +59,48 @@ function DataManager({
...
+      setApiData({
+        workflow: wf,
+        status: 'updating'
+      });
+
+      await wf.update(data).save();
+
+      // Note to self: hang on, will setting setApiData() cause the useMemo to update perpetually?
+      setApiData({
+        workflow: apiData.workflow,
```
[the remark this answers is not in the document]
eatyourgreens:
```suggestion
        workflow: newWorkflow,
```

eatyourgreens (commented):
Use a function to update loading state based on previous state.

shaunanoordin:
Thanks Jim! Here's a follow up to your suggestions

> ```
> setApiData(prevState => ({
>   ...prevState,
>   status: 'updating'
> });

👍 This is cleaner, I'll add this to the code.

> `const newWorkflow = await wf.update(data).save();`

Unfortunately, this didn't work - the newWorkflow and the wf are the same object.

```
const newWorkflow = await wf.update(data).save();
console.log('+++ are they the same? ', newWorkflow === wf);  // true
```

(Is the save() function [just returning 'this' on success?](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/json-api-client/resource.js#L71) I got a bit lost trying to follow the promise chain.)

> Total hack, I'm afraid. If you're working with the old Panoptes client, you have to listen to 'change' events in order to observe a resource for changes

Thanks Jim, I've taken a look at this. I had initially assumed that waiting for after `await wf.update().save()` would be sufficient to know when the context should be updated, but if the 'change' event is the more reliable trigger, I'll update the code accordingly.

(EDIT: _'on change'_ triggers when update() is called [(note 1)](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/json-api-client/resource.js#L35) (and when the resource is constructed?? [note 2](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/json-api-client/resource.js#L19)), while _'on save'_ triggers when save is successful? [(note 3)](https://github.com/zooniverse/panoptes-javascript-client/blob/master/lib/json-api-client/resource.js#L70) To Shaun: experiment and confirm. 'on change' should be sufficient since child components shouldn't need to wait until the changes are saved to Panoptes.)

My question about "is this hacky?" btw was in reference to using `setRandomFlagToPromptMemoUpdate()` to trigger useMemo.

eatyourgreens:
You're right `.update()` [updates the resource object in place](https://github.com/zooniverse/panoptes-javascript-client/blob/7ffa0471f4aff26184647816db4585a71ba21bea/lib/json-api-client/model.js#L79), rather than creating a new object. I thought that `.save()` might return a new workflow, since it fetches a new workflow from Panoptes.

I think the only way to detect that the stored resource has changed is to listen for `change` or `save` events.

Something else you could try is `apiClient.type('workflows').get(id, {})`, to disable internal caching in the client. That might force `.save()` to return  a new object. I'm not sure. The downside of that approach is possibly extra network traffic to Panoptes.

eatyourgreens:
Here's a thought: how about creating a new object yourself on save?

```js
const newWorkflow = await workflow.update(changes).save();

setApiData({
  workflow: { ...newWorkflow },
  status: 'ready'
});
```

I guess the problem with that is that it will remove any methods from the `workflow` object.

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -57,19 +59,50 @@ function DataManager({
...
+      // Does the workflow object have an internal version tracker?
+      // @shaunanoordin 20230902
+      // setRandomFlagToPromptMemoUpdate(Math.floor(Math.random() * 10000));
+    }
+
+    return {
+      workflow: apiData.workflow,
+      update
+    };
+  }, [apiData.workflow, randomFlagToPromptMemoUpdate]);
```
eatyourgreens, line 105:
```suggestion
  }, [apiData.workflow, apiData.status]);
```

This fixes the editor for me when I run it locally.

eatyourgreens: This might be a hack, but I think it’s cleaner than generating a random number on each state update.

shaunanoordin:
Oh, that's very useful - in [02de6aa](https://github.com/zooniverse/Panoptes-Front-End/pull/6795/commits/02de6aa2ada1ea1768fe53c3a1215118d635526a) I experimented with listening to _apiClient's_ change event, but that resulted in some odd triggers.

```
  useEffect(() => {
    function onWorkflowChange() { ... }
    apiClient.listen('change', onWorkflowChange)
    ...
  }, [apiClient]);

// would result in...

      const newWorkflow = await wf.update(data).save();

// ...that .save() triggering apiClient's change 6 times.
``` 

That oddness makes sense if I'm listening to the wrong thing. I'll try switching over to try listening to the _workflow_ object instead and observe.

eatyourgreens:
Also, that `payload` state does nothing in the project builder. It's only there to trigger a render. This was a replacement for the old `ChangeListener` code.

shaunanoordin:
> Also, that payload state does nothing in the project builder. It's only there to trigger a render. This was a replacement for the old ChangeListener code.

Ahh! OK, thank you for that explanation - my '...payload' was returning an empty array. OK, I'm going to compile today's experiment notes, give me a minute.

eatyourgreens:
Yeah, the trick is that `payload` is a new empty array after each update, so React will rerender even though it's empty.

shaunanoordin:
Follows previous [notes](https://github.com/zooniverse/Panoptes-Front-End/pull/6795#issuecomment-1705196680)

## PR Notes

After [987f1e0](https://github.com/zooniverse/Panoptes-Front-End/pull/6795/commits/987f1e0bbb1b92f0f9ac3f81fd267ac95b5c1c0e), I was looking for solutions to the problem of _how to monitor updates to the workflow, and subsequently trigger updates to the context._ I currently have three options

**Option A:** Listen to Status

Based on Jim's suggestion, this works.

```
const contextData = useMemo(() => {
  function update (data) {
    // manually change apiData.status on update
  }
  ...
}, [apiData.workflow, apiData.status])
```

- The hacky part here is that we're basically _manually_ triggering when the code should know the workflow has changed, instead of "organically" noticing the changes.

**Option B:** Listen to Workflow's Change Event

This method uses change listeners to listen to events baked into the Workflow resource.

```
const [updateCounter, setUpdateCounter] = useState(0)

useEffect(() => {
  function onWorkflowChange(data) {
    setUpdateCounter((uc) => uc + 1)
  }

  apiData.workflow?.listen('change', onWorkflowChange)
  return () => {
    apiData.workflow?.stopListening('change', onWorkflowChange)
  }
}, [apiData.workflow])

const contextData = useMemo(() => {
  ...
}, [api.workflowData, updateCounter])
```

- I still have to use a semi-hacky state (updateCounter) to trigger the context update via useMemo, but I'm coming around to accepting this.
  - This is still an improvement over randomFlagToTriggerMemo, though.
- Interestingly, the 'change' event seems to trigger more often than I initially expected.
  - it triggers _3x_ on the initial workflow fetch.
  - the line `wf.update(data).save()` triggers the event _4x._ (Note: save() performs a GET then a PUT)
  - (the line `wf.update(data)` alone correctly triggers the event only 1x.)
- This causes the useMemo() to trigger and update the context more often than expected (and hence trigger Tester re-renders)
  - ...but these wasted re-renders may still be acceptable.

**Option C:** Listen to Workflow's Save Event

As above, except change the event we're listening for.

```
useEffect(() => {
  function onWorkflowChange(data) {
    setUpdateCounter((uc) => uc + 1)
  }

  apiData.workflow?.listen('save', onWorkflowChange)
  return () => {
    apiData.workflow?.stopListening('save', onWorkflowChange)
  }
}, [apiData.workflow])
```

- This reduces the amount of re-renders, but it means the context isn't updated _until the workflow has successfully saved._
  - This is a major problem if the network is anything but lightning fast, so Option C is out. ❌ 

### Status

I'm in favour of Option B, so I'll clean up my code and get this PR ready for a review.

eatyourgreens:
Re. multiple updates firing, you could double check that the listener is only being attached once to the workflow. 

You could also compare with the listener in the main workflow editor. Maybe that also fires multiple times on save?

shaunanoordin:
## PR Update

Cleanup complete, this PR is now ready for review! 👌 

**Side note:** Jim, you were right to ask me to check the [main workflow editor.](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/workflow.jsx) Once I started monitoring when `.update()` and/or `.save()` triggered the 'change' events, it matched what I was seeing on DataManager.

eatyourgreens (approved):

PR #6822, opened 2023-09-12
Pages Editor: Basic Styling & Simple Workflow Settings Page

shaunanoordin:
## PR Overview

Part of: **Pages Editor MVP** project and **FEM Lab** super-project
Follows #6795
Staging Example: https://pr-6822.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging
Sean's designs: [on Figma](https://www.figma.com/file/gNOcFdlskfEXmVIFPvx7Yj/Pages-Editor?type=design&node-id=280-7689&mode=design)

The original intent of this PR is to add some **basic styling** and the **Workflow Settings Page,** but I had to scale back some ambition as even this stage had a lot going on.

Styling:
- added **lab-pages-editor.styl:** all our styling goes in here. It's gonna be a megalith.

Components:
- added PagesEditor's **PagesEditor:** the main component has been extracted out of index.jsx
- added **WorkflowSettingsPage**
  - 🅰️ this should follow HTML `<form>` standards
  - Changes are saved to Panoptes after _click_ and _blur_ events. (Not on _change_ events)
- reworked **WorkflowHeader**
  - not implemented: the Task and Workflow Settings buttons don't switch pages yet.

Other:
- Documentation: **README** updated with information on how we want PagesEditor to be as standalone as possible. 
- custom **icons** added: this will make replacing them with Grommet icons or etc in the future much easier.
- all text organised into **strings.json**

Minor stuff:
- re-modified **router.jsx:** OK, fine, lets use the standardised capital "ID" in workflowID
- modified **main.styl** 🍐 

🍐 indicates code that sits outside the `/lab-pages-editor` folder.
🅰️ indicate components that should have accessibility checks/reviews. Please flag any issues noted.

### Dev Notes

My initial idea of "just create the Workflow Config Page, it'll be easy!" has run arse-first into the reality that adding Subject Sets to Workflows can be a bit complex. (see [Subject Set Linker](https://github.com/zooniverse/Panoptes-Front-End/blob/master/app/pages/lab/workflow-components/subject-set-linker.jsx)) Probably linking Tutorials too.

Anyway,
- the PagesEditor needs to figure out which `project` it's editing, which means that variable either needs to be passed down from the parent (is there a parent?) or the DataManager needs a fetchProject() function and share that on the context. HMMM.
- but that's for a separate PR.
- so for now, the Workflow Settings Page in this PR will be simpler than originally intended: it's everything you see on Sean's design, _minus Tutorials and Subject Sets._

image

[...]

PR #6836, opened 2023-09-19
Pages Editor: improve Workflow Header with Tabs pattern

shaunanoordin:
## PR Overview

Follows #6822
Staging Example: https://pr-6836.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging
Tabs Pattern: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/

This PR adjusts the Workflow Header so that the navigation/paging tabs actually behaves like the ARIA-specified Tabs patten. This follows [feedback](https://github.com/zooniverse/Panoptes-Front-End/pull/6822#pullrequestreview-1630968196) from the previous PR.

- Elements should now have `tablist`, `tab`, and `tabpanel` roles
- Mouse: clicking on Task or Workflow Settings should switch to the appropriate panel (or placeholder panel)
- Keyboard:
  - tab-navigation (pressing the Tab key on a keyboard) should only allow users to focus on the active/selected tab button on the tabs list (e.g. so if the Task tab button is active, pressing the Tab key _shouldn't_ take you to the Workflow Settings tab button, but instead to the next focusable element on the page.)
  - _when_ the user is focused on the active/selected tab, then the left/right keys should switch between the tabs (similar to clicking on 'em). Note that you don't need to press Enter or Space to confirm that you want to switch tabs.

### Testing

Go to the staging example URL, and confirm that the "Tasks" and "Workflow Settings" tabs work as described by the ARIA Tabs pattern.

### Status

Ready for review!

### DIFF
--- a/app/pages/lab-pages-editor/components/WorkflowHeader.jsx
+++ b/app/pages/lab-pages-editor/components/WorkflowHeader.jsx
@@ -11,7 +11,7 @@ import { useWorkflowContext } from '../context.js';
 import strings from '../strings.json';
 
 export default function WorkflowHeader({
-  currentTab = 1000,
+  currentTab = 0,
   projectId = '',
   setCurrentTab = () => {}
 }) {
@@ -27,9 +27,27 @@ export default function WorkflowHeader({
     }
   ];
 
+  // When clicking a tab button, make that tab active. This is pretty straightforward.
   function onClick(e) {
     const { tab } = e.target.dataset;
     setCurrentTab(parseInt(tab));
+  }
+
+  // When a tab button has focus, the left/right keys will switch to the prev/next tab.
+  function onKeyUp(e) {
+    let changeTab = 0;
+    if (e.key === 'ArrowLeft') changeTab = -1;
+    if (e.key === 'ArrowRight') changeTab = +1;
+
+    if (changeTab !== 0) {
+      const newTab = (currentTab + changeTab + tabs.length) % tabs.length;
+      setCurrentTab(newTab);
+      e.preventDefault?.();
+      e.stopPropagation?.();
+      return false;
+    }
+
+    return true;
   }
 
   if (!workflow) return null;
@@ -51,6 +69,7 @@ export default function WorkflowHeader({
             key={`${tab.id}`}
             label={tab.label}
             onClick={onClick}
+            onKeyUp={onKeyUp}
             selected={(currentTab === index)}
           />
         ))}
@@ -70,6 +89,7 @@ function TabButton({
   index,
   label = '',
   onClick = () => {},
+  onKeyUp = () => {},
   selected = false
 }) {
   return (
@@ -79,6 +99,7 @@ function TabButton({
       data-tab={index}
       id={id}
       onClick={onClick}
+      onKeyUp={onKeyUp}
       role="tab"
       type="button"
     >
@@ -92,5 +113,6 @@ TabButton.propTypes = {
   index: PropTypes.number,
   label: PropTypes.string,
   onClick: PropTypes.func,
+  onKeyUp: PropTypes.func,
   selected: PropTypes.bool
 };
--- a/app/pages/lab-pages-editor/components/WorkflowHeader.jsx
+++ b/app/pages/lab-pages-editor/components/WorkflowHeader.jsx
@@ -13,19 +13,11 @@ import strings from '../strings.json';
 export default function WorkflowHeader({
   currentTab = 0,
   projectId = '',
-  setCurrentTab = () => {}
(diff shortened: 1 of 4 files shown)


app/pages/lab-pages-editor/components/WorkflowHeader.jsx
```diff
@@ -27,18 +60,62 @@ export default function WorkflowHeader({
...
+  index,
+  label = '',
+  onClick = () => {},
+  onKeyUp = () => {},
+  selected = false
+}) {
+  return (
+    <button
+      aria-selected={selected}
+      className={selected ? 'selected' : 'unselected'}
```
eatyourgreens, line 100: You can use attribute selectors to style an element based on its ARIA state eg.

```css
button[role=tab][aria-selected=true]
```

This has the advantage of making sure that your visual style represents an element’s state in the accessibility tree.

app/pages/lab-pages-editor/components/WorkflowSettingsPage.jsx
```diff
@@ -47,7 +47,12 @@ export default function WorkflowSettingsPage() {
   if (!workflow) return null;
 
   return (
-    <form className="workflow-settings-page" onSubmit={onSubmit}>
+    <form
+      aria-labelledby="pages-editor_workflow-header-tab-button_settings"
+      className="workflow-settings-page"
+      onSubmit={onSubmit}
+      role="tabpanel"
```
eatyourgreens, line 54: This breaks the form by overriding its default form role.

app/pages/lab-pages-editor/components/WorkflowSettingsPage.jsx
```diff
@@ -47,7 +47,12 @@ export default function WorkflowSettingsPage() {
   if (!workflow) return null;
 
   return (
-    <form className="workflow-settings-page" onSubmit={onSubmit}>
+    <form
+      aria-labelledby="pages-editor_workflow-header-tab-button_settings"
```
eatyourgreens, line 51: These IDs need to be maintained in a couple of different places, which makes the code brittle. You can make the component more flexible by passing IDs as component props.

eatyourgreens (changes requested):
Some initial thoughts on making the code more flexible. 

Don’t change the role of a form. Instead, wrap it in a tab panel.

app/pages/lab-pages-editor/components/WorkflowSettingsPage.jsx
shaunanoordin: Oh dang, that's good point - I'll fix this right away.

app/pages/lab-pages-editor/components/WorkflowHeader.jsx
```diff
@@ -27,18 +52,67 @@ export default function WorkflowHeader({
...
+  id,
+  index,
+  label = '',
+  onClick = () => {},
+  onKeyUp = () => {},
+  selected = false
+}) {
+  return (
+    <button
+      aria-selected={selected}
```
eatyourgreens, line 97: You can add `aria-controls=TAB-PANEL-ID` to link each tab to its corresponding tab panel.

app/pages/lab-pages-editor/components/WorkflowHeader.jsx
```diff
@@ -10,13 +11,37 @@ import { useWorkflowContext } from '../context.js';
...
 }) {
   const { workflow } = useWorkflowContext();
   const returnUrl = `/lab/${projectId}/workflows`;
 
-  function onClick() {
-    console.log('TODO');
+  // When clicking a tab button, make that tab active. This is pretty straightforward.
+  function onClick(e) {
+    const { tab } = e.target.dataset;
+    setCurrentTab(parseInt(tab));
```
eatyourgreens, line 25: Setting the state of component A from component B isn't recommended in React. Maybe have an `onChange` callback that runs when the state of `WorkflowHeader` changes?

eatyourgreens (approved):
Couple of minor comments to decouple `WorkflowHeader` state from the overall state of the page. Nothing blocking.

app/pages/lab-pages-editor/components/WorkflowHeader.jsx
eatyourgreens, line 97:
```suggestion
      aria-selected={selected ? 'true' : 'false'}
```

eatyourgreens:
Spotted a couple of small attribute errors after leaving that review. Nothing major.

shaunanoordin:
Thanks, Jim! 👍

PR #3711, opened 2017-04-10
Replace slideshow buttons with radio buttons

eatyourgreens:
Fixes #3514.

Describe your changes.
Adds keyboard support by replacing buttons with named radio inputs, which can be controlled using the cursor keys.

# Review Checklist

- [x] Does it work in all major browsers: Firefox, Chrome, Edge, Safari?
- [x] Does it work on mobile?
- [x] Can you `rm -rf node_modules/ && npm install` and app works as expected?
- [x] Did you deploy a staging branch? https://subject-viewer-pips.pfe-preview.zooniverse.org/
- [ ] Did you get any type checking errors from [Babel Typecheck](https://github.com/codemix/babel-plugin-typecheck)

## Optional

- [ ] If it's a new component, is it in ES6? Is it clear of warnings from ESLint?
- [ ] Have you replaced any `ChangeListener` or `PromiseRenderer` components with code that updates component state?
- [ ] If changes are made to the classifier, does the dev classifier still work?
- [ ] Have you added in [flow type annotations](https://flowtype.org/docs/type-annotations.html)?

PR #6868, opened 2023-10-07
Pages Editor: add Tasks page (UI only, non-functional)

shaunanoordin:
## PR Overview

Follow #6867
Staging branch URL: https://pr-???.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711

This PR adds the Tasks page to the Pages Editor, which _should have_ allowed project owners to add and edit tasks. The original scope of this PR was much larger, with an attempt to make the _add Tasks_ button actually working, but I've stripped that out due to being clunky (see Known Dev Flaw 1 below).

- This PR is now primarily a UI/design-only update, with a focus on making sure the HTML works correctly.
- Comments on accessibility very welcome
  - I've tested with Mac VoiceOver but I'm not sure how a vision-impaired project owner would actually experience the page. e.g. the Edit icon has an aria label that says "Edit Task T1" etc but I couldn't get the voiceover to read the _task instructions_ or _task key (e.g. T0)_ which is the core of the whole thing.
  - Maybe I need to make the `<li.task-item>` focusable?

⚠️ LIMITATIONS
- **Known dev flaw 1:** the "steps list" isn't working AT ALL. It's supposed to _list steps in workflow, and list task in each step,_ but instead at the moment it's _listing all tasks in the workflow, assuming each task is one step._ Blarp!
  - I want to address this in the next PR, when I have to really focus on how "Add New Task" should also 1. add a Step (Page), and 2. add the newly created Task into the Step (Page).
  - Follow by answering the question _how do you delete Steps (Pages)?_
- **Design feedback 1:** the grip-bar (or drag handle, which is a better name that just occurred to me goshdangit) is insufficient for its task of re-arranging steps. An "move up/move down" button could be implemented - I'll need to discuss with Sean about this.
  - (**Design feedback 2:** also to confirm with Sean, grip-bar/drag handle is for rearranging _steps..._ and rearranging _tasks_ should be done in the Edit Single Step modal... but what if user is interested in dragging one Task to another Step? Assume answer is "this isn't an option".)

image

### Status

Ready for review


eatyourgreens:
Some design feedback. The only icon here that's styled as a button is the T0, which isn't a button. How do I know, visually, which elements I can click (or tap)?

Visual icons that represent a task key and task type in the editor, along with SVG icon buttons to edit the task.

eatyourgreens:
> but I couldn't get the voiceover to read the task instructions or task key (e.g. T0) which is the core of the whole thing.

The list looks fine to me. Were you navigating the page in VoiceOver, or just tabbing through focusable elements? In the first case, VO will read through the list. In the second case, it won't, just as it will be skipped for a sighted keyboard user.

eatyourgreens:
WCAG 2.2 came out on Thursday, with a more restrictive definition of accessible drag-and-drop. TPGi has [a good summary](https://www.tpgi.com/new-success-criteria-in-wcag22/#dragging-movements).

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,99 @@
...
+          </button>
+          {/* Dev observation: the <select> should have some label to indicate it's for choosing the starting task. */}
+          <select
+            aria-label="Choose starting test"
+            className="flex-item"
+          >
+            <option disabled={true}>Choose starting test</option>
+          </select>
+        </div>
+        <ul className="steps-list">
```
eatyourgreens, line 48: All HTML landmark elements can be labelled, so you can label this list with either `aria-label` or `aria-labelledby`. 

Here's an example of a list with a linked heading, from project Team pages. 

https://github.com/zooniverse/front-end-monorepo/blob/76d0d94d2aa348b3a5ded97b76cff050ef67d3c3/packages/app-project/src/screens/ProjectAboutPage/ProjectAboutPage.js#L100-L119

eatyourgreens: [HTML landmarks](https://developer.mozilla.org/en-US/blog/aria-accessibility-html-landmark-roles/) can also be labelled, which can be useful for navigating complex pages, like these project builder pages. 

Use ctrl-opt-U to open the rotor in VoiceOver, then navigate to Landmarks to see a navigable list of page landmarks.

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,99 @@
...
+
+  return (
+    <div className="tasks-page">
+      <div className="workflow-title flex-row">
+        <h2 className="flex-item">{workflow?.display_name}</h2>
+        <span className="workflow-id">{`#${workflow?.id}`}</span>
+        {(isActive) ? <span className="status-active">Active</span> : <span className="status-inactive">Inactive</span>}
+      </div>
+      <section>
+        <h3>Tasks</h3>
```
eatyourgreens, line 31:
```suggestion
      <section aria-labelledby='workflow-tasks-heading'>
        <h3 id='workflow-tasks-heading'>Tasks</h3>
```

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,99 @@
...
+      <ul className="tasks-list">
+        <li className="task-item">
+          <div className="flex-row spacing-bottom-M">
+            <span className="task-key">{taskKey}</span>
+            <span className="task-icon">
+              {/* TODO: change icon and aria label depending on task type*/}
+              <i
+                aria-label="Task type: text"
+                className="fa fa fa-file-text-o fa-fw"
+              />
```
eatyourgreens, line 82: I don't think `<i>` can be labelled. Only elements with roles, like images, can be labelled. 

Does VoiceOver read this label? Does it show up in the accessibility tree?

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,99 @@
...
+function StepItem({ task, taskKey }) {
+  if (!task || !taskKey) return null;
+
+  const text = task.instruction;
+
+  return (
+    <li className="step-item">
+      <div className="grip-bar flex-row spacing-bottom-XS">
+        <GripIcon color="#A6A7A9" />
+      </div>
```
eatyourgreens, line 72: This should be a button.

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,99 @@
...
+      </section>
+    </div>
+  );
+}
+
+// WARNING/TODO: this should be handling steps, not tasks
+function StepItem({ task, taskKey }) {
+  if (!task || !taskKey) return null;
+
+  const text = task.instruction;
```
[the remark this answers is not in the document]
eatyourgreens: The APIs for individual tasks are also inconsistent. Some use `task.instruction`, others use `task.question`. This has been a pain point in the past.

app/pages/lab-pages-editor/components/TasksPage.jsx
eatyourgreens: You could add a keyboard handler to this button, so that the cursor keys move items up and down in the list, when the drag handle is focused.

eatyourgreens: Alternatively, make the whole `StepItem` a focusable element that responds to the cursor keys. That would be trickier, since each step item contains buttons and form inputs that also need to respond to the pointer and keyboard. 

An example of how that can go wrong is https://github.com/zooniverse/front-end-monorepo/issues/5422.

app/pages/lab-pages-editor/components/TasksPage.jsx
eatyourgreens: The browser doesn’t programmatically link sections to their headings, so you have to link them up in your code.

app/pages/lab-pages-editor/components/TasksPage.jsx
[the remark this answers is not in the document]
eatyourgreens: `workflow.strings['tasks.T0.instruction']` is how we get the instruction text now, but that would involve updating the lab to load workflow strings.

app/pages/lab-pages-editor/components/TasksPage.jsx
eatyourgreens: Font Awesome recommends `<span>` nowadays. Spans can't be labelled either, so I'm not sure what they mean by 'Accessibility' here:

https://fontawesome.com/docs/web/add-icons/how-to

eatyourgreens:
Here's a screen recording where I navigate through the page in VoiceOver on Safari, using the ctrl-opt-right arrow. Safari has better integration with VoiceOver than Chrome or Firefox, so results may differ in those browsers. I tried to go slowly, so that you can follow what I'm doing. A real user would zip through the page, and would have the speed of the reader set to 500% or more.


https://github.com/zooniverse/Panoptes-Front-End/assets/59547/319bac0f-4a1e-4702-afaa-eddc953151d3

A couple of observations:
- Some of the text doesn't make a lot of sense without any context eg. '#3711' or 'T0'.
- The text inputs aren't labelled at all. When  enter them I hear 'here' (the last word in the field) then an instruction to type without knowing that the field is already filled out. Labelling text inputs with a placeholder value, like this, is both confusing UX and very bad practice.

eatyourgreens:
Here's a second video, where I bring up the rotor (ctrl-opt-U) then use that to list all the headings, links, from controls and landmarks on the page. Finally, I select the 'main' landmark and jump to it. Hopefully this is easy to follow but ask if you have any questions.


https://github.com/zooniverse/Panoptes-Front-End/assets/59547/cfcd8315-a285-4555-984b-12bfc27450a2

app/pages/lab-pages-editor/components/TasksPage.jsx
shaunanoordin: I didn't realise this! Thanks Jim, this suggestion has now been committed.

app/pages/lab-pages-editor/components/TasksPage.jsx
shaunanoordin: Noted; (1) I'll read up on the MDN article, and (2) add labels to the lists.

Thanks again - I wasn't aware that lists could/should be labelled.

app/pages/lab-pages-editor/components/TasksPage.jsx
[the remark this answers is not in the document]
shaunanoordin: 👌 Noted, I'll add Panoptes translations to the ToDo. At the very minimum for this PR though, I should at least make that `const text = task.instruction || task.question || '???'` so we can see more task types when reviewing/testing.

app/pages/lab-pages-editor/components/TasksPage.jsx
eatyourgreens: Depends how complex the page is, to be honest. Sometimes, adding extra labels to lists can make them more confusing. 

Here, labelling the list as ‘edit tasks’ or ‘order tasks’ might help explain the purpose of the list.

app/pages/lab-pages-editor/components/TasksPage.jsx
shaunanoordin: I ran some tests, and on macOS + Chrome:
- mac VoiceOver seems to detect _"Task type: text"_ label
- The Accessibility panel sees this as a _generic_ element. (Hmm, thinking I should add role=img for this)

image

Noted that a `<span>` may make more sense.

shaunanoordin: All said though, I'm not too fond of going back to FontAwesome - I dislike how they're gating a lot of icons behind a subscription.

shaunanoordin: Testing update:
- `<i aria-label="Task type: text" className="fa fa-whatever" />` and `<span aria-label="Task type: text" className="fa fa-whatever" />` appear to be functionally the same.
  - Oh, and one thing I missed - the voiceover calls it "Task type: text, EMPTY GROUP"
- Adding `role="img"` sets this to "Task type: text, image"
- Adding `role="presentation"` causes the task key and task question to be read together, e.g. "T0 This is the first task"... which actually sounds good, but misses out on the task type information.

shaunanoordin:
Awesome, thanks Jim! 👍 Those video walkthroughs have been very helpful - I now realise my problem was that I didn't know how to navigate properly. I was _tab-navigating_ so I ended up only seeing the focusable elements.

Some follow up from your feedback:

> The only icon here that's styled as a button is the T0, which isn't a button. How do I know, visually, which elements I can click (or tap)?

Noted, will bring this up in the design review with Sean this week.

> WCAG 2.2 came out on Thursday, with a more restrictive definition of accessible drag-and-drop

Noted, Sean already proposed a new design for up/down arrows.

image

> HTML lists can be labelled, so you can label this list with either aria-label or aria-labelledby.

👍 Steps/Pages list and Tasks list now have labels.

> Future work: task.instruction is deprecated in favour of the Panoptes translations API.

Added this to the TODO

> I don't think <i> can be labelled. Only elements with roles, like images, can be labelled.

As noted in the conversation, I'm going to change `<i class=fa ...>` to `<span role=img class=fa ...>`

Update: this seems to work! But I'm going to look at this again ~~tomorrow~~ Thursday morning to make sure I'm not abusing the use of [role=img](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/img_role)

> Some of the text doesn't make a lot of sense without any context eg. #3711 or 'T0'.

Noted, I'll look into this; I think I'll bring it up on the design revision. I assume the assumption here is that Project Owners would understand those numbers/labels from experience... but let me check that, since that's two levels of _assuming_ on my part.

> The text inputs aren't labelled at all. When enter them I hear 'here' (the last word in the field) then an instruction to type without knowing that the field is already filled out. Labelling text inputs with a placeholder value, like this, is both confusing UX and very bad practice.

Noted; this is just a placeholder to test out the current design of a Text Task. (A pretty wonky placeholder at that - I think I should set it to disabled.) I have a feeling this will need to be revised further, as they're _not_ supposed to be actual/functional `<input>` elements, but just _presentational elements_ to indicate to the project owner what kind of Task they just made. 🤔

app/pages/lab-pages-editor/components/TasksPage.jsx
eatyourgreens: Generic elements can't be labelled and won't be read out by JAWS or NVDA, the two biggest screen readers. I'm pretty sure that's in the ARIA spec, so VO is ignoring the standard here. 

This relates to what I was saying about lists. Basically, labels are valid on elements that have a role.

> Because the generic role is nameless, the aria-labelledby and aria-label attributes are prohibited. Because the role is generic, the aria-roledescription and aria-brailleroledescription attributes are also prohibited.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/generic_role

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+            type="button"
+          >
+            Add a new Task +
+          </button>
+          {/* Dev observation: the <select> should have some label to indicate it's for choosing the starting task. */}
+          <select
+            aria-label="Choose starting test"
+            className="flex-item"
+          >
+            <option disabled={true}>Choose starting test</option>
```
eatyourgreens, line 45:
```suggestion
            <option disabled >Choose starting test</option>
```
You don't specify the `true` or `false` for boolean props in React.

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+          </select>
+        </div>
+        <ul className="steps-list" aria-label="List of Pages/Steps">
```
eatyourgreens, line 48: One small thing: `ul` is a list, and will be announced as a list, so 'List of...' in the label is redundant, similar to how you wouldn't write 'Link to...' in the text for a link.

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+  return (
+    <li className="step-item">
+      <div className="grab-handle flex-row spacing-bottom-XS">
+        <button aria-label={`Rearrange Page ${stepKey} upwards`} className="plain" type="button">
+          <span className="fa fa-caret-up" role="img" />
+        </button>
+        <GripIcon color="#A6A7A9" /> {/* TODO: add drag/drop functionality. Perhaps this needs to be wider, too. */}
+        <button aria-label={`Rearrange Page/Step ${stepKey} downwards`} className="plain" type="button">
+          <span className="fa fa-caret-down" role="img" />
+        </button>
```
eatyourgreens, line 81:
```suggestion
        <button aria-label={`Rearrange Page ${stepKey} upwards`} className="plain" type="button">
          <span className="fa fa-caret-up" />
        </button>
        <GripIcon color="#A6A7A9" /> {/* TODO: add drag/drop functionality. Perhaps this needs to be wider, too. */}
        <button aria-label={`Rearrange Page/Step ${stepKey} downwards`} className="plain" type="button">
          <span className="fa fa-caret-down" />
        </button>
```

The buttons have labels, so re-defining their contents as images might actually be confusing here.

app/pages/lab-pages-editor/components/TasksPage.jsx
```diff
@@ -0,0 +1,113 @@
...
+            <span className="task-text flex-item">{text}</span>
+            <button aria-label={`Delete Page/Step ${stepKey}`} className="plain" type="button">
+              <span className="fa fa-trash" role="img" />
+            </button>
+            <button aria-label={`Copy Page/Step ${stepKey}`} className="plain" type="button">
+              <span className="fa fa-copy" role="img" />
+            </button>
+            <button aria-label={`Edit Page/Step ${stepKey}`} className="plain" type="button">
+              <span className="fa fa-pencil" role="img" />
+            </button>
```
eatyourgreens, line 104:
```suggestion
            <button aria-label={`Delete Page/Step ${stepKey}`} className="plain" type="button">
              <span className="fa fa-trash" />
            </button>
            <button aria-label={`Copy Page/Step ${stepKey}`} className="plain" type="button">
              <span className="fa fa-copy" />
            </button>
            <button aria-label={`Edit Page/Step ${stepKey}`} className="plain" type="button">
              <span className="fa fa-pencil" />
            </button>
```

I think this might be an example of ['too much accessibility'](https://webaim.org/blog/overly-accessible/), where well-meaning developers add extra code that actually makes the page harder to use.

The [first rule of ARIA](https://www.w3.org/TR/using-aria/#rule1) is basically "don't use ARIA" so be sparing with roles and states. Add them when a task can't be performed without them, or without using native HTML elements and attributes.

eatyourgreens: By the way, if a button is labelled with an image, eg. an SVG icon, then the appropriate pattern is to label the image:
```html
<button><img alt='This is the button label' src='./icon.svg'></button>
```

eatyourgreens (approved):

Issue #6765, opened 2023-08-08
FEM Lab: add Simple Dropdown task editor


lcjohnso:
## Expected behavior
I can configure the FEM "Simple Dropdown" task (a simplified version of the PFE Dropdown task, lacking multiple cascading dropdowns or accepting user input) from the FEMLab version of the project builder workflow editor page.

## Current behavior
The FEMLab workflow editor page still shows full PFE dropdown task editor, including features that are not implemented in FEM.

## Additional information
We may need to take care when deploying this change with regard to potentially affected projects.  Example: will NfN still be able to make necessary edits?

PR #6872, opened 2023-10-10
Limit FEMLab Dropdown to only allow between 4-200 options.

kieftrav:
Staging branch URL: https://pr-6872.pfe-preview.zooniverse.org

# Overview
Limit the SimpleDropdown within FEMLab to between 4-200 options. 

Toward https://github.com/zooniverse/Panoptes-Front-End/issues/6765

[PFE without FEMLab enabled for a Dropdown](https://pr-6872.pfe-preview.zooniverse.org/lab/1996/workflows/3771?env=staging):
- [ ] Should not limit options to greater than 4
- [ ] Should not limit options to less than 200

[PFE with FEMLab enabled for a Dropdown should](https://pr-6872.pfe-preview.zooniverse.org/lab/1995/workflows/3770?env=staging):
- [ ] Should limit options to greater than 4
- [ ] Should limit options to less than 200

# Review Checklist
- [ ] Does it work on mobile?

### DIFF
--- a/app/classifier/tasks/dropdown/dropdown-dialog.cjsx
+++ b/app/classifier/tasks/dropdown/dropdown-dialog.cjsx
@@ -10,6 +10,9 @@ provincesCanada = require './presets/provinces-Canada' # value = two-letter post
 statesMexico = require './presets/states-Mexico' # value = three-letter ISO 3166-2 abbreviation
 # IMPORTANT: before adding preset options, confirm values are not duplicated in existing presets
 
+OPTIONS_MIN = 4
+OPTIONS_MAX = 250
+
 DropdownDialog = createReactClass
 
   getDefaultProps: ->
@@ -150,6 +153,13 @@ DropdownDialog = createReactClass
     if not @state.editSelect.title
       return window.alert('Dropdowns must have a Title.')
 
+    # for FEMLab we want to enforce a min and max number of options
+    numOptions = @state.editSelect.options['*'].length
+    if @props.pfeLab is false and numOptions < OPTIONS_MIN
+      return window.alert('Dropdowns must have at least 4 options.')
+    else if @props.pfeLab is false and numOptions > OPTIONS_MAX
+      return window.alert('Dropdowns must have fewer than 200 options.')
+
     selectTitles = @props.selects
       .filter (select) => select isnt @props.initialSelect
       .map (select) -> select.title
@@ -196,6 +206,9 @@ DropdownDialog = createReactClass
         {if select.condition?
           <span>Dependent on {@state.conditionalSelects[@state.conditionalSelects.length - 1]?.title}</span>}
       </p>
+      {if @props.pfeLab is false # for FEMLab we want to indicate a minimum number of options
+        <p><i>Enter a minimum of 4 choices</i></p>
+      }
 
       <label className="pill-button" title={dropdownEditorHelp.required}>
         Required <input type="checkbox" ref="required" checked={select.required} onChange={@editSelect}></input>


kieftrav:
The notification and enforcement of the limitation for FEMLab shows a standard window alert because that is what the current code validation does for a missing dropdown title. My preference is to follow the existing code style of using the window browser alert to create a consistent UI experience (acknowledging we all recognize it as sub-optimal). Attached are screenshots of the UI as discussed with @seanmiller26 and @shaunanoordin. If we'd like the UI to diverge from the pre-existing pattern we can do that with a relatively small amount of coding.

Test-display-4-option-limit
Existing-missing-title-alert
Less-than-4-alert
Greater-than-200-alert

seanmiller26:
Considering that we want a consistent UI experience, this LGTM.

eatyourgreens:
There’s an `Alert` component, which probably can be styled, but it’s probably less accessible than `window.alert`. 
https://github.com/zooniverse/Panoptes-Front-End/blob/ebfa8b33ef6ed8faa3ecf621c1d6ea5e8d3449d1/app/lib/alert.jsx

app/classifier/tasks/dropdown/dropdown-dialog.cjsx
```diff
@@ -150,6 +150,13 @@ DropdownDialog = createReactClass
     if not @state.editSelect.title
       return window.alert('Dropdowns must have a Title.')
 
+    # for FEMLab we want to enforce a minimum of 4 options and a maximum of 200
+    numOptions = @state.editSelect.options['*'].length
+    if @props.pfeLab is false and numOptions < 4
+      return window.alert('Dropdowns must have at least 4 options.')
+    else if @props.pfeLab is false and numOptions > 200
```
shaunanoordin, line 157: [suggestion] I'd abstract these numbers into `const MINIMUM_FEM_DROPDOWN_OPTIONS = 4` or etc

shaunanoordin (approved):
## PR Review

Affects: Dropdown Tasks editor (FEM Lab only)

This PR modifies the Dropdown Tasks editor, in the Project Builder (FEM Lab ver), so that it only accepts dropdown lists with a _minimum of 4_ and a _maximum of 200_ options.

- When the dropdown list modal is open, it will reject any attempt to _Save_ changes if the number of options doesn't meet the criteria.
  - In the screenshot below, after adding 3 items to the list, a `window.alert` appears when clicking 'Save', preventing the list of 3 items from being saved. As a project owner, I either need to add at least 1 more item, or press 'Cancel' to back out of making changes.
      image
- ⚠️ NOTE 1: it's still entirely possible to create a dropdown with **0 options,** by simply creating a Dropdown Task _and never editing the dropdown list._ However, I'm thinking that this is a non-breaking edge case that only happens when a project owner didn't actually complete their workflow to begin with, so it's an "OK, I guess?" from me.
- ⚠️ NOTE 2: when editing the dropdown list, the "Countries" preset will create **249 options,** which means the modal can't be closed/saved.

Functionality tests & code changes look pretty good, though I did add a small suggestion about wrapping those number (4, 200) into some variables for easier maintenance in the future.

### Testing

Tested with macOS + Chrome on staging URL, with a typical test project

Base functionality, in PFE Lab [(?pfeLab=true)](https://pr-6872.pfe-preview.zooniverse.org/lab/1966/workflows/3773?env=staging&pfeLab=true):
- User can create & delete Dropdown Tasks. (Assuming dropdown task experimental tool is enabled for project)
  - Dropdown Tasks are created with 1 dropdown list ("Main Dropdown") with 0 options.
- User can make changes to the Dropdown Task's main questions/help text/etc, and changes are saved immediately.
- User can edit dropdown lists:
  - they can change the list's title, required flag, etc
  - they can add/delete/rearrange options in the list
  - they can use **Presets** to quickly fill the list of options.
- 🍏   Dropdown lists require a manual save process (clicking "Save") to confirm changes.
  - "Save" only works if dropdown list is valid, meaning title can't be empty.
  - Clicking "cancel" reverts any changes.

Additional functionality, in FEM Lab [(?femLab=true)](https://pr-6872.pfe-preview.zooniverse.org/lab/1966/workflows/3773?env=staging&femLab=true)
- 🍎  Dropdown lists require a manual save process (clicking "Save") to confirm changes.
  - "Save" only works if dropdown list is valid, meaning title can't be empty _AND number of options must be between 4 and 200 (inclusive)._
  - Clicking "cancel" reverts any changes.

### Status

LGTM, but I have two _non-blocking_ requests before you merge this.

- First, please consider the small suggestion of turning those 4/200 numbers into more manageable vars/consts.
- Second, please check with Cliff(? Sam?) as to why the maximum number of 200, since it clashes with the 249 countries preset.
  - If 200 is a hard limit, then I'd recommend removing/hiding the countries preset from the presets list, when femLab=true
  - If 200 is just an arbitrary number, maybe crank that up to 250?

PR #6888, opened 2023-10-20
Pages Editor: implement basic "Add Tasks" functionality

shaunanoordin:
## PR Overview

Follow #6868
Staging branch URL: https://pr-6888.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR implements the functionality of the _"Add a New Task +"_ button. Clicking the button will: 1. create a new empty **(Text)** Task, 2. create a new Step/Page then add that Task to it, and 3. add the Task + Step/Page to the Workflow resource (i.e. save to Panoptes)

- Tasks are created with Task Keys like T0, T1, T2, etc.
- Steps/Pages are created with Task Keys like P0, P1, P2, etc.
- Tasks will be visible (listed) on the Tasks page.
- Temporary Debug code: "Reset" button added. 
- ⚠️ UI issue: the Steps List is actually only _listing Tasks, not Steps with Tasks._ This won't be relevant when running the "Add Tasks" functionality, but if you're looking at a Workflow with existing multi-Task Steps, you'll notice that everything looks wonky.
- ⚠️ Data issue: each Task created doesn't have a `task.next`, so if you try to preview the Workflow on e.g. frontend.preview, you'll notice that the Tasks don't link to each other.

### Testing

Testing has been done with macOS + Chrome

Testing steps
1. open the Staging Branch URL, then sign in & set to admin mode as necessary. Alternatively, open your own empty Workflow with the Pages Editor.
2. You should be looking at an empty workflow with no Tasks, no Steps. If you're not, press the "Reset" button to set `tasks={}` and `steps=[]`.
3. Click "Add a New Task" button.
4. You should immediately see a new Task listed on the Pages Editor's Tasks page.
5. Inspect the Workflow resource via dev console (i.e. see what's POSTed to Panoptes). You should see the workflow look something like... `workflow: { tasks: { T0: {...} }, steps: [ [P0, {taskKeys: ['T0']}] ] ... }`
6. Repeat steps 3-5 as much as you like. New Tasks and Steps/Pages should always have new Tasks Keys and Step/Page Keys.

### Status

Ready for review

### DIFF
--- a/app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
+++ b/app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
@@ -4,11 +4,7 @@ Returns 0 by default.
  */
 /* eslint-disable radix */
 
-import { TASK_KEY_PREFIX, STEP_KEY_PREFIX } from './constants.js';
-
 export default function convertKeyToIndex(taskKeyOrStepKey = '') {
-  // regular expression looks like /^(?:T|P)(\d+)$/ - only the '\d+' is captured.
-  const re = RegExp(`^(?:${TASK_KEY_PREFIX}|${STEP_KEY_PREFIX})(\\d+)$`);
-  const indexStr = taskKeyOrStepKey?.match(re)?.[1]; // [1] is the '\d+' capture group.
+  const indexStr = taskKeyOrStepKey?.replace(/^(\D)+/g, '');
   return parseInt(indexStr) || 0;
 }
--- a/app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
+++ b/app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
@@ -1,0 +1,14 @@
+/*
+Transforms "T1234" (string) to 1234 (number).
+Returns 0 by default.
+ */
+/* eslint-disable radix */
+
+import { TASK_KEY_PREFIX, STEP_KEY_PREFIX } from './constants.js';
+
+export default function convertKeyToIndex(taskKeyOrStepKey = '') {
+  // regular expression looks like /^(?:T|P)(\d+)$/ - only the '\d+' is captured.
+  const re = RegExp(`^(?:${TASK_KEY_PREFIX}|${STEP_KEY_PREFIX})(\\d+)$`);
+  const indexStr = taskKeyOrStepKey?.match(re)?.[1]; // [1] is the '\d+' capture group.
+  return parseInt(indexStr) || 0;
+}
--- a/app/pages/lab-pages-editor/helpers/createStep.js
+++ b/app/pages/lab-pages-editor/helpers/createStep.js
@@ -3,5 +3,5 @@ Creates a Step, with tasks, that's ready to be inserted into a Workflow.
  */
 
 export default function createStep(stepKey, taskKeys = []) {
-  return [stepKey, { taskKeys }];
+  return [stepKey, { stepKey, taskKeys }];
 }
--- a/app/pages/lab-pages-editor/helpers/createStep.js
+++ b/app/pages/lab-pages-editor/helpers/createStep.js
@@ -1,0 +1,7 @@
+/*
+Creates a Step, with tasks, that's ready to be inserted into a Workflow.
+ */
+
+export default function createStep(stepKey, taskKeys = []) {
+  return [stepKey, { taskKeys }];
+}
--- a/app/pages/lab-pages-editor/helpers/createTask.js
+++ b/app/pages/lab-pages-editor/helpers/createTask.js
@@ -5,13 +5,9 @@ function needs to assign the appropriate Task Key to this Task, and then add it
 to the Workflow.
  */
 
-export default function createTask(taskType = 'text') {
-  // TODO
-  // Placeholder only
-  return {
-    help: '',
-    instruction: 'This is an example Text Task.',
-    required: false,
-    type: 'text'
-  };
+import tasks from '../../../classifier/tasks';
+
+export default function createTask(taskType) {
+  console.log('+++ taskType: ', taskType, tasks)
+  return tasks[taskType]?.getDefaultTask();
 }
--- a/app/pages/lab-pages-editor/helpers/createTask.js
+++ b/app/pages/lab-pages-editor/helpers/createTask.js
@@ -1,0 +1,17 @@
+/*
+Creates an empty/placeholder Task that's ready to be inserted into a Workflow.
+NOTE: the Task Key isn't handled by this function. Whatever's calling this
+function needs to assign the appropriate Task Key to this Task, and then add it
+to the Workflow.
+ */
+
+export default function createTask(taskType = 'text') {
+  // TODO
+  // Placeholder only
+  return {
+    help: '',
+    instruction: 'This is an example Text Task.',
+    required: false,
+    type: 'text'
+  };
+}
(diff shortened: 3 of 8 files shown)


eatyourgreens (commented):
This seems complicated, and will break on existing transcription workflows, where the keys begin with `S`. Why not slice the first character from each string instead?

app/pages/lab-pages-editor/helpers/createStep.js
```diff
@@ -0,0 +1,7 @@
+/*
+Creates a Step, with tasks, that's ready to be inserted into a Workflow.
+ */
+
+export default function createStep(stepKey, taskKeys = []) {
+  return [stepKey, { taskKeys }];
```
eatyourgreens, line 6:
```suggestion
  return [stepKey, { stepKey, taskKeys }];
```
You've missed the `stepKey` and `next` properties for the new step. The properties of each Step object should be listed in the Step model in the classifier code.

app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
```diff
@@ -0,0 +1,14 @@
...
+Transforms "T1234" (string) to 1234 (number).
+Returns 0 by default.
+ */
+/* eslint-disable radix */
+
+import { TASK_KEY_PREFIX, STEP_KEY_PREFIX } from './constants.js';
+
+export default function convertKeyToIndex(taskKeyOrStepKey = '') {
+  // regular expression looks like /^(?:T|P)(\d+)$/ - only the '\d+' is captured.
+  const re = RegExp(`^(?:${TASK_KEY_PREFIX}|${STEP_KEY_PREFIX})(\\d+)$`);
```
eatyourgreens, line 11: This seems complicated, and will break on existing transcription workflows, where the keys begin with `S`. Why not slice the first character from each string instead?

app/pages/lab-pages-editor/helpers/createTask.js
```diff
@@ -0,0 +1,17 @@
...
+
+export default function createTask(taskType = 'text') {
+  // TODO
+  // Placeholder only
+  return {
+    help: '',
+    instruction: 'This is an example Text Task.',
+    required: false,
+    type: 'text'
+  };
```
eatyourgreens, line 16: This function should return `TextTask.getDefaultTask()`

https://github.com/zooniverse/Panoptes-Front-End/blob/d9834aa0b47b7f13053c59bf7dd133c162370923/app/classifier/tasks/text/index.jsx#L148-L153

eatyourgreens:
```js
import tasks from '../../classifier/tasks';

export default function createTask(type) {
  return tasks[type]?.getDefaultTask();
}
```

app/pages/lab-pages-editor/helpers/getNewStepKey.js
```diff
@@ -0,0 +1,20 @@
...
+import convertKeyToIndex from './convertKeyToIndex.js';
+
+export default function getNewStepKey(steps = []) {
+  let newIndex = 0;
+  const stepKeys = steps.map((step) => (step[0] || '')).filter((step) => (step.length > 0));
+  stepKeys.forEach((stepKey) => {
+    const index = convertKeyToIndex(stepKey);
+    newIndex = (newIndex <= index) ? index + 1 : newIndex;
+  });
+  return `${STEP_KEY_PREFIX}${newIndex}`;
```
eatyourgreens, line 19: Live transcription workflows have step keys like `S0` and `S1`. See Davy Notebooks or PRINT for example. Will they work with this function?

app/pages/lab-pages-editor/helpers/createStep.js
shaunanoordin: Thanks Jim - I missed the second `stepKey` since I was referring to an old [ADR](https://github.com/zooniverse/front-end-monorepo/blob/master/docs/arch/adr-10.md) instead of the updated FEM spec files.

app/pages/lab-pages-editor/helpers/createTask.js
shaunanoordin: 👍

app/pages/lab-pages-editor/helpers/convertKeyToIndex.js
shaunanoordin: Yeah, this is over-engineered, and actually breaks when the index is more than 1 digit long. (e.g. P1234 = 4) Let me simplify this.

app/pages/lab-pages-editor/helpers/getNewStepKey.js
shaunanoordin: Yeah, dealing with existing workflows (and a potential "autocleanup/autoconvert" function) is something I'm still sweating about. 😓 

To answer the question, with the earlier version of convertKeyToIndex, if an existing WF has `steps = S0, S1`, this function should return `P0`, so the WF would have an odd mix of `steps = S0, S1, P0`, but should still ~~technically~~ _teeechnically...?_ work.

With the newer version of convertKeyToIndex, this would now return `P3`, so we'd _still_ have an odd mix of `steps = S0, S1, P2`. 😞 Either case, this is something on my To Test list, once I get step chaining implemented (i.e. properly get that `.next` working) in the next (or more like _next next)_ PR.

shaunanoordin: Oh! Forgot to add context for this: with the switch of the user-facing nomenclature to favour "Page" over "Step", we wanted to see if `P0` was possible. This _isn't_ a non-negotiable parameter as far as I know, so if using P becomes a huge Pain, I'll just switch that STEP_KEY_PREFIX back to `S`

(Please don't ask me about the topic starting pages/steps as P1 instead of P0.)

shaunanoordin:
## PR Update

- createStep(): added missing `stepKey` value.
- createTask(): now uses PFE's pre-defined Task Default Values instead of an arbitrary Text Task config.
  - Still hardcoded to use Text tasks for now though.
- convertKeyToIndex(): simplified. Now it just strips off any number of letters at the start string to extract the numbers. e.g. `"P123"`, `"S123"`, `"T123"`, `"wakkawakka123"` all return `123`

app/pages/lab-pages-editor/helpers/getNewStepKey.js
eatyourgreens: That's a good point. I think some of the monorepo tests start workflows at `S1`.

eatyourgreens: `next` should be optional. Maps are iterables, in the order that items were added to the map. A linear workflow should be able to default to iterating through the steps map in order. 

`next` is needed for decision trees, where a workflow has more than one final step, or where steps branch.

eatyourgreens (approved):
Looks good to me. I'm not sure either whether step keys should be `P0, P1, P2…`, or `S0, S1, S2…`, or even `S1, S2, S3…`. It's probably worth loading a transcription workflow into this editor at some point, to see how it handles those.

eatyourgreens:
Here's a live workflow from Davy Notebooks.
- it starts at `S0`.
- the `taskKeys` array is the only required property of the step object.

Screenshot of workflow.steps, from a Davy Notebooks workflow, in the Firefox network inspector.

PR #6916, opened 2023-11-03
Update Linter rules for modern React

shaunanoordin:
## PR Overview

This PR updates PFE's ESLint rules to adapt to our more modern React coding style.

Changes include:
- Allow default boolean attributes.
  -  This is now OK: `<input disabled />`
  - Previously, we had to do: `<input disabled={true} />`
- Don't require `import React from 'react'` to appear at the start of every JSX file. 
- Allow easier default props for functional components.
  - The default rules would require something like `SirMixALot.defaultProps = { likes: "big boats", canLie: false }` for _both_ classic React components and functional React components.
  - By switching off the rule for functional compos, we can now do `function SirMixALot({ likes = "big boats", canLie = false })`
- Allow console messages.
  - There is no fighting this, we already have 100+ existing uses of console.logs/warnings/errors in the codebase.

### Status

Ready for review. Low priority.


.eslintrc
```diff
@@ -7,10 +7,16 @@
   },
   "parser": "@babel/eslint-parser",
   "rules": {
-    "react/jsx-boolean-value": ["error","always"],
+    "react/jsx-boolean-value": 0,
```
eatyourgreens, line 10: I think this should be set to 'never'. 
https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/jsx-boolean-value.md

eatyourgreens (approved):

PR #6915, opened 2023-11-03
Pages Editor: add "Add New Task" Dialog

shaunanoordin:
## PR Overview

Part of: **Pages Editor MVP** project and **FEM Lab** super-project
Follow #6898
Staging branch URL: https://pr-6915.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR adds the "Add New Task" Dialog.

Expected behaviour:
- On the Tasks Page:
  - clicking "Add a new Task" button should open the Add New Task dialog
- On the Add New Task Dialog:
  - clicking on a Task type should...
    - create a new Step, e.g. `P0`
    - create a new Task, corresponding to the Task type, e.g. `T0`.
    - add the new Task to the new Step.
    - update the workflow resource with the task and steps (NOTE: most of the above has a POC with experimentalAddNewTaskWithStep())
    - close the Add New Task dialog, ~~then open the Edit Step dialog with the new Step selected.~~
  - pressing Esc or clicking the 'x' close button should close the dialog
- ~~On the Edit Step dialog:~~
  - ~~pressing Esc or clicking the 'x' close button should close the dialog~~
  - _No other actions for now, to be resolved in a follow up PR._

Design: Tasks Page, with 0 Tasks

image

Design: "Add a new Task" dialog

image

Design: "Create Task" dialog (actually, it's the **"Edit Step"** dialog)

image

Code notes:
- We're using the standard HTML `<dialog>` for modals; see #6914 for additional details on some shenanigans that needed to be solved.

### Status

WIP

PR #6939, opened 2023-11-13
Pages Editor: link Steps together

shaunanoordin:
## PR Overview

Follows #6915
Staging branch URL: https://pr-6939.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR ensures that Steps are automatically linked to each other (i.e. Page 1 goes to Page 2, etc) 

- `linkStepsInWorkflow()` added.
  - This function is currently called every time a new Step is added...
  - ...and _will_ also be called when a Step is rearranged or deleted. 
- The rules:
  - If a Step ~~is branching~~ **can** branch OR is the last Step in the workflow, step.next = undefined
  - Otherwise, that Step's step.next = next step's ID. 

Additional thoughts:
- In the future, linkStepsInWorkflow() could be part of a cleanUp() workflow function, which is called every time the WF is updated, and does the following: 1. delete all Tasks that don't belong to a Step, 2. delete anything in Steps.tasks without a corresponding Task, 3. delete all empty Steps (i.e. no associated Tasks), and 4. automatically link all Steps together (linkStepsInWorkflow()).
- See [FEM's convertWorkflowToUseSteps.spec.js](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.spec.js) for an idea of how Workflows with Steps should look like.

### Status

~~WIP, I'm still running tests to ensure this is working as expected. Actually rearranging Steps will be done in a separate PR.~~ Ready for review.

Do not merge until 6915 is merged. Currently targets pages-editor-pt8 for reviewing purposes; this will be re-targeted to master when ready.

### DIFF
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -48,7 +48,6 @@ export default function TasksPage() {
 
   function experimentalLinkSteps() {
     const newSteps = linkStepsInWorkflow(workflow?.steps, workflow?.tasks);
-    console.log('+++ linkStepsInWorkflow()\n', workflow?.steps, '\n==>\n', newSteps);
     update({ steps: newSteps });
   }
 
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -34,7 +34,7 @@ export default function TasksPage() {
       ...workflow.tasks,
       [newTaskKey]: newTask
     };
-    const steps = [...workflow.steps, newStep];
+    const steps = linkStepsInWorkflow([...workflow.steps, newStep]);
 
     update({ tasks, steps });
   }
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -93,16 +93,18 @@ export default function TasksPage() {
           }}
         >
           <button
-            className="big primary"
+            className="big"
             onClick={experimentalReset}
             type="button"
+            style={{ margin: '0 4px' }}
           >
             RESET
           </button>
           <button
-            className="big primary"
+            className="big"
             onClick={experimentalLinkSteps}
             type="button"
+            style={{ margin: '0 4px' }}
           >
             LINK
           </button>
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -49,6 +49,7 @@ export default function TasksPage() {
   function experimentalLinkSteps() {
     const newSteps = linkStepsInWorkflow(workflow?.steps, workflow?.tasks);
     console.log('+++ linkStepsInWorkflow()\n', workflow?.steps, '\n==>\n', newSteps);
+    update({ steps: newSteps });
   }
 
   if (!workflow) return null;
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -9,6 +9,7 @@ import getNewTaskKey from '../../helpers/getNewTaskKey.js';
 import getNewStepKey from '../../helpers/getNewStepKey.js';
 import createTask from '../../helpers/createTask.js';
 import createStep from '../../helpers/createStep.js';
+import linkStepsInWorkflow from '../../helpers/linkStepsInWorkflow.js';
 // import strings from '../../strings.json'; // TODO: move all text into strings
 
 import NewTaskButtonAndDialog from './components/NewTaskButtonAndDialog.jsx';
@@ -45,6 +46,11 @@ export default function TasksPage() {
     });
   }
 
+  function experimentalLinkSteps() {
+    const newSteps = linkStepsInWorkflow(workflow?.steps);
+    console.log('+++ linkStepsInWorkflow()\n', workflow?.steps, '\n==>\n', newSteps);
+  }
+
   if (!workflow) return null;
 
   return (
@@ -77,13 +83,30 @@ export default function TasksPage() {
             />
           ))}
         </ul>
-        <button
-          className="big primary"
-          onClick={experimentalReset}
-          type="button"
+
+        {/* EXPERIMENTAL */}
+        <div
+          style={{
+            padding: '16px',
+            margin: '8px 0',
+            border: '2px dashed #c04040'
+          }}
         >
-          RESET
-        </button>
+          <button
+            className="big primary"
+            onClick={experimentalReset}
+            type="button"
+          >
+            RESET
+          </button>
+          <button
+            className="big primary"
+            onClick={experimentalLinkSteps}
+            type="button"
+          >
+            LINK
+          </button>
+        </div>
       </section>
     </div>
   );
--- a/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
+++ b/app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
@@ -47,7 +47,7 @@ export default function TasksPage() {
   }
 
   function experimentalLinkSteps() {
-    const newSteps = linkStepsInWorkflow(workflow?.steps);
+    const newSteps = linkStepsInWorkflow(workflow?.steps, workflow?.tasks);
     console.log('+++ linkStepsInWorkflow()\n', workflow?.steps, '\n==>\n', newSteps);
   }
 
--- a/app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
+++ b/app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
@@ -1,0 +1,30 @@
+/*
+Links Steps in a Workflow together.
+
+- Input: array of Steps
+- Output: copy of array of Steps, with their `step.next`s updated.
+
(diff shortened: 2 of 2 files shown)


eatyourgreens:
Noting that steps don’t need to be linked together for many workflows. Take a look at Davy Notebooks, for example. 

Only branching workflows need to explicitly link steps together. Linear workflows will use the order of the steps map by default.

shaunanoordin:
## PR Update

- linkStepsInWorkflow() now implements the branching check properly.
- ❗ I've changed `isStepBranching` and `isTaskBranching` (the latter taken from [FEM](https://github.com/zooniverse/front-end-monorepo/blob/master/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.js)) to `canStepBranch` and `canTaskBranch`
  - I belatedly realised that the Pages Editor isn't so much interested if the Step/Task **is** branching, but whether it **can,** since that logic determines whether or not we want to show the branching options to the Project Owner.
  - canStepBranch() returns the ID of the Task that branches. (In the case of multiple Tasks that can branch, return the FIRST.)
  - Checking whether a Task can branch is simple: _is it a Single Question Task?_

### Dev Notes

TODO:
- check if there's any task _other_ than the single question task that can branch.
- check what happens on FEM Classifier when a Question Task has 0 choices. It seems to automagically count as a "final Task/submit" 🤔 (EDIT: which I guess does make sense... but this needs to be made explicit in the Pages Editor, since visually it may look like a standard "move to next Step/Page" thing)
- find out what happens when a Step has MULTIPLE single question tasks.
  - See what happens in FEM classifier and/or dig up the workflow branching code.
  -  I'm just assuming that the _first_ branching task determines the branching behaviour, but this is based on nothing but fantasy.

> Noting that steps don’t need to be linked together for many workflows. Take a look at Davy Notebooks, for example.

👍 Noted, thanks! For the moment, I'll stick with `step.next` to make it explicit, even though (by definition) the Pages Editor is visually designed to lay out the Pages (steps) in an intuitive, linear way to the project owner. (Data-wise, my brain just works better seeing the explicit links.)

### Status

PR is now ready for review.

app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
```diff
@@ -0,0 +1,54 @@
...
+- If a Step *can* branch, then step.next = undefined.
+- If a Step is the final step in the array, then step.next = undefined. 
+- Otherwise, each Step's step.next will be the ID of the next Step in the array.
+ */
+
+export default function linkStepsInWorkflow(steps = [], tasks = {}) {
+  const newSteps = steps.map((step, index) => {
+    const [stepId, stepBody] = step;
+
+    const isFinalStep = index >= (steps.length - 1);
```
eatyourgreens, line 17: This is true for linear workflows, which don't need `step.next` anyway, but branched workflows will have more than one final step (one final step for the end of each branch.)

eatyourgreens: In this example, `S0` branches to either `S1` or `S2`, both of which are final steps ie. you can submit a classification either from `S1` or from `S2`.

https://github.com/zooniverse/front-end-monorepo/blob/45dde3a406671d480e97142789375d595afa086a/packages/lib-classifier/src/store/helpers/convertWorkflowToUseSteps/convertWorkflowToUseSteps.spec.js#L243-L259

shaunanoordin: Dangit, this variable name was more confusing than I realised. I'll rename it to `isFinalStepInArray` 👌 

> If a Step is the final step in the array, then step.next = undefined.

eatyourgreens: Can the final step in the array loop back to an earlier step, for recursive workflows? An example of this would be transcribing tables with a variable number of entries (Scarlets & Blues.)

shaunanoordin: Yes, that's possible, but the project owner would need to (1) make the final Step/Page (in the array) a Single Question task to routes users back to an earlier step, and (2) manually create an off-ramp at some earlier step (which would also be a Single Question Task)

Example 1:

image

Example 2: (trimmed)

image

image

I believe the goal for this new Pages Editor design is to encourage project owners to make the workflow _as linear as possible,_ by default.

eatyourgreens: Is that second example Scarlets & Blues? That’s exactly the workflow that I had in mind.

app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
```diff
@@ -45,6 +46,12 @@ export default function TasksPage() {
     });
   }
 
+  function experimentalLinkSteps() {
+    const newSteps = linkStepsInWorkflow(workflow?.steps, workflow?.tasks);
+    console.log('+++ linkStepsInWorkflow()\n', workflow?.steps, '\n==>\n', newSteps);
```
kieftrav, line 51: Probably want to remove this before merging

app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
```diff
@@ -0,0 +1,54 @@
+/*
+Links Steps in a Workflow together.
+
+- Input: array of Steps, and Tasks object.
+- Output: COPY of array of Steps, with their `step.next`s updated.
+
+Rules for linking steps:
+- If a Step *can* branch, then step.next = undefined.
+- If a Step is the final step in the array, then step.next = undefined.
```
kieftrav, line 9: My personal preference is to not conflate two purposes for a value of "undefined". An alternative is to make the final step = -1 or a somewhat nonsensical yet meaningful to the developer value that's easy to check. Not a requirement, but  I find it very useful to my future self figuring out how to debug something like this.

app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
```diff
@@ -0,0 +1,54 @@
...
+- If a Step *can* branch, then step.next = undefined.
+- If a Step is the final step in the array, then step.next = undefined. 
+- Otherwise, each Step's step.next will be the ID of the next Step in the array.
+ */
+
+export default function linkStepsInWorkflow(steps = [], tasks = {}) {
+  const newSteps = steps.map((step, index) => {
+    const [stepId, stepBody] = step;
+
+    const isFinalStepInArray = index >= (steps.length - 1);
```
kieftrav, line 17: Wouldn't it make more sense to use "===" instead of ">="? Is there any scenario in which the index will be greater than the steps.length - 1?

app/pages/lab-pages-editor/helpers/linkStepsInWorkflow.js
```diff
@@ -0,0 +1,54 @@
...
+    const canBranch = canStepBranch(step, tasks);
+    const nextStep = (isFinalStepInArray || canBranch) ? undefined : steps[index+1]?.[0];
+
+    const newStepBody = {
+      ...stepBody,
+      next: nextStep
```
kieftrav, line 23: I'd consider keeping this variable named "nextStep" so it's clear what exactly "next" means... Next what? Ah, it's the nextStep!

kieftrav (approved):
LGTM! Clearly a WIP stepping stone. My comments are suggestions with the strongest being for the "undefined" ambiguity and the "next" variable name disambiguation.

shaunanoordin:
Thanks @kieftrav ! 👍 I've added a quick cleanup for some of the items you spotted (stray console.logs, === vs >=), but the `step.next = undefined` actually deserves some extra explanation, since it's one of those _"under-the-hood Panoptes"_ things that may not be fully documented in one place.

OK, so: the `next` in `workflow.steps[n].next` has a specific/implicit meaning in the _Panoptes **workflow** resource._ (Example: `https://panoptes-staging.zooniverse.org/api/workflows/3711`)

Here's an example (simplified) workflow:
```
const workflow3711 = {
  steps: [  // (note1)
    ['Page1', { taskKeys: [ 'drawTaskA' ], next: 'Page3' } ],
    ['Page2', { taskKeys: [ 'drawTaskB' ], next: 'Page4' } ],
    ['Page3', { taskKeys: [ 'textTaskC' ] } ],
    ['Page4', { taskKeys: [ 'textTaskD' ] } ]
  ]
};
```

Page1 has a `.next`, which tells the classifier (note2) that the next step should be Page3. _(This is why I can't use '.nextStep'; the '.next' itself has implicit meaning in the data structure)_

Page3 doesn't have a `.next` (note3), which _implicitly_ tells the classifier that the next step is the next step in the array, i.e. Page4. _(My new code explicitly spells out the .next step, to reduce ambiguity.)_

Page4 doesn't have a `.next` AND it's the final item in the steps array, which _implicitly_ tells the classifier that the next "step" is to actually **submit** the classification.

Note that Page2 is completely skipped in this example. Goodbye, Page2!

- note1: _steps_ are _pages._ In Panoptes/FEM/PFE, the data node is called `steps`. When presenting this data to users, we're colloquially calling them "pages".
- note2: well, the _FEM_ classifier at least. I haven't gone through the PFE classifier code to see how it handles steps.
- note3: due to how [our PJC library](https://github.com/zooniverse/panoptes-javascript-client/) handles JavaScript-Object-to-JSON conversions, `step[n].next = undefined` basically means "delete the entire .next key/value from this step's data node". So `{ foo: 'bar', next: undefined }` in JS becomes `{ "foo": "bar" }` in JSON.
- note4: not discussed in this simple overview: (i) how Single Question Tasks create branching logic, and (ii) how to submit from a middle step.

PR #6947, opened 2023-11-16
Pages Editor: allow Steps to be rearranged

shaunanoordin:
## PR Overview

Follows #6939
Staging branch URL: https://pr-6947.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR allows Pages/Steps in the workflow.steps array to be rearranged.

- New helper: `moveItemInArray()`. This is a pretty useful general-purpose utility function, as it turns out.
- TasksPage: new `moveStep()` function can be called to rearrange Pages/Steps, _run a cleanup process that re-links the newly arranged Pages/Steps,_ and saves the changes to Panoptes.
- StepsItem:
  - New action: pressing the up/down buttons will rearrange the Page/Step.
  - **[TODO]** New action: Page/Step can be rearranged via dragging & dropping

### Status

WIP, I still need to do that drag & drop.

Do not merge until 6915 + 6939 are merged. Currently targets pages-editor-pt9 for reviewing purposes; this will be re-targeted to master when ready.

### DIFF
--- a/css/lab-pages-editor.styl
+++ b/css/lab-pages-editor.styl
@@ -305,11 +305,12 @@ $fontWeightBoldPlus = 700
       padding: 0
 
     .step-item
-      background: $grey5
-      border: 1px solid $grey1
-      border-radius: $sizeXS
-      margin-top: $sizeM
-      padding: $sizeS $sizeM
+      .step-body
+        background: $grey5
+        border: 1px solid $grey1
+        border-radius: $sizeXS
+        margin-top: $sizeM
+        padding: $sizeS $sizeM
 
       .step-controls
         .step-controls-left
--- a/css/lab-pages-editor.styl
+++ b/css/lab-pages-editor.styl
@@ -312,11 +312,17 @@ $fontWeightBoldPlus = 700
         border-radius: $sizeXS
         padding: $sizeS $sizeM
 
+        &:has(.move-button:focus)  // Note: :has() is experimental as of Nov 2023, and not supported on Firefox
+          outline: 2px solid $yellow
+
       .step-controls
         cursor: grab
 
         &:active
           cursor: grabbing
+        
+        .move-button:focus
+          outline: 1px solid $yellow
         
         .step-controls-left
           min-width: 5rem
--- a/css/lab-pages-editor.styl
+++ b/css/lab-pages-editor.styl
@@ -301,15 +301,15 @@ $fontWeightBoldPlus = 700
     
     .steps-list
       list-style: none
-      margin: $sizeM 0 0 0
+      margin: $sizeS 0 0 0
       padding: 0
 
     .step-item
+
       .step-body
         background: $grey5
         border: 1px solid $grey1
         border-radius: $sizeXS
-        margin-top: $sizeM
         padding: $sizeS $sizeM
 
       .step-controls
@@ -335,6 +335,9 @@ $fontWeightBoldPlus = 700
 
             .fa
               color: $grey3
+
+      .step-drop-target
+        padding: $sizeXS
 
     .tasks-list
       list-style: none
--- a/css/lab-pages-editor.styl
+++ b/css/lab-pages-editor.styl
@@ -313,6 +313,11 @@ $fontWeightBoldPlus = 700
         padding: $sizeS $sizeM
 
       .step-controls
+        cursor: grab
+
+        &:active
+          cursor: grabbing
+        
         .step-controls-left
           min-width: 5rem
         
--- a/css/lab-pages-editor.styl
+++ b/css/lab-pages-editor.styl
@@ -337,7 +337,11 @@ $fontWeightBoldPlus = 700
               color: $grey3
 
       .step-drop-target
-        padding: $sizeXS
+        padding: $sizeS
+
+        &.active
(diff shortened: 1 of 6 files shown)


shaunanoordin:
## PR Update

Drag & drop has been implemented!

- StepItem has been reconfigured to include a `.step-body` and a `<DropTarget>`
- "Drag & drop to rearrange pages/steps" works exactly as you might expect. 

Video: 4-page workflow that starts as [T2,T0,T1,T3] and is rearranged to [T0,T1,T2,T3]
https://github.com/zooniverse/Panoptes-Front-End/assets/13952701/3b764173-a498-4d02-a61d-702e50b28d5e

### Dev Notes

Room for improvement/discussions with @seanmiller26 :

- "Drop Targets"
  - At the moment, when a Page is dragged to a valid drop target (i.e. _the spaces between Pages,_ plus the space before & after all Pages), I've highlighted the target in BRIGHT GREEN and made the space BIGGER to indicate that, hey, you can drop here. There's probably a better visual design for this.
  - Drop targets are invisible until a Page is dragged on top of it. Not sure if users will intuitively know they can drag Pages to between other Pages.
  - There are always two "dead targets", i.e. if you drag Page to those drop targets, nothing happens. Because logically, nothing _can_ happen. Basically, if you drag Page 2 to "between Page 1 & Page 2", or to "between Page 2 & Page 3", then Page 2 gets _swapped with itself,_ meaning nothing happens. This isn't a functionality problem, but miiight be confusing UI/UX-wise.
- Animating Page swapping:
  - Yeah I ain't got nothing at the moment. Not sure how I can get CSS animations working smoothly with React re-renders.

We can discuss this more on the Monday checkin!

### Status

Ready for review, but I may update this PR further pending on feedback from Sean, or spin those updates into their own PR if they're sufficiently chunky. Or - as Sam might point out - some of these UI fine-tuning/tweaks could be reserved until we get the MVP in front of users so _they_ can decide if e.g. invisible drop targets are intuitive or not.

css/lab-pages-editor.styl
```diff
@@ -312,12 +312,18 @@ $fontWeightBoldPlus = 700
         border-radius: $sizeXS
         padding: $sizeS $sizeM
 
+        &:has(.move-button:focus)  // Note: :has() is experimental as of Nov 2023, and not supported on Firefox
```
eatyourgreens, line 315: It is planned to land in Firefox 121, though, so should be supported by next month.

shaunanoordin:
## PR Update

Some design updates:

- When dragging Steps/Pages, if _no rearrangement happens,_ the drop target _is no longer highlighted._ 
  - This means adjacent drop targets are now dead zones.
- When a Step/Page is rearranged via buttons, the Step/Page has a yellow/gold outline to indicate it was moved.
  - It's not CSS animation, but it's what we've got for now
- The "header" portion (.step-controls) of a Step/Page now indicates its grabbability with the grab/grabbing cursor
  - Previously, only the grab-handle had the grab-grabbing cursor.
  - Functionality hasn't changed, though: the whole StepItem remains grabbable.

seanmiller26:
Based on the latest version I think it would make sense if a Page that is currently being dragged also gets an outline. And it can stay when it is dropped into place, similar to when it is moved with the arrows. This should make the action a little more obvious.

Otherwise the functionality is working great!

kieftrav (approved):
Looks good, works as expected, and only very minor code style comments. Tests would be nice but optional.

kieftrav:
@shaunanoordin - code looks great! I'm not super familiar with the background on this but when rearranging the tasks I noticed that the label `T0` and `T3` follow the task instead of being rewritten so it's always `T0`, `T1`, ... Is that going to cause any problems later on? Mostly out of curiosity since I would personally be a little confused by joining a project and seeing them in all sorts of different orders with no actual significance to the numbering.

@seanmiller26 - a thought on the dropzone styling.. I've generally seen the attached as a guiding best practice. Instead of having a solid outline it's usually dashed and the color is a bit more muted / opaque. There's also usually some text that indicates this is a place to put the moved item. Another option is to put a skeleton loader or an opaque version of a clone of the actual item/html itself. Just some thoughts.

magnetism-drop-zoneinvisible-area

shaunanoordin:
Apologies for taking so long to merge this PR!

Travis, thank you again for reviewing this 👍

For your question: _"when rearranging the tasks I noticed that the label T0 and T3 follow the task instead of being rewritten so it's always T0, T1, ... Is that going to cause any problems later on?_

The answer is that it won't cause a problem on the _technical_ side. `T0`, `T1`, `T2`, etc are just _unique identifiers_ generated for each task, with the `T{n}` format being an arbitrary standard we chose. On the _project owner_ side, hmm, yeah. We're kinda assuming that they're familiar with the arbitrary `T{n}` format, but I can see how new project owners may mistakenly assume that `T{n}` implies a _sequence._ Hopefully, once Sean's designs are fully implemented (with big honking visual arrows pointing one Page to the next), there'll be no confusion, but I'm keen to hear user feedback. (And on the _volunteer_ side, they'll never see the Tasks' unique identifiers, so no problem there)

OK, let's get this merged, and I'll move on to the next PR!

PR #7046, opened 2024-02-16
Pages Editor: implement Delete Page action

shaunanoordin:
## PR Overview

Part of: **Pages Editor MVP** project and **FEM Lab** super-project
Follows #6991
Staging branch URL: https://pr-7046.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR adds the ability to **delete a Step/Page**.

- Clicking on the "trash can" icon on a Page will prompt a confirmation dialog, before deleting a Step/Page.
- Any _Branching Task_ with answers that pointed to the now-deleted Page will be updated, so the answers now point to nothing (i.e. "Submit").
  - A new `cleanupTasksAndSteps()` helper function has been added to remove orphaned references in branching tasks, (this has been implemented). This maaay be updated in the future to also remove Tasks that aren't associated with any Steps, and remove Steps that have 0 Tasks. We'll see.

Other changes:
- Fixed an issue where deleting Answers in a Single Question Task (by clicking on the "minus" button) didn't actually register changes.

Notes:
- ⚠️ The "Preview Workflow" button is _still_ hardcoded to WF 3711. Shaun: please remember to fix this by loading the Project resource in DataManager in the next PR.

### Status

Ready for review!

PR #7048, opened 2024-02-23
Pages Editor: implement Associated Subject Sets & fetch more data

shaunanoordin:
## PR Overview

Follows #7046
Staging branch URL: https://pr-7048.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR implements the "Associated Subject Sets" feature on the Workflow Settings sub-page.

- New component: AssociatedSubjectSets.
- New behaviour/feature: on Pages Editor (for any Workflow) -> Workflow Settings...
  - The _Associated Subject Sets_ feature should list all Subject Sets available to the Project _(caveat: up to a max of 250. Paging may be a future problem to solve.)_
  - The Associated Subject Sets feature _(and no, I will not shorten it to an acronym)_ should show which Subject Sets are linked to the Workflow via ticked checkboxes.
  - Clicking on the checkboxes should link/unlink Subject Sets to/from the Workflow.

Also, some background work was required to fetch the Subject Sets, so the DataManager was updated as well.
- Updated behaviour: DataManager now fetches the **project** resource.
- Misc update: the Preview Workflow link now works as intended. 

_Screenshot: example workflow with 3 available subject sets, 2 of which are linked._

image

### Testing

- Open the staging URL. You should be looking at a test workflow (e.g. staging wf 3711)
- Click on the Workflow Settings tab/button to view the sub-page.
- Look for the Associated Subject Sets feature.
  - It should list all available subject sets, and show which ones are already linked.
  - Click on a subject set to link/unlink them
  - Reload the page to confirm
  - Expectation: you should be able to link or unlink as many sets as you want. Changes should be committed to Panoptes.

### Status

Ready for review!

This PR is currently targeting pages-editor-pt14 for review purposes. Do not merge until 6974 is ready and this branch's merge is re-targeted to `master`.

### DIFF
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -2,6 +2,10 @@
 Data Manager
 Responsible for fetching, updating, and saving Workflow resources from/to
 the Panoptes service.
+
+Also fetches the project resource for secondary purposes. (e.g. figuring out
+which Subject Sets are available for the workflow, available experimental tools,
+and the workflow's preview URL.)
  */
 
 // ESLint: don't import global React, and don't use .defaultProps.
@@ -68,6 +72,7 @@ function DataManager({
   }, [workflowId]);
 
   // Listen for workflow changes, and update a counter to prompt useMemo to update the context.
+  // NOTE: we're not listening for *project* changes, as the DataManager isn't meant to update that resource.
   useEffect(() => {
     const wf = apiData.workflow;
     function onWorkflowChange() {
@@ -94,19 +99,21 @@ function DataManager({
 
       const newWorkflow = await wf.update(data).save();
 
-      setApiData({
+      setApiData((prevState) => ({
+        ...prevState,
         workflow: newWorkflow, // Note: newWorkflow is actually the same as the old wf, so useMemo will have to listen to status changing instead.
         status: 'ready'
-      });
+      }));
 
       return newWorkflow;
     }
 
     return {
+      project: apiData.project,
       workflow: apiData.workflow,
       update
     };
-  }, [apiData.workflow, updateCounter]);
+  }, [apiData.project, apiData.workflow, updateCounter]);
 
   if (!workflowId) return (<div>ERROR: no Workflow ID specified</div>);
   // if (!workflow) return null
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -122,7 +122,12 @@ function DataManager({
     <WorkflowContext.Provider
       value={contextData}
     >
-      <div>{apiData.status}</div>
+      {apiData.status === 'fetching' && (
+        <div className="status-banner fetching">Fetching data...</div>
+      )}
+      {apiData.status === 'error' && (
+        <div className="status-banner error">ERROR: could not fetch data</div>
+      )}
       {children}
     </WorkflowContext.Provider>
   );
--- a/app/pages/lab-pages-editor/DataManager.jsx
+++ b/app/pages/lab-pages-editor/DataManager.jsx
@@ -17,9 +17,11 @@ import { WorkflowContext } from './context.js';
 function DataManager({
   // key: to ensure DataManager renders FRESH (with states reset) whenever workflowId changes, use <DataManager key={workflowId} ... />
   children = null,
+  projectId = '',
   workflowId = ''
 }) {
   const [apiData, setApiData] = useState({
+    project: null,
     workflow: null,
     status: 'ready'
   });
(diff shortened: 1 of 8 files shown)


app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
```diff
@@ -140,7 +140,8 @@ export default function TasksPage() {
     update({ tasks: newTasks });
   }
 
-  const previewUrl = 'https://frontend.preview.zooniverse.org/projects/darkeshard/example-1982/classify/workflow/3711?env=staging';
+  const previewEnv = getPreviewEnv();
+  const previewUrl = `https://frontend.preview.zooniverse.org/projects/${project?.slug}/classify/workflow/${workflow?.id}${previewEnv}`;
```
kieftrav, line 144: Should we maybe do something different here if there isn't a project slug and workflow id? Because if its missing and it gets clicked on I would rather it send them somewhere predictable/debuggable instead of a default weird url with null/undefined/ect in it.

app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
```diff
@@ -230,3 +231,21 @@ export default function TasksPage() {
     </div>
   );
 }
+
+function getPreviewEnv() {
```
kieftrav, line 235: This function feels too important to bury in a file... I'd personally want something like this (with clear exceptions / responsive to our development URL schema) surfaced a bit higher. I'ts own file, part of a helper, or something that makes its purpose more surfaceable.

app/pages/lab-pages-editor/components/WorkflowSettingsPage/components/AssociatedSubjectSets.jsx
```diff
@@ -0,0 +1,82 @@
...
+        const ssets = await project.get('subject_sets', { sort: '+id', page_size: ARBITRARY_PAGE_SIZE });
+        if (!ssets) throw new Error('No subject sets');
+
+        setApiData({
+          subjectSets: ssets,
+          status: 'ready'
+        });
+
+      } catch (err) {
+        console.error('AssosicatedSubjectSets: ', err);
```
kieftrav, line 29: Misspelling and would be useful to indicate its an error in the console.log(). I personally like using the following format:

`console.log("AssociatedSubjectSets() fetchSubjectSets() error", err);`

That way you have a guide in the console on how to find that console log (my future self really appreciates these guideposts). Only a suggestion though.

kieftrav (approved):
LGTM with only a couple comments. One thing I wanted to surface is that I expected refreshing the page to keep me on the same tab "workflow settings".. Is this a bug? Or desired behavior? Personally, when a tab changes 50%+ of the page I assume refreshing the page / pushing the back button will take me back to the previous state. In this case, the back button also doesn't take me back to the "workflow settings" tab. This also might be future work so take this comment as is helpful!

app/pages/lab-pages-editor/DataManager.jsx
```diff
@@ -31,33 +37,42 @@ function DataManager({
...
         setApiData({
+          project: null,
           workflow: null,
           status: 'error'
         });
       }
     }
 
-    fetchWorkflow();
+    fetchProjectAndWorkflow();
```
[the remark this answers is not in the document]
shaunanoordin: An `await` is not needed here, as we want the fetch action to run asynchronously in the background; the `state.apiData.status` is set to `fetching` so users get to see a loading message while the data fetch occurs. 👌

app/pages/lab-pages-editor/components/TasksPage/TasksPage.jsx
shaunanoordin: Let me think about this. 🤔 

You're right that a URL like `https://frontend.preview.zooniverse.org/projects/undefined/classify/workflow/undefined` would be pretty wonky. But it's a fairly safe assumption every project has a slug and every workflow has an ID. But _but_ it's still good to have a safety for when those assumptions bork.

app/pages/lab-pages-editor/components/WorkflowSettingsPage/components/AssociatedSubjectSets.jsx
shaunanoordin: Derp! I definitely need to fix that misspelling, thank you.

And thank you for the suggestion, I might also need to revise how I write our error messages. (At the moment I'm making the standards up as I go.)

shaunanoordin:
Thanks Travis! I'm going to rebase, then address some of the smaller feedback in this PR, before merging. 👍 

To respond to the point you raised:
> I expected refreshing the page to keep me on the same tab "workflow settings".

Good point. Perhaps ideally, each "tab" should be its own sub-page with a URL, like `https://local.zooniverse.org:3735/lab/1982/workflows/editor/3711/workflow-settings`. However, at the moment the constraint I've imposed on myself is that the PagesEditor should be as self-contained as possible, so using a single path (`/lab/PROJECT_ID/workflows/editor/WORKFLOW_ID`) for _all_ the PagesEditor tabs/sub-pages/features allows me to side step adding router dependencies inside the PagesEditor (which would've been somewhat annoying should we plan to port this to FEM.)

EDIT: Wait, a thought occurs: `?tab=workflowSettings` param instead of `/workflow-settings` subpath to circumvent having to use react-router? Worth experimenting with later.

PR #7083, opened 2024-04-24
Email Settings: add "valid email" and "verified email" status

shaunanoordin:
## PR Overview

Staging branch URL: https://pr-7083.pfe-preview.zooniverse.org
Towards #7032

This PR adds "🟢 valid / 🔴 invalid email" and "🟢 verified / 🔴 unverified email" indicators to the [Email Settings page](https://www.zooniverse.org/settings/email) (under User -> Settings -> Email)

- These indicators are based on the `user.valid_email` and `user.confirmed_at` values of the User resource.

Screenshot 2024-04-24 at 02 55 55

### Status

Ready for review.

Next steps: the "request confirmation email" functionality requires either
1. a fairly custom (and complex) fetch() function inside email.jsx to POST to panoptes.zooniverse.org/api/users/confirmation with the auth token inside the code, or
2. an update to PJC to do the same, or
3. _as a placeholder,_ a simple link to https://panoptes.zooniverse.org/users/confirmation/new

For this PR, the "request confirmation email" has been disabled. (I got halfway through option 1 but ran into some bugs due to a malformed request, and option 3 hit a small snag when I realised I need to check the env to get the correct URL.)

Issue #7032, opened 2024-01-26
Add email verification and validity status to email settings page


lcjohnso:
## New Expected Behavior
Since 25 Jan 2024, due to https://github.com/zooniverse/panoptes/pull/4268, new users are required to confirm their email in order to post on Talk. As a result, we would like to provide two capabilities for users on the email settings page (https://www.zooniverse.org/settings/email):
1. Check their email confirmation / verification status.
2. Request a new confirmation email be sent to the user's email address.

Furthermore, email addresses can be marked invalid when an email from Zooniverse is reported as spam, experiences a hard bounce (e.g., user does not exist at stated domain), or the email address is manually marked as invalid via the user admin page. Related to this, we also wish to add the following third capability for users: 

3. Check their email validity status (and potentially edit their email if invalid).

## Feature Specification
Use the following fields that are serialized into the user `api/me` response to inform current status:
 
Email Validity States
- `"valid_email": true` for valid email address
- `"valid_email": false` for invalid email address

Email Confirmation / Verification States
- `"confirmed_at": null` for unconfirmed email address
- `"confirmed_at": 2024-01-25 17:33:01` for confirmed email address

Request New Confirmation Email (@zwolf please confirm or edit if below is not correct)
- send POST to https://panoptes.zooniverse.org/users/confirmation
    - similar to how https://panoptes.zooniverse.org/users/confirmation/new page works

## Current behavior
Email confirmation status and validity are not currently presented to the user.  For requesting new confirmation email: a page exists to make request by submitting an email address (https://panoptes.zooniverse.org/users/confirmation/new) but this page is not obvious or easy to access or find. 

## Accounts for Testing
On Prod:
- Confirmed User: https://www.zooniverse.org/admin/users/1901126
- Unconfirmed user: https://www.zooniverse.org/admin/users/2698362
- Invalid Email User: https://www.zooniverse.org/admin/users/2698365

On Staging:
- Confirmed User: https://master.pfe-preview.zooniverse.org/admin/users/1326467?env=staging
- Unconfirmed user: https://master.pfe-preview.zooniverse.org/admin/users/1326468?env=staging
- Invalid Email User: https://master.pfe-preview.zooniverse.org/admin/users/1326458?env=staging

eatyourgreens:
>Request New Confirmation Email (@zwolf please confirm or edit if below is not correct)
>
 >   send POST to https://panoptes.zooniverse.org/users/confirmation
[...]

PR #7117, opened 2024-06-01
Email verification link

shaunanoordin:
## PR Overview

Staging branch URL: https://pr-7117.pfe-preview.zooniverse.org/settings/email
Follows #7083, towards #7032

This PR adds a link to the [Email Settings page](https://www.zooniverse.org/settings/email) (under User -> Settings -> Email), which allows users to request a confirmation email from Panoptes.

image

New behaviour:
- This link only appears if your account's email is unverified.
- Clicking on the link will send a request to Panoptes, to send the confirmation email.
- Status messages:
  - Idle: "Resend confirmation email" (link)
  - Requesting: "Requesting..."
  - Success: "Confirmation email requested, please check your inbox"
  - Error: "ERROR: could not request confirmation email"

### Testing

- Login using a test account with an unverified email.
- Go to the Email Settings page.
- Click on the "Resend confirmation email" link.

### Status

⚠️ WIP - this PR is not yet functional.

As of commit aeaeca251b6e15242a2b3799eb80d33bd0bb944f, there's a problem where sending the request results in a **406 error,** at least for staging.


app/pages/settings/email.jsx
```diff
@@ -292,12 +313,27 @@ class EmailSettingsPage extends React.Component {
...
+                    <span>
+                      <Translate content="emailSettings.general.emailUnverifiedSuccess" />
+                    </span>
+                  )}
+                  {(this.state.requestConfirmationEmailStatus === 'error') && (
+                    <span style={{ color: 'red' }}>
+                      <Translate content="emailSettings.general.emailUnverifiedError" />
+                    </span>
+                  )}
                 </div>
```
eatyourgreens, line 337: Since the text here is changing, you’ll want to wrap it in one of the live region roles (probably ’status’), so that changes are announced in screen readers. 

Note that live regions are expected to be empty when they load/mount, so any initial status won’t be read out.

app/pages/settings/helpers/getBearerToken.js
```diff
@@ -0,0 +1,21 @@
+export default async function getBearerToken (authClient) {
+  let token = ''
+  if (authClient) {
+    try {
+      const authToken = await authClient.checkBearerToken()
+      // API method between PJC's auth and oauth clients are inconsistent.
+      // First party auth client returns a string of the bearer token
+      // Oauth client returns an object of the bearer token, expiration, and token type
```
eatyourgreens, line 8: The PFE user settings page doesn’t use third party OAuth.

If you’re logged in, you can always get your bearer token from `auth.checkBearerToken()`. 

See https://zooniverse.github.io/panoptes-javascript-client/#panoptes-javascript-client-auth

app/pages/settings/email.jsx
```diff
@@ -292,12 +313,27 @@ class EmailSettingsPage extends React.Component {
...
                   {' '}
                   <Translate content="emailSettings.general.emailUnverified" />
-                  {/*
                   {' | '}
-                  <a href="#" onClick={this.requestConfirmationEmail}>
-                    <Translate content="emailSettings.general.emailUnverifiedPrompt" />
-                  </a>
-                  */}
+                  {(this.state.requestConfirmationEmailStatus === 'idle') && (
+                    <a href="#" onClick={this.requestConfirmationEmail}>
```
eatyourgreens, line 318: This link links back to the current page. I think you meant to use a button here. In a screen reader, or even as a sighted user, I would expect the browser to navigate to a new location after clicking a link. I think the screen reader is more confusing, as this will be announced as a link.

app/pages/settings/email.jsx
eatyourgreens: I forgot to mention: the ‘Verify my email’ button should **not** be part of the live region. That would be weird. Something like this should work without discriminating against disabled people:

```jsx
<button onClick={verifyEmail}>
  Verify your email address
</button>
<span role=‘status’>
/*
  Changes to verification status
  will be announced here. The initial
  status must be an empty string. 
*/
  {verificationRequestStatus}
</span>

File: app/pages/settings/email.jsx (hunk 3 above)
Comment (eatyourgreens) [reply app/pages/settings/email.jsx:318]: See the [WebAIM accessibility cheat sheet](https://webaim.org/resources/htmlcheatsheet/) for more info about accessible links, but the code here would fail an accessibility audit.

File: app/pages/settings/email.jsx (hunk 1 above)
Comment (eatyourgreens) [reply app/pages/settings/email.jsx:337]: The HTML `<output>` element is also supported in all browsers, for live text. 

https://www.scottohara.me/blog/2019/07/10/the-output-element.html

**Comment by eatyourgreens** [comment]
By the way, [406: Not Acceptable](https://httpcats.com/406) is the response when content negotiation fails, so my first thought would be to look at the Accept headers.

File: app/pages/settings/email.jsx [hunk 4]
```diff
@@ -200,8 +204,25 @@ class EmailSettingsPage extends React.Component {
...
+  async requestConfirmationEmail() {
+    try {
+      this.setState({ requestConfirmationEmailStatus: 'posting' });
+      const url = `${apiHost}/users/confirmation`;
+      const bearerToken = await getBearerToken(authClient);
+      const res = await fetch(url, {
+        method: 'POST',
+        mode: 'cors',
+        headers: {
+          accept: 'application/vnd.api+json; version=1',
```
eatyourgreens, line 216:
```suggestion
          Accept: 'application/vnd.api+json; version=1',
```

This might be why content negotiation is failing.

eatyourgreens:
If I go to https://pr-7117.pfe-preview.zooniverse.org/settings/email, then change my email address to **a totally new email address**, the page still says my email is verified. Is that a bug that I can exploit in order to get around spam blocks?

I tested this by switching between an address on a personal domain that I own, and my work address on the `ox.ac.uk` domain.

app/pages/settings/email.jsx
eatyourgreens: Actually, looking at the auth client, these are the headers that it sends. Note that the Accept header is just `application/json`. Your Accept header here is requesting a JSON-API resource, but I don't think the auth APIs follow the JSON-API specs.

https://github.com/zooniverse/panoptes-javascript-client/blob/8157794dfacfbc1f5d41c5730b2f47aae6fc013a/lib/config.js#L75-L78

They are exported as `config.jsonHeaders`.

eatyourgreens:
Here's a weird bug (particularly if I'm using a screen reader, or colour blind, so don't see the change in colour of the input label.) When I delete my email address completely, I'm told the empty box is both valid and verified.


https://github.com/zooniverse/Panoptes-Front-End/assets/59547/6937ecde-04a9-44ff-ba37-0b878517f350

app/layout/account-bar.jsx
```diff
@@ -62,7 +62,7 @@ export default class AccountBar extends React.Component {
         messageCount: unread_count,
         unread: unread_messages
       });
-    });
+    }).catch((err) => { console.error(err) });
```
shaunanoordin, line 65: Whoops this was left in by accident because the "you're not allowed in this conversation" errors kept bugging me. This should be a separate PR, to be honest.

shaunanoordin:
## PR Update

Update: the **resend confirmation email** link should now be fully functional!

- 406 error has been resolved by actually sending the correct request format.
  - To be specific, the solution was: 1. POST to the correct URL, 2. header should only include Content-Type and charset, 3. body/data should NOT be in JSON format, 4. NOTHING ELSE
- Reviewing [Zach's example,](https://github.com/zooniverse/Panoptes-Front-End/issues/7032#issuecomment-1912527641) I realised I was overthinking what actually needed to be sent. As a result, all code require getBearerToken() has been stripped out, and all attempts to make the POST request resemble a PJC JSON API client request have been removed.

### Status

This PR is ready for review.

A follow up is required to address the accessibility feedback.

lcjohnso:
Great work, Shaun!  I can confirm this solution works when tested for an unconfirmed user account.

I agree with Jim's commend regarding link vs. button. When "Resend Confirmation Email" appears as a clickable link -- as Sean and I recommend -- it is a bit odd to see "https://pr-7117.pfe-preview.zooniverse.org/settings/email#" as the link destination. Now that I've used the feature, I'd prefer a button so I don't expect to navigate away to a different page.

Sorry for the indecision! @seanmiller26 -- are you OK with a button instead of a link?

seanmiller26:
Sure, that sounds good. Let's use this style of button design - #ffffff fill with #005d69 outline, 4pt rounded corners, 16pt font.

app/pages/settings/email.jsx
shaunanoordin: 👍 All status messages have now been placed in an `<output>`

shaunanoordin:
## PR Update

Thanks for the feedback, everyone!

- Accessibility updates:
  - Status messages are now inside `<output>`
  - "Resend email" link replaced with button. (@seanmiller26 see design notes below)

Design note 1: button looks like this. Font size is kept consistent with rest of text, which is... uh... 13.3333px? PFE is weird like that.

image

<details>
<summary>(If you're curious to see what the button looks like with font-size: 16px, click here)</summary>
image
</details>

Design note 2: hover & click has a different background.

image

Design note 3: status text

image

For further discussion, but won't be addressed in this PR:
- ❓ can an email be un-verified by changing it? Does changing an email mean it's still verified?

eatyourgreens:
Yep,once I've verified an account, I can change the email address to any valid email address at all, without the owner of that address being notified that their address is being used. 

A verified Zooniverse account for cliff+random@zooniverse.org

lcjohnso:
re: email notifications -- I've opened https://github.com/zooniverse/panoptes/issues/4351; that is out of scope for this PR and will be resolved separately.

app/pages/settings/email.jsx
```diff
@@ -292,12 +311,22 @@ class EmailSettingsPage extends React.Component {
...
-                  </a>
-                  */}
+                  </button>
+                  {' '}
+                  <output>
+                    {{
+                      'posting': <Translate content="emailSettings.general.emailUnverifiedRequesting" />,
+                      'success': <Translate content="emailSettings.general.emailUnverifiedSuccess" />,
+                      'error': <Translate content="emailSettings.general.emailUnverifiedError" />,
+                    }[this.state.requestConfirmationEmailStatus] || ''}
```
kieftrav, line 328: Clever!

kieftrav (approved):
Works as expected and code looks good!

PR #7124, opened 2024-06-08
refactor(collections): refactor collection search to match monorepo

### LINKED ISSUE #7121: Bad search results while adding a subject to an existing collection
## Expected behavior
When searching for an existing collection to add a subject in it, you start typing any word from the collection name and you get a list of your collections (in a pop-up list) containing the typed characters.
With some collections it seems to work fine. F.e. collection called *Corpses* gets listed after typing the first 3 characters "cor". Collection called *WOW* appears even after typing 2 characters "wo".

The difference seems to be that there's no other collection that would include 'wo' in the name while there's also a collection including 'co'  in the name (called 'COOL') - so I believe it requires 3 characters for the 'Corpses' to sort out whether I search for 'corpses' or 'cool'. I'm surprised it doesn't show me both after typing 'co' though, as it is something I expect as a user. But with other collections, it becomes even more weird.

## Current behavior
I'm having also two collections called *Birds of ...* (there's a project location instead of "..."), and I can also add images into a shared collection created by someone else called *C&S best of birds* that appears listed among my collections. As you can see, all the three collections include words 'birds'.

When searching for one of the first two collections (Birds of ...) and typing 'Birds' into the search box, only the third collection *'C&S best of birds'* appears in the list as if there were no other collections including 'birds' in its name. This is odd! (And I would say buggy.) Only after typing 'Birds of' (8 characters!) I get those two collections listed but I have to write all the 8 characters first. When I only type the first 7 ('Birds o') I get no results. Why it doesn't work the same as with the 'Corpses' where I only need to write 'cor'? I would prefer to see all the 3 collections that include 'birds' in their name listed after typing just 'bir'. This is so irritating. And why it displays 'C&S best of birds' after typing just 'birds' and not the other two that also include the word?

I also wonder if there were 3 collections, one called *'Australian birds - Kalbarri'*, second one called *'Australian birds - Jarrah forest'* and third called *'Juveniles of Australian birds'* how many characters of 'Australian' I would have to type to get any of them listed? (If there was also *'Australian mammals'* then it would require to type the whole 'Australian birds' or 'Australian mammals' to get either of the 4 collections listed?)

## Steps to replicate
Create the same collections I mentioned above, or your owns. You'll need at least 3 collections that include the same word (of at least 5 characters) in their name - two of them will start with that word, the third will have the word after another word(s). Try the same with two shared words in the name (f.e. 'Bird of'). Try it with a word made out of more than 5 characters etc.
Then go either to the classification of an "old" frontend project or to the Talk interface of a subject (image/video). Click on the 'Collect' button under the subject, then use the search box. Type slowly the characters of the word shared in names of the test-collections (f.e. 'Birds') and see what happens.

## Additional information
When adding a subject into an existing collection from the classification interface of a new frontend project, the search gives nice results - I get all three birds collections displayed immediately after typing 'bir'. Sadly, there seems to be a bug - if I type one more character ('bird'), it displays all my collections (Corpses, WOW, COOL, etc.) instead of just the three that actually include 'Birds' in their name. And after typing another one more character 'birds' it gives me the 'No matches found' result. That's very odd. Don't you think? :) Shall I start a new issue for this one?

- **Operating system:** Ubuntu
- **Browser:** Chromium (newest version)

eatyourgreens:
Panoptes full-text search requires at least 3 characters in order to return results. If your search string is longer than 3 characters, `collectionsSearch` passes it to the Panoptes full-text search API. Otherwise, it fetches as many collections as it can then filters for collections that include the search string in the collection name.

- Fixes #7121 for searches that contain 3 characters or less.
- see https://github.com/zooniverse/front-end-monorepo/pull/4867 for the collections search logic in the monorepo.

The Panoptes API returns odd results for search terms that are 4 or 5 characters (not whole words), but the API’s not something that can be fixed here.

https://github.com/zooniverse/Panoptes-Front-End/assets/59547/2307d4ff-00b3-4376-9fb5-0ecc6e60a002

# Required Manual Testing

# Review Checklist

- [ ] Does it work on mobile?

# Optional

### DIFF
--- a/app/collections/searchCollections.js
+++ b/app/collections/searchCollections.js
@@ -1,0 +1,26 @@
+import apiClient from 'panoptes-client/lib/api-client';
+
+function panoptesCollectionsSearch(text) {
+  const query = {
+    page_size: 100,
+    favorite: false,
+    current_user_roles: 'owner,collaborator,contributor'
+  };
+  if (text) {
+    query.search = text;
+  }
+  return apiClient.type('collections').get(query);
+}
+
+export async function collectionsSearch(text = '') {
+  const search = text.trim().toLowerCase();
+  // Search terms of more than 4 characters can use Panoptes full-text search.
+  if (search.length > 3) {
+    return panoptesCollectionsSearch(search);
+  }
+  // Otherwise, filter your collection names by the search term.
+  const allCollections = await panoptesCollectionsSearch();
+  return allCollections.filter(
+    collection => collection.display_name.toLowerCase().includes(search)
+  );
+}
--- a/app/collections/searchCollections.js
+++ b/app/collections/searchCollections.js
@@ -3,6 +3,11 @@ import apiClient from 'panoptes-client/lib/api-client';
 // Tune this value to determine when a search term is long enough to use Panoptes full-text search.
 const MIN_SEARCH_LENGTH = 3;
 
+/**
+ * Query the Panoptes collections search API
+ * @param {string} text - The search term.
+ * @returns {Promise<Array>} - A promise that resolves to an array of collections.
+ */
 function panoptesCollectionsSearch(text) {
   const query = {
     page_size: 100,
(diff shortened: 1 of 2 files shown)


app/collections/searchCollections.js
```diff
@@ -0,0 +1,29 @@
+import apiClient from 'panoptes-client/lib/api-client';
+
+// Tune this value to determine when a search term is long enough to use Panoptes full-text search.
+const MIN_SEARCH_LENGTH = 3;
```
eatyourgreens, line 4: Increasing this value might be a workaround for weird results when the search is 4 or 5 characters.

mcbouslog (approved):
Changes LGTM! Confirmed issue then confirmed these changes locally do indeed improve collections search similar to zooniverse/front-end-monorepo/pull/4867 .

app/collections/searchCollections.js
eatyourgreens: Updating to add that this breaks collections search for Gravity Spy, where there are 2-character collection names.

PR #7220, opened 2024-11-15
[Security] Update @sentry/browser to 7.120.0

shaunanoordin:
## PR Overview

Staging branch URL: https://pr-7220.pfe-preview.zooniverse.org/projects

See [Dependabot notice](https://github.com/zooniverse/Panoptes-Front-End/security/dependabot/92)

This PR updates `@sentry/browser` to address some security advice provided by Dependabot.

- Dependabot previously tried to update sentry-browser to v8, but that broke PFE. #7206 
- We reverted that change. #7209 
- This PR updates `@sentry/browser` to the latest v7 iteration instead, which seems to work on local and I _think_ will address Dependabot's concerns.
- There's a longer [v7 to v8 migration guide,](https://docs.sentry.io/platforms/javascript/migration/v7-to-v8/) but we won't invest dev time into that until it's necessary.

### Testing

Open the Projects Page in the staging branch. If the page loads & renders properly, this branch should be fine.

### Status

Ready for review.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8,7 +8,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@babel/runtime": "~7.26.0",
-        "@sentry/browser": "~7.100.1",
+        "@sentry/browser": "~7.120.0",
         "animated-scrollto": "~1.1.0",
         "chart.js": "~4.4.6",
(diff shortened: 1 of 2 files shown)


goplayoutside3 (approved):
Thanks Shaun!

PR #7282, opened 2025-03-20
Edit GitHub Actions workflow ymls to node v20.18

lcjohnso:
Replicating fix from https://github.com/zooniverse/front-end-monorepo/pull/6773 and https://github.com/zooniverse/front-end-monorepo/pull/6773 for failing tests and builds due to node version 20.19 incompatibility.

Staging branch URL: https://pr-7282.pfe-preview.zooniverse.org/projects


goplayoutside3 (approved):

eatyourgreens:
A better fix here, in terms of long term support and maintenance, would have been to add the mssing `type: commonjs` to `package.json`. See #7322.

PR #7298, opened 2025-04-17
Webpack: serve bundled image files from /lab subpath

### LINKED ISSUE #7295: PFE Production: Webpack-bundled image files not appearing
## UI Issue (but actually a Build Issue)

Related: Pages Editor (aka Workflow Editor) project.
Sorta kinda follows: zooniverse/static#392

Surprise! It turns out that on PFE _production_ specifically, all _image files_ bundled by Webpack _won't_ appear.

For example, compare this page on production[^1]...
Image

...with the similar page on PFE _staging_ [^2]

Image

The zooniverse-word-white.png (zooniverse logo, on top-right) appears on the staging page, but not the production page.

Analysis:
- In both cases, the PNG image was bundled by Webpack to a path like `(root)/5cf7528a3a51066295b1.png`
- Unfortunately, while there's a redirect path for `(root)/(filename).png` for PFE staging [(link: nginx file)](https://github.com/zooniverse/static/blob/3a722064c0c025b7050288d7438b630d49c36792/sites/star.pfe-preview.zooniverse.org.conf#L6), the redirect paths for PFE production [(link: nginx file)](https://github.com/zooniverse/static/blob/3a722064c0c025b7050288d7438b630d49c36792/nginx-pfe-redirects.conf) only work for specific subfolders, e.g. `(root)/lab/(filename).png`
  - I assume this is a legit choice, because `www.zooniverse.org` has pretty complicated routing since it houses both FEM and PFE servers(? instances? containers? thingamajigs?)

### Dev Notes

The fix then would be to modify the **[webpack.build.js](https://github.com/zooniverse/Panoptes-Front-End/blob/b719fd4/webpack.build.js)** file so that all assets/resources file are output to, say, `/lab` instead of `/`.

### Stat

Fixed required for the Pages Editor project.

Confirmed with commit [b719fd4](https://github.com/zooniverse/Panoptes-Front-End/commit/b719fd422513f501a5ac01a947976677f89ebab8)

[^1]: https://www.zooniverse.org/lab/1982/workflows/editor/3711?env=staging
[^2]: https://pr-7269.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging , but note that this staging preview will be deleted after that PR is merged.

shaunanoordin:
## PR Overview

Closes #7295

Staging branch URL: https://pr-7298.pfe-preview.zooniverse.org/lab/1982/workflows/editor/3711?env=staging

This PR updates the build process so that bundled image files are served from the `/lab` subpath, instead of the more sensible `/` root. Why change the path? Because we've configured the `www.zooniverse.org` domain to serve PFE code and FEM code on different paths, and the `/` root path is now the domain of FEM. Why the 'lab' subpath? Because in practice the only part of PFE that still uses bundled image files is the new Pages Editor.

- Changes are made to webpack.build.js (used for `www.zooniverse.org`) and webpack.dev.js (used on our local dev servers, i.e. on npm start)
- The changed subpath works properly on local, will work properly on the PFE staging link, but we won't know if the full production build will work properly until this PR is merged and deployed. ⚠️ 

### Testing

- Open the staging branch URL above ☝️ 
- On that Pages Editor page, you should see images (the Zooniverse word logo on the top right, and Task icons) served properly.
- You should not see broken images _anywhere_ on the Zooniverse.

### Status

Ready for review.

### DIFF
--- a/webpack.build.js
+++ b/webpack.build.js
@@ -32,6 +32,15 @@ module.exports = {
     path: path.join(__dirname, '/dist'),
     filename: '[name]-[chunkhash].js',
     chunkFilename: '[name]-[chunkhash].js',
+    // Image assets will be placed in the '/lab' subpath.
+    // This is due to how the www.zooniverse.org domain maps PFE and FEM to
+    // different subpaths; the root '/' path isn't set to serve PFE code, let
+    // alone PFE-related image assets.
+    // See https://github.com/zooniverse/static/blob/632be4f/nginx-pfe-redirects.conf
+    // The /lab subfolder was chosen since bundled image assets are primarily
+    // used ONLY by the Pages Editor system.
+    // See https://github.com/zooniverse/Panoptes-Front-End/issues/7295
+    assetModuleFilename: 'lab/[hash][ext][query]',
   },
   plugins: [
     new webpack.ProvidePlugin({
--- a/webpack.dev.js
+++ b/webpack.dev.js
@@ -28,6 +28,15 @@ var config = {
     publicPath: '/',
     path: path.join(__dirname, '/dist'),
     filename: '[name].js',
+    // Image assets will be placed in the '/lab' subpath.
+    // This is due to how the www.zooniverse.org domain maps PFE and FEM to
+    // different subpaths; the root '/' path isn't set to serve PFE code, let
+    // alone PFE-related image assets.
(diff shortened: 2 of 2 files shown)


goplayoutside3 (approved):
Looks good 👍

PR #7299, opened 2025-04-17
PFE Admin: add warnings to Manage Users -> (Semi-)Recent Classifications

shaunanoordin:
## PR Overview

Staging branch URL: https://pr-{NUMBER}.pfe-preview.zooniverse.org/admin/users/107?env=production

For whatever reason, our Admin page's Manage User section's "Recent Classifications" system is no longer working as intended. This PR adds a warning to our admin page, to remind us not to use it for debugging.

image

- Here's what no longer works: 1. searching/filtering by classification ID, and 2. recent classifications being listed by the most recent. (Which basically means it's less "recent classifications" and more "random jumble of classifications")
- I'm tempted to delete the Recent Classifications code altogether, but I imagine this could be fixed at some point, maybe, possibly.

### Status

Ready for review. Low priority.

Credit to Michelle for figuring out what was wrong. See [Slack](https://zooniverse.slack.com/archives/C06DCM0V9/p1744900853598099?thread_ts=1744684185.651289&cid=C06DCM0V9) for more info on the API side of things.

### DIFF
--- a/app/pages/admin/user-settings.jsx
+++ b/app/pages/admin/user-settings.jsx
@@ -112,7 +112,8 @@ class UserSettings extends Component {
           </ul>
         </details>
         <details>
-          <summary>Recent classifications {this.state.classifications.length}</summary>
+          <summary>Semi-recent classifications {this.state.classifications.length}</summary>
+          <p>⚠️ WARNING: classifications are <b>not correctly sorted</b> by most recently created. Order of classifications is semi-random.</p>
           <form onSubmit={this.updateSubjectID}>
             <label for="subjectId">Filter by subject ID: </label>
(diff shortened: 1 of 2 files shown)


yuenmichelle1:
> I'm tempted to delete the Recent Classifications code altogether, but I imagine this could be fixed at some point, maybe, possibly.

I'll put a fix into to Panoptes later on this coming up week so that the sort by `created_at` will work as expected. As far as I can tell, it's a very small change on panoptes to get that to work again (? unsure if it was working as expected at all before ). 


Filtering by subjects looks like it's working ? (though I only looked at a small sample set while testing) . But happy to be wrong and I'm ok to add some sort of warning  like you have here.

PR #7306, opened 2025-05-09
Add FEM redirect for cmc/mind-mapper

kieftrav:
Staging branch URL: https://pr-7306.pfe-preview.zooniverse.org


zwolf (approved):

zwolf (commented):
If this project is the renamed `cmcbrain/axonal-pathfinders`, then that line can be removed here instead of keeping both.

PR #7395, opened 2026-01-15
Edit the subject file-type instructions in the lab

goplayoutside3:
Staging branch URL: https://pr-7395.pfe-preview.zooniverse.org/lab

Can be merged/deployed anytime before Jan 27.

## How To Review
- See any /subject-sets tab for any project's lab. The updated text should display.

### DIFF
--- a/app/pages/admin/project-status/experimental-features.jsx
+++ b/app/pages/admin/project-status/experimental-features.jsx
@@ -77,6 +77,7 @@ class ExperimentalFeatures extends Component {
     return (
       <div className='project-status__section'>
         <h4>Experimental Features</h4>
+        <p>See the <a href='https://github.com/zooniverse/how-to-zooniverse/blob/master/ProjectLifecycle/experimental_tools.md'>how-to-zooniverse documentation</a> and its latest <a href='https://github.com/zooniverse/how-to-zooniverse/pull/324'>Pull Request</a> for more information</p>
         <AutoSave resource={this.props.project}>
           <div className='project-status__section-table'>
             {experimentalFeatures.map((feature) => (
--- a/app/pages/lab-fem/workflow.jsx
+++ b/app/pages/lab-fem/workflow.jsx
@@ -366,34 +366,6 @@ class EditWorkflowPage extends Component {
 
             <hr />
 
-            <div className="deprecated">
-              <AutoSave resource={this.props.workflow}>
-              <span className="form-label">Set annotation persistence</span><br />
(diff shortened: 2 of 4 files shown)


shaunanoordin (changes requested):
## PR Review

Affects: Project Builder -> Subject Sets -> Upload Subjects

The intent of this PR is to change the file types that the file uploader in the Edit Subject Set page supports. This mostly changes _content_ and _doesn't_ enforce any behaviour changes to the subject upload process. 

❗ As of ddd6b9f5aa734a922014fdfb404a7770a9c23036, I think there's a bug that needs to be fixed. The current code change...

```
Subject images can be up to {MAX_FILE_SIZE / 1024}KB and any of: {<span key={ext}><code>{ext}</code>{', ' if VALID_FEM_EXTENSIONS[i + 1]?}</span> for ext, i in VALID_SUBJECT_EXTENSIONS}{' '}
```

...only removes the commas `,` after ".m4a" (see screenshot below) and doesn't actually remove the now-unsupported file types (svg, mp3, m4a) from the list.

Image

❓  Also, I'm not sure if we wanted to enforce behaviour changes? If someone ignores the instructions and uploads an mp3 file anyway, do we want to quietly allow them to do so? (e.g. to support PFE backwards compatibility/for projects that still need to use PFE?)

### Status

Changes requested.

(Required)
- [ ] The "Subject images can up to X KB and any of..." line should be updated so that remaining instance of VALID_SUBJECT_EXTENSIONS is also replaced with VALID_FEM_EXTENSIONS instead.

(Optional/Conditional/) (Depends On Intent) 
- If (and only if) we wanted to enforce "users shouldn't be able to upload mp3, m4a, or svg files", then we should also change _findFilesInMetadata() to use VALID_FEM_EXTENSIONS.

goplayoutside3:
Thank you for the review @shaunanoordin and catching the incorrect array! I'll push a change to fix it.

> If (and only if) we wanted to enforce "users shouldn't be able to upload mp3, m4a, or svg files", then we should also change _findFilesInMetadata() to use VALID_FEM_EXTENSIONS.

There are Zooniverse projects with audio subjects we want to continue supporting via the PFE classifier, and we want them to be able to upload subjects via the project builder. Limiting upload based on subject file type would require reading the `pfeLab` flag in this section of the project builder (complicated since it's .cjsx code), and we'd also need to consider the other command line tools researchers use to upload subjects.

Support for .svg and audio subjects in FEM's classifier is coming, so this PR is a temporary change in the project builder instructions as requested by the leadership team to encourage researchers to contact us if they're unfamiliar with rules around supported subject file types.

PR #7400, opened 2026-01-21
Replace internal routes to /projects page with redirect to the static proxy

goplayoutside3:
Staging branch URL: https://pr-7400.pfe-preview.zooniverse.org/lab

To be merged and deployed Jan 27.

## Describe your changes
- Replace react-router Links to /projects with an anchor tag to www
- Config the /projects Route in router.jsx to redirect to the static proxy

## How to review
- Anchor tags to Projects in the header and footer should go to www
- Check out https://local.zooniverse.org:3735/rubin. The link to Projects should be to www
- Same with https://local.zooniverse.org:3735/accounts when you're signed in
- Try to visit https://local.zooniverse.org:3735/projects and you'll be redirected
- When signed out, https://local.zooniverse.org:3735/lab has a link to Projects too

### DIFF
--- a/app/pages/reset-password/reset-password.jsx
+++ b/app/pages/reset-password/reset-password.jsx
@@ -82,7 +82,6 @@ class ResetPasswordPage extends React.Component {
           resetSuccess: true,
         });
         alert(resolve => <LoginDialog onSuccess={resolve} />);
-        this.context.router.push('/projects');
       })
       .catch((error) => {
         this.setState({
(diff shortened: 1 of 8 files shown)


app/pages/reset-password/reset-password.jsx
```diff
@@ -82,7 +82,6 @@ class ResetPasswordPage extends React.Component {
           resetSuccess: true,
         });
         alert(resolve => <LoginDialog onSuccess={resolve} />);
-        this.context.router.push('/projects');
```
goplayoutside3: I'm not sure I've ever encountered this feature 🤔 Since /projects is now an "external" link to the static proxy, I think it's better not to push or replace the window location here.

PR #7392, opened 2026-01-15
Change the default lab to fem-lab

goplayoutside3:
Staging branch URL: https://pr-7392.pfe-preview.zooniverse.org/lab

For deploy on Jan 27

## Changes
- `isThisProjectUsingFEMLab()` is the source of truth for whether to show FEM-specific things in the lab. This PR changes that function to always return `true` unless a project has the `pfeLab` experimental flag, or `pfeLab = true` in the search query.
- A project's admin page now includes `pfeLab` checkbox in replacement of `femLab`.
- Update README and other documentation.

# How to Review

You can use your own test project or this one to test https://local.zooniverse.org:3735/lab/1928

PR #7409, opened 2026-02-02
Revise femLab v pfeLab override behavior

lcjohnso:
Staging branch URL: https://pr-7409.pfe-preview.zooniverse.org

Builds from and revises outcomes from #7392.

As part of switcharoo, the primary goals w.r.t. femLab and pfeLab behavior was
1. Change default lab to femLab for (almost) all projects
2. Manage project exceptions that use pfeLab along with redirect routing, via a _single setting_: the `PFE_SLUGS` variable in [app/slugList.js](https://github.com/zooniverse/Panoptes-Front-End/blob/main/app/slugList.js).

Combining the interface for defining redirects & defining pfe/fem lab preference leads to a desired simplification in process when the Zoo team wants to convert PFE projects to FEM in the future (i.e., removing them from the exceptions list). Also, the femLab/pfeLab experimental tools that previously governed pfe/fem lab preference are no longer needed.

This PR aims to handle an edge case: overriding default lab preferences. Certain situations call for the Zoo team and research teams to be able to manually override the lab preference. Example: in preparation for a move from PFE to FEM, it may be helpful to access the FEM lab even if the default rules specify use of the PFE lab -- e.g., enable an override to femLab even if the project is listed in the list of PFE_SLUGS exceptions.

The current code (as implemented in #7392) allows negation of the default `femLab=true` setting via URL query param for most projects, but doesn't allow negation of `femLab=false` for PFE exception projects. That is corrected here.

Also: we confirm that neither the femLab or pfeLab experimental tool settings are needed anymore, so femLab is removed. Documentation is also updated.

## For Review
- Test that for typical projects (those that default to femLab=true), the pfeLab workflow editor can be accessed via `?femLab=false` query param.
- Test that for PFE exception projects (those that default to femLab=false; e.g., https://www.zooniverse.org/projects/gaia-zooniverse/gaia-vari), the femLab workflow editor can be accessed via `?femLab=true` query param.

A convenient indicator of femLab true vs. false: whether the "Set as Default" option is included on the workflow editor page.

### DIFF
--- a/app/pages/admin/project-status/experimental-features.jsx
+++ b/app/pages/admin/project-status/experimental-features.jsx
@@ -13,7 +13,6 @@ const experimentalFeatures = [
   'enable subject flags',
   'expert comparison summary',
   'fan',
-  'femLab',  // Indicates that the Project should use the FEM-compatible Project Builder (lab) pages
   'freehandLine',
   'freehandSegmentLine',
   'freehandSegmentShape',


shaunanoordin (approved):
## PR Review

Affects: Project Builder -> Edit Workflow page
Also see: #7392 and #7397 

This PR re-introduces the ability to force either "PFE Lab" mode or "FEM Lab" mode on the Edit Workflow page, by adding `?femLab=...` to the URL's query parameters.

- For PFE projects[^1], the (now standard) FEM Lab mode can be enabled with `?femLab=true`
- For FEM projects[^2], the old PFE Lab mode can be enabled with `?femLab=false`[^3]

ALSO:
- the `femLab` experimental tool has been removed from the admin pages.

This functionality is intended primarily for admin use only. Code changes look good and testing looks good.

[^1]: which uses PFE Lab by default.
[^2]: which uses FEM Lab by default.
[^3]: prior to 7392, this override used to be `?pfeLab=true`

### Testing

Tested locally with macOS + Chrome 144.

- Selecting a personal test project, I either...
  - Add it to the list of PFE projects in `slugList.js` to ensure it's a PFE project
  - Or remove it from the list to ensure it's an FEM project.
- Editing [a random test workflow in that test project,](https://local.zooniverse.org:3735/lab/1982/workflows/3711),
  - I can confirm that by default it uses the correct PFE Lab/FEM Lab version of the Edit Workflow page.
  - I can confirm that I can override the PFE Lab/FEM Lab version via `?femLab=true` or `?femLab=false`

### Status

LGTM 👍

